PR #13, opened 2020-03-03
Dataplanes tags error fix

bloqhead:
This was caused when the number of tags exceeded the number of colors available in an array for the tags. The tags were originally going to be different colors but instead were simplified to one color. I've removed the unused code and simplified the tag handling in `DataOverview` and this fixes the error [seen here](https://github.com/Kong/kuma/issues/618).

I admittedly overlooked this because none of my test DPs had more than 8 tags.

Now we can have more than 8 tags with no errors: [screenshot](https://share.getcloudapp.com/qGuovYoG).

### DIFF
--- a/src/components/Skeletons/DataOverview.vue
+++ b/src/components/Skeletons/DataOverview.vue
@@ -64,7 +64,6 @@
               <span
                 class="entity-tags__label"
                 :class="`entity-tags__label--${item.label.toLowerCase()}`"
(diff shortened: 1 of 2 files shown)


bloqhead:
@jakubdyszkiewicz can you review when you have a chance? I would like to get a new version compiled.

subnetmarco:
Can we merge this?

PR #14, opened 2020-03-24
added publicPath to Vue config that sets all public assets to be relative

bloqhead:
This addresses a use case where a user proxies the GUI to a folder, e.g. something like `/kuma-cp/`, inside of a Kubernetes container. By default, Vue's asset path is absolute, which makes all assets 404 in this context. This would resolve that.


subnetmarco:
To replicate this issue we put Kuma in front of a Kong Ingress on AKS (Azure) by exposing only the GUI port (not the API port). When connecting to the GUI, the GUI was not able to fetch the static assets.

PR #20, opened 2020-04-21
Dataplane Wizard

bloqhead:
This is the first of 2 stages for the Dataplane Wizard. The first stage being the Kubernetes flow, and the second being Universal.

[Here is a quick video recording of it in action](https://share.getcloudapp.com/7Ku0yYd4)

### To test it:
1. Start up the GUI: `yarn install` (if you haven't already), and `yarn serve`
2. **For Kubernetes:** http://localhost:8080/#/wizard/kubernetes-dataplane / **For Universal:** http://localhost:8080/#/wizard/universal-dataplane
3. Go through the form and test each option available to see conditional handling and display

### Notes:
- The last step does not yet display code for creating a Dataplane. That part is still in progress.
- You will see a panel at the start of this wizard that will detect what environment Kuma is running in, and provide instructions to view the other environment's wizard.
- You'll see a "Debugger" panel in the right sidebar. As noted in the component itself, this is only displayed when running the GUI in development mode. The app runs in dev mode when using `yarn serve` in this repo, but runs in production mode when compiled and running from within Kuma. This panel exists simply to keep track of `v-model` data that will be used to assemble the code output.


bloqhead:
@subnetmarco I've finished the Universal Dataplane wizard but am stuck on what we should include for code in the last step where the user starts their new Dataplane. I am able to successfully create both Service and Gateway Dataplanes from the Wizard.

**To test the Universal DP wizard:**
1. `yarn serve` to run the app
2. Go to http://localhost:8080/#/wizard/universal-dataplane
3. Follow the steps provided

In the last step I'm currently including this code to start the Dataplane:
```
kuma-dp run \
  --name=your-dataplane-id \
  --mesh=your-mesh \
  --cp-address=http://localhost:5681`
```
But I'm required to create a `dataplane-token` first, so I am referencing [the Dataplane Token](https://kuma.io/docs/0.4.0/documentation/security/#dataplane-token) section in the docs. I don't have a token server or anything running at `http://localhost:5679/tokens` though, so I'm not sure what to do. Please advise.

PR #23, opened 2020-04-28
Native pagination

bloqhead:
All views now use native pagination based on the [new documentation](https://github.com/Kong/kuma-website/pull/179). For use cases where thousands of entities are present, this will make a **huge** improvement on GUI performance.

This has been tested on Universal and I'm working on testing it for Kubernetes. Please feel free to do the same and report back with any issues.

## Tasks
- [x] Pagination testing in Universal mode
- [x] Pagination testing in Kubernetes mode
- [x] Make the offset handler accept both an integer and a string (Kubernetes uses a long string for the offset, while Universal uses an integer -- my helper now properly takes both into account)

### Some notes
- The tables in each view are limited to 6 items per page. This can be adjusted accordingly
- I've disabled mock endpoints in development mode (via [env.development](https://github.com/Kong/kuma-gui/blob/master/.env.development)) so that we can test accurately with real data running in Kuma
- On the Global Overview (accessed at `http://localhost:8080/#/overview`) I've improved the list of meshes at the bottom to include native pagination as well


bloqhead:
@jakubdyszkiewicz @lobkovilya I've tested the pagination on both Kubernetes and Universal but would like some more eyes on it to ensure it works properly. Can you both test it out? I had to fix the offset handling to work in both environments.

@subnetmarco when I go to the [details page](https://app.fossa.com/projects/git%2Bgithub.com%2FKong%2Fkuma-gui/refs/branch/master/fecdbb58a2d654af1313f538f019e0e11916fae3/preview) for the License Compliance check, it says ["Maintainer Access Required"](https://share.getcloudapp.com/L1u7NoDn) -- do I needed to be added on the Fossa website?

lobkovilya:
I played with it a bit. I don't know whether this is a bug or an expected behavior:
1. Run 7 services in order to create 7 dataplanes
2. Go to Dataplanes tab, page 2
3. Delete couple dataplanes that located on the page 1
4. Click `Refresh`

Actual: I see the second page with 1 element.
Expected: See only the first page with 5 elements

`Refresh` doesn't update how entities paginated, but maybe that's fine

bloqhead:
@lobkovilya I'll see if I can recreate this scenario locally. I think on the refresh, I have to clear out the array that I pushed the `next` string into. It's weird that it would still go to a page though. The method for handling the API query will always fetch all of that data fresh. I think it's a simple fix but I'll have to test.

bloqhead:
@lobkovilya so I'm thinking more about this and I don't think there's a way for me to tell when the data has changed when using the Refresh control. In order to properly handle the previous control, I have to store each `next` offset value in an array before going to the next page so that I can properly reference in when going backwards. When you use the Refresh control, I don't modify the array where those offset values are because I don't want the user to lose their place. I could clear out the array when refreshing, but that would put the user back on page 1 which could lead to a frustrating user experience.

PR #40, opened 2020-07-01
Newsletter Form

bloqhead:
This PR adds a newsletter signup form to the Overview page. 

## ⚠️ Caveats
The current newsletter form handler on Salesforce has specific redirect URLs for both success and error responses. This means if we want this to work properly, we have to create a brand new handler just for the GUI that has custom success and error redirect URLs.

**The additional caveat:** if the user sets up Kuma to serve the GUI from a different port other than the default `5683`, the redirect URLs will be rendered useless.

### Why can't we submit the form with AJAX?
Salesforce's Pardot form handlers don't support AJAX. Trying to submit via AJAX results in CORS issues in the browser and and completely blocks the form submission.

---

## Revised approach
I'm going to create a proxy server in Node that simply does the form requests for us and allows AJAX. This will then carry over to the website so that we don't have to leverage SalesForce's weird redirect arrangement.


bloqhead:
Because of the heavy limitations to Salesforce Pardot form handlers, this will not work.

1. Pardot form handlers do not allow any sort of AJAX or even JSONP approaches. Instead, we would have to leverage the success and error redirects that are set within the form handler itself on the Pardot admin panel.
2. Redirects will not work in the context of the GUI because redirects require an absolute URL to return to once the form is submitted or returns an error. Because the GUI can be configured to run from a user-defined IP, this is not feasible and there is no way to return the user to the referring URL.
3. Salesforce is not doing any sort of CORS management on requests that go to their Pardot form handlers. So a POST request that works from within a tool like [Insomnia](https://insomnia.rest/), will be rejected entirely when running in other hosted environments.

PR #54, opened 2020-08-11
fix(*) API server URL handling

nickolaev:
Signed-off-by: Nikolay Nikolaev <nikolay.nikolaev@konghq.com>


src/main.js
```diff
@@ -111,6 +110,13 @@ function SETUP_VUE_APP () {
       /**
        * Always check the API URL and set it accordingly for the app to access.
        */
+      var apiUrl = response.data.guiServer.apiServerUrl
```
bloqhead, line 113: I would change `var` to `let` but this is minor.

PR #74, opened 2020-10-11
Added warning icon to show about rejected response on dataplanes

tharun208:
## Implementations
 * Added warning ⚠️  in total Updates row column to show about rejected updates from `envoy`.
 * Used `KoolTip` to show the warning message along with the total number of `rejections`.

Fixes: https://github.com/kumahq/kuma/issues/1056


bloqhead (approved):
@tharun208 this is a great addition!

My only feedback is that I wish we could make the "X rejections" in the KoolTip bold but that component doesn't accept HTML (maybe a PR for another day on Kongponents).

tharun208:
thanks, @bloqhead, Will happy to raise a PR on Kongponents. So, that we can show the rejections in warning color.

PR #78, opened 2020-10-15
Used KPopover component to highlight total rejections

tharun208:
Used KPopover component to highlight total rejections.


src/components/Skeletons/DataOverview.vue
```diff
@@ -103,17 +103,18 @@
...
+                  trigger="hover"
                 >
                   <KIcon
                     class="kong-icon--centered"
                     color="var(--yellow-300)"
                     icon="warning"
                     style="margin:auto"
                   />
-                </KoolTip>
+                  <div slot="content"><b>{{ row.totalRejectedUpdates }} </b> rejections. Some configuration changes were rejected by Envoy. The most common reason is that the configuration generated by the control plane may be invalid. Check Kuma DP sidecar logs for more details.</div>
```
bloqhead, line 116: I think this component works better! Some minor feedback:

- I think it would be helpful if we wrap the count and "rejections" in the bold tag too, like this: `<b>{{ row.totalRejectedUpdates }} rejections</b>`
- Maybe have `rejections` as `rejection(s)` instead to account for a single item or multiple

Please let me know your thoughts! Good work!

tharun208: agreed and will go with the above 👍

jakubdyszkiewicz (approved):

PR #82, opened 2020-11-03
feat(ingress) add columns with coordinates

jakubdyszkiewicz:
Removed Mesh column since Ingress should be in default mesh anyways and works with many meshes. Added public address and port columns

### DIFF
--- a/src/views/Entities/IngressDataplanes.vue
+++ b/src/views/Entities/IngressDataplanes.vue
@@ -200,9 +200,10 @@ export default {
           { key: 'actions', hideLabel: true },
           { label: 'Status', key: 'status' },
           { label: 'Name', key: 'name' },
-          { label: 'Mesh', key: 'mesh' },
           { label: 'Type', key: 'type' },
           { label: 'Tags', key: 'tags' },
+          { label: 'Public address', key: 'publicAddress' },
+          { label: 'Public port', key: 'publicPort' },
           { label: 'Last Connected', key: 'lastConnected' },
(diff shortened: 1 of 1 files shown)


src/views/Entities/IngressDataplanes.vue
```diff
@@ -541,6 +542,8 @@ export default {
               lastUpdated: lastUpdated,
               totalUpdates: totalUpdates,
               totalRejectedUpdates: totalRejectedUpdates,
+              publicAddress: (response.dataplane.networking.ingress.publicAddress || null),
+              publicPort: (response.dataplane.networking.ingress.publicPort || null),
```
nickolaev, line 546: I don't understand the language here, but is there a case where ingress can be `null` and we get dereference issues?

nickolaev (approved):

src/views/Entities/IngressDataplanes.vue
jakubdyszkiewicz: Ingress won't be null, because we are asking API for only ingress data planes (which are defined by having an ingress part).
The difference between
```
response.dataplane.networking.ingress.publicAddress
```
and
```
response.dataplane.networking.ingress.publicAddress || null
```
is that the first one will return `undefined` and the second one `null`. I think Vue when rendering is fine with both, but I see that we nullify all the values

bloqhead: Vue will handle the `undefined` but I like to nullify the items out of paranoia and consistency. It's probably overkill but it ensures I know exactly what the fallback will be.

PR #84, opened 2020-11-09
chore(menu): rename all data plane proxies menu entry

subnetmarco:


bloqhead:
I like the simplicity of this. That menu item is kind of long.

bloqhead (approved):

PR #85, opened 2020-11-10
fix(Dockerfile)

tharun208:
`Dockerfile` is outdated and fixed it by building from source. And also removed `docker-compose.yml` since `kuma-cp` itself comes with `gui`.


nickolaev:
Hmm why do we need this?

tharun208:
@nickolaev, Not a needed improvement since `gui` is served by `kuma-cp`. The old dockerfile is outdated. so, raised this to make consistent with the current context.

bloqhead:
Thanks for doing this @tharun208. That Docker file was from a previous project the GUI was modeled after and I just hadn't had time yet to revise it. I like the idea of an environment where we can run Kuma and the GUI from this repo in one contained setup. I'll find some time to review this and try it out.

PR #90, opened 2020-11-18
fix kuma-cp dp server address during dataplane onboarding

tharun208:
Fixes https://github.com/kumahq/kuma/issues/1167


jakubdyszkiewicz:
It has a disadvantage that when Kuma is deployed somewhere in the network, the URL will still be `https://localhost:5678`. We can do two things
1) keep always `https://localhost:5678` and say to a user to replace it with their value
2) Extract the hostname from which the user has connected to the GUI (Host header) and then use it to replace the port. There is a high chance that this hostname is a routable hostname available in the network.

jakubdyszkiewicz (approved):

PR #92, opened 2020-11-18
fix: display dp tags

jakubdyszkiewicz:
I noticed that tags in Ingress tab are broken so I looked into algorithm and noticed a couple of problems
* Ingress does not have tags in `networking.ingress.tags`, we should just list from the inbounds
* We cannot just take `inbound[0].tags` because they might be more tags
* There were two algorithms to display tags spread across 3 files, I created one function in `src/dataplane.js`

Some additional small stuff
* In ingress there should be no Certificate Insights in overview
* restored this one https://github.com/kumahq/kuma-gui/pull/82 because it was lost for some reason

### DIFF
--- a/src/dataplane.js
+++ b/src/dataplane.js
@@ -1,0 +1,55 @@
+/*
+dpTags takes a Dataplane received from backend and construct the list of tags in form of array of objects with label and value.
+It flattens common tags so we don't display them twice.
+
+Example:
+
+type: Dataplane
+mesh: default
+name: cluster-1.ingress-01
+networking:
+  inbound:
+    - port: 1234
+      tags:
+        kuma.io/service: backend
+        version: 1
+    - port: 1235
+      tags:
+        kuma.io/service: backend-api
+        version: 1
+
+Will produce:
+[
+  { label: 'kuma.io/service', value: 'backend'},
+  { label: 'kuma.io/service', value: 'backend-api'},
+  { label: 'version', value: '1'},
+]
+ */
+export function dpTags (dataplane) {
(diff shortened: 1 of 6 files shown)


src/dataplane.js
```diff
@@ -0,0 +1,55 @@
...
+  { label: 'kuma.io/service', value: 'backend-api'},
+  { label: 'version', value: '1'},
+]
+ */
+export function dpTags (dataplane) {
+  let tags = []
+
+  const inbounds = dataplane.networking.inbound || null
+  if (inbounds) {
+    tags = inbounds.map(inbound => inbound.tags)
```
lobkovilya, line 33: I think you can get rid of the first `map`:
```js
tags = inbounds.flatMap(inbound => Object.entries(inbound.tags))
```

src/dataplane.js
```diff
@@ -0,0 +1,55 @@
...
+  // gateway data plane has no inbounds, but has tags embedded in gateway branch
+  const gateway = dataplane.networking.gateway || null
+  if (gateway) {
+    tags = Object.entries(gateway.tags)
+      .map(([key, value]) => key + '=' + value)
+  }
+
+  tags = Array.from(new Set(tags)) // remove duplicates
+
+  return tags.map(function (tagPair) {
```
lobkovilya, line 47: Maybe we can do it in the same fashion:
```js
return tags.map(pair => pair.split('=')).map(([k,v]) => ({label:k, value: v}))
```

lobkovilya (approved):

PR #94, opened 2020-11-19
fix(univeral) dataplane onboarding

### LINKED ISSUE #93: Dataplane Wizard should promote the new Dataplane flow
Dataplane Wizard still says that we need to apply a Dataplane first and then start kuma-dp. We changed the flow that you can pass a Dataplane definition when running kuma-dp and the wizard should promote this approach

tharun208:
Fixed universal data plane onboarding as per recent changes to `kuma-dp`(https://github.com/kumahq/kuma/pull/999)

Fixes #93

### DIFF
--- a/src/views/Wizard/views/DataplaneUniversal.vue
+++ b/src/views/Wizard/views/DataplaneUniversal.vue
@@ -270,40 +270,38 @@
             </HelperTooltip>
           </FormFragment>
         </template>
+
         <template slot="complete">
           <div v-if="validate.meshName">
-            <div v-if="hideScannerSiblings === false">
-              <h3>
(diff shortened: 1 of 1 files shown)


jakubdyszkiewicz:
image

Merging for now since it is better than what we have right now, but I think there should be an extra step with putting this one
```
type: Dataplane
mesh: default
name: test-h9akdb
networking: 
  address: 127.0.0.1
  inbound: 
    - port: 20100
      servicePort: 20101
      serviceAddress: 127.0.0.1
      tags: 
        kuma.io/service: test
        kuma.io/protocol: tcp
```
into a file. Embedding this into command-line looks is a bit error-prone.

PR #96, opened 2020-11-22
fix dataplane entity views data load

tharun208:
Currently, Kuma rest API not supports filtering and pagination at the same time. This causes API calls to fail for dp views. 
So, removed pageSize from the API params.


jakubdyszkiewicz:
It does now. Have you checked with master?

tharun208:
closing this as this is fixed by https://github.com/kumahq/kuma/pull/1144

PR #97, opened 2020-11-23
chore(dataplanes) disable config rejected warning

jakubdyszkiewicz:
Unfortunately, we need to temporarily disable the config rejected warning.

When you use Kuma regularly and apply TrafficTracing you will see a config rejected. This is because of Envoy eventually consistency. First, the listener is configured with Tracing, but there is no cluster yet so Envoy rejects the config. But then cluster is sent so eventually listener has no problem anymore. Until we figure it out, we don't want to scare users with unnecessary warnings.


src/components/Skeletons/DataOverview.vue
```diff
@@ -100,22 +100,22 @@
...
-                  <KIcon
-                    class="kong-icon--centered"
-                    color="var(--yellow-300)"
-                    icon="warning"
-                    style="margin:auto"
-                  />
-                  <div slot="content"><b>{{ row.totalRejectedUpdates }} rejection(s) </b>. Some configuration changes were rejected by Envoy. The most common reason is that the configuration generated by the control plane may be invalid. Check {{ $productName }} DP sidecar logs for more details.</div>
-                </KPop>
-              </span>
+<!--              <span-->
```
lobkovilya, line 103: Can we maybe put here a link to the issue on github, to avoid having commented code without any explanation?

nickolaev (approved):

PR #98, opened 2020-11-24
fix(yaml) K8S YAML should contain mesh and namespace

jakubdyszkiewicz:
Two things
* K8S YAML should contain a mesh (if there is a Mesh, some objects are global scoped)
* K8S YAML should contain a namespace if the name is `name.namespace` from Kuma API

### DIFF
--- a/src/components/Skeletons/YamlView.vue
+++ b/src/components/Skeletons/YamlView.vue
@@ -205,7 +205,7 @@ export default {
       const kubernetes = () => {
         const newObj = {}
         const sourceObj = Object.assign({}, this.content)
-        const { name, type, metadata } = sourceObj
+        const { name, mesh, type } = sourceObj
 
         // all other parts of the Kubernetes object have to be placed under `spec`
         const spec = () => {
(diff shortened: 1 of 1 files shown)


src/components/Skeletons/YamlView.vue
```diff
@@ -224,15 +224,24 @@ export default {
...
         // assemble the main part of our object
         newObj.apiVersion = 'kuma.io/v1alpha1'
         newObj.kind = type
-        newObj.metadata = {
-          name: name
+        if (mesh !== undefined) { // mesh can be empty on global scoped objects
+          newObj.mesh = sourceObj.mesh
+        }
+
+        if (name.includes('.')) { // if name from Kuma has '.' it means it's k8s name joined with a namespace by dot
```
nickolaev, line 234: OK will this be working if we introduce the "parametrized' naming of services? Not sure if this aplies though

src/components/Skeletons/YamlView.vue
```diff
@@ -224,15 +224,24 @@ export default {
...
-        // we remove the `name` and `type` because they need to be placed under `metadata`
-        delete sourceObj.type
-        delete sourceObj.name
-
         // assemble the main part of our object
         newObj.apiVersion = 'kuma.io/v1alpha1'
         newObj.kind = type
-        newObj.metadata = {
-          name: name
+        if (mesh !== undefined) { // mesh can be empty on global scoped objects
```
lobkovilya, line 230: If `mesh` is empty, why do we need this field in `newObj`?

src/components/Skeletons/YamlView.vue
```diff
@@ -224,15 +224,24 @@ export default {
...
+          newObj.mesh = sourceObj.mesh
+        }
+
+        if (name.includes('.')) { // if name from Kuma has '.' it means it's k8s name joined with a namespace by dot
+          const [k8sName, namespace] = name.split('.')
```
lobkovilya, line 235: is it okay for multicluster where we sometimes have 2 dots in the name: `<name>.<original_ns>.default`?

src/components/Skeletons/YamlView.vue
jakubdyszkiewicz: it has nothing to do with parametrized naming of services. It's about naming our resources

lobkovilya (approved):

PR #101, opened 2020-12-02
feat(services) Internal and External services in the GUI

jakubdyszkiewicz:
image
image
image

I did not implement `All` because
1) We don't have an endpoint to aggregate ServiceInsights and ExternalServices and support pagination etc.
2) What would be the columns in the list? "Online dataplanes" does not make sense on external services and address does not make sense for internal services

Overall it can be done but with much work which I'd argue give not that much value.

Views are essentially copy paste with renaming with this modification
```
              this.tableData.data = this.tableData.data.map(entity => {
                entity.address = entity.networking.address
                entity.tlsEnabled = entity.networking.tls.enabled ? 'Enabled' : 'Disabled'
                return entity
              })
```

```
              this.tableData.data = this.tableData.data.map(entity => {
                // API can skip the field if there is no stat
                entity.offline = entity.offline || 0
                entity.online = entity.online || 0
                entity.total = entity.total || 0
                entity.totalOnline = `${entity.online} / ${entity.total}`
                return entity
              })
```

to support custom columns in the list

### DIFF
--- a/src/views/Entities/ExternalServices.vue
+++ b/src/views/Entities/ExternalServices.vue
@@ -1,0 +1,374 @@
+<template>
+  <div class="external-services">
+    <FrameSkeleton>
+      <DataOverview
+        :page-size="pageSize"
+        :has-error="hasError"
+        :is-loading="isLoading"
+        :empty-state="empty_state"
+        :display-data-table="true"
+        :table-data="tableData"
+        :table-data-is-empty="tableDataIsEmpty"
+        table-data-function-text="View"
+        table-data-row="name"
+        @tableAction="tableAction"
+        @reloadData="loadData"
+      >
+        <template slot="additionalControls">
+          <KButton
+            v-if="this.$route.query.ns"
+            class="back-button"
+            appearance="primary"
+            size="small"
+            :to="{
+              name: 'external-services'
+            }"
+          >
+            <span class="custom-control-icon">
+              &larr;
+            </span>
+            View All
+          </KButton>
+        </template>
+        <template slot="pagination">
+          <Pagination
+            :has-previous="previous.length > 0"
+            :has-next="hasNext"
+            @next="goToNextPage"
+            @previous="goToPreviousPage"
+          />
+        </template>
+      </DataOverview>
+      <Tabs
+        v-if="isEmpty === false"
+        :has-error="hasError"
+        :is-loading="isLoading"
+        :tabs="tabs"
+        initial-tab-override="overview"
+      >
+        <template slot="tabHeader">
+          <div>
+            <h3>{{ tabGroupTitle }}</h3>
+          </div>
+          <div>
+            <EntityURLControl :url="shareUrl" />
+          </div>
+        </template>
+        <template slot="overview">
+          <LabelList
+            :has-error="entityHasError"
+            :is-loading="entityIsLoading"
+            :is-empty="entityIsEmpty"
+          >
+            <div>
+              <ul>
+                <li
+                  v-for="(val, key) in entity"
+                  :key="key"
+                >
+                  <h4>{{ key }}</h4>
+                  <p>
+                    {{ val }}
+                  </p>
+                </li>
+              </ul>
+            </div>
+          </LabelList>
+        </template>
+        <template slot="yaml">
+          <YamlView
+            lang="yaml"
+            :title="entityOverviewTitle"
+            :has-error="entityHasError"
+            :is-loading="entityIsLoading"
+            :is-empty="entityIsEmpty"
+            :content="rawEntity"
+          />
+        </template>
+      </Tabs>
+    </FrameSkeleton>
+  </div>
+</template>
+
+<script>
+import { getSome, stripTimes } from '@/helpers'
+import EntityURLControl from '@/components/Utils/EntityURLControl'
+import sortEntities from '@/mixins/EntitySorter'
+import FormatForCLI from '@/mixins/FormatForCLI'
+import FrameSkeleton from '@/components/Skeletons/FrameSkeleton'
+import Pagination from '@/components/Pagination'
+import DataOverview from '@/components/Skeletons/DataOverview'
+import Tabs from '@/components/Utils/Tabs'
+import YamlView from '@/components/Skeletons/YamlView'
+import LabelList from '@/components/Utils/LabelList'
+
+export default {
+  name: 'ExternalServices',
+  metaInfo: {
+    title: 'External Services'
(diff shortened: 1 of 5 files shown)


nickolaev (approved):

src/views/Entities/ExternalServices.vue
```diff
@@ -0,0 +1,381 @@
...
+      this.$store.dispatch('updateSelectedTableRow', data.name)
+
+      // load the data into the tabs
+      this.getEntity(data)
+    },
+    loadData () {
+      this.isLoading = true
+
+      const mesh = this.$route.params.mesh || null
+      const query = this.$route.query.ns || null
```
bartsmykla, line 246: Is there a reason, you are assigning `null` values when these elems have falsy values? I don't see in this function strict checking for nulls

src/views/Entities/ExternalServices.vue
```diff
@@ -0,0 +1,381 @@
...
+
+      const params = {
+        size: this.pageSize,
+        offset: this.pageOffset
+      }
+
+      const endpoint = () => {
+        if (mesh === 'all') {
+          return this.$api.getAllExternalServices(params)
+        } else if ((query && query.length) && mesh !== 'all') {
```
bartsmykla, line 256: I think `&& mesh !== 'all'` is redundant here

src/views/Entities/ExternalServices.vue
```diff
@@ -0,0 +1,381 @@
...
+          return this.$api.getExternalService(mesh, query, params)
+        }
+
+        return this.$api.getAllExternalServicesFromMesh(mesh)
+      }
+
+      const getExternalServices = () => {
+        return endpoint()
```
bartsmykla, line 264:
```suggestion
      const getExternalServices = () => endpoint()
```

src/views/Entities/ExternalServices.vue
```diff
@@ -0,0 +1,381 @@
...
+          .then(response => {
+            const items = () => {
+              const r = response
+
+              if ('total' in r) {
+                if (r.total !== 0 && r.items && r.items.length > 0) {
```
bartsmykla, line 270: I prefer more explicit expressions as you provided here with `r.items.length > 0` but in other places I see you are using less type strict `r.items.length` which will check for `truthy` values, and if you are sure of the type of `r.items` (if `r.items` is present), you can use this more concise version (`r.items.length`)

src/views/Entities/InternalServices.vue
```diff
@@ -0,0 +1,383 @@
...
+  },
+  computed: {
+    tabGroupTitle () {
+      const entity = this.entity
+
+      if (entity) {
+        return `Internal Services: ${entity.name}`
+      } else {
+        return null
+      }
```
bartsmykla, line 173:
```suggestion
      if (entity) {
        return `Internal Services: ${entity.name}`
      } 

      return null
```

src/views/Entities/InternalServices.vue
```diff
@@ -0,0 +1,383 @@
...
+    },
```
bartsmykla, line 174: I'm not a huge ternary expression fan, but I can see they are being used in other places, so we can do
```suggestion
    tabGroupTitle () {
      return this.entity ? `Internal Services: ${entity.name}` : null
    },
```
or even if we are not strictly checking for nulls (I remember if you would like to render component in react you culdn't return any falsy value except nulls, so probably vuew is similar in that fashion)
```suggestion
    tabGroupTitle () {
      return this.entity && `Internal Services: ${entity.name}`
    },
```

src/views/Entities/InternalServices.vue
```diff
@@ -0,0 +1,383 @@
...
+      if (entity) {
+        return `Entity Overview for ${entity.name}`
+      } else {
+        return null
+      }
+    },
+    formattedRawEntity () {
+      const entity = this.formatForCLI(this.rawEntity)
+
+      return entity
```
bartsmykla, line 187: Why not just:
```suggestion
      return this.formatForCLI(this.rawEntity)
```

src/views/Entities/InternalServices.vue
```diff
@@ -0,0 +1,383 @@
...
+    },
+    entityOverviewTitle () {
+      const entity = this.entity
+
+      if (entity) {
+        return `Entity Overview for ${entity.name}`
+      } else {
+        return null
+      }
+    },
```
bartsmykla, line 183: Same comment as above

src/views/Entities/InternalServices.vue
```diff
@@ -0,0 +1,383 @@
...
+      const entity = this.formatForCLI(this.rawEntity)
+
+      return entity
+    },
+    shareUrl () {
+      const urlRoot = `${window.location.origin}#`
+      const entity = this.entity
+
+      const shareUrl = () => {
+        if (this.$route.query.ns) {
```
bartsmykla, line 194: Is `this.$route.query` always provided?

src/views/Entities/InternalServices.vue
```diff
@@ -0,0 +1,383 @@
...
+
+      getInternalServices()
+    },
+    getEntity (entity) {
+      this.entityIsLoading = true
+      this.entityIsEmpty = false
+
+      const mesh = this.$route.params.mesh
+
+      if (entity && entity !== null) {
```
bartsmykla, line 347: second part is unnecessary as it's impossible to meet the first expression and not meet the next one

src/views/Entities/InternalServices.vue
```diff
@@ -0,0 +1,383 @@
...
+    },
+    goToNextPage () {
+      this.previous.push(this.pageOffset)
+      this.pageOffset = this.next
+      this.next = null
+
+      this.loadData()
+    },
+    tableAction (ev) {
+      const data = ev
```
bartsmykla, line 230: Why not:
```suggestion
    tableAction (data) {
```

src/views/Entities/ExternalServices.vue
```diff
@@ -0,0 +1,381 @@
...
+              })
+
+              this.tableDataIsEmpty = false
+              this.isEmpty = false
+            } else {
+              this.tableData.data = []
+              this.tableDataIsEmpty = true
+              this.isEmpty = true
+
+              this.getEntity(null)
```
bartsmykla, line 321: I think that `null` here is unnecessary

jakubdyszkiewicz:
Talked offline that some approach to all views to fix them. Merging as is for now

PR #109, opened 2021-01-04
fix(dataplane) take into account application status as well

lobkovilya:
- Extract function for calculating Dataplane status - `getStatus`
- Take `inbound.health` into account when calculating Dataplane status
- Introduce new status `Partially degraded` if some of inbounds are not ready:
 
image
- In case of `Offline/Partially degraded` show the reasons in Overview tab:

image

### DIFF
--- a/src/dataplane.js
+++ b/src/dataplane.js
@@ -55,11 +55,11 @@ export function getStatus (dataplane, dataplaneInsight) {
     ? dataplane.networking.inbound
     : [{ health: { ready: true } }]
 
-  const everyInboundHealthy = inbounds
-    .every(item => item.health ? item.health.ready : true)
+  const someInboundUnhealthy = inbounds
+    .some(item => item.health ? !item.health.ready : false)
 
   const someSubscriptionOnline = dataplaneInsight.subscriptions
     .some(item => item.connectTime && item.connectTime.length && !item.disconnectTime)
 
-  return someSubscriptionOnline && everyInboundHealthy ? 'Online' : 'Offline'
+  return someSubscriptionOnline && !someInboundUnhealthy ? 'Online' : 'Offline'
 }
--- a/src/dataplane.js
+++ b/src/dataplane.js
@@ -44,7 +44,10 @@ export function dpTags (dataplane) {
   tags = Array.from(new Set(tags)) // remove duplicates
 
   return tags.map(tagPair => tagPair.split('='))
-    .map(([key, value]) => ({ label: key, value: value }))
+    .map(([key, value]) => ({
+      label: key,
+      value: value
+    }))
 }
 
 /*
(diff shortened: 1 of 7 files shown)


src/dataplane.js
```diff
@@ -46,3 +46,22 @@ export function dpTags (dataplane) {
...
+
+  const appStatus = inbounds
+    ? inbounds
+      .map(item => item.health ? item.health.ready : true)
+      .reduce((a, b) => a || b)
+    : true
+
+  const proxyStatus = dataplaneInsight.subscriptions
+    .map(item => item.connectTime && item.connectTime.length && !item.disconnectTime)
+    .reduce((a, b) => a || b)
```
austince, line 64: I think this is saying that 'Online' is at least 1 inbound is healthy (default `true` if no inbounds) and at least 1 proxy is healthy (default `false` if there are no proxies), right? If that's correct we may be able to use  [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) instead of mapping and reducing (which will also early exit as soon as the condition is met)? 

```js
const inbounds = dataplane.networking.inbound
 ? dataplane.networking.inbound
 : [ { health: { ready: true } } ];

const someInboundHealthy = inbounds.some(item => item.health ? item.health.ready : true);

// could also early return if all unhealthy?

const someProxyHealthy = dataplaneInsight.subscriptions
    .some(item => item.connectTime && item.connectTime.length && !item.disconnectTime);

return someInboundHealthy &&  someProxyHealthy ? 'Offline' : 'Online'
```

lobkovilya: > at least 1 proxy is healthy (default false if there are no proxies)

that's not entirely true, more correct to say "if 1 subscription has `connectTime` and doesn't have `disconnectTime` then we consider proxy `Online`"

But other than that yes, you are right! Thank you for your comment, I completely forgot about `some()` in js. Let me update the PR

lobkovilya: Oh, I think I mixed up things here, all inbounds should be true in order to be "Online", so I guess I need `every()`

austince: Ah got it, oops! Maybe we could also reverse the condition and still could use `some` for the early break? Like `someProxyUnhealthy` and `someInboundUnhealthy`?

austince: *just `someInboundHealthy`

lobkovilya: To be honest I'd prefer to have a single `return` value here even though it's less optimal than early break:
```js
return someSubscriptionOnline && everyInboundHealthy ? 'Online' : 'Offline'
```
I think it's a bit simpler to read and all constants are gathered together in one line. But I don't mind choose either option. What do you think?

austince: Oh yeah totally agree with the ending single return, I was just saying for the inbound condition, if we reversed that we could still use `some()` instead of `every()`, which breaks early as soon as one of the inbounds is unhealthy as `!someInboundUnHealthy === everyInboundHealthy`. Total nit at this point though, I think it looks good! Thanks for letting me jump in :) 


```js
const inbounds = dataplane.networking.inbound
 ? dataplane.networking.inbound
 : [ { health: { ready: true } } ];

const someInboundUnHealthy = inbounds.some(item => item.health ? !item.health.ready : false);

const someSubscriptionOnline = dataplaneInsight.subscriptions
    .some(item => item.connectTime && item.connectTime.length && !item.disconnectTime)

return !someInboundUnHealthy && someSubscriptionOnline ? 'Online' : 'Offline'
```

austince (approved):

src/views/Entities/Dataplanes.vue
```diff
@@ -625,9 +642,22 @@ export default {
                 return data
               }
 
+              const getDpStatus = async () => {
+                try {
+                  const res = await this.$api.getDataplaneOverviewFromMesh(entityMesh, entity.name)
```
jakubdyszkiewicz, line 647: Correct me if I'm wrong, but I see that we get a Dataplane in 596, then for this dataplane we've get overview at 606 and then overview again here. 3 requests. Can we make just one request for overview in 596?

lobkovilya: Yeah, I know, the initial idea was not to touch many lines with this PR, but it appears not so much of a work, so I fixed that. Please take look

jakubdyszkiewicz (approved):
I like the list of errors. Look pretty awesome.

PR #113, opened 2021-01-09
Fix/incorrect yaml for mesh

tharun208:
creating mesh through GUI was not able to generate exact YAML for k8's. This pr fixes by making GUI create mesh YAML for k8s. tested locally with both k8s and universal mode 👍🏻 

Fixes https://github.com/kumahq/kuma/issues/1388


bartsmykla:
Thank you for your helm with this and apologies for a late review!

bartsmykla (approved):

PR #114, opened 2021-01-18
feat(*) add info of dp and envoy versions mismatch

bartsmykla:
It's still WIP

Still missing:
- filled `versions.json` file
- show icon about version mismatch in dataplanes table
- copy logic from general `Dataplanes` view to `Standard`, `Ingress`, `Gateway`


bartsmykla:
It's not relevant anymore, as it was already shipped in: https://github.com/kumahq/kuma-gui/pull/142

PR #129, opened 2021-01-27
Refactor/use mesh insights in overview

bartsmykla:
- Refactored `fetch*` actions to remove duplicates
- Introduced new methods to call api endpoints for mesh insights
- Introduced `fetchAllMeshInsights` action + appropriate state
  and mutation
- Changed chained promises to fully use the `async/away` way,
  which was already being used but somehow mixed with raw promises
- Removed console.err when request failed from restClient,
  as it was not propagating errors higher and even failed
  request was marked as successful
- Fixed the type of store.version as it was failing in Header component
  as there was a data race
- Introduced concept of reducers (from react), a functions which are
  processing data in mutations.

  Maybe in the future, when the store will be divided into more
  modules, it would be good to get rid of reducers in favor of
  native Vue mutations, but currently it's much more readable
  to place these functions in one, separate file
- Introduced in state `meshInsightsFetching` property, which
  can be used in the future for some "loading" bar purposes
- Removed `dataplaneInsights` as they are currently unnecessary,
  as all data we were using it for are currently replaced by
  mesh insights
- Removed unnecessary mutations related to dataplane insights
- Refactored/fixed `fetchMeshInsights` action, which was wrong
  before as it is different, depending on mesh parameter
- Set default theme for amcharts charts to be animated
- Refactored `Overview` view to use mesh insights instead of multiple
  different endpoints

## Changes in charts

- Styles:
  - Restored radius on series
  - Added space (stroke/border) on series, which adds effect of
    white space between series
  - Because of removing top title, enlarged the chart

- Title/SumLabel:
  - Removed titles from top, and placed them inside instead of word
    `Total`
  - Set font color to main theme light `[rgba(41, 11, 83, 0.75)]`
  - Capitalized text
  - Set bold font weight for the amount number inside
- Labels:
  - Labels of series are now placed inside of series
  - Labels are now transparent for events, to not block clicks
  - Changed styles of series labels:
    - They are capitalized
    - They are white
    - Spacing between letters is set to 0.1em
- Tooltips:
  - Restored tooltips on non-disabled charts when hovering, with
    exception of `Zones` chart
  - Changed styles for tooltips:
    - Removed stroke
  - Tooltips now follows the cursor

- Other:
  - Introduced animation when charts initially load
  - Removed unused and unnecessary logic for setting `zIndex`
    property for charts
  - Change order of charts, now services are placed inside (second
    position), and dataplanes on the right (third position)

https://user-images.githubusercontent.com/11655498/105958876-7e3b6e00-607b-11eb-8da9-32b97b5ae31f.mov

Still missing:
- [ ] [Improve chart labels edge cases behaviour/look](https://github.com/kumahq/kuma-gui/pull/130)

### DIFF
--- a/src/components/OverviewCharts.vue
+++ b/src/components/OverviewCharts.vue
@@ -16,8 +16,9 @@ const getEmptyData = () => [{
   value: 1,
   color: am4core.color('#dadada'),
   opacity: 0.3,
+  strokeColor: am4core.color('#dadada'),
   strokeDasharray: '4,4',
-  strokeWidth: 1,
+  strokeWidth: 2,
 }]
 
 const ZONES_CHART_NAME = 'Zones'
@@ -32,7 +33,7 @@ export default {
       required: true,
     },
     dataplanes: {
-      type: Array,
+      type: Object,
       required: true,
     },
     services: {
@@ -41,12 +42,12 @@ export default {
     },
     selectedMesh: {
       type: String,
-      required: true
-    }
+      required: true,
+    },
   },
   data () {
     return {
-      charts: {}
+      charts: {},
     }
   },
   watch: {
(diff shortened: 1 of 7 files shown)


bartsmykla:
Closing in favor of: https://github.com/kumahq/kuma-gui/pull/133

PR #137, opened 2021-02-12
feat(services) change color for 'Partially degraded'

lobkovilya:
Before:
image

After:
image

### DIFF
--- a/src/assets/styles/components.scss
+++ b/src/assets/styles/components.scss
@@ -20,11 +20,11 @@
 }
 
 .entity-status.is-degraded {
-  color: var(--yellow-300);
+  color: #ff8040;
 }
 
 .entity-status.is-degraded .entity-status__dot {
-  background-color: var(--yellow-300);
+  background-color: #ff8040;
 }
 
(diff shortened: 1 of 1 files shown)


nickolaev (approved):

src/assets/styles/components.scss
```diff
@@ -20,11 +20,11 @@
 }
 
 .entity-status.is-degraded {
-  color: var(--yellow-300);
+  color: #ff8040;
```
tharun208, line 23: just a thought, `--yellow-400` won't work for this?

lobkovilya: not really :D 
image

tharun208: I referred kongponents for the color code - https://kongponents.konghq.com/style-guide/colors.html#colors for `yellow-400`.

lobkovilya: Oh I see, that's weird, I think `kuma-gui` depends on the previous version of kongponents, there are only 4 yellow colors like on the screenshot

PR #140, opened 2021-02-16
feat(services) timeout policy

lobkovilya:
GUI part for https://github.com/kumahq/kuma/pull/1554


src/router.js
```diff
@@ -315,6 +315,15 @@ export default (store) => {
...
           component: () => import(/* webpackChunkName: "retries" */ '@/views/Policies/Retries')
+        },
+        // timeouts
+        {
+          path: 'timeouts',
+          name: 'timeouts',
+          meta: {
+            title: 'Timeouts'
+          },
+          component: () => import(/* webpackChunkName: "retries" */ '@/views/Policies/Timeouts')
```
tharun208, line 326: can you change the chunk name to `timeouts` and also add the file extension to it. it should be '@/views/Policies/Timeouts.vue'

tharun208 (changes requested):

src/components/Sidebar/menu.js
```diff
@@ -126,6 +126,12 @@ export default [
               link: 'retries',
               title: false,
               parent: 'policies'
+            },
+            {
+              name: 'Timeouts',
+              link: 'timeouts',
+              title: false,
+              parent: 'timeouts'
```
tharun208, line 134: parent -> `policies`

tharun208 (changes requested):

src/router.js
lobkovilya: Thanks, fixed!

src/components/Sidebar/menu.js
lobkovilya: Thanks, fixed!

jakubdyszkiewicz (approved):

PR #149, opened 2021-03-24
UTM strings for Google Analytics

bloqhead:
I've added the UTM strings `utm_source` and `utm_medium` to URLs that go to the Kuma website from within the GUI so that we can track these events accordingly in Google Analytics.


subnetmarco:
@bartsmykla can we merge this?

bartsmykla:
@subnetmarco yes, lgtm

Issue #169, opened 2021-07-06
Dataplane install wizard suggests the `--mesh` flag when it shouldn't


jpeach:
The dataplane install wizard generates this `kuma-dp` commandline:
```bash
kuma-dp run \
      --name=example-g68sex \
      --mesh=default \
      --cp-address=https://127.0.0.1:5678/ \
      --dataplane="type: Dataplane
mesh: default
name: example-g68sex
networking: 
  address: 192.168.1.100
  inbound: 
    - port: 80
      servicePort: 6060
      serviceAddress: 127.0.0.1
      tags: 
        kuma.io/service: example
        kuma.io/protocol: tcp" \
      --dataplane-token-file=kuma-token-example-g68sex
```

But `kuma-dp` doesn't allow `--mesh` if it is also set in the dataplane resource. The wizard should omit the flag in this case.

tomaszwylezek:
Hi @jpeach!
I took a look at it and you are right. Also, it looks like it shouldn't produce flag `--name` as the result of the process. I will create a PR that is resolving both of these issues. Great that you found and shared that issue!

tomaszwylezek:
Closing, as it was fixed in #173

jpeach:
Thanks! FYI, you can reference the issue in your commit comments to link everything https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #173, opened 2021-07-07
fix(*) Remove --mesh and --name flags

tomaszwylezek:
It is fixing an issue with not needed flags, that was described here: https://github.com/kumahq/kuma-gui/issues/169


bartsmykla (approved):
lgtm

PR #175, opened 2021-07-09
fix(*) fix wrong dp type

tomaszwylezek:
The issue was because we weren't passing `dataplane` param into the function. After it has been provided, the issue is gone.
Also, it will remove `getDataplaneType` from being a prop as we have a generic function that is handling that behaviour properly.
Also removed `addDataFields` as it was doing nothing and did a small refactor following the boy scout rule.

Screenshots:
### All dps
Screenshot 2021-07-09 at 10 37 15

### Standard dps
Screenshot 2021-07-09 at 10 37 22

### Gateway dps
Screenshot 2021-07-09 at 10 37 27


bartsmykla (approved):
lgtm

PR #176, opened 2021-07-09
feat(*) add msw as mock for requests

tomaszwylezek:
In this PR we are going to replace `axios-mock-adapter` with [msw](https://github.com/mswjs/msw).
The reason behind that decision is whats written in FAQ of MSW [link](https://mswjs.io/docs/faq)

> most solutions provide requests interception on the application level, while Mock Service Worker intercepts requests on the network level. It also allows you to use the same mock definition not only for testing but for development and debugging as well.

Also with `axios-mock-adapter` it was impossible to see the whole flow, as it was mocking request before it was called ane because of that there was no request in the network which makes it harder to follow from where the data comes.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -271,7 +271,7 @@ view). Your route would instead be structured like so: `/meshes/default`
 
 ### Creating mock data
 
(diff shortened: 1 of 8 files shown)


bartsmykla (approved):
I love it! lgtm

PR #179, opened 2021-07-15
feat(*) adjust routes and wizards

tomaszwylezek:
### Changes:

- Create 2 independent main views (Shell and EmptyShell - naming may change in future) where one is related to "focus" mode (like wizards and onboarding) and the second one is a standard view with all sidebars/breadcrumbs etc.
- Change of paths in url. Previously we had inconsistency there. Right now 'mesh' depending routes will be with prefix `/mesh`. Below few examples of what has changed and how.
  - `/zones`-> `/zones`
  - `/zone-ingresses`-> `/zone-ingresses`
  - `/all/overview`-> `/mesh/all/overview`
  - `/all/dataplanes` -> `/mesh/all/dataplanes`
- Adjust validation on both universal/kubernetes wizard
- Fix router (right now when you are on step 1/2 and click back it takes you back to the previous step, previously it wasn't working fine)
- Adjust loading screen
- Adjust getting started page
- Adjustment of wizard
- Cleanup of actions
- Add test for DataplaneUniversal wizard
- Add tests in CI

### Video

https://user-images.githubusercontent.com/16152347/125760608-30c9a379-ac42-4fd7-8e6d-83d6992619c3.mov


src/store/index.ts
```diff
@@ -305,6 +305,66 @@ export default (api: Kuma) => {
...
+
+              if (lsVersion !== newVersion) {
+                // reload the app
+                window.location.reload()
+                // update the version in localStorage
+                localStorage.setItem('kumaVersion', newVersion)
+              } else {
+                store.commit('SET_GLOBAL_LOADING', { globalLoading: false })
+              }
+            })
```
bartsmykla, line 342: As the function is async, what do you think about instead of using promises' `.then` method here to use `await`?

src/store/index.ts
```diff
@@ -305,6 +305,66 @@ export default (api: Kuma) => {
...
+            if (routeMesh) {
+              // if the `mesh` param is present, use that
+              return routeMesh
+            } else if (lsMesh && lsMesh !== 'undefined' && lsMesh.length > 0) {
+              // or use what's available in localStorage
+              return lsMesh
+            } else {
+              // otherwise, fall back to the default value from our VueX store
+              return getters.getSelectedMesh
+            }
```
bartsmykla, line 356:
```suggestion
            }

            // otherwise, fall back to the default value from our VueX store
            return getters.getSelectedMesh
```

What do you think about it?

src/views/Onboarding/Complete.vue
```diff
@@ -138,6 +138,13 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+
+.welcome {
+  max-width: 640px;
+  margin-left: auto;
+  margin-right: auto;
+}
```
bartsmykla, line 146: I'm kinda curious if maybe we could just use tailwind's classes instead of writing custom styles here. Wdyt?

src/views/Onboarding/GetStarted.vue
```diff
@@ -379,15 +355,16 @@ export default {
   border-radius: 4px;
 }
 
+.welcome {
+  max-width: 640px;
+  margin-left: auto;
+  margin-right: auto;
+}
```
bartsmykla, line 362: The same question here. Could we use tailwind's built in classes instead of custom styles? Especially as it's duplicated here

src/views/Wizard/components/StepSkeleton.vue
```diff
@@ -90,10 +90,7 @@ export default {
       type: Array,
       default: () => {}
     },
-    advanceCheck: {
-      type: Boolean,
-      default: false
-    },
+
```
bartsmykla, line 93: Nit
```suggestion
```

src/views/Wizard/views/DataplaneKubernetes.vue
```diff
@@ -779,33 +780,31 @@ export default {
...
 
       return assembledBlock
+    },
+
+    nextDisabled() {
+      const mesh = this.validate.meshName
+
+      const {
+        k8sNamespaceSelection
+      } = this.validate
```
bartsmykla, line 790: Wdyt about it?
```suggestion
      const {
        mesh: meshName,
        k8sNamespaceSelection
      } = this.validate
```

src/views/Wizard/views/DataplaneKubernetes.vue
```diff
@@ -779,33 +780,31 @@ export default {
...
+        return true
+      }
+
+      if (this.$route.query.step === '1') {
+        if (k8sNamespaceSelection) {
+          return false
+        } else {
+          return true
+        }
+      }
```
bartsmykla, line 802: Wdyt?
```suggestion
        return !k8sNamespaceSelection
```

src/views/Wizard/views/DataplaneUniversal.vue
```diff
@@ -662,32 +611,35 @@ export default {
...
+      if (!mesh.length) {
+        return true
+      }
+
+      if (this.$route.query.step === '1') {
+        if (univDataplaneServiceName && univDataplaneId) {
+          return false
         } else {
-          this.nextDisabled = true
+          return true
```
bartsmykla, line 622: It can be simplified as in one of the suggestions before I think

src/views/Wizard/views/DataplaneUniversal.vue
```diff
@@ -594,64 +592,15 @@ export default {
...
-      deep: true
     },
 
-    '$route' () {
-      const step = this.$route.query.step
+    nextDisabled() {
+      const mesh = this.validate.meshName
 
       const {
         univDataplaneServiceName,
```
bartsmykla, line 657: You can destructurize the `.meshName` and change the name to `.mesh` the same as in one of the suggestions before I think

src/views/Wizard/views/DataplaneUniversal.vue
```diff
@@ -662,32 +611,35 @@ export default {
...
             univDataplaneNetworkServicePort &&
             univDataplaneNetworkDPPort &&
             univDataplaneNetworkProtocol
         ) {
-          this.nextDisabled = false
+          return false
         } else {
-          this.nextDisabled = true
+          return true
         }
```
bartsmykla, line 688: This can be simplified and if you don"t want to return with this complicated condition, we can get read of at least the `else` statement. Wdyt?

src/components/Sidebar/Sidebar.vue
```diff
@@ -56,10 +55,12 @@ import NavItem from '@/components/Sidebar/NavItem'
...
 import MeshSelector from '@/components/Utils/MeshSelector'
 
-import { getItemFromStorage, setItemToStorage } from '@/Cache'
-import { mapState, mapMutations } from 'vuex'
+import { mapState } from 'vuex'
 
 export default {
+
+  name: 'Sidebar',
+
```
bartsmykla, line 63:
```suggestion
  name: 'Sidebar',
```
Nit

bartsmykla (approved):

src/store/index.ts
tomaszwylezek: I thought about it but I decided to do it, later on, to do not change logic everywhere when doing cleanup.
Also when working with `async/await` we need to remember that it is blocking the flow of execution of the function, so if we just change everything to use `async/await` then every call needs to wait for the previous call to be executed.
The other option might be to aggregate all `Promises` and `await Promise.allSettled` to make sure we do not delay the resolution of it. But as I said, I will think about it later on.

src/views/Onboarding/Complete.vue
tomaszwylezek: I agree with you. But when I'm changing existing stuff I prefer to do not to change styles if I'm not introducing any changes to them. But nice catch that I duplicated that one, I was almost sure that I extracted in into `components.scss`. I will remove duplication then.

src/views/Onboarding/GetStarted.vue
tomaszwylezek: https://github.com/kumahq/kuma-gui/pull/179#discussion_r670411039

src/views/Wizard/views/DataplaneKubernetes.vue
tomaszwylezek: Nice catch, ofc its better

src/views/Wizard/views/DataplaneKubernetes.vue
tomaszwylezek: 🤦🏻  how I missed that one 😄

src/store/index.ts
tomaszwylezek: I would leave it as it is because the whole function is kind of messy and when I will split it into smaller chunks then I think it will be handed.

src/views/Onboarding/Complete.vue
bartsmykla: sgtm

src/store/index.ts
bartsmykla: sgtm

PR #180, opened 2021-07-19
Refactor/vuex module

tomaszwylezek:
This PR is the first one in which we are going to decouple the global state in Vuex.


bartsmykla (approved):
I love it

Issue #183, opened 2021-07-26
Kuma manager display issue on Safari


liyangau:
## Issue statement

Display issue when access Kuma Manager on Safari.

For example, `Skip to Dashboard` button, we can't see the text and the button background colour makes the text hard to read (white text + light gray backgroud)
![image](https://user-images.githubusercontent.com/71299093/126951027-bca8adf3-ab53-416f-bc70-d0e8fd868050.png)

tomaszwylezek:
Hi @liyangau 
Thanks for reporting this issue.
I already provided a solution for that in https://github.com/kumahq/kuma-gui/pull/184
I will keep you informed when it will be merged & backported.

tomaszwylezek:
@liyangau
The solution was merged and backported, closing the issue.

PR #184, opened 2021-07-26
bug(*) fix issue on safari with wrongly displayed buttons

tomaszwylezek:
It fixes a bug reported in: https://github.com/kumahq/kuma-gui/issues/183
Because tailwind set default appearance in Safari we need to disable it to allow us doing changes to how buttons look like.


bartsmykla (approved):
lgtm

Issue #171, opened 2021-07-06
Mesh creation wizard generated a config with "null" backend names


jpeach:
I used the mesh creation wizard and enabled but didn't enter any names for the logging, metrics, etc backends. I expected that the UI would use default names, but it generated an invalid mesh resource:

```
$ echo "type: Mesh
name: demo
mtls: 
  enabledBackend: null
  backends: 
    - name: null
      type: builtin
  enabled: true
tracing: 
  defaultBackend: null
  backends: 
    - name: null
      type: zipkin
      conf: 
        sampling: 100
        url: null
logging: 
  backends: 
    - name: null
      format: '{ "destination": "%KUMA_DESTINATION_SERVICE%", "destinationAddress": "%UPSTREAM_LOCAL_ADDRESS%", "source": "%KUMA_SOURCE_SERVICE%", "sourceAddress": "%KUMA_SOURCE_ADDRESS%", "bytesReceived": "%BYTES_RECEIVED%", "bytesSent": "%BYTES_SENT%"}'
      type: tcp
      conf: 
        address: '127.0.0.1:5000'
metrics: 
  enabledBackend: null
  backends: 
    - name: null
      type: prometheus
      conf: 
        port: 5670
        path: /metrics" | kumactl apply -f -
Error: Could not create a resource (Resource is not valid)
* logging.backends[0].name: cannot be empty
* tracing.backends[0].name: cannot be empty
* tracing.backends[0].config.url: cannot be empty

```

tomaszwylezek:
It was fixed in #186 and merged into kuma in https://github.com/kumahq/kuma/pull/2481

PR #186, opened 2021-07-29
fix(*) mesh creation required fields

tomaszwylezek:
### Summary
It fixes an issue presented in https://github.com/kumahq/kuma-gui/issues/171 by adding basic validation on the required fields.
Also, it adds a test that checks the whole mesh wizard behavior.
Also it unifies the way how we manage data/validation across wizards and cleanup the vuex state a bit.

### Movie of adjusted wizard

https://user-images.githubusercontent.com/16152347/127499305-69493d5b-c038-45a3-b6cc-55696a5d7fd8.mov


src/views/Wizard/views/Mesh.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    await userEvent.click(screen.getByText('Next ›'))
+  }
+
+  it('passes whole wizzard and render yaml', async () => {
+    const {
+      container,
+    } = renderWithVuex(Mesh, {
+      store: {
+
```
bartsmykla, line 38: Nit: unnecessary empty line

bartsmykla (approved):
Lgtm

src/views/Wizard/views/Mesh.spec.ts
tomaszwylezek: Nice catch. I think I will provide husky and pre-commit hook that will take care of these kinds of "linting" issues. For now, I will skip it.

Issue #135, opened 2021-02-11
KLoader has issue where progress can exceed 100


darrenjennings:
I noticed that the KLoader is using a copied version of an old component. We moved the Full screen loader to a kongponent (which resolves the >100 progress bug):

https://github.com/Kong/kongponents/blob/master/packages/KSkeleton/FullScreenKongSkeleton.vue#L54
https://github.com/kumahq/kuma-gui/blob/master/src/components/KLoader.vue#L43

Possible solution is that the KLoader can receive an asset name so that the assets can be customized so you could pass `../assets/images/kuma-loader-v1.gif?external` to KSkeleton, and then have kuma-gui use the kskeleton kongponent.

tomaszwylezek:
It was fixed in #187 and merged into kuma in kumahq/kuma#2481

PR #187, opened 2021-07-29
bug(*) fix wrong loading values

tomaszwylezek:
### Summary
It fixes an issue presented in #135  by disabling the possibility of width in the loader to exceed 100%.
Also, it adds a test that checks the behavior.


bartsmykla (approved):
Lgtm

PR #189, opened 2021-07-30
refactor(*) change the way how Tabs works

tomaszwylezek:
### Summary
This PR is about to get rid of vuex management of Tabs state and to make it a local state.
Also it get rids of some repetitive code and makes the code cleaner


src/components/Utils/Tabs.vue
```diff
@@ -137,12 +133,14 @@ export default {
     activeTab: {
       get () {
         if (!this.tabState) {
-          return this.$store.state.selectedTab
+          return this.tabs[0].hash
```
bartsmykla, line 136: What if there won't be any tabs?

src/components/Utils/Tabs.vue
```diff
@@ -137,12 +133,14 @@ export default {
...
         } else {
-          return `#${this.$store.state[this.tabState]}`
+          return `#${this.tabState}`
         }
```
bartsmykla, line 143: We can drop the else statement here

bartsmykla (approved):
Few nits but exept that lgtm

src/components/Utils/Tabs.vue
tomaszwylezek: Then it will fail. I just tried to adjust validation to avoid that one. But if makes no sense as then it fails in Ktabs from https://kongponents.konghq.com/components/tabs.html.
I will skip it then as it's worthless to wrap it with much more code to handle that scenario, especially that we are responsible for providing the data, so it shouldn't be a case where we miss that one.

tomaszwylezek: Well, I realized that the code was not needed. 😅

PR #199, opened 2021-08-05
refactor(*) create common function for fetching data

tomaszwylezek:
### Summary
It will align the way how the data is retrieved from API to be displayed in the GUI. Right now we have a lot of duplication of code. This is one of many steps to make sure we can better manage the data/duplication of code.

### DIFF
--- a/src/utils/tableDataUtils.ts
+++ b/src/utils/tableDataUtils.ts
@@ -1,14 +1,5 @@
 import { getOffset } from '@/helpers'
-
-interface TableDataParams {
-  getSingleEntity: (mesh: string, query: string | null, params: { size: number; offset: string | null }) => any
-  getAllEntities: (params: { size: number; offset: string | null }) => any
-  getAllEntitiesFromMesh: (mesh: string) => any
-  mesh: string
-  size: number
-  query: string | null
-  offset: string | null
-}
+import { TableDataParams } from './tableDataUtils.types'
 
 function sortEntities(
   items: {
@@ -62,7 +53,7 @@ async function getAPICallFunction({
   return await getAllEntitiesFromMesh(mesh)
 }
 
-async function getTableData({
+export async function getTableData({
   getSingleEntity,
   getAllEntities,
   getAllEntitiesFromMesh,
@@ -95,5 +86,3 @@ async function getTableData({
     next: response.next ? getOffset(response.next) : null,
   }
 }
-
-export { getTableData }
--- a/src/utils/tableDataUtils.ts
+++ b/src/utils/tableDataUtils.ts
@@ -1,0 +1,99 @@
+import { getOffset } from '@/helpers'
+
+interface TableDataParams {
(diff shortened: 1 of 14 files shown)


src/utils/tableDataUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+function sortEntities(
+  items: {
+    name: string
+    mesh: string
+  }[],
+) {
+  return [...items].sort((a, b) => (a.name > b.name ? 1 : a.name === b.name ? (a.mesh > b.mesh ? 1 : -1) : -1))
+}
+
+const getItems = (response: {
```
Pan-Bubr, line 13: Could we add a type signature? I am not sure what are we returning here

src/utils/tableDataUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+  total: number
+  items: {
+    name: string
+    mesh: string
+  }[]
+}) => {
+  const r = response
+
+  if (!r.total) {
+    return [r]
```
Pan-Bubr, line 23: If the total is 0 we return whole response as a single-element Array and lower we return [...items]

src/utils/tableDataUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+  query,
+  size,
+  offset,
+}: TableDataParams) {
+  const params = {
+    size,
+    offset,
+  }
+
+  if (mesh === 'all') {
```
Pan-Bubr, line 47: Do we need an await here? We can just return promise.

Pan-Bubr (changes requested):

src/utils/tableDataUtils.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+const getItems = (response: {
+  total: number
+  items: {
+    name: string
+    mesh: string
+  }[]
+}) => {
+  const r = response
```
Pan-Bubr, line 20: What is the purpose of this assignment?

src/utils/tableDataUtils.ts
Pan-Bubr: Probably
`return [...r.items]`

src/utils/tableDataUtils.ts
```diff
@@ -0,0 +1,88 @@
+import { getOffset } from '@/helpers'
+import { TableDataParams } from './tableDataUtils.types'
+
+function sortEntities(
+  items: {
```
Pan-Bubr, line 5: Items are used in a lot of places, i suggest that they deserve an Interface

src/utils/tableDataUtils.ts
tomaszwylezek: Good point, we don't need await there

src/utils/tableDataUtils.ts
tomaszwylezek: None, just copy-paste of function from which it was taken. Fixed.

Pan-Bubr (approved):

PR #201, opened 2021-08-09
Bug missing version zone standalone

tomaszwylezek:
### Summary
When working in standalone we weren't displaying version of cp.

### Before
Screenshot 2021-08-09 at 12 32 31

### After
Screenshot 2021-08-09 at 12 33 11


bartsmykla (approved):
lgtm

Pan-Bubr (approved):

PR #202, opened 2021-08-10
feat(*) disable onboarding when no requirement fulfilled to display it

tomaszwylezek:
### Summary
We no longer want to display onboarding for users which already have fulfilled the condition to display onboarding popup. 

So we are redirecting into onboarding view only if the user has default mesh and no dataplanes and did not seen already the onboarding view. In any other case we are no longer redirecting into onboarding and if user try to open that page it will be redirected into global overview.

### Explanation of changes 

As right now we have to wait in the router to make sure we can check condition if the page should be displayed or not (so we need to wait till all bootstrap action will be resolved, and `globalLoading` will change into `false`) it caused a flash of the blank screen after loading is finished (movie below). To avoid that I provided a delay on change on loading in `App.vue` to make sure that all calculations done in routing will be finished after that. It is something that will be changed in the future after we change the onboarding.

#### Without delay

https://user-images.githubusercontent.com/16152347/128845079-cf9f697d-0df7-4e77-b338-c6f1548cde67.mov

#### With delay 

https://user-images.githubusercontent.com/16152347/128845369-fb5228c1-f92c-4b14-8bd0-fd790438fd07.mov

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -40,7 +40,7 @@ export default {
     },
   },
   data() {
-    return { loading: true, timeount: null }
+    return { loading: true, timeout: null }
   },
   computed: {
     ...mapState({
@@ -51,9 +51,9 @@ export default {
     }),
   },
   watch: {
-    globalLoading: function (val) {
+    globalLoading: function (loading) {
       this.timeout = setTimeout(() => {
-        this.loading = val
+        this.loading = loading
       }, 200)
     },
   },
--- a/src/App.vue
+++ b/src/App.vue
@@ -36,23 +36,36 @@ export default {
     title: 'Home',
     titleTemplate: `%s | ${process.env.VUE_APP_NAMESPACE}`,
     htmlAttrs: {
-      lang: 'en'
-    }
+      lang: 'en',
+    },
+  },
+  data() {
+    return { loading: true, timeount: null }
   },
   computed: {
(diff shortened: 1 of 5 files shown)


Pan-Bubr (approved):
lgtm

src/App.vue
```diff
@@ -36,23 +36,36 @@ export default {
...
+      globalLoading: (state) => state.globalLoading,
     }),
     ...mapGetters({
-      status: 'config/getStatus'
-    })
+      status: 'config/getStatus',
+    }),
+  },
+  watch: {
+    globalLoading: function (val) {
```
Pan-Bubr, line 54: `globalLoading: function (loading: boolean) {`

src/App.vue
```diff
@@ -36,23 +36,36 @@ export default {
     title: 'Home',
     titleTemplate: `%s | ${process.env.VUE_APP_NAMESPACE}`,
     htmlAttrs: {
-      lang: 'en'
-    }
+      lang: 'en',
+    },
+  },
+  data() {
+    return { loading: true, timeount: null }
```
Pan-Bubr, line 43: TYPO
`timeout`

Pan-Bubr (changes requested):

Pan-Bubr (approved):

PR #203, opened 2021-08-11
feat(*) add tracking

tomaszwylezek:
### Summary
It will add tracking for:
- every nav click
- tabs clicks
- pagination clicks

Also, it removes some unused code related to sidebar.


Pan-Bubr (approved):
looks good with one suggestion

PR #205, opened 2021-08-12
Refactor/change api

tomaszwylezek:
### Summary
Skip step in between (request for api config) as we no longer need that, as the API url is constant.

This code was some legacy stuff which was based on some historical decisions where we had many servers and they weren't consolidated.


Pan-Bubr (approved):
good job

PR #209, opened 2021-08-17
refactor(*) cleanup of vuex store

tomaszwylezek:
### Sumary

As we introduced `mesh-insights` some time ago, there is no longer a need to request for each policy/dataplanes etc. independently, as we can retrieve it from aggregated data.
Because of that, it allows us to remove a lot of code and simplify the logic.


bartsmykla (approved):
lgtm

Pan-Bubr (approved):

PR #210, opened 2021-08-18
feat(*) add datadog-rum

tomaszwylezek:
### Summary
This PR introduces datadog rum which will allow us better track user behavior/problems which they encounter.

Also small cleanup of env variables.

### DIFF
--- a/.env
+++ b/.env
@@ -1,12 +1,15 @@
 VUE_APP_DATA_TIMEOUT=500
-VUE_APP_SENTRY_DSN=https://06e53ac131ba4456b589f6e8e956e9ef@o424646.ingest.sentry.io/5356675
 VUE_APP_INSTALL_URL=https://kuma.io/install/latest/
 VUE_APP_VERSION_URL=https://kuma.io/latest_version/
 VUE_APP_NAMESPACE=Kuma
 VUE_APP_SHOW_ENTERPRISE_BUTTON=true
 # this path is relative to 'src/assets/images/'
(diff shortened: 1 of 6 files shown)


Pan-Bubr (approved):

austince:
Is this opt-in?

austince:
Coming from community call notes as context: can we revert this PR until we discuss it further in the community, or at least block any release from including this?

subnetmarco:
The goal with this PR is to make sure that we collect a statistically quantifiable amount of interactions with the GUI in order to (a) improve the GUI and understand how it is being used (b) collect frontend errors in order to fix them across the variety of configurations that the user will be using (browsers, their versions, etc).

This PR does not provide any other benefit, but giving us a chance to make the product better. As such I suggest that we should definitely provide an explicit way to disable the feature, while at the same time make sure that we don't collect any personal information. Datadog offers the option to stop collecting IP addresses, which really is the only identifiable bit of information we can collect (since the GUI doesn't have any signup/registration).

Per Datadog's [website](https://docs.datadoghq.com/real_user_monitoring/browser/data_collected/): 

> Note: By default, Datadog stores the client IP address. If you want to stop collecting IP addresses, contact Support. This does not impact the collection of geolocation attributes listed above.

Of course, we could also not collect any data whatsoever, but that will leave us in the dark when it comes to products improvements that are derived from real usage.

tomaszwylezek:
@austince for now the PR was reverted in #211

PR #212, opened 2021-08-19
refactor(*) systematise pagination

tomaszwylezek:
### Summary

Previously we had in every view separate Pagination and logic which was handling it. 

But when I went over the application I realized that in every place we are repeating one pattern, so instead of having it repeated I decided to move Pagination and logic related to that inside `DataOverview` component which reduces the amount of duplicated code and will make code cleaner and easier to manage.

Also created a common event which I attached to reload.
Also provided tests which check if Pagination works fine.

### DIFF
--- a/src/views/Policies/ProxyTemplates.vue
+++ b/src/views/Policies/ProxyTemplates.vue
@@ -143,7 +143,6 @@ export default {
       rawEntity: null,
       firstEntity: null,
       pageSize: PAGE_SIZE_DEFAULT,
-
       next: null,
     }
   },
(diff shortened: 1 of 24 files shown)


src/views/Policies/ProxyTemplates.vue
```diff
@@ -152,10 +143,8 @@ export default {
       rawEntity: null,
       firstEntity: null,
       pageSize: PAGE_SIZE_DEFAULT,
-      pageOffset: null,
+
```
bartsmykla, line 146: nit: is this empty line intentional?

bartsmykla (approved):
lgtm

src/views/Policies/ProxyTemplates.vue
tomaszwylezek: nope, I overlooked that one, fixed

PR #215, opened 2021-08-24
chore(*) upgrade tests

tomaszwylezek:
### Summary

Because of the way how Jest works (and because it does not handle SVG well), we weren't able to test files that contain anything from charts.
That PR allows us to do so.

Also, it increases the amount of memory used in CircleCI as it started to throw ` ENOMEM: not enough memory, read` error. I Will observe if it will resolve the issue - if not then I will try to find some other solution for it.


bartsmykla (approved):
lgtm

PR #216, opened 2021-08-25
Chore/change endpoint

tomaszwylezek:
#### Summary
Previously we used `/status/zones` to get an overview of statutes of zones. But as we introduced `*-insights` then we should rely on them and use them to align all of our API usages. 

Also upgraded the mocks to be aligned with API responses and did some cleanup.

### DIFF
--- a/src/dataplane.ts
+++ b/src/dataplane.ts
@@ -1,4 +1,5 @@
 import { humanReadableDate } from '@/helpers'
+import { ONLINE, OFFLINE, PARTIALLY_DEGRADED } from '@/consts'
 import { satisfies } from 'semver'
 import Kuma from '@/services/kuma'
 
@@ -89,14 +90,14 @@ export function getStatus(dataplane: DataPlane, dataplaneInsight: TODO = {}) {
     const allInboundsOnline = errors.length === 0
 
     if (!proxyOnline || allInboundsOffline) {
-      return 'Offline'
+      return OFFLINE
     }
 
     if (!allInboundsOnline) {
-      return 'Partially degraded'
+      return PARTIALLY_DEGRADED
     }
 
-    return 'Online'
+    return ONLINE
   }
 
   return {
@@ -106,10 +107,10 @@ export function getStatus(dataplane: DataPlane, dataplaneInsight: TODO = {}) {
 }
 
 /*
-getStatus takes ZoneIngressInsight and returns the status 'Online' or 'Offline'
+getItemStatusFromInsight takes object with subscriptions and returns the status 'Online' or 'Offline'
  */
-export function getZoneIngressStatus(zoneIngressInsight: TODO = {}) {
-  const subscriptions = zoneIngressInsight.subscriptions ? zoneIngressInsight.subscriptions : []
+export function getItemStatusFromInsight(item: TODO = {}): { status: typeof ONLINE | typeof OFFLINE } {
+  const subscriptions = item.subscriptions ? item.subscriptions : []
 
   const proxyOnline = subscriptions.some(
     (item: TODO) => item.connectTime && item.connectTime.length && !item.disconnectTime,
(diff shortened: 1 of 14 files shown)


src/dataplane.ts
```diff
@@ -106,32 +107,28 @@ export function getStatus(dataplane: DataPlane, dataplaneInsight: TODO = {}) {
 }
 
 /*
-getStatus takes ZoneIngressInsight and returns the status 'Online' or 'Offline'
+getItemStatusFromInsight takes object with subscriptions and returns the status 'Online' or 'Offline'
  */
-export function getZoneIngressStatus(zoneIngressInsight: TODO = {}) {
-  const subscriptions = zoneIngressInsight.subscriptions ? zoneIngressInsight.subscriptions : []
+export function getItemStatusFromInsight(item: TODO = {}): { status: typeof ONLINE | typeof OFFLINE } {
+  const subscriptions = item.subscriptions ? item.subscriptions : []
```
bartsmykla, line 113:
```suggestion
  const { subscriptions = [] } = item
```
maybe like this?

src/store/index.ts
```diff
@@ -299,7 +301,7 @@ export default (): Module<RootInterface, RootInterface> => ({
           // if any dataplanes are offline
           for (let i = 0; i < Object.values(result).length; i++) {
             const statusVal = Object.values(result[i])[0]
-            const isOnline = !(statusVal === 'Offline' || statusVal === 'offline')
+            const isOnline = !(statusVal === OFFLINE || statusVal === OFFLINE.toLowerCase())
```
bartsmykla, line 304:
```suggestion
            const isOnline = statusVal.toLowerCase() !== OFFLINE.toLowerCase()
```
maybe like this?

src/store/index.ts
```diff
@@ -443,21 +450,30 @@ export default (): Module<RootInterface, RootInterface> => ({
...
-    setOverviewZonesChartData({ state, commit }, statuses = []) {
-      const total = statuses.length
-      const online = statuses.filter(({ active }: { active: boolean }) => active).length
+    setOverviewZonesChartData({ state, commit }, { data = [] }) {
+      const total = data.length
+
+      let online = 0
+
+      data.forEach((item: any): void => {
+        const status = getItemStatusFromInsight(item.zoneInsight).status
```
bartsmykla, line 459:
```suggestion
        const { status } = getItemStatusFromInsight(item.zoneInsight)
```

maybe like this?

src/utils/tableDataUtils.ts
```diff
@@ -27,13 +27,17 @@ function getAPICallFunction({
     offset,
   }
 
-  if (mesh === 'all') {
+  if (!mesh || mesh === 'all') {
     return getAllEntities(params)
   } else if (getSingleEntity && query && query.length && mesh !== 'all') {
     return getSingleEntity(mesh, query, params)
   }
```
bartsmykla, line 34:
```suggestion
  if (!mesh || mesh === 'all') {
    return getAllEntities(params)
  }

  if (getSingleEntity && query && query.length && mesh !== 'all') {
    return getSingleEntity(mesh, query, params)
  }
```
maybe like this, to be consistent, or if not, to add if statement below to another `else if`?

bartsmykla (approved):
just nits

src/dataplane.ts
tomaszwylezek: Good idea

src/store/index.ts
tomaszwylezek: I have no idea if `statusVal` might be or not `undefined`. In the case of `undefined` it will fail so for now I will leave it as it was.

src/store/index.ts
tomaszwylezek: good idea

src/utils/tableDataUtils.ts
tomaszwylezek: yeah, you are right, especially that we return the value so there is no option that it will enter the second one after it pass first one.

src/store/index.ts
bartsmykla: got it

PR #217, opened 2021-08-31
feat(*) Zones backend, dpp info & config

tomaszwylezek:
### Summary
This PR is about adding 3 thing into Zones page.

1. New backend column which informs about backend storage at zone
2. New field in overview - which inform about what data plane authentication is enabled
3. New tab in tabs - with the latest config of the zone

Screenshot 2021-08-31 at 10 12 41

Screenshot 2021-08-31 at 10 12 57


src/views/Entities/Zones.vue
```diff
@@ -339,11 +345,13 @@ export default {
     parseData(entity) {
       const { zoneInsight = {} } = entity
       let zoneCpVersion = '-'
+      let backend = ''
 
       if (zoneInsight.subscriptions && zoneInsight.subscriptions.length) {
         zoneInsight.subscriptions.forEach((item, index) => {
           if (item.version && item.version.kumaCp) {
             zoneCpVersion = item.version.kumaCp.version
+            backend = JSON.parse(item.config).store.type
```
jakubdyszkiewicz, line 354: technically you can still have old Zone CP from the previous major so the config might be null.
I think it would safer to do `if (item.config) {`

jakubdyszkiewicz (commented):
nice!

src/views/Entities/Zones.vue
tomaszwylezek: I will adjust it then

jakubdyszkiewicz (approved):

PR #219, opened 2021-09-02
reafactor(*) v-slots upgrade

tomaszwylezek:
### Summary

In this PR we are getting rid of old deprecated syntax of the slot and change it into a new one. It required some code adoption which was also included as a part of PR
This is also one of many steps which will make the upgrade to Vue 3 easier in the future.

Also, it is removing some unused code from the codebase

#### Vue slot info

> In 2.6.0, we introduced a new unified syntax (the v-slot directive) for named and scoped slots. It replaces the slot and slot-scope attributes, which are now deprecated, but have not been removed and are still documented here. The rationale for introducing the new syntax is described in this RFC.

https://vuejs.org/v2/guide/components-slots.html


Pan-Bubr (approved):
I'm proud

PR #221, opened 2021-09-07
refactor(*) remove not used code

tomaszwylezek:
### Summary
Across our app, we had a lot of code that was not used anymore. This PR is about to get rid of it && also it removed not needed props from some components to make them simpler to use.


lahabana (approved):
Dead code is good code :)

PR #227, opened 2021-09-14
fix(*) Minor adjustrments

tomaszwylezek:
### Summary

~~It will resolve issues with CORS on firefox and also~~ Fix minor issues (fix the case when null may be provided to semver or wrong icon size)


bartsmykla (approved):
lgtm

PR #229, opened 2021-09-14
Fix hardcoded docs links.

johnharris85:
Zone and ZoneIngresses both had hardcoded (and broken) docs links. Fixed these by grabbing the current version (similar to the way other components did).

Signed-off-by: John Harris john.harris@konghq.com


tomaszwylezek (approved):
Looks good to me!

PR #232, opened 2021-09-24
fix(*) properly fetch all resources

tomaszwylezek:
### Summary

Previously when we want to request all resources we wrongly increment the value of offset which caused hundreds of unnecessary calls and wrong results. This fix will resolve that issue.


bartsmykla (approved):
lgtm

PR #233, opened 2021-09-30
refactor(*) adjust styles and accordion

tomaszwylezek:
### Summary

• Get rid of an arrow which was by default in every button

Screenshot 2021-09-30 at 10 58 15

• Add the possibility to use accordion with two opened items. Add the arrow at the right side of the accordion head to indicate that it can be open which turns up/down based on if it's active or not.

Screenshot 2021-09-30 at 11 00 16


bartsmykla (approved):
lgtm

PR #241, opened 2021-11-04
feat(dataoverview) Add possibility to add custom slots into dataoverview

tomaszwylezek:
### Summary

Previously when we wanted to add a custom "display" for any key, we had to add it into DataOverview which caused enormous growth of that component. The change will allow passing slot through it - so instead of having all in DataOverview we will be able to leave dependency in a place where it's needed.


bartsmykla (approved):
lgtm

PR #242, opened 2021-11-05
refactor(dataoverview) extract tags

tomaszwylezek:
### Summary

Previously we had tags hardcoded in DataOverview component. This PR will extract it, reducing size of that component and making it reusable in other places.


bartsmykla (approved):
lgtm

PR #243, opened 2021-11-05
fix(meshes) policies bug

tomaszwylezek:
### Summary

In https://github.com/kumahq/kuma-gui/pull/238/files in Meshes was introduced bug which disable the possbility to use "Resource" tab. That PR fixes that and provides test which check if its working.


bartsmykla (approved):
lgtm

PR #248, opened 2021-11-10
fix(*) properly create share url

tomaszwylezek:
### Summary

Previously when clicked on  copy url (attachment) it copied wrong URL (http://localhost:5681#/mesh/default/circuit-breakers?ns=circuit-breaker-all-default)

Right now it should copy the proper URL which allows sharing link to a specific resource.

Screenshot 2021-11-10 at 14 12 56


bartsmykla (approved):
lgtm

PR #249, opened 2021-11-15
fix(*) adjustment of notification/url-share/onboarding alert

tomaszwylezek:
### Summary

It contains upgrade of text in Notification Manager when action available.
Also fix of share-url.


bartsmykla (approved):
lgtm

PR #251, opened 2021-11-15
fix(*) fix initial not generated insights and condition for notificat…

tomaszwylezek:
### Summary

It will fix an issue that happened when no initial insights. Also, provide missing getter for notification manager.


bartsmykla (approved):
lgtm

PR #252, opened 2021-11-16
refactor(api) change format

tomaszwylezek:
### Summary

The reason of this refactor is to align the format of parameters in API calls.
This will allow to easy access a single entity (in case of share-url or one entity table view) no matter if it's mesh-related or not, whereas previously we were limited to mesh-related resources.


bartsmykla (approved):
lgtm

Issue #240, opened 2021-11-03
Disable Datadog Logs and Event collection


austince:
In line with the CNCF guidelines found in https://github.com/kumahq/kuma/issues/3050, we should disable these collections since 1) they are published to Kong's DataDog account and 2) it is not opt-in unless I've missed something about how the initialization works.

 cc @jpeach

jpeach:
@Bradamant3 @subnetmarco

tomaszwylezek:
It will be disabled by default in Kuma 1.4

PR #253, opened 2021-11-18
feat(datadog) set reporting based on flag

tomaszwylezek:
### Summary

It will disable reporting by default and do it based on the flag from config. #240 


bartsmykla (approved):
lgtm

Issue #269, opened 2021-11-29
Add insights to dataplane view


lahabana:
### Description

When accessing a dp in `/gui/#/mesh/all/standard-dataplanes` it would be good to also see the dataplane-insights to debug issues.
This is done for Zones and it's very useful

tomaszwylezek:
Done in #274

PR #274, opened 2021-12-02
feat(dpp) add insights

tomaszwylezek:
### Summary

It will provide insights into the data plane proxy in the new tab. Also, adjust insights for ZoneIngresses

Related to: https://github.com/kumahq/kuma-gui/issues/269

Screenshot 2021-12-02 at 13 03 56


bartsmykla (approved):
lgtm

PR #276, opened 2021-12-08
fix(zones) handle obsolete zone cps

tomaszwylezek:

### Summary

When running multi-zone and having different versions of cp, there might occur a scenario that we are receiving ZoneInsighs which are calculated on obsolete zone - which cause obsolete content of ZoneInsighs - which caused an error as we were expected a value there.
The solution provides an additional check which makes sure that the value is available before trying to process it.


bartsmykla (approved):
lgtm

PR #282, opened 2022-01-04
fix(zones) properly detect ingress when associated

### LINKED ISSUE #280: [Zones] Ingress column displays wrong data
### What happened?

When a user is on the Zones view and has a zone with an ingress, there is information that there is no ingress for that zone.

tomaszwylezek:
### Summary

Fixes #280 and properly display ingress when associated with the zone.


bartsmykla (approved):
lgtm

PR #290, opened 2022-01-19
fix(wizzard) remove ca option from wizzard

tomaszwylezek:
Remove ca option from wizzard which is not available.


bartsmykla (approved):
lgtm

PR #295, opened 2022-01-28
chore(sidebar) simplify logic

tomaszwylezek:
It will simplify logic related to calculating sidebar insights and get rid of issues when we try to access non-existing resources.
Screenshot 2022-01-28 at 13 11 54


bartsmykla (approved):
lgtm

PR #298, opened 2022-01-31
feat(policies): more user friendly inspection of policies

jakubdyszkiewicz:
Instead of cryptic listeners addresses, we added `service` field to API so we can display it in the GUI. I pushed inbound/outbound address to subtitle.

Also, `dataplane` type of attachment was not handled. Name and Service, in this case, is empty, so we want to just display `Dataplane`.

Screenshot 2022-01-31 at 10 19 03

### DIFF
--- a/src/components/DataplanePolicies/DataplanePolicies.vue
+++ b/src/components/DataplanePolicies/DataplanePolicies.vue
@@ -17,11 +17,25 @@
             <template v-slot:accordion-header>
               <div class="flex items-center justify-between">
                 <div>
-                  <p class="text-lg">
-                    {{ item.name }}
+                  <p
+                    v-if="item.type === 'dataplane'"
+                    class="text-lg"
+                  >
+                    Dataplane
+                  </p>
+                  <p
+                    v-if="item.type !== 'dataplane'"
+                    class="text-lg"
+                  >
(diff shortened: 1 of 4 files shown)


src/components/DataplanePolicies/DataplanePolicies.vue
```diff
@@ -17,11 +17,25 @@
...
+                    {{ item.service }}
                   </p>
                   <p class="subtitle">
-                    {{ item.type }}
+                    <span v-if="item.type === 'inbound' || item.type === 'outbound'">
+                      {{ item.type }} {{ item.name }}
+                    </span>
+                    <span v-if="item.type === 'service' || item.type === 'dataplane'">
+                      {{ item.type }}
+                    </span>
```
bartsmykla, line 38:
```suggestion
                    <span v-if="item.type === 'inbound' || item.type === 'outbound'">
                      {{ item.type }} {{ item.name }}
                    </span>
                    <span v-else-if="item.type === 'service' || item.type === 'dataplane'">
                      {{ item.type }}
                    </span>
```

bartsmykla (changes requested):
Just one small suggestion

bartsmykla (approved):
lgtm

PR #300, opened 2022-02-02
feat(*): xds configuration for proxies

jakubdyszkiewicz:
Adds XDS configuration to Dataplane, ZoneIngress and ZoneEgress

image


lahabana (approved):
LGTM I'm going to open an issue to have a "json editor" like container for this so that user can collapse bits of the json and maybe search in the content as XDS conf can become overwhelming

bartsmykla (approved):
lgtm

PR #311, opened 2022-02-21
chore(*): remove v prefix in version header

jakubdyszkiewicz:
We will now release binaries with `v` prefix


lahabana:
Might want to update the mock data so the preview is good?

lahabana (approved):

lobkovilya (approved):

Issue #323, opened 2022-05-11
Consume version check in GUI


parkanzky:
### Description

With https://github.com/kumahq/kuma/issues/4203 we added compatibility checks to kuma CP which validate CP<->DP and Global CP<->Zone CP version compatibility. These are now stored within insights. The GUI still consumes the DP<->Envoy compatibility matrix to check if the DP is compatible with the CP.

* GUI should use the version check boolean in insights to mark entities as incompatible.
* Compatibility matrix and associated web service should be removed.

PR #326, opened 2022-05-18
feat(*): use new compat flags from Insights

lobkovilya:
Use new compat flags from the insights instead of compatibility matrix

Fix: https://github.com/kumahq/kuma-gui/issues/323


lahabana (approved):
This is great! It's good to see these kind of things be moved server-side where they can be done in a sustainable way

PR #333, opened 2022-07-07
feat: execute stats and clusters from kuma cp

jakubdyszkiewicz:
GUI for APIs introduced in [4557](https://github.com/kumahq/kuma/pull/4557).

image


slonka:
Getting an error trying to explore the preview app 
image

src/components/EnvoyStats/EnvoyStats.vue
```diff
@@ -0,0 +1,129 @@
...
+
+  methods: {
+    async fetchStats() {
+      this.hasError = false
+      this.isLoading = true
+
+      try {
+        let stats = ''
+
+        if (this.mesh !== '' && this.dppName !== '') {
```
slonka, line 98: This `else if`-ing on the type of stats we want to fetch looks a bit weird to me. Why don't we know which stats we want to download / how do we make sure that these properties are empty on zoneIngress / zoneEgress stats?

src/components/EnvoyClusters/EnvoyClusters.vue
```diff
@@ -0,0 +1,129 @@
+<template>
```
slonka, line 1: This is pretty much copy-paste of the second one, is there no easy way to re-use this?

slonka (commented):
Not confident enough to 👍 but if someone else feels it's ok to merge I won't stand in your way.

src/components/EnvoyClusters/EnvoyClusters.vue
jakubdyszkiewicz: I consulted this with @tomaszwylezek and his recommendation was to have separate components.
At first, I tried to generify them but ended up with a gigantic switch block of code. I can see how in the future we will improve each component separately. For example, give a better UI on clusters, or put some charts in the stats tab.

jakubdyszkiewicz:
I couldn't get the mocking library to work with nonjson files. That's why there is an error in preview

src/components/EnvoyStats/EnvoyStats.vue
jakubdyszkiewicz: we don't. You can define mesh, dppname, zoneingress and zoneegress and we will fetch then the dppName.

I personally don't see this as a problem since we are the only users of this component, but if someone has suggestions on how to implement this without adding tens of lines of code with validation then I'm happy to change it.

src/components/EnvoyClusters/EnvoyClusters.vue
bartsmykla: I believe this is a good idea for improvement, and as this is the way we are currently doing it everywhere else, I'm fine with it

bartsmykla (approved):
lgtm

slonka:
> I couldn't get the mocking library to work with nonjson files. That's why there is an error in preview

What was the error with the mocking library? maybe @tomaszwylezek or someone more familiar could help?

PR #340, opened 2022-08-05
fix: incorrect version path in docs URLs

kleinfreund:
Fixes URLs to the Kuma documentation website in cases where implementations of the open core versions have a version that differs from the underlying Kuma version.

Introduces a new `kumaDocsVersion` state which is derived from the `basedOnKuma` field found in the API response for `/`.

### DIFF
--- a/src/views/Wizard/views/__snapshots__/Mesh.spec.ts.snap
+++ b/src/views/Wizard/views/__snapshots__/Mesh.spec.ts.snap
@@ -548,7 +548,7 @@ metrics:
                
               <p>
                 <a
-                  href="https://kuma.io/docs/null/policies/mesh/?utm_source=Kuma&utm_medium=Kuma-GUI"
+                  href="https://kuma.io/docs/latest/policies/mesh/?utm_source=Kuma&utm_medium=Kuma-GUI"
                   target="_blank"
                 >
                   
--- a/src/views/__snapshots__/Overview.spec.ts.snap
+++ b/src/views/__snapshots__/Overview.spec.ts.snap
@@ -2179,7 +2179,88 @@ exports[`Overview.vue renders basic snapshot 1`] = `
       </div>
        
       <div>
-        <!---->
+        <div
+          class="kong-card card-item border"
+          data-v-0029352b=""
+        >
+          <div
+            class="k-card-header mb-4"
+            data-v-0029352b=""
+          >
+            <div
+              class="k-card-title"
+              data-v-0029352b=""
+            >
+              <h4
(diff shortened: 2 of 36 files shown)


src/views/Wizard/views/__snapshots__/Mesh.spec.ts.snap
```diff
@@ -548,7 +548,7 @@ metrics:
                
               <p>
                 <a
-                  href="https://kuma.io/docs/null/policies/mesh/?utm_source=Kuma&utm_medium=Kuma-GUI"
```
slonka, line 551: I know it's not a scope of this PR but there are other `nulls` in snapshots, like this - https://github.com/kumahq/kuma-gui/blob/1a9fe64e5591189e7f6ca17e10318d49ea670a47/src/views/__snapshots__/Shell.spec.ts.snap#L18 should I create an issue for it or is it ok in that place?

src/views/__snapshots__/Overview.spec.ts.snap
```diff
@@ -2179,7 +2179,88 @@ exports[`Overview.vue renders basic snapshot 1`] = `
...
+          <!---->
+          <div
+            class="k-card-body"
+            data-v-0029352b=""
+          >
+            <p
+              data-v-0029352b=""
+            >
+              
+      Join the Kuma community and ask questions:
```
slonka, line 2214: where is this defined?

slonka (approved):
Overall looks good, just two questions

src/views/Wizard/views/__snapshots__/Mesh.spec.ts.snap
kleinfreund: I don’t think an issue is necessary. The underlying reason for these incorrect values is that a lot of the unit tests in this project take care of mocking its dependencies/data sources. Most of the tests instead rely on the mock data returned by the tools that mocks all (well, most) API responses. Since not all data always comes from APIs (e.g. some times, data is passed to components via props) and since not all APIs are mocked, this data is just sometimes missing. In this case, the tagline is `null` so this is the result. In any case, this isn’t indicative of an actual bug. It’s just something that could be nicer in the tests. That being said, I don’t think it has priority over other things that need to happen.

src/views/__snapshots__/Overview.spec.ts.snap
kleinfreund: In `src/components/Resources.vue:8`. The string is partially dynamic.

PR #344, opened 2022-08-10
chore(deps): update vue-cli to version 5

kleinfreund:
Updates vue-cli to version 5. This change also updates the internal version of webpack to version 5. It also changes the internal version of jest to version 27. This makes a couple more changes necessary. As a positive side effect, this change set happens to update most of the dependencies for which known security vulnerabilities showed up in `yarn audit`.

Updates the vue.config.js file to provide several libraries as polyfills for Node.js built-ins which were previously pre-bundled with webpack 4 (and therefore vue-cli 4). These libraries are required my msw. As a result, the following new packages were installed: assert, browserify-zlib, https-browserify, os-browserify, process, stream-browserify, stream-http, timers-browserify, tty-browserify, and url. Note that this just makes these explicit dependencies. They were present in the project before, but not as direct dependencies of Kuma GUI.

Updates msw.

Updates a couple of tests that caused the unit test suite to exit with a non-zero exit code as a result of the upgrade of jest. They were missing API call mocks.

Updates the Node.js version in the CircleCI configuration from v12 to v16 because the project’s dependencies no longer support Node.js <16.


bartsmykla (approved):
lgtm

Issue #341, opened 2022-08-08
Dynamically generate policy-related content


kleinfreund:
**TL;DR**: In order to allow Kuma to introduce and change existing policies more easily, Kuma GUI should dynamically generate policy-related content based on available policy types that are discovered via the `/policies` API endpoint.

## Problem statement

Currently, all policies in Kuma GUI are represented via independent and static content or components. Each policy overview page is its own component and these components are pretty much identical to each other safe for some minor differences (e.g. titles, labels, URLs to documentation, etc.).

This leads to a maintenance problem. Changing the UI for the policy overview pages requires changes to all 14 components (one for each policy type). Also, introducing a new policy in Kuma requires introducing one more such component in Kuma GUI and updating all related places in Kuma GUI that deal with showing general policy content (e.g. data plane proxy overviews).

## Solution

By leveraging the `/policies` API endpoint, it’s possible to make this completely dynamic.

### Dynamically generating Vue router routes (non-blocking)

It’s not possible to programmatically add new child routes to an existing route in vue-router@^3. One can only add new top level routes. For this reason, it’s necessary to _delay_ the instantiation of Vue and Vue router until the available policy types have been fetched from the `/policies` endpoint. This is a very quick operation, but in principle, one would like avoid delaying the app instantiation process if possible.

### Missing data on `/policies` endpoint (blocking)

The current state of the `/policies` endpoint is insufficient to completely remove all non-dynamic policy data from Kuma GUI. Specifically, the following things are missing or need to be changed:

- **Missing property `displayNameSingular`**: In some places of the UI, the singular form of the display name is necessary.
  → **Triage outcome**: A new property will be added to the `/policies` endpoint.
- **Missing property `routePath`**: The route paths in Kuma GUI use slightly different strings than what’s provided with the path property (e.g. meshgatewayroutes is the API path and mesh-gateway-routes is the route path). In order to protect people who have bookmarked policies from broken URLs, we should add a routePath property to the endpoint. Alternatively, the route paths in Kuma GUI can be changed. Also, to disambiguate both properties, I recommend renaming `path` to `apiPath`.
  → **Triage outcome**: The `path` property will be used to construct the route paths. We consider breaking these URLs acceptable compared to maintaining Kuma GUI information in Kuma which is awkward.
[...]

PR #346, opened 2022-08-10
feat: dynamically generate policy-related content

kleinfreund:
Adds a new routine before instantiating Vue which loads all available policy types from the `/policies` endpoint. This populates the new `policies`, `policiesByRoute`, and `policiesByType` states which are subsequently used to populate the sidebar menu and set-up the respective routes in the router.

BREAKING CHANGE: Due to generating the routes from the `/policies` endpoint’s response, the route paths for the Mesh Gateways (`#/mesh/all/mesh-gateway-routes` → `#/mesh/all/meshgatewayroutes`) and Mesh Gateway Routes (`#/mesh/all/mesh-gateways` → `#/mesh/all/meshgateways`) views change.

Replaces the individual policy view components (e.g. `CircuitBreakers`, etc.) with one dynamic `PolicyView` component. The individual components were mostly duplicates of each other. This made it hard to maintain because the same changes to the UI would need to be made in all components. Introducing new policies would also require changes to both Kuma and Kuma GUI. Now, depending on which route is used to access a policy view, a prop tells the component which policy type to use and looks up the respective data via the `policiesByRoute` state. Among other things, this data enables the component to construct the necessary API endpoint paths.

Changes the `DataplanePolicies` and `Meshes` components to make use of the new policy states so that they, too, work completely dynamically based on which policies are available in Kuma and which are actually set-up.

Implements #341.


kleinfreund:
**Important**: This is blocked until the redirects for the docs website outlined in https://github.com/kumahq/kuma-gui/issues/341#issuecomment-1209014843 have been implemented.

src/components/DataplanePolicies/DataplanePolicies.vue
```diff
@@ -175,6 +172,30 @@ export default {
           dppName: this.dppName,
         })
 
+        for (const item of items) {
+          item.policyTypes = {}
+
+          // eslint-disable-next-line
```
michaelbeaumont, line 178: Can you add which rule is being ignore?

michaelbeaumont (approved):
LGTM

michaelbeaumont:
@kleinfreund there's an error with Timeouts: https://deploy-preview-346--kuma-gui.netlify.app/#/mesh/all/timeouts

The counts are wrong, but IIRC that's an issue with the mock data, right?

src/components/DataplanePolicies/DataplanePolicies.vue
kleinfreund: Sure, yes, I’ll ignore only the specific rule. The rule is `guard-for-in`.

kleinfreund:
> @kleinfreund there's an error with Timeouts: https://deploy-preview-346--kuma-gui.netlify.app/#/mesh/all/timeouts
> 
> The counts are wrong, but IIRC that's an issue with the mock data, right?

@michaelbeaumont You’re correct, this is an issue with the mock data or rather the absence of it. You can observe the same issue in the current `master` branch. There just isn’t anything mocked for timeouts. I’ve added the necessary data for it now and it’s working.

kleinfreund:
@michaelbeaumont I actually went through the entire application and found _a lot_ of inconsistencies like the one you reported in https://github.com/kumahq/kuma-gui/pull/346#issuecomment-1211199912. I assume that over time, policies, etc. got introduced and not all files were updated accordingly. A lot of places in the UI reported different counts of policies, dataplane proxies, etc. than what their respective detail views showed. I’ve tried my best to make these files consistent so that a view with 5 entries for different meshes actually showed 5 besides their entries in the sidebar.

src/components/DataplanePolicies/DataplanePolicies.vue
```diff
@@ -175,6 +172,30 @@ export default {
           dppName: this.dppName,
         })
 
+        for (const item of items) {
```
slonka, line 175: ok, so this maps policies from the API to our internal representation, right? I'd have it in a separate function or maybe not mutate the data coming from the API but return a new object?

src/views/Policies/PolicyView.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    store,
+    propsData: props,
+  })
+}
+
+describe('PolicyView', () => {
+  test('renders default view correctly', async () => {
+    const wrapper = await createWrapper({ policyPath: 'circuit-breakers' })
+
+    // For some magical reason, this exact arrangement is necessary to wait for the component to completely load. Yikes.
```
slonka, line 33: Can't we just await for some elements to be visible (like puppeteer `waitForSelector`), or [flush promises](https://v1.test-utils.vuejs.org/guides/testing-async-components.html#asynchronous-behavior-outside-of-vue)?

slonka (approved):
Overall seems ok, some comments + one thing that I don't like is mutating existing values, would rather have a copy of them but I'll leave it up to you to decide if you want to change that.

src/components/DataplanePolicies/DataplanePolicies.vue
```diff
@@ -175,6 +172,30 @@ export default {
...
+          item.policyTypes = {}
+
+          // eslint-disable-next-line guard-for-in
```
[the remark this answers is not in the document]
kleinfreund: There isn’t really a good reason to prevent unguarded iteration over a value that you know to be a plain object without a complex prototype chain. In my experience, I’ve never seen a stray prototype value leak into iteration and wreak havoc before. I don’t think this rule should apply in general, but tuning linting rules is a story left for another day.

src/views/Policies/PolicyView.spec.ts
kleinfreund: With Vue 3/Vue test utils 2, we can indeed simply use `flushPromises` which test utils handily provides since then. The current approach is admittedly not great, but then the upgrade is very close on the horizon. I should also just annotate that intend to use `flushPromises` which I did in a similar case elsewhere.

src/components/DataplanePolicies/DataplanePolicies.vue
kleinfreund: Yes, correct. When displaying this content before, we used the original `items` data structure and the methods `getPolicyTitle` and `getPolicyLink` to extract additional data. Calling methods from templates is one of the most common pitfalls the cause additional work: when a component’s template is being re-rendered, all method calls _must_ be executed again because there is no way to tell whether its result changed. By computing this data and adding it to the `items` data structure, this is avoided.

Now as for re-using the `items` data structure as it is returned by the API call, I generally agree with the approach to mutate data. In this case, I don’t see a problem however because `items` is _completely_ local to this function. `items` _is_ completely new and there can also not be any other references to it. So mutating it directly is more direct than creating a new structure. One could make an argument about it being less resilient for certain types of refactorings, but that’s not that pressing of a matter here.

In any case, externalizing this into a separate method is definitely making this a bit cleaner. I’ve done that.

Issue #351, opened 2022-08-16
Gateway -> Policies tab returns error


jakubdyszkiewicz:
### What happened?

Steps to reproduce.
1) Deploy Kuma
2) Deploy demo app
3) Go to GUI, Gateway -> Policies
![image](https://user-images.githubusercontent.com/5459824/184869441-47dec79b-8020-4ee9-a103-bf6c70add6fa.png)

Even that response is 200 from backend
![image](https://user-images.githubusercontent.com/5459824/184869496-a9ba6d85-a73a-4b22-83d7-108fd45371e0.png)

jakubdyszkiewicz:
response from the api

```
{
 "kind": "MeshGatewayDataplane",
 "gateway": {
  "mesh": "default",
  "name": "demo-app"
 },
 "listeners": [
  {
   "port": 80,
   "protocol": "HTTP",
   "hosts": [
    {
     "hostName": "*",
     "routes": [
      {
       "route": "demo-app",
       "destinations": [
        {
         "tags": {
          "kuma.io/service": "demo-app_kuma-demo_svc_5000"
         },
         "policies": {
          "CircuitBreaker": {
           "type": "CircuitBreaker",
           "mesh": "default",
           "name": "circuit-breaker-all-default",
           "creationTime": "2022-08-16T11:27:33Z",
           "modificationTime": "2022-08-16T11:27:33Z",
           "sources": [
            {
             "match": {
              "kuma.io/service": "*"
             }
            }
           ],
           "destinations": [
            {
             "match": {
              "kuma.io/service": "*"
             }
            }
           ],
           "conf": {
            "thresholds": {
             "maxConnections": 1024,
             "maxPendingRequests": 1024,
             "maxRetries": 3,
             "maxRequests": 1024
            }
           }
          },
          "Retry": {
           "type": "Retry",
           "mesh": "default",
           "name": "retry-all-default",
           "creationTime": "2022-08-16T11:27:32Z",
           "modificationTime": "2022-08-16T11:27:32Z",
           "sources": [
            {
             "match": {
              "kuma.io/service": "*"
             }
            }
           ],
           "destinations": [
            {
             "match": {
              "kuma.io/service": "*"
             }
            }
           ],
           "conf": {
            "http": {
             "numRetries": 5,
             "perTryTimeout": "16s",
             "backOff": {
              "baseInterval": "0.025s",
              "maxInterval": "0.250s"
             }
            },
            "tcp": {
[...]

PR #354, opened 2022-08-16
chore: avoid crash in dataplane policies tab

kleinfreund:
Changes the dataplane policies tab to not crash when receiving data associated to mesh gateway dataplanes whose API responses use a wholly different format than what the current code is able to process.

Related to #351.

### DIFF
--- a/src/components/DataplanePolicies/DataplanePolicies.vue
+++ b/src/components/DataplanePolicies/DataplanePolicies.vue
@@ -167,14 +167,19 @@ export default {
       this.isLoading = true
 
       try {
-        const { items, total } = await Kuma.getDataplanePolicies({
+        const { items, total, kind } = await Kuma.getDataplanePolicies({
           mesh: this.mesh,
           dppName: this.dppName,
         })
 
-        this.processItems(items)
-        this.hasItems = total > 0
-        this.items = items
+        // TODO: https://github.com/kumahq/kuma-gui/issues/351
+        // Stops processing responses from the dataplane policies endpoint that are in a new format
+        // because this code isn’t able to correctly process mesh gateway dataplanes.
+        if (kind === undefined || kind === 'SidecarDataplane') {
+          this.processItems(items)
+          this.items = items
+          this.hasItems = total > 0
+        }
       } catch (e) {
         console.error(e)
         this.hasError = true
--- a/src/services/mock/responses/dataplane-policies.json
+++ b/src/services/mock/responses/dataplane-policies.json
@@ -1,4 +1,5 @@
 {
+  "kind": "SidecarDataplane",
   "total": 6,
   "items": [
     {


src/components/DataplanePolicies/DataplanePolicies.vue
```diff
@@ -172,9 +172,14 @@ export default {
           dppName: this.dppName,
         })
 
-        this.processItems(items)
-        this.hasItems = total > 0
-        this.items = items
+        // TODO: https://github.com/kumahq/kuma-gui/issues/351
```
michaelbeaumont, line 175: Can you just switch on the `kind` field here? That's the intention behind the tagged union.

kleinfreund: Theoretically, that could be done, but the current responses for other dataplanes don’t have a `kind` field (see https://github.com/kumahq/kuma-gui/issues/351#issuecomment-1216585108). I’m not sure that improves much.

michaelbeaumont: It definitely improves things, it's the semantics of the field that it's a sum type now and it's backwards compatible with `kind` being unset.

Consider what happens if another response type is added with an `items` array, then the code breaks again. The problem is not that `items` is or is not an array, it's that this code expects a specific variant which we can express by simply checking `.kind === "SidecarDataplane" || .kind === undefined`

michaelbeaumont: I'm pretty sure the current responses have a `kind` set and if not that's a bug in the API server.

kleinfreund: I will need to produce an actual response for different dataplane types to really understand this.

Currently, the mock data indicates a fundamentally different data structure. Checking for `king` being `undefined` would work the same way as checking for `items` being an array. In both cases, we can process `items`. However, I don’t know whether the presence of the `kind` field and it being a different value (e.g. `SidecarDataplane`) has a data structure that the current code can process. That’s why it seems very counter-intuitive to even look at the `kind` field. I extract it from the response and only do something with the `response` if it wasn’t there.

I’ll now try to get a different response from the actual API so I can be confident about the possible formats this endpoint can produce.

michaelbeaumont: We initially had one type of `Dataplane` and these `/policies` responses always had the same structure.

Then we introduced `MeshGateway` and now the response has either:

* `kind: SidecarDataplane` with the exact same structure as before
* `kind: MeshGatewayDataplane` with a new structure

The code (before the refactor) could process only the original structure and therefore it worked if and only if:

* `kind: undefined`/`kind` missing
* `kind: SidecarDataplane`

michaelbeaumont: > Checking for king being undefined would work the same way as checking for items being an array. In both cases, we can process items.

No, because the latter will lead to bugs again if further variants ever have an `items` array. Checking `kind` is more precise and explicit about what we're able to process

kleinfreund: Thanks for explaining this. It really wasn’t obvious to me that a response that _does_ have the `kind` field would _not_ come in the new format. I’ve updated the check as suggested.

michaelbeaumont: Yeah sorry, I realize I left that unsaid. Thanks

michaelbeaumont (approved):

src/components/DataplanePolicies/DataplanePolicies.vue
michaelbeaumont: Maybe there's something we can do with pulling the `pkg/api-server/testdata` in here as mock data to avoid the mock data getting out of date.

Issue #359, opened 2022-08-18
Upgrade to Vue.js 3


kleinfreund:
### Description

**TL;DR**: Upgrade the project to use Vue.js 3.

---

All dependencies in the project have been updated and we can finally upgrade to Vue.js 3.

### vue-meta

The [vue-meta](https://github.com/nuxt/vue-meta) package is not very actively maintained and not yet compatible with Vue 3. However, it is easy to replace because we don’t use a lot of its features.

The short version: In `App.vue`, we can add a watcher like this:

```js
'$route.meta': function (routeMeta) {
  const siteTitle = `${process.env.VUE_APP_NAMESPACE} Manager`

  document.title = routeMeta?.title ? `${routeMeta.title} | ${siteTitle}` : siteTitle
}
```

### vue-prismjs

The [vue-prismjs](https://github.com/vue-formation/vue-prismjs) package is not actively maintained and also not compatible with Vue 3.

This is more complicated to solve since it involves syntax highlighting of code. Specifically, we have a couple of places in the GUI where JSON/YAML code needs to be rendered and highlighted. Here, we also might want to implement our own component since this would give us the flexibility of adding features to it (recently, a grep-like line filtering was brought up).

## Tests

The largest amount of changes will be needed in our unit test suite. That is because we now need to pull more weight when it comes to mocking a component’s dependencies: stores and routes. While this causes a lot of work, it’s good that we need to do this because it will our tests behave more like actual unit tests. Currently, we rely a lot of global application state and mechanisms. We also commonly render the components fully, including all their child components (regardless of source). This gives external dependencies like kongponents a big influence over the stability of the tests. Adding a single attribute to `KButton`, for example, would break all snapshot tests where `KButton` is used (which is really a lot of them).

kleinfreund:
Implemented in #361.

PR #361, opened 2022-08-19
chore(deps): upgrade to Vue.js 3

kleinfreund:
Upgrades the project to Vue.js 3. As part of it, the project is switching to @kong/kongponents@7. The incompatible packages vue-meta and vue-prismjs have been replaced with custom logic.

Removes vue-meta. Instead, some logic was added to App.vue to generate the appropriate value for `document.title` based on the active route.

Removes vue-prismjs. Instead, a new component `CodeBlock` was added. It uses the prismjs library which was added to the project.

Restructures how the mock server is set-up during development to avoid needing to delete msw’s `mockServiceWorker.js` file during builds.

Implements #359.


bartsmykla (commented):
> Info: On the left of the screenshots is always the new, upgraded to Vue.js 3 GUI, and on the right is the old one

## Notifications

1. Enterprise button to tall
image

## Every policy, data plane, service, zone and mesh related view

1. Missing border-bottom and table header to tall
image

## `Data Plane Proxies → All/Standard/Gateway`

1. Table is not scaling down well
image

## Wizard

### Universal ([ref](https://deploy-preview-361--kuma-gui.netlify.app/#/wizard/universal-dataplane))

#### General step

Description paragraphs not styled and Choose Mesh Section has wrong margins/paddings in multiple places
image

#### Topology step

image

#### Networking step

1. image
2. image
3. Protocol select looks like it's disabled, but it's not 
image

#### Install step

1. image
2. image

### Kubernetes

Every issue from Universal applies also for Kubernetes

#### Scope Settings step

image

kleinfreund:
@bartsmykla 

> 1. Enterprise button to tall

This is caused by a change between kongponents 6 and 7 which previously used a rule set to make the icons in buttons a certain smaller size. This rule is gone now and the icons being bigger makes the buttons be bigger. I’ve added a custom rule (they same it used to have) to our styles so that buttons generally appear in the same size.

> 1. Missing border-bottom and table header to tall

This is caused by a change between kongponents 6 and 7. Now every row but the last gets a border bottom. See also https://kongponents.konghq.com/components/table.html. In this case, I’d rather not diverge from the kongponents default style in Kuma GUI.

> 1. Table is not scaling down well

Same here. Table cells now have a `white-space: nowrap` rule which leads to this effect. Instead, tables can be scrolled now. I’ve overwritten this behavior in Kuma GUI now. However, we should perhaps consider making better use of the available screen space. I use a very large monitor and most of the screen is unused. Using a wider layout would avoid this issue entirely: the table wouldn’t need to wrap content _or_ scroll.

> Anything related to the dataplane wizard

This was caused by a change in how CSS is applied to slotted content which used throughout the wizard and some other parts of the UI. Previously, styles from a component with a slot could apply those styles to the slotted content as well. This is no longer possible because slotted content belongs to the component that is _using_ the component with the slot. I’ve addressed those issues now.

lahabana:
LGTM did we check the getting started wizard btw?

bartsmykla:
@lahabana yes, I did and @kleinfreund already fixed my findings

bartsmykla (commented):
## Create mesh wizard
I think spacing is off and the height of select input is different than text inputs
image
image

bartsmykla:
I have not tested Onboarding Wizard yet

kleinfreund:
@lahabana @bartsmykla I’ve tested the onboarding wizard and found a few incorrect colors that I fixed now.

@bartsmykla The create mesh wizard looks like this in the default branch already. There are really quite a few layout/spacing issues that need to be addressed across the UI which I rather do in a separate PR given that this one is already massive.

bartsmykla (approved):
lgtm

bartsmykla:
> @bartsmykla The create mesh wizard looks like this in the default branch already. There are really quite a few layout/spacing issues that need to be addressed across the UI which I rather do in a separate PR given that this one is already massive.

No problem, overall lgtm

kleinfreund:
Turns out that a breaking change in one of our transient dependencies broke our test suite which is why it started to fail. I needed to lock that package’s version. I also updated all other dependencies again.

PR #365, opened 2022-08-24
refactor(linters): reworks linter setup

kleinfreund:
Reworks the linter setup to use only a minimal set of project-specific rules. We already use reliable and well-curated linter rule presets. We should rely on them as much as possible. Alternatively, we could maintain our own rule preset, but then we shouldn’t use third party presets.


lahabana (approved):

bartsmykla (approved):
lgtm

Issue #367, opened 2022-08-25
Rework the way we depend on the gui from Kuma


lahabana:
### Description

We should have an easier and automated way for Kuma to get the latest GUI. We should also avoid having to build and checkin the whole UI in the Kuma repo

kleinfreund:
## Now

Currently, Kuma GUI is updated in Kuma using https://github.com/kumahq/kuma/blob/master/mk/generate.mk#L109-L115 (i.e. check out the latest commit of kuma-gui, run `yarn install` and `yarn build`, and put the generated distribution files in the right place).

This means that as an objective, all that Kuma needs are the latest distribution files of Kuma GUI (latest could mean actually latest or a version referenced by a commit reference of some kind).

## Possible approach (rough draft)

**Upload distribution files on push to Kuma GUI’s default branch**

Instead of building the files in Kuma, we should build them whenever a push to Kuma GUI’s default branch happens (i.e. a PR is merged). Then, we can upload them to some kind of repository. This could be a simple `kuma-gui-latest.tar.gz` archive (alternatively named using a commit hash so we have a stable reference) of the `dist` directory or a built Docker image that’s uploaded to a Docker repository. This depends on how Kuma is actually bootstrapping the GUI. Would a Docker image even be useful here?

## Notes

- It’s probably sensible to upload distribution files with a commit hash (i.e. not with a generic latest name) so Kuma can decide to use an older version in case we need to fix an issue in the GUI but want to stop the bleeding more quickly.

kleinfreund:
## Proposed solution

**Use a GitHub Action to create a PR in Kuma which updates a distribution files**

1. Create the distribution files (`yarn install`, `yarn build`).
2. Check-out Kuma’s default branch and prepare a pull request updating the distribution files.
3. Submit pull request.

In Kuma, an automatic process merges these PRs automatically.

Rollbacks would be done by rolling back changes in Kuma GUI.

kleinfreund:
In https://github.com/kumahq/kuma/pull/4954, I’m removing the upgrade/gui command from Kuma. Besides that, this is now done so I’m closing this as done.

PR #369, opened 2022-08-26
refactor: migrates from CircleCI to GitHub Actions

kleinfreund:
Migrates from CircleCI to GitHub Actions.

In preparation for #367.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,41 @@
+name: main
+
+on: push
+
+jobs:
+  test:
+    # Skip this job if the head commit’s message contains the string “[skip ci]”.
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        # https://github.com/actions/checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Node.js
+        # https://github.com/actions/setup-node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18'
+
+      - name: Cache node_modules and dist
+        # https://github.com/actions/cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            **/node_modules
+            dist
+          key: ${{ runner.os }}-node_modules-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install
+        run: yarn install --frozen-lockfile
+
+      - name: Lint
+        run: yarn run lint
+
+      - name: Test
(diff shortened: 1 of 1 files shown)


kleinfreund:
**Note**: it doesn’t look like workflows introduced via PR are considered when GitHub Actions checks whether to trigger one. I think to test this, the workflow needs to exist in the repository’s default branch.

.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
+name: main
+
+on: push
+
+jobs:
+  test:
+    # Skip this job if the head commit’s message contains the string “[skip ci]”.
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    name: Test
+    runs-on: ubuntu-latest
```
slonka, line 10: Shouldn't we run on the same version that is used to build?

bartsmykla (approved):
lgtm

.github/workflows/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - name: Checkout repository
+        # https://github.com/actions/checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Node.js
+        # https://github.com/actions/setup-node
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18'
```
slonka, line 20: Why are we using different version in `nvmrc` https://github.com/kumahq/kuma-gui/blob/master/.nvmrc#L1 ? shouldn't this be the same?

slonka (commented):
Some questions

.github/workflows/main.yml
kleinfreund: I intend to lift the versions of Node in general, but I want to do this separately for Netlify deployment previews (which use the `.nvmrc` file).

.github/workflows/main.yml
kleinfreund: ubuntu-latest is currently Ubuntu 20.04 (see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners). I don’t see an issue with this changing independent from the Netlify build images because there is really extremely little that could break due to a build image switch in this project. None of the npm packages we use need to be built on the OS plus the `-latest` image won’t just upgrade with the release of a new Ubuntu version as GitHub curates these images like Netlify does.

So, all in all, it’s just one less manual step that we have to do once GitHub deprecates their image for Ubuntu 20.04.

kleinfreund:
I’ve force-pushed to this branch because I needed to drop a commit.

PR #373, opened 2022-08-30
fix: not resetting state when switching between meshes

### LINKED ISSUE #364: Switching meshes dataplane to a failed request and back still fails
### What happened?

This happens when playing with the mesh selector at the top left.
This PR preview is where I've noticed this: https://deploy-preview-361--kuma-gui.netlify.app/#/mesh/hello-world/dataplanes pr #361 

1. Go to: https://deploy-preview-361--kuma-gui.netlify.app/#/mesh/all/dataplanes
2. switch to the hello-world mesh with the selector and see there's an error (likely a missing mock response which is fine)
3. switch back to all meshes with the selector

--> Still fails! If you refresh you get the data that you had in `1.`

kleinfreund:
Fixes an issue with not resetting a component’s boolean states when switching between meshes.

Fixes an issue with setting route params for links whose target routes don’t have params. This is no longer working (see https://github.com/vuejs/router/blob/main/packages/router/CHANGELOG.md#414-2022-08-22).

Updates vue-router explicitly to incorporate the breaking change from https://github.com/vuejs/router/blob/main/packages/router/CHANGELOG.md#414-2022-08-22.

Removes the params from route navigations in the onboarding components as the respective routes don’t actually have this param declared or use it.

Removes some dead code.

Changes the permitted version ranges for TypeScript to only patch versions because apparently TypeScript isn’t semantically versioned (see https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#correctness-changes).

Fixes #364.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "timers-browserify": "^2.0.12",
     "ts-jest": "^28.0.8",
     "tty-browserify": "^0.0.1",
-    "typescript": "^4.7.4",
+    "typescript": "~4.7.4",
     "url": "^0.11.0",
     "webpack-bundle-analyzer": "^4.6.1"
   },
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "semver-compare": "^1.0.0",
     "tailwindcss": "^3.1.8",
     "vue": "^3.2.37",
-    "vue-router": "^4.1.4",
+    "vue-router": "^4.1.5",
     "vuex": "^4.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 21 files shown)


lahabana:
Hmm everything seems to be "loading" when using all meshes

kleinfreund:
@lahabana Where are you seeing this? Are you using the deployment preview? I’ve just checked on several pages and while I can sometimes observe loading states, that’s only happening when data is actually being loaded. Switching between specific meshes and all meshes does also resolve the loading error state for me.

bartsmykla (approved):
lgtm

Issue #362, opened 2022-08-22
More descriptive error when access is missing


jakubdyszkiewicz:
### Description

If you don't have an access to view config dump / stats / clusters, right now backend returns 403 and the GUI says that it could not load the data. We should be more explicit that the access is missing.

jakubdyszkiewicz:
> Triage:
> step 1. present information based on status code
> step 2. (separate ticket?) present errors that are return from the backend

kleinfreund:
I’m working to improve error handling in #374. For API errors with a status code 404, we will now show:

> You currently don’t have access to this data.

![Screenshot from 2022-08-31 14-26-08](https://user-images.githubusercontent.com/5774638/187680920-df926b52-f198-4230-80e0-a36012e8a2fc.png)

Other errors will still show the familiar message unless the API responds with a simple error format. I’m going to check if we already have a format the backend would send in case I need to make adjustments to this logic. Currently, this is pretty much not a concern because we only do GET requests and so there is never a 400 error for validation issues or something like that.

kleinfreund:
With https://github.com/kumahq/kuma-gui/pull/374 being merged, errors in some places of the UI are now representing what we get from the API.

The UI is now able to handle error response formats such as these:

```json
{
  "code": "access_denied",
  "title": "Access denied",
  "details": "You currently don’t have access to this data.",
  "causes": [
    {
      "field": "id",
      "message": "This field is required."
    }
  ]
}
```

(Note: the API actually never includes an error code currently, but the UI will be able to process one should it be introduced.)

![image](https://user-images.githubusercontent.com/5774638/187898240-fc6db158-c7a2-48a0-9bc6-edb2f65ab482.png)

![image](https://user-images.githubusercontent.com/5774638/187898265-04bb2303-066a-4464-9adf-996b9551c282.png)

If no known error response format is found, we still use a custom message for response status code 403.

I’m closing this as done. I’m happy to improve this further, but let’s create new issues for that.

PR #374, opened 2022-08-31
feat: improves error handling

kleinfreund:
Creates a standard error class for API errors and adjusts the `makeRequest` utility to use it. This allows APIs to respond with an error object which will be used to populate the UI with more specific error messages and details.

Adds badges for error code and status code to the `StatusInfo` component which are shown when available (e.g. if we receive an error with code `'access_denied'` and status code `403`, two badges appear in the UI).

See #362.


kleinfreund:
I’ve added logic to handle `causes` on errors and show them in the UI. Currently, this never happens because we only do GET requests and there are never field-specific causes (as far as I know anyway.

image

bartsmykla (approved):
lgtm

PR #376, opened 2022-09-01
chore(.github): don't add date to kuma-gui upgrade PR title, look more like dependabot

michaelbeaumont:
IMO the date in the PR title is too noisy

### DIFF
--- a/.github/workflows/create-gui-pr.yml
+++ b/.github/workflows/create-gui-pr.yml
@@ -1,4 +1,4 @@
-name: create-gui-pr
+name: Update kuma with latest kuma-gui
 
 # Ensures that we only run one workflow per branch at a time.
 # Already running workflows will be cancelled.
@@ -18,7 +18,7 @@ env:
 jobs:
   # Creates a pull request in the main application to update its GUI.
   create-gui-pr:
-    name: Create “update GUI” PR
+    name: Create PR to update kuma-gui
     runs-on: ubuntu-latest
     steps:
       - name: Generate GitHub App token
@@ -29,10 +29,6 @@ jobs:
         with:
           app_id: ${{ secrets.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
-
-      - name: Get current date
-        id: date
-        run: echo "::set-output name=date::$(date +'%Y-%m-%d-%H-%M-%S')"
 
       - name: Check-out GUI
         uses: actions/checkout@v3
@@ -79,14 +75,14 @@ jobs:
           token: ${{ steps.generate-token.outputs.token }}
           path: main-application
           commit-message: |
-            chore(gui): updates GUI
+            chore(deps): bump kuma-gui to ${{ github.sha }}
 
-            Updates GUI.
+            Bumps kuma-gui to version ${{ github.ref_name }}@${{ github.sha }}
           committer: GitHub <noreply@github.com>
           author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
           signoff: true
-          branch: chore/update-gui-${{ steps.date.outputs.date }}
+          branch: chore/update-gui-${{ github.sha }}
           delete-branch: true
-          title: "chore(gui): updates GUI (${{ steps.date.outputs.date }})"
-          body: "Update GUI."
+          title: "chore(deps): bump kuma-gui"
+          body: "Bumps kuma-gui to version ${{ github.ref_name }}@${{ github.sha }}"
           draft: false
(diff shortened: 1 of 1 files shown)


.github/workflows/create-gui-pr.yml
```diff
@@ -1,4 +1,4 @@
-name: create-gui-pr
+name: Update kuma with latest kuma-gui
```
kleinfreund, line 1: I would rather not change this name. It’s the one that shows up in overviews like https://github.com/kumahq/kuma-gui/actions:

image

... and PR status checks like here:

image

kleinfreund (approved):
I don’t agree that the PR title is too noisy, but I don’t have strong feelings about it either, so I’m happy with this change.

It was also a conscious choice to not refer to project-specific terms in names, titles, etc. in order to make this workflow re-usable in other projects without the need to change too much.

.github/workflows/create-gui-pr.yml
```diff
@@ -79,14 +75,14 @@ jobs:
...
+            Bumps kuma-gui to version ${{ github.ref_name }}@${{ github.sha }}
           committer: GitHub <noreply@github.com>
           author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
           signoff: true
-          branch: chore/update-gui-${{ steps.date.outputs.date }}
+          branch: chore/update-gui-${{ github.sha }}
           delete-branch: true
-          title: "chore(gui): updates GUI (${{ steps.date.outputs.date }})"
-          body: "Update GUI."
+          title: "chore(deps): bump kuma-gui"
```
kleinfreund, line 86: **Suggestion**: add `skip ci` hint to skip CI for these PRs (see https://circleci.com/docs/skip-build).

```suggestion
          title: "chore(deps): bump kuma-gui [skip ci]"
```

kleinfreund (changes requested):

.github/workflows/create-gui-pr.yml
```diff
@@ -79,14 +75,14 @@ jobs:
...
-            chore(gui): updates GUI
+            chore(deps): bump kuma-gui to ${{ github.sha }}
 
-            Updates GUI.
+            Bumps kuma-gui to version ${{ github.ref_name }}@${{ github.sha }}
           committer: GitHub <noreply@github.com>
           author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
           signoff: true
-          branch: chore/update-gui-${{ steps.date.outputs.date }}
+          branch: chore/update-gui-${{ github.sha }}
```
kleinfreund, line 84: **Suggestion**: Re-using the same branch will update the PR so we can avoid to close and open new PRs if things get merged in Kuma GUI before the PR in Kuma is merged.

```suggestion
          branch: chore/update-gui
```

kleinfreund (changes requested):

.github/workflows/create-gui-pr.yml
lahabana: Can we put it in the body instead? It's a little noisy otherwise

Also should we add a `> Changelog: updateGUI` in the body so that changelog doesn't get too messy?

.github/workflows/create-gui-pr.yml
michaelbeaumont: Shouldn't they have a readable name?

michaelbeaumont: Or do you mean there's a problem with changing the name in general?

.github/workflows/create-gui-pr.yml
michaelbeaumont: Should we maybe just skip the Changelog entirely here?

.github/workflows/create-gui-pr.yml
kleinfreund: There is no technical problem with changing the name, no. They can be called like you suggested. I just think it’s quite noisy for the status checks, for example, to read “Update kuma with latest kuma-gui / Create PR to update kuma-gui”.

In any case, it’s a preference and a nitpick. I don’t mind to change them, but I personally wouldn’t.

.github/workflows/create-gui-pr.yml
kleinfreund: Yeah, I don’t see a benefit in having a changelog entry for this unless we actually compile the GUI changelog into this PR so that you actually know what changed. As far as troubleshooting goes, the git history will be good enough to see when one of these PRs was merged (though searching one markdown file might be quicker technically).

kleinfreund (approved):

.github/workflows/create-gui-pr.yml
lahabana: Ok so `> Changelog: skip`

michaelbeaumont: That's a good point though, that we probably want to include relevant `kuma-gui` changes into the changelog we publish

PR #377, opened 2022-09-02
refactor: migrates from vue-cli to vite

kleinfreund:
Migrates from vue-cli to vite.

Simplifies Jest configuration and moves it into its own (type-checked) file. All unnecessary options have been removed. All necessary options have been refined to be as minimal as possible (e.g. only transforming file types that we actually use).

Adds file extensions to all .vue imports which became necessary.

Removes a bunch of unused Vue logo files that were left over from the initial project setup.

Simplifies API URL detection to simply use `window.location.origin` in production and `http://localhost:5681` in development.


package.json
```diff
@@ -5,32 +5,33 @@
...
+    "build:ci": "VITE_MOCK_API_ENABLED=true yarn run build --mode development",
+    "lint": "eslint --ext .js,.ts,.vue --fix .",
+    "test:unit": "jest",
     "test:unit:watch": "yarn run test:unit --watch"
   },
   "dependencies": {
     "@amcharts/amcharts4": "^4.10.27",
     "@appscode/json2yaml": "^0.1.2",
-    "@datadog/browser-logs": "^4.17.2",
+    "@datadog/browser-logs": "^4.19.0",
```
slonka, line 20: I know it's not new but should we be collecting logs from the GUI?

src/components/Sidebar/__snapshots__/AppSidebar.spec.ts.snap
```diff
@@ -16,15 +16,18 @@ exports[`AppSidebar.vue renders snapshot 1`] = `
...
         >
           <a>
             <div
               class="nav-icon"
             >
               
-              <!--v-if-->
+              <!-- eslint-disable-next-line vue/no-v-html -->
+              <div>
+                [object Object]
```
slonka, line 29:
```
[object Object]
```

seems to indicate that something was serialised that shouldn't have.

src/configUrl.ts
```diff
@@ -1,30 +1,9 @@
...
-
 export function kumaDpServerUrl(): string {
-  const url = window.location
-  const envConfig = String(process.env.VUE_APP_KUMA_DP_SERVER_URL)
+  const dpServerUrl = String(import.meta.env.VITE_KUMA_DP_SERVER_URL)
 
-  if (process.env.NODE_ENV === 'development') {
-    return envConfig
+  if (import.meta.env.PROD) {
+    return dpServerUrl.replace('localhost', window.location.hostname)
```
slonka, line 5: I think I'm fine with this but in general I'm not really sure 😄. I think I'm most used to a `frontend` having just an env variable called `API_URL` that points to the API. In the future they might not be hosted on the same domain.

src/services/mocks.ts
```diff
@@ -1,146 +1,140 @@
...
-  'meshes/default/dataplanes+insights/cluster-1.backend-03',
-  'meshes/default/dataplanes+insights/cluster-1.ingress-02',
-  'meshes/default/dataplanes+insights/cluster-1.gateway-01',
-  'meshes/default/dataplanes+insights/backend',
-  'meshes/default/dataplanes+insights/frontend',
-  'meshes/default/dataplanes+insights/db',
-  'meshes/default/dataplanes+insights/no-subscriptions',
-
-  'global-insights',
-  'policies',
```
slonka, line 127: have you check in any automated way that you've not left out anything here?

src/views/Onboarding/components/OnboardingPage.vue
```diff
@@ -77,7 +77,7 @@ export default {
   height: 100vh;
   filter: grayscale(1);
   width: 100vw;
-  background: transparent url('@/assets/images/onboarding-background.svg?inline') no-repeat scroll 0% 0%;
+  background: transparent url('@/assets/images/onboarding-background.svg?url') no-repeat scroll 0% 0%;
```
slonka, line 80: why did this change?

src/views/__snapshots__/AppShell.spec.ts.snap
```diff
@@ -100,15 +100,18 @@ exports[`AppShell.vue renders component with route 1`] = `
...
             >
               <a>
                 <div
                   class="nav-icon"
                 >
                   
-                  <!--v-if-->
+                  <!-- eslint-disable-next-line vue/no-v-html -->
+                  <div>
+                    [object Object]
```
slonka, line 113: same as the one above somewhere

yarn.lock
```diff
@@ -10390,13 +7124,3 @@ yocto-queue@^0.1.0:
...
-
-yorkie@^2.0.0:
-  version "2.0.0"
-  resolved "https://registry.yarnpkg.com/yorkie/-/yorkie-2.0.0.tgz#92411912d435214e12c51c2ae1093e54b6bb83d9"
-  integrity sha512-jcKpkthap6x63MB4TxwCyuIGkV0oYP/YRyuQU5UO0Yz/E/ZAu+653/uov+phdmO54n6BcvFRyyt0RRrWdN2mpw==
-  dependencies:
-    execa "^0.8.0"
-    is-ci "^1.0.10"
-    normalize-path "^1.0.0"
-    strip-indent "^2.0.0"
```
slonka, line 10402: I'm guessing the app is going to be a lot lighter with so many dependencies removed from lockfile.

slonka (commented):
Left a couple of comments, in general looks good 👍 will approve after comments are resolved

package.json
slonka: FYI @lahabana ☝️

kleinfreund: I’m not sure we should be collecting logs for Datadog specifically, but I would like to consider _some_ form of error observability like Sentry which in my opinion is really valuable.

src/components/Sidebar/__snapshots__/AppSidebar.spec.ts.snap
kleinfreund: This only happens in tests. You will find this in places where we use SVGs. During tests, SVGs are stubbed using `jest-transform-stub` and this happens to result in `[object Object]` being rendered. Since the SVGs are otherwise rendered fine, I don’t think this is an issue.

src/views/Onboarding/components/OnboardingPage.vue
kleinfreund: Previously, these markers were custom defined in our webpack config (rather the vue.config.js file) and named somewhat arbitrarily. Now, we’re not using webpack anymore and I just used what was provided by vite-svg-loader which now does that job. Different marker, same result.

src/services/mocks.ts
kleinfreund: No, but I’ve changed using my editor’s multi cursor capabilities (i.e. not every line individually by hand).

Previously, this list had 97 entries. Now, it has 104.

9 more because I moved them up (they couldn’t previously be defined there). 1 less because there wasn’t a file for it (meshes/alpha-tango-mesh/circuit-breakers). 1 less because it was defined both in the list and the additional ones below.

97 + 9 - 1 - 1 = 104

src/views/__snapshots__/AppShell.spec.ts.snap
kleinfreund: Deferring this conversation to https://github.com/kumahq/kuma-gui/pull/377#discussion_r966842541.

yarn.lock
kleinfreund: Kind of. A lot of these are related to Node.js built in polyfills that we needed because of the way we imported `msw/node` in our mock code. Since this was a file that eventually got imported in our application, we needed to bundle them, too. But I split that out and could remove all of them. So kind of yes in the sense that no more unnecessary msw stuff is in our bundle and kind of no because it shouldn’t have been in there in the first place. So let’s take this as a yes :D

src/configUrl.ts
kleinfreund: Previously, we had `VUE_APP_KUMA_DP_SERVER_URL=https://localhost:5678/` set in `.env`.

In essence, this function becomes:

```js
function kumaDpServerUrl(): string {
  const url = window.location
  const envConfig = 'https://localhost:5678/'

  if (process.env.NODE_ENV === 'development') {
    return envConfig
  }

  return envConfig.replace('localhost', url.hostname)
}
```

Simplifying this:

```js
function kumaDpServerUrl(): string {
  if (process.env.NODE_ENV === 'development') {
    return 'https://localhost:5678/'
  } else {
    return 'https://localhost:5678/'.replace('localhost', window.location.hostname)
  }
}
```

and further:

```js
function kumaDpServerUrl(): string {
  if (process.env.NODE_ENV === 'development') {
    return 'https://localhost:5678/'
  } else {
    return `https://${window.location.hostname}:5678/`
  }
}
```

So yes, the concern of this not running on the same domain is valid: If it does run from a different domain, requests to this URL would fail.

However, this is already not configurable. You would need to rebuild the frontend application with a different environment variable value and that’s not something a user of Kuma can do. To make this a user choice, this would need to come from the API so we can make it dynamic (probably from `/` or `/config` [I wish the APIs were prefixed with `/api`]).

src/views/Onboarding/components/OnboardingPage.vue
slonka: if the result is the same then 👍

slonka (approved):

package.json
kleinfreund: Turns out we don’t initialize Datadog by default: https://github.com/kumahq/kuma-gui/blob/master/src/datadog.ts#L6-L8

slonka: then great, thanks! 👍

PR #384, opened 2022-09-12
feat: reworks data planes list view

### LINKED ISSUE #382: Rewrite and improve dataplane view
### Description

The dataplane list view is not ideal at the moment especially when you have a lot of DPP.

Let's redesign it:

- Provide a sidebar to the list with a "basic detail view"
- Provide a complete view to see "full details" (Currently what's at the bottom atm).
- Think ahead of future filters (search by dp names, tags ...)
- Make the list view nicer on large numbers of DPPs.
- On smaller screen have the bottom details fixed to the bottom of the viewport (switch from right content to bottom content when the view port becomes small)
- Don't differentiate between standard dpps and gateways but provide an easy way to filter the list.

We'll review with the team what actual content is useful.

kleinfreund:
## Changes

Reworks the data plane list view. The bottom portion of the view detailing the currently selected data plane was removed. As replacement, a condensed data plane summary of the selected data plane was added in the form of a fixed sidebar to the right of the data planes table. Each data plane summary includes a link to a newly added data plane detail view. The data planes table was tweaked to be more space-efficient. Only a few specific tags (kuma.io/service, kuma.io/protocol, and kuma.io/zone) are now represented in the table as separate columns. The sidebar lists all tags.

Adds a column selector to the data plane list view which allows controlling which columns are visible in the table. The selection is stored in the user browser’s local storage (i.e. its device-bound). Eventually, this selector will be added to all data tables in the UI.

Adds a data plane detail view and associated routes. It currently shows the same tabular overview that used to be shown below the data planes table in the list view.

Moves all data plane specific files into `src/data-planes`. This is motivated by wanting to structure the code base by the information hierarchy in the UI.

Creates LoadingBlock, ErrorBlock, and EmptyBlock components for simplifying the use of commonly found components (i.e. the empty states shown when loading or when content is empty or couldn’t be loaded).

Restructures the way the top level elements in the UI (i.e. the main sidebar, main content area, top bar, etc.) are laid out. This was necessary to gain a better control over the layout possibilities for the changes made to the data planes list view. As a nice side effect, a lot of unused or unnecessary code could be removed (e.g. controlling the hover state of the main navigation via CSS instead of JS).

Updates Kongponents to the latest version. This is necessary because some bug fixes made in between are necessary for the changes to the data plane views to work properly. Unfortunately, this made it necessary to regenerate all snapshots which accounts for the biggest number of files changed in this commit.

### With mock data

```sh
yarn run dev
```

### With fake data planes

Terminal window 1:

```sh
cd kuma
KUMA_DP_SERVER_AUTH_TYPE=none make run/universal/memory
```

Terminal window 2:

```sh
cd kuma
make run/xds-client
```

Terminal window 3:

```sh
cd kuma-gui
yarn run dev:with-api
```

## Screenshots

<details>
<summary>Before</summary>

Screenshot from 2022-09-14 10-39-49

</details>

<details>
<summary>After</summary>

image

</details>

Resolves #382.


kleinfreund (commented):
**Question**: Should data planes in the list view be sorted by associated service first?

lahabana:
In dataplane detail view. The link in the breadcrumb to "Data plane" is broken.

lahabana:
Should we get rid of "Entity Overview for backend" in the sidebar view I'm not sure it brings much

lahabana:
In the details view. When clicking yaml can we just call the `dataplane` endpoint to just retrieve the actual dataplane instead of the dataplane overview?

lahabana:
Ok how about a follow up fix to the detail view (I think YAML should just go in `general` and tabs should be reflected in the url)

src/data-planes/views/DataPlaneListView.vue
```diff
@@ -0,0 +1,654 @@
...
+      type: 'all',
+    },
+    {
+      label: 'Standard',
+      type: 'Standard',
+    },
+    {
+      label: 'Gateway',
+      type: 'Gateway / BUILTIN',
+    },
```
lahabana, line 165: Are we missing provided gateways here?

src/data-planes/views/DataPlaneListView.vue
```diff
@@ -0,0 +1,654 @@
...
+        name: 'mesh-child',
+        params: {
+          mesh,
+        },
+      }
+
+      /**
+       * Handle our tag collections based on the dataplane type.
+       */
+      const tags = dpTags(dataplane).filter((tag) => IMPORTANT_DATA_PLANE_TAG_LABELS.includes(tag.label))
```
lahabana, line 362: What's the of extracting this as a constant when the 3 entries in your constant are used just after? Seems like this would be better inlined.

src/dataplane.ts
```diff
@@ -1,70 +1,88 @@
...
+ * List of data plane tag labels that are considered important enough to be shown in the data plane list view.
+ */
+export const IMPORTANT_DATA_PLANE_TAG_LABELS: string[] = [
+  'kuma.io/protocol',
+  'kuma.io/service',
+  'kuma.io/zone',
+]
+
+/**
+ * Takes a data plane and constructs the list of tags. It flattens common tags so we don't display them twice.
```
lahabana, line 27: This seems to contradict your example

lahabana (approved):
Just some minor comments

src/dataplane.ts
kleinfreund: The way this utility was implemented considers uniqueness of tags as label-value pairs meaning that a network inbound tag `kuma.io/service: frontend` and a network outbound tag `kuma.io/service: backend` are distinct.

kleinfreund:
@lahabana 

> In dataplane detail view. The link in the breadcrumb to "Data plane" is broken.

What’s broken exactly? I fixed one issue where the URL still contained a `name` parameter when navigating away from the list view, but that seems unrelated.

> Should we get rid of "Entity Overview for backend" in the sidebar view I'm not sure it brings much

I agree; I don’t care much for it and in most places it’s really just noisy. I removed it.

> In the details view. When clicking yaml can we just call the `dataplane` endpoint to just retrieve the actual dataplane instead of the dataplane overview?

I’ve reworked the detail view to generally use the `Dataplane` entity which also addresses this point.

---

What’s your opinion on https://github.com/kumahq/kuma-gui/pull/384#pullrequestreview-1113809561?

src/data-planes/views/DataPlaneListView.vue
kleinfreund: I’ve added an entry for `Gateway / PROVIDED`. I *assume* this is how the type would be constructed in the UI though I’m not sure because there is no mock data that features this type. I suppose the `gateway.type` field would be set to `'PROVIDED'` for those?

src/data-planes/views/DataPlaneListView.vue
kleinfreund: I’m not sure I’m following.

Do you mean to suggest to look up service, protocol, and zone on `dataplane` directly rather than constructing the tag list and then finding them again?

lahabana: I'm just saying the global var seems not needed here. It's just asking the reader of the code to context switch when it's only used here and the 3 following lines would need to evolve accodingly to this global var.

src/dataplane.ts
lahabana: Ok might be worth clarifying the comment.

lahabana:
> Question: Should data planes in the list view be sorted by associated service first?

Good question I think you should keep as is for now. It would be nice to have it sorted but I'd do this separately

lahabana:
What I meant about breadcrumbs is here:

image


I'd expect that clicking on `Data plane` would redirect to the list view.

kleinfreund:
@lahabana

> I'd expect that clicking on `Data plane` would redirect to the list view.

This is now fixed.

src/data-planes/views/DataPlaneListView.vue
kleinfreund: Ah, now I understand. Done.

src/dataplane.ts
kleinfreund: I’ve added a note to it.

src/data-planes/views/DataPlaneListView.vue
kleinfreund: A few questions:

- I’ve seen some test examples in Kuma that suggest that `gateway.type` isn’t typically all uppercase. However, this is the case in our mock data. If it’s lowercase, I’m going to change the mock data.
- “Gateway / BUILTIN” looks strange to me. I suggest a slightly different format for the UI: “Gateway (builtin)” (and “Gateway (provided)”). What do you think?

lahabana: These are the 2 different ways to run gateways: https://kuma.io/docs/1.8.x/explore/gateway/

Going to try to find enough time to start the 2 gateways and get you the API output. FTM I think it's fine to merge as is and open a follow up issue if it's addressed before 2.0

PR #389, opened 2022-09-22
feat: adds service detail views

### LINKED ISSUE #385: Create service detail view
### Description

(Split-off from https://github.com/kumahq/kuma-gui/issues/383 so keep a better work scope and simplify reviews)

Add a service detail view and add links to it from all tables that currently have a service name column (e.g. data planes, services).

kleinfreund:
Adds detail views for service insights and external services.

Adds links to the new service insight detail views in the data plane list view.

Adds some basic logic that enables components to update the document title with more specific information (e.g. using the name of a service in the title).

Corrects a couple of issues with the mock data for services.

Adds some type annotations to API wrappers.

Resolves #385.


lahabana (approved):
LGTM

image

I didn't notice this on the previous PR but the fact that these are buttons endup padding the text in a weird way. Could we do better?

kleinfreund:
@lahabana

> I didn't notice this on the previous PR but the fact that these are buttons endup padding the text in a weird way. Could we do better?

You’re right. I’m not very fond of this look either. I kept it to avoid introducing much deviation from the basic building blocks the component library provides, but really, it’s more unsettling in this case. I turned all links inside tables into plain links as they would appear in other places, too.

PR #392, opened 2022-09-28
feat: moves YAML tab content into overview tab

### LINKED ISSUE #386: Move YAML tab content into Overview tab
### Description

For views with a tab widget which has a YAML tab, move its contents into the Overview tab.

kleinfreund:
## Changes

Moves the content of the YAML tab (universal and kubernetes configurations) into the overview tab so that they’re accessible more quickly.

Simplifies breadcrumbs component.

Resolves #386.

## Screenshots

image

### DIFF
--- a/src/views/Policies/PolicyView.vue
+++ b/src/views/Policies/PolicyView.vue
@@ -100,6 +100,15 @@
               </ul>
             </div>
           </LabelList>
+
+          <div class="config-wrapper">
+            <YamlView
+              :has-error="entityHasError"
+              :is-loading="entityIsLoading"
+              :is-empty="entityIsEmpty"
+              :content="rawEntity"
+            />
+          </div>
         </template>
 
         <template #affected-dpps>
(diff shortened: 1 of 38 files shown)


src/views/Policies/PolicyView.vue
```diff
@@ -327,6 +338,8 @@ export default {
 
 <style lang="scss" scoped>
 .config-wrapper {
-  padding: var(--spacing-md);
+  padding-right: var(--spacing-md);
```
slonka, line 341: Seen this padding 3 times, does it make sense to dry up?

slonka (approved):
Maybe I'm reviewing too much JS lately but I couldn't find anything obviously wrong here.

src/views/Policies/PolicyView.vue
kleinfreund: In CSS specifically, I prefer only setting properties that I truly want to set, so I will set three padding rules instead of using the shorthand where the forth property I don’t care for is set to 0. The primary reason for this is to avoid unintentionally overriding styles that are introduced by other rule sets or via the cascade.

For margins and paddings, this is perhaps overly cautious, but a good example is the `background` shorthand. Setting `background` instead of `background-color` sets **every single** property `background` is a shorthand for so specifying `background: tomato` sets `background-color: tomato` but also removes background images and everything else that might be set elsewhere.

PR #396, opened 2022-10-05
ci: add ci/skip-test label to upgrade PR and better message

lahabana:
This includes a link to the repo which is always useful

Signed-off-by: Charly Molter <charly.molter@konghq.com>

### DIFF
--- a/.github/workflows/create-gui-pr.yml
+++ b/.github/workflows/create-gui-pr.yml
@@ -75,19 +75,16 @@ jobs:
           token: ${{ steps.generate-token.outputs.token }}
           path: main-application
           commit-message: |
-            chore(deps): bump kuma-gui to ${{ github.sha }}
+            chore(deps): bump ${{ github.repository }} to ${{ github.sha }}
 
-            Bumps kuma-gui to version ${{ github.ref_name }}@${{ github.sha }}
-
-            [skip ci]
+            Bumps ${{ github.repository }} to version [${{ github.ref_name }}@${{ github.sha }}](https://github.com/${{ github.repository }}/tree/${{ github.sha }})
           committer: GitHub <noreply@github.com>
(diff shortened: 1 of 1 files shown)


michaelbeaumont (approved):

.github/workflows/create-gui-pr.yml
```diff
@@ -75,19 +75,16 @@ jobs:
...
-
-            [skip ci]
+            Bumps ${{ github.repository }} to version [${{ github.ref_name }}@${{ github.sha }}](https://github.com/${{ github.repository }}/tree/${{ github.sha }})
           committer: GitHub <noreply@github.com>
           author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
           signoff: true
           branch: chore/update-gui
           delete-branch: true
-          title: "chore(deps): bump kuma-gui"
+          title: "chore(deps): bump ${{ github.repository }} to ${{ github.sha }}"
```
kleinfreund, line 86: **Suggestion**: I think this should not refer to the git hash

```suggestion
          title: "chore(deps): bump ${{ github.repository }}"
```

I’m not sure whether the action would update the title of an existing PR, but in any case, I think it’s too verbose here.

kleinfreund (approved):

.github/workflows/create-gui-pr.yml
michaelbeaumont: I kind of agree, though I see the potential use of being able to search for the PR that updated to some hash

lahabana: It does update the title.

Moreover the PR title is what we use as first of git commit see: https://github.com/kumahq/kuma/commit/6b22e376f41d02665fd891366e04720ca995f6a9

kleinfreund: I don’t feel strongly about this so I’m happy to keep it.

PR #405, opened 2022-10-13
feat: removes all meshes scope

### LINKED ISSUE #397: Always be inside a mesh
### Description

At the moment we sometime see everything regardless of mesh.

We should not do this and always pick a mesh (potentially store it in local storage so you don't always have to reselect the one you are actually interested in).

kleinfreund:
## Changes

Removes the “all meshes” scope (i.e. the `all` option in the mesh selector) from the GUI entirely and restructures the navigation accordingly. The home page that was previously called “Overview” is now labeled “Home”. A new category “Mesh” containing a mesh overview, services, data plane proxies, and policies was added. The “Meshes” page has been removed and its content was moved into the mesh-specific “Overview” page. The mesh selector was also moved into the mesh-specific section. Zone entries (and zone charts in the main overview) aren’t shown anymore when the GUI is running in standalone mode.

Resolves #397.

## Screenshots

<details>
<summary>Before:</summary>

Screenshot from 2022-10-13 13-43-05

</details>

<details>
<summary>After:</summary>

image

</details>


lahabana:
1. This page: https://deploy-preview-405--kuma-gui.netlify.app/#/ is contextual on the mesh you select. I think it shouldn't it should always use all the data across all meshes. It should also likely show a number of meshes chart too.
2. Can we hide the "Zone" from the sidebar if running in Standalone mode?
3. Can we have what is in https://deploy-preview-405--kuma-gui.netlify.app/#/meshes (.i.e overview and resources) just be in https://deploy-preview-405--kuma-gui.netlify.app/#/mesh/default and get rid of the "meshes" entry in the sidebar?
4. These changes would enable us to have a top level of: `OVERVIEW`, `ZONES`, `MESHES` which would make things IMO clearer and remove the weirdness with "GLOBAL" as a top level section.

lahabana:
1. This page: https://deploy-preview-405--kuma-gui.netlify.app/#/ is contextual on the mesh you select. I think it shouldn't it should always use all the data across all meshes. It should also likely show a number of meshes chart too.
2. Can we hide the "Zone" from the sidebar if running in Standalone mode?
3. Can we have what is in https://deploy-preview-405--kuma-gui.netlify.app/#/meshes (.i.e overview and resources) just be in https://deploy-preview-405--kuma-gui.netlify.app/#/mesh/default and get rid of the "meshes" entry in the sidebar?
4. These changes would enable us to have a top level of: `OVERVIEW`, `ZONES`, `MESHES` which would make things IMO clearer and remove the weirdness with "GLOBAL" as a top level section.

kleinfreund:
@lahabana Those are great suggestions.

> * This page: https://deploy-preview-405--kuma-gui.netlify.app/#/ is contextual on the mesh you select. I think it shouldn't it should always use all the data across all meshes. It should also likely show a number of meshes chart too.

Right, I messed up in not fetching the relevant data by mesh here. The notion of "all meshes" was quite tightly integrated into some places. I’ve fixed this. I’ll add an number of meshes chart later.

> * Can we hide the "Zone" from the sidebar if running in Standalone mode?

Do you mean the entire Zones category or just specific entries here? For now I’ve removed the entire category when not in multizone mode. **Also, should the zone charts be removed in the main overview when not in multizone mode**?

> * Can we have what is in https://deploy-preview-405--kuma-gui.netlify.app/#/meshes (.i.e overview and resources) just be in https://deploy-preview-405--kuma-gui.netlify.app/#/mesh/default and get rid of the "meshes" entry in the sidebar?

Yep, that’s a good idea. I thought about it, too. 😁 I’ve implemented this now.

> * These changes would enable us to have a top level of: `OVERVIEW`, `ZONES`, `MESHES` which would make things IMO clearer and remove the weirdness with "GLOBAL" as a top level section.

This is now much cleaner. I think one item to consider now is whether we can somehow do without having two Overview items in the navigation.

lahabana:
> Do you mean the entire Zones category or just specific entries here? For now I’ve removed the entire category when not in multizone mode. Also, should the zone charts be removed in the main overview when not in multizone mode?

Yes it should also be removed good catch!

lahabana:
>  I’ll add an number of meshes chart later.

Np make sure not to forget that's all :)

lahabana:
> This is now much cleaner. I think one item to consider now is whether we can somehow do without having two Overview items in the navigation.

Rename the top level `Overview` to `Home` or just remove completely and have folks click on the logo to access this?

kleinfreund:
@lahabana I’ve added a number of meshes chart and removes the zone-related charts when not in mutlizone mode. The top-level overview is now called Home. Let’s see how this works out for people.

lukidzi:
Looks much better 🎉 , I have only one question. Is it possible to change maybe style of the window where we choose the mesh? It seems to not fit with the rest. 
image

kleinfreund:
> Looks much better  , I have only one question. Is it possible to change maybe style of the window where we choose the mesh? It seems to not fit with the rest.

@lukidzi The dialog that opens for this select widget _cannot_ be styled whatsoever and for simple use cases like our mesh selector, it’s not worth the complexity to implement custom widgets that address this. For more complex use cases, we have solutions in the Kongponents library that we can use.

lahabana:
>  For more complex use cases, we have solutions in the Kongponents library that we can use.

Is it that complicated to use a kongponent dropdown?

kleinfreund:
> Is it that complicated to use a kongponent dropdown?

No, it’s not, but it’s overkill when one just needs the basics of a select widget. I’d only use a custom dropdown implementation when its contents need to be searchable, the options should contain images, or one needs a multiselect-style select (the native HTML control for that is terrible).

kleinfreund:
I’ve also noticed that I recently accidentally committed a package-lock.json file which I removed now (presumably I accidentally ran `npm install` instead of `yarn install`).

lahabana (approved):

PR #407, opened 2022-10-18
feat: refreshes sidebar counters periodically

### LINKED ISSUE #327: The counters in the sidebar are not refreshed when the table refresh button is pressed
### What happened?

In the picture below, 6 standard DDPs are seen in the table, but the counter next to "Standard" in the side bar still shows "3" instead of "6".

Screen Shot 2022-05-18 at 3 34 13 PM

kleinfreund:
Adds a polling utility for refreshing the counters in the sidebar periodically. The polling interval is set to 10 seconds.

Resolves #327.


lahabana (approved):
Looks good to me. Is there anyway to skip the polling if the tab is not active? This is just to avoid hammering the api for no reason.

kleinfreund:
> Is there anyway to skip the polling if the tab is not active? This is just to avoid hammering the api for no reason.

Yep, I can add logic for this. We could also use exponential back off (while the tab is active) if we want.

lahabana:
I don't think exponential backoff is what we want though (There's no reason less change happens because you've been on the tab for a longer time).

kleinfreund:
> I don't think exponential backoff is what we want though

I agree; I wasn’t planning on adding it anyway.

I’ve added the logic now for disabling the polling while the window isn’t active.

PR #430, opened 2022-11-02
refactor: explicitly import components

kleinfreund:
Removes the global registration of Kongponents components which forces components to explicitly import them. This removes the need for tests to explicitly register them.


lahabana (approved):
Specific dependencies rather than global dependencies sounds like an improvement!

PR #431, opened 2022-11-02
chore(deps): updates all dependencies

kleinfreund:
Updates all dependencies. Addresses one transient dependency having a critical security vulnerability.


kleinfreund:
**Comment**: I will wait with merging this until https://github.com/Kong/kongponents/pull/872 was merged and published. I will then update Kongponents in this PR again.

package.json
```diff
@@ -16,73 +16,73 @@
     "test": "cross-env TZ=UTC jest",
     "test:watch": "yarn run test --watch"
   },
+  "engines": {
+    "node": "^18.0.0",
```
slonka, line 20: Should we add this to `/.nvmrc` as well?

package.json
```diff
@@ -16,73 +16,73 @@
...
+    "ts-jest": "^29.0.3",
+    "typescript": "~4.8.4",
+    "vite": "^3.2.2",
     "vite-svg-loader": "^3.6.0"
   },
   "browserslist": [
     "> 1%",
     "last 2 versions"
   ],
-  "engines": {
```
slonka, line 82: Just noticed that it was moved.

slonka (approved):
Looks good, the only thing that I'm curious about is some elements missing - `class="k-table-container"`.

package.json
kleinfreund: Some tools look at the .nvmrc (like Netlify), but I’d prefer this alongside the project dependencies here because those _are_ effectively project dependencies. Tools like npm, yarn, and friends will check against values in `engines`, too, so it’s good to have it here.

kleinfreund:
> the only thing that I'm curious about is some elements missing - class="k-table-container".

@slonka They’ve been added not removed.

slonka:
> They’ve been added not removed.

Yes, but that can break things as well. Just checked one of the tables and it looks the same so 👍

PR #432, opened 2022-11-02
refactor: uses vue router’s history mode

kleinfreund:
Switches from hash history mode to web history mode. Most notably this changes the GUI to have true URL paths without a `/#/` segment which allows us to use fragment identifiers ourselves.

Moves the vue router boilerplate to a Jest setup after env file which allows us to remove it from all tests.

Moves the vuex boilerplate to a Jest setup after env file which allows us to remove it from all tests.

### DIFF
--- a/jest/jest-setup.ts
+++ b/jest/jest-setup.ts
@@ -22,19 +22,7 @@ dotenv.config()
 Object.defineProperty(window, 'SVGPathElement', { value: class extends HTMLElement { } })
 
 /**
- * Kongponents v6 uses a call to `crypto.getRandomValues` which isn’t implemented in jsdom by default.
- * To make the tests run, we supply it via Node’s `crypto` library.
- *
- * See: https://github.com/jsdom/jsdom/issues/1612
- */
-Object.defineProperty(global.self, 'crypto', {
-  value: {
-    getRandomValues: (array: []) => require('crypto').randomBytes(array.length),
-  },
-})
-
-/**
- * Kongponents v6 uses generated UUIDs for several attribute values.
+ * Kongponents uses generated UUIDs for several attribute values.
  * This breaks the project’s snapshot tests since they’re based on fully-mounted components
  * which also includes those from external sources like Kongponents.
  *
--- a/jest/jest-setup.ts
+++ b/jest/jest-setup.ts
@@ -1,50 +1,8 @@
-import '@testing-library/jest-dom/extend-expect'
-import '@testing-library/jest-dom'
-
-// Polyfills `window.fetch` for Jest because it runs in a Node environment where fetch isn’t available. It initially looked like this would change with Node.js 18, but that is not so.
-import 'isomorphic-fetch'
-
 import dotenv from 'dotenv'
-
-import { replaceAttributesSnapshotSerializer } from './jest-replace-attribute-snapshot-serializer'
-import { setupMockServer } from '../src/api/setupMockServer'
 
 /**
  * Ensures the tests run with the project’s environment variables being available.
+ *
+ * It’s important to run the environment variable setup as part of a Jest setup file and before a Jest setup after env file because the latter might statically import a file that makes use of environment variables which wouldn’t be correctly resolved at the time the static imports executed.
  */
 dotenv.config()
-
-/**
- * amcharts need SVGPathElement defined for our tests to work.
- *
- * See: https://github.com/amcharts/amcharts4/issues/1387
- */
-Object.defineProperty(window, 'SVGPathElement', { value: class extends HTMLElement { } })
-
-/**
- * Kongponents uses generated UUIDs for several attribute values.
- * This breaks the project’s snapshot tests since they’re based on fully-mounted components
- * which also includes those from external sources like Kongponents.
- *
- * In order to stabilize the tests which otherwise fail because the attribute values are different every run,
- * we use a custom snapshot serializer to replace those attribute values with one fixed value.
- */
-expect.addSnapshotSerializer(replaceAttributesSnapshotSerializer([
-  'id',
-  'aria-describedby',
-  'aria-labelledby',
-  'aria-controls',
-  'data-tableid',
-]))
-
-const server = setupMockServer(import.meta.env.VITE_KUMA_API_SERVER_URL)
-
-// Establish API mocking before all tests.
-beforeAll(() => server.listen())
-// Reset any request handlers that we may add during the tests,
-// so they don't affect other tests.
-afterEach(() => server.resetHandlers())
-// Clean up after the tests are finished.
-afterAll(() => server.close())
-
-export { server }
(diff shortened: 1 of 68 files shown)


jest/jest-setup.ts
```diff
@@ -1,62 +1,8 @@
...
  */
-Object.defineProperty(window, 'SVGPathElement', { value: class extends HTMLElement { } })
-
-/**
- * Kongponents v6 uses a call to `crypto.getRandomValues` which isn’t implemented in jsdom by default.
- * To make the tests run, we supply it via Node’s `crypto` library.
- *
- * See: https://github.com/jsdom/jsdom/issues/1612
- */
-Object.defineProperty(global.self, 'crypto', {
```
slonka, line 30: I manually diffed the `jest-setup.ts` and `jest-setup-after-env.ts` - and I think we're missing this. Do we not need it any more or is this a bug?

src/app/App.vue
```diff
@@ -78,7 +78,7 @@ const isLoading = ref(store.state.globalLoading)
 const routeKey = computed(() => route.meta.shouldReRender ? route.path : 'NONE')
 const shouldShowAppError = computed(() => store.state.config.status !== 'OK')
 const shouldSuggestOnboarding = computed(() => store.getters['onboarding/showOnboarding'])
-const shouldShowNotificationManager = computed(() => !shouldSuggestOnboarding.value && store.getters['notifications/amountOfActions'] > 0)
+const shouldShowNotificationManager = computed(() => store.getters['notifications/amountOfActions'] > 0)
```
slonka, line 81: why are we not taking `shouldSuggestOnboarding.value` into account any more?

src/app/wizard/views/DataplaneUniversal.spec.ts
```diff
@@ -1,27 +1,10 @@
-import { createRouter, createWebHashHistory } from 'vue-router'
 import { rest } from 'msw'
 import { render } from '@testing-library/vue'
 import userEvent from '@testing-library/user-event'
 
 import DataplaneUniversal from './DataplaneUniversal.vue'
 import { store, storeKey } from '@/store/store'
```
slonka, line 6: nitpick: not introduced in this PR but if GH action is right and we can remove it then let's do it (like you did in ZoneIngresses).

src/store/modules/notifications/notifications.spec.ts
```diff
@@ -1,23 +1,38 @@
-import { createStore } from 'vuex'
+// import { createStore } from 'vuex'
```
slonka, line 1: if it's really not needed this should be removed

src/store/modules/sidebar/sidebar.spec.ts
```diff
@@ -1,15 +1,26 @@
...
+        sidebarModule.mutations[type](state, status)
+      }
+
+      function dispatch(type: string): void {
+        // @ts-ignore I can’t be bothered to battle Vuex’s loose types right now.
+        return sidebarModule.actions[type]({ commit, state, rootState })
+      }
 
-      await store.dispatch('sidebar/getInsights')
+      // @ts-ignore I can’t be bothered to battle Vuex’s loose types right now.
```
slonka, line 20: should we create an issue for this?

slonka (changes requested):
Looks good apart from one place where there is commented out code. I'll click around the gui now.

jest/jest-setup.ts
kleinfreund: Yep. I checked whether some of the things in here were still necessary and this wasn’t.

Automaat (approved):

src/app/App.vue
kleinfreund: I don’t know why it was there in the first place because it’s not relevant here.

The notification manager is only invoked by user interaction from the top bar so this doesn’t mean to show the modal right away, it just _allows_ it do be shown in the first place (i.e. when the user clicks on the notification icon in the top bar).

slonka:
Old links do not work now (e.g. https://deploy-preview-432--kuma-gui.netlify.app/#/mesh/default/circuit-breakers does not link anywhere), and some people can have direct links to something in guides / tutorials / etc. We should document this in "upgrading" and also probably announce during the next release.

slonka:
I clicked around the GUI and the router seems to work so 👍 from that side

src/store/modules/sidebar/sidebar.spec.ts
kleinfreund: We can’t really fix this in a fashion that wouldn’t be more annoying than leaving this (also annoying) comment. I tried manually setting a narrower type for the actions that only accepts action _handlers_ and not action _objects_ (this union is the cause of the TypeScript error I’m suppressing here), but then the `this` reference of the internal Vuex type structures breaks and I don’t know how to fix it. I tried a few things and in the end it was the least amount of friction to just `@ts-ignore` it.

Without exception all our actions are functions and not plain objects so while this is a true type error, the code path isn’t actually broken.

kleinfreund:
> Old links do not work now (e.g. https://deploy-preview-432--kuma-gui.netlify.app/#/mesh/default/circuit-breakers does not link anywhere), and some people can have direct links to something in guides / tutorials / etc. We should document this in "upgrading" and also probably announce during the next release.

@slonka I added redirects for old-style URL paths in https://github.com/kumahq/kuma-gui/pull/432/commits/af448b2bba399f92d902f29376d6711d2659ccbc. Sorry, I accidentally rebased while editing commits so needed to force-push.

src/store/modules/notifications/notifications.spec.ts
kleinfreund: Cleaned this up.

src/app/wizard/views/DataplaneUniversal.spec.ts
kleinfreund: Cleaned this up.

slonka:
I thought we agreed on not force pushing between reviews :(

kleinfreund:
> I thought we agreed on not force pushing between reviews :(

@slonka My bad. It happens.

src/api/mocks.ts
```diff
@@ -10,12 +10,13 @@ async function loadMockFile(importFn: () => Promise<any>): Promise<any> {
 
 function getBaseInfo(): Info {
   return {
-    hostname: 'Tomaszs-MacBook-Pro-16-inch-2019',
+    hostname: 'control-plane-5d94cb99c6-rzr96',
     tagline: import.meta.env.VITE_NAMESPACE,
     version: '1.7.1',
```
slonka, line 15: nitpick: we could just update to 2.0.0 😉

slonka (approved):
Looks good 👍

src/api/mocks.ts
kleinfreund: Sure, we could, but it’s not really relevant. It only affects using mock data and gives us a fuzzy way to see if URL links to older docs sources still work.

PR #436, opened 2022-11-07
chore(data-planes): expands test coverage

kleinfreund:
* chore(data-planes): expands test coverage

  Expands the test coverage for data plane views.
* chore(api): parameters incorrectly being optional
* chore: removes non-standard public/private fields

### DIFF
--- a/src/app/data-planes/views/DataPlaneListView.spec.ts
+++ b/src/app/data-planes/views/DataPlaneListView.spec.ts
@@ -1,54 +1,63 @@
 import { flushPromises, mount } from '@vue/test-utils'
 
 import DataPlaneListView from './DataPlaneListView.vue'
-import { store } from '@/store/store'
 import { router } from '@/../jest/jest-setup-after-env'
 
 async function renderComponent() {
-  router.currentRoute.value.name = 'home'
-  router.currentRoute.value.params.mesh = 'default'
-
-  await store.dispatch('fetchPolicies')
+  await router.push({ name: 'data-plane-list-view', params: { mesh: 'default' } })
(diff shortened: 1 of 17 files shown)


slonka:
please resolve conflicts :)

kleinfreund:
> please resolve conflicts :)

@slonka Done.

src/app/data-planes/views/DataPlaneListView.spec.ts
```diff
@@ -72,4 +81,23 @@ describe('DataPlaneListView.vue', () => {
...
+    const wrapper = await renderComponent()
+
+    const secondTableRow = wrapper.findAll('[data-testid="data-overview-table"] tbody tr')[1]
+    expect(secondTableRow.html()).toContain(dppName)
+
+    expect(secondTableRow.element.classList.contains('is-selected')).toBe(false)
+    expect(wrapper.find('[data-testid="data-plane-proxy-title"]').html()).toContain('backend')
+    expect(window.location.search.includes('name=backend'))
+
+    // The click event listener for selecting a table row is unfortunately added to table cells instead of rows so we need to trigger a click on a cell instead of the row.
```
slonka, line 96: There is no functional error here right? This is just perf / proper implem issue right? Can we create an issue for this?

kleinfreund: Indeed, it’s not a functional error. It’s not an error of any form really. The fact that the event is fired from table cells just makes it a bit annoying to replicate it in tests. I don’t think a change in the implementation is necessary.

PR #445, opened 2022-11-08
fix: incorrect router base path

kleinfreund:
Fixes the router base path being incorrect when serving the GUI application from a non-root path.

### DIFF
--- a/src/router/router.ts
+++ b/src/router/router.ts
@@ -3,6 +3,7 @@ import { createRouter as createVueRouter, createWebHistory, NavigationGuard, Rou
 import { store } from '@/store/store'
 import { PolicyDefinition } from '@/types/index.d'
 import { ClientStorage } from '@/utilities/ClientStorage'
+import { getAppGuiPath } from '@/utilities/getAppGuiPath'
 
 function getPolicyRoutes(policies: PolicyDefinition[]): RouteRecordRaw[] {
   return policies.map((policy) => ({
@@ -281,8 +282,10 @@ export function createRouter(policyDefinitions: PolicyDefinition[] = []): Router
     },
   ]
 
+  const appGuiPath = getAppGuiPath(window.location)
+
   const router = createVueRouter({
-    history: createWebHistory(import.meta.env.BASE_URL),
+    history: createWebHistory(appGuiPath),
     routes,
   })
 
--- a/src/utilities/getAppBaseUrl.spec.ts
+++ b/src/utilities/getAppBaseUrl.spec.ts
@@ -1,0 +1,26 @@
+import { getAppBaseUrl } from './getAppBaseUrl'
+
+describe('getAppBaseUrl', () => {
+  test.each([
+    ['https://example.com/gui', 'https://example.com'],
+    ['https://example.com/gui/', 'https://example.com'],
+    ['https://gui.example.com/gui/', 'https://gui.example.com'],
+    ['https://example.com/dev/gui', 'https://example.com/dev'],
+    ['https://example.com/dev/gui/', 'https://example.com/dev'],
+    ['https://gui.example.com/dev/gui/', 'https://gui.example.com/dev'],
+    ['https://example.com/guillermo/dev/gui', 'https://example.com/guillermo/dev'],
+    ['https://example.com/guillermo/dev/gui/', 'https://example.com/guillermo/dev'],
+    ['https://gui.example.com/guillermo/dev/gui/', 'https://gui.example.com/guillermo/dev'],
+    ['https://example.com/gui/dev/gui', 'https://example.com'],
+    ['https://example.com/gui/dev/gui/', 'https://example.com'],
+    ['https://gui.example.com/gui/dev/gui/', 'https://gui.example.com'],
+    ['https://example.com', 'https://example.com'],
+    ['https://example.com/', 'https://example.com'],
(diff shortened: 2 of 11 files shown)


slonka (changes requested):
I think the `/gui` detection code should be be more bullet proof and probably we could use a couple of unit tests there.

src/router/router.ts
```diff
@@ -281,8 +281,16 @@ export function createRouter(policyDefinitions: PolicyDefinition[] = []): Router
     },
   ]
 
+  // Dynamically detects the URL path from which the GUI application is served as it is unknown to us at build time. The application can be served from an arbitrarily nested path so we can’t assume anything about the `/gui` path segment. Without this, router-created links in the application will be broken.
+  let base
+  const guiPathSegment = '/gui'
+  const guiPathSegmentIndex = window.location.pathname.indexOf(guiPathSegment)
```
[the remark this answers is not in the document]
kleinfreund: You’re completely right. I’ve extracted this functionality (and the one for the API base URL) into utility functions and wrote extensive unit tests for them in https://github.com/kumahq/kuma-gui/pull/445/commits/72bf3c1f93d0cc40b9a4fe6467e95cc1290c072e.

src/utilities/getAppBaseUrl.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { getAppGuiPath } from './getAppGuiPath'
```
slonka, line 1: shouldn't this be `getAppBaseUrl`?

src/utilities/getAppGuiPath.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { getAppGuiPath } from './getAppGuiPath'
+
+describe('getAppGuiPath', () => {
```
slonka, line 3: this looks like we are testing this twice

src/utilities/getAppBaseUrl.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  test.each([
+    ['https://example.com/gui', '/gui'],
+    ['https://example.com/gui/', '/gui'],
+    ['https://gui.example.com/gui/', '/gui'],
+    ['https://example.com/dev/gui', '/dev/gui'],
+    ['https://example.com/dev/gui/', '/dev/gui'],
+    ['https://gui.example.com/dev/gui/', '/dev/gui'],
+    ['https://example.com/guillermo/dev/gui', '/guillermo/dev/gui'],
+    ['https://example.com/guillermo/dev/gui/', '/guillermo/dev/gui'],
+    ['https://gui.example.com/guillermo/dev/gui/', '/guillermo/dev/gui'],
```
slonka, line 13: what's the correct behaviour for `/gui/dev/gui/` ? should we match the last one or the first one? Can we add a test case for that?

src/utilities/getAppBaseUrl.spec.ts
kleinfreund: Oh, I copied the wrong tests while splitting up the files 🤦🏻

kleinfreund: Fixed.

src/utilities/getAppGuiPath.spec.ts
kleinfreund: Fixed. See https://github.com/kumahq/kuma-gui/pull/445#discussion_r1017696184.

src/utilities/getAppBaseUrl.spec.ts
kleinfreund: I’d say that’s a pretty arbitrary choice. I’ve added tests for matching the first segment.

slonka: Isn't the last one safer option? I know it's pretty weird to have `/gui/something/gui/` but if we use the first one then the app would just show a 404 on the index load and all the links would "escape" that prefix right? If we use the last one then I guess we are good in that scenario right? Also how did this problem originate? (this might be a stupid question) Isn't just using relative links everywhere the solution to all of this?

kleinfreund: > I know it's pretty weird to have `/gui/something/gui/`

I don’t see why anything would ever link to `/gui/something/gui/`. That’s not something I think we need to protect against. If that should ever be a use case, I’d rather introduce a setting on the Kuma side that we discover via what’s proposed in https://github.com/kumahq/kuma/issues/3600#issuecomment-1308703684.

kleinfreund: > Also how did this problem originate? (this might be a stupid question) Isn't just using relative links everywhere the solution to all of this?

Relative URLs are a maintenance nightmare. If you’re at http://localhost:1234/a/b/c and you want to link to http://localhost:1234/d/e/f, the relative link target would need to be `../../../d/e/f`. It’s much more straightforward to have absolute or root-relative URLs that are the same no matter what page you’re on. For either, you need to know from where an application is served from. At the very least that means having a `<base href="http://localhost:1234/example/gui" />` (this makes all relative URL reference _relative_ to this base value) or an otherwise discoverable GUI base path.

kleinfreund: I think this is sufficiently tested now so I’m resolving this.

slonka (approved):

PR #449, opened 2022-11-11
refactor: removes testing library

kleinfreund:
Removes testing library and rewrites all tests using it using vue test utils.


johncowen (approved):
So this is just the switch of testing libraries and the changes related to that right? If so looks good to me!

One question I had was, all the `isloading="true"`s were changed to `isloading="false"`s in the App.spec file for the donut charts, wondered why that was needed?

kleinfreund:
> So this is just the switch of testing libraries and the changes related to that right? If so looks good to me!

Yep, it’s just removing testing library and changes affected tests.

> One question I had was, all the `isloading="true"`s were changed to `isloading="false"`s in the App.spec file for the donut charts, wondered why that was needed?

This change is caused by the point in time of asserting the snapshot being at a different point in time of the component’s life cycle. A call to `flushPromises` was added to get the timing correct and that effectively delays said assertion to a slightly _later_ point at which the loading prop is already switched to false. This is actually more accurate as that test specifically wants to wait for loading to be done.

johncowen:
Gotcha, yeah makes sense, thanks!

PR #456, opened 2022-11-16
refactor: use @jest/globals

kleinfreund:
Uses the package @jest/globals to explicitly import global Jest symbols and removes the global type definitions for jest.

### DIFF
--- a/jest/jest-replace-attribute-snapshot-serializer.ts
+++ b/jest/jest-replace-attribute-snapshot-serializer.ts
@@ -1,3 +1,6 @@
+// Imports the appropriate plugin type from the library Jest uses for snapshot serialization.
+import { NewPlugin } from 'pretty-format'
+
 const DEFAULT_UUID = 'aaaabbbb-cccc-dddd-eeee-ffffffffffff'
 
 function hasUuidAttribute(element: Element, attribute: string): boolean {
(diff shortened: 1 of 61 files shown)


jest/jest-replace-attribute-snapshot-serializer.js
```diff
@@ -1,17 +1,14 @@
 const DEFAULT_UUID = 'aaaabbbb-cccc-dddd-eeee-ffffffffffff'
 
-function hasUuidAttribute(element: Element, attribute: string): boolean {
+function hasUuidAttribute(element, attribute) {
```
kleinfreund, line 3: **Comment**: I turned this file into a plain JS file because it doesn’t look like `@jest/globals` exports the `SnapshotSerializerPlugin` type.

johncowen: Wondered if you could you import from `pretty-format` instead?

https://github.com/facebook/jest/blob/main/packages/pretty-format/src/types.ts#L69-L98

kleinfreund: Good idea. That’s as good as it’s gonna get. Done.

johncowen (approved):
👍

Issue #451, opened 2022-11-11
The wizard is providing the wrong Kubernetes commands


subnetmarco:
### What happened?

The wizard is showing `kumactl` examples even when explicitly looking for `Kubernetes` examples. This could be happening for every other wizard command.

I am following the instructions to create a virtual mesh:

<img width="830" alt="Screen Shot 2022-11-11 at 9 18 37 AM" src="https://user-images.githubusercontent.com/964813/201359031-993b8732-32a6-4ba4-82b8-cc6fbcdab951.png">

But the suggested command doesn't work because it's using `kumactl` instead of `kubectl`:

```
$ echo "type: Mesh
> name: asd" | kumactl apply -f -
Error: On Kubernetes you cannot change the state of Kuma resources with 'kumactl apply' or via the HTTP API. As a best practice, you should always be using 'kubectl apply' instead. You can still use 'kumactl' or the HTTP API to make read-only operations. On Universal this limitation does not apply.
```

PR #457, opened 2022-11-16
fix: Show correct 'ctl command in mesh wizard code block

johncowen:
Makes the command dynamic depending on which tab you have selected, plus
adds some data-test-* attributes for testing purposes

Fixes https://github.com/kumahq/kuma-gui/issues/451

### DIFF
--- a/src/app/wizard/views/Mesh.vue
+++ b/src/app/wizard/views/Mesh.vue
@@ -499,7 +499,7 @@
                 <template #kubernetes>
                   <CodeBlock
                     id="code-block-kubernetes-command"
-                    data-test-codeblock="kubernetes"
+                    data-testid="kubernetes"
                     language="bash"
                     :code="codeOutput"
                   />
@@ -507,7 +507,7 @@
                 <template #universal>
                   <CodeBlock
                     id="code-block-universal-command"
-                    data-test-codeblock="universal"
+                    data-testid="universal"
                     language="bash"
                     :code="codeOutput"
                   />
--- a/src/app/wizard/views/Mesh.vue
+++ b/src/app/wizard/views/Mesh.vue
@@ -499,6 +499,7 @@
                 <template #kubernetes>
                   <CodeBlock
                     id="code-block-kubernetes-command"
+                    data-test-codeblock="kubernetes"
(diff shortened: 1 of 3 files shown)


src/app/wizard/views/Mesh.vue
```diff
@@ -499,13 +499,15 @@
                 <template #kubernetes>
                   <CodeBlock
                     id="code-block-kubernetes-command"
+                    data-test-codeblock="kubernetes"
```
kleinfreund, line 502: **Suggestion**: Can we stick to the commonly present `data-testid` style format here? I know it doesn’t matter, but I rather have _one_ attribute name with _different_ values_ than both name and value being different.

kleinfreund (approved):

src/app/wizard/views/Mesh.vue
johncowen: Using a single one means they can easily get overridden:

Screenshot 2022-11-16 at 14 17 37

Not sure if overwriting that existing `data-testid` does anything right now, but might be worth chatting through how to avoid this, say deciding on a pattern using `data-test-component` for top level component node? Do you know if Kongponents uses `data-test-*` attributes?

Let me know if you still want to tweak anything here (or change to `data-testid` anyway), otherwise I'll merge a little later.

kleinfreund: Why do you want to avoid overwriting them? In my experience, you rarely want to rely on attributes you don’t control. It’s also pretty rare that you even need to find and use components from external dependencies.

I don’t think any more sophistication is needed here. `data-testid="kubernetes-code-block"` does the job just as well. Happy to reconsider this is we do have such conflicts.

(Kongponents uses `data-testid="*"` attributes exclusively)

johncowen: > Why do you want to avoid overwriting them?

I suppose because I've no idea if something was relying on that elsewhere (in this instance I guess it's not as no tests broke).

> (Kongponents uses data-testid="*" attributes exclusively)

Ok good to know

Will switch up to use `data-testid`, gimme 2 tics

PR #461, opened 2022-11-17
feat(dpp): adds rules API data to policies tab

### LINKED ISSUE #387: Update inspect view to support rule based view and others
### Description

Atm the moment the inspect view could be a little nicer (data is not very readable). Also as part of: https://github.com/kumahq/kuma/issues/5036 there's a whole new set of data to display

kleinfreund:
## Changes

Adds rules from the inspect API to the policies tab.

Reworks layout for policies tab content to be grouped by policy type.

Resolves #387.

## Screenshots

<details><summary>Before:</summary>

Screenshot from 2022-11-22 12-12-22

</details>

<details><summary>After:</summary>

image

</details>


src/app/common/PolicyTypeTag.vue
```diff
@@ -0,0 +1,114 @@
...
+      class="policy-type-tag-icon"
+      :src="policy.iconUrl"
+      alt=""
+    >
+
+    <KIcon
+      v-else
+      icon="brain"
+      size="24"
+    />
```
lahabana, line 14: IIUC the "brain" icon is shown whenever there's no dedicated icon for this policy?

src/app/data-planes/components/PolicyTypeEntryList.vue
```diff
@@ -0,0 +1,151 @@
...
+            <template #config="{ rowValue, rowKey }">
+              <template v-if="rowValue !== null">
+                <details>
+                  <summary>Conf</summary>
+
+                  <CodeBlock
+                    :id="`${index}-${rowKey}-code-block`"
+                    :code="rowValue"
+                    language="yaml"
+                  />
```
lahabana, line 76: Can we maybe get rid of the line number of this? I find the expanded conf a little `bulky` also because it has to fit in a column it's quickly wrapped when there's a lot of space .e.g:

image

Wondering if there's something a bit nicer to be done here

src/api/mock-data/dataplane-rules.json
```diff
@@ -0,0 +1,92 @@
+{
+  "total": 3,
+  "items": [
+    {
+      "type": "clientSubset",
+      "name": "192.168.0.1:80",
+      "service": "backend",
```
lahabana, line 7: @jakubdyszkiewicz is this actually possible? Do clientSubset actually have service?

lahabana (commented):
So questions and nits but this is already looking very cool and such an improvement over the old UI.

I think we should be able to improve a bit the from entry it wasn't clear to me that `backend`  and `cluster-1.backend-02` were services in this view:

image

I'll try to think some more what could be done but if you have ideas don't hesitate :)

src/app/common/PolicyTypeTag.vue
kleinfreund: Yes, we’re missing 5 icons in total (the new Mesh* policies, basically). I don’t know who creates them for kuma.io, but we probably can get some new ones, right?

lahabana: Yeah we could. I think for the new policies we can reuse the old ones (as they are replacements really). But in any case we should keep the old stuff just because there's usually a delay between icons and actual implem (and we're also disucssing having this fully pluggable for users)

kleinfreund: I’ve updated the following to use existing icons of their predecessors:

- MeshAccessLog → TrafficLog
- MeshTrace → TrafficTrace
- MeshTrafficPermission → TrafficPermission

kleinfreund:
> I think we should be able to improve a bit the from entry it wasn't clear to me that `backend` and `cluster-1.backend-02` were services in this view:

@lahabana Does something like the following make sense?

| From          | To            |
| ------------- | ------------- |
| web (service) | backend (DPP) |

src/api/mock-data/dataplane-rules.json
jakubdyszkiewicz: It's a service associated with an inbound 192.168.0.1:80 so we don't have to look for it in Dataplane object.

kleinfreund: In other words: this is a convenience field to save a lookup

lahabana: Ha so the `kuma.io/service` line 10 should be `backend` no?

src/app/data-planes/components/PolicyTypeEntryList.vue
lahabana: Yes can't find of a good option here... Let's see if someone else comes up with something :)

kleinfreund: I did some small adjustments here: the code block no longer shows the copy button here and I’ve also removed the disclosure widget around it.

slonka:
I like the fact that it still looks nice even with long names

image

slonka:
Something that's probably not related is that refreshing a page causes a 404 and you have to go back to the root of the page because none of the links "in the middle" work as well.

https://user-images.githubusercontent.com/2753650/204740190-a11eee45-4c84-4121-9c95-c4d71ff6b512.mov

slonka:
The IPs have multiple ports, it's probably a bug in the mocked response but just making sure

image

kleinfreund:
> The IPs have multiple ports, it's probably a bug in the mocked response but just making sure

@slonka This mock data for `/policies` is taken from Kuma (e.g. `pkg/api-server/testdata/inspect_dataplane.json`) so I assume that has some significance. If not, we can change this easily as it’s not in any way constructed.

slonka:
What's the difference between `-` vs `kuma.io/service: *`? When I saw this:

image

I thought it meant "all" but then I saw this:

image

so I thought why are there two ways of presenting "all". ?

kleinfreund:
> What's the difference between `-` vs `kuma.io/service: *`?

@slonka There are technically three scenarios that are relevant here:

* Showing an EM dash (i.e. `—`): This is synonymous with “this current data plane proxy”
* Showing the wildcard service tag from a `*/policies` source: matched policies from `*/policies` can have the wildcard service tag as a source. So when you see it in the upper section (i.e. above "Rules"), that’s just the value from the API response
* Showing the wildcard service tag **from empty `clientSubset` rule subsets**: for client subsets from `*/rules`, if `subset` is empty, we *assume* this to mean the wildcard service tag and add it on the UI’s side.

This (specifically the third point) might be confusing either if we do it or if we don’t (summoning @jakubdyszkiewicz and @lahabana who both had light opinions about this) so I’d greatly appreciate some input on whether this difference of "this DPP" versus "all services" is significant. In the case of the second point, I think this should definitely show the wildcard service tag because that’s what’s coming from the API.

slonka:
Sometimes I'm not really sure if a "thing" I'm noticing is just the way mocked data is set-up or there might be something in the GUI code. Maybe we could think about re-doing the mocked data to better reflect a real scenario.

slonka:
> Showing an EM dash (i.e. —): This is synonymous with “this current data plane proxy”

I'd have an explicit "This DPP" instead of —.

> difference of "this DPP" versus "all services" is significant

Just to make sure I understand "this DPP" - it means that it was selected by some selector but not via "all services" wildcard ("*") - right?

kleinfreund:
> I'd have an explicit "This DPP" instead of —.

@slonka Happy to run a poll for this. I initially had exactly that, but through feedback and iteration, dropped it again.

> Just to make sure I understand "this DPP" - it means that it was selected by some selector but not via "all services" wildcard ("*") - right?

@slonka Not sure I get you right, but really "this DPP" just means "the DPP you’re looking at right now for which you went to the Policies tab". I don’t know if the wildcard service tag implies semantically the same source or if there is a distinction.

slonka:
Also, have we thought about explicitly stating that the top ones are "source & destination" policies and the bottom ones instead of "Rules" - "targetRef policies"? WDYT @lahabana ?

kleinfreund:
> Also, have we thought about explicitly stating that the top ones are "source & destination" policies and the bottom ones instead of "Rules" - "targetRef policies"? WDYT @lahabana ?

@slonka I thought about that, but I couldn’t come up with a handy name that makes sense to users. Happy to hear suggestions because I would also prefer to have a heading for both. In the long run (but perhaps not now), we would not split these structures up I think so then it would be fine to not have a heading at all (the tab would be the quasi heading).

lobkovilya:
I don't think `N/A` is correct in the MeshAccessLog example:
image

It means "not applicable" and because of that it feels like MeshAccessLog can't have a "from" section in general

kleinfreund:
> It means "not applicable" and because of that it feels like MeshAccessLog can't have a "from" section in general

I think there are some subtle differences here overall and so I’d probably prefer to go with the non-significant em dash ("—").

src/app/data-planes/components/DataplanePolicies.vue
```diff
@@ -161,44 +188,128 @@ function getPolicyRoutes(policies: Record<string, PolicyType> | undefined): Mesh
...
+    const connection: PolicyTypeEntryConnection = { sourceTags, destinationTags, name, config, origins }
+    policyEntry.connections.push(connection)
   }
 
-  return dataplanePolicies
+  const policyTypeEntries = Array.from(policyEntriesByType.values())
+
+  policyTypeEntries.sort((policyEntryA, policyEntryB) => policyEntryA.type.localeCompare(policyEntryB.type))
+
+  return policyTypeEntries
```
johncowen, line 312: I found all of this a bit 'intense' to look at, I can see that its mostly reshaping data for easier lookup (I assume). I had two things I was wondering, is it worth organising it a little more to make it unit testable? Or another idea was adding some more comments in here just giving a brief explanation of what's happening and why. Both of which would probably help when wondering what shape we are going from > to. Appreciate you are working with the code thats already there so maybe no is not the right time? Thought I'd ask in case.

I'm also getting some editor errors down on line 304 assuming because sourceTags and destinationTags can be `undefined` rather than `null`, are you getting those? Not a biggie

src/app/data-planes/components/PolicyTypeEntryList.vue
```diff
@@ -0,0 +1,161 @@
...
+}
+
+.origin-list {
+  display: flex;
+  flex-direction: column;
+}
+</style>
+
+<style lang="scss">
+.policy-type-table.policy-type-table td {
```
johncowen, line 152: Is this maybe your editor again? I don't think you need `policy-type-table` twice here?

johncowen (commented):
I had a quick look through the code and had a few questions. Apart from that my first thought was, do we want to resize that imagery down to the size we are using? I'm guessing we can't convert them all to svgs, which is a shame - there's no way of easily removing the white background from that big jpg is there? Then at least they are all transparent pngs.

src/app/common/CodeBlock.vue
```diff
@@ -23,39 +24,33 @@ import { ClientStorage } from '@/utilities/ClientStorage'
...
-   * The syntax language of `props.code` (e.g. `'json'`).
-   */
   language: {
     type: String as PropType<AvailableLanguages>,
     required: true,
   },
 
-  /**
-   * Shows an actions bar with a search input and related action buttons.
-   */
```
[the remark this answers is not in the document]
kleinfreund: Because that’s just props passed along to KCodeBlock where all of this is documented again. No point in having this here.

src/app/common/CodeBlock.vue
```diff
@@ -78,7 +73,6 @@ const reformattedCode = computed(() => props.language === 'yaml' ? reformatYaml(
 async function handleCodeBlockRenderEvent({ preElement, codeElement, language, code }: CodeBlockEventData): Promise<void> {
   isProcessing.value = true
 
-  await nextTick()
```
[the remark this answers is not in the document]
kleinfreund: It is no longer required because I addressed the need for it in KCodeBlock where it's adequate. Consuming components shouldn’t care for this sort of timing detail: the correct way to do this was to delay the event until the rendering actually happened.

src/app/data-planes/components/PolicyTypeEntryList.vue
```diff
@@ -0,0 +1,161 @@
...
+    <AccordionItem
+      v-for="(policyEntry, index) in props.policyTypeEntries"
+      :key="index"
+    >
+      <template #accordion-header>
+        <h3>
+          <PolicyTypeTag :policy-type="policyEntry.type">
+            {{ policyEntry.type }} ({{ policyEntry.connections.length }})
+          </PolicyTypeTag>
+        </h3>
```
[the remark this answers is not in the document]
kleinfreund: Not hard, but it’s actually the DPP name that’s in the wrong place. We don’t have an `h2` heading in the document at this point at all. In fact we don’t even have an `h1`. Our views in general have pretty inconsistent document outlines.

I’m gonna change the DPP heading (and the same ones in other views) to h2 for now. We should consider at one point what each view should have as a document title (i.e. `h1`) and adjust the headings accordingly.

src/app/data-planes/components/PolicyTypeEntryList.vue
```diff
@@ -0,0 +1,161 @@
...
+
+              <template v-else>
+                —
+              </template>
+            </template>
+
+            <template #config="{ rowValue, rowKey }">
+              <template v-if="rowValue !== null">
+                <CodeBlock
+                  :id="`${index}-${rowKey}-code-block`"
```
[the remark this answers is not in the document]
kleinfreund: Because the component contains elements that have `for`/`id` attribute relationships that must be unique within the document. While adding the component to Kongponents, we decided to make the `id` mandatory so one can’t forget to explicitly take care of it.

src/app/data-planes/components/PolicyTypeEntryList.vue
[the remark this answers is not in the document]
kleinfreund: I’ve changed the document outline to look like this now:

* DPP: <dpp name>
  * Policies
    * <policy type entry headings>
  * Rules
    * <policy type entry headings>

src/app/data-planes/components/SidecarDataplanePolicyList.vue
```diff
@@ -1,157 +1,46 @@
...
-            :key="`${index}-${matchPolicyIndex}`"
-          >
-            <p class="policy-type">
-              {{ matchedPolicy.pluralName }}
-            </p>
+  <div
+    v-if="policyTypeEntriesFromRules.length > 0"
+    class="mt-2"
+  >
+    <h2>Rules</h2>
```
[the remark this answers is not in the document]
kleinfreund: See conversation in https://github.com/kumahq/kuma-gui/pull/461#discussion_r1039413688.

vite.config.js
```diff
@@ -24,7 +24,13 @@ export default defineConfig(function ({ mode }) {
...
        */
       pluginRewriteAll(),
-      vue(),
+      vue({
+        template: {
+          compilerOptions: {
+            whitespace: 'preserve',
+          },
+        },
+      }),
```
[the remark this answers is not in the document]
kleinfreund: No. It’s for any view where you render a template like this:

```html
<p>
  one
  two
  three
</p>
```

With the default setting of Vue CLI/Vite (`'condense'`), the resulting HTML looks like this:

```html
<p>onetwothree</p>
```

Which is completely in contrast with how HTML normally renders which would be (unsurprisingly):

```html
<p>
  one
  two
  three
</p>
```

This is quite important to me because the browser displays both things differently. Since the spaces between the words are removed in one example, the browser would actually show “onetwothree” while in the other example it would be “one two three”.

I don’t know why the Vue folks decided on this non-standard whitespace processing default for templates, but I strongly dislike it.

src/app/common/CodeBlock.vue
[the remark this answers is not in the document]
johncowen: Ah gotcha, thankyou!

src/app/common/CodeBlock.vue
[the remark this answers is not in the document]
johncowen: Ah thanks, maybe thats where I saw it last, oh wow ok I just noticed the `isProcessing` thing also, interesting

src/app/data-planes/components/PolicyTypeEntryList.vue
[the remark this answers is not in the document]
johncowen: Oh I see, for the code searching form input? In this case that means there can only ever be one instance of this particular `PolicyTypeEntryList` component in the page at once? 

Do you know if Vue has its own recommended native uuid generator, I had a very brief look but didn't see one see unless I missed it? Sorry getting a bit off topic, but just getting a feel for things still.

vite.config.js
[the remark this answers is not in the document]
johncowen: Oh wow yeah thats unfortunate 😄 . Could you point me to where that was an issue in this PR, surprised we've got this far without it being an issue, oh mind you I suppose its since we've moved to vite maybe?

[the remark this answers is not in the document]
kleinfreund: Since moving to tables, this didn’t happen in this PR, but I’ve worked around this in other instances by turning things into horizontal lists. Don’t remember where though.

src/app/data-planes/components/DataplanePolicies.vue
johncowen: > I process the data in such a way that it fits the use case of respective templates as close as possible to avoid doing bigger chunks of logic in templates.

Yeah for sure, makes sense 👍 

> This process is inadvertently complex

Yeah that's why I wondered if unit test might have helped me (no action here by the way if you are happy with how it is)

> It would help to know which parts are particularly hard to follow.

It wasn't anything in particular, just a general "looks like theres a lot of stuff going on here", might be partly down to my unfamiliarity with the data also.

src/app/data-planes/components/PolicyTypeEntryList.vue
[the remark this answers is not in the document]
kleinfreund: > Oh I see, for the code searching form input?

Yes

> In this case that means there can only ever be one instance of this particular PolicyTypeEntryList component in the page at once?

I’m not sure what you mean by this, but I did notice an issue with the IDs here. Since this component is used twice on this page, I actually need to make these IDs more unique. Fixed now.

> Do you know if Vue has its own recommended native uuid generator, I had a very brief look but didn't see one see unless I missed it? Sorry getting a bit off topic, but just getting a feel for things still.

Don’t think so, but there are utilities for doing this (Kongponents uses one).

src/app/data-planes/components/DataplanePolicies.vue
kleinfreund: I agree there is a lot going on with this. I just can’t think of something clever to comment that doesn’t amount to "well basically that one child component needs it to look like that"

src/app/data-planes/components/PolicyTypeEntryList.vue
johncowen: > What do you mean by this?

There's was another thing I picked up on that was just something you said your editor did (the adding `js` to the imports) which I wasn't aware of, I thought this could have been similar.

> Yes, you do. Without increasing the specificity of this rule, the KTable default overrides it to vertical-align: middle which I don’t want for this table. The cells should be top-aligned.

Oh I see 😄 , not necessary to change unless you like it better, but I wondered if something like `.policy-type-table .k-table td` might have read more like "I'm doing this specifically to overwrite a rule that is only troublesome in k-tables" - not sure if that has exactly the same effect - just the double class looked weird/jumped out to me.

kleinfreund: > I wondered if something like `.policy-type-table .k-table td` might have read more like "I'm doing this specifically to overwrite a rule that is only troublesome in k-tables" - not sure if that has exactly the same effect - just the double class looked weird/jumped out to me.

The effect would be the same (adding one class selector to the selector list); however, I _specifically_ duplicate an _existing_ selector if possible because in my opinion, that does the job better and with less ambiguity:

- `.policy-type-table .k-table td`: more ambiguous. It doesn’t necessarily implicate that this is a specificity hack but rather that the selector needs to be scoped to `td`s within `.k-table`s within `.policy-type-table`s. One could think "ah, but `.k-table` here isn’t needed so I can remove it". That would be wrong and it would be hard to spot.
- `.k-table.k-table td`: too general. As written, this looks to be generally applicable to KTable uses which isn’t intentional here. It so happens that this only has KTables we want to change, but that’s just circumstantial.
- `.policy-type-table.policy-type-table td`: Indicates that this is a specificity hack which as far as I know is the only reason one would want to duplicate a class selector like this. Scopes the rule to only the tables we want to scope it to.

src/app/data-planes/components/PolicyTypeEntryList.vue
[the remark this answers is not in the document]
johncowen: > In this case that means there can only ever be one instance of this particular PolicyTypeEntryList component in the page at once?

>> I’m not sure what you mean by this, but I did notice an issue with the IDs here. Since this component is used twice on this page, I actually need to make these IDs more unique. Fixed now.

Ah! I didn't even notice this was on this page twice 😄 , but yeah exactly that is what I meant.

> (Kongponents uses one).

Do you know whereabouts that is? I had a quick look but didn't see it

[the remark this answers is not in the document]
kleinfreund: Any component that imports the following (e.g. KTable, 10 in total):

```js
import { v1 as uuidv1 } from 'uuid'
```

johncowen (approved):
I think it was mainly the big 42kb white background jpg here:

https://github.com/kumahq/kuma-gui/blob/2072de0c7ff98bf127fa6dd9eab14398a22b4899/src/assets/images/policies/Timeout.jpg

I was most concerned about, but I think this needs to go in so from my side ✅  , not sure if you need an approval from the other folks on here also or not

src/app/data-planes/components/PolicyTypeEntryList.vue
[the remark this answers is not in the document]
johncowen: Oh I see, hmm not quite what I was thinking but thanks, good to know

PR #465, opened 2022-11-18
revert: switching to history mode

kleinfreund:
Reverts the relevant changes that switched the vue router to history mode.

### DIFF
--- a/src/app/__snapshots__/App.spec.ts.snap
+++ b/src/app/__snapshots__/App.spec.ts.snap
@@ -12,7 +12,7 @@ exports[`App.vue fails to renders basic view 1`] = `
       <a
         aria-current="page"
         class="router-link-active router-link-exact-active"
-        href="/"
+        href="#/"
       >
         <img
           alt="Kuma Logo"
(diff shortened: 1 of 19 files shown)


johncowen (commented):
Not a biggie, but I noticed that bookmarking/linking to a specific "Data Plane Proxy Preview" doesn't look like it's working now with this revert? For example linking here `http://localhost:8080/?name=cluster-1.backend-02#/mesh/default/data-planes` redirects me back to `?name=backend`. From what I remember the query parameter addition wasn't there when we were using hash fragments, should we revert adding those to the URL also if they aren't working with this revert?

Other than that all looks good to me!

johncowen (approved):
Cool nice one, that works better for me!

lahabana:
So this is not needed anymore: https://github.com/kumahq/kuma/issues/5306 ?

kleinfreund:
@lahabana I’d still love to get https://github.com/kumahq/kuma/issues/5306 in place. Working with hash-based URLs is not great and they prevent us from ever using anchors anywhere else in the application. It’s easy enough (looking at the minimal changes) to redo this later so for now reverting this is more important than waiting for a possible fix.

lahabana:
I guess this can now be canned right?

Issue #447, opened 2022-11-09
Add a button to file an issue


lahabana:
### Description

where to put it is an open question

lahabana:
Also maybe link to top level docs

lahabana:
Speak with @johnharris85 about details and scope

PR #466, opened 2022-11-21
feat: Add help menu to the header navigation

johncowen:
Add a new Help menu to the top right of the GUI containing a link to the docs and a link to file an issue on Kuma.

Wasn't totally sure whether the create issue link should go to the Kuma or KumaGUI repo

Notes:

I made a `useEnv` composable to centralise our environment variables and some of our static strings that aren't yet environment variables but potentially should be (the root url for the docs site for example). One important thing to note here is that it looks like the version is always set on application boot (unless something is more drastically wrong) i.e. before we display anything.

This `useEnv` also hides references to our build tooling (vite) so we aren't using that naming etc within the rest of the application code.

---

There are probably a multitude of ways to do this, but this was my preferred and `useEnv` could potentially be used for replacing these:

https://github.com/search?q=repo%3Akumahq%2Fkuma-gui+kuma.io%2Fdocs+language%3AVue&type=code&l=Vue

.. so we can centrally configure our URLs per environment.

Fixes https://github.com/kumahq/kuma-gui/issues/447


src/composables/useEnv.ts
```diff
@@ -0,0 +1,20 @@
+type EnvVars = {
+  KUMA_VERSION: string
+  KUMA_DOCS_URL: string
+  KUMA_UTM_QUERY_PARAMS: string
+  KUMA_FEEDBACK_URL: string
+}
+export const useEnv = ({ version = '' }) => {
+  const env: EnvVars = {
+    KUMA_VERSION: version,
```
johncowen, line 9: I need to change the name of this as I think this isn't the version but the 'docs version' i.e. the minor version `0.0.x`. Need to double check and come back and change if so.

src/app/AppHeader.vue
```diff
@@ -46,29 +46,71 @@
...
-          icon="gearFilled"
-          color="currentColor"
-          title="Diagnostics"
-        />
-
+      <KDropdownMenu
+        v-if="env('KUMA_VERSION') !== ''"
+        class="help-menu"
+        icon="help"
+        button-appearance="outline"
```
kleinfreund, line 53: **Suggestion**: Positions the dialog at the bottom right corner.

```suggestion
        button-appearance="outline"
        :kpop-attributes="{ placement: 'bottomEnd' }"
```

src/composables/useEnv.ts
johncowen: Yeah I getcha, but this `version` comes from `getKumaDocsVersion`

https://github.com/kumahq/kuma-gui/pull/466/files#diff-690147818f34213a982d8623466a4657d508ec2f07081366e3adb0725ceda5ceR114

So `KUMA_VERSION` the 'minor-ified' version `1.7.x` not the full `1.7.0` version.

> I need to change the name of this as I think this isn't the version but the 'docs version' i.e. the minor version 0.0.x. Need to double check and come back and change if so.

Looking at it a little better and knowing whats happening in other PRs it probably makes sense to keep the naming and then 'minor-ify' the version in one place for the docs URL here:

https://github.com/kumahq/kuma-gui/pull/466/files#diff-83628cc2270f76402f0d38798b33cec37e0f44bfe660fa723b9998ef87080c27R10

I might keep this PR hanging until that other work is in.

.env
```diff
@@ -5,5 +5,7 @@ VITE_NAMESPACE=Kuma
 VITE_KUMA_API_SERVER_URL=http://localhost:5681/dev/
 VITE_KUMA_DP_SERVER_URL=https://localhost:5678/
 VITE_AMCHARTS_LICENSE=''
+VITE_DOCS_ROOT=https://kuma.io/docs/
```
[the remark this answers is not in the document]
johncowen: Good point! I switched this up

src/composables/useEnv.ts
```diff
@@ -0,0 +1,20 @@
+type EnvVars = {
+  KUMA_VERSION: string
+  KUMA_DOCS_URL: string
+  KUMA_UTM_QUERY_PARAMS: string
+  KUMA_FEEDBACK_URL: string
+}
+export const useEnv = ({ version = '' }) => {
```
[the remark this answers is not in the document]
johncowen: I think I originally passed it in from where the value already was to avoid changing things too much. Probably won't need to pass a version from the `store` in any more once we can get that value from elsewhere, so this will change slightly once other code comes up here.

mmorel-35:
Hi ! Concerning this : 

> Wasn't totally sure whether the create issue link should go to the Kuma or KumaGUI repo

Maybe you can provide a link for both of them with an explicit title saying something like "Feedback > Kuma Manager" or "FeedBack > Kuma" ?

johncowen:
I've a few tweaks here following upstream changes:

1. version now comes from the gui config that is passed to the GUI via the index.html file.
2. semver-like parsing is added here, super basic parsing just enough to more or less follow what it was doing previously (see https://github.com/kumahq/kuma-gui/blob/7f1d17145684e5c09578992753d7332725e16363/src/store/modules/config/config.ts#L68-L92)
3. Moved the files to `@/utilities`

Remaining work after this PR

1. Replace other instances where we use these variables with `useEnv`/`env`
2. Remove the old docs string parse thing that lives in the vuex store

I wanted to ask about `basedOnKuma` - the code here doesn't use that as yet, so it's theoretically removing functionality that we need to keep. You mentioned we use this in Kong Mesh? I understand by that that the only time there is a difference between `version` and `basedOnKuma` is if we are running from something that isn't kuma specifically but uses kuma (like Kong Mesh)? I'm just wondering exactly what to do here, as its obviously missing right now, and it should be added into the gui config we get from the index.html (along with `tagline`). If it's just something thats only necessary in Kong Mesh that would be good to know.

kleinfreund:
> I wanted to ask about `basedOnKuma` - the code here doesn't use that as yet, so it's theoretically removing functionality that we need to keep. You mentioned we use this in Kong Mesh?

Yes. In Kong Mesh, `basedOnKuma` is the version of the underlying Kuma version which is needed to construct appropriate docs URLs.

> I'm just wondering exactly what to do here, as its obviously missing right now, and it should be added into the gui config we get from the index.html (along with `tagline`). If it's just something thats only necessary in Kong Mesh that would be good to know.

Well, we should do ... what we do right now, no? That logic was specifically added to account for this difference.

It’s not complex logic and only in a very, very limited way the same as "parsing a version string into its semantic versioning components". We just replace the last number with an `x` and set "dev" for preview versions.

Also, take note of the semver package that we have already installed in the project.

I’ll also re-iterate what I said before: I don’t think there should be _any_ logic whatsoever in a utility for retrieving environment variables. Anything that’s a computed result of some stuff, among which parts are environment variables, is application logic and not an environment variable anymore.

johncowen:
> Also, take note of the semver package that we have already installed in the project.

Oh perfect 😆 , I was just thinking I don't want to include a big package just for this, but if we are already using it then 🤷 , should have looked!

johncowen:
Hmm 🤔 turns out we aren't using the `semver` package on the frontend, only for build tooling. Thanks for the nudge but I think I'm happier sticking with what I have here at least for the moment (unless I'm mistaken that we are actually using it?)

Lemme know, it could do with being removed from package.json if we are ok to just stick with the split/joining until we come to need something more involved.

johncowen:
> I’ll also re-iterate what I said before: I don’t think there should be any logic whatsoever in a utility for retrieving environment variables. Anything that’s a computed result of some stuff, among which parts are environment variables, is application logic and not an environment variable anymore.

It's a centralized place to access variables that don't really ever change without having to refresh the page i.e. they are not reactive. These variables might come from build tooling environment variables, runtime backend 'environment variables' passed via our index.html file, or even in the future potentially user/engineer runtime 'environment variables' either via local storage or cookies.

`useEnv` and it's resulting `env` hides all those details away so we don't have to worry about them. We just pretend they are all 'env vars'. The fact that we figure some out from another for ease/centralizing doesn't really bother me, and avoids repeating the same logic in multiple places throughout the app, something I'm seeing a lot of.

johncowen:
I've reverted this back to what I had before the upstream work so we can go with that for the moment. Using that upstream work can be a followup PR.

src/utilities/useEnv.ts
```diff
@@ -0,0 +1,21 @@
...
+type EnvVars = {
+  KUMA_DOCS_URL: string
+  KUMA_UTM_QUERY_PARAMS: string
+  KUMA_FEEDBACK_URL: string
+}
+export const useEnv = () => {
+  const store = useStore()
+  const docsVersion = store.getters['config/getKumaDocsVersion']
+  const env: EnvVars = {
+    KUMA_DOCS_URL: `${import.meta.env.VITE_DOCS_BASE_URL}/${docsVersion}/`,
```
kleinfreund, line 11: **Question**: Why is this variable not created and stored in the store? In my opinion, this is where all computation of some application-wide stored data should happen.

This utility (i.e. `useEnv`) is effectively a secondary store for static/computed data. However, the store we create using Vuex already serves that purpose. Why do we need this store on top of it?

Since this store accesses the Vuex store using a composable (`useStore` wraps the Vuex `useStore` composable), it cannot be used in a context where the Vuex store can’t be used.

On a side note, I also think the name is slightly misleading. `useEnv` implies you’re accessing environment variables and while that’s certainly part of what the utility does, it does more than it — stuff environment variables usually don’t. I’d rather not muddle the waters and refer to environment variables in a utility that does more than access environment variables.

src/utilities/index.ts
```diff
@@ -0,0 +1 @@
+export { useEnv } from './useEnv'
```
kleinfreund, line 1: **Question**: Why do we need a reexport file? I never understood the concept of these files. What problem does it solve that a direct/explicit import can’t/doesn’t address?

If you absolutely want to use them, I don’t mind that, but I think they just add maintenance work without improving anything.

src/utilities/useEnv.ts
johncowen: 1. We are moving away from Vuex.
2. We are moving away from these values being "user data". They are environment variables not API exposed user data. No reason to use all the previous code to manage runtime state when that state never changes.
3. Whether we do the computation as a build step or a runtime step, to be honest I don't really mind. It's just easier to do this here for now so we can move on and then begin to pull in all the other changes. This PR was meant to be "add a button" not "change how we pass in 'burnt in' environment variables". As it stands now, we achieve the "add a button" so we can merge this but also begin to prepare the way for the "change how we pass in burnt in environment variables"  which is happening upstream.
4. Naming-wise, I really don't want the name of our build tooling and other related things tightly coupled to our application.

kleinfreund: > We are moving away from Vuex.

That is news to me. When was this decided and by who and why?

> We are moving away from these values being "user data". They are environment variables not API exposed user data. No reason to use all the previous code to manage runtime state when that state never changes.

I’m not sure I understand this argument. All this data *is* runtime state and while it might never change that doesn’t mean it shouldn’t be stored in a Vuex store. Just because Vuex can recompute state in terms of Vue reactivity doesn’t mean this adds an unnecessary burden on storing static information in it.

State doesn’t immediately imply "stuff that’s constantly being recomputed". 

> Whether we do the computation as a build step or a runtime step, to be honest I don't really mind.

This computation doesn’t happen at build time in either approach.

> It's just easier to do this here for now so we can move on

It’s just as easy to do this in the Vuex store.

> and then begin to pull in all the other changes.

What do you mean by this?

> "change how we pass in burnt in environment variables" which is happening upstream.

I’m wholly unaware what you mean by this. What is "happening upstream"?

> Naming-wise, I really don't want the name of our build tooling and other related things tightly coupled to our application.

That’s an argument for introducing a utility that abstracts access to environment variable, but this utility does more than access environment variables. It assembles computed state. I think our respective definition of environment variables doesn’t match.

---

What I really like answered, however, is

> **Why** is this variable not created and stored in the store? In my opinion, this is where all computation of some application-wide stored data should happen.

What is the problem with the current approach of storing and accessing application state and what is it that you intend to change about it?

johncowen: Honestly, I just want to add a button 😆 🤷

kleinfreund (approved):
I’m approving this as the core intention of this PR was adding the help menu in the top bar which was done.

I do however want to discuss the effort of newly introduced storage facilities which isn’t very well motivated so far.

johncowen:
I pulled it all out again, you might want to give it another check before I merge. Otherwise I'll resolve the outstanding conversation and merge away!

kleinfreund:
> you might want to give it another check before I merge.

Looks all good to me.

johncowen:
oh hey @mmorel-35 👋 , sorry I wanted to come back to you on this:

> Maybe you can provide a link for both of them with an explicit title saying something like "Feedback > Kuma Manager" or "FeedBack > Kuma" ?

We did chat about this option and figured if we just had one place for people to click then we could just transfer the issue in GH during triage if it needed to go specifically to GUI rather than backend. Kinda easier for us and other folks for there to just be one option/link and then we can triage it from here when it comes in.

But thanks for the suggestion, and good to see you around here 🎉

PR #468, opened 2022-11-22
refactor: adds BaseGuiPath variable to index file

kleinfreund:
Adds the `{{.BaseGuiPath}}` string in the index.html file generated when building for production. This variable is going to be replaced when being served by the Go application. It is expected to be replaced with the GUI base path without a leading or trailing slash (e.g. "", "gui",  "dev/gui", "a/b/c/gui", etc.).

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,21 +1,26 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width,initial-scale=1" />
 
-    <link rel="icon" href="/favicon.png" />
-    <title>Manager</title>
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width,initial-scale=1" />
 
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1,21 +1,26 @@
...
-  <body>
-    <noscript>
-      <strong>
-        We're sorry but this site doesn't work properly without JavaScript enabled. Please enable it to continue.
-      </strong>
-    </noscript>
+  <link rel="icon" href="/favicon.png" />
+  <title>Manager</title>
+
+  <script type="application/json" id="kuma-config">{{.}}</script>
```
lahabana, line 11: So in dev you rely on something different?

kleinfreund: We’d do something like this: When building for production, only the if branch code will remain.

```js
if (import.meta.env.PROD) {
  const pathConfig = JSON.parse(document.querySelector('#kuma-config')?.textContent)
  console.log(pathConfig.guiBasePath)
} else {
  // Read some mock data during development
}
```

Issue #464, opened 2022-11-18
Add Gateway view


lahabana:
### Description

Add an entry in the sidebar that links to a page that shows the gateways `service+insights?gateway=true` for builtin gateways ability to drill in and see MeshGateways and MeshGatewayRoutes attached to it.

For the moment don't try to show the full builtin gateway just the policies attached to it. We can later think some more about a complex view but this is already a good first step.

johncowen:
@lahabana just a note here after closing with a PR

> ability to drill in and see MeshGateways and MeshGatewayRoutes attached to it.

From what I understand I haven't done anything to cover that ^, but not totally sure if thats related to what philipp is working on, not sure if we want to create an extra issue off here to cover the fact that this bit of this issue isn't done as far as I'm aware?

PR #469, opened 2022-11-22
feat: Add new gateway listing page 

johncowen:
Adds a new Gateway listing and detail page by reusing the DataplaneListing/Detail pages/views. The `Type` filter was also removed from the Dataplane Proxies page (seeing as it now only shows one type) but kept in the Gateways page to filter between builtin and delegated. The `Type` column was removed entirely from the Dataplane Proxy listing.

I wasn't totally sure whether to reuse this like this or not, but it worked out ok, and considering Gateways are still classed as 'Dataplane Proxies' it kinda  feels ok.

Screenshot 2022-11-22 at 16 51 23

Main approach sets a top level type on the routes meta and then decides on the tiny details between the two types dependant on that. The rest of the changes are trying to work in with what is already there for the moment, ~so just filtering out/in gateway type dataplane proxies where appropriate.~

The current implementation makes new HTTP requests whenever the `Type:` dropdown is changed, using the `gateway=` query param the backend then responds with the correctly filtered dataset.

There is at least one more general 'bug' I found whilst doing this in that the pagination buttons display on this listing whether the listing has data or not, also seems to show them whether you have other pages or not also. This is something to look at separately.

Fixes https://github.com/kumahq/kuma-gui/issues/464 (partly)


lahabana:
nit: protocol doesn't make sense on the gateway.

src/app/data-planes/views/DataPlaneListView.spec.ts
```diff
@@ -82,6 +115,32 @@ describe('DataPlaneListView', () => {
       expect(tableHeadHtml).toContain(expectedColumnHeaders[i])
     }
   })
+  test('shows correct default table columns for gateways', async () => {
+    const expectedColumnHeaders = [
+      'Status',
+      'Name',
+      'Mesh',
```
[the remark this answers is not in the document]
kleinfreund: Resolving this as it’s already being addressed elsewhere.

src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -181,18 +178,25 @@ const nextUrl = ref<string | null>(null)
 const filteredDataPlaneType = ref('All')
 const pageOffset = ref(props.offset)
 const dataPlaneOverview = ref<DataPlaneOverview | null>(null)
-
 const isMultiZoneMode = computed(() => store.getters['config/getMulticlusterStatus'])
 const dataplaneWizardRoute = computed(() => ({ name: store.getters['config/getEnvironment'] === 'universal' ? 'universal-dataplane' : 'kubernetes-dataplane' }))
+
+const isGateway = (item: DataPlaneOverview) => {
+  return typeof item.dataplane.networking.gateway !== 'undefined'
```
kleinfreund, line 185: **Suggestion**: Is there any benefit in not comparing the values directly?

```suggestion
  return item.dataplane.networking.gateway !== undefined
```

kleinfreund:
**Suggestion**: Please change `src/app/App.vue:82` to include the new view so that the layout is taking up all available space.

```diff
- const isWideContent = computed(() => typeof route.name === 'string' && ['data-plane-list-view', 'service-list-view'].includes(route.name))
+ const isWideContent = computed(() => typeof route.name === 'string' && ['data-plane-list-view', 'service-list-view', 'gateway-list-view'].includes(route.name))
```

kleinfreund:
> There is at least one more general 'bug' I found whilst doing this in that the pagination buttons display on this listing whether the listing has data or not, also seems to show them whether you have other pages or not also. This is something to look at separately.

@johncowen I’m almost certain this is just an consequence of the mock data at `src/api/mock-data/meshes/default/dataplanes+insights.json` containing a value for the `next` field. Setting it to `null` makes it so that no pagination button is shown.

kleinfreund (changes requested):

src/app/data-planes/views/DataPlaneListView.vue
johncowen: Half habit, half consistency with things like `typeof item.method === 'function'`

lahabana:
Minor thing is that there's still a "details" sidebar if you change gateway type to delegated even if the table is empty. I guess that when you toggle you should clear the selection.

Also just so you know the api has been enriched with this filters for `gateway=`:

- true: list all gateways of any type
- false: list no gateway
- provided: list only provided gateways (which would give proper pages vs client side filtering)
- builtin: list only builtin gateways

I think it might useful to leverage this feature now no?

johncowen:
> Minor thing is that there's still a "details" sidebar if you change gateway type to delegated even if the table is empty. I guess that when you toggle you should clear the selection.

Ah good spot, will update in a sec

> Also just so you know the api has been enriched with this filters for gateway=:

true: list all gateways of any type
false: list no gateway
provided: list only provided gateways (which would give proper pages vs client side filtering)
builtin: list only builtin gateways
I think it might useful to leverage this feature now no?

Yes definitely! I figured towards the end of last week that if we are doing paging on the backend, that means we also have to do sorting/filtering/searching on the backend also, otherwise the paging and the data is out of sync. I was gonna look through the docs/ask if we have ability to do these things in the HTTP API and looks like for this one we have. I'll take better look at this one also in a sec, ty!

johncowen (commented):
I left a few inline comments in a self-review. Also to note I'm pretty sure I found a weird paging bug whilst looking at this, but inside the DataOverview component - definitely would like to look at the paging impl. we use a little better

src/app/data-planes/views/DataPlaneListView.spec.ts
```diff
@@ -2,10 +2,24 @@ import { describe, expect, test } from '@jest/globals'
...
+      ...{
+        name: 'data-plane-list-view',
+        params: { mesh: 'default' },
+        meta: {
+          type: 'standard',
+        },
+      },
+      ...props || {},
+    },
+  )
```
kleinfreund, line 22: **Suggestion**:

* Avoid calling this `props` to not confuse the initial route with the quasi-default `props` parameter passed to other `renderComponent` functions.
* Use vue-router’s `RouteLocationNamedRaw` type to pass complete usable route objects

```suggestion
const DEFAULT_ROUTE: RouteLocationNamedRaw = {
  name: 'data-plane-list-view',
  params: {
    mesh: 'default',
  },
  meta: {
    type: 'standard',
  },
}

async function renderComponent(initialRoute: RouteLocationNamedRaw = DEFAULT_ROUTE) {
  await router.push(initialRoute)
```

src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -410,16 +422,23 @@ async function loadData(offset: number): Promise<void> {
...
+    })
 
     if (Array.isArray(items) && items.length > 0) {
-      items.sort(function (overviewA, overviewB) {
-        if (overviewA.name === overviewB.name) {
-          return overviewA.mesh > overviewB.mesh ? 1 : -1
-        } else {
-          return overviewA.name.localeCompare(overviewB.name)
-        }
-      })
```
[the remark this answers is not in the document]
kleinfreund: I’m fine with that. Feel free to resolve.

src/app/data-planes/views/DataPlaneListView.spec.ts
kleinfreund: I think the reason why passing `meta` to a `router.push` doesn’t work with the types available is that `meta` is part of what _defines_ a route but not part of the things _identify_ a route **in navigation**. In other words: `router.push` accepts things that are represented in the URL and meta isn’t taken into account when vue-router identifies a matching route. I’m pretty sure you could call `router.push` with an arbitrary value for `meta` (assuming one were to disable type checking for that call) and it wouldn’t break that call as long as the rest of the route location identifies an existing route.

What you’re trying in this test is "navigate to the data plane list view but act as if its meta data field `type` is `gateway`". Instead, you should just ask the test to navigate to the `gateway-list-view`.

**Setup**:

```suggestion
const DEFAULT_ROUTE: RouteLocationNamedRaw = {
  name: 'data-plane-list-view',
  params: {
    mesh: 'default',
  },
}
async function renderComponent(initialRoute: RouteLocationNamedRaw = DEFAULT_ROUTE) {
  await router.push(initialRoute)
```

**Usage**:

```ts
const wrapper = await renderComponent({
  name: 'gateway-list-view',
  params: {
    mesh: 'default',
  },
})
```

---

> Not quite sure what you meant by this? Are you referring to the name of the local argument for the function? If so happy to change the name if it's just that, but it's a local variable so I don't think it matters that much, how about `partialRouteLocation`?

My point is merely one of naming. Props has a very fixed and defined meaning in a Vue application and the parameter expected here is _not_ a prop but a route location.

johncowen: Ah I getcha, I think. Is what you are saying is I just need to change the `name` to either 'dataplane-list-view' or 'gateway-list-view' here? Then whatever the `meta` is in the routing configuration in router.ts will then be applied?

I'll give that a go.

As an aside/observation, if I've understood you correctly, this definitely feels like we are testing the entire routing/app rather than just the isolated component. All in all I think I'd rather just use component props, I think I tried that before trying out meta, might give that another go instead also 🤔

lahabana:
Use delegated instead of standard in the type column please

src/app/data-planes/views/DataPlaneListView.spec.ts
kleinfreund: > Ah I getcha, I think. Is what you are saying is I just need to change the name to either 'dataplane-list-view' or 'gateway-list-view' here? Then whatever the meta is in the routing configuration in router.ts will then be applied? i.e. I shouldn't be setting meta at all.

Yes, that’s correct. `router.push` will resolve the route location provided to it against the route definitions the router instances has registered. It will try to match by name and params to find the correct definition and use that as the route meaning that the meta information will be set correctly.

> As an aside/observation, if I've understood you correctly, this definitely feels like we are testing the entire routing/app rather than just the isolated component.

I don’t think that’s the case. We have some components that just need access to route-specific information and for that, it’s important to supply this in the tests.

In this test, we navigate to the data-plane-list-view in the router because that’s accurate, but we only do that because of various access to `route` information like the mesh parameter or now the meta `type` property. This doesn’t test whether the navigation takes places. Think of it more like what you’d do with mocking. You can achieve similar results by mounting the component with global mocks:

```js
mount(Component, {
  global: {
    mocks: {
      // Works for accessing `this.$router`, won’t work for accessing `router` that comes from `useRouter`.
      $router: {
        params: { mesh: 'default' },
      },
    },
  },
})

kleinfreund: Given that we don’t have an end-to-end test environment (yet), I care less about the philosophical purity of the test approach. As long as the test works reasonably well and does relevant assertions, it’s good for me.

Technically speaking, these tests can’t be end-to-end tests because they’re run in a Node.js environment with DOM simulation and all that.

johncowen: Right yeah, so these should aim to be isolated component rather than e2e

kleinfreund: Yes, but calling `router.push` here doesn’t make them end-to-end tests or less isolated.

kleinfreund (approved):

src/app/data-planes/views/DataPlaneListView.spec.ts
johncowen: In my mind it's depending on router functionality, which makes the component test less isolated. But whichever, I don't think it matters too much, but I'm gonna have a look a removing the router either via mocking or look again at trying to pass props straight into the component, even if its just for the learning experience.

PR #470, opened 2022-11-23
chore(deps): setup dependabot

mmorel-35:
Provides dependabot configuration to keep npm and github-actions up-to-date.
This is based on [kuma configuration](https://github.com/kumahq/kuma/blob/master/.github/dependabot.yml) 

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"    
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
```
kleinfreund, line 3: **Issue**: This project uses yarn instead of npm.

```suggestion
  - package-ecosystem: "yarn"
```

kleinfreund (changes requested):
Hey, thanks, you’re getting ahead of us with this one. I had intentions to set-up dependabot as well. One thing that I want to do before this goes in is pinning the dependencies by hand. I know dependabot would do this on its first PR, but I want to ensure this works by hand.

.github/dependabot.yml
mmorel-35: Take a look over here : https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem 

package-ecosystem for yarn is also npm

kleinfreund: Ah, yeah I remember I ran into this before. Certainly very confusing.

kleinfreund (approved):

PR #471, opened 2022-11-23
chore(deps): updates all dependencies

kleinfreund:
Updates all dependencies.

Removes `data-v` attributes from snapshots to avoid minor/patch releases of Kongponents to breaking our tests solely on account of all the `data-v` attributes changing.

Adds workaround plugin for Vite issue https://github.com/vitejs/vite/issues/2415.


kleinfreund:
**Comment**: Need to tweak the removal of data-v-* attributes a bit more.

jest/jest-replace-attribute-snapshot-serializer.ts
```diff
@@ -33,6 +33,13 @@ export function replaceAttributesSnapshotSerializer(
         }
       }
 
+      // Removes “data-v-*” attributes.
+      for (const attribute of value.attributes) {
+        if (attribute.name.startsWith('data-v-')) {
+          value.removeAttribute(attribute.name)
+        }
+      }
+
```
johncowen, line 42: I understood that this is removing the data-v attributes? Wasn't sure whether i should still be seeing them in the snapshots? On a brief scan of the diff here, it looks like they've been removed from some places but not others?

johncowen (approved):
Looks good to me, I had a question on the data-v attributes but thats non-blocking

jest/jest-replace-attribute-snapshot-serializer.ts
kleinfreund: I'm struggling a bit with getting the custom serializer to go through all nodes. I got most of them removed locally but there are still a handful left. Fixing that next week.

johncowen: Awesome, sounds great! Will be good once they are all gone

kleinfreund: Figured it out. Much happier with removing those attributes now. This should minimize the amount of times we have to regenerate these snapshots.

PR #474, opened 2022-11-30
refactor: reads GUI base path from DOM

### LINKED ISSUE #454: Use index.html content to determine API base URL and GUI base path
### Description

In https://github.com/kumahq/kuma/pull/5357, we added templating capabilities to the index.html that's being served by the Kuma application.

* If present, a `<script id="kuma-config">{{.}}</script>` tag is populated with a JSON string representing a JavaScript object with the following interface:

  ```ts
  {
    apiUrl: string
    baseGuiPath: string
    version: string
  }
  ```

  Example:

  ```
  {
    "apiUrl": "https://foo.bar.com:8080/foo",
    "baseGuiPath": "/ui",
    "version":"0.0.0"
  }
  ```
* Any references to `{{.BaseGuiPath}}` are replaced with the base GUI path which is a path segment that **always** has a leading slash and that **only** has a trailing slash if that’s also a leading slash (i.e. the value is `"/"`).

## To do

* Prepare the index.html file for templating by including `<script id="kuma-config">{{.}}</script>` before the `</body>` tag.
* Ensure that **every single file-relative reference to GUI assets** is adjusted to start with `{{.BaseGuiPath}}`. Such references may not be prefixed with any other character (see discussion in https://github.com/kumahq/kuma/pull/5357#discussion_r1030137519).
  * This will require post-processing the index.html file after the build step as Vite won’t allow us to set `base` to `"{{.BaseGuiPath}}"` without adding leading and trailing slashes to it.

<details><summary>Outdated description</summary>

### Description

In https://github.com/kumahq/kuma/pull/5295, we’re adding a new static resource that’s served relative to the GUI’s `index.html` file (e.g. under http://localhost:5681/gui/config.json). It will contain (1) the GUI base path and (2) the API base URL.

Make a request to this config.json file before any other API calls and specifically before initializing the Vue router and Vuex store. With its result, ensure that:

* the router is initialized with GUI base path (instead of the current logic which can be removed)
* the RestClient used by kumaApi is initialized with the API base URL. This will automatically cause all future API calls to use it.

**Blocked by**:

* https://github.com/kumahq/kuma/pull/5295

</details>

kleinfreund:
Adds a `<base>` tag to the index.html file which will be set to an actual base path during development and the GO template placeholder `{{.BaseGuiPath}}` for production. Upon serving the GUI application, the placeholder will be replaced with the actual GUI base path.

Adds the `kuma-config` script tag containing `{{.}}` to the index.html file which will be replaced via Go templating with a basic data structure that, among other things, will hold the actual API base URL and the GUI base path.

Reads the API base URL and the GUI base path from the index.html file replacing any attempt of determining it from the current URL which is error-prone and inflexible.

Resolves #454.


kleinfreund:
**Issue**: This worked great in the development environment and in the Netlify deployment preview, but it’s utterly broken in Kuma proper. I’ve ran a build and copied the files into Kuma for testing:

```sh
cd kuma-gui
yarn run build

cd ../kuma
git pull
rm -rf ~/kong/kuma/app/kuma-ui/pkg/resources/data \
  && mkdir -p ~/kong/kuma/app/kuma-ui/pkg/resources/data \
  && cp -r ~/kong/kuma-gui/dist/* ~/kong/kuma/app/kuma-ui/pkg/resources/data
make dev/tools
make build
KUMA_DP_SERVER_AUTH_TYPE=none make run/universal/memory
```

With …

* `<base href="/gui" />` in the `head`,
* … asset references like `<link rel="stylesheet" href="./assets/index.dc35afea.css">` in index.html,
* … and the absolute resource URL http://localhost:5681/gui/assets/index.dc35afea.css resolving

the GUI doesn’t load at all because the resources are being requested from a URL missing the GUI path (e.g. http://localhost:5681/assets/index.dc35afea.css).

src/main.ts
```diff
@@ -54,13 +59,55 @@ async function initializeVue() {
     store.dispatch('fetchPolicies'),
   ])
 
-  const router = await createRouter(store.state.policies)
+  const guiBasePath = readGuiBasePathFromDom()
```
johncowen, line 62: How come we need this (and the function itself) I understood we can get almost the same value from `pathConfig.baseGuiPath` just without the trailing slash, which we can just add on the end? In fact I think there's something a little of here considering the current default fallback in the `readGuiBasePathFromDom` function doesn't end in a slash and the `<base href="">` value will always end in a slash.

src/main.ts
```diff
@@ -20,6 +21,7 @@ import '@/assets/styles/typography.scss'
 import '@/assets/styles/inputs.scss'
 import '@/assets/styles/components.scss'
 import '@/assets/styles/transitions.scss'
+import { PATH_CONFIG_DEFAULT } from './constants.js'
```
johncowen, line 24: How come this references a `.js` file?

src/utilities/kumaDpServerUrl.ts
```diff
@@ -1,5 +1,5 @@
 export function kumaDpServerUrl(): string {
-  const dpServerUrl = String(import.meta.env.VITE_KUMA_DP_SERVER_URL)
+  const dpServerUrl = String(import.meta.env.VITE_KUMA_DP_SERVER_URL) + '/'
```
johncowen, line 2: Do you know why this is wrapped in a `String`, tiny nit, but would this be fine as a template literal?

vite.config.js
```diff
@@ -1,32 +1,46 @@
...
+      vue(),
+      svgLoader(),
+      createHtmlPlugin({
+        inject: {
+          data: {
+            /**
+             * Adds the appropriate GUI base path placeholder to the index.html file. It is going to be replaced using server-side templating when serving the GUI application in production.
+             */
+            baseGuiPath: mode === 'production' ? '{{.BaseGuiPath}}' : '/gui',
+            // **TIP**: Change this value to test various GUI base path scenarios.
```
johncowen, line 31: What are your thoughts on switching the `{{.}}` in an out here instead of dealing with it inside the application? I think we mentioned this idea elsewhere also. Feels like you could then reference `{baseGuiPath: '/gui'}.baseGuiPath` somehow here also.

```javascript
const config = {
  baseGuiPath: '/gui',
  ...
}
...
baseGuiPath: mode === 'production' ? '{{.BaseGuiPath}}' : config.baseGuiPath,
config: mode === 'production' ? '{{.}}' : JSON.stringify(config)
```

That kinda thing ^, then we remove all references to the string `/gui` from the application code entirely, its all just config.

johncowen (commented):
I left a few questions before taking it for a spin! Be right back once I've taken it for a spin.

src/main.ts
kleinfreund: VS Code adds the extension when auto-completing imports and I forgot to remove it. Preferably, I would use the .ts extension, but TypeScript forbids this still.

src/utilities/kumaDpServerUrl.ts
kleinfreund: Historically, environment variables where types as something like `string | undefined` to honor the fact that accessing `process.env.STUFF` resolves to `undefined` if the variable isn’t actually defined.

Now, we have explicit types for the variables that are always defined and this can be dropped.

Also, I prefer not to use template literals if they don’t add any value. Here, the concatenation is neither more clear or shorter with template literals. Regular string concatenation is just fine.

vite.config.js
kleinfreund: Yeah I like that idea. I’ve changed this around now.

src/main.ts
kleinfreund: You’re right; it makes more sense to read this from the config object as well.

One thing that’s somewhat bothersome now is the following:

* With replacing the config in non-production environments in the Vite configuration, we need to define the default path config object there. Specifically, we need to define it _after the environment variables have been populated_ because the object needs to have access to `VITE_KUMA_API_SERVER_URL`.
* This means we can’t define this object in `src/constants.ts` and import it in vite.config.js because the import would happen _before_ we have a chance to run `dotenv.config()` which allows us access to environment variables in vite.config.js.
* This means we can’t import the same default config in main.ts

johncowen: > This means we can’t import the same default config in main.ts

Maybe I don't follow but my question would be, why do we need a default config in main.ts? We rely on the config always being set and passed to us from the backend (or the build tool during dev/testing/preview), so in that case there is no 'default' as the config is always passed to us. There shouldn't be a case where we aren't passed a config.

kleinfreund: Yes, this is ideally the case, but this is a critical point of failure: if for some reason, the templating fails, the entire application breaks immediately despite us being able to offer most users a working product seeing that the majority of people do use the quasi standard /gui GUI base path and the default API URLs. I consider this a defensive measure against this unlikely but severe risk.

johncowen: In the unlikely event of the templating failing should that not be dealt with in the backend? I'm only saying as it means the problem you've outlined above becomes a non-problem and we don't have to worry about it.

kleinfreund: How do you imagine the backend anticipating a failure in templating? I don’t think that makes more sense than doing it right here. There is nothing inherently wrong with duplicating this structure and while it would be nice to re-use the same object, we can’t (at least I don’t know how).

johncowen: > How do you imagine the backend anticipating a failure in templating?

Well thats up to backend I would say, but at a guess they could just fallback to a default JSON the same as you are suggesting we do here (or even a default non-templated HTML file). Honestly I don't think this is ever going to be a major issue, and if it does I'd say it should be surfaced in the backend and dealt with there i.e. fix it where it broke.

All this aside, is this is causing no issues for you or you have a workaround you are happy with then I really don't mind either way.

vite.config.js
```diff
@@ -1,32 +1,61 @@
...
+dotenv.config()
+
+/**
+ * @type {PathConfig}
+ */
+const PATH_CONFIG_DEFAULT = {
+  baseGuiPath: '/gui',
+  // **TIP**: Change this value to test various GUI base path scenarios.
+  // baseGuiPath: '/dev/gui',
+  apiUrl: /** @type {string} */ (process.env.VITE_KUMA_API_SERVER_URL),
```
johncowen, line 20: Just incase, I think the default for this in prod from the backend is `/` (the env var looks to be full hostname:port) i.e. without the hostname:port. Should we mirror that so its just `/` here also? Not sure if there is a restriction on running msw on the same port as vite, I would guess not?

johncowen (approved):
I took this for a spin, and all looks good to me!

I had one more question which was just around consistency with our most likely production configuration values that I left inline, no biggie though

vite.config.js
kleinfreund: Let me try this in Kuma with a local build. I think this should work.

In our `RestClient` class, we’d always set the correct URL during development (where the API origin is _different_ from the GUI origin) and in production we’d use `window.location.origin`. I’ve made those changes in https://github.com/kumahq/kuma-gui/pull/474/commits/66ca34e7dfc81d0351be44d73f7edbf7cf579afe.

kleinfreund: Perhaps we can get rid of this environment variable by proxying requests from 8080 to 5681 during development, but I’ll consider this separately from this PR.

PR #476, opened 2022-11-30
fix: accessing dataplaneInsight prematurely

### LINKED ISSUE #475: Failed to render dpp list with no subscription
### What happened?

With the response to `dataplane+insights`:

```json
{
 "total": 1,
 "items": [
  {
   "type": "DataplaneOverview",
   "mesh": "default",
   "name": "redis-1",
   "creationTime": "2022-11-30T11:27:36.245951+01:00",
   "modificationTime": "2022-11-30T11:27:36.245951+01:00",
   "dataplane": {
    "networking": {
     "address": "192.168.0.1",
     "inbound": [
      {
       "port": 9000,
       "servicePort": 6379,
       "tags": {
        "kuma.io/service": "redis"
       }
      }
     ]
    }
   }
  }
 ],
 "next": null
}
```

I get:

```
index.c8163df9.js:1 TypeError: Cannot read properties of undefined (reading 'subscriptions')
    at Fee (index.c8163df9.js:30:15896)
    at yd.fn (DataPlaneListView.89096fce.js:1:2908)
    at yd.run (index.c8163df9.js:1:6278)
    at get value [as value] (index.c8163df9.js:1:14369)
    at ie (index.c8163df9.js:1:13446)
    at Proxy.<anonymous> (DataPlaneListView.89096fce.js:1:4055)
    at ic (index.c8163df9.js:1:17784)
    at yd.Y [as fn] (index.c8163df9.js:1:41522)
    at yd.run (index.c8163df9.js:1:6278)
    at b.update (index.c8163df9.js:1:41803)
```

We wrongly assume there's always a subscription which is not the case if no dpp was ever started

kleinfreund:
Fixes a premature access to the `dataplaneInsight` property which can be absent.

Fixes #475.

### DIFF
--- a/src/app/data-planes/components/DataPlaneDetails.vue
+++ b/src/app/data-planes/components/DataPlaneDetails.vue
@@ -300,7 +300,7 @@ const dataPlaneVersions = computed(() => getVersions(props.dataPlaneOverview.dat
 const rawDataPlane = computed(() => stripTimes(props.dataPlane))
 const mtlsData = computed(() => parseMTLSData(props.dataPlaneOverview))
 const insightSubscriptions = computed(() => {
-  const subscriptions = Array.from(props.dataPlaneOverview.dataplaneInsight.subscriptions)
+  const subscriptions = Array.from(props.dataPlaneOverview.dataplaneInsight?.subscriptions ?? [])
 
   subscriptions.reverse()
 
@@ -316,7 +316,7 @@ const kumaDocsVersion = computed(() => {
 const filteredTabs = computed(() => warnings.value.length === 0 ? tabs.filter((tab) => tab.hash !== '#warnings') : tabs)
 
 function setWarnings() {
-  const subscriptions = props.dataPlaneOverview.dataplaneInsight.subscriptions
+  const subscriptions = props.dataPlaneOverview.dataplaneInsight?.subscriptions ?? []
 
   if (subscriptions.length === 0 || !('version' in subscriptions[0])) {
     return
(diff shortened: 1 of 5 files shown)


johncowen (commented):
I wondered if there was a more centralized place to do this so we don't have to remember to check whether its undefined every time its used?

Say for example when the request comes in in `kumaApi`? Could we detect, as soon as we receive the request, whether any of the `items` has an undefined `dataplaneInsight/subscriptions` and if so set to the empty array? Then we can guarantee everywhere else in our app that `dataplaneInsight/subscriptions` is an array (even if its empty) so we don't have to add these 'alarms' and there are 'no surprises'

kleinfreund:
> I wondered if there was a more centralized place to do this so we don't have to remember to check whether its undefined every time its used?
> 
> Say for example when the request comes in in `kumaApi`? Could we detect, as soon as we receive the request, whether any of the `items` has an undefined `dataplaneInsight/subscriptions` and if so set to the empty array? Then we can guarantee everywhere else in our app that `dataplaneInsight/subscriptions` is an array (even if its empty) so we don't have to add these 'alarms' and there are 'no surprises'

This is far beyond the scope of this PR. Since `dataplaneInsight` can be omitted, I’ll treat it as such in all the current code _in order to fix the bug_. I don’t think it makes sense to automatically serialize some convenient fallback structure in general. If it did, we would change the API to provide that structure automatically. It’s a conscious choice of our APIs to generally omit empty collections rather than including it (i.e. the collection property is absent rather than present and being set to `[]`). While I don’t like this style, it’s not something we should revert in the frontend. That would introduce a level of complexity that we don’t have a clear need for in my opinion.

johncowen (approved):
> That would introduce a level of complexity that we don’t have a clear need for in my opinion.

Up to you, although I think it would pretty much be a one-liner as opposed to all the changes here (which includes all the widening to ` | undefined`), plus it works for everyone who works on this in the future without them having to think about it.

I'm guessing there are no other occurrences of this other than the ones in the PR here?

PR #478, opened 2022-12-01
refactor: removes unnecessary fonts

kleinfreund:
Moves CSS entry points into the HTML (where they belong).

Removes all unused fonts and removes legacy font formats. The remaining cuts of Roboto are regular, italic, bold, and bold-italic which cuts down our used font weights to normal (i.e. 400) and bold (i.e. 700).


index.html
```diff
@@ -11,8 +11,11 @@
   <title>Manager</title>
 
   <script type="application/json" id="kuma-config"><%- config %></script>
-
   <script type="module" src="/src/main.ts"></script>
+
+  <link rel="stylesheet" href="/node_modules/@kong/kongponents/dist/style.css">
+  <link rel="stylesheet" href="/node_modules/@kong/kongponents/dist/_variables.scss">
```
johncowen, line 17: How come you chose to put these here instead of importing from the CSS?

johncowen (approved):
1 question but ✅

index.html
kleinfreund: Moved them into main.scss now.

PR #488, opened 2022-12-07
chore(deps): removes semver-compare

kleinfreund:
Removes semver-compare in favor of using the semver provided compare function. We had both packages installed and only used semver-compare. The latter has last been updated 8 years ago and its source code is no longer available.


johncowen (approved):
🚀

PR #491, opened 2022-12-08
refactor: replaces Tailwind’s base styles

kleinfreund:
Removes the last two Kongponents variable overrides. These changed the theme of buttons which affects us as follows: (1) primary buttons now have the Kongponent default blue background and (2) buttons that navigate to a wizard (e.g. "Create mesh") now use the "creation" appearance giving them a green background.

Removes all remaining web fonts in favor of a system font stack. While this might cause the UI to look subtly different, most probably will probably not notice this change. On my system, the difference between Roboto and the system fonts I have installed are very small.

Removes the Tailwind base and component styles in favor of our own custom styles. We didn’t make use of the component styles and the base styles have some questionable defaults (like resetting the list styles of `ul` and `ol`, the sizes of all headings, etc.).


src/store/storeConfig.ts
```diff
@@ -284,8 +286,13 @@ export const storeConfig: StoreOptions<State> = {
...
     // update the selected mesh
-    updateSelectedMesh({ commit }, mesh) {
-      ClientStorage.set('selectedMesh', mesh)
+    updateSelectedMesh({ commit }, mesh: string | null) {
+      if (mesh !== null) {
+        ClientStorage.set('selectedMesh', mesh)
+      } else {
+        ClientStorage.remove('selectedMesh')
+      }
+
```
johncowen, line 295: Is this a little bugfix for something? Is it worth splitting this out into a separate PR if so?

johncowen (commented):
Looks to my eye like there is a bit of stuff in here that fixes little bugs that have nothing to do with replacing tailwind styles, refactoring CSS or anything else in the description

I like that we are removing the overwrites though, I also like that we are using icons for icons instead of HTML entities.

Are you ok splitting off the little extra fixes in here into a separate PR, I find it helps when people (users and engineers) see something not working, and then they see a PR fixing it. Also helps with the amount of change in every PR.

Lastly just to say, like an occasional "oh I just fixed that up while I was here" is fine by me, just I think I've seen a lot of mixed PRs and PRs with lots of change - but maybe I'm just used to a different approach. I dunno, all in all completely up to you.

kleinfreund:
@johncowen Sure, no problem, I’ll split this up in smaller parts.

src/store/storeConfig.ts
kleinfreund: See https://github.com/kumahq/kuma-gui/pull/493.

johncowen (approved):
Looks good to me!

Just noting what we chatted about yesterday, that moving forwards we are preferring not to use Tailwind when possible, so this is great!

PR #493, opened 2022-12-09
fix: using non-existing mesh from local storage

kleinfreund:
Fixes an issue with using a stored value for the `selectedMesh` state when that value doesn’t represent an existing mesh. Then, we’d still let users navigate to routes associated to this resource. This happens more commonly for developers frequently switching around their data during development.

### DIFF
--- a/src/store/storeConfig.ts
+++ b/src/store/storeConfig.ts
@@ -243,6 +243,8 @@ export const storeConfig: StoreOptions<State> = {
         const newStoredMesh = ClientStorage.get('selectedMesh')
         if ((newStoredMesh === null || newStoredMesh === 'all') && state.meshes.items.length > 0) {
           dispatch('updateSelectedMesh', state.meshes.items[0].name)
+        } else if (newStoredMesh !== null && !state.meshes.items.some((mesh) => mesh.name === newStoredMesh)) {
+          dispatch('updateSelectedMesh', null)
         }
       }
 
@@ -284,8 +286,13 @@ export const storeConfig: StoreOptions<State> = {
     },
 
     // update the selected mesh
-    updateSelectedMesh({ commit }, mesh) {
-      ClientStorage.set('selectedMesh', mesh)
+    updateSelectedMesh({ commit }, mesh: string | null) {
+      if (mesh !== null) {
+        ClientStorage.set('selectedMesh', mesh)
+      } else {
+        ClientStorage.remove('selectedMesh')
+      }
+
       commit('SET_SELECTED_MESH', mesh)
     },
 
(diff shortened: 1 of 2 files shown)


src/store/storeConfig.ts
```diff
@@ -243,6 +243,8 @@ export const storeConfig: StoreOptions<State> = {
         const newStoredMesh = ClientStorage.get('selectedMesh')
         if ((newStoredMesh === null || newStoredMesh === 'all') && state.meshes.items.length > 0) {
           dispatch('updateSelectedMesh', state.meshes.items[0].name)
+        } else if (newStoredMesh !== null && !state.meshes.items.some((mesh) => mesh.name === newStoredMesh)) {
+          dispatch('updateSelectedMesh', null)
```
johncowen, line 247: Kinda not related to this PR, but how does 'all' come into this? Is that like a backcompat/upgrade thing? Unlikely, but if I have a mesh called `all` it gets reset to the first mesh in the list. Not a blocker as it's unlikely, but just to explain as to why the `all` value got me wondering.

johncowen (approved):
🚀 thanks for doing this!

src/store/storeConfig.ts
kleinfreund: Yeah, this is purely an artifact of removing a feature. You used to be able to select "all" in the mesh selector which would remove any views being mesh specific. E.g. the data plane list view would show _all_ data plane proxies for _all_ meshes.

That being said, I think the newly added logic would also handle this, so maybe I can replace the special casing of all with it.

johncowen: Gotcha, thanks for the explanation

kleinfreund: Simplified this in https://github.com/kumahq/kuma-gui/pull/493/commits/6e640aecd24cd79fba48c56508e2b7237ea35258.

johncowen: Ah wait looking at the previous commit 😆 all good! ... I think 👀

johncowen: Yeah defo think a test, I think we can get to an undefined `state.meshes.items` here? A test would definitely help I think.

kleinfreund: How can `state.meshes.items` become `undefined`? Even if `items` is omitted in the response, `fetchMeshList` has logic against this ensuring that `items` is always an array.

One thing that can technically happen is the list being empty and the fallback logic trying to pick its first element. I added tests for this and updated the logic.

johncowen: Yeah that's what I meant 😸 

> looks like state.meshes.items can be zero length?

Thanks for adding that check and the tests!

PR #494, opened 2022-12-09
chore: uses dedicated icons instead of HTML entities

kleinfreund:
Uses dedicated icon components instead of HTML entities and sets standard button icon sizes.

### DIFF
--- a/src/app/common/DataOverview.vue
+++ b/src/app/common/DataOverview.vue
@@ -137,29 +137,6 @@
             </span>
           </template>
 
-          <template #selected="{row}">
-            <a
-              class="data-table-action-link"
-              :class="{ 'is-active': selectedRow === row.name }"
-            >
-              <span
-                v-if="selectedRow === row.name"
-                class="action-link__active-state"
-              >
-                ✓
-
-                <span class="sr-only">Selected</span>
-              </span>
-
-              <span
-                v-else
-                class="action-link__normal-state"
-              >
-                View
-              </span>
-            </a>
-          </template>
-
           <!--- dp Version --->
           <template #dpVersion="{ row, rowValue }">
             <div
@@ -211,7 +188,8 @@
                   :icon="row.warnings.length > 0 ? 'warning' : 'info'"
                   :color="row.warnings.length > 0 ? 'var(--black-75)' : 'var(--blue-500)'"
                   :secondary-color="row.warnings.length > 0 ? 'var(--yellow-300)' : undefined"
-                  size="20"
+                  size="16"
+                  hide-title
                 />
               </template>
               Details
@@ -444,25 +422,6 @@ function getRowAttributes({ name }: any): Record<string, string> {
 
 .with-warnings {
   color: var(--yellow-500);
-}
-
-.data-table-action-link {
-  padding: 0;
-}
-
-.action-link__active-state {
-  --size: 18px;
-
-  display: block;
-  width: var(--size);
-  height: var(--size);
-  line-height: var(--size);
-  border-radius: 50%;
-  background-color: var(--logo-green);
-  margin: 0 0 0 var(--spacing-xxs);
-  color: var(--white);
-  font-size: 13px;
-  text-align: center;
 }
 </style>
 
--- a/src/app/common/NotificationIcon.vue
+++ b/src/app/common/NotificationIcon.vue
@@ -5,8 +5,9 @@
     @click="openModal"
   >
     <KIcon
-      color="var(--yellow-300)"
       icon="notificationBell"
+      color="var(--yellow-300)"
+      hide-title
     />
 
     <span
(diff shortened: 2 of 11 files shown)


src/app/common/NotificationIcon.vue
```diff
@@ -5,8 +5,9 @@
     @click="openModal"
   >
     <KIcon
-      color="var(--yellow-300)"
       icon="notificationBell"
+      color="var(--yellow-300)"
+      hide-title
```
johncowen, line 10: Kind of a consequence of kongponents I think, but did we lose the `Notifications` word out of the HTML here? Quick aria-label or visually hidden span to put it back?

src/app/common/__snapshots__/PaginationWidget.spec.ts.snap
```diff
@@ -45,7 +45,36 @@ exports[`PaginationWidget.vue renders pagination 1`] = `
...
+      <svg
+        fill="none"
+        viewBox="0 0 20 20"
+        xmlns="http://www.w3.org/2000/svg"
+      >
+        
+  
+        <title>
+          Back
+        </title>
```
johncowen, line 60: `Back Previous`? Does this one and `Next Forward` need hiding? Or is it because the snapshots are reporting the DOM before kongponents has had a chance to remove it?

src/app/wizard/components/StepSkeleton.vue
```diff
@@ -244,73 +244,29 @@ export default {
...
-    > * {
-      margin-bottom: 16px;
-    }
-
-    .wizard-steps__sidebar {
-      border-radius: 6px;
-    }
-  }
+.wizard-steps__sidebar {
+  border-radius: 6px;
```
johncowen, line 269: What's all this change for?

johncowen (commented):
Spotted some things I had questions on, I think its mainly down to kongponents being how they are tho, but I left a couple of questions

src/app/common/__snapshots__/PaginationWidget.spec.ts.snap
kleinfreund: Hm, it’s confusing to me why the `hide-title` prop doesn’t hide it here but in the notification icon. Let me look into that.

kleinfreund: I can only imagine this to be an issue with the KIcon implementation. It does some DOM manipulation and maybe it doesn’t always remove the `title` correctly.

I added `aria-hidden="true"` to those icons now to prevent those from being announced in screen readers. I don’t think we can fix this on our side.

src/app/wizard/components/StepSkeleton.vue
kleinfreund: Unrelated. Removed it.

src/app/common/NotificationIcon.vue
kleinfreund: I’ve opted to removes `hide-label` here because this doesn’t have a good label otherwise, yeah.

johncowen: Sounds good!

src/app/common/__snapshots__/PaginationWidget.spec.ts.snap
johncowen: LGTM!

What's the deal with kongponents, is it worth filing an issue over there or do you think it is just a test/snapshotted too early thing?

src/app/common/DataOverview.vue
```diff
@@ -445,25 +423,6 @@ function getRowAttributes({ name }: any): Record<string, string> {
...
-  width: var(--size);
-  height: var(--size);
-  line-height: var(--size);
-  border-radius: 50%;
-  background-color: var(--logo-green);
-  margin: 0 0 0 var(--spacing-xxs);
-  color: var(--white);
-  font-size: 13px;
-  text-align: center;
-}
```
johncowen, line 466: What's all this change/removal?

kleinfreund: I couldn’t find any use of this column. I think we used to highlight the selected row with some sort of green check mark icon, but that’s not present anymore in any view.

src/app/common/__snapshots__/PaginationWidget.spec.ts.snap
kleinfreund: Not sure. I don’t exactly know where it breaks and it’s definitely not a pressing matter. If I have some extra time, I can look into it more, but that’s not today.

src/app/common/DataOverview.vue
johncowen: Gotcha ty!

johncowen (approved):
🚀

PR #495, opened 2022-12-09
refactor: replaces web fonts with system font stack

kleinfreund:
Removes all remaining web fonts in favor of a system font stack. While this might cause the UI to look subtly different, most probably will probably not notice this change. On my system, the difference between Roboto and the system fonts I have installed are very small.


src/assets/styles/_variables.scss
```diff
@@ -1,8 +1,8 @@
 $logo-purple: #290b53;
 
 :root {
-  --base-font-family: Roboto, -apple-system, system-ui, BlinkMacSystemFont, 'Segoe UI', 'Helvetica Neue', Arial, sans-serif;
-  --code-font-family: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
+  --font-family-sans: Roboto, -apple-system, system-ui, BlinkMacSystemFont, 'Segoe UI', 'Helvetica Neue', Arial, sans-serif;
+  --font-family-mono: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
```
johncowen, line 5: Is the Roboto here just incase someone has Roboto installed on their machine, seeing as we've removed Roboto fonts here? If so should we add `Roboto Mono` to mono also?

kleinfreund: Huh, strange, this not intended. That’s just the old value again. I had actually used the same font stack GitHub uses (i.e. `-apple-system,BlinkMacSystemFont,"Segoe UI","Noto Sans",Helvetica,Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji"`). I’ve fixed that now.

Very few people should’ve Roboto installed natively on their system so there isn’t much use in including it anyway.

johncowen: > Very few people should’ve Roboto installed natively on their system

😆 Yeah thats what I thought!

src/assets/styles/main.scss
```diff
@@ -1,9 +1,9 @@
 @import '/node_modules/@kong/kongponents/dist/style.css';
 @import '/node_modules/@kong/kongponents/dist/_variables.scss';
 @import 'third-party/tailwind.css';
-@import 'third-party/kongponents';
-@import 'fonts';
+@import 'kongponents-overrides';
 @import 'variables';
 @import 'base';
+@import 'kongponents-overrides';
```
johncowen, line 7: Double import?

kleinfreund: Yes, good catch.

johncowen (approved):
:shipit:

PR #496, opened 2022-12-12
refactor: uses Kongponents colors

kleinfreund:
Removes the last two Kongponents variable overrides. These changed the theme of buttons which affects us as follows: (1) primary buttons now have the Kongponent default blue background and (2) buttons that navigate to a wizard (e.g. "Create mesh") now use the "creation" appearance giving them a green background.

Removes all custom colors in favor of Kongponents standard colors.

### DIFF
--- a/src/app/common/EntityStatus.vue
+++ b/src/app/common/EntityStatus.vue
@@ -46,7 +46,7 @@ const entityStatusClassNames = computed(() => {
 }
 
 .entity-status.is-online {
-  color: var(--green-400);
+  color: var(--green-500);
 }
 
 .entity-status.is-offline {
@@ -54,7 +54,7 @@ const entityStatusClassNames = computed(() => {
 }
 
 .entity-status.is-degraded {
-  color: var(--custom-orange);
+  color: #ff8040;
 }
 
 .entity-status.is-not-available {
--- a/src/app/common/PolicyTypeTag.vue
+++ b/src/app/common/PolicyTypeTag.vue
@@ -107,7 +107,7 @@ const policy = computed(() => POLICIES[props.policyType])
 
 <style lang="scss" scoped>
 .policy-type-tag {
-  display: inline-flex;
+  display: flex;
   align-items: center;
   gap: var(--spacing-xs);
 }
--- a/src/app/data-planes/views/DataPlaneListView.vue
+++ b/src/app/data-planes/views/DataPlaneListView.vue
@@ -72,7 +72,6 @@
           </KDropdownMenu>
 
           <KButton
-            class="add-dp-button"
             appearance="creation"
             :to="dataplaneWizardRoute"
             icon="plus"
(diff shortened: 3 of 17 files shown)


src/app/common/PolicyTypeTag.vue
```diff
@@ -107,7 +107,7 @@ const policy = computed(() => POLICIES[props.policyType])
 
 <style lang="scss" scoped>
 .policy-type-tag {
-  display: inline-flex;
+  display: flex;
```
johncowen, line 110: Just out of interest, what was the issue here?

src/app/common/EntityStatus.vue
```diff
@@ -46,15 +46,15 @@ const entityStatusClassNames = computed(() => {
...
+  color: var(--green-500);
 }
 
 .entity-status.is-offline {
   color: var(--red-600);
 }
 
 .entity-status.is-degraded {
-  color: var(--custom-orange);
+  color: #ff8040;
```
johncowen, line 57: doh this is a shame isn't it. None of these are ideal but:

1. Did you see theres a `yellow-400`/`yellow-500` in kongponents that is reasonably close to this, not sure maybe its a bit off
2. Is it worth making an `orange-400` just to try and keep getting rid of hex's everywhere

None of the above are ideal for different reasons, but what do you think?

Hmm I just spotted this also:

yellow

After seeing that, I think `yellow-500` might be the best thing instead of having a custom orange, whether with a custom variable or hex? wdyt?

src/assets/styles/_variables.scss
```diff
@@ -1,20 +1,8 @@
-$logo-purple: #290b53;
-
 :root {
   --font-family-sans: -apple-system, BlinkMacSystemFont, 'Segoe UI', 'Noto Sans', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji';
   --font-family-mono: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
 
-  --logo-purple: #{$logo-purple};
-  --logo-purple-light: #{rgba($logo-purple, 0.75)};
-  --logo-purple-lighter: #{rgba($logo-purple, 0.08)};
```
johncowen, line 9: 🎉 Glad we are getting rid of these word based scales

src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -485,10 +484,6 @@ function selectDataPlaneOverview(name: string | null): void {
 </script>
 
 <style lang="scss" scoped>
-.add-dp-button.add-dp-button {
-  background-color: var(--logo-green);
-}
-
```
johncowen, line 491: Glad these overwrites are going 👏

johncowen (commented):
Generally a-ok by me, I had a suggestion for the orange instead of using a hex code, lemme know what you think

src/app/common/PolicyTypeTag.vue
kleinfreund: I’ve seen a case where the `inline-flex` nature of this element was interfering with the layout in that it added a vertical space which I would circumvent by also adding `vertical-align: middle`. I do think however that this was related to the Tailwind base style work. In any case, either style works in either change set.

johncowen (approved):

PR #505, opened 2022-12-14
feat: adds a filter to data plane proxy tables

### LINKED ISSUE #338: Filtering by tag support on dataplane view
### Description

We should add filtering support to the columns for all our tabular output, would make finding things in larger environments easier.

kleinfreund:
## Changes

Adds a filter bar to data plane proxy tables which allows to filter the table for a list of known fields (i.e. `tag` and `namePrefix`).

Resolves #338.

## Notes & further details

* **`QueryParameter`**: Abstracts getting/setting query parameters into a small `QueryParameter` utility object (which replaces the `patchQueryParam` utility) to account for the need to both write _and_ read query parameters.
* **`ShortcutManager`**: In this PR, a `ShortcutManager` utility is added. This matches exactly the utility and behavior of the shortcut management of the `KCodeBlock` component in Kongponents. I’m considering to generalize the filter bar into a reusable component that can be added to Kongponents in which case the utility can be dropped from this project where it won’t be used anymore.

## Testing

Since our mock data API won’t behave realistically when queried with filters, the interaction between filter bar and API can only verified by using the real Kuma API. This can be set-up locally by starting the development server using the following command:

```sh
yarn run dev:real-api
```

... and starting Kuma with, for example, the following:

```sh
make dev/tools && make build && KUMA_DP_SERVER_AUTH_TYPE=none make run/universal/memory
```

## Screenshots

Screenshot from 2023-01-10 10-33-14

Screenshot from 2023-01-10 10-34-02

Screenshot from 2023-01-10 10-34-27


johncowen:
Kinda unrelated but vaguely related, I found that when trying to test this with paging that the paging seems to be broken?

Not sure, but is the paging not working when listing proxies and gateways? If I'm not mistaken it looks like it broke in https://github.com/kumahq/kuma-gui/pull/479 (paging previous to this PR seems fine)

Is paging working for you on this PR?

kleinfreund:
> Kinda unrelated but vaguely related, I found that when trying to test this with paging that the paging seems to be broken?
> 
> Not sure, but is the paging not working when listing proxies and gateways? If I'm not mistaken it looks like it broke in #479 (paging previous to #479 seems fine)
> 
> Is paging working for you on this PR?

@johncowen You’re correct. I fixed this in https://github.com/kumahq/kuma-gui/pull/505/commits/76646bd82666ed6eb402e08b2e5b78071bd34274.

johncowen:
Cool! Cheers for fixing the paging.

I took it for another spin, I think overall it looks great. I noticed that there is some race-y behaviour happening (I think?). 

Sometimes when I refresh the page it shows me the full list of dataplane proxies, even though it has a search that should only show me one, also notice the 2 requests for the listing instead of just the one with the search, which doesn't seem right?

filter-bar

I'm gonna keep trying it in the meantime, so there may or may not be something else that comes up before I take a look at the code.

kleinfreund:
> I noticed that there is some race-y behaviour happening (I think?).
> 
> Sometimes when I refresh the page it shows me the full list of dataplane proxies, even though it has a search that should only show me one, also notice the 2 requests for the listing instead of just the one with the search, which doesn't seem right?

@johncowen This is a consequence of storing the filter query in the URL while for loading we need the set of fields that result from the filter query.

This isn’t trivial to solve.

* 1, the DataPlaneListView component governs the loading of data
* 2, the KFilterBar component governs transforming a query into a set of fields (which become parts of the filter parameters for the data loading process)

At the point when (1) happens, we will only have access to the filter query from the URL. Only when (2) happens do we get access to fields which we need to adjust the API request for loading data. And (2) happens after (1).

A possible solution could be to persist both the filter query and the fields value in the URL so that we can use the fields immediately, but this is rather messy.

Another approach could be to skip the very first loading routine **if** a filter query value is found in the URL. This sounds weird, but is quite simple and covers all cases as far as I can tell.

johncowen:
If I understood correctly this sounds like the solution to me:

> Another approach could be to skip the very first loading routine if a filter query value is found in the URL. This sounds weird, but is quite simple and covers all cases as far as I can tell.

Or put another way, detect if the filter query value is found in the URL and use that for the initial HTTP query/loading routine for the view.

Are you gonna work on this here, I'll hold off on code review until those commits are in if so, in the meantime I'll try it out a little more. Lemme know either way

kleinfreund:
@johncowen I’ve now found that this isn’t as simple in some scenarios after all.

For the service detail view, we only show the data plane proxy table for internal services. However, we only know the type of a service _after_ loading has completed. So, logically, before the component is even initialized, we must load the data. This in itself makes perfect sense.

However, this prevents us from skipping the loading until the filter bar has turned the query into a set of fields because, well, before loading, the component doesn’t even exist.

I’m now trying the other approach of also persisting the filter fields in the URL.

One other conflict I need to solve is an infinite loading loop on service detail views.

I’m moving this PR into draft for the time being.

johncowen:
Gotcha, if it would help to rubber duck with someone lemme know.

kleinfreund:
@johncowen I’ve now implemented the approach of persisting both query and fields in the URL to solve the cyclical conflict explained above. This seems to be working.

I’ve also made sure that in the service detail view, we’re only requesting to load DPPs instead of all data which is unnecessary. This solves a common foot gun where causing to load the service again would toggle the condition for when the DPP table is rendered which interfered with the table display.

johncowen:
I'm still getting 2 identical requests to `/meshes/default/dataplanes+insightstag=kuma.io%2Fservice%3Aredis&size=50&offset=0&gateway=false`.

Oh wait, they aren't identical. On one `gateway=false` is at the end, on the other its in the middle of the query string, maybe that helps?

- `/meshes/default/dataplanes+insightstag=kuma.io%2Fservice%3Aredis&size=50&offset=0&gateway=false`
- `/meshes/default/dataplanes+insightstag=kuma.io%2Fservice%3Aredis&gateway=false&size=50&offset=0`

I'm refreshing on `/mesh/default/data-planes?filterQuery=service%3A%2Bredis&filterFields=%7B%22tag%22%3A%5B%22kuma.io%2Fservice%3Aredis%22%5D%7D&dpp=redis-58f9d884db-ll5f6.kuma-demo`

Is that expected?

johncowen:
I think apart from the 2 request there is a couple of nice to haves that I saw:

1. If I just type `redis` without specifying anything else, should it do a search on name/service/tag/zone
2. Clicking the little arrow things at the end of the menu can repeat things

I wouldn't call either of the above blockers, but point 1 I guess might be easier to sort than 2 and might be worth doing if so.

Gif below to help illustrate:

filter-bar

Lemme know when you are done here code wise and I'll take a look there. I don't think there'll be anything more from me from a user perspective, I think it looks great overall.

kleinfreund:
> I'm still getting 2 identical requests to `/meshes/default/dataplanes+insightstag=kuma.io%2Fservice%3Aredis&size=50&offset=0&gateway=false`.

Right you are. I’ve implemented another guard for avoiding this redundant load. Overall, I’m not happy with the complexity that this added. While it’s nice to avoid an unnecessary API call, I wonder whether that was worth it.

> 1\. If I just type `redis` without specifying anything else, should it do a search on name/service/tag/zone

I don’t want to implement a free-form search for now. We only have a few specific filters for now and being able to do a free-form search could indicate that one can search for data on DPP objects in general which isn’t the case. It’s also unclear which field or set of fields this should be performed on.

> 2\. Clicking the little arrow things at the end of the menu can repeat things

That’s totally expected in my opinion. There is no good reason why applying a suggestion multiple times should behave differently between the first and subsequent applications. That’s keeping this interface simple.

src/types/api.d.ts
```diff
@@ -29,12 +29,14 @@ export interface DataPlaneOverviewParameters extends PaginationParameters {
   /**
    * **Example**: `?gateway=delegated`
    */
-  gateway?: 'builtin' | 'delegated' | boolean
+  gateway?: 'builtin' | 'delegated' | 'true' | 'false' | boolean
```
johncowen, line 32: Non blocker, kinda thinking out loud: I keep seeing little places with gateway = true/false as opposed to gateway = 'true'/'false' i.e. boolean vs string. I wonder if we should remove the boolean so its always a string. Don't think it's that important, but it would help with things like - "should this be 'true' or true? 🤔 ". wdyt?

src/utilities/ShortcutManager.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+function normalizeKeyCode(code: string): string {
+  // Returns relevant modifier keys as the empty string which is going to be filtered out.
+  if (MODIFIER_KEY_CODES.includes(code)) {
+    return ''
+  }
+
+  return code.replace(/^Key/, '').toLowerCase()
+}
```
johncowen, line 123: This looks to be forked from kongponents, is that right? Was wondering how comes we can just use the one in the dependency?

src/utilities/tokenizeFieldFilterQuery.ts
```diff
@@ -0,0 +1,67 @@
...
+  const fieldMatches = prefixedQuery.matchAll(/ ([-\s\w]+):\s*/g)
+  const fieldEntries: [string, string][] = []
+
+  for (const match of Array.from(fieldMatches)) {
+    if (match.index === undefined) {
+      continue
+    }
+
+    // Normalizes field names. Allows field names to be written in dash or space-delimited form (e.g. “name-prefix” or “name prefix” instead of “name”).
+    const fieldName = getCamelCasedField(match[1])
```
johncowen, line 21: Not a blocker at all, and not sure if I understood this right: Just wondering if we are ever going to need to support being able to type `name-prefix`, `namePrefix` or `name prefix` - that's what a lot of this is doing right, supporting all of those? Wouldn't we just use `name` and nothing else?

johncowen (commented):
I had a few questions I left, but I think this is pretty much good to go, gonna scan one more time while you come back, but unless I spot something else this should be good to go!

Nice feature 👏

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -13,9 +13,18 @@
         :next="props.nextUrl !== null"
         :page-offset="props.pageOffset"
         @table-action="($event: any) => selectDataPlaneOverview($event.name)"
-        @load-data="loadData(props.pageOffset)"
+        @load-data="loadData"
       >
         <template #additionalControls>
+          <KFilterBar
+            id="data-plane-proxy-filter"
```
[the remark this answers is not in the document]
kleinfreund: Yes, that’s technically possible, but it’s a minor improvement for which I don’t see a pressing need at the moment.

Technically, this would be a utility that creates a string value that’s unique across the application (e.g. `data-plane-proxy-filter-${ID}` or just `${ID}`).

src/types/api.d.ts
kleinfreund: Right, it’s nicer to only deal with string and number values (or possibly arrays of strings or numbers) on the parameter level. Technically, this wouldn’t be a problem because whatever piece of code turns these parameters into query parameters into a URL would always convert booleans appropriately.

I’m gonna remove them as booleans.

src/utilities/tokenizeFieldFilterQuery.ts
kleinfreund: For multi-word strings, we’d likely have camel-cased fields on the API layer. It’s very easy to convert space-delimited and dash-delimited names to camel-cased variants which is nice to have so you can do what you said: type name prefix or name-prefix and have it work.

Now since we renamed `namePrefix` to `name`, this is technically not relevant right now as we don’t have multi-word parameters, but the functionality comes at hardly any extra cost so I’m leaving this in.

src/utilities/ShortcutManager.ts
johncowen: Ah gotcha sorry 🤦 😆 , cool yeah makes sense, thanks!

src/types/api.d.ts
kleinfreund: Done in https://github.com/kumahq/kuma-gui/pull/505/commits/a091db498913daaae358c4fb2f4730bff4aef68e.

johncowen (approved):
Thanks for the extra info, I also looked at the boolean commit, all looks good to me!

PR #506, opened 2022-12-14
fix(service): not loading correct data when switching service

kleinfreund:
Fixes an issue with not loading the service data when switching between different services while on a service detail view. This happened because our router doesn’t trigger a re-render in these cases.


lahabana:
Confirmed it's visually fixed

johncowen (approved):
:shipit:

PR #507, opened 2022-12-15
chore(docs): Move all the docs links to use env vars

johncowen:
This PR continues on from https://github.com/kumahq/kuma-gui/pull/466, which I pulled back not to include the commit here, so the initial commit messages may read a little strange unless you know the context.

This PR replaces all instances of static docs URLs with static env vars. Importantly, env vars are vars that never change without having to install a new version of Kuma, in which case you'll want to refresh your page.

I went a little further here than in the previous PR but not too far by splitting the `useEnv` utility/composable to use an Env service and then a very simple service container just to get us this far. The service container is likely to be switched out at some point to use something that can auto build our services for us. Important thing for me is how we use this thing, we can change how it works later.

Extra Notes:

1. I changed the default version here to use 1.7.0 instead of 0.0.0 to get less snapshot churn in this PR.
2. Added `KUMA_CHAT_URL`
3. I changed a few `PRODUCT_NAME`s while I was here (just the ones I came across whilst making this change). Interestingly, depending on where you are in the code we seem to use different methods of getting this value. Sometimes we'd use `@/constants` (which is just referring to `import.meta.env.VITE_NAMESPACE`) and sometimes computing the value from `tagline` from the store, which as far as I understand is exactly the same thing. There are still some `PRODUCT_NAME/tagline` things in the codebase and probably a few other things that can now use `useEnv/env` but that can be for future PRs.
4. All docs links now have UTMs
5. This cleaned up some `null`s in the snapshots (just thinking out loud, I wonder if theres an easy way to test for `null`s in the snapshots, might make them a little more useful 🤔 )
6. Added some very minimal tests for `semver` and `useEnv`
7. I probably would like to change `KUMA_NAME` to `KUMA_PRODUCT_NAME` at some point, but not hugely important


src/services/env.ts
```diff
@@ -0,0 +1,73 @@
...
+  KUMA_API_URL: config.apiUrl,
+  KUMA_BASE_PATH: config.baseGuiPath,
+  // TODO(jc): not totally sure we need to use a regex here, maybe just split and join if not
+  KUMA_DOCS_URL: `${import.meta.env.VITE_DOCS_BASE_URL}/${version.patch === '0.0.0' ? 'dev' : version.patch.replace(/\.\d+$/, '.x')}`,
+  // remove the ? for now incase we need to append to an already query
+  // param'ed URL
+  KUMA_UTM_QUERY_PARAMS: import.meta.env.VITE_UTM.substring(1),
+  KUMA_FEEDBACK_URL: import.meta.env.VITE_FEEDBACK_URL,
+  KUMA_CHAT_URL: import.meta.env.VITE_CHAT_URL,
+})
```
kleinfreund, line 41: **Suggetion**: The sole purpose of the `env` utility is to look up static primitive values. This shouldn’t be based on a class whose only functionality is to look up object values by property.

```suggestion
export const env: EnvVars = {
  KUMA_NAME: `${import.meta.env.VITE_NAMESPACE}`,
  KUMA_VERSION: version.pre,
  KUMA_API_URL: config.apiUrl,
  KUMA_BASE_PATH: config.baseGuiPath,
  // TODO(jc): not totally sure we need to use a regex here, maybe just split and join if not
  KUMA_DOCS_URL: `${import.meta.env.VITE_DOCS_BASE_URL}/${version.patch === '0.0.0' ? 'dev' : version.patch.replace(/\.\d+$/, '.x')}`,
  // remove the ? for now incase we need to append to an already query
  // param'ed URL
  KUMA_UTM_QUERY_PARAMS: import.meta.env.VITE_UTM.substring(1),
  KUMA_FEEDBACK_URL: import.meta.env.VITE_FEEDBACK_URL,
  KUMA_CHAT_URL: import.meta.env.VITE_CHAT_URL,
}
```

src/utilities/useEnv.ts
```diff
@@ -0,0 +1,8 @@
+import container from '@/services/container'
```
kleinfreund, line 1: **Question**: The following works correctly with object property lookup:

```ts
import { env } from '@/services/env'

console.log(env.KUMA_DOCS_URL)
```

Why do we need the `useEnv` utility to abstract away this access?

src/services/container.ts
```diff
@@ -0,0 +1,19 @@
+import { env } from '@/services/env'
```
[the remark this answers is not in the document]
johncowen: Its a ServiceContainer a centralized place to access (or inject from The Outside) Services, it'll be swapped out for something a little more featured later (i.e. with auto instantiation, lazy instantiation etc etc)

src/services/env.ts
johncowen: I can see your point here, right now, but moving forwards all Services are probably going to be classes or maybe functions, not bare objects.

src/utilities/useEnv.ts
johncowen: Mainly because it looks like most `vue` usage i.e. `useSomething`, I'd probably prefer something like:

```javascript
import { env } from '@/utilities'

console.log(env('KUMA_DOCS_URL'))
```

but I'm trying to stick to Vue's `useSomething()` pattern i.e. as it will probably/maybe help later on

```javascript
import { useEnv } from '@/utilities/env'
const env = useEnv()
console.log(env('KUMA_DOCS_URL'))
```

src/services/env.ts
kleinfreund: I’m not a fan of abstracting away simple utilities into more complex utilities for the sake of having all utilities follow the same pattern (i.e. a service pattern).

Why do we want this?

None of the properties in this object need any sort of additional handling or routine. It’s merely looking up one primitive value for each use. So while perhaps some utilities/application services are more complex in nature, looking up static data isn’t. That should remain a simple task and easy maintenance even if other services are used with said pattern.

src/services/container.ts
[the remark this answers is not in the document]
kleinfreund: I’d prefer for such a pattern to be justified by a use case that makes use of features as you named them. I’m curious to hear of examples.

In the context of this PR (i.e. streamlining the access of static data for the purpose of avoiding hard-coded docs URLs), this doesn’t seem to be needed.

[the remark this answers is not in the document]
johncowen: One use case is to avoid merge conflicts between repositories.

Examples of auto instantiation would be the service container doing all the dependency lookup for a service when you first instantiate it, lazy instantiation, as sometimes you don't want to instantiate a service until you need it (here we instantiate Env and add it to the container manually even though we don't necessarily need it yet)

> In the context of this PR (i.e. streamlining the access of static data for the purpose of avoiding hard-coded docs URLs), this doesn’t seem to be needed.

A fair point I guess, but you can do a lot of things differently, but depends if you are thinking short term or long term as how you might go about doing something.

Just to point out, you don't even need to know that it's using a ServiceContainer if you are just using it as a consumer (i.e. `const env = useEnv()`, in fact it looks just like any other Vue composable. If we wanted to change our building tooling again or how the app is fed these variables again, the app doesn't need to know about it, we just tweak the service or inject a new one from The Outside.

src/utilities/useEnv.ts
kleinfreund: This pattern in Vue has a specific purpose. It’s commonly used for accessing composables which are indeed Vue specific utilities. However, *all* our utility functions are just plain JavaScript. They haven’t got anything to do with Vue. We don’t _need_ to follow a naming pattern of a framework/library we’re using just because it’s otherwise common.

Specifically for composables, we’re talking about utilities that make use of Vue’s state management functionalities (see https://vuejs.org/guide/reusability/composables.html). Composables typically return an object that contains _reactive_ state which is the used _inside a Vue context_. So the composable pattern is used for framework-specific utilities, but it isn’t used for _all_ utilities.

Therefore, **if** we access the `env` utility via a factory function, we shouldn’t use the `use*` naming convention because it’s for composables which `env` isn’t.

---

All that being said, my question is of a different nature: I’m asking why we even need a factory function for doing an object property lookup? We can already directly look up properties on the `env` object. An abstraction around it would be pointless (i.e. it would do that exact same job).

johncowen: I don't think the fact that a 'utility' doesn't use vue specific state management functionalities (at the moment) matters as to whether we happen to call something `use*` or not - more important to me is that everything looks the same (whether the innards use vue specific things right now or not). If we were to change a utility to use vue specific state management features would that mean we should change the name of it to `use*` everywhere, and vice versa?

Honestly, I don't think it's a big deal that this is called `useEnv` or just `getEnv` or whatever, it's just a name. But I'd rather stick to a pattern whether the innards use something special or not. From the outside/as a consumer I don't need to know.

I think another thing is, which I mentioned last time, I've seen other `use*` utilities that don't contain anything vue specific also, so I don't think this is that off the wall.

src/services/container.ts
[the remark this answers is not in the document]
kleinfreund: What you describe may well apply to complex utilities, but it doesn’t apply to this use case.

> One use case is to avoid merge conflicts between repositories.

Would you mind elaborating how that is accomplished for looking up static data? As far as I can tell, you’d get the same merge conflict whether you make changes to a line referencing `env('PROPERTY')` as you’d get for `env.PROPERTY`. The same would be true for the code where these values are defined.

> A fair point I guess, but you can do a lot of things differently, but depends if you are thinking short term or long term as how you might go about doing something.

That’s why I’m asking. Without a service that has sufficient complexity to warrant the feature richness you mentioned, this adds a maintenance burden in the short term and possibly also in the long term (depending on if and when such a service becomes necessary).

> Just to point out, you don't even need to know that it's using a ServiceContainer if you are just using it as a consumer (i.e. const env = useEnv()), in fact it looks just like any other Vue composable. If we wanted to change our build tooling again (or the build tooling itself changed) or change how the app is fed these variables again, the app doesn't need to know about it, we just tweak the service or inject a new one from The Outside.

That’s also true for accessing a plain object with `env.PROPERTY`. It’s plain JavaScript and not framework or tooling specific. The things that would change on the `env` utility are (1) new properties being added, (2) properties being removed, or (3) properties being renamed. All three scenarios have the exact same impact regardless of whether they’re accessed directly from an object or via a service abstraction.

src/utilities/useEnv.ts
kleinfreund: > I don't think the fact that a 'utility' doesn't use vue specific state management functionalities (at the moment) matters as to whether we happen to call something use* or not

It definitely matters to me.

> more important to me is that everything looks the same (whether the innards use vue specific things right now or not).

I don’t think everything should look the same just because that’s technically possible. Using the `use*` naming convention for all utilities would indicate to people familiar with Vue that they’re composables when they’re not.

> If we were to change a utility to use vue specific state management features would that mean we should change the name of it to use* everywhere, and vice versa?

Absolutely yes. Why not? That’s a good distinction to make. Then you know at a glance what’s a pure, plain JavaScript utility function and what relies on Vue internals.

A plain JavaScript function, you can copy as is into any other JavaScript code base and use. You can’t do that with a composable unless that code base also happens to use Vue 3.

These distinctions matter because it gives users of the code (i.e. us) a clear idea of what’s what. Composables and utility functions aren’t on the same hierarchy of things. Composables _are_ utility functions, but utility functions aren’t necessarily composables.

src/services/env.ts
johncowen: > Why do we want this?

Because we want to use a consistent pattern to inject dependencies from The Outside without having to change code. Granted this is a simple service that could be done differently, but because its simple, it's a good opportunity to get the ball rolling.

src/utilities/useEnv.ts
kleinfreund: > Honestly, I don't think it's a big deal that this is called useEnv or just getEnv or whatever, it's just a name. But I'd rather stick to a pattern whether the innards use something special or not.

I explained [here](https://github.com/kumahq/kuma-gui/pull/507#discussion_r1049649751) that this wasn’t my primary concern.

My primary concern remains that we access the `env` utility via a service abstraction when there is no need for that.

> From the outside/as a consumer I don't need to know.

As a developer, you **have** to know the difference though. You can’t use composables and utilities interchangeably.

> I think another thing is, which I mentioned last time, I've seen other use* utilities that don't contain anything vue specific also, so I don't think this is that off the wall.

That doesn’t align well with the documentation on composables though. They’re described very specifically and narrowly as stateful Vue utilities. https://vuejs.org/guide/reusability/composables.html

johncowen: Oh another thing probably worth mentioning, whilst I'm more than happy to change the name of this here

> Ok so if I change this to getEnv are we good?

Longer term all these accessor functions will hopefully be autogenerated (we're hiding the fact we are using a ServiceContainer), and therefore they'll all use the same naming pattern, if I need to look to see if the function/service is using something Vue specific or not before deciding on a name, we probably won't be able to do the autogeneration.

Just to clarify on the last comment:

> > Ok so if I change this to getEnv are we good?

It sounds like from your last comment, that if I did just change the name here (which I'm happy to do) you still wouldn't be happy with this?

src/app/zones/components/MultizoneInfo.vue
```diff
@@ -45,14 +46,9 @@ export default {
 
   data() {
     return {
+      env,
```
kleinfreund, line 49: **Tip**: In Options API components, you can set non-reactive properties on the top level of the default export (e.g. beneath `name: 'MultizoneInfo',`) and subsequently access them in the template using `$options.env`.

This is better because setting `env` in the `data` function will turn it into a reactive object which we don’t want.

src/services/env.ts
```diff
@@ -0,0 +1,73 @@
...
+  KUMA_VERSION: version.pre,
+  KUMA_API_URL: config.apiUrl,
+  KUMA_BASE_PATH: config.baseGuiPath,
+  // TODO(jc): not totally sure we need to use a regex here, maybe just split and join if not
+  KUMA_DOCS_URL: `${import.meta.env.VITE_DOCS_BASE_URL}/${version.patch === '0.0.0' ? 'dev' : version.patch.replace(/\.\d+$/, '.x')}`,
+  // remove the ? for now incase we need to append to an already query
+  // param'ed URL
+  KUMA_UTM_QUERY_PARAMS: import.meta.env.VITE_UTM.substring(1),
+  KUMA_FEEDBACK_URL: import.meta.env.VITE_FEEDBACK_URL,
+  KUMA_CHAT_URL: import.meta.env.VITE_CHAT_URL,
```
kleinfreund, line 40: **Suggestion**: I think it would be better if none of these values were prefixed with `KUMA_` so that we minimize the chance for merge conflicts between this project and possible forks of it that won’t be called Kuma.

(Though I would use `PRODUCT_NAME` and `GUI_BASE_PATH`)

kleinfreund (commented):
I’ve checked many changed instances in the UI and things are working. I do like that this makes it more consistent to access this kind of data.

I don’t think this is the right time to be introducing a service abstraction as the utility for which it is used doesn’t need its features. It would make more sense to propose such a pattern for a service/utility that _does_ make use of more advanced features.

src/utilities/useEnv.ts
kleinfreund: > It sounds like from your last comment, that if I did just change the name here (which I'm happy to do) you still wouldn't be happy with this?

I’m afraid so. The functionality in this PR needs a typed object with primitive property values. That covers all the needs of the uses we currently have and that we will have in the short-term future (as far as I’m aware of).

Introducing the proposed service pattern and its abstractions, in its current form, adds logic and code that is not needed and made good use of.

I’m happy to discuss the merits of this pattern as part of a use case that needs it, but I strongly believe accessing static data isn’t one of them.

src/app/zones/components/MultizoneInfo.vue
johncowen: Oh nice! Will take look at that now, ty!

src/services/env.ts
johncowen: I would prefer to prefix things really, especially is some of these variables are coming from your environment. Right now I'd prefer to get this merged as there's still work to be done on top of this. But another option would be `KM_` (i.e. KuMa). Again it's just naming really so happy either way as long as its prefixed. If you feel strongly about this now happy to change up to use `KM_` here, otherwise we can revisit later.

kleinfreund: It’s just that every time, you’d change a line with an `env` reference or an adjacent line and a fork has this value renamed, you’ll get a merge conflict. This can be completely avoided by not prefixing them with something product-specific.

kleinfreund: > Right now I'd prefer to get this merged as there's still work to be done on top of this.

In the interest of this, I would prefer to consider this PR with the scope of abstracting away static data access but without introducing the service container pattern. We clearly differ greatly in our opinions on that pattern and so there is need for further exchange.

johncowen: The product is always going to be a using this repository i.e. Kuma, even as a base. Whether the product is called Kuma or something else, naming of properties in code doesn't matter really, main thing is avoiding picking up environment variables that aren't for this project. Most projects that use environment variables use prefixes.

src/app/zones/components/MultizoneInfo.vue
johncowen: I moved both `env` and `productName` (that was also hanging on `data`) to the root of the exported object, not super keen on the magical global-like `$options` thing, but I suppose we'll eventually convert this to composition API at some point, so it will go.

kleinfreund: Yeah, that’s one of the reasons I strongly prefer composition API style components because almost all things are explicit (i.e. no magic `this.$*` things).

src/utilities/useEnv.ts
johncowen: Ok lets put it another way, are you happy that:

1. This uncovered a bunch of tests that should have been failing but weren't (the `null`s).
2. Provides a centralized access point to static environment variables no matter where those actually come from (similar to `useStore()` being for accessing user data, or `useRouter` being for accessing the router)
3. Removes some of the complexity that made it so we were accessing the same value via totally difference approaches in different parts of the codebase.
4. Provides a consistent way to inject alternative data and/or _behaviour_, whether that be within tests or within other different environments.

If we can get past the fact that we are calling a function rather than accessing a property, and naming, we might be able to move forwards a little?

kleinfreund: > This uncovered a bunch of tests that should have been failing but weren't (the nulls).

Yes, but while that is a small issue, those aren’t representing application bugs, but improperly regenerated snapshots.

> Provides a centralized access point to static environment variables no matter where those actually come from

Yes, I am. I pointed this out in https://github.com/kumahq/kuma-gui/pull/507#pullrequestreview-1219287192

> Removes some of the complexity that made it so we were accessing the same value via totally difference approaches in different parts of the codebase.

Yes, very.

> Provides a consistent way to inject alternative data and/or behaviour, whether that be within tests or within other different environments.

No. I will need better examples for this argument. The current complexity of adding the service container pattern is not warranted in my opinion because the alternative (i.e. accessing property values from an object) is strictly simpler, requires less code, and has no disadvantages. I’m not bringing this up because of stylistic preference of how this data is accessed. I’m opposed to adding abstractions to things that don’t need them and accessing static data like we need to really doesn’t.

I’d like to ask you to propose this pattern in context of a justified use case and outside the scope of this PR.

johncowen: > No. I will need better examples for this argument.

Here's one example, we are injecting functionality, not static values, and not using our current (but soon to change) test utilities, i.e. we could just as well do this for a different environment where we don't have test utilities available.

https://github.com/kumahq/kuma-gui/pull/507/files#diff-5cb6e61dca3b13ba5c02a1b586af105e5e92eb5d080e3cea1a4ccc8bf86cd738R14

> The current complexity of adding the service container pattern is not warranted in my opinion

We use a similar approach on the backend to avoid merge conflicts, the GUI currently has a merge conflict problem, so this approach is a common solution to that, I would say it's warranted (either this or another alternative). I think the bigger picture is, we need to inject behaviour aswell as just static variables, unless we are happy with conditonal-y feature flags throughout the codebase, something I thought we were in agreement that we had to avoid.

I get that this could be done simpler, by just having an object and hanging some properties off of it, but as I said up there ^ - this is about more than that. And even then what we have here isn't really that complex, and I've talked about some of the advantages of the overall approach. I'm also a little confused as you were happy to approve this https://github.com/kumahq/kuma-gui/pull/466#pullrequestreview-1208150922 with reservations (but different ones) and we've briefly chatted about this approach offline. Overall if you want to put more discussion into this, I'd be happy to but its probably going to have to wait now 😿 

If you'd rather leave this hanging for today, please let me know and we can continue the discussion another time. I'm easy either way.

src/services/env.ts
johncowen: > In the interest of this, I would prefer to consider this PR with the scope of abstracting away static data access but without introducing the service container pattern. We clearly differ greatly in our opinions on that pattern and so there is need for further exchange.

Ok but can we consider the naming subject resolved? Sounds like that from our discussions offline?

kleinfreund: Yes, sure.

src/app/__snapshots__/App.spec.ts.snap
```diff
@@ -916,32 +916,27 @@ exports[`App.vue fails to renders basic view 1`] = `
                      Join the Kuma community and ask questions: 
                   </p>
                   <ul>
-                    
                     <li>
                       <a
                         href="https://kuma.io/docs/1.7.x/?utm_source=Kuma&utm_medium=Kuma-GUI"
-                        target="_blank"
```
mmorel-35, line 923: Why don’t you keep the blank target. IMHO,  the blank target shall be the default behavior when the href point to a  different domain from the current one.

johncowen: Oh! Good spot! Yeah agree, all these external links should all be target="_blank", let me take a look at that in a sec, thanks! 🙏

johncowen: K these are back in, thanks for that!

src/services/env.ts
lahabana: What about adding computed properties? For example clearly the UTM stuff can be computed.

johncowen: We had a quick chat about this. We can calculate the utm from the `VITE_NAMESPACE`/`KUMA_NAME`, which means we can drop `VITE_UTM` entirely, gonna add a commit on the end here for that.

johncowen:
@kleinfreund @lahabana is there anything else at the moment needed here for another review?

src/services/container.ts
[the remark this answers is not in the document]
kleinfreund: It was decided to move forward with this implementation because it works and because moving between the approaches in question is easy.

(For the record, I still don’t think it’s warranted to introduce this kind of abstraction for such simple utilities.)

kleinfreund:
@johncowen I’ve completed my review for this PR as per https://github.com/kumahq/kuma-gui/pull/507#pullrequestreview-1219287192. Please resolve (or address where you see fit) the remaining open threads as I’m not sure whether you consider them handled. If you want I can do another pass of reviewing the functionality to be working. Let me know.

johncowen:
If you are happy that its working (putting the approach to one side), I just need an approval to merge

kleinfreund (approved):

Issue #517, opened 2022-12-23
When adding a zone failed to show zones


lahabana:
### What happened?

on the GUI while a second zone was added the Zone http://localhost:5681/gui/zones failed with an error like:"cannot read property Some"

I guess the content of the zone response is unexpected.

jakubdyszkiewicz:
> Triage:
> Deploy Global CP. Deploy Zone CP, refresh quickly, wait until Zone CP appears

johncowen:
<img width="745" alt="Screenshot 2023-01-04 at 14 18 19" src="https://user-images.githubusercontent.com/554604/210574872-129e705f-aeb1-4d99-9d1e-566002022b26.png">

lahabana:
The fix that was added was for dataplaneInsight while this issue is in MeshInsight are we sure this is the correct fix?

johncowen:
I think so yeah, the same function is reused for both it's reused here in `ZonesView` https://github.com/kumahq/kuma-gui/blob/6e540fd433274ae6eaa39f5c849ce3e35ac1f4e5/src/app/zones/views/ZonesView.vue#L289 which was where it was erroring.

lahabana:
Can we fix the type and parameter name then?

PR #530, opened 2023-01-05
refactor(insight): rename and refactor status related functions

johncowen:
Whilst fixing https://github.com/kumahq/kuma-gui/issues/517 with https://github.com/kumahq/kuma-gui/pull/523 we realized that the type naming didn't quite make sense:

https://github.com/kumahq/kuma-gui/issues/517#issuecomment-1371985785

So we tweaked the types slightly so that things didn't look as weird.

Whilst I was there we noticed a couple of other things we could do regarding `getItemStatusFromInsight` and `getStatus` which led to being able to compose those two functions to DRY things out a little, simplifying the innards of them slightly and a rename of `getStatus`

@lahabana could you double check the changes in the second to last commit (https://github.com/kumahq/kuma-gui/commit/50fb7d99ef5d98f8eb94300e4502d6e0a14dd882), a tweaked it a little more following our catch up and it would be good for someone with more context than me to make sure its doing the right thing. To me it reads a lot better now

### DIFF
--- a/src/utilities/dataplane.spec.ts
+++ b/src/utilities/dataplane.spec.ts
@@ -27,7 +27,7 @@ describe('utilities/dataplane', () => {
           subscriptions: [
             {
               connectTime: '1',
-            } as DiscoverySubscription,
+            },
           ],
         }],
         expected: 'online',
@@ -39,7 +39,7 @@ describe('utilities/dataplane', () => {
             {
               connectTime: '1',
               disconnectTime: '2',
-            } as DiscoverySubscription,
+            },
           ],
         }],
         expected: 'offline',
@@ -50,11 +50,11 @@ describe('utilities/dataplane', () => {
           subscriptions: [
             {
               connectTime: '1',
-            } as DiscoverySubscription,
+            },
             {
               connectTime: '1',
               disconnectTime: '2',
-            } as DiscoverySubscription,
+            },
           ],
         }],
         expected: 'online',
--- a/src/utilities/dataplane.spec.ts
+++ b/src/utilities/dataplane.spec.ts
@@ -133,7 +133,7 @@ describe('utilities/dataplane', () => {
         }],
         expected: {
           status: 'offline',
-          reason: [''],
+          reason: [`Inbound on port ${1} is not ready (kuma.io/service: ${'service'})`],
         },
       },
       {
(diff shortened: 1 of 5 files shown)


src/utilities/dataplane.ts
```diff
@@ -70,32 +68,36 @@ export function dpTags(dataplane: { networking: DataPlaneNetworking }): LabelVal
...
+// getStatusAndReason takes Dataplane and DataplaneInsight and returns a
+// {status: 'online' | 'offline' | 'partially_degraded', reason: errors[]}
+export function getStatusAndReason(dataplane: { networking: DataPlaneNetworking }, insight: { subscriptions: DiscoverySubscription[] } | undefined = { subscriptions: [] }): { status: StatusKeyword, reason: string[] } {
+  const inbound = dataplane.networking.inbound ?? []
+  const errors = inbound
+    .filter(item => item.health && !item.health.ready)
+    .map(item => `Inbound on port ${item.port} is not ready (kuma.io/service: ${item.tags['kuma.io/service']})`)
+
+  let status: StatusKeyword
+  switch (true) {
```
lahabana, line 87: I dataplane.networking.inbound is empty or nil we should have status = "online"

lahabana: I think adding a first:

```ts
case inbound.length == 0:
  status = 'online'
  break
```

is likely the nicest thing you can do here

johncowen: Thanks @lahabana, I added this in, also added a quick bunch of unit-y tests which more or less cover what I'd understood

src/utilities/dataplane.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+              port: 1,
+              tags: {
+                'kuma.io/service': 'service',
+              },
+            }],
+          },
+        }],
+        expected: {
+          status: 'offline',
+          reason: [''],
```
lahabana, line 136: nit: Would it be more readable if reason was `reasonCount` or something like this? It's kind of weird to see this here and then see it works because we assert on length of the array

lahabana (approved):

src/utilities/dataplane.spec.ts
johncowen: Yeah maybe, I started off trying to keep it the same as shape of the return from the function, and then realized these were strings that needed more mock data adding to get the correct string, then I kinda figured "wait a second I don't want to test the string content of the `reason`s, just that there are the right amount", and left it like this.

Me personally I'd keep `expected` to be the same shape as the return from the function, but I don't mind changing it to `reasonLength` if you would prefer? Let me know what you think after reading this, happy to change if you are still keen.

lahabana: I don't mind either way really. Just pointing out it's reading these tests put me off-guard a little

johncowen:
```suggestion
          reason: ['fake error message to fill the length of `reason`'],
```
Would something like that ^ have helped? Means we keep the shape and it might be clearer when reading it?

lahabana: Though it feels hacky

johncowen: Oh I just realized I've left the the correct values in from when I decided not to do it that way 😆 I'm half way there, one sec

lahabana: LG

Issue #525, opened 2023-01-04
Surface policies/resources in Mesh overview instead of within a tab


lahabana:
### Description

This page: http://localhost:5681/gui/mesh/default

- Remove the overview and resources tabs to have a single page which shows all the content
- Add links to policies page from the policies
- Add donut chart with policy count (and link to policies page)
- Experiment with moving "add a virtual mesh" etc down or up the page (maybe a way to delete that).

PR #531, opened 2023-01-05
feat(policies): Add a new column and donut chart for policies, move starter grid down

johncowen:
In https://github.com/kumahq/kuma-gui/pull/503 we removed all the Policies from the side nav bar as it was getting a little busy (and that will only grow), but we noticed that it was a little weird not being able to see which PolicyTypes had non-zero counts in them at a glance.

We also noticed that the information we were after was in the [Resources] tab on the Mesh Overview page.

This PR removes the [Overview] and [Resources] tabs and combines those views, therefore when you click into the Mesh/Mesh Overview page you can immediately see which policies are populated for that Mesh. Note we don't show zero count PolicyTypes. This keeps the sidebar uncluttered but still gives you the ~Droids~information you are looking for.

Additionally we added a new DonutChart showing amount of PolicyTypes for the mesh, and lastly we moved the 'starter blocks' below all the information/data.

Actually, in writing this, I'm thinking we can maybe split the DonutChart into segments of policy types maybe? That can be a followup PR if so? WDYT?

Screenshot 2023-01-04 at 17 10 36

Fixes https://github.com/kumahq/kuma-gui/issues/525


lahabana:
I see 27 policies in total is it still the same bug? If yes can you fix the mocks (I've fixed the bug earlier today)

johncowen:
Oh yeah - this one you are referring to right? : https://github.com/kumahq/kuma/pull/5577 . I defo should fix up the mock data also, will sort that now, thanks for the nudge!

src/app/mesh-overview/components/MeshCharts.vue
```diff
@@ -73,7 +82,7 @@ function loadData() {
 }
 
 .chart-1\/4 {
```
kleinfreund, line 84: **Suggestion**: Ought to rename this according to the change in ratio.

```suggestion
.chart-1\/5 {
```

kleinfreund (approved):

src/app/mesh-overview/components/MeshCharts.vue
johncowen: 😆 ah defo yeah, thanks for the spot

I made a slightly different change that takes the value out of the class name, see last commit

PR #534, opened 2023-01-09
refactor: streamlines selected table row persisting

kleinfreund:
Reworks slightly how selected rows are persisted in DPP and service tables. The associated props and query parameters were given more accurate names to avoid conflicts and give a clearer indication of what they’re representing.

Also fixes one or two instances of persisting selected rows not actually working.

### DIFF
--- a/src/app/data-planes/components/DataPlaneList.vue
+++ b/src/app/data-planes/components/DataPlaneList.vue
@@ -179,7 +179,7 @@ const props = defineProps({
     default: 0,
   },
 
-  name: {
+  selectedDppName: {
     type: String,
     required: false,
     default: null,
@@ -284,7 +284,7 @@ function onCreateClick() {
 async function initializeData(): Promise<void> {
   try {
     if (Array.isArray(props.dataPlaneOverviews) && props.dataPlaneOverviews.length > 0) {
-      selectDataPlaneOverview(props.name ?? props.dataPlaneOverviews[0].name)
+      selectDataPlaneOverview(props.selectedDppName ?? props.dataPlaneOverviews[0].name)
       tableData.value.data = await Promise.all(props.dataPlaneOverviews.map((item) => parseData(item)))
     } else {
       selectDataPlaneOverview(null)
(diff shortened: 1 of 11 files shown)


src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -298,10 +298,10 @@ async function initializeData(): Promise<void> {
 function selectDataPlaneOverview(name: string | null): void {
   if (name && props.dataPlaneOverviews.length > 0) {
     dataPlaneOverview.value = props.dataPlaneOverviews.find((data) => data.name === name) ?? props.dataPlaneOverviews[0]
-    patchQueryParam('name', dataPlaneOverview.value.name)
+    patchQueryParam('dpp', dataPlaneOverview.value.name)
```
johncowen, line 301: Kinda off topic question: Do we still need `patchQueryParam` now we've moved to history API routing?

src/router/router.ts
```diff
@@ -114,7 +114,7 @@ export function createRouter(baseGuiPath: string = '/', policyDefinitions: Polic
                 title: 'Gateways',
               },
               props: (route) => ({
-                name: route.query.name,
+                selectedDppName: route.query.dpp,
```
johncowen, line 117: I don't think I mind either way, but noticed that this could also be `?gateway` instead of `?dpp=` seeing as its the gateway listing, not a blocker at all completely up to you

johncowen (commented):
I had a couple of non-important questions, but also - how much effort would it be to make the  zones/policies list views work the same? Is it something that we could do here seeing as this PR is:

> Also fixes one or two instances of persisting selected rows not actually working.

Or would that be best for follow up PRs?

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: Yes and the reason for this is that using the router’s push method will cause a navigation when we don’t want that. We _only_ want to change the URL. We don’t want a navigation to happen.

src/router/router.ts
kleinfreund: I’ve changed this to conditionally set the right parameter.

kleinfreund:
@johncowen I’ve added support for persisting selected rows in the URL for both policy and zone-related views now.

johncowen (approved):
:shipit:

Issue #526, opened 2023-01-04
Policies page should be on the route `/policies`


lahabana:
### What happened?

At the moment /gui/mesh/default/circuit-breakers this should be: /gui/mesh/default/policies/circuit-breakers

PR #543, opened 2023-01-10
feat(policy): Add dynamic policy route instead of generating static ones

johncowen:
Previously we were looping through all possible policyTypes and generating static routes for each of them.

This PR changes that to use a dynamic route for all of them instead `/policies/:policyType` instead. All policies now live under `/policies/*` instead of just `/*` meaning clashes are less likely.

Setting the title is a little hacky here, and setting the page title does seems to be possible via a `store.dispatch` to the data layer. But when I looked at setting the breadcrumb, it didn't seem possible with the current code without setting the route.meta.title within the route, ideally breadcrumbs I'd vote for breadcrumbs (and titles) being set within the *View.vue, but I didn't want to go about changing too much just for this - either via that alternative or a different one.

I think I removed all the code we no longer need, but please shout if there is something I've missed as I'm not totally sure.

Fixes https://github.com/kumahq/kuma-gui/issues/526

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -42,11 +42,12 @@ async function initializeVue() {
   await Promise.all([
     // Fetches basic resources before setting up the router and mounting the application. This is mainly needed to properly redirect users to the onboarding flow in the appropriate scenarios.
     store.dispatch('bootstrap'),
-    // Loads available policies in order to populate the necessary routes.
+    // Loads available policies in order to populate the necessary route information.
+    // i.e. pluralDisplayName and amounts of policyTypes
     store.dispatch('fetchPolicies'),
   ])
 
-  const router = await createRouter(pathConfig.baseGuiPath, store.state.policies)
+  const router = await createRouter(pathConfig.baseGuiPath)
 
   app.use(router)
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -40,10 +40,12 @@ async function initializeVue() {
   app.use(store, storeKey)
 
   await Promise.all([
-    // Fetches basic resources before setting up the router and mounting the application. This is mainly needed to properly redirect users to the onboarding flow in the appropriate scenarios.
+    // Fetches basic resources before setting up the router and mounting the
+    // application. This is mainly needed to properly redirect users to the
+    // onboarding flow in the appropriate scenarios.
     store.dispatch('bootstrap'),
-    // Loads available policies in order to populate the necessary route information.
-    // i.e. pluralDisplayName and amounts of policyTypes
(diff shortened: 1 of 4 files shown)


kleinfreund:
This is how you can set the document title from within the policy view component:

```ts
watch(() => policy.value, function (newPolicy, oldPolicy) {
  if (newPolicy.name !== oldPolicy?.name) {
    store.dispatch('updatePageTitle', newPolicy.name)
  }
}, { immediate: true })
```

src/main.ts
```diff
@@ -42,11 +42,12 @@ async function initializeVue() {
   await Promise.all([
     // Fetches basic resources before setting up the router and mounting the application. This is mainly needed to properly redirect users to the onboarding flow in the appropriate scenarios.
     store.dispatch('bootstrap'),
-    // Loads available policies in order to populate the necessary routes.
+    // Loads available policies in order to populate the necessary route information.
+    // i.e. pluralDisplayName and amounts of policyTypes
```
kleinfreund, line 46: **Comment**: `fetchPolicies` _only_ loads policy type definitions, not the amounts. That happens as part of the bootstrap action I believe.

johncowen:
I noticed that there were a bunch of router links to fix up in DataplanePolicies (see last couple of commits), which led to me realising that we don't have an actual 'detail' page for a policy, we just reuse the listing page with a list with one policy in it. I found this all a little confusing from a user perspective (it looked to me like I only had one policy). The code seems to use a `ns` query param which again is a little confusing seeing as `ns` (namespace?) is being used to show a single policy.

I tried not to increase the scope of what we are doing here, so I fixed things up so everything worked as before - but I feel like we need a policy detail page to replace what the `DataplaneProxies > Expand Policy Types > Click single Policy Origin > View single Policy` flow is doing.

> This is how you can set the document title from within the policy view component:

Yeah gotcha, I tried this out, but then saw there was no way to set the breadcrumb like this so backed out of that:

> Setting the title is a little hacky here, and setting the page title does seems to be possible via a store.dispatch to the data layer. But when I looked at setting the breadcrumb, it didn't seem possible

I found I could just change the meta from `props()` at which point I have the string title, it's a shame that `meta` can't use `meta()` so you can calculate meta on the fly, feels a bit wrong to set meta from `props()` but 🤷 . I'd like to change how titles and breadcrumbs are set/displayed/announced at some point in the future.

src/main.ts
johncowen: I tweaked the comment a little to be more exact.

kleinfreund (approved):

PR #546, opened 2023-01-13
refactor: streamlines naming of policy type

kleinfreund:
Renames state and other variables related to policy types to distinguish the concepts of policy types (e.g. CircuitBreaker) from policies (e.g. one specific instance of a policy type).

Removes all uses of the policy type fields `singularDisplayName` and `pluralDisplayName` to be consistent with how we refer to these entities in the docs.

### DIFF
--- a/src/api/kumaApi.ts
+++ b/src/api/kumaApi.ts
@@ -16,8 +16,8 @@ import {
   Mesh,
   MeshGatewayDataplane,
   MeshInsight,
-  PolicyDefinition,
   PolicyEntity,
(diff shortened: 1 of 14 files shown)


lahabana:
That's @michaelbeaumont 's favourite PR

johncowen (approved):
Very much like that we have `Policies` and `PolicyTypes` as 2 different things, looks like it half became `PolicyDefinition` at some point, so I like that this is all consistently two separate things now, and if it matches up with our docs then even better.

PR #548, opened 2023-01-16
chore(deps): requires Node.js v18.13

kleinfreund:
Requires Node.js v18.13 throughout the project to address a breaking change in v18.13 (see https://github.com/nodejs/node/issues/46123).


johncowen (approved):
🎱

PR #551, opened 2023-01-16
feat: refine policy tab

### LINKED ISSUE #542: policies view improvement
### What happened?

Some improvements to policy views are needed:

- clearer direction (from and to)
- Should “origin” be “policy”? (or “origin policy”?)
- “Origins” should be styled as a list
- Discuss with backend what needs to be done with the "on" column.

kleinfreund:
## Changes

* chore: renames origins to origin policies
* chore: styles origin policies as a list
* chore: updates policy type mock data

Resolves #542.


src/app/data-planes/components/DataplanePolicies.vue
```diff
@@ -250,80 +252,103 @@ function getPolicyEntryConnections(policy: MatchedPolicyType, policyType: Policy
...
   let sourceTags: LabelValue[] | null
   let destinationTags: LabelValue[] | null
 
+  // Determines source tags.
   if (type === 'clientSubset') {
+    // For client subsets, the source is represented by `subset` (i.e. tags)
     // Sets the wildcard service tag for client subsets without subset values.
     const sourceSubset = subsetEntries.length > 0 ? subsetEntries : [['kuma.io/service', '*']]
-    // For client subsets, the source is represented by `subset` (i.e. tags); for destination subsets and single items, the source is the DPP.
     sourceTags = sourceSubset.length > 0 ? sourceSubset.map(([label, value]) => ({ label, value })) : null
```
johncowen, line 316: I know this is from a previous PR but maybe a "whilst you are here", plus it would help with an overall opinion here that this function is a little hard to look at, and there are a few things that I'm looking at and wondering about things. Also it looks like this function is the main chunk of the PR (the rest looks like mostly mocks and variable renaming)

Anyway:

Isn't `sourceSubset` here always guaranteed to be `> 0` length?

i.e. If `subsetEntries` is `> 0` length, then we use `subsetEntries` for `sourceSubset` and hence `sourceSubset` is `> 0` length, and if `subsetEntries` isn't ` > 0` length then we set it to something that is `> 0` length and hence `sourceSubset` is `> 0` length.

Therefore `sourceSubset` is always `> 0` length?

I dunno maybe I read it wrong?

src/app/data-planes/components/DataplanePolicies.vue
```diff
@@ -250,80 +252,103 @@ function getPolicyEntryConnections(policy: MatchedPolicyType, policyType: Policy
...
   } else {
+    // For destination subsets and single items, the source is the DPP for which we don’t want to show anything.
     sourceTags = null
```
johncowen, line 319: If `sourceTags` was:

```javascript
const sourceTags: LabelValue[] = []
```
instead of 
```javascript
let sourceTags: LabelValue[] | null
```

you could remove this fork/line, then in (and anywhere else): 

https://github.com/kumahq/kuma-gui/pull/551/commits/ed32c8c85549326024ccacfa7d2f6622081274f9#diff-49d71fc15211d9665640235e79525ac0136f3ed7584cd8d56c133f14f73f9fb9R28-R55

```javascript
Array.isArray(rowValue.sourceTags) && rowValue.sourceTags.length > 0
```

You could remove all the isArray checks and just stick with the non-zero length checks. The typing allows `null`, but I also noticed you added that type here too (`RuleEntryConnection`), is there anything preventing these from being just `LabelValue[]` instead of `LabelValue[] | null`? Maybe I missed something? Is there a point that we do something different when this is `null` rather than zero length?

src/app/data-planes/components/DataplanePolicies.vue
```diff
@@ -250,80 +252,103 @@ function getPolicyEntryConnections(policy: MatchedPolicyType, policyType: Policy
...
+    } else {
+      // For destination subsets with empty or absent `subset` field and empty or absent `service`, we can set the wildcard service tag to indicate that the traffic goes to all services.
+      destinationTags = [
+        {
+          label: 'kuma.io/service',
+          value: '*',
+        },
+      ]
+    }
+  } else if (type === 'clientSubset' && typeof service === 'string' && service !== '') {
```
johncowen, line 344: Is there any reason not to include this in the `type === 'clientSubset'` fork higher up (starting at line 312)?

johncowen (commented):
I mostly looked at `getRuleEntryConnections` (looks like the rest of this PR is mostly just a variable rename) and left a few questions. I think overall this function looks like it might benefit from a bit of reworking, which would help when looking at it (either now or in the future).

Most of the queries are small things which maybe on there own I wouldn't particularly mind about, but there's quite a few things in that one function where I'm like 🤔 and the more I looked at it the more I 🤔 'ed.

Have a look and see what you think.

Functionality wise, looks great!

src/app/data-planes/components/__snapshots__/DataplanePolicies.spec.ts.snap
```diff
@@ -356,15 +356,17 @@ exports[`DataplanePolicies.vue renders snapshot 1`] = `
...
-                        <div
+                        <ul
                           class="origin-list"
                         >
                           
-                          <a>
-                            fi1.kuma-system
-                          </a>
+                          <li>
+                            <a>
```
johncowen, line 364: Sorry I forgot to ask. I'm guessing this (the anchor with no href) is a consequence of us removing router link in the tests right (I vaguely remember thats what we are doing)?

Would be nice if these failed when the routes didn't exist. Would it be a lot of effort to make the snapshots include the `href`'s? (btw I might have misunderstood whats happening here, if so please ignore!)

src/app/data-planes/components/DataplanePolicies.vue
kleinfreund: You’re correct. In this branch, we can simplify the logic slightly. I’ve adopted the same explicit if statement branching as I’ve used below for setting `destinationTags`.

src/app/data-planes/components/DataplanePolicies.vue
kleinfreund: I want to explicitly distinguish between "sourceTags are applicable" and "sourceTags are not applicable" by using `null` (which indicates the latter). However, it seems to be the case that the lists of tags can’t actually be empty if they’re applicable for a rule (which can be attributed to the notion that we set the wildcard service tag). Therefore, I can indeed change the semantic of those variables to have `[]` indicate "not applicable".

Technically, I think this is worse to read _semantically_ because an empty list by itself doesn’t mean "this isn’t applicable", but it makes for a slightly improved readability overall of code here.

src/app/data-planes/components/DataplanePolicies.vue
kleinfreund: Yes, I want to separate setting source tags from setting destination tags because that makes it easier to follow the individual conditions.

src/app/data-planes/components/__snapshots__/DataplanePolicies.spec.ts.snap
kleinfreund: It turns out the test also works without the router stub so I’m removing it. This leads to the snapshot being generated with correct links.

kleinfreund:
> I think overall this function looks like it might benefit from a bit of reworking, which would help when looking at it (either now or in the future).

@johncowen I’m very happy to hear suggestions about how it should be reworked. The data transforming functions in this file do exactly what they need to do and don’t do anything extra or unnecessary. I can separate their logic into smaller functions, but the complexity of the task at hand wouldn’t get smaller, it would only be more abstract. I really don’t know how to make this easier to look at while not making it hopelessly obscure.

Please suggest an approach of making this be less overwhelming or explain what about it makes it hard to look at.

> Most of the queries are small things which maybe on there own I wouldn't particularly mind about, but there's quite a few things in that one function where I'm like thinking and the more I looked at it the more I thinking 'ed.

Please be more specific. I really need to know what "where I'm like thinking and the more I looked at it the more I thinking 'ed." (note: GitHub somehow doesn’t let me copy the emojis) actually means in order to understand the problem.

johncowen:
> Please suggest an approach of making this be less overwhelming or explain what about it makes it hard to look at.

I left the suggestions as inline comments, from you comments on those it sounds like you are already working on it?

> Please be more specific. I really need to know what "where I'm like thinking and the more I looked at it the more I thinking 'ed." (note: GitHub somehow doesn’t let me copy the emojis) actually means in order to understand the problem.

---

> but there's quite a few things in that one function where I'm like 🤔 and the more I looked at it the more I 🤔 'ed.

> but there's quite a few things in that one function where I'm like "I'm thinking about this and it looks strange" and the more I looked at it the more I "thought it looked strange".

The specific-ness is partly in the inline comments I made, but it's also partly just an overall feeling that things didn't make sense to me 🤷 , I'll leave it to you, but it sounds like you agreed with at least some of my feedback?

kleinfreund:
> The specific-ness is partly in the inline comments I made, but it's also partly just an overall feeling that things didn't make sense to me  .

I’m really not sure what to do about this. To the specific task of transforming this data from one format into another, I’ve tried to do my best to make it minimal (in the sense of logic, not space) and easy to follow.

> but it sounds like you agreed with at least some of my feedback?

Yes, I do.

johncowen (approved):
Ok great, thanks for taking another look at it.

PR #558, opened 2023-01-17
feat: always uses full width layout

### LINKED ISSUE #550: Content is sometimes fixed width
### What happened?

We decided to move away from fixed-width cotent. We should do this everywhere

kleinfreund:
Switches the UI to always use full-width layouts instead of doing so only for some views.

Resolves #550.


johncowen (approved):
Looks way better 👏

PR #559, opened 2023-01-17
feat: persist selected tab in URL

### LINKED ISSUE #544: Persist opened tab in URL where applicable
### Description

Update the tab widget to optionally allow persisting the open tab item in the URL.

kleinfreund:
Persists a selected tab of the `TabsWidget` component in the URL.

Resolves #544.

### DIFF
--- a/src/app/common/TabsWidget.vue
+++ b/src/app/common/TabsWidget.vue
@@ -20,9 +20,9 @@
 
       <div class="tab__content-container">
         <KTabs
-          v-model="tabState"
+          v-model="activeTabHash"
           :tabs="tabs"
-          @changed="hash => switchTab(hash)"
+          @changed="switchTab"
         >
           <template
             v-for="tab in tabsSlots"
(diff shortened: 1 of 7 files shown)


johncowen:
I hit this, which I think I can reliably reproduce:

Screenshot 2023-01-17 at 13 42 05

On another note, have you considered doing this via nested routing? If it was me, I'd much prefer that over using queryParams

kleinfreund:
> I hit this, which I think I can reliably reproduce:

You’re running into an issue unrelated to this PR. I can reproduce this on the `master` branch as well.

Steps to reproduce:

1. Navigate to a gateway (e.g. http://localhost:8080/mesh/default/gateways/cluster-1.gateway-01)
2. Navigate to the policies tab.

kleinfreund:
> On another note, have you considered doing this via nested routing? If it was me, I'd much prefer that over using queryParams

No, I haven’t. In my opinion, the URL **path** isn’t a better place to store this information. Our URL routes are modeled mostly after the resources represented in any given view. So, for example, a data plane proxy detail view shows details about a single DPP and the URL path corresponds to that: `/mesh/:mesh/data-planes/:dataPlane`. However, clicking around inside the detail view and opening different tabs, in my opinion, doesn’t constitute _navigating_ to a different view; thus, the URL path shouldn’t change. For DPP detail views one could argue that the individual tab content could be considered its individual view and to some degree it’s arbitrary which model is followed.

That being said, implementing this as individual routes is also quite a bit more work while being functionally identical so I prefer this approach.

johncowen:
Oh missed one! 😆 There's another I just spotted also, will PR it in a sec. Lemme take another look at this

johncowen (approved):
LGTM!

On the query params vs nested routes thing. Seems like we differ on opinion here. I'd say this is exactly one of the things nested routes are there/good for.

For example, there seems to a good bit of weirdness here (going from `#`'s to `?`s) plus the query param shim thing, and maybe more (?). All of which I guess could be removed by nesting the routes, it also has the benefit that all the tabs could be controlled (added/removed) from The Outside via the routing configuration.

Top level: in my mind different URL segments show you different data (whether the new data is nested or not), queryParams then control the view of that same data i.e. sorting

Anyway, agree not the time to go into detail now if you are happy with this, just wanted to leave alternative opinion, as it's something I'd like us to talk about moving to.

Issue #521, opened 2023-01-03
Modularize the codebase to make things more easily composable


lahabana:
### Description

The current codebase is not very configurable. We have some use cases (entreprise fork or test mocks) where we should be able to replace entire components with another and add an extra functionality (for example TLS reports in the sidebar).

We should write a MADR to explain how we'll proceed and then refactor the codebase to make this work.

lahabana:
https://github.com/Kong/kong-mesh-gui/pull/135

PR #560, opened 2023-01-17
chore(di): Move service container to use brandi

johncowen:
In https://github.com/kumahq/kuma-gui/pull/507 we added an `env` service/function to help us get hold of 'environment variables' from lots of different places (build-time CLI, run-time DOM etc). We also introduced a very rudimentary service container to make the behaviour of this service/function centralised and more flexible.

This PR goes one more step (towards https://github.com/kumahq/kuma-gui/issues/521) and introduces a more featured service container based on `brandi`.

Further notes:

1. As noted in #507 I wanted to move `KUMA_NAME` to `KUMA_PRODUCT_NAME`, so I did this here.
2. 'Services' live in `@/services/service-name/`. They are folders. This means we can add different implementations, tests, and docs inside this folder and the service itself is like a 'little module'. I almost made an `index.ts` file in this folder to get 'default setups' I'm 99% sure I want to do something like that and only ~90% sure of exactly what I want to go in the `index.ts` file. 
3. Docs for the service live in the same service folder - I added one here, which is relatively sparse, but as these services are built up I would imagine these READMEs will be super handy as we add more and more of them.
4. I included a `CookiedEnv`, an alternative implementation of `Env`, which will let us change environment variables when we can't and/or it's hassle-y to rebuild the application, i.e. on our Netlify preview sites. This also gave me the opportunity to test out the whole use case/flow ( `We have some use cases (enterprise fork or test mocks) where we should be able to replace entire components with another and add an extra functionality`  ) in the one single repository, so it was nice to have something to check the use case with here.
5. I added an alternative vite config using a "development container", but I didn't as yet add this into our default package scripts, we can discuss if/when we want to do that, but for the purposes of this PR you can use `npx vite -c vite.config.development.js`

P.S. Oh I forget to mention

6. I added some very thin helper functions over `brandi`


vite.config.js
```diff
@@ -53,4 +53,5 @@ export default defineConfig(function ({ mode }) {
       },
     },
   }
-})
+}
+export default defineConfig(({ mode }) => config({ mode }))
```
kleinfreund, line 57: **Issue**: This prevents all type checking that is provided through `defineConfig` to happen when defining `config`.

I see why you’re doing it like that, but please somehow restore the type checking for the `config` object. Ideally, the development config should have the same type checking capabilities. In the development config, the parameters of the config function (e.g. `mode`) are now completely untyped, too.

vite.config.development.js
```diff
@@ -0,0 +1,17 @@
...
+
+// https://vitejs.dev/config/
+export const config = ({ mode }) => {
+  const config = prodConfig({ mode })
+  //
+  config.resolve.alias = {
+    ...config.resolve.alias,
+    ...{
+      '/src/services/index.ts': '/src/services/development.ts',
+    },
```
kleinfreund, line 12: **Suggestion**: No need for destructuring an object literal when you can just use its fields:

```suggestion
    '/src/services/index.ts': '/src/services/development.ts',
```

src/services/env/Env.ts
```diff
@@ -1,48 +1,55 @@
...
+    const env = (str: keyof EnvInternal, d: string = '') => this.var(str, this.var(str, _env?.[str] ?? d))
+
+    const config = readPathConfigFromDom()
+    const version = semver(env('KUMA_VERSION', config.version))
+
+    _env = {
+      ..._env,
+      ...{
+        KUMA_UTM_QUERY_PARAMS: `utm_source=${env('KUMA_PRODUCT_NAME')}&utm_medium=${env('KUMA_PRODUCT_NAME')}-GUI`,
+      },
```
kleinfreund, line 33: **Suggestion**: Please replace the unnecessary destructuring with setting properties directly.

vite.config.js
johncowen: I moved the `vite.config` files to typescript, no idea why they weren't typescript already, guessing having these as typescript files is fine and all seems fine to me. Little note here: later on we could completely remove `mode` and just use the separate files, but we can sort that once we have all this finished up.

vite.config.development.js
johncowen: I used `mergeConfig` for this instead.

johncowen:
Those bits are done, lemme know if there's anything else.

src/app/common/UpgradeCheck.spec.ts
```diff
@@ -1,29 +1,36 @@
...
 function renderComponent() {
   return mount(UpgradeCheck)
 }
 
 describe('UpgradeCheck.vue', () => {
   test('renders snapshot', async () => {
-    const env = container.get('env')
-    container.set('env', {
-      var: (key: string) => {
+    container.capture?.()
```
kleinfreund, line 14: **Question**: What is this doing (same for the restore call)?

src/app/common/UpgradeCheck.spec.ts
```diff
@@ -1,29 +1,36 @@
...
           return '1.2.0'
         }
-        return env.var(key)
-      },
-    })
+        return super.var(...rest)
+      }
+    }
+    TOKENS.Env = service(TestEnv, { description: 'Env' })
+    injected(TestEnv, TOKENS.EnvVars)
```
kleinfreund, line 26: **Comment**: This is very complicated now.

Previously this was achieved using:

```js
store.state.config.version = '1.2.0'
```

From what I can tell, we seem to be replacing the _entire_ `Env` utility with a custom one specific to this test. Is that correct? I had really hoped for this to be simpler.

One thing that confuses me is that the symbols `service` and `injected` are very unclear in their naming. What is `service` doing? Constructing a service? Returning an existing service? Same for `injected`. The names of those functions do not indicate what they’re doing.

I’d appreciate a slim piece of documentation in the project that explains how one makes use of ... this stuff. And what things are called (e.g. what’s a service? what’s a token? etc.).

src/services/env/Env.ts
```diff
@@ -1,48 +1,51 @@
...
+      KUMA_VERSION: version.pre,
+      KUMA_API_URL: env('KUMA_API_URL') ?? config.apiUrl,
+      KUMA_BASE_PATH: env('KUMA_BASE_PATH') ?? config.baseGuiPath,
+    }
   }
 
-  var(key: keyof EnvVars) {
-    return this.env[key]
+  var(key: keyof EnvVars, d: string = '') {
+    return this.env?.[key] ?? d
```
kleinfreund, line 45: **Question**: I don’t think `this.env` can actually be `undefined` after construction. Can you please check this?

johncowen: On line 24:

```javascript
const env = (str: keyof EnvInternal, d: string = '') => this.var(str, this.var(str, _env?.[str] ?? d))
```
If you call `env` before you set `this.env` (on line 33) - which we do, then `this.env` is `undefined`. More to the point tho, I think there's one too many `this.var`s on line 24 😆

kleinfreund: I see, but does it even make sense to call `this.var` there? It can’t ever find a value because `this.env` will be `undefined` and so it would fall back to the default value.

johncowen: Yes, because if you override `var` so that that it checks elsewhere first before checking `this.env` (i.e. Cookies) then it will find a value.

kleinfreund: Hm. I see. It’s really quite intransparent to account for that in a utility that shouldn’t really have the outside knowledge.

In any case, the declaration of `env` in line 21 needs to be typed as `env: EnvVars | undefined` to account for this.

src/app/common/UpgradeCheck.spec.ts
johncowen: This is just overriding a method on a class in order to provide a mock, I personally don't think that's a very complicated thing, and probably pretty common. Is this a blocker, or just an opinion?

It's almost:

`env.var = () => '1.2.0'`

> From what I can tell, we seem to be replacing the entire Env utility with a custom one specific to this test. Is that correct?

Yes, that's correct.

`injected` comes from brandi which we've said we are going to use as a service container. Docs for `injected` specifically are here https://brandi.js.org/reference/pointers-and-registrators#injectedtarget-tokens.

`service` is a thin helper function the following (`service` is easier to write than the following, especially when the vast majority of our services will be singletons):

```javascript
const TOKENS = {
  apiService: token<ApiService>('ApiService') 
};
container.bind(TOKENS.apiService).toInstance(ApiService).inSingletonScope();
```

https://brandi.js.org/reference/binding-scopes#insingletonscope

A `token` is an identifier for a `service` that carries its type with it instead of just using strings which would mean you lose the type.

https://brandi.js.org/reference/pointers-and-registrators#tokentdescription

All the docs for all this stuff is at https://brandi.js.org/, and if we want to keep the thin helper functions, then I'd be happy to add some brief docs for those if it would help.

> And what things are called (e.g. what’s a service? what’s a token? etc.).

Take a look at brandi's docs, a lot of this was covered in the MADR along with links to these things.

src/app/common/UpgradeCheck.spec.ts
johncowen: `capture` saves the state of the container so you can change things so you can then call `restore` to put it back to how it was, perfect for before/after hooks.

https://brandi.js.org/reference/container#capture
https://brandi.js.org/reference/container#restore

src/app/common/UpgradeCheck.spec.ts
kleinfreund: > This is just overriding a method on a class in order to provide a mock, I personally don't think that's a very complicated thing, and probably pretty common.

Absolutely speaking, it’s not very complicated, but compared to what we did before, it is very much more complicated.

> Is this a blocker, or just an opinion?

I label my PR comments very clearly (following roughly what’s outlined in https://conventionalcomments.org/) and so this is a comment.

> It's almost:
>
> env.var = () => '1.2.0'

But it’s really not and it’s not close. If it were that, it wouldn’t seem complicated to do, but as it is, the amount of boilerplate needed for mocking stuff is undoubtedly much more complicated.

---

Thanks for writing up some explanations on those concepts. I understand where these symbols are coming from.

So, in summary `service` retrieves a service and `injected` is ... registering that service again because it was changed? If so, In my opinion, `injected` is a terrible name.

> All the docs for all this stuff is at https://brandi.js.org/, and if we want to keep the thin helper functions, then I'd be happy to add some brief docs for those if it would help.

I don’t think we should be asking people who onboard in this project that they need to be learning Brandi. Not only is this something we said we definitely want to avoid by hiding away even the fact that we are using brandi (or whichever library we use for DI), but it also asks for a degree of understanding of the intricacies of Brandi that really, really shouldn’t be required for a project like Kuma GUI.

What I’m looking for is a short primer that tells me or a new developer how to create a new service and how to use an existing one and how to change one in a context where it needs to be changed.

---

As a point of personal opinion, I want to know as little as I need to be able to use these utilities. I understand at which points we need this kind of complexity, but for the rest of the day-to-day programming activities in this project, I rather keep things simple.

kleinfreund: Thanks again for clarifying some of the terms used here.

If you don’t think documentation for the use of the DI pattern in Kuma GUI, that’s fine by me. I think that would be pretty sensible seeing that it’s not a pattern many people will be overtly familiar with.

kleinfreund (approved):
Save for the one outstanding thread, I have nothing more.

src/app/common/UpgradeCheck.spec.ts
johncowen: No problem, I added more documentation around this in the root of the `@/services` folder just incase (see https://github.com/kumahq/kuma-gui/blob/e15711deda3b51b1aecfc2e93ff79ad37ff31764/src/services/README.md). 

Take into account that the MADR as plenty of info and links to more info also.

src/services/env/Env.ts
johncowen: Neither my text editor nor VSCode pokes me about this, I'm guessing because it is set in the constructor (which is quite common), i.e. after you've constructed an instance it _is_ defined. I added it here anyway even though as a consumer it is pretty much always set. In the scheme of things I don't mind either way.

kleinfreund: You can observe the issue with this by changing the implementation of `var` as follows:

```ts
  var(key: keyof EnvVars, d: string = '') {
    return this.env[key] ?? d
  }
```

It doesn’t result in a TypeScript error but it should (hence why optional chaining is needed) and that’s because the type of `this.env` is incomplete.

Issue #549, opened 2023-01-16
Don't warn on observability when using MeshAccessLog or MeshTrace


lahabana:
### What happened?

This warning shouldn't show when using observability features in policies

![image](https://user-images.githubusercontent.com/702355/212655956-77937349-efac-4aef-a988-ec2294a39642.png)

PR #562, opened 2023-01-18
fix(notifications): Additionally check for MeshAccessLog/MeshTrace before showing notice

johncowen:
Previously we only checked the `mesh.logging` property to decide whether to show a notification to remind you to set up logging, this PR also adds checking to see if there are a MeshAccessLog or MeshTrace policies.

Note, I originally just accessed `rootState.meshInsight.policies` instead of using the getter, but I kinda assumed using the getter would be preferable/was more the done thing (even though this lost the type) - if this is the wrong assumption lemme know!

Fixes https://github.com/kumahq/kuma-gui/issues/549

### DIFF
--- a/src/store/modules/notifications/notifications.spec.ts
+++ b/src/store/modules/notifications/notifications.spec.ts
@@ -29,7 +29,12 @@ describe('notifications module', () => {
         },
       }
 
-      const result = notificationsModule.getters.singleMeshNotificationItems(notificationsModule.state(), getters, rootState, undefined)
+      const rootGetters = {
+        getMeshInsight: {
+          policies: {},
+        },
+      }
+      const result = notificationsModule.getters.singleMeshNotificationItems(notificationsModule.state(), getters, rootState, rootGetters)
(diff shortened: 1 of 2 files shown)


kleinfreund (approved):
As for the question of accessing state directly or via getters: Often, you’ll find applications with getters that directly access the original state themselves. In those cases, it really doesn’t matter which you access. The interesting consideration is when the getter actually does something. Consider a piece of state that’s a list of things. A getter might provide access to that list filtered down by some criteria. Now that’s a case of it being required and useful to access the getter. A getter can also be useful if it abstracts away some extra safe guard checks. But in general, if the getter is itself just direct state access, it doesn’t matter.

johncowen:
🙌 Yeah my thoughts exactly, plus you never know if you need to add something to the getter when previously it didn't do anything extra, and you assume that everything is using getters instead of the state directly, and little did you realise that there was just this one place that was accessing state directly ... boom

So glad I chose the getter then

PR #563, opened 2023-01-20
refactor(zones): migrate components to composition API

kleinfreund:
Migrates all zone-related (and a few simple ones) to composition API.

This aids in providing better type system support for component code which is lackluster for components written in Options API style.

### DIFF
--- a/src/app/zones/views/__snapshots__/ZonesView.spec.ts.snap
+++ b/src/app/zones/views/__snapshots__/ZonesView.spec.ts.snap
@@ -99,7 +99,7 @@ exports[`ZonesView.vue renders snapshot when multizone 1`] = `
                         <span
                           class="sr-only"
                         >
-                          actions
+                          Actions
                         </span>
                         
                         <!---->
@@ -211,7 +211,7 @@ exports[`ZonesView.vue renders snapshot when multizone 1`] = `
                         <span
                           class="sr-only"
                         >
-                          warnings
+                          Warnings
                         </span>
                         
                         <!---->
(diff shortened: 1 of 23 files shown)


src/app/zones/views/__snapshots__/ZonesView.spec.ts.snap
```diff
@@ -888,6 +888,30 @@ exports[`ZonesView.vue renders snapshot when multizone 1`] = `
...
+                                    >
+                                      
+                                      online
+                                      
+                                    </div>
+                                  </div>
+                                  <!---->
+                                </div>
+                              </p>
+                            </li>
```
johncowen, line 914: We gained a `status` output here, but I don't see any mocks that changed, so whilst we might need to show `status` I'm guessing some logic changed somewhere to mean that this now shows?

johncowen (approved):
I had one question just incase, but I think even that is fine, if it is then 👍

src/app/zones/views/__snapshots__/ZonesView.spec.ts.snap
kleinfreund: That’s right. The component clearly indicated that the status was meant to be shown (hence the table column for it). There was other code surrounding statuses, too. However, this was partially broken and resulted in the statuses never making it into the page. Specifically, the component’s `getEntity` method didn’t add the status to the `entity` object and so the table, while trying to display it, never had a value.

So, technically that’s a small bug fix.

johncowen: Gotcha, merge away!

PR #564, opened 2023-01-20
feat: updates dpp policies tab to handle new rule format

kleinfreund:
Updates the DPP policies tab to handle the new rule format. As part of this, the "On" column was replaced with an "Addresses" column.


lahabana (approved):
Thx for the quick fix

PR #566, opened 2023-01-20
chore(routing): Move routes to the top of the application

johncowen:
This PR moves routes to the top of the application into the composition root.

1. I tweaked `createInjections` while I was here, just to make passing the container optional
2. I externalised the routing config and adds it as a service to our container
3. I moved things around and plumbed in the routing through to `createRouter` 

I would guess the exact plumbing of this will evolve slightly. I tried to concentrate on getting this work done without worrying too much about the exact plumbing of things, but I'm pretty sure the exact plumbing will change slightly as we use this pattern for other things moving forwards.

One thing I wanted to note is that ideally the `use*` helpers would only be used from inside Vue components (and I'm considering throwing an error from within them if they are used for anything else), they are only really necessary when we are unable to use constructor injection.

For the moment I'm using `get(TOKEN.name)` to manually inject a few things - but that will potentially change later to rely on the container to automatically resolve and inject things for us.

Top level benefit here is, if we ever have development only routes (such as for embedding developer documentation into the app itself, just as one example) we can now do that in our different development composition root.


src/router/routes.ts
```diff
@@ -0,0 +1,336 @@
+import type { Store } from 'vuex'
+import type { State } from '../store/storeConfig'
```
kleinfreund, line 2: **Suggestion** (nitpick):

```suggestion

import type { State } from '@/store/storeConfig'
```

src/services/production.ts
```diff
@@ -1,5 +1,7 @@
 import { service, constant, get, injected } from './utils'
 import Env, { EnvArgs, EnvVars } from '@/services/env/Env'
+import routes from '../router/routes'
```
kleinfreund, line 3: **Suggestion** (nitpick):

```suggestion
import routes from '@/router/routes'
```

kleinfreund (approved):

PR #569, opened 2023-01-23
fix: policy route using wrong name/param

kleinfreund:
Fixes a small issue with links to specific policy views using the route name for the general policies route and a `policyType` param which doesn’t exist.

### DIFF
--- a/src/app/mesh-overview/views/MeshOverviewView.vue
+++ b/src/app/mesh-overview/views/MeshOverviewView.vue
@@ -71,9 +71,9 @@
                     >
                       <router-link
                         :to="{
-                          name: 'policies',
+                          name: 'policy',
                           params: {
-                            policyType: item.path
+                            policyPath: item.path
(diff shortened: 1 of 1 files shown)


johncowen (approved):
This would be good in the upcoming release right? ( @Automaat FYI. Not sure if its too late if this is something that would be good to go in the upcoming release)

kleinfreund:
> This would be good in the upcoming release right? (@Automaat FYI. Not sure if its too late if this is something that would be good to go in the upcoming release)

I think this is such a minor issue that it doesn’t necessarily need to be included in the release of 2.1.0. In any case, it’s also not difficult to do as long as we haven’t done more than prepare the release branch in Kuma. We can merge it which would automatically propagate it to Kuma’s `master` branch from where it will have to be cherry-picked into the `release-2.1` branch. Leaving this up to @Automaat as the release manager to decide.

PR #578, opened 2023-01-24
fix: not correctly reformatting yaml multiline strings

kleinfreund:
Fixes the YAML reformatting logic to handle multiline strings set on array objects which wasn’t accounted for.


johncowen:
If I use `js-yaml` and run the yaml used in the test case here through `load` and then `dump` and put the output in a `<pre>` I get this:

Screenshot 2023-01-25 at 11 00 23

... which seems correct, unless I'm missing something?

I can think of one reason why we might not want to use this module, but putting that one aside for the moment, historically was there any reason why aren't using approach this for all our YAML displaying purposes? Feels like YAML parsing etc should be something we depend on.

lmk!

kleinfreund:
@johncowen To use a different library that already supports our reformatting needs is a good idea. I’m going to try if this works for all our use cases.

kleinfreund:
I’ve put up a working implementation of switching to js-yaml in https://github.com/kumahq/kuma-gui/pull/578/commits/62e1f754ee32bb040f0b18ffd85ccfdf0bb639bf. There are some differences that I accounted for and a few others that I’ve seen so far that shouldn’t block this change. Most notable is that depending on the source structure, we will now see more quoted values because the data contains the field as a string. This is mostly fine. In some cases, I needed to make adjustments to make sure the underlying data is of type number if the config is supposed to have a number field (e.g. for ports, see data plane proxy wizard).

johncowen:
Ok sounds good

> Most notable is that depending on the source structure, we will now see more quoted values because the data contains the field as a string

^ That's another bug fix right? We saw some values that should have been quoted but weren't yesterday I think.

> In some cases, I needed to make adjustments to make sure the underlying data is of type number if the config is supposed to have a number field (e.g. for ports, see data plane proxy wizard).

From what I understood from your comment there are some places were without your extra fix in here js-yaml sometimes shows strings instead of numbers. I took a quick scan (I really like that you made a thin helper function over `js-yaml`), but will have a proper look in the PR now

kleinfreund:
> That's another bug fix right? We saw some values that should have been quoted but weren't yesterday I think.

Hm, good question. As far as I’m aware, strings can be unquoted in YAML as long as it’s unambiguous that it’s a string value. In any case, I prefer strings being quoted so that there is very visibly no ambiguity possible.

> From what I understood from your comment there are some places were without your extra fix in here js-yaml sometimes shows strings instead of numbers. I took a quick scan (I really like that you made a thin helper function over `js-yaml`), but will have a proper look in the PR now

It’s rather that the object passed to `toYaml` contains string values when they should be numbers. With the previous utility, this didn’t matter because it would never quote them, but with js-yaml, string values are always quoted. Now if the value is actually intended to be a number, that would be wrong (e.g. `port: '8080'` when port is expected to be a number)

johncowen:
Ok gotcha. Yeah agree, all strings should be quoted.

I would imagine if the API gives us an unquoted number then it should be a number, whereas if its quoted it should be a string i.e. we are just displaying what the API says and if it's wrong then it's because the data we are given is wrong.

I'm not totally sure of all the usecases/potential historical things tho. Also, maybe I'm missing something re: js-yaml. Lemme look properly at this PR and maybe we can catch up if I have more q's

johncowen:
> It’s rather that the object passed to toYaml contains string values when they should be numbers. With the previous utility, this didn’t matter because it would never quote them, but with js-yaml, string values are always quoted. Now if the value is actually intended to be a number, that would be wrong (e.g. port: '8080' when port is expected to be a number)

Ah ok so looking at the PR it looks like this is only when the 'numbers' are provided by the user via an input field (which would make them strings) so we have to cast them to numbers first before running them through js-yaml. If this is the case, this all makes sense to me.

I'm guessing that all the API responses will be correct as numbers/strings

Probably should re-word this:

> Ok gotcha. Yeah agree, all strings should be quoted.

Ok gotcha. all non-strings should be quote-less.

I don't think its important that strings are quoted or not considering we are displaying YAML, whereas if its a number it shouldn't be quoted.

Anyway pretty sure this is good to go, let me take on last scan

johncowen (approved):

kleinfreund:
Yes, the special cases that I needed to handle separately come from data that isn’t sourced from the API but in some way from the client side. I haven’t seen anything out of the ordinary for YAML based on API data.

PR #581, opened 2023-01-26
chore: turns off module preload option

kleinfreund:
Turns off the `build.modulePreload` option in the Vite configuration because it interferes with our tests and isn’t strictly necessary.


johncowen (approved):
🎱

PR #583, opened 2023-01-27
fix: crashing on reading non-existing stored mesh

kleinfreund:
Fixes yet another instance of the type of bug that causes the application to crash when reading a stored selected mesh that doesn’t exist.

Also fixes an issue of displaying nav items for mesh-related views when no meshes exist. (To test this, one can empty the mock data for `src/api/mock-data/meshes.json` and `src/api/mock-data/mesh-insights.json`).


src/store/storeConfig.ts
```diff
@@ -255,6 +255,7 @@ export const storeConfig: StoreOptions<State> = {
           // Updates the selected mesh if one wasn’t read earlier or if it’s not an existing mesh.
           const newStoredMesh = ClientStorage.get('selectedMesh')
           if (newStoredMesh === null || !state.meshes.items.some((mesh) => mesh.name === newStoredMesh)) {
+            await dispatch('sidebar/getInsights')
```
johncowen, line 258: From what I remember from last time we said that we should be validating the value of `selectedMesh` before we use it for `sidebar/getInsights`, is this to address that, and if so do we need to remove line 249 (https://github.com/kumahq/kuma-gui/pull/583/files#diff-7487c8e94db54374c1bf362cbee3cce6a715793c1b2fba31289e684994fd154cR249) where we call `sidebar/getInsights`, otherwise this request gets made twice does it?

I also noticed when trying this out that when I have `selectedMesh:all` in localStorage it still makes a HTTP request asking for meshInsights for `all`. Everything seems to fail nicely still, but we shouldn't be making that request at all ideally. I think this code here is possibly all related to that issue.

Super fine by me to address that in a future PR if you didn't want to do it here. This PR is an improvement on its own. Lemme know what you want to do.

johncowen (approved):
Approving, lemme know if you wanted to address that extra HTTP call here or in a follow up tho.

Oh also, do we have to think about zones at all here, guessing not but thought I'd check?

src/store/storeConfig.ts
kleinfreund: I’ve decided to rewrite the logic of this action to be more minimal.

Essentially, the call to `dispatch('sidebar/getInsights')` is now delayed as long as possible. Before doing that, we now check in order whether a stored mesh is present and whether it exists. If meshes are present, we fall back to the first one if the stored mesh can’t be found. If meshes aren’t present at all, we clear the stored value.

kleinfreund:
> Oh also, do we have to think about zones at all here (i.e. is there a case anywhere there that we don't get any and we expect to), guessing not but thought I'd check?

Luckily, no. Zones are truly global resources and our views handle their absence.

If you could have another pass over the changes from https://github.com/kumahq/kuma-gui/pull/583/commits/262e613c4a98a8b5fadbb18a6d27973cae578ae4, that’d be great.

johncowen:
Looks good to me!

```javascript

import { beforeEach, describe, expect, jest, test } from '@jest/globals'
import { ActionHandler, ActionTree, Dispatch } from 'vuex'
import { Mesh } from '@/types/index.d'
import { store } from '@/store/store'


import { State, storeConfig } from '@/store/storeConfig'
import { ClientStorage } from '@/utilities/ClientStorage'

...

  test('bootstrap clears selected mesh when meshes are empty', async () => {
    const rootState = {} as State
    const rootGetters = undefined as any
    const actions = storeConfig.actions as ActionTree<State, State>
    const bootstrap = actions.bootstrap as ActionHandler<State, State>

    const commit = jest.fn()
    expect.assertions(1)
    const dispatch: Dispatch = async function (type: string, payload?: any) {
      if (type === 'updateSelectedMesh') {
        expect(payload).toBeNull()
      }
    }
    const getters = {
      'config/getStatus': 'OK',
    }
    const state = storeConfig.state as State
    state.meshes = {
      items: [] as Mesh[],
      total: 0,
      next: null,
    }

    jest.spyOn(ClientStorage, 'get').mockImplementation(() => 'default')

    await bootstrap.apply(store, [{ commit, dispatch, getters, state, rootState, rootGetters }])
  })
...
```
Something like that ^ quietens `tsc` instead of  `tsc-ignore` on Line 29 if you prefer that:

https://github.com/kumahq/kuma-gui/pull/583/files#diff-deb4c3647ea27617da67ed885e99fe424087cbf431c24372e79392ab440ccf9aR29

...also theres another on 57. Up to you that one though, not sure which you prefer, although guessing `apply`ing could be more correct? Guessing `this` might get used in `bootstrap`?

Feel free to merge either way though, I don't get that extra HTTP request anymore.

PR #584, opened 2023-01-27
chore: Add a component composition root with KumaLogo and GHButton

johncowen:
This PR slightly changes the way we consume components, in that we consume them from a single composition root in `@/components`. I wanted to try it out with multiple components so used the KumaLogo and the recently added GHButton so see/show how the application level code looks when using multiple components like this.

Notes:

1. Whilst exporting functions that return the component seems a little redundant (as opposed to just exporting the component), this matches up with the `use*()` that we use for services and other things, therefore everything works the same.
2. We have a separate composition root specifically for components, ideally we want a single composition root (we already have `@/services`, now we have `@/components` and `@/services`). This avoids problems with circular imports which is caused by not being able to use constructor injection for components. There are probably other ways around this, but this is the simplest, and having 2 composition roots that work exactly the same is fine by me.
3. I haven't used our service container for components as yet (maybe never). The fact that we can't use constructor injection and take advantage of automatic dependency resolution and injection means using that would be kinda redundant. Although I did notice that using the service container here _may_ help with exporting these functions when/if the amount of these grows. We can try that a later date though, we can change the composition root however we like in the future as long as the exports stay the same, which in turn would mean we don't have to change any application code.

### DIFF
--- a/src/app/common/KumaLogo.vue
+++ b/src/app/common/KumaLogo.vue
@@ -1,0 +1,15 @@
+<script lang="ts" setup>
+import { useEnv } from '@/utilities'
+const env = useEnv()
+</script>
+<template>
+  <img
+    src="@/assets/images/product-logo.png"
+    :alt="`${env('KUMA_PRODUCT_NAME')} Logo`"
+  >
+</template>
+<style lang="scss" scoped>
+img {
+  max-height: 36px;
+}
+</style>
--- a/src/app/AppHeader.vue
+++ b/src/app/AppHeader.vue
@@ -2,11 +2,7 @@
   <header class="app-header">
     <div class="horizontal-list">
       <router-link :to="{ name: 'home' }">
-        <img
-          class="logo-image"
-          src="@/assets/images/product-logo.png"
-          :alt="`${store.state.config.tagline} Logo`"
-        >
+        <KumaLogo />
       </router-link>
 
       <GithubButton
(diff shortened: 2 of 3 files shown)


johncowen:
I changed my initial approach here half way through see:

> P.S. Ah of course! It would also be handy to mock out/inject components during testing 🤔

..in the description.

I subsequently changed this to use our container, which meant I could mock/stub components out during testing.

There is an additional snapshot change here as I'd made `<GithubButton>` display `<github-button-stub>` during testing:

https://github.com/kumahq/kuma-gui/blob/master/src/app/App.spec.ts#L22-L24

```javascript
        // keeps the github-button as a <github-button> instead of a span in
        // the snapshot so its as close to actual usage as possible
        GithubButton: true,
```

we lost that in https://github.com/kumahq/kuma-gui/commit/2192080503d5ab852ef493ac5c13373a78d22d1e#diff-2d5de16271ef26e8711852e31081b0a215e86861571b601185eee9024c515b29R22, and it was displaying a `span` again and I couldn't figure out why.

This change puts it back so that the stubbed `<GithubButton>` displays `<github-button>` (which is even better than `<github-button-stub>`) which is what I'd originally wanted, was also an interesting learning and answered a question/query/interest I had around making an empty components easily.

Lastly, I altered `createInjections` slightly now I know we are only ever going to use the one single container.

src/app/common/KumaLogo.vue
```diff
@@ -0,0 +1,15 @@
...
+  <img
+    src="@/assets/images/product-logo.png"
+    :alt="`${env('KUMA_PRODUCT_NAME')} Logo`"
+  >
+</template>
+<style lang="scss" scoped>
+img {
+  max-height: 36px;
+}
+</style>
```
kleinfreund, line 15: **Suggestion**: Can we please stick to one style of arranging single file components and use some empty new lines to separate blocks of things?

```suggestion
<template>
  <img
    src="@/assets/images/product-logo.png"
    :alt="`${env('KUMA_PRODUCT_NAME')} Logo`"
  >
</template>

<script lang="ts" setup>
import { useEnv } from '@/utilities'

const env = useEnv()
</script>

<style lang="scss" scoped>
img {
  max-height: 36px;
}
</style>
```

kleinfreund (approved):

src/app/common/KumaLogo.vue
johncowen: Yeah sure, added a couple of blank lines

PR #587, opened 2023-01-30
refactor: migrates onboarding components to composition api

kleinfreund:
Migrates all onboarding-related components to composition API.

Moves `shouldSuggestOnboarding` getter into global state as it only accesses global state.

Replaces all Tailwind utility classes in the onboarding module with either custom CSS or utility classes that are also present via Kongponents (which already override the Tailwind ones).

Removes some unnecessary router link stubs from some component tests (no more tests now use RouterLinkStub which allows their actual URLs to be visible).

### DIFF
--- a/src/app/App.spec.ts
+++ b/src/app/App.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, jest, test } from '@jest/globals'
-import { flushPromises, mount, RouterLinkStub } from '@vue/test-utils'
+import { flushPromises, mount } from '@vue/test-utils'
 import { withVersion } from '@/../jest/jest-setup-after-env'
 
 import App from './App.vue'
@@ -29,11 +29,11 @@ function renderComponent(status: string) {
       stubs: {
         // Let’s not unnecessarily render all that chart markup.
         DonutChart: true,
-        'router-link': RouterLinkStub,
       },
     },
   })
 }
+
 describe('App.vue', () => {
   test('renders main view when successful', async () => {
     withVersion('10.2.0')
--- a/src/app/App.vue
+++ b/src/app/App.vue
@@ -76,7 +76,7 @@ const isLoading = ref(store.state.globalLoading)
  */
 const routeKey = computed(() => route.path)
 const shouldShowAppError = computed(() => store.state.config.status !== 'OK')
-const shouldSuggestOnboarding = computed(() => store.getters['onboarding/showOnboarding'])
+const shouldSuggestOnboarding = computed(() => store.getters.shouldSuggestOnboarding)
 const shouldShowNotificationManager = computed(() => store.getters['notifications/amountOfActions'] > 0)
 
 watch(() => store.state.globalLoading, function (globalLoading) {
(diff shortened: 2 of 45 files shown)


johncowen:
I had a quick look at this last night and I noticed at least a few design differences with this refactor, was that intentional?

For example:

## Previous
 
Screenshot 2023-01-30 at 17 38 04

## With this PR

Screenshot 2023-01-30 at 17 38 17

kleinfreund:
@johncowen Yeah a lot of small things were broken in the onboarding views. The heading was hidden on account of some CSS rules clipping it away (see screenshots actually). The background would often clip through the header elements that didn’t have a background and the welcome view for some reason was using a different but near-identical scaffold for its layout making it appear much more narrow than all the subsequent views. I still think these views need quite a bit of work, but I wanted to at least some sort of unified experience stepping through these pages.

johncowen:
> The heading was hidden on account of some CSS rules clipping it away (see screenshots actually)

Yeah that's a nice fix definitely, but to my eye I think there are other problems we've introduced in this PR:

## Previous

Screenshot 2023-01-31 at 09 49 41

## With this PR

Screenshot 2023-01-31 at 09 48 41

To my eye (and for sure that's a subjective thing), the previous layout looked a lot better.

I also took a look at KM, and the bug mentioned above (the hidden header) isn't in there, in fact it generally looks a lot better (again to my eye). I wonder if we should just stick to the refactor here and then try to make sure these are in sync. Then if we do want to change the design of all this we can do that as a separate thing. i.e.

1. Fix the visual bugs in kuma (possibly just a sync and/or a tweak)
2. Refactor related components to the new component API (which shouldn't visually change anything)
3. Discuss/implement design changes.

wdyt?

kleinfreund:
@johncowen With this PR, I’m trying to improve things in places that we’re currently not actively working on in the few moments where I actually have time to work on them.

Personally, I don’t think the current state looks a lot better. In fact I think the opposite is true. I think there is no rhyme or reason to the spacing and layout of the views as they are. So could you do me a favor and please raise what you’re concerned about in my changes in a more specific fashion than "the previous layout looked a lot better"?

I don’t want to assign a higher priority to this work as I’m already doing as there isn’t any plans for doing so, so if this becomes a bigger topic, I’d rather not work on it at all and leave it to the future.

johncowen:
Ok sounds like you didn't think much of that then! 😆 Sorry didn't want to sounds dismissive in my previous comment ^, I was trying to figure how to review/check this. Appreciate you are trying to find places to improve the code when you have downtime and I think moving everything to more modern Vue is a good thing.

If it was me, considering this PR was about "refactor: migrates onboarding components to composition api". I'd probably stick to doing that without the visual changes for the moment, that's an improvement and it makes it far easier for me to review (I'm not sidelined with things that I thought could have been unintentional) and avoids conversations that boil down to just differences of opinion on design. But that's just me.

> So could you do me a favor and please raise what you’re concerned about in my changes in a more specific fashion than "the previous layout looked a lot better"?

Off the top of my head, we are now mixing center alignment and left alignment quite a bit. I preferred the "modal-like centrally aligned content" look to this i.e. my concern is that we are now left aligning text whilst the design revolves around a centrally aligned look.

The more general concern is that this PR seems to be a refactor to upgrade APIs from old Vue style to new Vue style - therefore in my mind any visual/behavioural changes are probably unintentional. But from what it sounds like from your comments they are intentional, and therefore my concern is whether we should be making these changes?

Overall I think there are lots of really nice things here, such as the API upgrade and the removal of the router stubs I'd be super happy to just 👍 those if they were on their own. I think thats why I suggested the 3 steps in the above comment.

PR #594, opened 2023-01-30
chore: Adds an injectable logger

johncowen:
This PR adds an injectable logger so we can inject alternative implementations, in this case a DisabledLogger during development.

I've left the little `PROD` check conditional in the code the moment, just as we haven't added the development container into all our package scripts as yet. There are a couple more things I'd like to do before I make using the development container the default container during development, so I'll enable that by default in a future PR. At which point this `PROD` check can be removed.

Couple of extra things:

- I'm not totally aware of whereabout and how we actually log in the code, but on a brief scan, its look like in the future we could potentially use our `logger` directly instead of accessing `datadogLogs.logger` directly. One for the future tho.
- I've added some README in here which is empty for the moment, but is likely to be filled out in the future. I know this is a little redundant right now, but I would rather each 'service' have a README, even if its empty to begin with.


src/services/logger/Logger.ts
```diff
@@ -0,0 +1,26 @@
+import { datadogLogs } from '@datadog/browser-logs'
+
+import { kumaApi } from '@/api/kumaApi'
+import type Env from '@/services/env/Env'
+
+export default class Logger {
```
kleinfreund, line 6: **Suggestion**: Can you please rename this to something more specific? We don’t use Datadog as a general purpose logging utility, but only in a few places.

kleinfreund (approved):

src/services/logger/Logger.ts
johncowen: Here's an example of one place where we use it (but not as yet via the service):

https://github.com/kumahq/kuma-gui/blob/9c37737efeb8b8054aa9a190b1e6bc2813345e77/src/app/common/PaginationWidget.vue#L59-L66

I partly imagined at some point in the future if we wanted to, we could change the:

```javascript
datadogLogs.logger.info(datadogLogEvents.PAGINATION_NEXT_BUTTON_CLICKED) 
```

to

```javascript
const { logger, loggerEvents } = useLogger()
logger.info(loggerEvents.PAGINATION_NEXT_BUTTON_CLICKED) 
```

or similar, which we could switch out during development for a local logger of some sort.

Apart from `Logger` do you have another name in mind? We could call this one `RemoteLogger` perhaps?

kleinfreund: I would much rather call this DatadogLogger. It should be obvious when we’re logging something to an external place *and* what that external place is.

johncowen: K I changed this to DatadogLogger

PR #600, opened 2023-02-01
refactor: remove unnecessary router stubs

kleinfreund:
Removes some unnecessary router link stubs from some component tests (no more tests now use RouterLinkStub which allows their actual URLs to be visible).


johncowen (approved):
🚀

PR #601, opened 2023-02-01
refactor: cleans up onboarding store

kleinfreund:
Moves `shouldSuggestOnboarding` getter into global state as it only accesses global state.

Stops using store setter in component.

Stops setting computed property in component (they should be read-only).

### DIFF
--- a/src/app/App.vue
+++ b/src/app/App.vue
@@ -76,7 +76,7 @@ const isLoading = ref(store.state.globalLoading)
  */
 const routeKey = computed(() => route.path)
 const shouldShowAppError = computed(() => store.state.config.status !== 'OK')
-const shouldSuggestOnboarding = computed(() => store.getters['onboarding/showOnboarding'])
+const shouldSuggestOnboarding = computed(() => store.getters.shouldSuggestOnboarding)
 const shouldShowNotificationManager = computed(() => store.getters['notifications/amountOfActions'] > 0)
 
 watch(() => store.state.globalLoading, function (globalLoading) {
--- a/src/app/onboarding/views/AddNewServices.vue
+++ b/src/app/onboarding/views/AddNewServices.vue
@@ -9,7 +9,7 @@
         <ServiceBox
           :active="mode === 'demo'"
           class="cursor-pointer"
-          @clicked="update('demo')"
+          @clicked="setMode('demo')"
         >
           <div>
             <img src="@/assets/images/new-service-demo.svg?url">
@@ -27,7 +27,7 @@
         <ServiceBox
           :active="mode === 'manually'"
           class="cursor-pointer"
-          @clicked="update('manually')"
+          @clicked="setMode('manually')"
         >
           <div class="cursor-pointer">
             <img src="@/assets/images/new-service-manually.svg?url">
@@ -54,7 +54,8 @@
 </template>
 
 <script>
-import { mapGetters, mapMutations } from 'vuex'
+import { mapState } from 'vuex'
+
 import OnboardingNavigation from '../components/OnboardingNavigation.vue'
 import OnboardingHeading from '../components/OnboardingHeading.vue'
 import OnboardingPage from '../components/OnboardingPage.vue'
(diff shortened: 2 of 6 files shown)


johncowen:
I had a quick look at the code and wondered what the intention was here, are we looking to get rid of the `onboarding` module? If not I would have thought `shouldSuggestOnboarding` should be part of that module?

Also generally, it looks like we are moving the state from the store to the component, i.e. it looks like the string `demo` that comes from the store never gets updated from anywhere in the application (unless I missed it) i.e. we are moving the 'mode' state into the component. Is that right?

If so is there any point in getting the initial state (`demo`) from the store in the first place, if in the store it always equals `demo`?

It also looks to me like the `changeMode` function added here isn't being called anywhere, unless I missed something? Should this be called from somewhere?

I suppose I'm a little confused as to the reasoning behind the PR just by looking at the code and the PR description, could you explain "The Why' a little more maybe, like what's the objective here? That might help.

Looking at how it works on `master` if I select `Manually` in the wizard, and then click Back and the Next again (or Next and Back browser button), `Manually` stays selected - whereas with this PR it doesn't stay select and goes back to `Demo`, I'm guessing that loss of functionality isn't intentional?

kleinfreund:
> I had a quick look at the code and wondered what the intention was here, are we looking to get rid of the `onboarding` module? If not I would have thought `shouldSuggestOnboarding` should be part of that module?

While the name seems to suggest as much, that is not the case. This value is only used specifically in non-onboarding views where it controls when to suggest the onboarding process. To me, it makes much more sense to have this state be computed closely to where its sources are stored, too.

> Also generally, it looks like we are moving the state from the store to the component, i.e. it looks like the string `demo` that comes from the store never gets updated from anywhere in the application (unless I missed it) i.e. we are moving the 'mode' state into the component. Is that right?
>
> If so is there any point in getting the initial state (demo) from the store in the first place, if in the store it always equals demo?
>
> It also looks to me like the changeMode function added here isn't being called anywhere, unless I missed something? Should this be called from somewhere?

That’s my bad. When I split the previous PR apart, I re-refactored this incorrectly. I’ve changed this to what I intended to be doing from the start which is use the action to set the onboarding state `mode`.

johncowen:
Gotcha, lemme take another look

kleinfreund:
> Gotcha, will take another look once you've updated

In case that wasn’t clear from my previous message: I had already made those changes.

johncowen:
Oh sorry I thought you were gonna make more changes on top of this PR? If not, I think those questions still stand then:

> Also generally, it looks like we are moving the state from the store to the component, i.e. it looks like the string demo that comes from the store never gets updated from anywhere in the application (unless I missed it) i.e. we are moving the 'mode' state into the component. Is that right?

> If so is there any point in getting the initial state (demo) from the store in the first place, if in the store it always equals demo?

> It also looks to me like the changeMode function added here isn't being called anywhere, unless I missed something? Should this be called from somewhere?

kleinfreund:
> Oh sorry I thought you were gonna make more changes on top of this PR? If not, I think those questions still stand then:

Then please reevaluate the PR. Those points do not still stand.

The component, as presented in this PR, does not maintain its own state but merely accesses the store’s state. It also causes the appropriate action to be triggered to change said store state.

johncowen:
Ok I just took another look, could you show me where `changeMode` is being called as from what I can see it isn't being called. Both, I could not see where it is being called in the code and ten when I drop a console.log in `changeMode` and take it for a spin I don't see anything logged out into console. Maybe I don't follow how `changeMode` is being called?

I think this could be the overall problem if the intention is to call `changeMode` somewhere.

It would also solve this issue I would think:

> Looking at how it works on `master` if I select `Manually` in the wizard, and then click Back and the Next again (or Next and Back browser button), `Manually` stays selected - whereas with this PR it doesn't stay select and goes back to `Demo`, I'm guessing that loss of functionality isn't intentional?

When you give it a try can you reproduce this regression ^ ?

kleinfreund:
It looks like the remote branch wasn’t actually updated. I (force-)pushed my local branch again and now the changes I mentioned are actually here.

Now you should be able to re-review the PR.

johncowen:
Ah yeah, quick look at that now and at a brief look that answers everything I think, lemme take it for a spin! brb

johncowen (approved):
Yeah looking at the code all makes perfect sense now, and that regression is no longer there.

One last comment on the `shouldSuggestOnboarding` thing. Ideally if something is a 'module' then everything to do with that module should be inside the module, then if you uninstall a module, it's gone entirely - there is nothing else in my code that references `onboarding`.

I actually think this is more about wording, so essentially a nitpick, but how about if we called it `isFreshInstall/isDefaultInstall` or something similar, as thats actually what you are asking when you call that function, i.e. only the default mesh with no dataplanes (but `isOnlyTheDefaultMeshWithNoDataplanes` is a bit long 😆 )

That means if in future we need to know whether something is a fresh install for something that isn't to do with `onboarding`, nobody is going to ask 'should I suggest onboarding?' for that thing that isn't to do with onboarding, more to do with 'is this a fresh install?'.

Anyway as I said its a bit of a nit pick, so no need to change if you don't prefer what I said (if what I said makes sense 😅 )

Otherwise LGTM!

PR #602, opened 2023-02-01
chore: Adds injectable API client

johncowen:
This PR moves our `kumaApi` client to be an injectable/injected object.

In order to do this I didn't want to provide a `useAPI` or similar and go through the entire codebase changing everything as that would take quite a bit of churn. So I replaced the instantiation of `kumaApi` in `@/api/kumaApi` with a reference to the new API service in the service container.

As the `store` files where importing `kumaApi` and we are already importing `store` in the service container/composition root, this meant we have to avoid the circular dependencies. So I made some minimal tweaks to store so that kumaApi could be injected there also, and used a similar approach where we instantiate store so that the entire application uses the store from the service container.

I've also moved all the kumaApi related files/class into its own service folder in `@/services/kuma-api` (a lot of the green in the changset is this), but I've stopped short doing this for store related files for the moment at least.

~Which reminds me I've not added a README.md here for the new service, I'll do that shortly and add a commit on the end with that but please don't let that hold up initial review on the rest.~


johncowen:
Had a last minute change of mind here and decided to pass the entire config object through to `setup` instead of just a `boolean`. This should be ready for 👀 now.

src/store/index.ts
```diff
@@ -0,0 +1,664 @@
+import { StoreOptions } from 'vuex'
```
kleinfreund, line 1: **Suggestion**: Can I ask you to undo the renaming of this file, please? It’s not necessary and I don’t want to promote the possibility of importing from a path for a directory with an index file. Importing from `$PATH/storeConfig` is far more clear than importing from `$PATH` and is also much less ambiguous for module resolution algorithms which have to run more checks for resolving paths when importing using that style.

src/api/kumaApi.ts
```diff
@@ -1,227 +1,2 @@
...
-  }
-
-  getSinglePolicyEntity({ mesh, path, name }: { mesh: string, path: string, name: string }, params?: any): Promise<PolicyEntity> {
-    return this.client.get(`meshes/${mesh}/${path}/${name}`, { params })
-  }
-}
-
-export const kumaApi = new KumaApi()
+import { get, TOKENS } from '@/services'
+export const kumaApi = get(TOKENS.api)
```
kleinfreund, line 2: **Question**: I noticed that it’s no longer possible to jump to the kumaApi definition from a reference. For example, in my editor, when I Alt+Click on `kumaApi` in a piece of code, it used to take me to its implementation. This is now the definition in this file instead of the actual implementation of `kumaApi`.

Looking at this further, this seems to be the case for all services/utilities accessed via the dependency inversion setup because all of their definitions are now subject to the indirection of the dependency inversion pattern (i.e. where their respective service instances are retrieved).

This is quite a downgrade in terms of my developer experience. Is there anything we can do to fix that?

kleinfreund (changes requested):

src/api/kumaApi.ts
johncowen: Not sure if I follow, you want to jump to the definition of `kumaApi`?

The following is with VSCode (but its the same thing in vim etc) but I'm switching 'Alt-Click' for 'Command-Click' - which I think is the same thing in my world:

1. Open up say 'ServiceDetailView.vue', goto line 103 where it says `kumaApi`, I Command-Click on that (Goto Definition) and it takes me to `@/app/kumaApi.ts`. I've moved the definition of kumaAPI into a separate file now, so that's why you don't see the class here.
2. I Command-Click on the `api` part of `TOKENS.api` in `@/app/kumaApi.ts` and it takes me to the definition of `TOKENS.api` on line 24. This shows me that `TOKENS.api` is a `KumaAPI` service.
```javascript
api: service(KumaApi, { description: 'api' }),`
```
4. I Command-Click on `KumaApi` and it takes me to `@/services/kuma-api/KumaApi.ts` (what used to be in `@/app/kumaApi.ts`

Is this a downgrade? Or did I misunderstand? I think it's maybe one click/hop more, but to be fair the class probably should have been in a separate file anyway.

Also, probably worth noting that keeping `@/api/kumaApi.ts` is just a temporary workaround as I don't want to go through the entire codebase right now changing all the `import { kumaAPI } from '@/api/kumaApi'`s to `import { useAPI } from '@/services'`

src/store/index.ts
johncowen: I'm pretty sure it is necessary for all the contents not to be `storeConfig` otherwise we have circular dependency issues (we access the container in the `storeConfig` file so we can't then import that same file from the container file). This is just down to history/baggage, i.e. most of the app currently accesses the store from a hardcoded import instead of an injectable dependency. I don't want to go through the codebase right now changing all this, so I did the smallest amount of churn possible to keep the same functionality throughout the app, but make the app always access the store via the service container.

This was all because we need to make KumaApi injectable (the store uses KumaApi a great deal, but now we need to inject that in)

Happy to rename it to anything you like, let me know what works for you. I just chose `index` as its an _index_ of everything to do with the `store` (apart from the actual instances, which usually would come straight from the service container) and I couldn't think of anything else off the top of my head. It also makes things quite nice to work with imo as you can think of `index` exposing a public interface, and the rest should be private to the module - kinda like a JS module.

kleinfreund: > I'm pretty sure it is necessary for all the contents not to be storeConfig otherwise we have circular dependency issues (we access the container in the storeConfig file so we can't then import that same file from the container file).

Can you please explain what this means? Where does a circular dependency happen? Why does this prevent you from calling this file anything but index.ts?

> Happy to rename it to anything you like, let me know what works for you.

storeConfig.ts is perfectly adequate.

> I just chose index as its an index of everything to do with the store

The current file name of storeConfig does a better job at indicating what it contains because there is likely only going to be one of it in the entire code base whereas with the `$PATH/index.ts` pattern, there are potentially dozens of index files. It also allows developers to quickly open this file using an editor’s command palette by typing `storeConfig` (or some form of abbreviation of it). That’s also not possible if everything is in an index.ts file.

src/api/kumaApi.ts
kleinfreund: Yes, it’s precisely what you outline in those steps: it now takes me 4 steps instead of 1 and I need to remember that the token definition will actually have a reference to the actual API wrapper class. That is a downgrade.

It just would’ve been nice if that were simpler.

src/store/index.ts
johncowen: > Why does this prevent you from calling this file anything but index.ts?

It doesn't prevent me from calling it anything but `index.ts` it prevents me from putting all this in `storeConfig.ts`, anything else is fine.

> storeConfig.ts is perfectly adequate.

it prevents me from putting all this in `storeConfig.ts`, anything else is fine.

> Where does a circular dependency happen? 

If you look in storeConfig as its stands in this PR on the first line:

```
import { get, TOKENS } from '@/services'
```

If the things I need to import from inside the '@/services' file (i.e. `production` or `development`) import the the `@/services` file itself (as is the case with `storeConfig`) then there is a circular dependency.

src/api/kumaApi.ts
johncowen: Just so I undertstand correctly, its not that you can't find the definition, its because you have to click one more time than you did previously?

src/store/index.ts
kleinfreund: Can you think of a way to _not_ rename this file? I really don’t want to lose the ability to look up implementation files like this by popping open the command palette and typing `storeConfig`.

johncowen: You don't lose anything, you just type `store/index` instead of `storeConfig`, or am I not following here? I just changed the name of the file?

I can change it to `storeConfig` if you like, but then I have to change the name of the current `storeConfig` to something else. I don't mind, just let me know what fits you if you don't want `index`, just we need 2 files not 1.

johncowen: Ok so as discussed offline, I removed the one place where we were previously importing `storeConfig` (on `master` we import it from 2 places, but one of those was removed in this PR in favour of using the service container instead). This means we don't need to import the container into `storeConfig.ts` which then means I can then use `storeConfig.ts` for the contents of `index` without running into circular dependency issues, and lastly `index` went in the bin 🎉 

I think thats all that sorted now then?

kleinfreund: Yep, that’s all good now. 👍🏻

kleinfreund:
**Issue**: Something isn’t quite right currently. In the deployment preview, the following error happens and the page isn’t rendering:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'enabled')
```

johncowen:
> Issue: Something isn’t quite right currently. In the deployment preview, the following error happens and the page isn’t rendering:

Ah I kinda half noticed this during dev, but figured as we check for `import.meta.PROD` it wouldn't be an issue as yet. Sounds like preview is `PROD` so logging is enabled, but how the code was ordered the mock server hadn't been setup yet (seems like preview is kinda PROD with mocks, whereas dev isn't PROD but has mocks)

I changed the order here, and also changed the async/await-ness of this so that we aren't waiting for that `getConfig` call before we start the app, which is how it worked before.

kleinfreund (approved):

PR #604, opened 2023-02-02
chore: Ensure utm values are url friendly

johncowen:
There's always the possibility that the product name could contain a space, if that happens make sure its 'url friendly'

### DIFF
--- a/src/services/env/Env.ts
+++ b/src/services/env/Env.ts
@@ -27,9 +27,10 @@ export default class Env {
     const config = readPathConfigFromDom()
     const version = semver(env('KUMA_VERSION', config.version))
 
+    const productName = env('KUMA_PRODUCT_NAME').split(' ').join('-')
     _env = {
       ..._env,
-      KUMA_UTM_QUERY_PARAMS: `utm_source=${env('KUMA_PRODUCT_NAME')}&utm_medium=${env('KUMA_PRODUCT_NAME')}-GUI`,
+      KUMA_UTM_QUERY_PARAMS: `utm_source=${productName}&utm_medium=${productName}-GUI`,
     }
     this.env = {
       ..._env as EnvVars,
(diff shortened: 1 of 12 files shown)


src/services/env/Env.ts
```diff
@@ -27,9 +27,10 @@ export default class Env {
     const config = readPathConfigFromDom()
     const version = semver(env('KUMA_VERSION', config.version))
 
+    const productName = env('KUMA_PRODUCT_NAME').split(' ').join('-')
```
lahabana, line 30: isn't replace a better way to do this?
```suggestion
    const productName = env('KUMA_PRODUCT_NAME').replace(' ', '-')
```
Also should we should uriEncode instead? (We don't really use this utm atm so breaking change should be fine)

lahabana: Also if you clean this remove the (GUI) bit in the utm. GUI is the most common place no need to be specific IMO

kleinfreund: I agree; if we use `encodeURIComponent`, this would allow a consuming side to decode it reliably whereas with dashes, you don’t know what should be a dash and what should be a space.

```suggestion
    const productName = encodeURIComponent(env('KUMA_PRODUCT_NAME'))
```

If you go with the replace call, it should use `replaceAll` to replace all instances not just the first.

```suggestion
    const productName = env('KUMA_PRODUCT_NAME').replaceAll(' ', '-')
```

kleinfreund (approved):

src/services/env/Env.ts
johncowen: > isn't replace a better way to do this?

Yes probably, but watch for one of my most common mistakes - not using `replaceAll` :highfive: 😄 . Much prefer to uri encode it so I'll do that instead, and remove the `GUI` bit, gimme 2 tics. Oh! 🤔 do we even need the `utm_medium=` bit?

johncowen:
@kleinfreund snap/jinx! 😆

johncowen:
Moved to `encodeURIComponent` and remove the `-GUI`, will just wait for tests to pass and then merge

PR #607, opened 2023-02-03
fix: looking up external service with wrong name

### LINKED ISSUE #599: Service view is broken with ExtneralService
### What happened?

When there is an external service like:
```yaml
type: ExternalService
mesh: default
name: httpbin
networking:
  address: httpbin.org:80
tags:
  kuma.io/protocol: http
  kuma.io/service: backend_es
```

service view returns an error:

image

it sends a request to `:5681/meshes/default/external-services/backend_es`, but it should send a request to `http://localhost:5681/meshes/default/external-services/httpbin`

kleinfreund:
Fixes a bug where in the service list view, an external service resource was looked up using the service insight name which isn’t correct. For now, in the absence of the lookup ability via API, we find the corresponding external service in the list of all external services.

Fixes #599.


src/app/services/views/ServiceListView.vue
```diff
@@ -176,7 +176,16 @@ async function loadService({ mesh, name }: { mesh: string, name: string }): Prom
...
+    // The following code is a hotfix for https://github.com/kumahq/kuma-gui/issues/599 until we implement the lookup of `ExternalService` resources by `ServiceInsight` name.
+    const { items } = await kumaApi.getAllExternalServicesFromMesh({ mesh })
+
+    if (Array.isArray(items)) {
+      const foundExternalService = items.find((externalService) => externalService.tags['kuma.io/service'] === name)
+
+      if (foundExternalService !== undefined) {
+        externalService.value = foundExternalService
+      }
+    }
```
johncowen, line 188: This looks fine to me, but how about putting this code inside `kumaApi.getExternalService()` that way if it's needed anywhere else, the code is centralised in the API client. Plus, the fact that we need to do this at all is hidden by our api client. Lastly, this bug shows that the current implementation of `kumaApi.getExternalService()` doesn't work as intended.

Not a blocker at all

johncowen (approved):
LGTM! I left a suggestion inline, not a blocker tho

johncowen:
P.S. hurrah 🎉  for the `release-2.1` base branch 👏

src/app/services/views/ServiceListView.vue
kleinfreund: I specifically don’t want to add it to the API wrapper because it insinuates that the API of retrieving an external service is broken which is not true. If you provide it with the name of an external service, it will work, but that’s not what we’ve been doing in the service views.

However, I noticed that we do actually need this in one other place (the service detail view) and so I externalized this code into a function which I intend to delete again once we fix this properly with the appropriate backend changes.

johncowen:
Can you hold up on the merge for this so I can take another look?

johncowen (changes requested):
What are the reasons for not putting this new function on the `KumaApi` class? P.S. sorry didn't see your comment but wanted to retract the approval (not sure how to do that). Lemme read up.

johncowen:
Yeah this still stands:

> What are the reasons for not putting this new function on the KumaApi class? 

Unless we have a good reason not to use the centralised KumaAPI class for doing this kinda of stuff, then we should move it there. Is there a good reason not to put it as a new method on there?

kleinfreund:
@johncowen And this is why I don’t think we should:

> I specifically don’t want to add it to the API wrapper because it insinuates that the API of retrieving an external service is broken which is not true. If you provide it with the name of an external service, it will work, but that’s not what we’ve been doing in the service views.

johncowen:
> I specifically don’t want to add it to the API wrapper because it insinuates that the API of retrieving an external service is broken which is not true.

Without getting into whether `kumaApi.getExternalService()` is broken or not, why can't we add an entirely new method to the `KumaAPI` service?

kleinfreund:
> Without getting into whether kumaApi.getExternalService() is broken or not, why can't we add an entirely new method to the KumaAPI service?

We can, but I didn’t. It doesn’t matter. We don’t intend to keep this code around and only employ it as a hotfix. Is this preventing the fix from working? If not I’d appreciate to move on with this matter.

johncowen:
> We can

Ok I'd say we do that then if you are ok doing that, then I can re-approve.

kleinfreund:
I’ve made these changes now.

src/api/kumaApi.ts
```diff
@@ -207,6 +207,19 @@ class KumaApi {
     return this.client.get(`meshes/${mesh}/external-services/${name}`, { params })
   }
 
+  // The following code is a hotfix for https://github.com/kumahq/kuma-gui/issues/599 until we implement the lookup of `ExternalService` resources by `ServiceInsight` name.
+  async getExternalServiceByServiceInsightName(mesh: string, name: string): Promise<ExternalService | null> {
+    const { items } = await kumaApi.getAllExternalServicesFromMesh({ mesh })
```
johncowen, line 212: Should `kumaApi` here be `this`?

kleinfreund: Yes, good catch!

johncowen (approved):
Thanks for changing that up 🙏

kleinfreund:
@johncowen I’ve also included the commit for supporting the release branch workflow https://github.com/kumahq/kuma-gui/pull/607/commits/bbebc46b6e65c4768da7fc3699409ddb46d618b7 because I wasn’t totally sure whether it works if it’s not in the branch.

johncowen:
Ok yeah gotcha I think, I'm still thinking about how this is working branching wise, sorry. I initially figured this fix should also be in `master` until the backend addition lands - but looking at this again now and after our chat I don't think that is the plan. i.e. `master` will be broken until the backend fix lands, that's the plan right?

kleinfreund:
> I initially figured this fix should also be in `master` until the backend addition lands - but looking at this again now and after our chat I don't think that is the plan. i.e. `master` will be broken until the backend fix lands, that's the plan right?

No no, I definitely intend to also merge this into master regardless of how we fix this long-term.

johncowen:
Ah ok, I think we are on the same page more or less then, just going about it differently! Anyway this is sorted, thanks again!

PR #608, opened 2023-02-03
refactor: migrates onboarding child components to composition api

kleinfreund:
Migrates all onboarding-related child components to composition API.

Fixes an issue of the onboarding heading not being visible in Chrome on account of Chrome only supporting a vendor-prefixed CSS property we didn’t use.

Replaces a few Tailwind styles in migrated components.

### DIFF
--- a/src/app/onboarding/components/ItemStatus.vue
+++ b/src/app/onboarding/components/ItemStatus.vue
@@ -1,10 +1,10 @@
 <template>
-  <li class="flex items-center mb-2">
-    <span class="circle">
+  <li class="item-status mb-2">
+    <span class="circle mr-2">
       <KIcon
         v-if="props.status"
         icon="check"
-        size="10"
+        size="14"
         color="var(--kuma-purple-1)"
       />
     </span>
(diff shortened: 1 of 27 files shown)


johncowen (commented):
Does this PR cover the fix in https://github.com/kumahq/kuma-gui/pull/606 also? If so shall I close that one or merge it and bring it up here?

I took a quick look at this and there is a least one issue that we'd need to fix up here before merge, whereas I think the other one is good to go, which means I can move 606 over to KM if we merge it.

Let me know, I'm easy either way.

Here's the thing I just spotted now, but I haven't looked in great detail yet.

Screenshot 2023-02-03 at 13 45 10

Note the missing left hand margin.

kleinfreund:
> Does this PR cover the fix in #606 also? If so shall I close that one or merge it and bring it up here?

Right, that’s the exact same fix. Yours is just as good so let’s merge that.

kleinfreund:
> Note the missing left hand margin.

I’ve now added a general padding to the content container of the OnboardingPage container. With the next step of also migrating the view components, I’ll pass over this again and check whether we have duplicated margins now.

johncowen (approved):
Looks good to me! I much prefer the slot approach for text that could potentially be HTML - rather than using attributes 👍

PR #611, opened 2023-02-03
fix: reading subscriptions too early

### LINKED ISSUE #585: Handle missing ZoneInsight
### Description

We can have Zone that does not have ZoneInsight yet

image

kleinfreund:
Fixes an issue of accessing a zone insight’s subscriptions too early when briefly after creating a zone, the zoneInsight object is not available, yet.

Fixes #585.


johncowen (approved):
:shipit:

PR #612, opened 2023-02-03
chore(policies): Remove hardcoded policy listing

johncowen:
Policies are pluggable so we can't refer to them statically in the code.

If you could give this one an extra look that'd be great, whilst I'm gaining more context on the application, I probably still don't have enough to be totally sure there isn't somewhere where we are trying to do `undefined + 0`. Pretty sure what is here is fine after following the code paths I could find through, but would be good for a second opinion on that.


kleinfreund (approved):
Looking at how the involved state is used, this change should be fine.

The most relevant reference is `src/app/mesh-overview/views/MeshOverviewView.vue:169`: `store.state.meshInsight.policies[policyType.name]?.total ?? 0`. It handles the absence of a policy type gracefully.

PR #614, opened 2023-02-03
refactor: migrates onboarding components to composition api

kleinfreund:
Migrates all onboarding-related components to composition API.

Streamlines color usage within the onboarding module to use variables where necessary (mainly replacing hexadecimal colors in templates).

Replaces all Tailwind utility classes in the onboarding module with either custom CSS or utility classes that are also present via Kongponents (which already override the Tailwind ones).


johncowen (approved):
Looks good to me, only question would be, should we put this bit in the center to keep with the central alignment? Which do you think goes better?

## Left aligned:

Screenshot 2023-02-07 at 11 28 15

## Centered:

Screenshot 2023-02-07 at 11 26 46

kleinfreund:
@johncowen I’ve centered the welcome view content now.

I also added another commit to replace hard-coded colors in the graphs with custom properties as discussed earlier.

johncowen:
Sounds good 👍

PR #646, opened 2023-02-13
ci: adds dispatch merged PR notification workflow

kleinfreund:
Adds a workflow for dispatching a merged PR notification which triggers a corresponding workflow in the notify repo.

(Note that the workflow won’t work on this branch as it exists on my fork of Kuma GUI and so it doesn’t actually have access to kumahq/kuma-gui secrets.)


johncowen (approved):
:shipit:

PR #647, opened 2023-02-13
chore: Makes AppHeader and AppSidebar injectable

johncowen:
Moves AppHeader and AppSidebar to our component container.

I also removed the `readonly` routes here as that seems to be unnecessary.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,7 +17,7 @@ import type { State } from '@/store/storeConfig'
 
 export function useApp(
   env: (key: keyof EnvVars) => string,
-  routes: readonly RouteRecordRaw[],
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -17,7 +17,7 @@ import type { State } from '@/store/storeConfig'
 
 export function useApp(
   env: (key: keyof EnvVars) => string,
-  routes: readonly RouteRecordRaw[],
+  routes: RouteRecordRaw[],
```
kleinfreund, line 20: **Question**: Why is `readonly` being removed here?

kleinfreund (approved):

PR #648, opened 2023-02-13
ci: cleans up some code style inconsistencies

kleinfreund:
Makes workflow and job name the same as they’re always in a 1 to 1 relationship in our case.

Replaces double quotes with single quotes where applicable.

Removes spaces inside array bracket boundaries as we were using that style inconsistently.

Re-orders create-gui-pr workflow steps slightly.


johncowen (approved):
:shipit:

PR #662, opened 2023-02-14
chore: Adds type-annotation-spacing spacing lint rule

johncowen:
I noticed that I occasionally miss a space between `:` and type (i.e. `var:Type`). I much prefer the `var: Type` style (with a space after the colon), and it also seems by far to be the style already existing informally in the application.

This PR adds this style to our CI/package scripts so its applied for for any/all contributors.

Couple of corrections included.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -84,15 +84,6 @@ const INLINE_NON_VOID_ELEMENTS = [
         'no-undef': 'off',
         // Turns off some non-TypeScript rules in favor of their specific TypeScript rules to avoid false negatives:
         indent: 'off',
-        '@typescript-eslint/type-annotation-spacing': ['error', {
-          before: false,
-          after: true,
-          overrides: {
-            arrow: {
-              before: true,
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -84,6 +84,15 @@ const INLINE_NON_VOID_ELEMENTS = [
...
         indent: 'off',
+        '@typescript-eslint/type-annotation-spacing': ['error', {
+          before: false,
+          after: true,
+          overrides: {
+            arrow: {
+              before: true,
+            },
+          },
+        }],
```
kleinfreund, line 95: **Suggestion**: Please move this to another place that’s not in between the two indent rules. Those logically relate to each other so it’s nice to keep them next to each other, too.

kleinfreund (approved):

.eslintrc.js
johncowen: Oh! Sure yeah, thats done!

PR #664, opened 2023-02-15
ci: uses .nvmrc file for node

kleinfreund:
Looks like this is supported but not listed on https://github.com/actions/setup-node.


johncowen (approved):
Ah yeah much better, thanks for sorting! I found the docs here:

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

PR #665, opened 2023-02-15
fix: Revert to using /gui for dev and preview environments

johncowen:
TBD


package.json
```diff
@@ -12,6 +12,7 @@
     "preview": "vite preview -c ./vite.config.development.ts",
     "build": "vite build -c ./vite.config.ts",
     "build:ci": "cross-env VITE_MOCK_API_ENABLED=true VITE_FAKE_MULTIZONE=true yarn run build --mode preview -c ./vite.config.development.ts",
+    "build:preview": "yarn run build:ci && mkdir gui && mv dist/* gui && mv gui dist/gui && mv dist/gui/mockServiceWorker.js dist/ && cp _redirects dist/_redirects",
```
kleinfreund: **Question**: Isn’t changing the build script for this purpose a conflict with how users of Kuma are able access the GUI and configure its path?

When we ship Kuma GUI, we do so using the `build` script which produces a `dist` directory whose contents are bundled with the binary distribution of Kuma. Users of Kuma then access the GUI from an arbitrary and at build-time unknown to us GUI path (most commonly `/gui`).

As a consequence, we must be able to serve the GUI from such an arbitrary GUI path without changing the location of Kuma GUI distribution files because users won’t have control over the location of those files as they will be served by a running Kuma environment.

With this change, we would manually intervene with the build script to make things work with a different (i.e. other than root) GUI path.

So, my question is: is this fine to do for the sake of being able to test the most common case of a GUI path in spite of it not exactly representing the capabilities of users to serve the GUI?

Perhaps we can instruct Netlify to know of this altered base directory (see https://app.netlify.com/sites/kuma-gui/settings/deploys) which would perhaps be a better mirror of how users configure the GUI path in Kuma?

johncowen: Hey, just wanted to confirm if we are both on the same page here.

I would like the preview and dev sites to be available at `/gui` instead of `/`. How we go about making sure that is the case is another question that I don't mind either way about and we can get to once I know if we are on the same page with this question or not.

Could you let me know if you would like the preview and dev sites to be available at `/gui` instead of `/`?

kleinfreund: Yes, I’d like for the deployment previews to be served from /gui.

johncowen: Great we are both on the same page then. Let me know if you have another solution to consider, otherwise I'd be happy to go with this one at least for the short term.

> Perhaps we can instruct Netlify to know of this altered base directory (see https://app.netlify.com/sites/kuma-gui/settings/deploys) which would perhaps be a better mirror of how users configure the GUI path in Kuma?

Yeah maybe, let me know what's possible, in the meantime are we good to merge this if this achieves what we both want?

lahabana: If the option suggested by Philipp works that means this would be a config only change.
Therefore if we merge this and then set the config, we'll need to revert this right? Therefore it seems like exploring the alternative is probably something to do prior to merging this

lahabana: I think you need to do this sort of thing: https://docs.netlify.com/routing/redirects/rewrites-proxies/#history-pushstate-and-single-page-apps

johncowen: > If the option suggested by Philipp works that means this would be a config only change.
Therefore if we merge this and then set the config, we'll need to revert this right? Therefore it seems like exploring the alternative is probably something to do prior to merging this

I think we are getting distracted by the fact that this contains an unfortunately necessary change to how we deploy/serve the preview site.

Can we back up a bit?

1. When we are developing the application we currently view our work at http://localhost:8080/ (no path, i.e. root). This isn't how, probably 100%, of our users view our work. They view it at http://localhost:8080/gui. All the time we spend viewing/working on the application and the indirect manual testing that this includes omits the fact that we aren't actually doing this at the correct path.

2. In https://github.com/kumahq/kuma-gui/pull/474 we added a "fallback to default configuration if the JSON config parsing fails" feature. In the conversation on that PR (https://github.com/kumahq/kuma-gui/pull/474#discussion_r1037059981) you can see I didn't think that was that necessary, whereas Philipp pretty strongly did. The thing is if we are going to add something like this then it should be correct, otherwise we may as well just remove it.

In https://github.com/kumahq/kuma-gui/pull/486 we added something to fix the deployment previews. This fix for our deployment previews changed the "fallback default configuration" to have a path that is not the "default" (i.e. '/' instead of '/gui'). If a user does hit upon a problem with the JSON parsing, then their base URL setting is set to '/'. In my mind in fixing our preview deployments by changing the application code we introduced a potential bug. I still don't think this is a huge deal, but Phillip seemed keen to keep this code. In my opinion if we are to keep this "fallback default configuration", it should at least be correct, otherwise we may as well remove it.

---

With this PR our development application (i.e. what we work on locally) is now back to how I think it should be, the same as how probably 100% of our users see it.

The problem is that the preview site needs changing to reflect this. In the grand scheme of things I generally work on this locally so I don't particularly mind what the preview site does as long as it works. Obviously I would rather everything was correct and reflected what I have here locally, but the most important thing to me is what I am working on locally is correct.

---

Coming back around. I don't mind either way how we set up the preview site, as long as it works and preferably is correct (i.e. is served from `/gui`). The rest of the code apart from the one line in the "build:preview" script is all still required for things to work correctly, so rather than revert, if Phillip wants to add something later then he could just change the build script rather than the application code.

If it's just the preview site deploy we are discussing here, and the rest of the code is fixing a bug, we can always merge this and come back with whatever Philipp wants to do afterwards. Although without wanting to sound against a different approach, I would prefer if whatever is done is in code rather than a Netlify web GUI. To be clear, but that's not a strong preference, I really don't mind either way. All in all I don't mind how the preview site is deployed as long as everything else is correct.

Lastly, if this wants to hang here until Philipp comes up with a different solution for the previews then thats also fine by me, as long as the application code included here eventually goes in (or something else to address the same issue)

johncowen: Out of interest, I just went and had a look at what we do in the kuma code:

https://github.com/kumahq/kuma/blob/06c1f62e68e137b725be5d1fac302a2a69aed8d9/pkg/api-server/server.go#L184

Is there a difference between putting a folder at `/gui` and serving it via static webserver vs adding a handler to `restful` at `/gui` that serves the same static files (I think that's whats happening?)?

kleinfreund: In the interest of unblocking this, I’m resolving this thread.

kleinfreund (approved):

johncowen:
> I added a temp commit on the end of here, just so things worked and I could test out the flow without changing netlify's settings, but I would like to change the name of the build script to yarn build:preview which would then use yarn build:ci. If you ignore the final commit here you will see what I mean. Once this PR is merged we'll need to update netlify to use yarn build:preview. (and I can remove the last commit off the end of here before merge)


^ This has been updated in Netlify now. I also added an additional commit here as I realised the `console.error`'ing is noisy in tests (tests are in the CLI so don't have access to the HTML config). Ideally the tests would use a different form of config, and therefore not error out, but I've left that for a future PR.

PR #669, opened 2023-02-17
fix: env not using base URL

kleinfreund:
Fixes RestClient not using `env` for its base URL.

### DIFF
--- a/src/services/env/Env.ts
+++ b/src/services/env/Env.ts
@@ -1,4 +1,4 @@
-import { PATH_CONFIG_DEFAULT } from '@/pathConfigDefault'
+import { getPathConfigDefault } from '@/pathConfigDefault'
 import { PathConfig } from '@/types/index'
 
 export type EnvArgs = {
@@ -8,6 +8,7 @@ export type EnvArgs = {
   KUMA_INSTALL_URL: string
   KUMA_VERSION_URL: string
   KUMA_DOCS_URL: string
+  KUMA_API_URL: string
 }
 type EnvProps = {
   KUMA_VERSION: string
@@ -65,7 +66,7 @@ export default class Env {
 
     // Falls back to a sensible default when encountering a malformed JSON payload
     // or non-replaced template.
-    return PATH_CONFIG_DEFAULT
+    return getPathConfigDefault(import.meta.env.PROD ? window.location.origin : import.meta.env.VITE_KUMA_API_SERVER_URL)
   }
 }
 
--- a/src/pathConfigDefault.ts
+++ b/src/pathConfigDefault.ts
@@ -1,9 +1,11 @@
 import { PathConfig } from './types/index'
 
-export const PATH_CONFIG_DEFAULT: PathConfig = {
-  baseGuiPath: '',
-  // **TIP**: Change this value to test various GUI base path scenarios.
-  // baseGuiPath: '/dev/gui',
-  apiUrl: '/',
-  version: '1.7.0',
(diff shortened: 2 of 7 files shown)


src/services/env/Env.ts
```diff
@@ -65,7 +66,7 @@ export default class Env {
 
     // Falls back to a sensible default when encountering a malformed JSON payload
     // or non-replaced template.
-    return PATH_CONFIG_DEFAULT
+    return getPathConfigDefault(import.meta.env.PROD ? window.location.origin : import.meta.env.VITE_KUMA_API_SERVER_URL)
```
lahabana, line 69: Suggestion: Could this if be moved out into the vite config by having `getPathConfigDefault` take a lambda: `(): string {}` seems like it could help keeping the complexity in a single place or do a systematic if `VITE_KUMA_API_SERVER_URL || window.location.origin`?

lahabana (approved):
LGTM but checking my assumptions are right as I'm super unfamiliar with this:

1. `mode` in vite.config.js is something like 'production' when using KM/Kuma and something else when on Konnect.
2. In the Konnect case we don't pass a template we actually generate something with the right values (.i.e : VITE_KUMA_API_SERVER_URL).

kleinfreund:
> checking my assumptions are right as I'm super unfamiliar with this:
>
> 1. mode in vite.config.js is something like 'production' when using KM/Kuma and something else when on Konnect.

Yes, `mode` is either `'production'` or `'development'`. For Mesh in Konnect GUI, we’ll likely only have `'production'` for PR previews as we want to communicate with real APIs and that’s the biggest influence the mode has: selecting the right value for the API URL (the other one being to not try to log to Datadog in development).

> 2. In the Konnect case we don't pass a template we actually generate something with the right values (.i.e : VITE_KUMA_API_SERVER_URL).

For Mesh in Konnect GUI, we’ll be serving the index.html with a pre-replaced config object like so https://github.com/Kong/kong-mesh-gui/pull/282/files#diff-80940bbea9a896922d15cb6b3cb838b0526b481748c8998529abe29294917453R10-R16 for now. This is likely to change with whatever mechanism we come up that gives us the correct API to talk to in the various stages of deploying MinK to preview and release environments.

lahabana:
Cool so for `2` the value is not a value api url but just a path right (by default: `/` or whatever the value of: `KUMA_API_SERVER_ROOT_URL` is)? So is the right approach to look at the value in our json and if it's a full url use just `apiUrl` and if it's just a path use `window.location.origin + apiUrl`?

kleinfreund:
> Cool so for `2` the value is not a value api url but just a path right (by default: `/` or whatever the value of: `KUMA_API_SERVER_ROOT_URL` is)? So is the right approach to look at the value in our json and if it's a full url use just `apiUrl` and if it's just a path use `window.location.origin + apiUrl`?

Kind of. Preferably, we just simply get an API URL from the right source (currently, the index.html config bit; for MinK eventually something else I reckon) so we don’t have to do additional concatenation of some default base API URL with the changing API path.

In any case, our RestClient class has logic for setting its base URL (and therefore KumaApi’s) depending on whether the provided value is a URL or just a path: in the latter case, the initially provided default base URL is concatenated to it and this is `window.location.origin` in production environments (that’s the bit of code you see changed in this PR, too).

In any case, as I said in https://github.com/kumahq/kuma-gui/pull/669#discussion_r1111752166, I think this could do with some streamlining so that this becomes injectable from one location.

Issue #661, opened 2023-02-14
chore: Move our jest testing setup to use a similar composition/wiring root as our main app


johncowen:
### Description

During https://github.com/kumahq/kuma-gui/pull/644 we re-arranged the majority of our application so the vast majority of our application used a top-level `main.ts` that simply requested the application automatically configured for our application from the service container.

https://github.com/kumahq/kuma-gui/blob/dc34da7a77598fe1fc996abe112ff763643fd804/src/main.ts#L1-L5

This means that we should be able to remove some of our `vite` resolve/aliasing and just use different `main.ts` instead, reducing the slightly strange aliasing method we are currently using

We should also look at using a similar approach for our test harness, to help us further reduce the amount of aliasing we use (especially as currently we have to use a different approach for our jest based test harness to our vite based application).

If we can remove the aliasing entirely from the application it will simplify the approach a little more and reduce some of the awkward paths we have in some configuration files.

TLDR; Use a top-level `main`-like approach for our test harness

johncowen:
Closing this with https://github.com/kumahq/kuma-gui/pull/765

PR #670, opened 2023-02-19
chore: Move get(env) calls into functions ready to inject

johncowen:
Moves some remaining calls to `get` inside our mocks into functions 

Part of https://github.com/kumahq/kuma-gui/issues/661


johncowen:
This has sat in draft for a little bit, but I'm fine for this to go in as is, so marking as ready for review.

slonka (approved):

PR #683, opened 2023-02-21
refactor: extracts app mounting into entry point

kleinfreund:
Moves the call to `app.mount` into the entry point file so that consuming applications can make further `app.use` calls **before** the application is being rendered. This is especially relevant for registering components before trying to render them.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,11 +10,10 @@ import type KumaApi from '@/services/kuma-api/KumaApi'
 import type { State } from '@/store/storeConfig'
 
 /**
- * Initializes and mounts the Vue application.
- *
  * This is a good place to run operations that should ideally be initiated or completed before the Vue application instance exists.
+ *
+ * @returns a factory creating an initialized Vue application with installed store and router without mounting it.
  */
-
 export function useApp(
   env: (key: keyof EnvVars) => string,
   routes: RouteRecordRaw[],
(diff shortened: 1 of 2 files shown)


lahabana (approved):
This LGTM however, it feels like what you are trying to do here is bypassing our standard way to do DI.

I guess you'd want to pass in useApp a `plugins: VuePlugin[]`

and then define vue Plugins with a function:

```js
interface {
function apply(app : vue.App) {
}
}
```

That way you'd update in your `useApp`

```js
const router = await createRouter(routes, store, env('KUMA_BASE_PATH'))
app.use(router)
plugins.each((p) => {
    p.apply(app)
})

app.mount('#app')
return app
```

This way you'll still be able to set these plugins in `kuma/src/services/production.ts` or equivalent no?

Then again I know little about JS so take this as pure suggestion :)

kleinfreund:
@lahabana This isn’t bypassing DI. We’re not passing plugins into `useApp` currently or with this PR. This changeset doesn’t change that. It really just allows us to delay calling `app.mount`.

We could add plugins as a parameter to `useApp` which would work the same for the current setup, but if we need more than just `app.use` calls in a specific order, this wouldn’t work (for example as we do for creating the router).

As it stands, the injectable pieces around this entry point remain unchanged (they are the top-level `App` component and the router) and therefore part of our DI system.

PR #685, opened 2023-02-27
fix: base API URL being set incorrectly

kleinfreund:
Fixes an issue with the base API URL being set incorrectly. To fix this issue, the logic for ensuring the API base URL is indeed an absolute URL and not just a path segment relative to the origin was moved into the Env.ts file where the Kuma config object is being fetched. `RestClient` now always expects an absolute URL without any trailing slashes.

### DIFF
--- a/src/services/env/Env.ts
+++ b/src/services/env/Env.ts
@@ -67,7 +67,7 @@ export default class Env {
         const config = JSON.parse(pathConfigNode.textContent.trim()) as PathConfig
 
         // Ensures the config always has an absolute URL.
-        config.apiUrl = getAbsoluteUrl(window.location.origin, config.apiUrl)
+        config.apiUrl = getNormalizedApiUrl(config.apiUrl)
 
         return config
       } catch {
@@ -76,7 +76,7 @@ export default class Env {
     }
     // console.error('Unable to parse kuma config. Falling back to defaults')
 
-    return getPathConfigDefault(import.meta.env.PROD ? window.location.origin : import.meta.env.VITE_KUMA_API_SERVER_URL)
+    return getPathConfigDefault(import.meta.env.PROD ? '/' : import.meta.env.VITE_KUMA_API_SERVER_URL)
   }
 }
 
@@ -92,17 +92,13 @@ export function semver(version: string): { major: string, minor: string, patch: 
 }
 
 /**
- * @returns an absolute URL given the current origin and a base URL or base path string. When a path is provided, it is concatenated to the current origin; when a URL is provided, no concatenation takes place. In both cases, the URL is returned without any trailing slashes.
+ * @returns a normalized API URL or URL path without trailing slashes. URL paths will always have a leading slash.
  */
-function getAbsoluteUrl(origin: string, baseUrlOrPath: string): string {
-  let baseUrl
-
+function getNormalizedApiUrl(baseUrlOrPath: string): string {
   if (baseUrlOrPath.startsWith('http')) {
-    baseUrl = baseUrlOrPath
+    return baseUrlOrPath.replace(/\/+$/, '')
   } else {
-    const basePath = baseUrlOrPath.replace(/^\/+/, '')
-    baseUrl = [origin, basePath].filter((segment) => segment !== '').join('/')
+    const basePath = (baseUrlOrPath.startsWith('/') ? '' : '/') + baseUrlOrPath
+    return basePath === '/' ? '/' : basePath.replace(/\/+$/, '')
   }
-
-  return baseUrl.replace(/\/+$/, '')
 }
--- a/src/services/env/Env.ts
+++ b/src/services/env/Env.ts
@@ -62,9 +62,9 @@ export default class Env {
 
     // TODO: Uncomment noisy console errors (we don't want them during testing
     // but we do want them for our users)
-    if (pathConfigNode instanceof HTMLScriptElement) {
+    if (pathConfigNode instanceof HTMLScriptElement && pathConfigNode.textContent) {
       try {
-        const config = JSON.parse(pathConfigNode.innerText.trim()) as PathConfig
+        const config = JSON.parse(pathConfigNode.textContent.trim()) as PathConfig
 
         // Ensures the config always has an absolute URL.
         config.apiUrl = getAbsoluteUrl(window.location.origin, config.apiUrl)
--- a/src/services/env/Env.ts
+++ b/src/services/env/Env.ts
@@ -64,9 +64,14 @@ export default class Env {
     // but we do want them for our users)
     if (pathConfigNode instanceof HTMLScriptElement) {
       try {
-        return JSON.parse(pathConfigNode.innerText.trim())
-      } catch (e) {
-        // console.error(e)
+        const config = JSON.parse(pathConfigNode.innerText.trim()) as PathConfig
+
+        // Ensures the config always has an absolute URL.
+        config.apiUrl = getAbsoluteUrl(window.location.origin, config.apiUrl)
+
+        return config
+      } catch {
+        // Handled by falling back to a default value.
       }
     }
     // console.error('Unable to parse kuma config. Falling back to defaults')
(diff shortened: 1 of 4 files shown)


src/services/kuma-api/RestClient.spec.ts
```diff
@@ -15,14 +15,8 @@ describe('RestClient', () => {
   })
 
   test.each([
-    ['api', 'http://localhost:5681/api'],
-    ['/api', 'http://localhost:5681/api'],
-    ['/api/', 'http://localhost:5681/api'],
-    ['test/api', 'http://localhost:5681/test/api'],
-    ['/test/api', 'http://localhost:5681/test/api'],
-    ['/test/api/', 'http://localhost:5681/test/api'],
```
lahabana, line 23: Why do we remove these test cases? They look like genuine options. (At lest the 2nd and 5th).

kleinfreund: Because `RestClient` now always expects an absolute URL. The scenario of there only being a root-relative URL is now covered when we read from the Kuma config object.

src/services/env/Env.ts
```diff
@@ -58,7 +58,12 @@ export default class Env {
 
     if (pathConfigNode instanceof HTMLScriptElement) {
       try {
-        return JSON.parse(pathConfigNode.innerText.trim())
+        const config = JSON.parse(pathConfigNode.innerText.trim()) as PathConfig
+
+        // Ensures the config always has an absolute URL.
+        config.apiUrl = getAbsoluteUrl(window.location.origin, config.apiUrl)
```
johncowen, line 64: How come we are still reading `window.location.origin` shouldn't we just be using whatever is passed into the gui via the HTML vars?

kleinfreund: What we get from Kuma is not an absolute URL but a root-relative URL path like `'/'`. Without the concatenation with `window.location.origin`, this would result in requests made to `http:///meshes/...` which is wrong. We need this as long as we can’t guarantee that we always deal with absolute URLs.

johncowen: Sorry bear with me I'm just looking into this, I would have thought having `/` as a base path would be the same as 'on whichever host I am on' i.e. root.

It looks to me at first glance that the problem in production is that the URLs passed through to `makeRequest` look like `//policies` not `/policies`.

johncowen: Just before I have another look, if this is the case:

> It looks to me at first glance that the problem in production is that the URLs passed through to makeRequest look like `//policies` not `/policies`.

Would you do anything in the PR differently?

lahabana: As mentioned by @kleinfreund there are 2 options for the var in the index.html:

1. a complete url (containing a scheme)
2. an absolute path (but not an absolute URL).

If we're in 2 it is assumed that we're serving the api from the same url as the static files (so window.location.origin).

src/services/kuma-api/RestClient.spec.ts
lahabana: Though it sounds like this test doesn't really make sense anymore (the setter does nothing special)

src/services/env/Env.ts
```diff
@@ -85,3 +90,19 @@ export function semver(version: string): { major: string, minor: string, patch:
     pre: `${major}.${minor}.${patch}${pre !== undefined ? `-${pre}` : ''}`,
   }
 }
+
+/**
+ * @returns an absolute URL given the current origin and a base URL or base path string. When a path is provided, it is concatenated to the current origin; when a URL is provided, no concatenation takes place. In both cases, the URL is returned without any trailing slashes.
+ */
+function getAbsoluteUrl(origin: string, baseUrlOrPath: string): string {
```
lahabana, line 97: Maybe this should be tested?

src/services/env/Env.ts
johncowen: Right yeah gotcha. I think where I might not be following is, it looks like if you specify an absolute path right now i.e. `/`, it makes requests to `//policies` which seems to expand to `http://policies`. If the slashes didn't get doubled up in the frontend code then the requests would be `/policies` not `//policies` and therefore just use localhost i.e. 'from the same url as the static files'?

Maybe I'm not following here?

kleinfreund: Requests in the production build are currently made to URLs of the form `http://policies/` which is just broken structurally.

I’m just ensuring that we always concatenate the origin correctly before making requests. Whether the URLs are explicitly constructed using `window.location.origin` or `/` doesn’t really matter (they’re the same as far as `fetch` is concerned).

johncowen: Ok well it just felt like a lot of code/complexity for what I would have thought would be just `${stripLastSlashIfThereIsOne(config.apiURL)}/policies`.

> Without the concatenation with window.location.origin, this would result in requests made to http:///meshes/... which is wrong. We need this as long as we can’t guarantee that we always deal with absolute URLs.

I guess I wanted to point out that I think the bug isn't that we aren't prepending `window.location.origin`, its that by default in prod we are using double slashes on everything.

So to maybe finish up here, it sounds like you wouldn't do anything differently here considering the above?

src/services/env/Env.ts
kleinfreund: I’ve added tests for this in https://github.com/kumahq/kuma-gui/pull/685/commits/e8bf1314182c3bbda787cd5b4d6129be28ca32c0.

src/services/env/Env.ts
kleinfreund: Not that it matters, but I slightly simplified the implementation to not insist on making URLs absolute.

johncowen: Ok I took another brief scan, so we are not using `window.location.origin` anymore and ensuring we don't double up the slashes, and just use root `/` instead (or whatever comes from `config.apiUrl`)? If so, that sounds good, and I can take another look at it.

kleinfreund: I did not check we don’t use `window.location.origin` at all anymore. It is not relevant to this PR so I’d prefer it to not play a role in review.

This PR remains reviewable.

johncowen:
If I set apiURL to `/` (prod default) I'm still getting calls to `http://policies` with this PR? I just double checked that I've pulled this correctly but I thought I'd mention incase while I triple check what I'm seeing here locally. If I log the urls in `RestClient.raw` I still get `//`s also

johncowen:
Just looking at `KumaApi` would it make more sense to prefix all the calls to `this.client.get` with `/` i.e.:

```javascript
return this.client.get('/policies')
// instead of
return this.client.get('policies')
```

They'd read a whole lot better then also.

Then `client.get` can literally do:

```javascript
`${baseURL}${url}`.

// (baseURL = stripLastSlashesIfThereAreAny(config.apiURL))
// config.apiURL coming from env('KUMA_API_URL')
```

Anyway just a suggestion, that's how I'd personally prefer, happy to still go the route thats here already also.

kleinfreund:
> If I set apiURL to `/` (prod default) I'm still getting calls to `http://policies` with this PR? I just double checked that I've pulled this correctly but I thought I'd mention incase while I triple check what I'm seeing here locally. If I log the urls in `RestClient.raw` I still get `//`s also

I’ve fixed this in https://github.com/kumahq/kuma-gui/pull/685/commits/990c7a9e537f486cbdc199976c16f508f67fa9b2. This didn’t happen with the variant of always using absolute URLs.

kleinfreund:
> Just looking at `KumaApi` would it make more sense to prefix all the calls to `this.client.get` with `/` i.e.:

I don’t want to give this degree of freedom to consumers of the RestClient. I don’t think we want to allow requests to be made _relative to the current URL path_. A call to `restClient.get('policies')` should be identical to `restClient.get('/policies')` and `restClient.get(window.location.origin + '/policies')`.

I’ve added tests ensuring that one can pass both `'policies'` and `'/policies'` style paths and end up with a root-relative request and not a current-path-relative one.

src/services/kuma-api/RestClient.ts
```diff
@@ -60,7 +56,15 @@
...
    */
   async raw(urlOrPath: string, options?: RequestInit & { params?: any }): Promise<{ response: Response, data: any }> {
-    const url = urlOrPath.startsWith('http') ? urlOrPath : [this.baseUrl, urlOrPath].join('/')
+    let url
+
+    if (urlOrPath.startsWith('http')) {
+      url = urlOrPath
+    } else {
+      url = [this.baseUrl, urlOrPath]
+        .map((pathSegment) => pathSegment.replace(/\/+$/, '').replace(/^\/+/, ''))
```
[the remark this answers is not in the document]
kleinfreund: So far have I not seen a helpful instance of CodeQL code scanning.

johncowen (approved):
I messed around quite a bit with it (in dev mode, not built into the binary) and I could no longer see the doubled up slashes, so LGTM!

PR #699, opened 2023-03-02
refactor: create base Api class

kleinfreund:
Creates base Api class for multiple implementations to extend.


src/services/kuma-api/Api.ts
```diff
@@ -0,0 +1,34 @@
+import { RestClient } from './RestClient'
+import Env from '@/services/env/Env'
```
johncowen, line 2: I think this one would be nicer as `import type` also, super nitpick tho.

johncowen (approved):
LGTM! Couple of inline comments but nothing necessary for merge

src/services/kuma-api/Api.ts
kleinfreund: I’ve changed this.

PR #700, opened 2023-03-02
refactor: changes tag style to KBadge

kleinfreund:
Switches the TagList style to use KBadge.

### DIFF
--- a/src/app/services/components/ServiceSummary.vue
+++ b/src/app/services/components/ServiceSummary.vue
@@ -51,7 +51,7 @@
           </div>
         </div>
 
-        <div v-if="tags.length > 0">
+        <div v-if="tags !== null">
           <h2>Tags</h2>
 
           <TagList :tags="tags" />
@@ -139,9 +139,9 @@ const status = computed(() => {
 
 const tags = computed(() => {
   if (props.service.serviceType === 'external' && props.externalService !== null) {
-    return Object.entries(props.externalService.tags).map(([label, value]) => ({ label, value }))
+    return props.externalService.tags
   } else {
-    return []
+    return null
   }
 })
 
--- a/src/app/common/TagList.vue
+++ b/src/app/common/TagList.vue
@@ -88,6 +88,6 @@ function getRoute(tag: LabelValue): RouteLocation | undefined {
 }
 
 .tag-badge a {
-  color: currentColor;
+  text-decoration: none;
 }
 </style>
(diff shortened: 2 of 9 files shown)


johncowen (commented):
I really like the idea of trying to re-use this component for some of our things.

I took a quick look at this though and I think using it like we have done here makes the text look a bit lost/squashed, is there anything we can do via the component to prevent that?

For example:

Before:

Screenshot 2023-03-02 at 12 57 23

With this PR:

Screenshot 2023-03-02 at 12 57 14

When hovered on a table row:

Screenshot 2023-03-02 at 13 03 15

johncowen:
Just wanted to note here that we have been discussing this offline and we are waiting on some design input.

kleinfreund:
On the matter of design, I’ve also got some input. There are significant changes planned and underway in Kongponents that address some of the issues we noted with the badges’ appearance. For this reason, it was recommended to use the badges as they currently come from Kongponents.

src/app/services/components/ServiceSummary.vue
```diff
@@ -139,9 +139,9 @@ const status = computed(() => {
 
 const tags = computed(() => {
   if (props.service.serviceType === 'external' && props.externalService !== null) {
-    return Object.entries(props.externalService.tags).map(([label, value]) => ({ label, value }))
+    return props.externalService.tags
   } else {
-    return []
+    return null
```
johncowen, line 144: Am I right in thinking that if this always returns an array, then you don't have to do all the `Array.isArray` stuff higher up in this changeset?

https://github.com/kumahq/kuma-gui/pull/700/files#diff-ad73132c252f8aacc8adf35d8c8797f251a68e98b1c414cecba95bc63851aefeR40

kleinfreund: No. The checks for this are intended to accommodate the tags as they can appear on our APIs. Only when we get an array (specifically `LabelList[]`) do we skip passing `props.tags` through the `getLabels` function. The latter handles records, `null`, and `undefined` which covers the shapes now possible in `ServiceSummary`. For this code, it wouldn’t matter whether to use `null` or `[]` because the result would be the same.

kleinfreund:
> Juts coming back to this could you let me know if the color differences here are intentional?

Those are our standard link colors and in order to achieve a text color change on hover, I decided to just re-use them instead of introducing a new set of link and hover colors. As a nice side effect, it provides a bit of an indicator that something about those badges is different. Not as good as underlines of course.

johncowen (approved):
LGTM!

PR #703, opened 2023-03-06
chore: Move to a 2 step "define then build" process

johncowen:
This PR is an evolution/iteration of our service container configuration and favours a 2 step "define and then build" approach rather than "build as you go". Consequently we remove `set` and `injected` and set/inject it all in one place, whilst making things more self explanatory ("this is the service, these are its arguments")

---

This PR is "I think" as close as I can get to what I actually want whilst maintaining the types.

Just declarative configuration:

```yaml
env:
  service: Env
  arguments:
    - EnvVars
kumaApi:
  service: KumaApi
  arguments:
    - Env
  labels:
    - api
```

We can then merge/compose configurations as we want, and `build` as a final step.

Notes:
- We are currently building both in prod and dev (so twice in dev). This isn't an issue right now as it's just resetting everything again and isn't intensive in any way. A little longer down the line building will happen in the application itself, not the configuration file, so building will happen once.
- I made some 'decorator' functions for some of the things we duplicated due to inheritance reasons.

Left for later:

- Tagging/labelling
- Remove everything apart from the configuration from the configuration files (e.g. move `build` into the application code)
- Either detect `constant`s behind the scenes during building, or decide on whether to use a function or a property of `ServiceDefinition`. I use a mix of both here as in some places it feels one way is better in other places the other. We should hide `constant` (i.e. detect its a constant behind the scenes) or just use one approach.

More on tagging/labelling:

I'd like to be able to do things like:

```yaml
policyRoutes:
  service: policyRoutes
  arguments:
    - store
   labels:
    - routes
dataplaneRoutes:
  service: dataplaneRoutes
  arguments:
    - store
  labels:
    - routes
```

```javascript
get(TOKENS.routes) // a concatenation of all the services tagged/labelled as routes
```

This makes it easier to compose things together without the app having to know about what exactly things are composed of, it should also make it easier in the future to decorate services, again without knowing what the service is we are decorating is exactly, only that it is of say a `logger` type for example.

None of the "left for later" is super necessary now, but I think what is here is a good iteration on what we have.

### DIFF
--- a/src/services/kuma-api/MockKumaApi.ts
+++ b/src/services/kuma-api/MockKumaApi.ts
@@ -4,22 +4,18 @@ import { setupMockWorker } from '@/api/setupMockWorker'
 import Env from '@/services/env/Env'
 
 export const mockApi = function (parent: typeof KumaApi) {
-  const name = `Mock${parent.name}`
-  const mocked = {
-    [name]: class extends parent {
-      mocks: Mocks
-      constructor(env: Env, mocks: Mocks) {
-        super(env)
-        this.mocks = mocks
-      }
+  return class MockApi extends parent {
+    mocks: Mocks
+    constructor(env: Env, mocks: Mocks) {
+      super(env)
+      this.mocks = mocks
+    }
 
-      setBaseUrl(baseUrl: string): void {
-        super.setBaseUrl(baseUrl)
-        if (this.env.var('KUMA_MOCK_API_ENABLED') === 'true') {
-          setupMockWorker(setupHandlers(this.env.var('KUMA_API_URL'), this.mocks))
-        }
+    setBaseUrl(baseUrl: string): void {
+      super.setBaseUrl(baseUrl)
+      if (this.env.var('KUMA_MOCK_API_ENABLED') === 'true') {
+        setupMockWorker(setupHandlers(this.env.var('KUMA_API_URL'), this.mocks))
       }
-    },
+    }
   }
-  return mocked[name]
 }
--- a/src/services/kuma-api/MockKumaApi.ts
+++ b/src/services/kuma-api/MockKumaApi.ts
@@ -3,20 +3,6 @@ import { setupHandlers, Mocks } from '@/api/mocks'
 import { setupMockWorker } from '@/api/setupMockWorker'
 import Env from '@/services/env/Env'
 
-export default class MockKumaApi extends KumaApi {
-  mocks: Mocks
-  constructor(env: Env, mocks: Mocks) {
-    super(env)
-    this.mocks = mocks
-  }
-
-  setBaseUrl(baseUrl: string): void {
-    super.setBaseUrl(baseUrl)
-    if (this.env.var('KUMA_MOCK_API_ENABLED') === 'true') {
-      setupMockWorker(setupHandlers(this.env.var('KUMA_API_URL'), this.mocks))
-    }
-  }
-}
 export const mockApi = function (parent: typeof KumaApi) {
   const name = `Mock${parent.name}`
   const mocked = {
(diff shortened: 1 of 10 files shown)


src/services/kuma-api/MockKumaApi.ts
```diff
@@ -1,19 +1,25 @@
...
-    }
+      setBaseUrl(baseUrl: string): void {
+        super.setBaseUrl(baseUrl)
+        if (this.env.var('KUMA_MOCK_API_ENABLED') === 'true') {
+          setupMockWorker(setupHandlers(this.env.var('KUMA_API_URL'), this.mocks))
+        }
+      }
+    },
   }
+  return mocked[name]
```
kleinfreund, line 24: **Question**: Why does the structure of `MockKumaApi` and `DisabledLogger` have to change?

This implementation and that of DisabledLogger are turning increasingly more abstract and hard to follow.

From what I can glimpse, `mockApi` is a function that takes a `KumaApi` class as argument and then returns a class that extends from it. Isn’t this just ... what `MockKumaApi` has been but more complicated?

In short, why is this change needed? Where you call `mockApi` now, why is constructing a new instance of `MockKumaApi` not enough?

johncowen: Because there are times when you have two different classes for example:

```javascript
class LoggerOne {}
class LoggerTwo {}
```

We currently extend this to disable things:

```javascript
class DisabledLogger extends LoggerOne {
// ...stuff
}
class DisabledLogger extends LoggerTwo {
// ...duplicated stuff
}
```

This PR means we can just do:

```javascript
disable(LoggerOne)
disable(LoggerTwo)
```

i.e. we just decorate the classes

kleinfreund: Where do we currently need to create two independent extensions like that?

johncowen: FYI: This was explained offline

kleinfreund: I feel like it’s actually an advantage to need to actively disable services in a project rather this being abstracted away, but it seems you’re pretty set on this approach, so I’m not going to argue for it.

src/services/logger/DisabledLogger.ts
```diff
@@ -1,8 +1,17 @@
...
+export const disabledLogger = function (parent: typeof c) {
+  const name = `Disabled${parent.name}`
+  const disabled = {
+    [name]: class extends parent {
+      setup(_config: ClientConfigInterface) {
+        console.log('Logging disabled')
+      }
+    },
+  }
+  return disabled[name]
```
kleinfreund, line 16: **Suggestion**: I’m not sure why you’re first creating an object with a named key only to return the value assigned to that key. Wouldn’t the following be equivalent?

```suggestion
  return class extends parent {
    setup(_config: ClientConfigInterface) {
      console.log('Logging disabled')
    }
  }
```

Same suggestion applies to `MockKumaApi`.

johncowen: It makes sure we keep a name during debugging. Honestly I don't really mind either way, in fact I'd be happier with the suggestion you made here, but in other PRs you made a clear preference to try to ensure we use named functions for debugging purposes, and this ensures its named. Happy to remove it if you don't want it, it was only for your benefit.

kleinfreund: > in other PRs you made a clear preference to try to ensure we use named functions for debugging purposes,

I did that in the very specific context of a warning that made use of a function’s name so this doesn’t apply here.

I’d rather have the simpler version.

kleinfreund (approved):

src/services/logger/DisabledLogger.ts
johncowen: just to check you are fine me using this form right?

```javascript
return class DisabledLogger extends parent ()
///
return class MockApi extends parent ()
```

i.e. statically named classes instead of dynamic? Let me know either way, I'm good with or without.

kleinfreund: Yeah that’s fine

johncowen: Ok just a quick ping to say this is done, will merge in 5mins or so if I hear nothing back 👍

PR #713, opened 2023-03-07
refactor: integrates RestClient into Api

kleinfreund:
## Changes

* **refactor: integrates RestClient into Api**

  Integrates `RestClient` into `Api`. Both classes were always used together and effectively covered the same purpose. This makes specific Api instances a little easier because they can just call `this.get`, etc. instead of accessing these methods via `this.client`.

* **refactor: moves setting base URL out of bootstrap**

  Moves setting the Kuma API base URL out of the bootstrap function so that it doesn’t get overridden when setting it before bootstrap elsewhere.

* **refactor: allows bypassing initial API calls**

  Adds an optional `isAllowedToMakeApiCalls` to the bootstrap function that, if set to `false`, will cause the function to not make any API calls. In that case, it explicitly sets the `config.status` state to `'OK'` to not falsely indicate a not working API.

* **refactor: isolates mock worker from specific API instances**

  Removes `MockKumaApi` in order to isolate the management of mock service worker from specific API instances like `KumaApi` as there should only be one mock server worker handling all mocked APIs.

  Changes `setupMockWorker` to take `apiName` and `handlers` parameters. The function now keeps a record of which handlers belong to which API (as identified by `apiName`). This way, we can store independent handlers per API which might need to be constructed in different ways.


src/services/kuma-api/Api.ts
```diff
@@ -1,34 +1,151 @@
...
+   */
+  async raw(urlOrPath: string, options?: RequestInit & { params?: any }): Promise<{ response: Response, data: any }> {
+    // Normalizes URL and, for URL paths, concatenates the base URL and the URL path.
+    let url
+
+    if (urlOrPath.startsWith('http')) {
+      url = urlOrPath
+    } else {
+      url = [this.baseUrl, urlOrPath]
+        .map((pathSegment) => pathSegment.replace(/\/+$/, '').replace(/^\/+/, ''))
```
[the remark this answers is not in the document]
kleinfreund: No thank you.

src/main.ts
```diff
@@ -2,8 +2,13 @@ import { TOKENS, get } from '@/services'
 import App from './app/App.vue'
 
 async function mountVueApplication() {
+  const env = get(TOKENS.env)
+  const kumaApi = get(TOKENS.api)
+  kumaApi.baseUrl = env('KUMA_API_URL')
```
kleinfreund, line 7: **Comment**: Setting the base URL outside of the bootstrap function allows for re-using it with a different base URL.

src/api/setupMockWorker.ts
```diff
@@ -1,17 +1,29 @@
...
  * Sets up the msw-based mock server.
  */
-export function setupMockWorker(handlers: RestHandler[]): void {
-  const worker = setupWorker(...handlers)
+export function setupMockWorker(apiName: string, handlers: RestHandler[] = []) {
+  handlersMap.set(apiName, handlers)
+
+  const allHandlers = Array.from(handlersMap.values()).flat()
+  // Stops an existing service worker so new listeners are being registered. The warning this causes in the dev tools console does **NOT** seem to be accurate.
+  mockWorker?.stop()
```
kleinfreund, line 18: **Comment**: This cost me several hours of development work. When adding it, MSW logs a message that calling stop for a running worker has no effect and that it should be removed. Only much later did I realize that this isn’t correct.

johncowen:
Hey sorry I know this is still in progress, but just wanted to ask mainly about the following approach in this PR:

> refactor: integrates RestClient into Api

I quite liked the idea that we had a separate HTTP/Rest API client, that we could pass into the API class (i.e. "the place we get ask for data"). That separation means we can pass in different HTTP/RestClient to the API classes if we need to inject different functionality for making data requests (in our case we need HTTP), but if we want to make the base url for this dynamic for example, we can pass in different clients with functionality to detect and use different base paths.

Did you not prefer the injectable http client approach?

src/services/kuma-api/KumaApi.ts
```diff
@@ -29,13 +32,20 @@ import type {
 import type { ClientConfigInterface } from '@/store/modules/config/config.types'
 
 export default class KumaApi extends Api {
+  setupMockWorker() {
+    if (this._env.var('KUMA_MOCK_API_ENABLED') === 'true') {
+      const handlers = setupHandlers(this.baseUrl, get(TOKENS.mocks))
+      setupMockWorker('KumaApi', handlers)
+    }
+  }
```
kleinfreund, line 40: **Comment**: Responsibilities of the individual `Api` instances are pretty much what they were before (then in encapsulated in the `MockKumaApi` class): setting up API instance-specific mock handlers. This remains necessary as the handlers for various API instances may be constructed differently.

kleinfreund:
> I quite liked the idea that we had a separate HTTP/Rest API client, that we could pass into the API class (i.e. "the place we get ask for data"). That separation means we can pass in different HTTP/RestClient to the API classes if we need to inject different functionality for making data requests (in our case we need HTTP), but if we want to make the base url for this dynamic for example, we can pass in different clients with functionality to detect and use different base paths.

> Did you not prefer the injectable http client approach?

The thing I wanted to address with this change _primarily_ is that `RestClient` included many general pieces of functionality that aren’t specific to any specific API instance. These are now all part of `Api`.

The one piece of core information we know is different between API instances is how their complete API URLs are being constructed. And that’s still a piece of information that can be controlled externally and changed per API instance.

In short: we don’t really need a injectable class for each API instance to only change one string.

In our current plans, nothing indicates that we are ever dealing with anything but an HTTP REST API. If that comes to happen, we can think of better abstractions, but right now, this separation was pointless and just split apart responsibilities to the same end goal.

PR #714, opened 2023-03-08
chore: adds eslint-import-resolver-alias

kleinfreund:
* **chore: adds eslint-import-resolver-alias**

  Allows ESLint to know which aliases for paths belong to the same group.

* **chore: runs linter script**


johncowen (approved):
:shipit:

PR #715, opened 2023-03-09
fix: infinite requests when next URL doesn't change

kleinfreund:
Avoids an infinite request loop in the `fetchAllResources` utility that occurs when the `next` URL in the response is present but doesn’t change.


johncowen (approved):
Nice to have these kind of things, I like it! :shipit:

PR #716, opened 2023-03-09
refactor: moves setting base URL out of bootstrap

kleinfreund:
Moves setting the Kuma API base URL out of the bootstrap function so that it doesn’t get overridden when setting it before bootstrap elsewhere.

### DIFF
--- a/src/services/production.ts
+++ b/src/services/production.ts
@@ -32,7 +32,6 @@ const $ = {
 }
 
 export const services: ServiceDefinition[] = [
-
   // Env
   [$.EnvVars, {
     constant: {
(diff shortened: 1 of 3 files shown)


src/services/production.ts
```diff
@@ -112,7 +111,6 @@ export const services: ServiceDefinition[] = [
   [$.bootstrap, {
     service: useBootstrap,
     arguments: [
-      $.env,
```
kleinfreund, line 115: @johncowen One think to look out for I just noticed: This will no longer tell you if the arguments are wrong for the service construction. Previously, we would get a TypeScript error when trying to pass in env here while removing the argument from the `useBootstrap` function. This doesn’t happen here now.

johncowen: Oh right, I will take look at that! Thanks! 👍

PR #717, opened 2023-03-09
feat(di): Adds labelling abilities to our di config

johncowen:
Follow up of #703 as detailed in there , this PR adds labelling abilities to our service/di configuration.

This lets us define multiple things and refer to them before we actually define them. For example, routes in multiple modules, mocks in multiple modules, 'api clients' in multiple modules.

From #703:

> More on tagging/labelling:

> I'd like to be able to do things like:

```yaml
policyRoutes:
  service: policyRoutes
  arguments:
    - store
   labels:
    - routes
dataplaneRoutes:
  service: dataplaneRoutes
  arguments:
    - store
  labels:
    - routes
```
```javascript
get(TOKENS.routes) // a concatenation of all the services tagged/labelled as routes

```

> This makes it easier to compose things together without the app having to know about what exactly things are composed of, it should also make it easier in the future to decorate services, again without knowing what the service is we are decorating is exactly, only that it is of say a logger type for example.

### DIFF
--- a/src/services/utils.ts
+++ b/src/services/utils.ts
@@ -11,6 +11,7 @@ import {
 
 export {
   token,
+  token as label,
 } from 'brandi'
 
 type InjectionHooks<T extends TokenValue[]> = {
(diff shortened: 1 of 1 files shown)


kleinfreund (approved):

kleinfreund:
@johncowen LGTM

PR #718, opened 2023-03-09
refactor: allows bypassing initial API calls

kleinfreund:
Adds an optional `isAllowedToMakeApiCalls` to the bootstrap function that, if set to `false`, will cause the function to not make any API calls. In that case, it explicitly sets the `config.status` state to `'OK'` to not falsely indicate a not working API.


johncowen (approved):
:shipit:

PR #719, opened 2023-03-09
feat: enables mock worker to handle multiple api instances

kleinfreund:
Changes `setupMockWorker` to take `apiName` and `handlers` parameters. The function now keeps a record of which handlers belong to which API (as identified by `apiName`). This way, we can store independent handlers per API which might need to be constructed in different ways.

Makes sure to stop the mock worker before starting it again which seems to be necessary despite a logged warning indicating the opposite. Without it, newly added handlers have no effect.

### DIFF
--- a/src/services/production.ts
+++ b/src/services/production.ts
@@ -1,10 +1,9 @@
-import {
-  RouteRecordRaw,
-} from 'vue-router'
+import { RouteRecordRaw } from 'vue-router'
 import { createStore, StoreOptions, Store } from 'vuex'
 
 import { ServiceDefinition, token, build } from './utils'
 import { useApp, useBootstrap } from '../index'
+import type { Mocks } from '@/api/mocks'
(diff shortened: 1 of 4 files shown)


src/services/production.ts
```diff
@@ -58,6 +58,9 @@ export const services: ServiceDefinition[] = [
   }],
 
   // KumaAPI
+  [$.mocks, {
+    constant: [],
+  }],
```
johncowen, line 63: Do you need to add this if we don't use mocks in production?

johncowen (approved):
I had one question but otherwise 👍

src/services/production.ts
kleinfreund: Rather so that the symbol `TOKENS.mocks` can be resolved in code.

johncowen: ty!

PR #730, opened 2023-03-15
feat: adds support for more request methods

kleinfreund:
Adds support for `delete`, `post`, `put`, and `patch` methods on `RestClient` as our APIs are involving.

Sets the `credentials` option to default to `'include'` for all instances of `RestClient`. Environments in which no session state is transferred via cookies aren’t affected as there wouldn’t be any cookie stored to send with requests. Environments which do need to transfer such session state in turn will benefit from not needing to set `credentials: 'include'` on all the individual API instances.

Ensures URLs are consistently constructed without a trailing slash and adds test cases for that edge case.

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 VITE_INSTALL_URL=https://kuma.io/install/latest/
-VITE_VERSION_URL=https://kuma.io/latest_version/
+VITE_VERSION_URL=https://kuma.io/latest_version
 VITE_NAMESPACE=Kuma
 VITE_KUMA_API_SERVER_URL=http://localhost:5681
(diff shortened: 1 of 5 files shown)


johncowen (approved):
🎱

johncowen:
On that last commit, did you want to just check for a trailing `+json`? Oh doesn't matter, its merged.

kleinfreund:
> On that last commit, did you want to just check for a trailing `+json`?

I suppose it won’t matter, but I’d rather anticipate a rather fixed set of content types here. I really hope we won’t use more than ... two of them.

It would also be a bit more complex to check for a content type ending in `+json` because the content-type header can have more content than just the MIME type: `application/json; charset=utf-8`.

PR #731, opened 2023-03-15
feat: allows hiding breadcrumbs on route level

kleinfreund:
feat: allows hiding breadcrumbs on route level

Allows hiding the breadcrumbs on a per-route basis.


johncowen (approved):
:shipit:

PR #733, opened 2023-03-16
feat: allows hiding alerts

kleinfreund:
Streamlines how general pieces of the UI may show up. Only if their respective entry in `state.defaultVisibility` is `true` become any of their other specific checks relevant.

### DIFF
--- a/src/store/storeConfig.ts
+++ b/src/store/storeConfig.ts
@@ -1,4 +1,5 @@
 import semverCompare from 'semver/functions/compare'
+import { RouteMeta } from 'vue-router'
 import { StoreOptions } from 'vuex'
 
 import { ConfigInterface } from './modules/config/config.types'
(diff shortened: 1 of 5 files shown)


src/store/storeConfig.ts
```diff
@@ -85,6 +97,12 @@ interface BareRootState {
 const initialState: BareRootState = {
   menu: null,
   globalLoading: true,
+  defaultVisibility: {
+    appError: true,
+    notificationManager: true,
+    onboardingNotification: true,
+    breadcrumbs: true,
+  },
```
johncowen, line 105: Are these always `true` in this environment?

kleinfreund: Yes.

kleinfreund: What would "something different" be? Currently, this is only related to the strictly boolean question of whether something shows up in the UI.

johncowen (approved):
🎱

PR #736, opened 2023-03-17
docs(links): fix chat link

slonka:
<Explain your change!>

Did you sign your commit? [Instructions](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md#sign-your-commits)

Have you read [Contributing guidelines](https://github.com/kumahq/.github/blob/main/CONTRIBUTING.md)?


bartsmykla (approved):

johncowen:
I think the GUI uses https://kuma-mesh.slack.com somewhere, does that need to be updated also do you know?

(https://github.com/kumahq/kuma-gui/blob/master/.env#L8)

slonka:
> does that need to be updated also do you know?
I don't think so, this is the chat link, it still works, the only thing that broke is the chat **invite** link.

slonka:
Also, there is a flake somewhere because this passed only on the second run.

PR #737, opened 2023-03-20
fix: stripping slash from absolute URL

kleinfreund:
Fixes stripping trailing slashes from absolute URLs in the RestClient. Generally, our APIs don’t require a trailing slash. But we make one request for getting the latest Kuma version to our docs site and depending on the web server configuration, requesting without a trailing slash can fail because the server responds with a redirect.

### DIFF
--- a/src/services/kuma-api/RestClient.spec.ts
+++ b/src/services/kuma-api/RestClient.spec.ts
@@ -163,6 +163,8 @@ describe('RestClient', () => {
     ['http://example.org/', '/path/', 'http://example.org/path'],
     ['http://example.org/', '', 'http://example.org'],
     ['http://example.org/', '/', 'http://example.org'],
+    ['http://example.org/', 'http://konghq.tech/path', 'http://konghq.tech/path'],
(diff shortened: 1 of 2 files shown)


johncowen:
Could you explain the issue a little more? I think we are talking about:

https://github.com/kumahq/kuma-gui/blob/12b18817675052b07f231c60f0379d453e1f1729/.env#L2

Is that right?

---

If so, it currently doesn't have a trailing slash in the config, it was removed a few days ago in #730 ?

https://github.com/kumahq/kuma-gui/commit/fdaf597e93af6e44d2d2f9c7059dc1ebbc0a55c3#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055dL2

🤔 I'm not sure if I understand the issue correctly, but do we need to put the trailing slash back so the URL in the configuration is correct? We need the trailing slash right? It's not that it shouldn't be there i.e. what's the correct URL `https://kuma.io/latest_version` or `https://kuma.io/latest_version/`?

johncowen (approved):
Questions answered offline. I like that we don't edit any absolute URLs! LGTM

PR #748, opened 2023-03-21
refactor: makes mockApi re-usable

kleinfreund:
Generalizes the `mockApi` utility to work with multiple API instances by accepting the new `mocks` and `setupHandlers` parameters: they’re the details that can vary between API instances and are now provided upfront.

Also renames the `api` token to `kumaApi` to be more explicit.


kleinfreund:
@johncowen I’m closing it though I would’ve of course preferred not to do this work for nothing. Not sure it was entirely necessary to completely refactor the approach to setting up the mock API instances in #751, but since you did, I rather avoid any conflict of interest.

PR #750, opened 2023-03-22
fix: not always determining tag links correctly

kleinfreund:
Fixes `TagList` not always being able to correctly determine tag routes. Specifically, when used on a page where no mesh parameter is in a route, it can’t know which one to use for routes that need it. Adding it explicitly fixes this.


kleinfreund:
> Looks good, the only thing I'd say is, could you remove that last commit ([28536d0](https://github.com/kumahq/kuma-gui/commit/28536d08d624e1679c04cb44731fbaae9466502f))?

Done

johncowen (approved):
Great stuff ty!

Issue #522, opened 2023-01-03
Use generated data for mocks instead of hand written


lahabana:
### Description

Currently all our mocks are manually generated. While this is useful as data is contextually correct it's hard to get an idea of how the GUI behaves when we have 1000s of datapoint.

We should be able to use #521 and provide 2 types of mocks:

1. Correct mocks likely extracted from a running instance
2. Auto generated mocks

johncowen:
This is mostly completed by https://github.com/kumahq/kuma-gui/pull/767.

As it stands now if an old mock exists for a URL it will use that, if one doesn't exist it will use the new style dynamic one, meaning we should get way less 404s when navigating the UI.

There are probably a few follow ups to be done here, but I would guess at a later date

PR #751, opened 2023-03-22
chore: Start moving to dynamic mocks

johncowen:
Currently our mocks are pretty static, its difficult to create infinite amounts of data to mock a large Kuma installation (a new file would be needed for every mesh/dataplane for example)

We would like to move towards having more dynamic mocks so we can view the GUI with infinite amounts of data without having 404 pages show up when visiting certain pages. In the future our tests (both small jest/vitest unit style, and large e2e Cypress-like style) should then be able to reuse these to setup various mocked environments ready for testing without having to make amends to mock files that then breaks things elsewhere.

The test side of things is TBD, but there is a tiny bit of work already in the code that should help with this, I've also tried to make sure that the current tests pass correctly. I've only added 2 new style mock files here (edit: I've added a couple more since) to give an idea of what they will look like, this also means I don't need to be concerned about the test side of things just yet. But it does mean we have to support both styles of mocks. The idea being, once the test side of things is clearer we can slowly but surely replace old style with new.

The whole idea is based around `faker-js` usage (we don't use this in the mock files as yet, but eventually we will).

We also use the labelling work from https://github.com/kumahq/kuma-gui/pull/717 to simply adding both 'types' of mocks both old style and new style, and then having them picked up by MSW.

I'm going to do a self-review myself now and I may ~(or may not)~ add some more inline comments to help review.

P.S. Oh I wanted to at least mention, there are some "naming" things and "where to put things" things I'm still deciding on, so "naming" and "where to put things" are a little subject to change, maybe 🤔 But that sort of stuff shouldn't be a blocker, if things need renaming or moving, we can do that later.

This PR is part of https://github.com/kumahq/kuma-gui/issues/522 (but would not close it on its own)


src/api/mocks/index.ts
```diff
@@ -0,0 +1,37 @@
...
+      const response = fetch(req)
+      return res(ctx.json(response.body))
+    })
+  })
+}
+
+export const fs: Record<string, FakeEndpoint> = {
+  '/': _0,
+  '/config': _1,
+}
```
johncowen, line 37: These will eventually be in a separate (preferably auto-generated) file

src/services/development.ts
```diff
@@ -1,36 +1,102 @@
...
+const bootstrap: ServiceDefinition | undefined = prodServices.find(([token, _]) => token === $.bootstrap)
+if (bootstrap) {
+  bootstrap[1].service = (...rest: Parameters<typeof useBootstrap>) => {
+    const env = get($.Env)
+    if (env.var('KUMA_MOCK_API_ENABLED') === 'true') {
+      get($.msw)
+    }
+    return useBootstrap(...rest)
+  }
+}
```
johncowen, line 100: At some point in the future we can have a think about doing proper from The Outside decorators (a.k.a. sidecars)

src/services/kuma-api/FakeKuma.ts
```diff
@@ -0,0 +1,46 @@
...
+        'grpc',
+        'tcp',
+        'kafka',
+      ],
+    )
+  }
+}
+export default class FakeKuma extends Faker {
+  kuma = new KumaModule(this)
+}
```
johncowen, line 46: We don't use this just yet, but we will, and we'll probably expand on this.

This file gives faker a utility of kuma specific things/words

johncowen (commented):
I left some inline notes, so needed to submit a 'comment review'

jest/jest-setup-after-env.ts
```diff
@@ -40,7 +39,7 @@ expect.addSnapshotSerializer(replaceAttributesSnapshotSerializer([
   'data-tableid',
 ]))
 
-const server = setupMockServer(setupHandlers(import.meta.env.VITE_KUMA_API_SERVER_URL, mocks))
+const server: SetupServer = setupServer(...get(TOKENS.mswHandlers))
```
kleinfreund, line 42: **Suggestion**: This is already correctly inferred and shouldn’t have a type annotation.

```suggestion
const server = setupServer(...get(TOKENS.mswHandlers))
```

src/api/mocks/src/config.ts
```diff
@@ -0,0 +1,258 @@
+import type { EndpointDependencies } from '@/api/mocks/index'
+export default <T extends EndpointDependencies>(_deps: T) => () => {
+  return {
+    headers: {},
+    body: {
```
kleinfreund, line 5: **Suggestion**: Can we please hook up the body with the appropriate type (i.e. `ClientConfigInterface`) so we get type checking in mock data generation? That would be a significant advantage over the current JSON file format.

src/services/development.ts
```diff
@@ -1,36 +1,102 @@
...
+    ],
   }],
-  [$.api, {
-    service: mockApi(KumaApi),
+
+  // Mock Service Worker
+  [$.msw, {
+    service: (handlers: RestHandler[]) => {
+      const worker = setupWorker(...handlers)
+      worker.start({
```
kleinfreund, line 45: **Issue** (maybe): This removes all the logic I added to ensure mocking works when providing msw with new handlers during runtime.

Specifically stopping the worker and starting it with new handlers was essential to make this work. With the removal of `MockKumaApi.ts`, I see no more logic that would ensure handlers are being set again when changing an API instance’s base URL.

Please especially test that this works.

src/services/development.ts
```diff
@@ -1,36 +1,102 @@
...
+      $.Env,
+    ],
   }],
-  [$.api, {
-    service: mockApi(KumaApi),
+
+  // Mock Service Worker
+  [$.msw, {
+    service: (handlers: RestHandler[]) => {
+      const worker = setupWorker(...handlers)
```
kleinfreund, line 44: **Suggestion**: Please keep this message. It helped me several times during debugging the mock setup.

```suggestion
      const worker = setupWorker(...handlers)

      console.warn(
        '%c ✨You are mocking api requests.',
        'background: gray; color: white; display: block; padding: 0.25rem;',
      )
```

src/services/development.ts
```diff
@@ -1,36 +1,102 @@
...
+    labels: [
+      $.mswHandlers,
+    ],
   }],
-])
-build(services)
+]
+
+// straight-forwards bootstrap decorator
+const bootstrap: ServiceDefinition | undefined = prodServices.find(([token, _]) => token === $.bootstrap)
```
kleinfreund, line 91: **Suggestion**: This type is already correctly inferred.

```suggestion
const bootstrap = prodServices.find(([token, _]) => token === $.bootstrap)
```

jest/jest-setup-after-env.ts
johncowen: Done!

src/services/development.ts
johncowen: Ah sorry missed that other `warn`, I've added this back in

src/services/development.ts
johncowen: Done!

src/api/mocks/src/config.ts
johncowen: HTTP responses aren't typed. In my mind any typing should happen in the application layer, not in the HTTP mocks.

src/services/development.ts
```diff
@@ -1,36 +1,102 @@
...
+if (bootstrap) {
+  bootstrap[1].service = (...rest: Parameters<typeof useBootstrap>) => {
+    const env = get($.Env)
+    if (env.var('KUMA_MOCK_API_ENABLED') === 'true') {
+      get($.msw)
```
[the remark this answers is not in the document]
johncowen: At bootstrap time (not application creation time) it `get`s (and therefore initialises MSW)

src/api/mocks/src/config.ts
kleinfreund: I really disagree with this assessment. This is not some closed-off box where we can’t have types (as is the case for actual HTTP responses being surfaced through fetch, etc.). We can easily have types because we already have the definitions and because we write the JavaScript for the mock definitions. We would have a much easier time to spot mistakes in the mock data with this. And there isn’t an better place for this either: once this mock data is served, we’re looking at a running application and are well past a stage where the TypeScript compiler can tell us anything.

Having types here is strictly better than not having them on account of "HTTP responses aren't typed".

src/services/development.ts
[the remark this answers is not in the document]
kleinfreund: I’d like a comment explaining this as the code looks like it’s not doing anything.

src/api/mocks/src/config.ts
kleinfreund: It’s a suggestion and I’m happy to propose the change in a follow-up PR.

That being said, this can be easily accomplished by adding `: { headers: {}, body: ClientConfigInterface }` as the return type for the inner function and importing the appropriate type like so:

```ts
import type { ClientConfigInterface } from '@/store/modules/config/config.types'
```

johncowen: FYI I tried this out:

> That being said, this can be easily accomplished by adding : { headers: {}, body: ClientConfigInterface } as the return type for the inner function and importing the appropriate type like so:

```javascript
import type { ClientConfigInterface } from '@/store/modules/config/config.types'
```

And all the tests/linting failed.

I can fix that up by adding `any` to `MockResponder`, but I'm trying to avoid using `any` if I can help it, so I'm getting into rabbit hole territory I feel.

On a separate point and already mentioned point, I'm still unsure whether I want the functions that I would like to return bare untyped JSON (i.e. `Record<string, unknown>`) return an application type instead. I hope to know a little more once I work on the testing side of this.

Right now, I'm 'pretty sure' I want them to return bare untyped JSON. I may change my mind on this later, but even then I don't think I would type them with our frontend application types. So of right now this is where I am.

All in all, even if we were ok with `any`s, I'd very much rather leave them untyped, at least for the moment.

kleinfreund: > And all the tests/linting failed.

Exactly _one_ TypeScript error arises which propagates into the tests failing because virtually all of them import this file.

Avoiding `any` for a piece of data that we won’t read from in application logic in trade for having typed mock definitions is not worth it, but anyway.

src/services/development.ts
kleinfreund: Have you tested that this works when changing the base URL during runtime?

src/api/mocks/src/config.ts
johncowen: Appreciate your take on this, and I'm super happy to revisit at a later date. I want to try some more things out before I decide exactly what I would like and we can discuss further. Just so I understand correctly - is this convo resolved in the meantime?

src/services/development.ts
johncowen: Do you mean the 'base' URL or the 'prefix' path?

If you mean can the prefix path change at runtime, then yes, tested. All works fine as far as I can see.

If you mean the base URL, then that doesn't really change at runtime as far as I understand it. It's set once, before MSW initialises and it stays the same throughout the lifetime of the application.

We need to clear up the difference between the base URL and a prefix path.

I know in the application we currently change the entire base URL over and over again, like there is no prefix path. But as this is the mocks/MSW, not the application, it works differently ie. `${baseURL}/${prefix}/${url}`

---

For the lifetime of MSW:

The baseURL doesn't change once set, ever.
The prefix path doesn't change, but contains a dynamic segment which is variable. Sometimes we use a prefix path, sometimes we don't (see next point for how we do this)
The url is our traditional URL, that doesn't change either but contains more dynamic segments that are variable. We sometimes dynamically prefix these URLs with the prefix path via a loop-de-loop through the majority of URLs from The Outside (i.e. from our service container).

---

If you think about it like that ^ it makes the whole thing an awful lot easier.

Hopefully that's clear enough for how I think about it, but let me know if you have any questions/queries or you need anything clarifying.

kleinfreund: All I wanted to know is if you tested changing API URLs to be reflected correctly in the mocks and you replied yes (be it in a very long manner).

kleinfreund (approved):

PR #755, opened 2023-03-23
fix: not bundling all font files

kleinfreund:
Fixes a bug with the build not bundling all font files which was caused by the extracted woff2 files actually being _identical_ between different weights which Rollup actually seems to detect and optimize away. Deleting the redundant files and using the same file name for all `@font-face` rules seems to fix the issue.


johncowen:
Was this one of those "looks fine in dev, but not when built" things?

kleinfreund:
> Was this one of those "looks fine in dev, but not when built" things?

Yes. And it’s really quite intransparent. You might even check the built files and find references to woff2 files and the woff2 files themselves and miss that it’s only 4 files instead of 8. I would expect for such an optimization to take into account that there might be references to the assets that are being "optimized" away, but that doesn’t seem to be the case. Maybe it’s even an actual bug in that process of the build tool.

johncowen (approved):
🚀

PR #765, opened 2023-03-29
chore: Adds merging mocker function and uses for DataPlaneListView.spec

johncowen:
This PR is a first draft at re-using our new dynamic mocks for testing whilst ensuring we can setup mocks with as little fixture-like code as possible, and still maintaining support for snapshotting (at least in the short term)

There are some things in this PR that are more important to me than others:

Important:

1. Start moving the jest-setup file to use a similar service container approach to our `main.ts` files. This file is almost '`main.ts` for tests'. Just to note: our spec files are similar to Vue components in that we cannot do any sort of constructor/parameter injection, so we will always need to use out `use*` service location functions in test files the same as with components.
2. Getting the functionality for the test mocker right: Provide a url, some env vars, and potentially a way to tweak the response. Note: I've noticed are beginning to get more reusable now we have our service container.
3. Keeping the same coverage for the test/spec that was changed here, but instead of relying on the fact that the mock had the expected data, explicitly setting up the exact data we expect inside the test for each test.

Semi-important:
1. Making sure we can keep snapshotting functionality along with dynamic mocks. Hardcoding a faker seed for the test keeps the data stable between snapshotting.

Still important but less so here but will become more important in a later PR:
1. The innards of the `mocker` function (and our `fakeApi` function). This is one of those things that will have to improve as things become clearer/come together a little more. Right now I can almost see which bits we can reuse or generalise to reduce `mocker` and `fakeApi` to the same function. I didn't want to do that here though. Firstly because punting this means I can work on adding more mocks _and_ generalising these functions at the same time (without this PR in I can't keep going with the mocks) and secondly, I want to have a think about how this would look from a Cypress/browser runner perspective. Ideally devmode/jest/vite/cypress etc etc would all use the same functionality/tests with or without mocks. Just to be clear, the code in `@/api/mocks/index` and `@/test-support/index` will probably end up being merged into one file/import path at `@/test-support`.

As always please ask if there are any questions or anything isn't clear.

I have one question myself:

Currently all our mocks and things live in `@/api` just due to historical reasons. You can probably see in this PR we also will have a handful of test utils (which I've started adding to a `@/test-support` path). I will probably move all test related/mocking things into the `@/test-support` folder, meaning we can remove `@/api` altogether once we've moved everything over to the new style mocks. Does anyone have any thoughts there?

(I may add some inline comment to help review)

### DIFF
--- a/jest/jest-setup-after-env.ts
+++ b/jest/jest-setup-after-env.ts
@@ -2,77 +2,93 @@ import { afterAll, afterEach, beforeAll, beforeEach, expect, jest } from '@jest/
 // Polyfills `window.fetch` for Jest because it runs in a Node environment where fetch isn’t available. It initially looked like this would change with Node.js 18, but that is not so.
 import 'isomorphic-fetch'
 import { config } from '@vue/test-utils'
-import { rest, MockedRequest as Request } from 'msw'
 import { setupServer } from 'msw/node'
 
 import { replaceAttributesSnapshotSerializer } from './jest-replace-attribute-snapshot-serializer'
 import { TOKENS as COMPONENT_TOKENS } from '../src/components'
-import { createRouter } from '../src/router/router'
-import { TOKENS, get, container, build } from '../src/services/development'
+import { TOKENS, get, container, build, merge, createInjections } from '../src/services/development'
+import { token } from '../src/services/utils'
 import Env from '@/services/env/Env'
-
-type MockFunction = (_opts: Record<string, unknown>, cb: (req: Request, resp: Record <string, any>) => Record<string, unknown>) => void
+import { mocker } from '@/test-support'
+import type { Mocker } from '@/test-support'
+import type { RestHandler } from 'msw'
 
 // jest can't import this module properly due to transpiling issues
 // mock this out with a blank element
 jest.mock('vue-github-button', () => ({ template: '<span />' }))
-
-/**
- * Adds the application’s router to vue test utils. This way tests don’t have to set-up a new router instance on their own.
- */
-const router = createRouter(get(TOKENS.routes), get(TOKENS.store))
-config.global.plugins.push(router)
-
-/**
- * Kongponents uses generated UUIDs for several attribute values.
- * This breaks the project’s snapshot tests since they’re based on fully-mounted components
- * which also includes those from external sources like Kongponents.
- *
- * In order to stabilize the tests which otherwise fail because the attribute values are different every run,
- * we use a custom snapshot serializer to replace those attribute values with one fixed value.
- */
-expect.addSnapshotSerializer(replaceAttributesSnapshotSerializer([
-  'id',
-  'aria-describedby',
-  'aria-labelledby',
-  'aria-controls',
-  'data-tableid',
-]))
-
-const server = setupServer(...get(TOKENS.mswHandlers))
-
-beforeAll(() => server.listen())
-afterEach(() => server.resetHandlers())
-afterAll(() => server.close())
-
-// unless we actually use COMPONENT_TOKENS it won't actually get executed
-// probably due to tree shaking/rollup import ordering. This mixed with
-// container capturing/restoring to make our tests isolated means that
-// potentially we can capture an empty container before all the tokens are set,
-// then set the TOKENS/fill the container during a test then the container can
-// get restored to empty whilst we still have TOKENS with now non-existent
-// services accessing TOKENS before we do anything means we set the TOKENS and
-// fill the container with the default services i.e. before we capture if we
-// ever make a test mocking utility to mock out components (similar to
-// withVersion) will will then use COMPONENT_TOKENS here also, which means we
-// can remove the following line
-beforeAll((_ = COMPONENT_TOKENS) => {})
 //
-beforeEach(() => container.capture?.())
-afterEach(() => container.restore?.())
 
-// add a utility to easily setup/mock out API endpoints
-const re = /\+/g
+const $ = {
+  ...TOKENS,
+  mock: token<Mocker>('mocker'),
+};
 
-const useMock = (url: string, response: Record<string, unknown>): MockFunction => {
-  return (_opts, cb) => {
-    server.use(
-      rest.get(`${import.meta.env.VITE_KUMA_API_SERVER_URL.slice(0, -1)}${url.replace(re, '\\+')}`, (req, res, ctx) => {
-        return res(ctx.json(cb(req, JSON.parse(JSON.stringify(response)))))
-      }),
-    )
-  }
-}
+(async () => {
+  const services = merge([
+    [$.msw, {
+      service: (handlers: RestHandler[]) => {
+        return setupServer(...handlers)
+      },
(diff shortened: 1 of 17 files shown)


src/api/mocks/fs.ts
```diff
@@ -0,0 +1,22 @@
...
+  '/': _0,
+  '/config': _1,
+  '/meshes/:mesh/traffic-permissions': _2,
+  '/meshes/:mesh/dataplanes+insights': _3,
+  '/meshes/:mesh/dataplanes+insights/:name': _6,
+  '/meshes/:mesh/dataplanes/:name': _5,
+  '/meshes/:mesh/dataplanes/:name/clusters': _7,
+  '/meshes/:mesh/dataplanes/:name/stats': _4,
+  '/meshes/:mesh/dataplanes/:name/xds': _8,
+}
```
johncowen, line 22: I started splitting this off, I would imagine this could potentially be auto generated in the future.

src/api/mocks/index.ts
```diff
@@ -1,31 +1,28 @@
...
+type Pager = (total: string | number, req: RestRequest, self: string) => {
   next: string | null,
   pageTotal: number,
   total: number,
   offset: number,
   size: number
 }
-const pager: Pager = (total, req: RestRequest, self) => {
+const pager: Pager = (_total: string | number, req: RestRequest, self) => {
+  const baseUrl = 'http://localhost:5681'
```
johncowen, line 18: We'll end up using `env()` here also

src/app/data-planes/views/DataPlaneListView.spec.ts
```diff
@@ -46,13 +92,43 @@ describe('DataPlaneListView', () => {
...
+                  gateway: {
+                    type: 'BUILTIN',
+                  },
+                },
+              },
+            },
+          ],
+        },
+      })
+    })
```
johncowen, line 131: Note we don't have to provide the entire response here, just the bits we are interested in.

src/services/development.ts
```diff
@@ -5,20 +5,36 @@ import { merge, build, ServiceDefinition, token, get } from './utils'
...
+  kumaFS: token<Record<string, FakeEndpoint>>('fake.fs.kuma'),
+  /**
+   * @description
+   * Service Label for labeling MSW handlers for consumption via setupWorker
+   */
   mswHandlers: token<RestHandler[]>('msw.handlers'),
+  /**
+   * @description
+   * Service Label for labeling fake FSs for consumption via MSW
+   */
```
johncowen, line 36: Oh I wasn't gonna add these here! Are these helpful or not? I'm hoping they'll help when you need to use these tokens in other files.

src/services/production.ts
```diff
@@ -85,6 +86,15 @@ export const services: ServiceDefinition[] = [
...
 
+  // Router
+  [$.router, {
+    service: createRouter,
+    arguments: [
+      $.routes,
+      $.store,
+    ],
+  }],
+
```
johncowen, line 97: This is pulled into our test `main` file, not yet our application `main` file, but we may as well do the same in our application `main` file soon.

johncowen (commented):
Just adding self-review notes:

src/services/development.ts
kleinfreund: Yes, those comments will be added to tooltips in editors when hovering over, for example, `TOKENS.api`. Though personally I would not add `@description` because it doesn’t really add any value. The main comment content will be added, too, just with the difference that it won’t be prefixed with `@description`.

jest/jest-setup-after-env.ts
```diff
@@ -2,77 +2,93 @@ import { afterAll, afterEach, beforeAll, beforeEach, expect, jest } from '@jest/
...
+    }],
+    [$.mock, {
+      service: mocker,
+      arguments: [
+        $.env,
+        $.msw,
+        $.fakeFS,
+      ],
+    }],
+  ])
```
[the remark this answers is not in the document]
kleinfreund: Yeah, I would much rather have the service definitions in a testing.ts file than here.

jest/jest-setup-after-env.ts
```diff
@@ -2,77 +2,93 @@ import { afterAll, afterEach, beforeAll, beforeEach, expect, jest } from '@jest/
...
 
-const server = setupServer(...get(TOKENS.mswHandlers))
+  build(
+    services,
+  )
```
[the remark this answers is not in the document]
kleinfreund: This makes sense to me, too.

jest/jest-setup-after-env.ts
```diff
@@ -2,77 +2,93 @@ import { afterAll, afterEach, beforeAll, beforeEach, expect, jest } from '@jest/
...
-expect.addSnapshotSerializer(replaceAttributesSnapshotSerializer([
-  'id',
-  'aria-describedby',
-  'aria-labelledby',
-  'aria-controls',
-  'data-tableid',
-]))
+(async () => {
+  const services = merge([
+    [$.msw, {
```
kleinfreund, line 28: **Suggestion**: I’d prefer this service to be called `mswServer` and the worker in development.ts `mswWorker` to not create the illusion that this service is interchangeable. While they serve the same purpose, they’re functionally different and won’t ever be able to be used in the respective other environment (i.e. you can’t use the MSW server in a browser context and you can’t use the MSW worker in a Node.js context) so there isn’t a benefit to making them the same service.

johncowen: What happens if you have application code that does `get($.msw)` in a browser _or_ CLI. Do you have to add application code to detect whether you are in a browser or node and call `get($.mswWorker)` or `get($.mswServer)` accordingly?

Alternatively (as we do here) we keep it as `get($.msw)` and inject a different implementation from The Outside?

johncowen: > Where would you need that in the first place? MSW is not used in application code but as part of the setup of the individual testing or development environments.

We use it here, which I would count as application code:

https://github.com/kumahq/kuma-gui/blob/1972104661ae1dd4ace8a0b46fc4af9460a1e099/src/services/development.ts#L97-L107

If I were to run this code in a browser I want it to get the worker, whereas if I run it in some node CLI based tests I want it to return the equivalent in node i.e. the node version of the worker.

Should I add a conditional to detect whether this is browser or node here? Or should I also overwrite `bootstrap` yet again during testing? Alternatively (as we do here) we keep it as `get($.msw)` and inject a different implementation from The Outside?

jest/jest-setup-after-env.ts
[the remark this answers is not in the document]
johncowen: 👍 I added [b2147b6](https://github.com/kumahq/kuma-gui/pull/765/commits/b2147b6e15d3ef6f9ee6264426025ef75ea1b2cc) to cover this

jest/jest-setup-after-env.ts
johncowen: Ok thanks for the viewpoint, if it becomes problematic I'll happily change it.

Lemme know if theres anything else in the overall PR.

PR #766, opened 2023-03-29
fix: amount sometimes being incorrectly positioned

kleinfreund:
**fix: amount sometimes being incorrectly positioned**

Removes the absolute positioning from the nav item amount which caused it to be pulled down with an increase in bottom padding.

**chore: removes border bottom from nav items**

Removes the functionality for adding a bottom border to certain nav items and tweaks the space between groups of nav items a little to account for it.

### DIFF
--- a/src/app/AppNavItem.vue
+++ b/src/app/AppNavItem.vue
@@ -2,7 +2,7 @@
   <div
     class="nav-item"
     :class="{
-      [`nav-item-${props.routeName}`]: true,
+      [`nav-item-${props.routeName}`]: props.routeName !== '',
       'nav-item--is-category': targetRoute === null,
       [`nav-item--is-${props.categoryTier}-category`]: props.categoryTier !== null,
     }"
--- a/src/app/AppNavItem.vue
+++ b/src/app/AppNavItem.vue
@@ -2,10 +2,11 @@
   <div
     class="nav-item"
     :class="{
+      [`nav-item-${props.routeName}`]: true,
       'nav-item--is-category': targetRoute === null,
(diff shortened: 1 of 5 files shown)


johncowen:
I think this looks fine.

How much extra work would it be to put `props.routeName` as a class on each AppNavItem?

It looks like we already do `:data-testid="props.routeName"`, so could we copy that to `:class="props.routeName"` also? (or if you wanted to be extra namespace-y`:class="nav-item-${props.routeName}"`, either is fine by me)

Lmk, I'd give this a 👍 whichever way.

kleinfreund:
> How much extra work would it be to put `props.routeName` as a class on each AppNavItem?
> 
> It looks like we already do `:data-testid="props.routeName"`, so could we copy that to `:class="props.routeName"` also? (or if you wanted to be extra namespace-y`:class="nav-item-${props.routeName}"`, either is fine by me)

That would be simple, yes. What do you want to use it for?

kleinfreund:
> Well, it would be handy to have if we wanted to tweak a menu item from the CSS, shall we add that while we are here if its simple?

I’ve added this now.

johncowen:
Sorry one last thing: How comes we needed to add that `|| undefined` in that last commit? I'm asking as we added that for the data-testid and not the className. Is props.routeName ever falsey?

kleinfreund:
> Sorry one last thing: How comes we needed to add that || undefined in that last commit?

It prevents adding an empty `data-testid` to categories which don’t have a route name.

> I'm asking as we added that for the data-testid and not the className. Is props.routeName ever falsey?

I actually need to add the same protection here, too

PR #770, opened 2023-03-31
fix: using non-existing black color property

kleinfreund:
**fix: using non-existing black color property**

Replaces `--black-75` with `--black-500` because `--black-75` isn’t actually a color we define anywhere (see https://kongponents.konghq.com/guide/styles/colors.html for reference).

Also sets a default color that’s not quite as black as #000 (i.e. `--black-500`) on the `body` element.

**fix: using slightly different color for some borders**

Fixes the border for the top header and sidebar being a slightly different grey than the border we see usually in the app (e.g. the frame around many blocks/panels, the border beneath the Home nav item, etc.).


johncowen (approved):
Oh weird did black-75 used to be in kongponents or something? (doesn't matter if you don't know, just curious)

LGTM!

kleinfreund:
> Oh weird did black-75 used to be in kongponents or something? (doesn't matter if you don't know, just curious)

I don’t even know. Maybe it’s never been in there. 🤷🏻‍♂️

kleinfreund:
@johncowen Actually, I forgot something that I’ve been meaning to do as part of this: https://github.com/kumahq/kuma-gui/pull/770/commits/735cdcd98dbe83e58567c3e1bd0dadf102cd45d4

Could you have a look at this, please? I’m just re-using the `--KCardBorder` property for identical border definitions so it’s all neatly tied up in one place.

johncowen:
Nice 👍 looks good to me also!

kleinfreund:
@johncowen I realized that I’ve been a bit overeager with https://github.com/kumahq/kuma-gui/pull/770/commits/735cdcd98dbe83e58567c3e1bd0dadf102cd45d4. Some of those shouldn’t have this property and I reverted this partially in https://github.com/kumahq/kuma-gui/pull/770/commits/8972448a09dac41a230ea62579170a4a062cef0b.

Also, I noticed that in some place where I changed this, the CSS was actually wholly unused so I removed it instead: https://github.com/kumahq/kuma-gui/pull/770/commits/9c3a0f8ab6586d78a839a1f310c79061cc68656b

johncowen:
Ok well maybe let's let this sit for a bit before I give another approval in case you want to tweak a bit more, I'll come back in a few hours 👍

kleinfreund:
> Ok well maybe let's let this sit for a bit before I give another approval in case you want to tweak a bit more, I'll come back in a few hours

I’m done now though.

johncowen (approved):
LGTM!

PR #784, opened 2023-04-04
chore: Remove all old style mocks

johncowen:
This PR:

- Removes all the old style 'fixture-like' static mocks, we no longer need to support both formats.
- Moves all our new mocks to `@/test-support`.
- Move the code used for 'serving' mocks files during development and 'mocking' mock files during testing over to `@/test-support`. I'm still unsure on the naming of things like `fakeApi` and `mocker`, at least these are all in one place in `@/test-support` now - so if the names change at any point we can just use import aliasing if need be. All in all though its just easier that these 'test-support' functions are in one place (not in a couple of different places)

The PR is quite large unfortunately, so I tried to split it into easier to digest commits, so looking at those might help review. The majority of the PR is setting up tests with the `mock` function to ensure that the test pass without relying on the old static mocks. As always, any questions please ask 👍 

(I may do some inline self-review comments shortly to help)

Closes https://github.com/kumahq/kuma-gui/issues/522


src/app/App.spec.ts
```diff
@@ -56,17 +56,25 @@ describe('App.vue', () => {
...
+    const wrapper = renderComponent()
+
     expect(wrapper.find('[data-testid="app-progress-bar"]').exists()).toBe(true)
 
     await flushPromises()
     await store.dispatch('updateGlobalLoading', false)
     expect(wrapper.find('[data-testid="app-progress-bar"]').exists()).toBe(false)
 
-    expect(wrapper.element).toMatchSnapshot()
+    expect(wrapper.html()).toContain('Unable to reach the API')
```
johncowen, line 78: I removed the snapshot match here as the previous test setup was still loading in other API requests perfectly fine, and was creating all the donut charts etc. Really we are testing for whether the `Unable to reach the API` message is showing, so I changed it to test just for that.

src/services/development.ts
```diff
@@ -80,44 +81,29 @@ export const services: ServiceDefinition[] = [
...
       $.mswHandlers,
     ],
   }],
 
-  // new style kuma mocks
   [$.kumaFS, {
     constant: fs,
-    arguments: [
-      $.Env,
-    ],
```
johncowen, line 105: This is cruft left over from a previous change, you can't pass arguments to `constant`s

src/services/development.ts
```diff
@@ -80,44 +81,29 @@ export const services: ServiceDefinition[] = [
...
     ],
   }],
-
-  // old style kuma mocks
-  [$.mswKumaMocks, {
-    service: (env: Env) => setupHandlers(env.var('KUMA_API_URL'), mocks, env),
+  [$.mswFakeApiHandlers, {
+    service: (env: TokenType<typeof $.env>, fs: FS) => fakeApi(env, rest, fs),
     arguments: [
-      $.Env,
```
johncowen, line 93: All the old style mocks have gone now.

src/services/development.ts
```diff
@@ -80,44 +81,29 @@ export const services: ServiceDefinition[] = [
...
-  [$.mswKumaHandlers, {
-    service: (env: TokenType<typeof $.env>, fs: FS) => fakeApi(env, fs),
-    arguments: [
-      $.env,
-      $.fakeFS,
-    ],
-    labels: [
-      $.mswHandlers,
-    ],
-  }],
```
johncowen, line 120: This was pretty much just moved, token name changed, and I also added the extra `rest` argument (`fakeApi(env, rest, fs)`). You'll see the final result over in the green of the changeset.

johncowen, line 120: This was pretty much just moved, token name changed, and I also added the extra `rest` argument (`fakeApi(env, rest, fs)`). You'll see the final result over in the green of the changeset.

src/utilities/helpers.ts
```diff
@@ -138,9 +138,9 @@ export async function fetchAllResources<T = Object>(endpoint: (params: Object) =
         allTotal = total
       }
 
-      if (total !== allTotal) {
-        throw new Error('Mismatch between "total" values between requests')
-      }
+      // if (total !== allTotal) {
+      //   throw new Error('Mismatch between "total" values between requests')
+      // }
```
johncowen, line 143: I wanted to double check this before complete removal.

This seems to check that the `total` stays the same throughout all pages of a paged request, and if it doesn't then an error is thrown.

This feels to me like it is actually expected behaviour that the total could change between requesting page one and page two of a request. Erroring if the total changes is probably not quite what we want to do.

I came across this as during testing. As our mocks no longer have state, it is quite possible that the total is different between requesting page one and page two (etc etc).

All in all I think deleting this is safe, and desirable, but thought I'd ask for a second opinion incase.

I'll be either deleting or uncommenting this before merge (probably deleting)

johncowen (commented):
Self-review inline notes:

lahabana (approved):

PR #802, opened 2023-04-13
fix: empty charts not rendering a grey arc

kleinfreund:
* fix: empty charts not rendering a grey arc
* fix: missing gap from app header

### DIFF
--- a/src/app/AppHeader.vue
+++ b/src/app/AppHeader.vue
@@ -146,6 +146,7 @@ const mode = computed(() => store.getters['config/getMulticlusterStatus'] ? 'Mul
   display: flex;
   justify-content: space-between;
   align-items: center;
+  gap: var(--spacing-lg);
   padding-right: var(--spacing-lg);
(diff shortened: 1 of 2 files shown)


johncowen (commented):
LGTM!

How easy would it be to remove the rollover state also? Lmk. If its not straightforwards I can just approve this as is.

(also, probably need to change the base branch of this back to `master`)

kleinfreund:
> How easy would it be to remove the rollover state also? Lmk. If its not straightforwards I can just approve this as is.

Probably not hard, but I don’t think it should be removed as we use the same appearance for no data than for when we do have data save for the difference in arc color.

> (also, probably need to change the base branch of this back to `master`)

I’ll be merging the current fixes I have open PRs for into the release branch still because it’s more hassle to update the PRs than to backport the changes. Also, this way, in the event of releasing a patch version of 2.2, the fixes will be included.

johncowen (approved):

Issue #798, opened 2023-04-11
di: Move service config to use composition rather than inheritance


johncowen:
### Description

We initially set up our service container usage using a "build the container as you go", inheritance based approach to easily get started splitting our application up into separate modules/environments.

During https://github.com/kumahq/kuma-gui/pull/703 we moved to a 2 step "define and build" approach for our service container (rather than "build as you go").

The above PR also set things up ready to move the entire service config to use composition rather than inheritance via `merge` and `build`, but we left actually doing that to a later date.

From https://github.com/kumahq/kuma-gui/pull/703:

> We are currently building both in prod and dev (so twice in dev). This isn't an issue right now as it's just resetting everything again and isn't intensive in any way. A little longer down the line building will happen in the application itself, not the configuration file, so building will happen once.

> Remove everything apart from the configuration from the configuration files (e.g. move build into the application code)

The code currently works with the `development` service config importing the `production` service config, then merging and building the config together inside the `development` file. We then import _only_ the development file from `main.ts`, giving us an inheritance-like structure (the fact that `production` is imported is invisible to the application).

Preferably, instead of configuration files importing each other in series, we would import everything we need in the `main.ts` file and then compose and build in the main file itself (i.e. the application itself), giving us a more compositional pattern and more visibility into how the application is made up in the `main` file itself.

Our testing service config is already hints at this improved structure:

The config: https://github.com/kumahq/kuma-gui/blob/master/src/services/testing.ts
Consuming the config: https://github.com/kumahq/kuma-gui/blob/db6a5e4dec523dec4136a9eb0b48f133ef95f0ba/jest/jest-setup-after-env.ts#L23-L29

Once we have shifted things around a little to use the above structure, we _should_ be able to stop "vite-aliasing" `@/services`, which should make things clearer moving forwards as well as setting the codebase up to become even more modular/pluggable.

cc @lahabana

PR #803, opened 2023-04-17
feat: Adds cypress based fakeAPI mocking plus cucumber usage example

johncowen:
This PR uses Cypress as a browser based test runner to run our more e2e-like tests. Additionally it does this by using cucumber as a thin layer over test assertions etc. Before explaining further, I just wanted to point out that when I say 'e2e', I mean 'frontend e2e' i.e. these tests should not use any backend infrastructure for testing (at least for the moment) These are 'e2e' tests, but without testing the things that we don't own.

We have a longish running desire to run our e2e-like tests in a browser rather than using Node/CLI via Jest, this PR is an approach to that task.

I've used the existing CLI based e2e-like tests from DataPlaneListView.spec (https://github.com/kumahq/kuma-gui/blob/1c7311e7d093397b853ce7f50dd224f7daf13493/src/app/data-planes/views/DataPlaneListView.spec.ts) as a starting point for browser based testing, and you'll see that the new tests here replicate the DataplaneListView tests that we are currently using a CLI to test. I've immediately seen one benefit in using this approach as it prevents things like:

https://github.com/kumahq/kuma-gui/blob/1c7311e7d093397b853ce7f50dd224f7daf13493/src/app/data-planes/views/DataPlaneListView.spec.ts#L292

where we've told the tests to `expect` without performing an assertion, so the above line currently doesn't assert anything in the places that we use it. In cucumber you'd get an error if you did something similar saying the step doesn't exist.

Another benefit seems to be non-clashing jest/cypress types.

There is work still todo here which I would like to do as a relatively soon follow up task[s], but it's currently at a point where we move all of our e2e tests forwards without more noodling with code (there are typescript types to sort out and probably a bit of duplication to remove). I've kept all the code that could potentially be noodled with in the one place in `@/test-support`, and as we are using cucumber for syntax we can noodle all we like with the underlying code (when the time is right) and it doesn't effect any of our potentially 1000s of tests.

Briefly:

1. The main bit of work was getting our dynamic mocks to work within Cypress, but this time using `cy.intercept` rather than MSW. It doesn't look like Cypress can perform the variable substitution from URLs for you (i.e. `/meshes/:mesh/` > `{mesh: "default"}`, so I've used URLPattern to do that since I'm also using that for something else/elsewhere. This would also be useful on the MSW side of things.
2. We currently use MSW types for a Request data object (i.e. `params` and `query` etc). For the moment I've just `@ts-ignore`d the places where the types originating from Cypress don't match the MSW types we currently use. These will need making more generic/less MSW-coupled when we come back to noodle with this. (edit: see comment I make below, whilst waiting on review I've noodled a bit and I think I've sorted this)
3. We will need to use our `@/services/development` file so we can re-use all the mock file setup (aswell as the ability to label these for them to automatically picked up) similar to what we do during development/jest testing. Currently Cypress gives a tonne of log errors related to loading `.vue` files. There is a task here https://github.com/kumahq/kuma-gui/issues/798 to split development out so it doesn't also load `production` so that might help here. Although it might also be better to configure Cypress to load these if that's possible. I'm unsure about this right now/still thinking.
4. `env` will also need to come from the main application (as it does when using MSW)
5. Probably more comments... (I'll edit in here if/when I think of them)

All in all I wanted to get this up now for 👀 rather than noodle more.

This is not set up in CI yet and shouldn't be until we are a bit further ahead with it, so for the moment you should run `yarn dev:real-api` (we don't want MSW mocking) on port `8080` and run `npx cypress open` in order to see this working (you might have to do the ol' `cypress install` thing previous to that also)

As always, any questions around getting it running or anything else please ask. We also have a GUI sync up tomorrow so it might be a good opportunity to discuss this further then.

### DIFF
--- a/src/test-support/mocks/src/dataplanes+insights.ts
+++ b/src/test-support/mocks/src/dataplanes+insights.ts
@@ -1,10 +1,9 @@
 import type { EndpointDependencies, MockResponder } from '@/test-support'
 export default ({ fake, pager, env }: EndpointDependencies): MockResponder => (req) => {
-  const params = req.params
   const { offset, total, next, pageTotal } = pager(
     env('KUMA_DATAPLANE_COUNT', `${fake.datatype.number({ min: 1, max: 1000 })}`),
     req,
-    `/meshes/${params.mesh}/dataplanes+insights`,
+    '/dataplanes+insights',
   )
   return {
     headers: {},
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,4 +1,5 @@
 import { addCucumberPreprocessorPlugin } from '@badeball/cypress-cucumber-preprocessor'
+// eslint-disable-next-line import/no-named-as-default
 import createEsbuildPlugin from '@badeball/cypress-cucumber-preprocessor/esbuild'
 import createBundler from '@bahmutov/cypress-esbuild-preprocessor'
 import { defineConfig } from 'cypress'
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,0 +1,36 @@
+import { addCucumberPreprocessorPlugin } from '@badeball/cypress-cucumber-preprocessor'
+import createEsbuildPlugin from '@badeball/cypress-cucumber-preprocessor/esbuild'
+import createBundler from '@bahmutov/cypress-esbuild-preprocessor'
+import { defineConfig } from 'cypress'
+
+export default defineConfig({
+  e2e: {
+    specPattern: '**/*.feature',
+    // TODO Env var
+    video: false,
+    async setupNodeEvents(
+      on: Cypress.PluginEvents,
+      config: Cypress.PluginConfigOptions,
+    ) {
+      // This is required for the preprocessor to be able to generate JSON reports after each run, and more,
+      await addCucumberPreprocessorPlugin(on, config)
+
+      on('task', {
+        log(message: unknown) {
+          console.log(JSON.stringify(message))
+          return null
+        },
+      })
+
+      on(
+        'file:preprocessor',
+        createBundler({
+          plugins: [createEsbuildPlugin(config)],
+        }),
+      )
+
+      // Make sure to return the config object as it might have been modified by the plugin.
+      return config
+    },
+  },
+})
--- a/cypress/support/e2e.ts
+++ b/cypress/support/e2e.ts
@@ -1,0 +1,20 @@
+// ***********************************************************
+// This example support/e2e.ts is processed and
+// loaded automatically before your test files.
+//
+// This is a great place to put global configuration and
+// behavior that modifies Cypress.
+//
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+// import './commands'
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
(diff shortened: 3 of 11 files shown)


src/test-support/mocks/src/dataplanes+insights.ts
```diff
@@ -1,10 +1,9 @@
 import type { EndpointDependencies, MockResponder } from '@/test-support'
 export default ({ fake, pager, env }: EndpointDependencies): MockResponder => (req) => {
-  const params = req.params
   const { offset, total, next, pageTotal } = pager(
     env('KUMA_DATAPLANE_COUNT', `${fake.datatype.number({ min: 1, max: 1000 })}`),
     req,
-    `/meshes/${params.mesh}/dataplanes+insights`,
+    '/dataplanes+insights',
```
johncowen, line 6: I spotted this was wrong so fixed this up whilst I was here, would be nice for `Pager` to automatically figure these out in the future.

johncowen: I think there's a couple. One is at root based and one is mesh based:

https://github.com/kumahq/kuma-gui/blob/1c7311e7d093397b853ce7f50dd224f7daf13493/src/services/kuma-api/KumaApi.ts#L134-L140

This one is the root one. Not totally sure if we still use the root based one in the GUI. 👀 

At first glance it looks like we don't so we could completely remove the root one (and any other similar root ones).

lahabana: Ok feel free to open an issue to figure this out later and ignore this comment :)

johncowen:
After PRing this I tried some aforementioned noodling to remove some duplication and remove the `@ts-ignore`.

Completely informal work in progress right now, but here's a link to that https://github.com/johncowen/kuma-gui/compare/feature/cypress...johncowen:kuma-gui:chore/cypress-refact I'm likely to PR this separately soon after. Thought I'd mention here and give a changeset to view even though I can't nest a PR here. I'll try and describe more in the desc for it once I PR it.

johncowen:
I think the other thing I wanted to ask here was: What would next steps be now that we are in a place where we can standup a production build and run e2e tests on it in a browser using mocks?

Options (there may be more than this)

1. CI integration
2. Convert our current CLI based tests to browser based
3. Start creating net new tests (and if so are there any places in particular where we would like to start adding i.e. are there places that are currently lighter in tests than other areas?)

For the CI integration, this will potentially involve standing up/serving a build in CI. We can't use the preview site as they are MSW-mocked by default  (unless we can base the MSW feature flag on a runtime value, say a based on a cookie or something)

johncowen:
Before I forget there's at least one TODO to do here before we merge:

1. Take a look at moving the `.feature` files into the `/cypress` folder instead of being colocated along with the Route/View Component file. We only have one feature file at the minute so shouldn't be an issue

Just to note that ^ shouldn't block review but should be looked at before merge.

kleinfreund:
FYI, you will need to rebase/merge with master in order to run the tests using the updated GitHub Actions workflow since a job in it was renamed.

This is necessary because branch protection rules for which status checks are required are based on job names and so currently, this PR is waiting for a "Tests" job to have succeeded when it was previously called "Tests & build".

kleinfreund (approved):

johncowen:
Just wanted to mention before merging this, that there are 2 follow up PRs for this mentioned above - the last one should be ready to go, the previous might be a little out of date and may need a bit of massaging before PRing.

Apart from that I should be able to work on re-using our `development` service config now `development` doesn't import `production`. I can also start looking at bringing this all into CI.

PR #816, opened 2023-04-18
ci: updates workflows to use correct reference branch

kleinfreund:
Updates the create-gui-pr.yml workflow to use a correct reference branch instead of the branch associated with the workflow file used for execution (i.e. `master` in the case of the workflow being triggered via `workflow_run`). Without this changes, triggering this workflow from a release branch leads to running it in a default branch context instead.


johncowen (approved):
Quite a bit of change here, but guessing there's a bit of re-org here on purpose. Thanks for adding the comments!

PR #818, opened 2023-04-18
chore: Composable Service Config

johncowen:
This PR moves our service config to use composition rather than inheritance.

We we began using a service container approach it was a lot quicker to get where we wanted to get to by having a pretty linear "inheritance-like" reuse structure of config (a development file would import the production file)

As we've moved forwards and added distinct `main`/wiring roots, expanded our usage and new usecases have appeared (using the parts of the same service config for our Cypress tests/mocks). We've kind of begun to outgrow the inheritance approach

This PR therefore does the following:

- Changes service config to be composable via `...services(TOKENS)` functions. This means we can pick and choose sets of config we need rather than using an "all or nothing" approach.
- Adds decorators for service config via `decorates` i.e. `decorates: $.bootstrap`. This means we don't have to keep repeatedly overwriting services again due to the `production is imported by/before development which is imported by/before production which is imported by/before development` etc. approach.
- Removes as much vite aliasing as possible. There are no more magic `@/services` files that switch between prod and dev 🎉 . Originally this was handy before we had a top-level main/wiring root, but now I'm pretty sure the aliasing is unnecessary and moreover confusing.
- Added `useBootstrap`. This probably should have been there from when I added bootstrap originally, all Vue components should use `use*` helpers not `get(TOKEN...)`. I'm hoping that eventually in the not too distant future, that `get` will no longer be an export. This will help us to keep us to "_in Vue Components use `use*` helpers, otherwise use constructor injection_". I _think_ there's only one place left that doesn't follow this.
- Added runtime error if undefined arguments are found during service container building (this happens at container build time, i.e. at the very start of the app).

Closes https://github.com/kumahq/kuma-gui/issues/798

One thing I've noticed whilst doing this, this frees up `@/services/index` (a.k.a `@/services`), and whilst there is no rush whatsoever to do this, it means we can move `@/utilities` to `@/services` which to me makes much more sense.

Lastly, I also added a bunch of inline comments to help review.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,16 +1,25 @@
-import App from './app/App.vue'
-import { TOKENS, get } from '@/services'
+import { TOKENS as $, services as production } from '@/services/production'
+import { build } from '@/services/utils'
 
 async function mountVueApplication() {
-  const env = get(TOKENS.env)
-  const kumaApi = get(TOKENS.api)
-  // During development setBaseUrl also optionally installs MSW mocking via MockKumaApi
-  kumaApi.setBaseUrl(env('KUMA_API_URL'))
-
-  const app = await get(TOKENS.app)(App)
+  const get = build(
+    // production service container configuration
+    production($),
+    // any DEV-time only service container configuration
+    import.meta.env.DEV
+      ? await (async () => {
+        const dev = await import('@/services/development')
+        return dev.services({
+          ...$,
+          ...dev.TOKENS,
+        })
+      })()
+      : [],
+  )
+  const app = await get($.app)((await import('./app/App.vue')).default)
   app.mount('#app')
 
-  get(TOKENS.bootstrap)()
+  get($.bootstrap)()
 }
 
 mountVueApplication()
--- a/src/services/utils.ts
+++ b/src/services/utils.ts
@@ -37,23 +37,64 @@ type DependencyDefinition = {
   arguments?: Array<RequiredToken | OptionalToken>
   shared?: boolean
   labels?: Token[]
+  decorates?: Token
 }
 export type ServiceDefinition = [Token, DependencyDefinition]
+export type ServiceConfigurator = <T extends Record<string, Token>>(app: T) => ServiceDefinition[]
 export type Alias<T extends (...args: never[]) => unknown> = (...rest: Parameters<T>) => ReturnType<T>
-
+export type Decorator<T extends TokenValue> = TokenType<T>
+export type ReturnDecorated<T extends TokenValue> = () => TokenType<T>
 export const container = createContainer()
 
 //
 export const merge = (...definitions: Array<ServiceDefinition[]>): ServiceDefinition[] => {
   return [...new Map([...definitions.flat()]).entries()]
 }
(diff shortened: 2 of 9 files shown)


src/main.ts
```diff
@@ -1,16 +1,25 @@
...
+    import.meta.env.DEV
+      ? await (async () => {
+        const dev = await import('@/services/development')
+        return dev.services({
+          ...$,
+          ...dev.TOKENS,
+        })
+      })()
+      : [],
+  )
```
johncowen, line 18: I'm pretty sure this gives us a reliable "don't add the code in here to a production build" wrapper, if someone could just give that a second set of eyes that'd be much appreciated.

It's done in this way, as a function wrapper rather than a single line, so we can add more to it in other contexts. It's a little wieldy so an improvement on this that achieves the same would be good also, no matter if not though its code that will hardly ever be changed and is only in one place.

(this was one usecase where previously we were relying on vite aliasing)

src/main.ts
```diff
@@ -1,16 +1,25 @@
-import App from './app/App.vue'
-import { TOKENS, get } from '@/services'
+import { TOKENS as $, services as production } from '@/services/production'
+import { build } from '@/services/utils'
 
 async function mountVueApplication() {
-  const env = get(TOKENS.env)
-  const kumaApi = get(TOKENS.api)
-  // During development setBaseUrl also optionally installs MSW mocking via MockKumaApi
-  kumaApi.setBaseUrl(env('KUMA_API_URL'))
```
johncowen, line 8: Pretty sure we've got rid of the need to do this here?

src/main.ts
```diff
@@ -1,16 +1,25 @@
...
-
-  const app = await get(TOKENS.app)(App)
+  const get = build(
```
johncowen, line 5: `get` is now returned by `build` and should (in the future) be the only place we should need to use it i.e. start up the application straight after we've built the service container.

src/services/utils.ts
```diff
@@ -37,23 +38,64 @@ type DependencyDefinition = {
...
+          service: (): ReturnDecorated<DecoratedToken> => () => get(TOKEN),
+        })
+
+        map.set(decoratedToken,
+          {
+            ...definition,
+            arguments: [
+              WRAPPER_TOKEN,
+              // adding further arguments is currently unsupported
+              // ...(definition.arguments || []),
```
johncowen, line 81: We don't need to do this right now, but if we do need this (decorators with arguments), this is where it would be done.

kleinfreund:
FYI, you will need to rebase/merge with master in order to run the tests using the updated GitHub Actions workflow since a job in it was renamed.

This is necessary because branch protection rules for which status checks are required are based on job names and so currently, this PR is waiting for a "Tests" job to have succeeded when it was previously called "Tests & build".

src/main.ts
kleinfreund: This should work. The combination of compiler-analizeable condition (i.e. the ternary involving `import.meta.env.DEV`) and a dynamic import should remove any trace of the development file in a non-dev build.

kleinfreund (approved):

johncowen:
Gotcha, thanks for the heads up!

PR #826, opened 2023-04-20
feat: reworks LabelList component

kleinfreund:
## Changes

Turns our `LabelList` into a `DefinitionList` component that’s paired with `DefinitionListItem`s. It is now more opinionated about its internal structure and enforces a term-and-details structure implemented using HTML `dl`, `dt`, and `dd`.

Turns all uses of `LabelList` into `DefinitionList` and turns custom definition list implementations from service and data plane proxy summaries into instances of `DefinitionList`.

## Screenshots

Data plane proxy details (_before_ isn’t using `LabelList`):

| Before | After |
|--------|-------|
| Screenshot from 2023-04-24 14-34-07 | Screenshot from 2023-04-24 14-33-53 |

Zone details:

| Before | After |
|--------|-------|
| Screenshot from 2023-04-24 14-35-34 | Screenshot from 2023-04-24 14-35-43 |

Zone insights:

| Before | After |
|--------|-------|
| Screenshot from 2023-04-24 14-36-57 | Screenshot from 2023-04-24 14-36-48 |

Mesh details:

| Before | After |
|--------|-------|
| Screenshot from 2023-04-24 14-38-11 | Screenshot from 2023-04-24 14-38-21 |


src/app/common/subscriptions/SubscriptionDetails.vue
```diff
@@ -1,59 +1,66 @@
...
 
-      <ul>
-        <li v-if="details.globalInstanceId">
-          <strong>Global Instance ID:</strong>
-          <span>{{ details.globalInstanceId }}</span>
-        </li>
+      <DefinitionList>
+        <DefinitionListItem
+          v-if="details.globalInstanceId"
+          term="Global instance ID"
```
johncowen, line 11: Do you think we'll ever need to include HTML in a `<dt>`, say an `<em>` or `<span>`, or maybe a `<a href>`. I can imagine a feature that term could have a little (i) icon, clicking on which would take you to some documentation, or show a tooltip or something, wdyt?

src/app/common/subscriptions/SubscriptionDetails.vue
```diff
@@ -1,59 +1,66 @@
 <template>
   <div>
-    <div v-if="details.globalInstanceId || details.connectTime || details.disconnectTime">
+    <div v-if="details.globalInstanceId || details.controlPlaneInstanceId || details.connectTime || details.disconnectTime">
```
johncowen, line 3: Is this a bug fix?

src/app/data-planes/components/DataPlaneDetails.spec.ts
```diff
@@ -33,7 +33,7 @@ describe('DataPlaneDetails', () => {
 
   test('has expected content and UI elements', async () => {
     const expectedOverviewStrings = ['DPP: backend', 'test-mesh', 'online', 'kuma.io/service', 'backend', 'kuma.io/protocol', 'http', 'envoy', '1.16.2', 'kumaDp', '1.0.7', 'coredns', '1.8.3']
-    const expectedInsightStrings = ['Connect time: February 17, 2021 at 7:33:36 AM', 'Control Plane Instance ID', 'Disconnect time: February 17, 2021 at 7:33:36 AM']
+    const expectedInsightStrings = ['Connect time: February 17, 2021 at 7:33:36', 'CP instance ID', 'Disconnect time: February 17, 2021 at 7:33:36']
```
johncowen, line 36: Looks like we lost the `AM` here?

src/app/policies/views/PolicyListView.vue
```diff
@@ -370,10 +366,10 @@ async function getEntity(selectedEntity: { mesh: string, path: string, name: str
       const selected = ['type', 'name', 'mesh']
 
       entity.value = getSome(item, selected)
-      QueryParameter.set('policy', entity.value.name)
+      QueryParameter.set('policy', selectedEntity.name)
       rawEntity.value = stripTimes(item)
     } else {
-      entity.value = {}
+      entity.value = null
```
johncowen, line 372: Could you let me know what this was for? Is it a bugfix or something else?

johncowen (commented):
I had a few more questions after looking at the code, not taken it for a spin yet

src/utilities/camelCaseToWords.ts
```diff
@@ -0,0 +1,8 @@
+export function camelCaseToWords(str: string): string {
+  const words = str
+    .split(/([A-Z][a-z]+)/)
+    .join(' ')
+    .replace(/\s+/g, ' ')
+    .trim()
+  return words.charAt(0).toUpperCase() + words.substring(1)
+}
```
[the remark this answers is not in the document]
kleinfreund: > Didn't we remove this kind of stuff in https://github.com/kumahq/kuma-gui/pull/579 ?

We decided not to alter resource names (e.g. policy types), but the purpose of _this_ utility is _much_ more general: we often display field-value pairs from API data in the UI and for fields, we use camelcase. Instead of displaying `globalInstanceId`, `responsesSent`, `type`, `name`, etc. we display `Global instance ID`, `Responses sent`, `Type`, `Name`, etc.

Using this approach, for now, is by far the most simple to meet this style.

> In any case (no pun intended!) any massaging should be done outside the DefinitionList components. For example in this instance what would happen if we wanted to show a definition list with a term that we needed to be displayed camelCased, that wouldn't be possible.

I think this component is a perfectly adequate place to handle this for the time being. The "what if" of displaying a field in its raw form is currently not a use case as far as I can tell and such a functionality could trivially be added to the component.

I don’t want to overcomplicate this for now.

src/app/common/subscriptions/SubscriptionDetails.vue
kleinfreund: I initially had implemented `DefinitionListItem` to take both a `term` and a `details` slot for the flexibility. I then realized that we never had a case where that was necessary for term: it was always a plain text node without any special-cased formatting. So then I simplified the component to accept a default slot instead of the two named slots because it really cuts down on markup used to use the component.

One default slot:

```vue
<DefinitionListItem term="Global instance ID">
  {{ details.globalInstanceId }}
</DefinitionListItem>
```

Two named slots:

```vue
<DefinitionListItem>
  <template #term>
    Global instance ID
  </template>

  <template #details>
    {{ details.globalInstanceId }}
  </template>
</DefinitionListItem>
```

I think the use case that additional stuff becomes necessary to be added is much more likely for the details part of the definition list so for now I prefer the simplification in use over the long-term flexibility.

src/app/common/subscriptions/SubscriptionDetails.vue
kleinfreund: Perhaps. Looking at the current implementation, all uses of these fields are guarded with `v-if`s indicating that they can be empty/unset. If so, the top-level condition is indeed missing this check. However, I’m not sure if this field _can_ be empty/unset. So it’s cautionary, but I haven’t seen this actually be broken in practice.

src/app/data-planes/components/DataPlaneDetails.spec.ts
kleinfreund: On purpose. This is one of the tests that broke with the two breaking changes in Node.js 18.13 and 18.15 where the space before `AM` or `PM` was changed. I’ve removed it here because asserting for the presence of `AM` as part of the string doesn’t improve the quality of the tests and removing it _does_ avoid it from breaking again once we upgrade past Node.js 18.15.

src/app/policies/views/PolicyListView.vue
kleinfreund: It’s just adopting the style of indicating "entity is not currently available" that we use in all the other 5 instances of list view component. It’s not a functional difference in this component as all template conditionals are currently also covered by checks against `isEmpty` (which technically is redundant information) and the change from `:selected-entity-name="entity.name"` to `:selected-entity-name="entity?.name"` is also not different because both will evaluate to `undefined`.

src/utilities/camelCaseToWords.ts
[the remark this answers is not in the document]
johncowen: I'm still looking at this PR in its entirety and thanks for answering my questions on here already! Super helpful

I wanted to note on this particular one that I feel pretty strongly about not using this `camelCaseToWords` utility, and definitely not inside the component.

In my eyes the correct tool here is an i18n utility i.e. `t(machineTerm)`. Or, if we don't want to use one of those right now for whatever reason, then a poor mans version i.e. a `humanWords[machineTerm] ?? machineTerm` map, like we've used elsewhere in lieu of an i18n utility.

Using a map for machine terms to human words means we can do things like `responsesSent` > `Successful Responses Sent`, `DELEGATED` > `Delegated Gateway` if/when we need to. (I think we are already in 'if' and 'when' territory)

Lastly, and to add to my reasoning here, this utility currently turns `mTLS` into `MTLS` which we don't want, and I would guess there are other things, and will be other things that will just end up looking wrong with or without us noticing. i.e. the `XMLHTTPRrequest` > `XML HTTP Request` already in the test should always be rendered as `XMLHTTPRequest`

If we add a i18n utility, it's done then, and we can just use it for these things and not have to worry about it or add unnecessary code (plus tests required for the unnecessary code) and error prone automatic conversion utilities.

Now to me feels like a good time to add a i18n utility.

I'm going to keep looking over this PR in the meantime, but if you could give this a bit of consideration that would be amazing.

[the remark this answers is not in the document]
kleinfreund: > In my eyes the correct tool here is an i18n utility i.e. t(machineTerm). Or, if we don't want to use one of those right now for whatever reason, then a poor mans version i.e. a humanWords[machineTerm] ?? machineTerm map, like we've used elsewhere in lieu of an i18n utility.

I simply don’t want to introduce and maintain a complete map of terms for all our API terms instead of a simple utility like `camelCaseToWords` that almost always works correctly.

I’m happy to add the ability to use a map (i.e. `TermLabels[props.term] ?? camelCaseToWords(props.term)`) so we can special-case, for example, the field `mtls` to be labeled as `mTLS` (which currently is displayed as "MTLS" and in this PR as "Mtls"), but by default, I’m going to keep this utility because it does a good job of showing human-readable labels in our UI.

This would also allow for use cases, should they arise, like you mentioned:

> Using a map for machine terms to human words means we can do things like responsesSent > Successful Responses Sent, DELEGATED > Delegated Gateway if/when we need to. (I think we are already in 'if' and 'when' territory)

---

> If we add a i18n utility, it's done then, and we can just use it for these things and not have to worry about it or add unnecessary code (plus tests required for the unnecessary code) and error prone automatic conversion utilities.
>
> Now to me feels like a good time to add a i18n utility.

I really don’t see a justified use case for a i18n utility here because as far as I can tell from current use cases and the near future, how we want to map API fields to UI labels is covered by `camelCaseToWords` in the vast majority of cases already.

As I said above, I’m happy to introduce the option to provide overrides for special-cased terms like `mTLS`, but those special cases are very much the exception right now. I’m happy to reconsider this approach separately when we have a concrete use case for deviating from API fields in a way that `camelCaseToWords` isn’t able to cover.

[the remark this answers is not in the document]
johncowen: Ok gotcha, I'm happy with `TermLabels[props.term] ?? camelCaseToWords(props.term)` as long as I understood correctly that you plan on doing this outside of the DefinitionList components. Oh wait, I just looked at the commit on the bottom here, I see its inside the DefinitionList component.

Ummm, if we don't want to add a 3rd party util right now, could we just make our own `t` function and use it from The Outside?

```javascript
// component
const { t } = useI18n()

<DefinitionListItem :term="t(`http.api.mTLS`)" />
// or
<DefinitionListItem :term="t(`http.api.${term}`)" />

// i18n.js : i.e. pseudo code `t`
const terms = {
   http: {
      api: {
        mTLS: 'mTLS'
      }
   }
}
const usei18n = () => {
  t: (term) => terms[term] ?? camelCaseToWords(term)
}
```

The above achieves lots of things, most notably:

1. Doesn't couple machine > human mapping (plus un-camelCasing) to any specific presentational component i.e. `DefinitionList` 
2. In keeping the mapping out of a specific component it means we can also use it when not using that single `DefinitionList` component.
3. Sets us up for if we ever want to add a i18n layer, we just switch out ours for the new one.
5. Keeps using your camelCaseToWords as an overridable default, but the code doing that is in one single place (considering we don't want it to live in the single `DefinitionList` component)

Most importantly that ^ seems to be a good compromise to me?

> I really don’t see a justified use case for a i18n utility

I'm guessing going to have to use a i18n helper in our application in the very near future, and I think we already are using one, and probably should be for things like pluralization and date formatting.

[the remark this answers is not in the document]
kleinfreund: > Ok gotcha, I'm happy with TermLabels[props.term] ?? camelCaseToWords(props.term) as long as I understood correctly that you plan on doing this outside of the DefinitionList components. Oh wait, I just looked at the commit on the bottom here, I see its inside the DefinitionList component.

Is there a reason why having it outside is currently needed or advantageous? Moving it to the outside would mean you have to always manually use and import the helper. That’s vastly more work than what would currently be required.

> 1. Doesn't couple machine > human mapping (plus un-camelCasing) to any specific presentational component i.e. DefinitionList

The job of this utility, to transform a machine-provided term into a human-readable label, is strictly presentational though. I think having it in the component is currently ideal.

> 2. In keeping the mapping out of a specific component it means we can also use it when not using that single DefinitionList component.

We can easily add a utility of the following form to provide access to this transformation.

```typescript
import { camelCaseToWords } from '@/utilities/camelCaseToWords'
import { TermLabels } from '@/utilities/TermLabels'

export function getTermLabel(term: string): string {
  return TermLabels[term] ?? camelCaseToWords(term)
}
```

> 3. Sets us up for if we ever want to add a i18n layer, we just switch out ours for the new one.

Sure, but since it’s not clear whether we will need or have one, this is work that doesn’t have to be done now.

> 4. Keeps using your camelCaseToWords as an overridable default, but the code doing that is in one single place (considering we don't want it to live in the single DefinitionList component)

See response to point 2.

---

To summarize, currently, adding a translation utility and externalizing this code from DefinitionList only creates more work for me without any _current_ benefits.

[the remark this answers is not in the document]
johncowen: > Is there a reason why having it outside is currently needed or advantageous?

I can think of a few yeah. It's "good practice". It's de-coupled. It sets us up for the future. Its reusable. It means we can use DefinitionList with a camelCased term if we need to.

> Moving it to the outside would mean you have to always manually use and import the helper.

Fair point I suppose, but thats fine right? You have to import DefinitionList also, why shouldn't I just use a normal HTML `dl` because I don't want to import that every time? I think lots of folks use a `$t` global seeing as its such a common util.

> We can easily add a utility of the following form to provide access to this transformation.

Could we rearrange this function slightly to follow `useI18n().t(term)` and then export it from `@/utilities`? This seems to be almost exactly what I was suggesting just named differently?

> Sure, but since it’s not clear whether we will need or have one, this is work that doesn’t have to be done now.

Ok gotcha, fair enough if you aren't clear, but its super clear to me. I also remember some meetings we've had offline where folks have been requesting I18n helper functionality, I would guess its clear to them also. But you're right it doesn't have to be done right now necessarily - if we use the same `t` interface, we can just switch out the util in one place later.

> To summarize, currently, adding a translation utility and externalizing this code from DefinitionList only creates more work for me without any current benefits.

It might create a tiny bit of very simple work for you now, but it means reduces the amount of work for everybody else from now on into the future - that is a clear benefit. By doing a small bit of planning/work upfront, you help everyone else. Next time I need to do some machine > human mapping I can be like "oh! Philipp added a lovely t function for me to do this with, amazing! that saved me a bit of work, I don't have repeat whatever he did in that DefinitionList"

In my mind its been more work for both of us discussing it (in this and other PRs, with the bugs and issues from those) than to just use a very well recognised design pattern which we are already using elsewhere.

If this doesn't convince you then I think we should just give it a bit of time to think about it, and we can maybe discuss in our sync in a couple of days.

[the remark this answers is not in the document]
kleinfreund: I’ve now changed `getTermLabel` to receive a `i18n` object in https://github.com/kumahq/kuma-gui/pull/826/commits/28f6778a5503e94deb555b8b63932c3441f68695. This is as far as I’m willing to stretch the scope of this PR beyond what’s necessary.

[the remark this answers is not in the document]
kleinfreund: Please also tell me whether that’s a blocking concern for you.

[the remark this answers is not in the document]
johncowen: Outcome of team sync: https://github.com/kumahq/kuma-gui/issues/843.

Approval for this PR incoming.

src/utilities/useI18n.ts
```diff
@@ -0,0 +1,10 @@
+export function useI18n(): Record<string, any> {
+  return {
+    // A map of (machine-readable) terms (e.g. fields that come from API data) to (human-readable) labels to used in the UI.
+    termLabels: {
+      mtls: 'mTLS',
+    },
+  }
+}
+
+export type I18n = ReturnType<typeof useI18n>
```
johncowen, line 10: Just a small note here (I noticed this was amended since I last looked at this PR)

Preferably this would return a pseudo-standard `t('json.path.term')` function instead of just a JSON blob.

`t('json.path.term')` to me says "this is text/copy" rather than just `json.path.term` which to me says "this is data"

We make and document these decisions in the MADR though

johncowen (approved):
LGTM, sorry for the delay but thanks for the catchup

src/utilities/useI18n.ts
kleinfreund: Until we’ll decide on how to proceed with matters of internationalization, I’ll leave this as-is.

johncowen: Yep no prob

PR #830, opened 2023-04-24
feat: adds detail views to zones

### LINKED ISSUE #805: Make sure there's a details page for everything that exists in a list view
### Description

This should work with matching breadcrumbs. These pages should exist even if the content is no different from the "item" view from the list view.

We should first have a look at all the list views and see where we're missing details page.

kleinfreund:
**chore: adds basic zone ingress and egress mocks**

**chore: changes the default pagination size from 12 to 50**

Changes the `PAGE_SIZE_DEFAULT` from 12 to 50 and uses it consistently in all list views.

**refactor: adds routes for zone detail views**

Adds routes for zone, zone ingress, and zone egress detail views.

**feat: adds detail views to zones**

Adds a dedicated detail view to the three zone resource pages. All other resource pages (services, gateways, data plane proxies, and policies) already have a dedicated detail view.

Note: I’ll be adding links to the detail views in a more structured manner in a follow-up PR in which I’m refactoring (read: normalizing) the way we generate table data from API data and generally how we feed our standard DataOverview component.

Resolves #805.


johncowen (commented):
I was wondering where I can click to get to the new detail pages? I was expecting by clicking the zone here for example:

Screenshot 2023-04-25 at 17 28 31

but that takes me to the old style "list with detail page"

lmk!

johncowen:
Oh gotcha! 🤦

So for the minute then I'm looking here https://deploy-preview-830--kuma-gui.netlify.app/gui/zones/alarm-0

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -407,7 +405,7 @@ async function parseData(dataPlaneOverview: DataPlaneOverview) {
   let zoneRoute: RouteLocationNamedRaw | undefined
   if (zone !== undefined) {
     zoneRoute = {
-      name: 'zones',
+      name: 'zone-list-view',
       query: {
         ns: zone,
```
johncowen, line 410: Do you think we should get rid of the `ns` param name entirely and use `zone` or something (I think thats what we did for policies). `ns` to me means `namespace` which doesn't make sense to me here. lmk

kleinfreund:
> So for the minute then I'm looking here https://deploy-preview-830--kuma-gui.netlify.app/gui/zones/alarm-0

Yes, that’s right. It’s the same structure for all resource pages.

src/router/routes.ts
```diff
@@ -38,39 +38,84 @@ export default (store: Store<State>): RouteRecordRaw[] => {
...
-        selectedZoneName: route.query.zone,
-        offset: getLastNumberParameter(route.query.offset),
-      }),
-      component: () => import('@/app/zones/views/ZonesView.vue'),
+      children: [
+        {
+          path: '',
+          name: 'zone-list-view',
+          meta: {
+            title: 'Zone',
```
johncowen, line 46: Should this be `Zones`?

src/test-support/mocks/src/zoneingresses+insights/_.ts
```diff
@@ -0,0 +1,89 @@
...
+                version: '1.18.3',
+                build: '98c1c9e9a40804b93b074badad1cdf284b47d58b/1.18.3/Clean/RELEASE/BoringSSL',
+              },
+            },
+          },
+        ],
+      },
+    },
+  }
+}
```
johncowen, line 89: We should try to get used to making these as dynamic as possible when we add them (i.e. use `fake` to make all the values different). That way we get a good coverage of different types of data when developing. Obvs if there isn't time to do this for whatever reason then we can always circle back when there is. For example I would like to go through all the newly added ones at some point and make them way more dynamic. No action here if you don't want to do this now btw.

src/store/storeConfig.ts
```diff
@@ -629,7 +629,7 @@ export const storeConfig = (kumaApi: KumaApi): StoreOptions<State> => {
               title: lastSubscription.version.kumaCp.version,
               data: 1,
               route: {
-                name: 'zones',
+                name: 'zone-list-view',
```
johncowen, line 632: No action here really unless you want to, and kinda thinking out loud.  But, I quite like `zones` meaning "the collection of `zones`" due to it being plural. I know its a `-view` because its a view component, it probably can't be anything other than a `view` so "view" is a bit redundant, and I know its a `-list` (or `-table`, or `-grid` a.k.a. `-collection`) because of the pluralization.

---

It would be nice to eventually end up with route names being of the form `mesh.zones.show.config` seeing as the visiting this route is keyed by `mesh`, `zones`, `single-zone-identifier`, `config`, i.e. at the `mesh.zones.show.config` route you would see information specific to the `config` of the `single-zone` within all `zones` in a `mesh`. Chances are this is all nested/child-routed in some way.

Just to be clear, not necessary to amend this at all right now, just sharing thoughts.

johncowen (changes requested):
LGTM, I think theres possibly one tiny tweak to be done here from `Zone`> `Zones`, oh and possibly the `ns` removal aswell (depends what you think there), hence the "Request Changes".

Let me know on those ^ and I can approve this.

Couple more questions / potential follow up PRs:

- Would it be a good idea to start using path segments for tabs here instead of query params? Or maybe you'd prefer to do that application wide in one go?
- Could we add the missing mocks for zone stats, config etc. I know there are bits of mocks missing still but we should try and get it 100% filled out as we add/work on new things.

kleinfreund:
> Would it be a good idea to start using path segments for tabs here instead of query params? Or maybe you'd prefer to do that application wide in one go?

Since it’s unclear how general site navigation will evolve with regards to navigational tabs and our sidebar, I’d very much rather hold off to make any changes in that area.

> Could we add the missing mocks for zone stats, config etc. I know there are bits of mocks missing still but we should try and get it 100% filled out as we add/work on new things.

I’ll have a look at that.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: This will be handled in the follow-up PR I have prepared for this. It removes all routes using the `ns` param and targets detail view routes instead.

src/router/routes.ts
kleinfreund: Yes, indeed. I’ll change that in my follow-up PR as that deals mostly with the list views anyway.

johncowen: 👍

src/test-support/mocks/src/zoneingresses+insights/_.ts
kleinfreund: I’ve made some updates to the added mocks to make them more random.

src/store/storeConfig.ts
kleinfreund: > It would be nice to eventually end up with route names being of the form mesh.zones.show.config seeing as the visiting this route is keyed by mesh, zones, single-zone-identifier, config, i.e. at the mesh.zones.show.config route you would see information specific to the config of the single-zone within all zones in a mesh. Chances are this is all nested/child-routed in some way.

I think this specifically is to be avoided. It makes changing routing hierarchies unnecessarily hard because if you move a route to a hierarchically different position, you now end up with a misleading route name. And you can rename it, but that defeats the purpose of targeting routes by name and not by path: to allow you to move routes and change paths without breaking existing route targets in the process.

In that sense, these names are deliberately chosen to avoid ambiguity. And while it might seem redundant to call the list view for zones "zone list view" (etc.), it’s much nicer to search the project for references to it (one could improve this further by only using constants instead of string literals): a search for `zones` (for example) is much more likely to yield false positives when searching for route names. That’s nearly impossible when using the naming convention I established for routes.

kleinfreund:
This is ready for re-review.

src/app/data-planes/components/DataPlaneList.vue
johncowen: Gotcha

src/test-support/mocks/src/zoneingresses+insights/_.ts
johncowen: ty!

src/store/storeConfig.ts
johncowen: WeIl suppose that leads to another bugbear of mine. Generally I prefer to use the word `items` for variables instead of `zones` (you know an `item` is a `Zone` as its the Zone component, change the name of the component to Area and `item` can still be `item` as it now refers to an `Area`). This way the term `zones` is probably only ever used for string routeNames. But in the absence of that, I think its maybe more likely that the inclusion of a `-` makes the searching whilst excluding javascript varnames easier

> I think this specifically is to be avoided. It makes changing routing hierarchies unnecessarily hard because if you move a route to a hierarchically different position, you now end up with a misleading route name. And you can rename it, but that defeats the purpose of targeting routes by name and not by path: to allow you to move routes and change paths without breaking existing route targets in the process.

What happens when you have two routes using the same view i.e. say a service detail page on its own "service-detail-view" and another service detail page alongside the service listing (say in a drawer as a nested route). They can't both be named "service-detail-view" can they? One of them is inside `services.` and one isn't. I also think that possibly the idea of "relative route names" would help with what you explain there.

Anyway I can see there is a large difference of opinion here so glad I asked so I know but happy to leave this one here for now at least.

(just incase, nothing to be done here, just trading ideas)

johncowen:
Did you say you were gonna do this here:

> Could we add the missing mocks for zone stats, config [(also the ingress/egress ones)]

src/store/storeConfig.ts
kleinfreund: > What happens when you have two routes using the same view i.e. say a service detail page on its own "service-detail-view" and another service detail page alongside the service listing (say in a drawer as a nested route). They can't both be named "service-detail-view" can they? One of them is inside services. and one isn't. I also think that possibly the idea of "relative route names" would help with what you explain there.

That could be called `service-summary-view`, for example.

kleinfreund:
> Did you say you were gonna do this here:

I’ll look into this in follow-up work because I don’t want to another another review cycle to this PR.

johncowen (approved):

Issue #820, opened 2023-04-19
ci: Run Cypress tests in CI


johncowen:
### Description

We are currently working on running all our View Component tests in a browser as e2e (see https://github.com/kumahq/kuma-gui/pull/803)

Once the test runner is complete/merged, it should be added to CI but will require https://github.com/kumahq/kuma-gui/issues/819 to be completed first.

In tandem, we can continue to expand our browser based, production build, e2e test coverage.

PR #832, opened 2023-04-24
chore(ci): Integrate e2e tests into CI

johncowen:
Integrates Cypress based e2e tests into CI.

There are several tasks done here that are semi-inter-related:

1. I forgot that when serving our `index.html` we need to serve all 404s as index.html also. i.e. check the file exists, if it does serve it, if not serve `index.html`. This PR adds this functionality to the vite preview config
2. Makes necessary changes to the `package.json` file for CI to use. The majority of the changes are to split or join commands in a way that makes CI as nice as possible.
3. Makes CI run linting, cli-testing, browser-testing in parallel instead of in serial
4. Removes CI commit message skip functionality which I _think_ is provided natively at a workflow level by GH actions. edit: I've since confirmed its provided natively.

Further notes:

I removed the majority of the GHA `name:`s and comments here. Controversial I know, but the reasoning being:

- If I need to look at the docs for native actions such as `actions/checkout` I know where I can find those as they are native GH actions.
- I know that `actions/cache` is doing "Caching" and that `actions/checkout` is doing "Checking out".
- I know that `yarn lint:lockfile` is doing "Lint lock file"

All in all, I found that the above things were adding noise to the code rather than helping. Moreover, now everything is parallel these things would be repeated an additional 3 or 4 times throughout the file on top of the amount of times that they were already being repeated making the code look a lot bigger/major than it actually was.

On the otherhand, I _did_ add a comment to explain that there is a branch protection rule on a job called `"Tests"`, something that isn't clear only from looking at the code.

For me this is the important thing, is this thing clear from just looking at the code?

Lastly, I'm not totally happy with the names of the yarn scripts ("What's 'preview' mean?. What's 'staging' mean?"), just due to how things have shaken out. I've gone with something I'm vaguely happy with for the moment. But I'm 99% sure that now we have all our necessary environments/tasks setup (I think?), we can probably rename some things to make them more self-explanatory. But any refactor/rename of these things can be a follow up.

~Currently waiting on https://github.com/kumahq/kuma-gui/pull/831~

Closes https://github.com/kumahq/kuma-gui/issues/820


.github/workflows/main.yml
```diff
@@ -10,73 +7,101 @@ on:
...
-        uses: actions/setup-node@v3
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
           node-version-file: '.nvmrc'
-
-      - name: Cache node_modules
-        # https://github.com/actions/cache
-        uses: actions/cache@v3
+      - uses: actions/cache@v3
```
kleinfreund, line 38: **Suggestion**: Let’s change this to only restore the cache. We restore and store the cache in `install-dependencies` and don’t expect any of the subsequent jobs to change its contents anyway and can save the store action.

```suggestion
      - uses: actions/cache/restore@v3
```

.github/workflows/main.yml
```diff
@@ -10,73 +7,101 @@ on:
...
-        run: yarn run lint:ts
+  cli-tests:
+    needs: [install-dependencies]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+      - uses: actions/cache@v3
```
kleinfreund, line 59: **Suggestion**: Let’s change this to only restore the cache. We restore and store the cache in `install-dependencies` and don’t expect any of the subsequent jobs to change its contents anyway and can save the store action.

```suggestion
      - uses: actions/cache/restore@v3
```

.github/workflows/main.yml
```diff
@@ -10,73 +7,101 @@ on:
...
-        run: yarn run test
+  browser-tests:
+    needs: [install-dependencies]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+      - uses: actions/cache@v3
```
kleinfreund, line 77: **Suggestion**: Let’s change this to only restore the cache. We restore and store the cache in `install-dependencies` and don’t expect any of the subsequent jobs to change its contents anyway and can save the store action.

```suggestion
      - uses: actions/cache/restore@v3
```

.github/workflows/main.yml
```diff
@@ -10,73 +7,101 @@ on:
...
+          yarn run test
 
-      - name: Test
-        run: yarn run test
+  browser-tests:
+    needs: [install-dependencies]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
```
kleinfreund, line 74: **Question**: Can you check if Cypress needs Node.js to run? Maybe we don’t need this for browser tests.

.github/workflows/main.yml
```diff
@@ -1,7 +1,4 @@
 name: 'Tests'
```
kleinfreund, line 1: **Suggestion**: Seeing that you added a comment about branch protection rules, here is another gotcha, that we should probably document:

This workflow name is the trigger for the create-gui-pr.yml workflow:

```
  workflow_run:
    workflows: ['Tests']
```

(Yes, it’s very unfortunate how GitHub Actions uses _job_ names, _step_ IDs, and _workflow_ names as identifying references.)

.github/workflows/main.yml
```diff
@@ -10,73 +7,101 @@ on:
...
+            **/node_modules
+            /home/runner/.cache
+          key: ${{ runner.os }}-node-modules-${{ hashFiles('**/yarn.lock') }}
+      - run: |
+          yarn build:staging
+      - run: |
+          yarn preview:built &
+          yarn run test:browser
+
+  post-checks:
```
kleinfreund, line 90: **Suggestion**: Instead of this job, why don’t we simply add the three jobs to the branch protection rule? Then we could remove this entire job. It only needs those jobs to be named.

.github/workflows/main.yml
johncowen: Ah! Nice spot!

Ok umm, I was considering changing `Tests` to `tests` also so it looks like a machine identifier rather than a human arbitrary name, wdyt?

.github/workflows/main.yml
johncowen: I would think it does, but in any case we need node to build and run the preview build that we test against which we bootup in this job also.

.github/workflows/main.yml
johncowen: Because doing that is in a web gui rather than in code.

We can't avoid adding a single branch protection rule in a web gui as far as I know. But if we are changing the number of these from now and in the future, it means its all tracked in git in code with history etc etc.

I'm still fiddling with this unfortunately, and I think its possible its just getting it right. If for whatever reason I find its not possible then we'll have to resort to adding these via a web gui.

.github/workflows/main.yml
johncowen: nice, much better yeah. I'm going to apply these shortly

.github/workflows/main.yml
kleinfreund: Hm. The thing is, technically, a workflow also has an ID (I think it’s just the file name without `.yml`, e.g. `main` for this one). Maybe we can make this less painful by using a prefix convention like the following:

- Workflows are always referenced by name. Workflows must be named with a `workflow-` prefix (e.g. `workflow-tests`).
- Jobs are always referenced by ID (i.e. the name of the key under `jobs:`) were possible (e.g. in `needs: [install-dependencies]`). Were _not_ possible (e.g. in branch protection rules), jobs are referenced by name. Jobs must be named with a `job-` prefix (e.g. `job-tests`).
- Steps are always referenced by ID. Steps are not prefixed as accessing them in workflows already includes a `steps` key (e.g. `steps.node-modules-cache.outputs.cache-hit`).

johncowen: Ah, just double checking one thing, so:

https://github.com/kumahq/kuma-gui/blob/master/.github/workflows/main.yml#L1

is a completely different thing to:

https://github.com/kumahq/kuma-gui/blob/master/.github/workflows/main.yml#L48

? i.e. The strings here just happen to use the same "Tests" string, they are not really related in any way.

If thats the case, I would like to change the workflow name to something like

`name: main`

and the branch protection name to something like:

`name: main-success`

`name: tests` and `name: tests-success` would also be fine.

This way the identifiers that link things together are sufficiently generic enough not to have to keep changing them if we change the contents/purpose of the things that are linked. They also look like identifiers rather than `'` wrapped strings.

I'm not super keen on redundant prefixes, i.e. I know its a workflow as its in a workflow array and I know its a job as its in the jobs object

```
workflow_run:
    workflows: [main]
---
jobs:
  install-dependencies:
```

But all in all I/we are getting a little bike-sheddy so I'm not particular precious about any of this.

johncowen:
Just to note I want to keep this broken test in here until I get an approval so I can see/test what happens with an approval but a failing branch protection. So post approval there will be additional commit to fix the breaking test introduced on purpose here https://github.com/kumahq/kuma-gui/pull/832/commits/29751ee7e541bafc06dc7a5dd95b67f3c9b6acf4. I verified in advance that fixing that test make the branch protection go green.

johncowen:
OK sorry this PR was a bit noisy, but I think I'm done with this (apart from https://github.com/kumahq/kuma-gui/pull/832#issuecomment-1523185405) so just needing another review/approval here.

kleinfreund (approved):

PR #844, opened 2023-04-26
refactor: streamlines generation of table data for list views

kleinfreund:
**refactor: renames zone list views**

Renames the existing zone components for list views to follow the established `*ListView` convention.

**refactor: streamlines generation of table data for list views**

Streamlines how we generate table data for list views that’s fed to `KTable` via `DataOverview`. Every instance of `DataOverview` already receives a standard `tableData` prop. All consumers of `DataOverview` now generate this table data from API data using a `transformToTableData` function that transforms a raw API resource into a table row format. All these transformations now also have a complete type definition.

Removes the “Details” column from data plane table as we don’t use that approach to linking to detail views anywhere else and because we now have one established way of doing so (via linked cells in the “Name” column).

**refactor: moves general state props out of definition list**

Uses `StatusInfo` around one `DefinitionList` instance that was using state props which allowed for the removal of the state props from `DefinitionList` altogether.


src/app/common/DataOverview.vue
```diff
@@ -78,55 +78,55 @@
             <TagList :tags="rowValue" />
           </template>
 
-          <template #name="{ row, rowValue }">
+          <template #entity="{ rowValue, row }">
```
kleinfreund, line 81: **Comment**: For every use of `DataOverview`, I now made sure that the table row data includes an `entity` object which is the raw resource object. Paired with it, there is also always a `detailViewRoute` which is a route location for the entity.

That allows every list view to easily link to a corresponding detail view by providing `detailViewRoute` in the table data.

For other linked cells, I couldn’t think of a better way than the slotted content that follows below (i.e. one for mesh, service, and zone).

src/app/common/DataOverview.vue
```diff
@@ -173,28 +173,6 @@
               size="20"
             />
           </template>
-
-          <template
```
kleinfreund, line 177: **Comment**: For the time being, I’m removing the details column as we’re only using it for data plane proxies.

We might consider re-introducing this for all list views, but let’s discuss this separately.

johncowen: Cool yeah sounds good, we should add that to our next GUI sync

src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -114,13 +114,8 @@ async function loadData(offset: number, dppParams: DataPlaneOverviewParameters =
...
 
-    if (Array.isArray(items) && items.length > 0) {
-      dataPlaneOverviews.value = items
-      nextUrl.value = next
-    } else {
-      dataPlaneOverviews.value = []
-      nextUrl.value = null
-    }
+    nextUrl.value = next
+    dataPlaneOverviews.value = items ?? []
```
johncowen, line 118: Nice little simplification. Just out of interest, apart from when the HTTP request throws an error, is there any eventuality where items isn't an array?

src/app/policies/views/PolicyListView.vue
```diff
@@ -243,142 +225,99 @@ watch(() => route.params.mesh, function () {
...
-})
-watch(() => route.query.ns, function () {
-  isLoading.value = true
-  isEmpty.value = false
-  entityIsLoading.value = true
-  entityIsEmpty.value = false
-  entityHasError.value = false
-  tableDataIsEmpty.value = false
-  error.value = null
-
```
johncowen, line 265: Very glad all this is going!

src/app/common/DataOverview.vue
```diff
@@ -391,9 +366,12 @@ function getCellAttributes({ headerKey }: any): Record<string, string> {
...
 
-function getRowAttributes({ name }: any): Record<string, string> {
-  const entityName = props.selectedEntityName || props.tableData.data[0].name
-  const className = name === entityName ? 'is-selected' : ''
+function getRowAttributes(row: any): Record<string, string> {
+  if (!row || !row.entity) {
+    return {}
+  }
+
+  const className = row.entity.name === selectedEntityName.value ? 'is-selected' : ''
```
johncowen, line 374: Little side question: Do you have any theories on how comes GH can't syntax highlight this code?

johncowen (commented):
Wow quite a lot of change here 😅 , but going from what you are doing that's probably unavoidable.

I left a few questions, and a couple of things you'd maybe change/amend here but generally I think this looks good, hopefully maybe a couple of amends here followed by a quick approval.

There's a good deal of deletion of things I didn't like in here, and any improvement in how we render our tables from JSON data is a good thing, its always felt like such a lot of code to me just to render JSON in a table.

Don't take this the wrong way, as I really like what you've done here. But I think longer term we should probably get rid of the DataOverview component and split it up into smaller more reusable/composable parts. I think it does far too much as one component and is difficult to use when we want to tailor a table view slightly (like say if we don't want to render a refresh button for a specific table). Whether we do that or not is probably quite a way off, so if this PR makes things easier in the meantime then thats great, and we may not even ever do that.

I think one thing I don't like in this very positive PR is that we have types for DataObjects and then also a type for TableRows for each individual DataObject which longer term feels a little overkill (I can see why you've done it here, so I'm all good with that right now)

Anyway, again wanted to share thoughts rather than say what we should or shouldn't do.

Generally all looks good to me, lemme know on those few questions 👏 oh! and thanks for leaving some inline comments, those where handy.

src/app/common/DefinitionList.vue
```diff
@@ -1,41 +1,9 @@
...
-
-  <ErrorBlock v-else-if="props.hasError" />
-
-  <EmptyBlock v-else-if="props.isEmpty" />
-
-  <dl
-    v-else
-    class="definition-list"
-  >
+  <dl class="definition-list">
```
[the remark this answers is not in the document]
kleinfreund: There are indeed no more uses of DefinitionList where these props are provided. There used to be two remaining instances: this one and one in the ZoneListView (formerly ZonesView) which was removed in https://github.com/kumahq/kuma-gui/pull/830.

src/app/common/__snapshots__/DataOverview.spec.ts.snap
```diff
@@ -57,7 +57,6 @@ exports[`DataOverview.vue renders additional scoped slot 1`] = `
   <tbody>
     
     <tr
-      class="is-selected"
```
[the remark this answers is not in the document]
kleinfreund: At first I was trying to fix the test to ensure this value remains but I realized that it was unintentionally there:

In the current implementation of DataOverview, when `props.selectedEntityName` is not provided, it will be the empty string. Then, in `getRowAttributes`, `entityName` is set to `props.selectedEntityName || props.tableData.data[0].name` which evaluates to `undefined` in this test as the table data doesn’t have a `name` property. And again since the table data doesn’t have a `name` property, the `name` from `getRowAttributes`’s first argument is _also_ `undefined` and so `name === entityName ? 'is-selected' : ''` evaluates to `'is-selected'` when that shouldn’t be the case.

That being said, I made some tweaks to how this state is produced to ensure that, as long as there is a `name` property`, a default row can always be marked as selected.

src/app/data-planes/views/DataPlaneListView.vue
kleinfreund: As far as I can tell, in general, our APIs prefer omitting a field when it’s empty (e.g. a field would be missing instead of being `[]`).

However, it seems `items` as returned by API endpoints for collection resources, when empty, is actually provided as `null` instead of `[]` or being omitted. Since we always use `items ?? []`, we would be fine in the case of `null` being provided _or_ the field being omitted. I wonder if we should mark the field as optional, too, in the event that it is actually omitted, but I’m not aware of that actually being possible.

(Personally I would prefer empty lists to be `[]` always and never `null` or omitted, but that’s not what we do)

cypress/support/step_definitions/index.ts
```diff
@@ -91,7 +91,9 @@ Then(/^the "(.*)" element[s]? exist[s]? ([0-9]*) time[s]?$/, async (selector: st
 Then(/^the "(.*)" element(s)? contain[s]?$/, async (selector: string, multiple = '', table: DataTable) => {
   const rows = table.rows()
   if (multiple === 's') {
-    $(selector).each((el, i) => {
+    const tableHeads = $(selector).should('have.length', rows.length)
+
+    tableHeads.each((el, i) => {
```
johncowen, line 96: Oh sorry I meant use this existing step:

https://github.com/kumahq/kuma-gui/blob/93a255521efe545f50aadd97177d7ec9bfc2b5f6/features/mesh/gateways/Index.feature#L45

If we add this code you've added here it means I can't just check the first few elements of a selector, i.e. it does too much, plus we already have a step that tests length / times occurring.

src/app/common/DataOverview.vue
kleinfreund: I think this had to do with the presence of HTML comments that used three dashes (e.g. `<!--- test --->`) like so (it actually reproduces here, too):

```vue
<!--- test --->
<p>Test</p>
```

I’ve removed them now and that fixes it.

src/app/data-planes/views/DataPlaneListView.vue
kleinfreund: Yeah, it would be a good idea to make this maybe a bit chaotic even so we catch missed edge cases more easily (e.g. allow them to be `omitted`, too, so a check explicitly against `!== null` would be incorrect).

src/app/data-planes/views/DataPlaneListView.spec.ts
```diff
@@ -222,16 +233,16 @@ describe('DataPlaneListView', () => {
...
       'Service',
       'Protocol',
       'Last Updated',
       'Kuma DP version',
-      'Details',
     ]
     const wrapper = await renderComponent()
 
     const tableHeads = wrapper.findAll('[data-testid="data-overview-table"] th')
+    expect(tableHeads.length).toBe(expectedColumnHeaders.length)
```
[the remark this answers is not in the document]
kleinfreund: I’ve added the `Then` clause for this as suggest. Leaving this test file as it is now then.

cypress/support/step_definitions/index.ts
kleinfreund: I think now I got it right

src/app/data-planes/views/DataPlaneListView.spec.ts
```diff
@@ -282,13 +293,11 @@ describe('DataPlaneListView', () => {
...
     expect(secondTableRow.element.classList.contains('is-selected')).toBe(false)
     expect(wrapper.find('[data-testid="data-plane-proxy-title"]').html()).toContain('fake-backend')
-    expect(window.location.search.includes('name=fake-backend'))
 
     // The click event listener for selecting a table row is unfortunately added to table cells instead of rows so we need to trigger a click on a cell instead of the row.
     await secondTableRow.find('td').trigger('click')
 
     expect(secondTableRow.element.classList.contains('is-selected')).toBe(true)
     expect(wrapper.find('[data-testid="data-plane-proxy-title"]').html()).toContain('fake-frontend')
-    expect(window.location.search.includes('name=fake-frontend'))
```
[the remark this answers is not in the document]
kleinfreund: Leaving this file as is now.

src/app/data-planes/views/DataPlaneListView.vue
johncowen: Looks like at least this endpoint returns an empty array, so if all of the endpoints are the same (🤞 I would assume so) I think we are good for the mocks at least, actually I'll check in with backend offline to confirm

Screenshot 2023-04-28 at 09 53 37

johncowen: > However, it seems items as returned by API endpoints for collection resources, when empty, is actually provided as null instead of [] or being omitted.

From my checking and confirming offline with the backend peeps it seems like that is not the case ^. An empty `items` property will always return `[]` (as opposed to `null` or omission), and if it doesn't its a bug.

Could you double check whatever it was that made you think different for me just in case I'm missing something?

kleinfreund: At least `meshes/default/service-insights` seems to return `null` for `items`. I’m not changing this for the time being even though it would be nice to get rid of the `| null` type. Right now, this is the safer bet.

johncowen: Yeah for sure, makes sense 👍 . Glad this came up though. Did you want to make an issue for `meshes/default/service-insights` in kuma and link it here?

kleinfreund: Issue here: https://github.com/kumahq/kuma/issues/6639

johncowen (approved):
Thanks for the tweaks here. LGTM!

PR #848, opened 2023-04-28
feat: teaches breadcrumbs to read chains of parent routes

kleinfreund:
**feat: teaches breadcrumbs to read chains of parent routes**

Teaches the breadcrumbs component to read a complete chain of parent routes instead of only ever just one route-parent route relationship. This allows us to display, for example, "Policies → CircuitBreaker → cb-1".

Ensures that the title of a parent route is constructed in the same way as the one for the current route (i.e. taking `getBreadcrumbTitle` and `breadcrumbTitleParam` into account).

**refactor: streamlines policy route definitions**

Renames the `policy` route to `policy-list-view`.

Replaces `route.meta.title` assignment via `props` callback with dedicated `getBreadcrumbTitle` logic to set the breadcrumb title and a use of the `updatePageTitle` action.

Nests policy routes the same way we do with other detail/list view pairs to avoid accidental mistakes in the routes’ `path` values.


johncowen:
Before looking at this, just wanted to ask what you thought of re-working things at somepoint to just put the "breadcrumbs" in the `*View.vue` file?

Whilst these things are called breadcrumbs and look like breadcrumbs, they are actually just sub navigation i.e. they don't dynamically record _where you have been_ rather _where you are_). I'm pretty sure they can be completely static i.e. hardcoded into the `*View.vue` file, i.e. if you are in the PolicyDetailView we know where you are and the "breadcrumbs" / subnav should be `Policies > policy.Type > policy.Name` no matter which page you come here from.

In doing that we could probably get rid of quite a lot of code around what we currently do to render breadcrumbs.

Anyway, just a question at the moment to get your opinion.

kleinfreund:
> Before looking at this, just wanted to ask what you thought of re-working things at somepoint to just put the "breadcrumbs" in the `*View.vue` file?
> 
> Whilst these things are called breadcrumbs and look like breadcrumbs, they are actually just sub navigation i.e. they don't dynamically record _where you have been_ rather _where you are_). I'm pretty sure they can be completely static i.e. hardcoded into the `*View.vue` file, i.e. if you are in the PolicyDetailView we know where you are and the "breadcrumbs" / subnav should be `Policies > policy.Type > policy.Name` no matter which page you come here from.
> 
> In doing that we could probably get rid of quite a lot of code around what we currently do to render breadcrumbs.
> 
> Anyway, just a question at the moment to get your opinion.

Perhaps moving some of this logic into *View components could work. Each *View component could just do something like this:

```vue
await store.dispatch('setBreadcrumbItems', items)
```

... where `items` would be specific to this view and could be arbitrarily complex.

However, inadvertently, you introduce duplicated logic with this: consider the views hierarchically belonging to "Meshes > $meshName". Those would always have to add those breadcrumb trails themselves.

An alternative to letting only the "last" (i.e. the one associated with the last entry in the breadcrumbs) view component handle breadcrumbs would be to render `<router-view />` for all routes that need a breadcrumb item: For example, the route that has `path: '/mesh/:mesh'` would also have `component: () => import('@/app/meshes/views/MeshAbstractView')` and that component would just render a `<router-view />` and could do something like this:

```vue
await store.dispatch('addBreadcrumbItems', meshBreadcrumbItems)
```

And so multiple view components could contribute to adding breadcrumb items. With this approach you would also need to clear this list on every navigation.

I think this makes handling breadcrumbs unnecessarily complex. The hierarchical information of which breadcrumbs we want is already in direct relation to our route definitions so I think it’s nice to actually define breadcrumb matters there.

Probably there is an easier way than referencing route names via `route.meta.parent`:

The `route.matched` property already gives us the current route chain. Provided all routes in the chain for which we want breadcrumb entries have the right meta properties set, we probably could just loop through them instead of resolving `parent` references like we do right now.

All in all, I think we can get this a lot simpler but I don’t think moving breadcrumb code into view components is doing that necessarily.

johncowen:
I was thinking something like:

```vue
<AppPage
  :breadcrumbs="[{
    to: {
       name: 'dataplane-list-view'
     },
     text: 'Dataplane Proxies'
},
  {
      to: {
         name: 'dataplane-detail-view'
       },
     text: dpp.name
  }]"
>
*View Content in here
</AppPage>
```

even 

```vue
<AppPage>
  <KBreadcrumbs
    :breadcrumbs="[{
      to: {
         name: 'dataplane-list-view'
       },
     text: 'Dataplane Proxies'
  },
  {
      to: {
         name: 'dataplane-detail-view'
       },
     text: dpp.name
  }]"
   />
*View Content in here
</AppPage>
```

But the first one means we can add a `nav` in the AppPage component application wide:

```vue
// AppPage.vue
<template>
      <nav
        aria-label="Breadcrumb"
      >
        <KBreadcrumbs
          :items="props.breadcrumbs"
        />
      </nav>
      <slot name="default"></slot>
</template>
```

It's pretty much all just declarative config, zero javascript.

Anyway, gonna take a look at the PR, thanks for the chat

johncowen:
> Each view component must be wrapped in another component that allows slotting AppBreadcrumbs or KBreadcrumbs respectively into the right place.

I don't this as a problem, in fact I'd see it as a benefit. I think its a good thing to have every Route / View Component wrapped in the same AppSpecific root component. There are probably other things we'd use it for.

> Each view component must declare all its breadcrumbs. This is inadvertently redundant because we would need to add any ancestor breadcrumb items to all leave node breadcrumbs.

I also don't see this as a problem. If we wanted to reduce the teensy-tiny bit of redundancy, nested AppPage's could just look at their parent (similar to your JS) but in our case seeing as our breadcrumbs are pretty shallow I'm not sure I'd bother. I think the important point is the breadcrumb itself (whether thats a single one or multiple) is a view concern and belongs in the view layer i.e. the template.

Anyway none of this is asking for any work here, this is all for conversation more than anything, thanks again for the chat!

johncowen:
P.S. Having looked very briefly at the changes here in  https://github.com/kumahq/kuma-gui/commit/0be6bb5c0b9ea7e7f9c08851dc742c7b22a675b9

> Defines a route record as participating in breadcrumb creation. Defaults to `false`.

For this kind of stuff ^, with the "breadcrumbs in the route template" approach, if you wanted breadcrumbs for a page, you'd fill in the breadcrumbs, and if you didn't you just wouldn't set them.

Again just for convo for future.

src/app/AppNavItem.vue
```diff
@@ -113,33 +113,24 @@ const isActive = computed(() => {
...
-  }
+  try {
+    return props.routeName && route.matched.some((matchedRoute) => {
+      const resolvedRoute = router.resolve(matchedRoute)
 
-  return props.routeName && currentRoute.matched.some((route) => props.routeName === route.name || props.routeName === route.redirect)
+      return props.routeName === resolvedRoute.name
+    })
+  } catch {
+    return false
```
johncowen, line 132: Is not having something visible in console here likely to be a gotcha if we make a mistake and are trying to figure out why something isn't working? I'm not sure if after this change a console error would still surface - if it does you can ignore this.

src/app/policies/views/PolicyListView.vue
```diff
@@ -311,7 +319,7 @@ async function loadEntity({ name, mesh, path }: { name?: string | undefined, mes
 
 function changePolicyType(item: {value: string}) {
   router.push({
-    name: 'policy',
+    name: 'policy-list-view',
```
johncowen, line 322: Just something I keep noticing, not important here at all, but GH this doesn't syntax highlight this either 🤔

src/router/routes.ts
```diff
@@ -110,7 +118,7 @@ export default (store: Store<State>): RouteRecordRaw[] => {
           name: 'zone-egress-detail-view',
           meta: {
             title: 'Zone Egress',
-            parent: 'zone-egress-list-view',
+            isBreadcrumb: true,
```
johncowen, line 121: Despite what I was saying in our more general chat, I think this is definitely in improvement on the previous code.

src/router/routes.ts
```diff
@@ -151,7 +162,7 @@ export default (store: Store<State>): RouteRecordRaw[] => {
               name: 'gateway-detail-view',
               meta: {
                 title: 'Gateway',
-                parent: 'gateway-list-view',
+                isBreadcrumb: true,
                 breadcrumbTitleParam: 'dataPlane',
```
johncowen, line 166: I know its not part of this PR and more part of the ongoing breadcrumb chatter, but these 'specify the data property param to use for the breadcrumb text' type things would go away if we just did this in the template. (just to be clear no action requested here)

johncowen (commented):
I think this looks good, I had one question on whether an error gets surfaced or not, lmk!

src/app/AppNavItem.vue
kleinfreund: You’re right, we should probably log errors here to have better visibility on what’s likely an unintentional error case. I’ve added `console.error` statement.

johncowen (approved):

PR #393, opened 2022-09-30
feat: unify services list views

kleinfreund:
## Changes

Unifies the internal and external services list views into one view.

Replaces the few uses of Lodash utilities with native alternatives.

Pins msw versions while waiting for the fix for https://github.com/mswjs/msw/issues/1411.

Signed-off-by: Philipp Rudloff <philipp.rudloff@konghq.com>

Resolves #383.

## Screenshots

**Before**:

image

**After**:

image

PR #869, opened 2023-05-02
feat: adds Zone creation flow

kleinfreund:
Adds dedicated WizardTitleBar and FormProgressBar components for future re-use in other wizards.

Changes the top-level layout so that new style wizards can claim all space and not show either sidebar or top bar.

Adds the ZoneUpdateView component and the corresponding routes and controls in the zone list view for creating and deleting zones. The functionality for accessing the zone creation flow is enabled using the `VITE_ZONE_CREATION_FLOW` environment variable which, for now, is only enabled in development mode (like a feature flag).

Resolves #393.


kleinfreund:
**Comment**: Locally and in CI, I’m running a lot into the error that the GitHub button component causes sometimes:

```
Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.
```

lahabana:
image

Should we collapse 3 and 4?
Something like:

3. Set-up secret
Your token is: 
... (current token with copy option) 

You should add it as a kubernetes secret:
...

package.json
```diff
@@ -16,7 +16,7 @@
     "build:ci": "yarn run build --mode preview -c ./vite.config.development.ts",
     "build:gui": "mkdir gui && mv dist/* gui && mv gui dist/gui && mv dist/gui/mockServiceWorker.js dist/",
     "build:preview": "yarn run build:ci && yarn build:gui && cp _redirects dist/_redirects",
-    "build:staging": "yarn build && yarn build:gui",
+    "build:staging": "VITE_ZONE_CREATION_FLOW=enabled yarn build && yarn build:gui",
```
kleinfreund, line 19: **Comment**: Not sure if there is a better place to put this.

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,137 @@
...
+  zone: {zoneName}
+  kdsGlobalAddress: {globalKdsAddress}
+  secrets:
+    - Env: KUMA_MULTIZONE_ZONE_KDS_AUTH_CP_TOKEN_INLINE
+      Secret: cp-token
+      Key: token
+ingress:
+  enabled: {zoneIngressEnabled}
+egress:
+  enabled: {zoneEgressEnabled}`,
```
kleinfreund, line 100: **Comment**: A downside of putting config templates like this into i18n is that you can’t do more fine-grained JavaScript logic to _add_ `ingress.enabled` only when the setting is enabled, you can just flip the boolean.

features/zones/Index.feature
```diff
@@ -0,0 +1,54 @@
...
+      | environment-universal-radio-button  | [data-testid="environment-universal-radio-button"]  |
+      | environment-kubernetes-radio-button | [data-testid="environment-kubernetes-radio-button"] |
+      | ingress-input-switch                | [data-testid="ingress-input-switch"]                |
+      | egress-input-switch                 | [data-testid="egress-input-switch"]                 |
+    When I visit the "/zones/create-zone" URL
+
+  Scenario: The form shows only the initial elements
+    Then the "$name-input" element exists
+    Then the "$create-zone-button" element exists
+    Then the "$create-zone-button" element is disabled
```
johncowen, line 16: Not super important but if the the "$create-zone-button" is disabled, that means it exists right? Also, is there anything that would make the `$name-input` not exist? i.e. Are we just testing Vue here?

Just as a note, when I was moving some of the old CLI tests I wanted to replicate everything that was there, even the things that when I was doing it I was thinking "this is just testing Vue".

Moving forwards I would like to try to think about only testing things we own and getting as much value as possible out of the tests by choosing what to test and what _not_ to test and getting a good balance between the two. Like these two 'exists' checks here, I'm not totally sure if there is any code of ours directly in the component that would cause these to fail? They don't seem to be using any logic at all to render/not-render them.

Anyway as I said not super important, but just wanted to communicate that here.

johncowen (commented):
I took a look at the code and left a bunch of comments, some more important than others.

I also took it for a quick spin and generally it looks super good! I noticed that when you save the zone it seems to make 2 GET requests to the zone after saving, should it just make the one?

I'm still having a little play with the feature but wanted to submit this review so we can progress with it.

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,137 @@
...
+      confirmModalErrorText: 'An unexpected error occurred',
+    },
+  },
+}
+
+function translate(path: string, placeholders: Record<string, string> = {}): string {
+  let value = get(messages, path)
+
+  if (value === undefined) {
+    throw new Error(`[useI18n]: Could not find translation for path “${path}”.`)
```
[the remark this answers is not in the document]
johncowen: I'm also happy with a hard error

cypress/support/step_definitions/index.ts
```diff
@@ -15,18 +15,27 @@ Before(() => {
   urls = new Map()
   useServer()
 })
-type Cy = typeof cy;
-const $ = (selector: string): ReturnType<Cy['get']> => {
+
+const $ = (selector: string) => {
```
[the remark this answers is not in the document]
johncowen: Amazing, thankyou!

src/app/zones/views/ZoneListView.vue
```diff
@@ -30,20 +45,44 @@
...
+      modal-id="delete-zone-modal"
+      :action-button-text="i18n.messages.zones.delete.confirmModalProceedText"
+      :title="i18n.messages.zones.delete.confirmModalTitle"
+      @cancel="toggleDeleteModal"
+      @delete="handleDelete"
+    >
+      <template #body-content>
+        <p>{{ i18n.messages.zones.delete.confirmModalText1 }}</p>
+
+        <p>{{ i18n.messages.zones.delete.confirmModalText2 }}</p>
```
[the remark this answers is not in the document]
ValeryG: it might be already that it does. Take a look at: 
image

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,137 @@
...
+      confirmModalText1: 'Are you sure you want to delete the Zone {zoneName}?',
+      confirmModalText2: 'This action cannot be reversed.',
+      confirmModalProceedText: 'Yes, delete',
+      confirmModalTitle: 'Delete Zone',
+      confirmModalErrorText: 'An unexpected error occurred',
+    },
+  },
+}
+
+function translate(path: string, placeholders: Record<string, string> = {}): string {
```
ValeryG, line 118: why are we inventing translate function. could we use `core/i18n`, I think we should. And make it better if it doesn't handle the cases you guys need.

src/app/zones/components/ZoneCreateKubernetesInstructions.vue
```diff
@@ -0,0 +1,179 @@
+<template>
+  <div>
+    <h3>1. {{ i18n.messages.zones.form.kubernetes.prerequisitesTitle }}</h3>
```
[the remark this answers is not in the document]
ValeryG: yes, please, let's not only use t('path.to.key') , but let'
s consider to use `core/i18n` package https://github.com/Kong/public-ui-components/blob/main/packages/core/i18n/README.md and make it inline with other apps as well as the tools tech writers are using for defining those strings. And let's make this package better if need be instead gf having each app implement its own.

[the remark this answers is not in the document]
johncowen: hey @ValeryG 👋 

Quick TLDR on this one is, we will be using the `core/i18n` package very soon, in fact it cannot come sooner really!

Longer form: The plan as it was a few hours before the i18n stuff was added here was: In order to be able to merge this PR and then add `core/i18n` without one PR holding up the other, to follow the exact same interfaces in this PR as `core/i18n` i.e. `t('path.to.key')`. That way when the `core/i18n` comes in the next few PRs it will "Just Work"(tm). What we _don't want to do_ is start using object notation for i18n.

The locale files will in all likelihood be external also.

From the other comments you've written here (the markdown one is super interesting!) you and I (and Adam) seem to be on the exact same page 🙌

.env.development
```diff
@@ -0,0 +1 @@
+VITE_ZONE_CREATION_FLOW=enabled
```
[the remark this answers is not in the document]
kleinfreund: I’m avoiding "true" and "false" for .env entries on purpose because usually, .env file entry values are parsed as string and you’d need to check against `'true'` and `'false'` anyway. To not suggest that this will contain a true boolean, I’m using "enabled".

In any case, it works the same so I’m leaving this as is.

kleinfreund:
> Staging is supposed to be as close to production as we can get. If anything, we should be testing that production users cannot see the Zone Creation Flow. But obvs we still need to test the Zone Creation Flow.

> Not totally sure whether this is just gonna be a "pretend that Kuma in production has this enabled". Lemme think about this whilst we look at the other stuff here, its kinda of like we could do with a version where in production this is always enabled.

I rather avoid inverting this flag to "on by default" because it’s riskier. Forgetting to turn it off could leak this into a production build which can’t happen by accident if it’s "off by default".

Not sure what the alternative to setting it in the "build:staging" script is. Setting `KUMA_ZONE_CREATION_FLOW: 'enabled',` in `src/services/e2e.ts` doesn’t seem to work which I guess is owed to the fact that `build:staging` runs the production build using production.ts.

And I really do want to test this flow here because it’s where the code is (and where it can be broken).

Ideally, we could turn this flag on on a per-test basis so we could test both sides: enabled and disabled.

cypress/support/step_definitions/index.ts
```diff
@@ -58,6 +67,28 @@ When('I wait for {int} milliseconds/ms', function (ms: number) {
...
+          options.force = true
+        }
+
+        cy.wrap($el).click(options)
+      })
+
+      break
+    }
+  }
+})
```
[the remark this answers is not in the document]
kleinfreund: Always using `{ force: true }` is problematic because you risk accidental passes of commands that shouldn’t pass because they’re _actually_ currently occluded (e.g. by a modal dialog) or invisible (e.g. hidden via `v-show` or something like that).

When interacting with UI components whose `input` elements are visually hidden so some custom UI can be built that goes beyond what, for example, `input[type="checkbox"]` offers, you basically have two options when `click()`ing them using Cypress:

- `cy.get('input').click({ force: true })`
- Get the associated label element and click that as it’s unlikely to ever be hidden

Now the second option is more complex because finding the `label` for an `input` again has two options: the label is a parent or the label is associated to the input with a `for` attribute matching the input’s `id` attribute. We do use both approaches in Kongponents and so would need to check for both.

Conditionally using `{ force: true }` based on a class name seems more reliable even though it could of course break if that class name changes. That however should be marked a breaking change and so shouldn’t pose a risk.

src/app/zones/views/ZoneListView.vue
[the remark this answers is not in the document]
kleinfreund: Note that this PR is only adding a minimal effort interface for the i18n layer while not integrating with any dedicated library for internationalization, yet. So as the utilities present in this PR stand, they can’t do that.

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,137 @@
...
+
+type I18nMessages = Record<string, any>
+
+type I18nUtilities = {
+  t: typeof translate
+}
+
+export type I18n = { messages: I18nMessages } & I18nUtilities
+
+const messages: I18nMessages = {
```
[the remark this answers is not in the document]
kleinfreund: The i18n layer as present in this PR is just a minimal effort to allow us using it. It’s not the final implementation of it and messages aren’t supposed to be stored here.

That being said, JSON is perhaps an obstacle for us seeing as we’re storing multi-line strings here.

src/utilities/useI18n.ts
kleinfreund: See https://github.com/kumahq/kuma-gui/pull/869#discussion_r1189484381.

kleinfreund:
> If I understood you correctly, I think once we get a proper i18n lib under this we can probably do things like that.

The only way I can see this being improved is by splitting up the config into individual entries and making those placeholders in the top-level one. This we could to today but I would much rather have one entry for the full config.

Are you thinking of a different approach perhaps?

src/test-support/mocks/src/zones.ts
```diff
@@ -20,3 +30,16 @@ export default ({ fake }: EndpointDependencies): MockResponder => (_req) => {
...
+    headers: {},
+    body: {
+      name: req.body.name,
+      creationTime: '2020-07-22T19:37:28.442793+03:00',
+      modificationTime: '2020-07-22T19:37:28.442793+03:00',
+      enabled: false,
+      token: 'eyJhbGciOiJSUzI1NiIsImtpZCI6IjEiLCJ0eXAiOiJKV1QifQ.eyJab25lIjoid2VzdCIsIlNjb3BlIjpbImNwIl0sImV4cCI6MTY2OTU0NjkzOSwibmJmIjoxNjY2OTU0NjM5LCJpYXQiOjE2NjY5NTQ5MzksImp0aSI6IjZiYWYyYzkwLTBlODYtNGM2Mi05N2E3LTc4MzU4NTU4MzRiYyJ9.DJfA0M6uUfO4oytp8jHtzngiVggQWQR88YQxWVU1ujc0Zv-XStRDwvpdEoFGOzWVn4EUfI3gcv9qS2MxqIzQjJ83k5Jq85w4hkPyLGr-0jNS1UZF6yXz7lB_As8f91gMVHbRAoFuoybV5ndDtfYzwZknyzott7doxk-SjTes2GDvpg0-kFNGc4MBR2EprGl7YKO0vhFxQjln5AyCAhmAA7-PM7WRCzhmS-pUXacfZtP2VulWYhmTAuLPnkJrJN-ZWPkIpnV1MZmsgWbzTpnW-PhmCMQfD5m2im1c_3OlFwa9P9rZQQhdhbTp0ofMvW-cdCAcG_lOJI5j60cqPh2DGg',
+    },
+  }
+}
```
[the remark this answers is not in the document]
kleinfreund: How would you do it instead?

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,137 @@
...
+      zoneIngressEnabledLabel: 'Enabled',
+      zoneEgressLabel: 'Zone Egress',
+      zoneEgressEnabledLabel: 'Enabled',
+      connectZone: 'Connect Zone',
+      scanWaitTitle: 'Waiting for Zone to be connected …',
+      scanCompleteTitle: 'Done!',
+      scanCompleteDescription: 'The Zone “{name}” is now connected.',
+      scanCompleteButtonLabel: 'Go to Zone “{name}”',
+      scanErrorTitle: 'Could not connect Zone',
+      scanErrorDescription: 'We were unable to connect Zone.',
```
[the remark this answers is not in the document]
kleinfreund: I’ve made some changes to this now and nested more things were useful.

package.json
kleinfreund: I rather avoid inverting this flag to "on by default" because it’s riskier. Forgetting to turn it off could leak this into a production build which can’t happen by accident if it’s "off by default".

Not sure what the alternative to setting it in the "build:staging" script is. Setting `KUMA_ZONE_CREATION_FLOW: 'enabled',` in `src/services/e2e.ts` doesn’t seem to work which I guess is owed to the fact that `build:staging` runs the production build using production.ts.

And I really do want to test this flow here because it’s where the code is (and where it can be broken).

Ideally, we could turn this flag on on a per-test basis so we could test both sides: enabled and disabled.

src/app/zones/components/ZoneCreateKubernetesInstructions.vue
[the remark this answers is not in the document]
kleinfreund: I’ve changed the rudimentary i18n implementation to longer allow access to `i18n.messages` and used `i18n.t` throughout.

package.json
johncowen: Ideally the code would live in the place where we actually need it now (we could always just move it here when we want to enable it here). That would solve all of this.

I think another solution would be to add yet another run of tests on a dev build (i.e. `yarn dev`), which whilst isn't too much hassle to do, longer term it means we are complicating matters a little. 🤔 Actually, we could probably use a tag to only run the one set of tests in dev and exclude that same one set of tests from prod. In fact I think this is my favourite option considering everything/everyone. What do you think of adding a cucumber `@dev` tag (or similarly named) to your tests here, and then only running `@dev` tagged tests against `yarn dev` and anything else runs against `yarn build:staging`. The idea being, once this 'goes live' we just remove the `@dev` tag from the tests. We could even add a test testing the feature is not enabled (without the `@dev` tag of course)

Another option and perhaps my least favourite but easiest, is to just accept that we aren't really testing a production build, which is what we've done with this change here.

Just to be clear, my fave is the second option. What do you think of using a cucumber `@dev` tag?

cypress/support/step_definitions/index.ts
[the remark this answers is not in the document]
kleinfreund: We can’t add attributes to `label` elements that come via Kongponents components as usually attributes are bound to the `input` element (that’s a good thing). 

I’ve now combined the two step utilities and removed the `{ force: true }` from the select branch as it doesn’t need it.

.env.development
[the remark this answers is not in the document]
johncowen: > you’d need to check against 'true' and 'false' anyway.

We already use `"true"|"false"` to determine whether something is on/off, enabled/disabled, yes/no. Its just easier to stick to the same combo for everything.

Once we use less strings in `env` use pretty much just `booleans` we can make `env` return only a boolean-ish instead of only strings. Now we are adding a `i18n` layer that is closer to happening.

🤔 In fact I'd actually prefer `"1"|""` I think. The only env var we have using `"true"|"false"` currently is the MSW enabling one right?

Lemme know, I think I'm preferring to change this to `"1"|""` but I'm not totally sure if the MSW enabling var is the only one that currently uses `"true"|"false"`

cypress/support/step_definitions/index.ts
```diff
@@ -116,6 +147,15 @@ Then(/^the "(.*)" element( does| doesn't| don't)? exist[s]?$/, function (selecto
...
 })
+
+Then(/^the "(.*)" element[s]? (?:is|are) (enabled|disabled|checked|not checked)$/, (selector: string, word: string) => {
+  const prefix = ['enabled', 'not checked'].includes(word) ? 'not.' : ''
+  const property = ['disabled', 'enabled'].includes(word) ? 'disabled' : 'checked'
+  const chainer = `${prefix}be.${property}`
+
+  $(selector).should(chainer)
+})
+
```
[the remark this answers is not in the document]
kleinfreund: I’ve generalized this a bit now covering some likely word cases as suggested.

src/test-support/mocks/src/zones.ts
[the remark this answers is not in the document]
johncowen: I'd lose the functions and put the JSON in place, and not as important, but I'd maybe use `switch` if we come to add something like `DELETE` etc so we don't end up with a big run of 'else if`.

src/utilities/useGetGlobalKdsAddress.ts
```diff
@@ -0,0 +1,3 @@
+export function useGetGlobalKdsAddress() {
+  return () => 'grpcs://<global-kds-address>:5685'
+}
```
johncowen, line 3: This feels like it should be part of our API client, plus from our conversation yesterday and from what I understand of the feature (if I understood correctly), I think this might belong there. Wdyt?

kleinfreund: Not really. It’s not clear whether this bit of information will come from the API or whether we’ll be able to determine it by means of some convention based on information already known in the environment.

Long story short, we simply don’t know yet, how we will get this.

johncowen: Ah ok, I'd misunderstood then, I thought we said it should be something callable (i.e. a function) as we are likely to get this from the backend. I think it also just feels 'network-y'/'http' like, but no bother, thanks for the extra detail.

kleinfreund: I mean it _is_ callable; the service factory returns a function and that is called (hence the slightly awkward name `useGetGlobalKdsAddress`).

package.json
johncowen: > This is to degree less convenient that I’m not willing to consider it (again).

😆  I'd guessed that, hence the second option being my fave.

> It seems like quite some work to add the logic for running against two different builds in CI.

Only running certain tests based on a tag can be done via env vars and the functionality is built into cucumber already. Pointing cypress to a different build is also an env var.

> Another option and perhaps my least favourite but easiest, is to just accept that we aren't really testing a production build, which is what we've done with this change here.

Sounds like you're not keen, so I think we should just go with the third option for the moment then. i.e. no action here if you are happy with this, thanks for the convo.

Oh I was going to ask it looks like we don't actually need `cross-env` is that right?

cypress/support/step_definitions/index.ts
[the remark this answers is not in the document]
johncowen: I'm curious as to why:

1. We can't use `n't` like other steps and which is how people usually talk/write.
2. We can't use `${prefix}be.${word}`

Also, am I reading this wrong or can I say `the "selector" element is "writeable"` and it translates that to `be.readonly`, still looking at this but I'm a bit 🤔  Would it be easier implement points 1 and 2 here ^?

features/zones/Index.feature
johncowen: > I’m removing cases of double-asserting like this.

not sure if you are gonna remove things here or not lemme know.

kleinfreund: I already have?

johncowen: Oh sorry was a bit difficult to tell on GH, thanks!

src/utilities/useGetGlobalKdsAddress.ts
johncowen: Yeah for sure, it is callable, and I'd understood that the reason for it needing to be callable was that we were gonna make a HTTP request with it in places to ask the backend for this info. 

> I thought we said it should be something callable (i.e. a function) as we are likely to get this from the backend

i.e. I wasn't saying it wasn't callable, I was saying we said it should be callable for the reason that we are likely to use it to get info from the backend, and that reason was the important point I was trying to make.

No matter, no action here.

cypress/support/step_definitions/index.ts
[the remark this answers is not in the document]
kleinfreund: > We can't use n't like other steps and which is how people usually talk/write.

We can, yes, but the possibility tree becomes significantly larger.

I guess your intention is actually to only allow actual attribute names for `word` (e.g. `disabled` but not `enabled`), is that correct?

> Also, am I reading this wrong or can I say the "selector" element is "writeable" and it translates that to be.readonly, 

That would be `'not.be.readonly'`.

cypress/support/step_definitions/index.ts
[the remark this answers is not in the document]
johncowen: Why don't we put the data-testid on the parent `radio-button-group` and then `data-testid=['ingress-input-switch'] label`? Makes for a better test also I think as a human clicks on the label:

```vue
          <div class="radio-button-group" data-testid="ingress-input-switch">
            <KInputSwitch
              id="zone-ingress-enabled"
              v-model="zoneIngressEnabled"
              
            >
              <template #label>
                {{ i18n.t('zones.form.zoneIngressEnabledLabel') }}
              </template>
            </KInputSwitch>
          </div>
        </div>
```

Would something like that work? Means we can get rid of the classname check in the step also

cypress/support/step_definitions/index.ts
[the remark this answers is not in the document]
johncowen: > That would be 'not.be.readonly'.

Ah yeah read it wrong 😆 

> I guess your intention is actually to only allow actual attribute names for word (e.g. disabled but not enabled), is that correct?

I think we should allow anything, ie "does the step have `n't`? then add `.not`, other wise just pass through the property". In the positive sense it would fail if it wasn't there, granted in the negative it wouldn't I suppose, but I'm kinda trying to avoid adding a big list of strings that will never be complete here just to check against. If you are super keen on having a big list of stuff then happy to leave this one aside.

I think my main thing is to try and keep the steps consistent, we use `n't` elsewhere so I'm like "can we keep this the same?", I'm strongest on that really as anything else we can change as we go. Is the `n't` absolute no-go?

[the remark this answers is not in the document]
kleinfreund: I’ve simplified the two utilities around these assertions and made them support the same style of assertions now.

cypress/support/step_definitions/index.ts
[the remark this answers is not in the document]
kleinfreund: I didn’t try it because I don’t think it’s a better approach. It would rely on us keeping up this structure which I think is less likely than the components in Kongponents changing theirs.

Also I think determining the labels is probably the most generic and robust way to do this safe from having a dedicated utility for switches (and possibly other components with that implementation detail, not a fan of the sound of that).

I’m gonna leave this as it is now.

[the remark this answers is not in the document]
johncowen: Ok sounds good, thanks for working through it with me, appreciate the extra effort here.

package.json
kleinfreund: I just don’t know how to add this tag-based dev build logic. We’d need to add support for a `@dev` tag (I assume the functionality being built into Cucumber refers to the ability to add custom tags not the presence of a `@dev` tag, yes?) and create two builds in two separate directories (I assume pointing Cypress to a different build refers to setting a different `KUMA_GUI_DIST_ROOT` for a dev CI build, yes?).

> Oh I was going to ask it looks like we don't actually need cross-env is that right?

_We_ technically don’t because our operating systems understand this syntax, but Windows doesn’t. cross-env abstracts this away so that one can run these commands on Windows systems. (Though that is currently broken through the use of `&&` which Windows also doesn’t understand.)

johncowen: > (I assume the functionality being built into Cucumber refers to the ability to add custom tags not the presence of a @dev tag, yes?)

yup

> We’d need to add support for a @dev tag

We just add `@dev` to the top of the test and then from what I remember we can specify which tags to run via env vars. Some deets here https://github.com/badeball/cypress-cucumber-preprocessor/blob/master/docs/tags.md

> (I assume pointing Cypress to a different build refers to setting a different KUMA_GUI_DIST_ROOT for a dev CI build, yes?).

Kinda, yup, I think its `KUMA_BASE_URL` (look in in `visit.ts`) i.e. we would run `yarn dev` and point Cypress to it using `KUMA_BASE_URL` (FYI: I made MSW auto disable whenever you point Cypress at it)

All in all this was a good chat, but I'm happy for us to leave as is for now if its not straight forward to do the `@dev` thing. I do think it would be a good "utility" for us to have in our testing toolkit though, so maybe something for the future.

> We technically don’t because our operating systems understand this syntax, but Windows doesn’t. cross-env abstracts this away so that one can run these commands on Windows systems. (Though that is currently broken through the use of && which Windows also doesn’t understand.)

I'd be happy to remove cross-env as a dependency if you also were, sounds like its just cruft. Maybe we make an issue for the removal?

Just to be clear especially considering all the conversation on this one, there's no request for change here unless you wanted to have a go at the `@dev` thing (even then we could do that in a PR at some point)

src/utilities/useI18n.ts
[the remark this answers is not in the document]
johncowen: Hey Val

> Longer form: The plan as it was a few hours before the i18n stuff was added here was: In order to be able to merge this PR and then add core/i18n in a separate PR without one PR holding up the other we should follow the exact same interfaces in this PR as core/i18n i.e. t('path.to.key'). 

I think the most important thing for us here was the `t` usage. Honestly I wouldn't have minded it if here the `t` usage didn't work at all, i.e. it just printed the `path.to.key` to the page, as long as we were using the same `t` shape as `core/i18n`. That way when the PR comes with the addition of `core/i18n` it all just works

I think we've just got a little ahead of ourselves more than anything, sorry for the confusion! If you have any more question or queries I'd probably give @lahabana a shout.

features/application/MainNavigation.feature
```diff
@@ -96,7 +96,7 @@ Feature: index
       | $main-nav .nav-item-service-list-view a      | Services           |
       | $main-nav .nav-item-gateway-list-view a      | Gateways           |
       | $main-nav .nav-item-data-plane-list-view a   | Data plane proxies |
-      # TODO: This should say Circuit Breakers
+      # TODO: This should say CircuitBreaker. It currently can’t because the mocks will randomly change which policies exist.
```
johncowen, line 99: Q: Just for future knowledge, can't we mock things out so we know the title of the first policy page? (not asking for this to be done here)

kleinfreund: Huh, this test failed now even though I added the right mock change and ran the test locally 😕

kleinfreund: Oh I think this is flaky. I got the Zones and CircuitBreaker one failing locally now and on second run they all passed.

kleinfreund: Okay, that‘s way too flaky, I’m undoing this again. Strange.

features/application/MainNavigation.feature
```diff
@@ -83,8 +83,16 @@ Feature: index
...
 
   Scenario Outline: Visiting the "<Title>" page
+    # Given the URL "/mesh-insights/default" responds with
+    #   """
+    #   body:
+    #     items:
+    #     - policies:
+    #         CircuitBreaker:
+    #           total: 1
+    #   """
```
johncowen, line 93: If you change this to:

```yaml
body:
  policies:
    CircuitBreaker:
      total: 0
    FaultInjection:
      total: 1
```

And then obvs:

```gherkin
      | $main-nav .nav-item-policies a               | FaultInjection |
```

Then this works fine (I chose to not choose the naturally first policy type)

There's a swallowed error pretty deep in the application that made this not obvious. I think we should make an effort to stop trying to cover up when things go wrong with the backend/binary. The GUI should act as a thin client, when things go wrong, as a user, I want to know about it, at the very least in the javascript error console.

Just in case to be clear here, things haven't gone wrong with the backend, but the above commented mocks here mocked out our mock backend to make it so things had gone wrong. Hence the test wouldn't pass (which is actually a good thing). Just it was harder than it should be to figure out what was wrong as the error was swallowed and we tried to cover up the problem in the frontend instead of just acting like a thin client and saying "hey the backend gave us something we weren't expecting"

I might add something to next weeks GUI sync to chat about this.

johncowen: P.S. to be fair we probably shouldn't make these changes here as they have nothing to do with the PR. If it was me I'd remove all this commenting/changes here and draft a new PR to specifically address this bit. But up to you 🤷

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,154 @@
...
+            command: 'kuma-cp run -c config.yaml',
+            description2: 'Next, run the following command to connect the Zone.',
+          },
+        },
+        kubernetes: {
+          prerequisites: {
+            title: 'Prerequisites',
+            step1Label: 'Kubernetes cluster',
+            step1LabelAddendum: 'with load balancer',
+            step1Description: 'Konnect is compatible with all distributions of Kubernetes, including Minikube, AKS, EKS, GKE and OpenShift.',
```
lahabana, line 63:
```suggestion
            step1Description: '{productName} is compatible with all distributions of Kubernetes, including Minikube, AKS, EKS, GKE and OpenShift.',
```

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,154 @@
...
+            step2Label: 'kubectl or oc access',
+            step2Description: 'You have kubectl or oc (if working with OpenShift) installed and configured to communicate to your Kubernetes TLS.',
```
lahabana, line 65:
```suggestion
            step2Description: 'You have kubectl or oc (if working with OpenShift) installed and configured to communicate to your Kubernetes cluster.',
```

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,154 @@
...
+            step3LinkTitle: 'Helm 3',
+            step3Tail: 'is installed.',
+          },
+          helm: {
+            title: 'Set-up Helm',
+            description: 'On your local machine, create a namespace in your Kubernetes cluster and pull down the kong Helm repo.',
```
lahabana, line 71:
```suggestion
            description: 'On your local machine, create a namespace in your Kubernetes cluster and pull down the Helm repo.',
```

src/utilities/useI18n.ts
```diff
@@ -1,10 +1,154 @@
...
+            step1Description: 'Create a namespace',
+            step1Command: 'kubectl create namespace kuma-system',
+            step2Description: 'Add the Kong charts repository',
```
lahabana, line 74:
```suggestion
            step2Description: 'Add the charts repository',
```

features/application/MainNavigation.feature
johncowen: Just for future anyone reading this, see above comment on this

kleinfreund:
@johncowen I’ve made some of the changes to the step implementations we discussed yesterday. I’ve also cleaned up the git commits in this PR so they’re now combined in one commit: https://github.com/kumahq/kuma-gui/pull/869/commits/e8d91eb898e7978fe185b963bb0a6bca40673589

Not sure if I forgot something.

Generally, this PR is now ready for another pass.

johncowen:
Amazing! Thanks for the ping, will take another look shortly. Did you manage to look at why it makes 2 GET requests to the same URL with the same response when you save the zone? Also I had another question on something I noticed which might be a kongponent thing:

Screenshot 2023-05-11 at 09 38 46

Should the word `Enabled` there ^ change when you click the switchers?

Also just wanted to say appreciate there was a lot of convo/asks here so I appreciate you working through it with me 🙌 . Would be great to get this in today

kleinfreund:
> Amazing! Thanks for the ping, will take another look shortly. Did you manage to look at why it makes 2 GET requests to the same URL with the same response when you save the zone?

Are you referring to the requests to the `/zones+insights/:name` endpoint?

Now I did. The reason was `EntityScanner` not `await`ing `props.loaderFunction`. Fixed in https://github.com/kumahq/kuma-gui/pull/869/commits/6a6e681a7da70417fa813ff2533f4d11770a68f5.

kleinfreund:
> Should the word `Enabled` there ^ change when you click the switchers?

It shouldn’t. When toggling a switch (or a similar control), its _state_ changes (e.g. a checkbox is checked or unchecked, a toggle switch is enabled or disabled [sometimes also called checked actually]), but its label mustn’t; otherwise, you signal _no change_:

- "Zone Ingress Enabled, toggle switch, checked"
- "Zone Ingress Disabled, toggle switch, unchecked" which essentially is equal to "Zone Ingress Enabled, toggle switch, checked"

cypress/support/step_definitions/index.ts
```diff
@@ -110,12 +145,24 @@ Then('the URL {string} was requested with', (url: string, yaml: string) => {
   })
 })
 
-Then(/^the "(.*)" element( does| doesn't| don't)? exist[s]?$/, function (selector: string, assertion: string) {
-  $(selector).should(`${(assertion || 'does').trim() !== 'does' ? 'not.' : ''}exist`)
+Then(/^the "(.*)" element[s]?( don't | doesn't | do | does | )exist[s]?$/, function (selector: string, assertion: string) {
```
johncowen, line 148: I think we said yesterday that we were going to keep this auxiliary verb thing to the minimum of options rather than let folks write this in many different ways.

I think I said yesterday that I probably shouldn't have included the `does` in the first place, and then we said we were going to just keep it at supporting `don't` and `doesn't`, giving us:

```
the "" element exists
the "" elements exist
the "" element doesn't exist
the "" elements don't exist
```

Which covers everything we might want to do with this and it reads nice and natural with the contraction.

cypress/support/step_definitions/index.ts
```diff
@@ -58,19 +67,45 @@ When('I wait for {int} milliseconds/ms', function (ms: number) {
   cy.wait(ms)
 })
 
-When(/^I "(.*)"(.*)? on the "(.*)" element$/, (event: string, object: string | number | undefined, selector: string) => {
+When(/^I "(.*)" the "(.*)" element(?: and select "(.*)".*)?$/, (event: string, selector: string, value?: string) => {
```
johncowen, line 70: Just to note here our discussion form yesterday (no action here just noting).

We discussed the fact that I found out that you can't `"click" on the select` and `then "click" on the option` with Cypress, so I originally went with a "select" verb so it you can just replace "click" with "select".

We chatted through this potentially being weird due to any differences between a native select and a custom select, where probably with a custom select you would be clicking to open and then clicking to select. but as and when that becomes something we need to do we can hide the difference in the step.

We also chatted through the tiny nit-like `on` removal which was there to make all triggers/events read nice i.e:

```
When I "click" the "button" element
When I "hover" the "button" element 
vs
When I "click" on the "button" element
When I "hover" on the "button" element
```
"hover on" reading nicer than just "hover"

Bit of a nit so I'm happy either way really, but thats why the `on` was originally there. No change necessary for approval here.

cypress/support/step_definitions/index.ts
```diff
@@ -58,19 +67,45 @@ When('I wait for {int} milliseconds/ms', function (ms: number) {
...
   switch (event) {
-    case 'select':
-      if (typeof object === 'undefined') {
-        throw new Error()
+    case 'check':
```
johncowen, line 72: Did we say we were gonna remove the "check" and just use "click"? Seeing as you "click" on the checkbox or in a lot of cases "click" on the label? Not sure, not super important this, just asking in case we did say that.

johncowen (commented):
I think this looks fine!

I left a couple more last (promise!) comments, I think its mainly the removal of too many auxiliary options in the steps, the rest I think is just to note things here.

One more note, as we spoke about yesterday. It's really important for me to keep the amount of steps to a minimum and make them as sufficiently re-usable as possible. It's also really important to try and maintain consistency in how they are written, the more natural they read/write the better. So whilst some things are kinda nitty, I'll always think carefully when we add or change a steps human words.

Let me know on those last comments, and then I can approve this now. Thanks for all the work here 👏

cypress/support/step_definitions/index.ts
kleinfreund: I don’t care much for the difference so I’ve removed it, making `event` no longer variable so I removed that, too.

cypress/support/step_definitions/index.ts
kleinfreund: I’ve removed these two words now.

cypress/support/step_definitions/index.ts
```diff
@@ -58,19 +67,38 @@ When('I wait for {int} milliseconds/ms', function (ms: number) {
...
-  switch (event) {
-    case 'select':
-      if (typeof object === 'undefined') {
-        throw new Error()
-      }
-      $(selector).select(parseInt(object.toString()), { force: true })
-      break
-    default:
-      $(selector).trigger(event, { force: true })
+When(/^I click the "(.*)" element(?: and select "(.*)")?$/, (selector: string, value?: string) => {
```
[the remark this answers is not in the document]
johncowen: Ok can we leave the "" in for the moment please? This PR is about adding a zone creation view, not discussing already existing patterns and I think we are getting very far away from the task here. If you want to change this then we should discuss separately and make a decision that way, otherwise all of this is needlessly holding up your PR, considering the original PR didn't have this change in it at first, it's these changes here that are holding it up.

If you want to discuss these little details further let me know. I'd be happy to chat through again.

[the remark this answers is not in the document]
kleinfreund: You suggested very strongly and explicitly to keep things as simple as possible when it comes to these step utilities and allow as little as possible variety in how they can be used. In this interest, I applied several of your suggestions beyond what I had initially presented in this PR.

So to say that me deviating from the core of this PR’s task is what’s holding up the PR is getting it backwards: you suggested this work.

Is it really necessary to block this PR over what you call a little detail that has absolutely no functional consequence and is really a matter of taste?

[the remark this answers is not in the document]
kleinfreund: I’m resolving this because it’s purely a matter of taste and was only changed on detailed suggestion.

package.json
kleinfreund: I’d like to implement this, but I don’t have the time, so I’m leaving this as it is.

johncowen (approved):
Ok I wanted to underline here that refactorings and suggestions for refactorings are welcome, but they are best/ideally kept out of large pieces of feature work like this. The smaller the PR for the feature work and the less unrelated changes it has the quicker any discussion on the PR can be resolved and subsequently get approved.

If you could keep that in mind when you are working on feature work that'd be great, I've noticed that these types of PRs quite often have little refactorings in them that tend to warrant discussion.

The feature here looks good to me and I would really like to get it in so we can integrate/use `t()`/`core/i18n` here.

PR #871, opened 2023-05-03
fix: not correctly resolving unnamed routes

kleinfreund:
Fixes a few edge cases where the breadcrumbs component couldn’t properly resolve unnamed routes. For some reason that’s unclear to me, routes that are definitely defined and loaded only resolve to a proper route location with a resolved `path` property (e.g. the actual mesh name instead of `:mesh`) if they have a name. For this reason, I’ve added a `name` property to all routes that didn’t have one, yet.

Furthermore, changes how breadcrumbs create route targets so that instead of `{ name: '...' }`, they now use `{ path: '...' }` as navigating to, for example, `{ name: 'zone-abstract-view' }` will **only** navigate to that particular route without further resolving children routes for any routes with empty path. However, this _does_ happen when navigating to paths which is what we want here. All in all, `router.resolve` has some surprising and confusing behavior that we address with these changes.

### DIFF
--- a/src/app/AppNavItem.vue
+++ b/src/app/AppNavItem.vue
@@ -38,14 +38,13 @@
 <script lang="ts" setup>
 import { datadogLogs } from '@datadog/browser-logs'
 import { computed, PropType } from 'vue'
-import { useRoute, useRouter, RouteLocationNamedRaw } from 'vue-router'
+import { useRoute, RouteLocationNamedRaw } from 'vue-router'
 
 import { useStore } from '@/store/store'
 import { datadogLogEvents } from '@/utilities/datadogLogEvents'
 import { get } from '@/utilities/get'
 
 const route = useRoute()
-const router = useRouter()
 const store = useStore()
 
 const props = defineProps({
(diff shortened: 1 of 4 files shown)


kleinfreund:
**Issue**: Turning this back into a draft as these changes introduce an issue with determining the active sidebar navigation item.

src/app/AppNavItem.vue
```diff
@@ -117,21 +122,13 @@ const isActive = computed(() => {
     return true
   }
 
-  const currentRouteSubpath = route.path.split('/')[2]
-  if (currentRouteSubpath === targetRoute.value.name) {
+  const hasAnchorRoute = props.anchorRouteName !== '' && route.matched.some((matchedRoute) => matchedRoute.name === props.anchorRouteName)
```
kleinfreund, line 125: **Comment**: Save from listing _all_ routes in `src/app/getNavItems.ts` that belong to a nav item, this is the simplest I could come up with.

It turns out we will already need to add names to routes that we don’t necessarily want to name because of https://github.com/vuejs/router/issues/1829. TL;DR: unnamed and unresolved routes (i.e. what `route.matched` will contain) can’t be resolved using `router.resolve` so naming them is necessary to get breadcrumb generation to work.

In that sense the only additional work is manually stating the anchor route when defining nav items.

PR #873, opened 2023-05-03
fix: prematurely accessing policy types

kleinfreund:
**fix: prematurely accessing policy types**

Fixes some premature accesses of the current policy type in `PolicyListView`. Some instances were replaces with `route.params.policyPath` (i.e. when specifically the policy path was required). Other instances received additional type guard checks.

**chore: renames some incorrect references to policy types**

Renames some instances where the policy (type) path (e.g. circuit-breakers) was incorrectly referred to as policy type (e.g. CircuitBreaker).

**test(e2e): switches to more accurate title test case**

### DIFF
--- a/features/application/Titles.feature
+++ b/features/application/Titles.feature
@@ -15,6 +15,4 @@ Feature: The HTML title is correct on each page
       | /mesh/default/services                  | Services           |
       | /mesh/default/gateways                  | Gateways           |
       | /mesh/default/data-planes               | Data plane proxies |
-      # TODO: This should say Circuit Breakers
-      | /mesh/default/policies/circuit-breakers | Manager            |
-      # | /mesh/default/policies/circuit-breakers | Circuit Breakers   |
+      | /mesh/default/policies/circuit-breakers | CircuitBreaker     |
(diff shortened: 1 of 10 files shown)


johncowen (approved):
LGTM. This and https://github.com/kumahq/kuma-gui/pull/871, makes me think even more that we should do something like:

```vue
// PolicyDetailView.vue
   <AppPage
    :breadcrumbs="[
      {
        to: {
          name: 'policy-list-view'
        },
        text: 'Policies'
      },
      {
        to: {
          name: 'policy-type-view',
          params: {
            type: policy.type
          }
        },
        text: t(`http.${policy.type}`)
      },
      {
        to: {
          name: 'policy-detail-view',
          params: {
            type: policy.type,
            name: policy.name
          }
        },
        text: policy.name
      }
    ]"
  >
    ViewContent
  </AppPage>
```

Similar to the convo in https://github.com/kumahq/kuma-gui/pull/848#issuecomment-1527490825

Also in relation to our chat yesterday, this would mean that titling and breadcrumbing would be moved out of `App.vue` and into `AppPage`. `AppPage` can then be used in View components that are nested children of `App.vue` or even in View components that are _not_ nested children of `App.vue`.

I also like that any changes to breadcrumbing or titling would be done in each separate view component (the View layer) rather than in a big global routing configuration (ie. the FrontController).

Just to be clear, I'm happy to approve this, but I think we need to start thinking about making it so we can make changes to the application in smaller less sprawly ways and I want to make sure I share places where I think I can see that we might be able to do that.

PR #881, opened 2023-05-05
refactor: simplifies EntityScanner control flow

kleinfreund:
Simplifies the control flow of EntityScanner to use fewer conditionals and instead relying on short-circuiting of mutually exclusive states (loading, error, complete). Also removes `props.shouldStart` as it was always set and so it being `true` can be implied.

Related to #869 (splitting up parts of it so review is easier).


johncowen (approved):
I think this is fine, there are no functionality changes here right, just a refactor?

I thought a bit about whether it's ok that `hasError` can be `true` at the same time as either `isRunning` or `isComplete` (unless I read it wrong) but I don't think it matters in the code.

kleinfreund:
> I thought a bit about whether it's ok that `hasError` can be `true` at the same time as either `isRunning` or `isComplete` (unless I read it wrong) but I don't think it matters in the code.

`isRunning` should always take precedence because it means another (or the first) scan has been triggered. In that scenario, we don’t know whether the scan will complete successfully nor do we know if an error will occur. Even if the previous scan resulted in an error, a new scan might discover a different result. In any case, showing the loading UI is the right call here.

As for `props.hasError`, this should only be shown if a scan isn’t currently running.

PR #895, opened 2023-05-11
fix(zones): zone egress not generally being available

### LINKED ISSUE #771: Zone Egress should be available regardless of being in multizone mode
### What happened?

A bug happened!

kleinfreund:
Fixes zone egress not being available in the navigation when in standalone mode.

Resolves #771.

### DIFF
--- a/features/application/MainNavigation.feature
+++ b/features/application/MainNavigation.feature
@@ -4,6 +4,23 @@ Feature: index
       | Alias         | Selector                         |
       | main-nav      | .app-sidebar                     |
 
+  Scenario Outline: The navigation shows Zone items depending on mode
+    Given the URL "/config" responds with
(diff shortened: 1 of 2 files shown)


johncowen (commented):
Partly in the interest of starting to backfill our tests as we go. Would you be ok adding a small feature/e2e test here for checking that when multizone is enabled/disabled the correct items show/don't show?

I think a feature test would be good as I'm not entirely sure how the `isMultizoneMode` gets all the way through into here, plus we are changing this soon I believe, so the having a feature test to update would ensure this keeps working once its changed. Lmk

kleinfreund:
I’ve added a test for both scenarios now.

johncowen (approved):
LGTM!

PR #899, opened 2023-05-15
chore: Adds @kong-ui-public/i18n

johncowen:
Adds `@kong-ui-public/i18n` as our i18n layer.

There are a couple of extra PRs to come after this I think, details are in the accompanying MADR. Mostly around moving to YAML but also some changes to `DefinitionListItem` to pull `t` usage out of that component.

I tried to make as few changes as possible to existing things whilst keeping things working as we move the old implementations to `@kong-ui-public/i18n`.

I may add some inline comments or more desc here at a later date.


src/locales/en-us/index.ts
```diff
@@ -0,0 +1,134 @@
...
+          description2: 'Add it as a Kubernetes secret.',
+          command: `echo "
+apiVersion: v1
+kind: Secret
+metadata:
+name: cp-token
+namespace: kuma-system
+type: Opaque
+data:
+token: {token}
```
kleinfreund, line 89: **Issue**: This needs to be indented one level.

```suggestion
  token: {token}
```

src/locales/en-us/index.ts
```diff
@@ -0,0 +1,134 @@
...
+zone: {zoneName}
+kdsGlobalAddress: {globalKdsAddress}
+secrets:
+  - Env: KUMA_MULTIZONE_ZONE_KDS_AUTH_CP_TOKEN_INLINE
+    Secret: cp-token
+    Key: token
+ingress:
+enabled: {zoneIngressEnabled}
+egress:
+enabled: {zoneEgressEnabled}`,
```
kleinfreund, line 107: **Issue**: Keys following `controlPlane:` need to be indented one level until and excluding `ingress`. Both `enabled` keys need to be indented one level.

```suggestion
  mode: zone
  zone: {zoneName}
  kdsGlobalAddress: {globalKdsAddress}
  secrets:
    - Env: KUMA_MULTIZONE_ZONE_KDS_AUTH_CP_TOKEN_INLINE
      Secret: cp-token
      Key: token
ingress:
  enabled: {zoneIngressEnabled}
egress:
  enabled: {zoneEgressEnabled}`,
```

kleinfreund (changes requested):

src/locales/en-us/index.ts
johncowen: This has already been fixed in the followup PR https://github.com/johncowen/kuma-gui/compare/chore/core-i18n...johncowen:kuma-gui:chore/i18n-yaml . It'll be easier to just leave this as is here.

src/services/utils.ts
```diff
@@ -8,6 +8,7 @@ import {
   OptionalToken,
   injected,
 } from 'brandi'
+import deepmerge from 'deepmerge'
```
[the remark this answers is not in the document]
johncowen: Oh good spot! I've moved this out of devDeps

src/locales/en-us/index.ts
kleinfreund: Fine by me.

src/locales/en-us/index.ts
kleinfreund: Fine by me. Make sure to de-indent the `ingress` and `egress` blocks one level in that changeset though.

src/services/i18n/I18n.ts
```diff
@@ -0,0 +1,40 @@
...
+export default <T extends I18nRecord>(strs: T) => {
+  const i18n = createI18n<typeof strs>('en-us', strs, true)
+  return {
+    ...i18n,
+    t: function (...rest: Parameters<typeof i18n['t']>) {
+      const key = rest[0]
+      try {
+        if (typeof get(strs, key) === 'undefined') {
+          if (key.startsWith('http.api.')) {
+            throw new I18nError(key)
```
kleinfreund, line 25: **Question**: Why is this handled via exception? To me, this looks like it could just be `return $fallbackMechanism`, no?

src/services/i18n/I18n.ts
```diff
@@ -0,0 +1,40 @@
...
+  [key: string]: I18nRecord | string
+}
+
+class I18nError extends Error {
+  get key() {
+    return this.message
+  }
+}
+
+export default <T extends I18nRecord>(strs: T) => {
```
kleinfreund, line 16: **Suggestion**: Could we call this `messages` like the parameter is called in `@kong-ui-public/i18n`?

```suggestion
export default <T extends I18nRecord>(messages: T) => {
```

src/services/i18n/I18n.ts
johncowen: I expected `core/i18n` to throw an error when a key wasn't found. It does throw an error but then it swallows that error and deals with it itself by just erroring to console, so we can't override that. Ideally via bubbling errors so thats how I continued to write it and I still prefer this.

I think apart from that we are probably going to be getting rid of this pretty soon anyway and just using `core/i18n` erroring and all our term > human mappings will be pre-built. This was just to support what we currently have until we do that.

johncowen:
All the "Requested Changes" are addresses/resolved in some way, let me know if there's anything else necessary here @kleinfreund

kleinfreund (approved):

PR #913, opened 2023-05-16
chore: YAML for i18n layer

johncowen:
Adds a YAML layer for our i18n strings.

---

I've also added the ability to tag YAML text as being markdown formatted, which may make things easier for folks in the future. We don't use this as yet but I'm guessing I will use it in the future, so I've left it in here for now (unless there are any concerns with doing that)

See MADR for more details

Additionally I've added a couple of amends to the Kuma YAML configs contained in the YAML i18n strings here. That have been spoken about previously. I'll inline comment them so its clear where they are.

I also noticed that having camelCase i18n property names can be problematic (I've added both forms of mTLS that we use in the API), so I would recommend lowercase i18n property names where possible, and I'm considering `toLowerCase()`ing all our paths that are passed to `t` at some point in the future to prevent this.

`jest` yaml loader required until we get rid of `jest,` at which point it can be deleted.


src/app/zones/locales/en-us/index.yaml
```diff
@@ -0,0 +1,105 @@
...
+          multizone:
+          zone:
+            name: {zoneName}
+            globalAddress: {globalKdsAddress}
+          kuma:
+          multizone:
+            zone:
+              kds:
+                auth:
+          cpTokenInline: {token}
```
kleinfreund, line 41: **Issue**: A few indentation fixes needed here as well. Note that I kept the `kuma` key even though that one will definitely not exist in Kuma; it just makes it easier to see where this part would be in other places.

See https://github.com/kumahq/kuma/blob/master/docs/generated/kuma-cp.md for reference.

```suggestion
          environment: universal
          mode: zone
          multizone:
            zone:
              name: {zoneName}
              globalAddress: {globalKdsAddress}
          kuma:
            multizone:
              zone:
                kds:
                  auth:
                    cpTokenInline: {token}
```

kleinfreund (changes requested):

vite.config.ts
```diff
@@ -34,6 +38,18 @@ export const config: UserConfigFn = ({ mode }) => {
         },
       }),
       svgLoader(),
+      yamlLoader(
```
kleinfreund, line 41: **Issue** (minor): Not exactly sure where this happens, but in this PR, the YAML rendered on page now has an extra empty line at the end which shouldn’t be there.

The rest looks good to me.

kleinfreund (changes requested):

vite.config.ts
johncowen: Oh good spot, lemme take a look

johncowen: trimmed things

kleinfreund (approved):

Issue #738, opened 2023-03-20
Remove global text-decoration: underline for links


johncowen:
### Description

We should probably default to no underline (and then probably underline on hover/focus instead) If we decide a specific piece of text should be underlined, we then add it using a specific CSS selector rather than using a global 'underline all the links'

PR #922, opened 2023-05-16
chore: Moves mesh sub items out of main nav and into tabs

johncowen:
Removes several side menu navigation items and moves them to "In Mesh Tabs". Probably best explained with a screengrab:

Screenshot 2023-05-16 at 17 40 08

Services, Gateways, Data Plane Proxies and Policies are now sub tabs of Meshes.

---

The longer story: Whilst this PR looks relatively large, its mainly as a result of splitting our biiiiig routes file into separate files within the applications 'modules' (right now in `@/app/*`), I'll try to bullet point an overview of what changed:

1. Every 'module' now has an `index.ts` and `routes.ts` file.
2. Some `routes.ts` files have 2 sets of routes one for list page and one for detail pages, these are exported as `items` and `item` properties and are split like this to later compose `item` onto `items` within the `routes.ts` file. This gives us the ability to have the following path segment based deep links; `/service/:service-name` as a detail page, `/services/` as a list page and `/services/:service-name` as a combined listing/detail page, a.k.a a drawer. Currently the combined listing/detail routes are commented out.
3. Ideally we would have something to walk over route collections and combine them similar to as if they were in a tree-like structure. But as we don't have this yet, I've done this manually using `...routes` as 'placeholders'. In mind mind the tree-like-walker thing should be a TODO. At which point we can just combine all these by labelling them as `$.routes` and forget about it.
4. I moved the enabling of the zone creation flow routes to the wiring root i.e. `main.ts`. This means you now get a 404 instead of a redirect to home which makes more sense to me. I did this with a little `constant` trick that makes me think we should probably keep this DI util (I mentioned a while ago possible deciding between `constant()` and `constant:`)
5. Whilst doing the above I noticed that you can no longer have a zone called `create-zone`. If you tried to view the detail page for a zone called `create-zone` you would get the zone creation page. Therefore I changed the URL to be `/zones/-create`. By using an invalid DNS identifier for the path segment we can pretty much guarantee these kinds of things won't happen. There are other options (like maybe `create/zone`) but this is my preferred right now.
6. Several little test changes to reflect the above changes.

One thing I will note is that it took me quite a while to organize the routing configurations to get the breadcrumbs correct, for a long while I could get it to link to meshes properly (the URL would look literally like `/mesh/:mesh`). I don't know how I managed to get it to print it properly, but it seems to be fine now, but I think there is a degree of unnecessary nesting in the route configuration just to get the breadcrumbs correct. I would like to change how we do breadcrumbing, but again that is something I'm punting for the minute.

Oh lastly KTabs! I almost forgot. I managed to do this with the new additions to KTabs to allow us to have Tabs as links rather than clickable HTMLElements. But! I had to do a little bit of hackery still to get it to work. I also noticed that it seems to be impossible to dynamically display a collection of tabs i.e. loop through and array of tabs and dynamically print the tab/panel. This might be down to my poor Vue experience though.

All in all I'm relatively happy with what's here, and I tried to get by with using what we have already rather than adding/changing to much. But a few things have come up here that was should work in future.

I'll try to put some comments inline as this is getting bigger than the PR 😆 !


features/application/MainNavigation.feature
```diff
@@ -116,11 +64,7 @@ Feature: index
     Examples:
       | Selector                                     | Title              |
       | $main-nav .nav-item-home a                   | Overview           |
-      | $main-nav .nav-item-zone-list-view a         | Zones              |
+      | $main-nav .nav-item-zone-list-view a         | Zone CPs           |
```
johncowen, line 67: I kinda guessed this should say `Zone CPs` but lemme know if not and I can put back

features/mesh/Index.feature
```diff
@@ -1,9 +1,10 @@
 Feature: mesh / index
   Background:
     Given the CSS selectors
-      | Alias         | Selector                         |
-      | main-nav      | .app-sidebar                     |
-      | mesh-selector | [data-testid='mesh-selector']    |
+      | Alias           | Selector                      |
+      | main-nav        | .app-sidebar                  |
+      | mesh-selector   | [data-testid='mesh-selector'] |
+      | mesh-breadcrumb | .k-breadcrumbs:nth-child(2)   |
```
johncowen, line 7: We now use the breadcrumb to test that a mesh selection has worked correctly

src/app/data-planes/components/__snapshots__/DataPlaneDetails.spec.ts.snap
```diff
@@ -281,7 +281,7 @@ exports[`DataPlaneDetails matches snapshot 1`] = `
                               
                               <a
                                 class=""
-                                href="/mesh/default/services/backend"
+                                href="/mesh/default/service/backend"
```
johncowen, line 284: All of the snapshot changes are as a result of having detail pages at `services` as opposed to `services`

src/app/data-planes/routes.ts
```diff
@@ -0,0 +1,58 @@
...
+                    title: 'Data plane proxies',
+                  },
+                  props: (route) => ({
+                    selectedDppName: route.query.dpp,
+                    offset: getLastNumberParameter(route.query.offset),
+                  }),
+                  component: () => import('@/app/data-planes/views/DataPlaneListView.vue'),
+                  // children: [
+                  //   ...item(prefix),
+                  // ],
```
johncowen, line 48: These are ready for the "list plus detail view" a.k.a drawers

src/app/meshes/views/MeshView.vue
```diff
@@ -0,0 +1,149 @@
+<template>
+  <KTabs
+    :tabs="items"
+    :model-value="(items.find(item => (router.currentRoute?.value.name ?? '').toString().startsWith(item.hash)) ?? items[0]).hash"
+    @click="change"
+  >
+    <template #mesh-detail-view-anchor>
```
johncowen, line 7: I would love to know if there was a way to loop through the tabs and add the templates instead of having to hardcode the names in like this. I tried `v-for`s in various places but either got a double up tab bar or a big Vue error. I would imagine at some point we will need to dynamically add tabs based on something.

I have an idea for a PR on kongponents that should be backwards compatible if its not currently possible.

src/app/meshes/views/MeshView.vue
```diff
@@ -0,0 +1,149 @@
...
+    title: 'Policies',
+  },
+]
+const change = (e: MouseEvent) => {
+  const target = e.target as HTMLElement
+  if (target.nodeName.toLowerCase() !== 'a') {
+    e.preventDefault()
+    target.querySelector('a')?.click()
+  }
+}
```
johncowen, line 136: Hackery 😬 but I really really want to use KTabs

src/app/meshes/views/MeshView.vue
```diff
@@ -0,0 +1,149 @@
...
+.tab-link {
+  height: 100%;
+}
+.tab-link a {
+  text-decoration: none;
+  color: var(--KTabsColor)
+}
+li.active .tab-link a {
+  color: var(--KTabsActiveColor)
+}
```
johncowen, line 148: More hackery

src/app/zones/routes.ts
```diff
@@ -0,0 +1,139 @@
+import { getLastNumberParameter } from '@/router/getLastParameter'
+import type { RouteRecordRaw } from 'vue-router'
+export const actions = (): RouteRecordRaw[] => {
+  return [{
+    path: '-create',
```
johncowen, line 5: Hurrah for DNS! (for once 😆 )

src/services/production.ts
```diff
@@ -123,11 +147,64 @@ export const services: ServiceConfigurator<SupportedTokens> = ($) => [
...
+  [$.wizardRoutes, {
+    service: wizardRoutes,
+  }],
+  [$.onboardingRoutes, {
+    service: onboardingRoutes,
+  }],
+  [$.diagnosticsRoutes, {
+    service: diagnosticsRoutes,
+  }],
+
```
johncowen, line 207: A lot of this could be avoided if we make something to walk the routing config and merge it as if it was a tree of routes.

johncowen (commented):
Added some inlines to help review

kleinfreund:
**Changes to URL paths for detail views**:

I don’t think the URLs (e.g. `/mesh/${mesh}/services/${service}`) for detail views should change to singular form (e.g. `/mesh/${mesh}/service/${service}`).

The current URL structure, in my opinion, is good already (with the exception that `/mesh` should be `/meshes`). It allows for URL hacking and mirrors how the Kuma API endpoints are structured. Why do they need to change?

One aspect to this which we might get to is that we might add a kind of summary drawer to list views and that we want the state of that drawer (i.e. being visible or not) to be persisted in the URL. This raises the question of how that is done, but I don’t think it should make it necessary to break the current URL structure.

Can I ask you to please undo those changes in this PR and propose them separately? They’re not necessary for the changes made here (i.e. to use tabbed navigation for mesh modules) and I want the change to the URL structure to be discussed and agreed upon.

features/application/MainNavigation.feature
kleinfreund: Hm, probably, yes. In any case, we might not have these sidebar entries for long anyway.

src/app/zones/routes.ts
kleinfreund: > Whilst doing the above I noticed that you can no longer have a zone called create-zone. If you tried to view the detail page for a zone called create-zone you would get the zone creation page. Therefore I changed the URL to be /zones/-create. By using an invalid DNS identifier for the path segment we can pretty much guarantee these kinds of things won't happen. There are other options (like maybe create/zone) but this is my preferred right now.

I’m not really sure I’m following this here. How would this not have the exact same problem but for `-create` instead of `create-zone`?

src/app/AppMeshSelector.vue
```diff
@@ -51,9 +50,14 @@ function changeMesh(event: Event): void {
 
   store.dispatch('updateSelectedMesh', mesh)
 
-  const name = 'mesh' in route.params ? route.name as string : 'mesh-detail-view'
-
-  router.push({ name, params: { mesh } })
+  router.push(
+    {
+      name: 'mesh-overview-view',
```
kleinfreund, line 55: **Question**: Why did you remove the functionality to stay on the current page if the route is a mesh route?

johncowen:
We had a discussion offline in Slack about this a couple of days ago, where you asked the same question/query and I gave you the reasons behind this change and the explained how all opinions would be catered for with this change, I've not heard anything more from you since then until right now.

To give you those reasons/explanations again in shorter form:

- It's more future proof like this, and this is a future that we have designs for already.
- We shouldn't be expecting users to edit URLs, users will generally just click on navigation. Despite this URLs are still "hackable".
- If this becomes an issue we can add a redirect from `/service` to `/services`

Is there anything left after reading those ^ that doesn't address your concerns?

Additionally:

I'm against using query parameters for things like this as it leads to people stuffing route components with too many views making them overly complicated and also writing their own routing layer based around query parameters and I want to move away from both those things. This to me makes the current approach not "good already".

Seeing as we are changing all our navigation now, now feels like a good time to do this.

src/app/meshes/views/MeshView.vue
```diff
@@ -0,0 +1,149 @@
...
+    <template #policies>
+      <RouterView
+        v-slot="routerView"
+      >
+        <component
+          :is="routerView.Component"
+          :key="routerView.route.path"
+        />
+      </RouterView>
+    </template>
```
kleinfreund, line 101: **Suggestion**: You can make this dynamic like so:

```suggestion
    <template
      v-for="item in items"
      :key="`${item.hash}-anchor`"
      #[`${item.hash}-anchor`]
    >
      <router-link
        :to="{
          name: item.hash,
        }"
      >
        {{ item.title }}
      </router-link>
    </template>

    <template
      v-for="item in items"
      :key="item.hash"
      #[item.hash]
    >
      <RouterView v-slot="routerView">
        <component
          :is="routerView.Component"
          :key="routerView.route.path"
        />
      </RouterView>
    </template>
```

johncowen: Oh perfect! The `v-for` on the slot of course! ty!

src/app/meshes/views/MeshView.vue
```diff
@@ -0,0 +1,149 @@
+<template>
+  <KTabs
```
kleinfreund, line 2: **Issue**: As `KTabs` isn’t built with links in mind and so its styles don’t accommodate that use well: currently, the clickable area of the link is just the box of the text but not the padding box of the tab making only the text behave like a link.

We will need to override `KTabs` and make it so that the link takes up the whole space of the tab for it to work correctly.

Is this also what the `click` event handler on `KTabs` is for?

src/app/AppMeshSelector.vue
johncowen: Based on the meeting we had last week where John said it was more useful to land on the overview page everytime we select a Mesh (this was when I asked that we were loosing this by moving to a list mesh selection rather than a dropdown).

I believe we are removing this dropdown menu very soon anyway?

src/app/zones/routes.ts
johncowen: The vast majority of our "entity" names need to be DNS compatible. A `-` at the start of a name I believe isn't DNS compatible, which means we won't have an entity with this name.

src/app/AppMeshSelector.vue
kleinfreund: > I believe we are removing this dropdown menu very soon anyway?

That’s true.

kleinfreund:
@johncowen I’m still opposed to changing the URL structure like that. Why should the detail views have a different URL path than they do now? It works perfectly well and doesn’t prevent us from adding summary drawers. They would just have to be on a different URL path.

This is a blocker for me.

johncowen:
> Why should the detail views have a different URL path than they do now?

You also said: "One aspect to this which we might get to is that we might add a kind of summary drawer to list views and that we want the state of that drawer (i.e. being visible or not) to be persisted in the URL. This raises the question of how that is done, but I don’t think it should make it necessary to break the current URL structure."

> I don’t think it should make it necessary to break the current URL structure.

I'm against using query parameters for things like this as it leads to people stuffing route components with too many views making them overly complicated and also writing their own routing layer based around query parameters and I want to move away from both those things. This to me makes the current approach not "good already".

> It works perfectly well and doesn’t prevent us from adding summary drawers

No, here you are right. It doesn't _prevent_ us at all, but doing it like this is far more future proof and avoids the overly complicated things I mention above. Depends what sort of code base we want to work in I suppose and whether we want to be spending time writing things that should be dealt with by Vue's router.

---

> Is there anything left after reading those ^ that doesn't address your concerns?

Is there anything left that doesn't address your concerns?

You concerns as far as I understand them are:

1. The URLs are not hackable. : Adding a redirect from `/service` to `/services` addresses this.
2. It's good already. Thats an opinion which I disagree with. It currently works "more or less", but it is not a good way of doing this (reasons above)
3. Mirrors how the Kuma API endpoints are structured. Not totally sure how important this but we already don't mirror the Kuma API endpoints so I don't think this is that valid.

Let me know on that last question so I have all your concerns clear.

kleinfreund:
Okay, I’m definitely not convinced this was necessary and I would’ve really appreciated this to not happen as part of this PR, but I see this is not moving and I don’t want to argue it. Consider the matter of changing URL paths unblocked.

src/app/meshes/views/MeshView.vue
johncowen: 🤔 yeah kinda between a rock and a hard place here, not totally sure whats best long term, umm I'm gonna change things up to loop through the tabs dynamically, so I might take another look at this also, not totally sure.

src/app/meshes/views/MeshView.vue
johncowen: This has been changed up, thanks again!

src/app/meshes/views/MeshView.vue
johncowen: I took another look at this whilst doing the tab loop-de-loop thing. And I figured how I've left it right now is the least intrusive.

- The CSS I'm using isn't doing any overwrites and is only effecting slots which I control
- The CSS only affects colors, and the colors I'm using are the KTab variable colors, so thats pretty safe as an overwrite.
- There is one more overwrite for removing underlines, and I know there is an issue (https://github.com/kumahq/kuma-gui/issues/738) which will make this unnecessary pretty soon.

The JS approach I'm using is a little weird considering, but there are no overwrites and it's pretty acceptable - just that its a bit weird that we have to add this to do anchored tabs but no biggie really. I'm pretty sure that any changes to KTabs wouldn't break this unless explicitly stopped bubbling events.

src/app/meshes/views/MeshView.vue
johncowen: This has been amended following a suggestion during review, these are now dynamically generated with a couple of loops 🎉 

KTabs could still do with a tiny bit of massaging to make it work with hashes or with just normal path segments/anchors, but for the moment I got far enough so that I'm happy enough with whilst still continuing to use current KTabs which is what I wanted to do.

kleinfreund: (Just FYI, this was the PR that was meant to provide better support in KTabs for links: https://github.com/Kong/kongponents/pull/1251)

src/app/meshes/views/MeshView.vue
kleinfreund: In my opinion this is functionally broken. Buttons are not links and making this work correctly requires minimal overriding of Kongponents styles which is far less problematic than fake link activation triggered via `click` event. Below is what I just tried locally to make the clickable area as big as the tab:

In `src/app/meshes/views/MeshView.vue`:

- Add class `nav-tabs` to `KTabs`
- Add class `nav-tab-link` to `router-link`
- Add the following rule set to the scoped style block:

  ```scss
  .nav-tab-link {
    display: block;
    padding: var(--spacing-md);
  }
  ```

- Add the following unscoped style block:

  ```vue
  <style lang="scss">
  .nav-tabs .tab-item {
    padding: 0 !important;
  }
  </style>
  ```

johncowen: Not keen on the unscoped CSS block with an `!important` hiding in a component somewhere deep the app folder structure but 🤷 , as long as we don't use `.nav-tabs` elsewhere.

Oh actually:

Screenshot 2023-05-19 at 10 13 42

I suppose this could be fixed with a `.nav-tab > ul .tab-item`? But it makes me think I'd rather stick with standard DOM event approach for the moment just as we know for sure it doesn't affect anything else like the global `!important` might.

Lemme know. If you are still super strong on the global `!important` CSS rule then I could fix this up with a `>`. I'm still not that precious about either approach really.

kleinfreund: > I suppose this could be fixed with a .nav-tab > ul .tab-item?

This would indeed avoid that issue.

> But it makes me think I'd rather stick with standard DOM event approach for the moment just as we know for sure it doesn't affect anything else like the global !important might.

For what it’s worth, you could remove the `!important` by increasing the selector’s specificity sufficiently (e.g. `.nav-tabs > ul > .tab-item:not(.increase-specificity)` would do the trick).

I definitely want this addressed. Triggering a click via JavaScript isn’t enough to make the tab behave like a link.

johncowen: > Triggering a click via JavaScript isn’t enough to make the tab behave like a link.

What's the issue here?

johncowen: Good points, I switched to CSS but used a classname that is highly unlikely to be reused elsewhere.

kleinfreund (approved):

PR #923, opened 2023-05-17
feat: always shows resources in universal format

### LINKED ISSUE #696: Always display policies in Universal and offer an option to copy it as k8s
### Description

Who cares about kubernetes when you are using universal and vice-versa.

Let's just show the policies in the format of the global CP. 

Note that this is only for policies (mostly because policies are managed by the global CP). It might be trickier for dataplanes for example

kleinfreund:
## Changes

**chore: adds SingleResourceParameters type**

Adds a `SingleResourceParameters` type including common query parameters available on all API endpoints for retrieving single resources and uses it for the appropriate API wrapper methods.

**chore: adds copyTextToClipboard utility**

**refactor: deduplicates policy details**

Creates a `PolicyDetails` component to get rid of the duplication of code in `PolicyListView` and `PolicyDetailView`.

**feat: always shows resources in universal format**

Reworks the `YamlView` component into a more independent `ResourceCodeBlock` component with the built it ability to fetch the right resource via the API: instead of getting the resource as an object, it now gets a wrapper around the appropriate API method to retrieve the resource by itself. The component also no longer offers the two display formats Universal and Kubernetes in the form of tabs but rather always shows Universal with the option to copy the Kubernetes format.

Resolves #696.

## Screenshots

image


johncowen:
Are you planning on using https://github.com/kumahq/kuma-gui/pull/897 for this?

src/app/common/ResourceCodeBlock.vue
```diff
@@ -0,0 +1,141 @@
...
+        >
+          <KIcon
+            color="currentColor"
+            icon="copy"
+            size="18"
+            :title="i18n.t('common.copyKubernetesText')"
+          />
+
+          {{ i18n.t('common.copyKubernetesShortText') }}
+        </KButton>
```
johncowen, line 30: We discussed in GUI sync whether this should re-use the reusable CopyButton component from https://github.com/kumahq/kuma-gui/pull/897. Personally I think it should whether here or in a quick follow up.

src/app/common/ResourceCodeBlock.vue
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+}
+
+function toYamlRepresentation(resource: Entity): string {
+  const { creationTime, modificationTime, ...resourceWithoutTimes } = resource
+
+  return toYaml(resourceWithoutTimes)
+}
+</script>
```
johncowen, line 141: Personally I don't think we should be embedding data retrieval/loading/error support in a presentation component. I can kinda understand how you're thinking, but we should work on moving towards only passing data down into presentational components, thus decoupling them from how they end up with their data. Not a blocker right now tho.

src/app/data-planes/components/DataPlaneEntitySummary.vue
```diff
@@ -126,13 +126,12 @@
           </div>
         </section>
 
-        <section class="config-section">
-          <YamlView
-            id="code-block-data-plane-summary"
-            :content="dataPlane"
-            code-max-height="250px"
-          />
-        </section>
```
johncowen, line 135: Would it be possible for the moment to re-use this inside of ResourceCodeBlock? That way we keep a presentational component that just requires data to be passed into it.

src/utilities/copyTextToClipboard.ts
```diff
@@ -0,0 +1,48 @@
...
+  try {
+    isSuccess = document.execCommand('copy')
+  } catch {
+    isSuccess = false
+  } finally {
+    document.body.removeChild(textArea)
+  }
+
+  return isSuccess
+}
```
johncowen, line 48: Also waiting on https://github.com/kumahq/kuma-gui/pull/897#discussion_r1201882557

johncowen (commented):
Few things comments, non of them blocking apart from the one waiting on whether we need to use the kongponents thing for copying or not. If we can get that resolve I can 👍 this.

src/app/common/ResourceCodeBlock.vue
kleinfreund: I’ve looked into this and it’s not quite as straightforward.

The "Copy as Kubernetes" button’s action fetches the Kubernetes format of the resource on demand. However, in order to use said CopyButton component, we need to know the text that’s to be copied up-front. We could fetch the Kubernetes version up front, but that’s a waste of network resources. Alternatively, CopyButton could have a prop `getText` which returns the text to be copied.

src/app/data-planes/components/DataPlaneEntitySummary.vue
kleinfreund: Not sure I understand your question. Re-use what inside `ResourceCodeBlock`? `YamlView`?

kleinfreund:
@johncowen Please re-review this PR.

src/app/data-planes/components/DataPlaneEntitySummary.vue
johncowen: Yeah, maybe I misunderstood the change set, but was YamlView a presentational component that didn't deal with any loading of things? If we want to load things in ResourceCodeBlock we could do that and then pass the result to YamlView, then we still have a presentational component to use.

This isn't a biggie though, especially if I misunderstood

johncowen:
I think we are going to use https://github.com/kumahq/kuma-gui/pull/897 for this right? Brief look at the new changeset makes it seem like we are? Could we merge that and rebase now its approved?

kleinfreund:
I’ve rebased this now.

johncowen:
👍 Awesome lemme take a quick look

kleinfreund:
> And wondered why I was seeing the kubernetes button when I'm running in standalone mode. I figured I should only see the YAML I'm installed on, and then come to think of it only a single copy button ("Who cares about kubernetes when you are using universal and vice-versa").

Yeah, we settled on just always showing the same state: Universal by default, button to copy as Kubernetes. It would be relatively simple from the state of this PR to switch to show Universal or Kubernetes conditionally.

Possible would also be a variant where ResourceCodeBlock has a select widget for selecting the format.

src/app/data-planes/components/DataPlaneEntitySummary.vue
kleinfreund: With what we settled on implementing, there isn’t a need for having a component that can handle both views. We might get back to this in a different form (i.e. without tabs) at some point.

And with that, it’s also true that YamlView was "more" presentational. In a sense ResourceCodeBlock is a more opinionated/application-specific component (hence the name implying its a code block for showing resources) so personally I’m fine with it being provided with a bit of logic to fetch the content it displays.

PR #927, opened 2023-05-19
feat: adds default slot to block components

kleinfreund:
Wraps the title content of our block components in a default slot so you can set the content from the consumer side.

### DIFF
--- a/src/app/common/StatusInfo.vue
+++ b/src/app/common/StatusInfo.vue
@@ -3,7 +3,7 @@
     <LoadingBlock v-if="isLoading" />
 
     <ErrorBlock
-      v-else-if="hasError"
+      v-else-if="hasError || error !== null"
       :error="error"
     />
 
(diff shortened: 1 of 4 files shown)


src/app/common/StatusInfo.vue
```diff
@@ -3,7 +3,7 @@
     <LoadingBlock v-if="isLoading" />
 
     <ErrorBlock
-      v-else-if="hasError"
```
johncowen, line 6: I think the slot things is definitely what we should be doing, just wondering why this is here also?

In what cases should `hasError` be different to whether `error` is empty or not? Kinda not related to this PR but can we just check whether `error` is set to determine whether we `haveAnError`?

kleinfreund: Some uses of this component don’t have an error object and don’t provide it so we need to check for both. Maybe we can only deal with error objects in some future in which case everything related to `props.hasError` can go.

johncowen: Ok just a note this would have been a much easier 👍 without this change, from what I can see it wasn't required here

johncowen (approved):
LGTM!

PR #929, opened 2023-05-19
fix: triggering global loading when re-running bootstrap

kleinfreund:
Fixes triggering the global loading state when re-running the bootstrap function.


johncowen (approved):
Approving this as it gets us past a little issue and the code seems good to me, but I think the correct fix is bigger than this, and we are tracking that elsewhere

PR #930, opened 2023-05-19
feat: replaces mesh selector with mesh list view

kleinfreund:
Replaces the mesh selector with selecting the current mesh through the mesh list view.

With it, the need to categorize sidebar item is no longer there and so we just have up to 5 items in the sidebar now and can remove some code from related components.


src/app/AppSidebar.vue
```diff
@@ -1,45 +1,39 @@
...
-const navItems = computed(() => getNavItems(store.getters['config/getMulticlusterStatus'], store.state.meshes.items.length > 0))
-const meshes = computed(() => store.state.meshes.items)
+const navItems = computed(() => getNavItems(store.getters['config/getMulticlusterStatus']))
 
-watch(() => store.state.selectedMesh, () => {
-  store.dispatch('sidebar/getMeshInsights')
+watch(() => route.params.mesh, (newMesh, oldMesh) => {
+  if (newMesh !== oldMesh && newMesh) {
+    store.dispatch('sidebar/getMeshInsights', newMesh)
+  }
```
johncowen, line 36: If we no longer switch the mesh from the side bar, should this be here in `AppSidebar`? Feels like we should only be getting mesh insights when we visit the mesh overview page?

src/app/common/TagList.vue
```diff
@@ -66,10 +65,15 @@ function getRoute(tag: LabelValue): RouteLocation | undefined {
         })
       }
       case 'kuma.io/service': {
+        // Annotations by themselves don’t have information about a service’s associated mesh. The easiest solution is to read this information from the current route. A better approach could be to provide the current mesh as an optional prop to `TagList`.
+        if (!('mesh' in route.params)) {
+          return undefined
+        }
+
```
johncowen, line 72: agree that the better approach would to provide this from The Outside, so thanks for the comment here, should we add a TODO on the front?

Not totally aware of how TagList currently works though, but at a brief look we could get rid of the entire switch statement by providing a possible href for tags from The Outside.

Anyway thanks again for the comment, definitely one to avoid working on here 👏 We should follow up at some point though.

src/app/getNavItems.ts
```diff
@@ -43,33 +40,10 @@ export function getNavItems(isMultizoneMode: boolean, hasMeshes: boolean): NavIt
...
-      },
-    ]
-
-  return [
     {
-      name: 'Home',
-      routeName: 'home',
+      name: 'Meshes',
+      routeName: 'mesh-list-view',
+      anchorRouteName: 'mesh-abstract-view',
```
johncowen, line 46: Just a comment/thought. I kinda think we could do away with `getNavItems` entirely and just configure the Sidebar in the sidebar itself, feels even more indirection-y now we barely have anything in the sidebar. Just to be clear no action here just thinking out loud.

features/mesh/Index.feature
```diff
@@ -32,8 +37,14 @@ Feature: mesh / index
...
+    When I visit the "/meshes" URL
+
+    When I click the "<Selector>" element
+    Then the URL contains "/mesh/<Mesh>"
+    And the "$mesh-breadcrumb" element contains "<Mesh>"
+
+    Examples:
+      | Mesh                                      | Selector                                         |
+      | aalphabetically-second-because-of-default | [data-testid='data-overview-table'] tbody tr:nth-child(2) [data-testid='detail-view-link'] |
+      | default                                   | [data-testid='data-overview-table'] tbody tr:nth-child(1) [data-testid='detail-view-link'] |
```
johncowen, line 50: Should we make CSS selector aliases for these? We can probably get rid of the ridiculously named mesh now 😆 .

Also did I see this test no longer requires `/mesh-insights` mocking out? Or does it still need it but only need at least 1 item maybe?  not sure lmk.

johncowen (commented):
Looks good! Mostly unimportant comments, lmk and we can get this approved

johncowen:
P.S. Sorry forgot to ask if this requires integration changes elsewhere? (re: getNavItems)

src/app/AppSidebar.vue
kleinfreund: Probably (or rather: hopefully) not, but there are quite a few things tied into calls like this I don’t want to carve all of those out right now and check if they’re really not needed. They might not be needed here, but maybe the rest of the code relies on it being executed still.

src/app/common/TagList.vue
kleinfreund: The only way to be able to provide the links up front would be to have this exact logic on the outside. That logic should really be here so we can always rely on it. Only then can we just show tags for the annotations coming from our resources without thinking about it.

That being said, it makes sense to let the consuming side provide a mesh value if one is available: it seems logical that only mesh-specific resources have annotation referring to mesh resources. I’ll leave a comment.

features/mesh/Index.feature
kleinfreund: I renamed the second mesh and removed the mesh insights mock. Adding aliases here doesn’t really help, does it?

src/app/AppSidebar.vue
johncowen: gotcha, will be nice once we have a bigger suite of tests

features/mesh/Index.feature
johncowen: > Adding aliases here doesn’t really help, does it?

Well if we are in here it reduces the duplication, up to you, no biggie.

kleinfreund: But how? We could have a partial alias for the start and end bit of the overall selector, but it’s the middle bit that’s different.

johncowen: Something like:

https://github.com/kumahq/kuma-gui/blob/8cdae53804e11491ff218fb81551f3401931a5c3/features/mesh/dataplanes/Index.feature#L7-L9

https://github.com/kumahq/kuma-gui/blob/8cdae53804e11491ff218fb81551f3401931a5c3/features/mesh/dataplanes/Index.feature#L69-L72

kleinfreund:
@johncowen Please re-review this PR.

kleinfreund:
I’ve reverted the last two commits I added. Without support for the aliases to be correctly resolved (why you wouldn’t want this feature to work correctly isn’t really clear to me; why should aliases only work in the first position of a selector?), using them is really not an improvement at all.

The approach used now is simple and works.

johncowen:
Please see my last comment, I don't think there's anything more to add

johncowen (approved):

PR #957, opened 2023-05-25
fix: passing invalid data to mutations

kleinfreund:
Fixes two `commit` calls that weren’t passing any arguments to their respective mutation implementations which would lead to an exception.

### DIFF
--- a/src/store/storeConfig.ts
+++ b/src/store/storeConfig.ts
@@ -419,7 +419,7 @@ export const storeConfig = (kumaApi: KumaApi): StoreOptions<State> => {
 
           commit('SET_INTERNAL_SERVICE_SUMMARY', await fetchAllResources(endpoint))
         } catch {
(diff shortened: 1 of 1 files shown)


johncowen (commented):
When you say exception did you mean like a console error? If so could you explain to me how I repro it before applying this PR? ty!

kleinfreund:
> When you say exception did you mean like a console error? If so could you explain to me how I repro it before applying this PR? ty!

I do mean exception. The mutations `SET_INTERNAL_SERVICE_SUMMARY` and `SET_EXTERNAL_SERVICE_SUMMARY` are currently called without argument when `fetchAllResources` throws which happens when the underlying Kuma API wrappers throw and that happens for any error (e.g. 404) during the processing of the network request.

johncowen (approved):
ty!

PR #960, opened 2023-05-25
fix: nav tab bar having border bottom

kleinfreund:
Removes the bottom border from the KTabs component forming our nav tabs in the MeshView component. This is most notable in the views with a summary sidebar (i.e. services and DPPs).

### DIFF
--- a/src/app/meshes/views/MeshView.vue
+++ b/src/app/meshes/views/MeshView.vue
@@ -56,7 +56,7 @@ const items = [
   },
 ]
 </script>
-<style scoped>
+<style lang="scss" scoped>
 .tab-link a {
   display: block;
   padding: var(--spacing-md);
@@ -68,6 +68,10 @@ li.active .tab-link a {
 }
 </style>
 <style lang="scss">
+.route-mesh-view-tabs ul:not(.increase-specificity) {
+  border-bottom: none;
+}
+
 .route-mesh-view-tabs > ul .tab-item {
   padding: 0 !important;
 }


johncowen (commented):
Bit concerned that we are doing more overwriting. i.e. these tabs are now the only ones without bottom borders as far as I can follow. Is there an option to move the content with the top border slightly away from the tabs? Not sure if that would look ok, but if there was an option to stop doing more overwrites to KTabs I would like that.

kleinfreund:
> Bit concerned that we are doing more overwriting. i.e. these tabs are now the only ones without bottom borders as far as I can follow. Is there an option to move the content with the top border slightly away from the tabs? Not sure if that would look ok, but if there was an option to stop doing more overwrites to KTabs I would like that.

I’m not sure I’m following. Why should there be a bottom border on the navigational tabs? The way we use them, they always come before a bordered container (e.g. a card) so we’d always have a two borders and it’s much more robust to override the border of KTabs than to remove the top border of the following container. So, put another way: there always will already be a border beneath the navigation tabs.

lahabana:
I don't see what changes with the PR

kleinfreund:
**Before**:

Screenshot from 2023-05-31 09-27-58

**After**:

Screenshot from 2023-05-31 09-27-28


Note how in the first screenshot, there is a border stretching the entire width of the content area. It’s most clear in the gap between listing and summary.

This border shouldn’t be there at all.

lahabana:
Could this be a bug in the actual component?

johncowen:
My thought was to give the content a small top margin to space it away from the tabs. I'm not totally sure whether that idea is better than what @kleinfreund has done here, but it avoids overwriting more and more CSS things which is generally what we should try to avoid. The exception here of course is that we are already overwriting so much that we are nearly in "Triggers Broom / Ship of Theseus" territory.

On another note, I'm just working/thinking through our options for KTabs usage. I desperately want to keep using them, but for something like a tabbed interface, they are causing more issues that they are solving and therefore we are having to overwrite most of it. I'm catching up with some folks this afternoon on some ideas I had and I'll know more then. If possible I'd like to wait until that convo before we merge anything more that is to do with tabs so we know what our approach.

TLDR; This is kinda similar to https://github.com/kumahq/kuma-gui/pull/962

@kleinfreund lemme know if you wanted to come along / chat through what I had in mind first, ping me offline. I'm guessing this can wait until at leats this afternoon?

kleinfreund:
All those reasons are why I had used a custom `NavTabs` component initially as I was unsuccessful in the attempt of adding adequate support for navigational tabs to KTabs and using KTabs just seemed more trouble than not. That has its own downsides (mainly attached to needing to duplicate a lot of styles to make it look like KTabs), but it avoided overriding KTabs.

I’m happy to wait for this conversation to resolve, but I also think it’s hardly the case that we’re overriding "most things" of KTabs. In addition, some of those overrides wouldn’t be necessary if https://github.com/Kong/kongponents/pull/1251 could’ve been accepted which I maintain should be a consideration.

johncowen:
See https://github.com/kumahq/kuma-gui/pull/962#issuecomment-1570518885 (trying to move this conversation over there)

johncowen:
Just an additional note more specific to this very PR, did we try adding a bit of space under the tabs and see if that's something that looks ok, thus avoiding having tabs with no bottom border? Is that something that would work?

kleinfreund:
I did try that and in my opinion it doesn’t look good.

In any case, this PR is not _meant_ to change the aesthetic but to fix the bug of their being two borders on top of each other when that shouldn’t be the case.

lahabana:
I think we've discussed with design to move away from this view in favour of a tray in the future. As this is a minor aesthetic bug should we just shelve this and it will be transparently addressed when we use trays no?

kleinfreund:
> I think we've discussed with design to move away from this view in favour of a tray in the future. As this is a minor aesthetic bug should we just shelve this and it will be transparently addressed when we use trays no?

Not sure this relates to a view using an overlay tray instead of a sidebar column. The issue with the borders being duplicated would exist there as well.

kleinfreund:
> I don't think so as the issue is not present when there's no box to the right of the tab (you can see this with a smaller screen). If there were a tray there wouldn't be a second box.

The issue exists also when there is only one box beneath the tabs. Then you still have two borders. They happen to have the same color, so visually you can only tell by the fact that that border seemingly has a width of 2 pixels (2 times 1 pixel more specifically).

lahabana:
> The issue exists also when there is only one box beneath the tabs. Then you still have two borders. They happen to have the same color, so visually you can only tell by the fact that that border seemingly has a width of 2 pixels (2 times 1 pixel more specifically).

But that's already

Issue #318, opened 2022-03-30
Improve UI for Zone {ingress,egress}


johnharris85:

lahabana:
I think we should likely have 1 entry in the sidebar: `zones` in this we'll be able to go in and see the CPs, the zoneIngress, zoneEgress, a future schema showing that all zones connect to one another...

lahabana:
xref 406

lahabana:
xref #1008 let's check that the new view doesn't have this issue anymore

kleinfreund:
@lahabana 

Fixing https://github.com/kumahq/kuma-gui/issues/1008 in https://github.com/kumahq/kuma-gui/pull/1019.

PR #962, opened 2023-05-26
refactor: extracts nav tabs component

kleinfreund:
**fix: not using var() when using custom property**

**refactor: extracts nav tabs component**

Extracts a NavTabs component out of MeshView so it can be re-used.

Changes the detection for the current nav tab to be based on `route.matched` so it doesn’t rely on parent route names being prefixes of child route names (currently, this is used to identify `policies-list-view` as a matching route for `policies`).


src/assets/styles/_components.scss
```diff
@@ -6,7 +6,7 @@ To be used in places where we solely need KCard’s frame properties.
 
 .kcard-border {
   border: var(--KCardBorder);
-  border-radius: --KCardBorderRadius;
+  border-radius: var(--KCardBorderRadius);
```
johncowen, line 9: Good spot! Could you split this out into a different PR?

johncowen (commented):
Thought I'd best leave a message here to mention the branch I have that aims to address an issue with out KTabs usage (take a look at the HTML we currently render, i.e. the completely empty tabs panel divs).

I'm waiting to have a chat with some other folks on how to approach this, and I'll know more later on today hopefully.

johncowen:
Just to give an update here, we have a plan for a way forwards to we can hopefully just use the native/bare kongponent for our usecase with no overwrites. There should be something ready to review/use in the next day or so. Are you good to wait for that as it hopefully means we can just use the native/bare kongponent moving forwards? Lmk if you had any questions around that. (will post this on the other issue also as I think some of the convo is other there unfortunately, apologies for the duplication)

johncowen:
Also wanted to make it clear on this PR what the main issue is with our current usage of KTabs (apart from all the overwriting / Triggers Broom / Ship of Thesus thing)

As we render the "tab panel" outside of KTabs whose content updates depending on the browser's URL/Vue's router (see https://github.com/kumahq/kuma-gui/pull/932), KTabs currently adds additional empty div's inside itself, that we don't use or need, additionally these empty divs/tab-panels are all linked to the tab buttons via aria attributes

kleinfreund:
This PR is not meant to address the shortcomings of KTabs. The point is to create a re-usable abstraction so we can have multiple instances of navigational tabs without duplicating the usage of KTabs unnecessarily.

This PR should not be blocked on addressing the shortcomings of KTabs. Once they are addressed, the usage of KTabs in NavTabs can be updated and this PR doesn’t makes that simpler because it avoids there being multiple uses of KTabs as nav tabs code.

Also, it blocks me on https://github.com/kumahq/kuma-gui/issues/318.

lahabana:
What part of this PR will survive once KTabs is updated to support usages? Feels like we'll be left with a component that brings nothing and we'd likely just undo what is done here no?

kleinfreund:
> What part of this PR will survive once KTabs is updated to support usages? Feels like we'll be left with a component that brings nothing and we'd likely just undo what is done here no?

That depends on the implementation changes made to KTabs, but even if it addresses all the issues we’ve encountered using it, most of this code will still be needed so that you can use it like this:

```ts
<template>
  <NavTabs :tabs="TABS" />
</template>

<script lang="ts" setup>
import NavTabs, { NavTab } from '@/app/common/NavTabs.vue'

const TABS: NavTab[] = [
  {
    title: 'Overview',
    routeName: 'mesh-detail-view',
  },
  {
    title: 'Services',
    routeName: 'services-list-view',
    activeRouteNames: ['service-detail-view'],
  },
  {
    title: 'Gateways',
    routeName: 'gateways-list-view',
    activeRouteNames: ['gateway-detail-view'],
  },
  {
    title: 'Data Plane Proxies',
    routeName: 'data-planes-list-view',
    activeRouteNames: ['data-plane-detail-view'],
  },
  {
    title: 'Policies',
    routeName: 'policies',
    activeRouteNames: ['policies-list-view', 'policy-detail-view'],
  },
]
</script>
```

instead of like this:

```ts
<template>
  <div class="nav-tabs-wrapper">
    <KTabs
      class="nav-tabs"
      :tabs="kTabs"
      :model-value="currentTabHash"
    >
      <template
        v-for="tab in TABS"
        :key="`${tab.routeName}-anchor`"
        #[`${tab.routeName}-anchor`]
      >
        <router-link :to="{ name: tab.routeName }">
          {{ tab.title }}
        </router-link>
      </template>
    </KTabs>
  </div>
</template>

<script lang="ts" setup>
import { KTabs, Tab } from '@kong/kongponents'
import { computed } from 'vue'
import { useRoute } from 'vue-router'

export interface NavTab {
  title: string
  routeName: string
  activeRouteNames?: string[]
}

const route = useRoute()

const TABS: NavTab[] = [
  {
    title: 'Overview',
    routeName: 'mesh-detail-view',
  },
  {
    title: 'Services',
    routeName: 'services-list-view',
    activeRouteNames: ['service-detail-view'],
  },
  {
    title: 'Gateways',
    routeName: 'gateways-list-view',
    activeRouteNames: ['gateway-detail-view'],
  },
  {
    title: 'Data Plane Proxies',
    routeName: 'data-planes-list-view',
    activeRouteNames: ['data-plane-detail-view'],
  },
  {
    title: 'Policies',
    routeName: 'policies',
    activeRouteNames: ['policies-list-view', 'policy-detail-view'],
  },
]

const kTabs = computed<Tab[]>(() => TABS.map((tab) => {
  const { title, routeName } = tab

  return {
    title,
    hash: '#' + routeName,
  }
}))
const currentTabHash = computed(() => {
  const activeTab = TABS.find((tab) => {
    if (tab.routeName === route.name) {
      return true
    }

    if (Array.isArray(tab.activeRouteNames) && tab.activeRouteNames.includes(route.name as string)) {
      return true
    }

    return false
  })
  const routeName = activeTab?.routeName ?? TABS[0].routeName

  return '#' + routeName
})
</script>

<style lang="scss" scoped>
.nav-tabs-wrapper {
  display: flex;
  align-items: stretch;
  gap: var(--spacing-md);
}
</style>
```

So, even in an ideal case of KTabs, we’d still have to do some logic to determine which tab is active, etc. An abstraction (i.e. `NavTabs` as found here) will be necessary.

adamdehaven:
If you just need the Tabs UI (without the corresponding panels below) why will the solution I proposed yesterday not work for this use-case? 

The PR is still a WIP, but here's the link: https://github.com/Kong/kongponents/pull/1426

johncowen:
Ok so as far as I can see the work for https://github.com/kumahq/kuma-gui/issues/318 is done over here: https://github.com/kleinfreund/kuma-gui/pull/2, so I don't think being able to get work done is "blocked", just that we can PR/merge things, maybe I'd misunderstood but I'd understood that not merging this was stopping you from working.

I would love to wait and see what happens in kongponents, which I believe is mostly done and being PRed imminently, so we can look at that and see if it covers everything we need to do/are currently overwriting.

I think my main concern with all of this is that until yesterday there was a fair chance that all of this could be done in a very different way, and seeing as this was PRed before that and way before work on https://github.com/kumahq/kuma-gui/issues/318 began (until yesterday I was assigned to be working on that one) I didn't want us to be duplicating work that would ideally be done in kongponents.

TLDR; Lets wait on kongponents there's no rush from what I understand?

adamdehaven:
I don't understand how every other Kong app that uses Kongponents can use the tabs natively and there is such an issue in this repository, even as we are trying to augment the component to meet the exact need here. 

It completely defeats the purpose of a consistent look & feel across our apps if you augment base-level components. 

This isn't a valid solution IMO and I still fail to see why the several solutions we proposed are unacceptable.

adamdehaven:
> We will at the very least always need to determine the current tab based on the active route.

Why can't you bind the current route to the [`model-value`](https://kongponents.konghq.com/components/tabs.html#v-model)?

kleinfreund:
> Why can't you bind the current route to the [`model-value`](https://kongponents.konghq.com/components/tabs.html#v-model)?

That alone doesn’t cover all our use cases. We want to show navigational tabs also for detail views and mark the associated tabs as active, so structurally, what we need is:

- Services tab -> links to service list view
  - Service list view: Services tab should be active
  - Service detail view: Services tab should be active

kleinfreund:
> I don't understand how every other Kong app that uses Kongponents can use the tabs natively and there is such an issue in this repository, even as we are trying to augment the component to meet the exact need here.
> 
> It completely defeats the purpose of a consistent look & feel across our apps if you augment base-level components.
> 
> This isn't a valid solution IMO and I still fail to see why the several solutions we proposed are unacceptable.

I really need to ask you to reconsider what this PR is about and what not and what it does change with use of KTabs and what it does not. I’m **moving** the use of KTabs into an abstraction so it’s easier for us to use navigational tabs while making use of KTabs for that purpose. I’m not implementing an alternative.

I’m very much using KTabs natively here as far as that’s possible.

This does not sacrifice the consistent look and feel across various applications.

The alterations we do make to KTabs stem from the fact that a proposed change to Kongponents (i.e. https://github.com/Kong/kongponents/pull/1251) could not be merged. It would’ve kept the contents of NavTabs as found in this PR to an absolute minimum (namely the determination of the active tab hash). Without it, what we _need_ to alter in KTabs are the following things:

- Override the tab item padding so that the slotted links take up the entire tab item area and so can be activated as links on that entire area.
- Set a view properties on those slotted in links so that they look exactly like native KTabs items using the for that purpose designated custom properties.

These two items are not a deviation from the look and feel native to Kongponents components, rather they ensure that this look and feel is maintained. And again, they would not be needed with a change like that proposed in https://github.com/Kong/kongponents/pull/1251.

kleinfreund:
This was updated and can be reviewed again.

I did the update to Kongponents manually as Dependabot will currently not do it. Kongponents now has a version limit for Vue <3.3 but we had updated to Vue 3.3 before that. I guess this limit is temporary.

src/app/meshes/views/MeshView.vue
```diff
@@ -1,30 +1,7 @@
...
-          <router-link
-            :to="{
-              name: item.hash.substr(1),
-            }"
-          >
-            {{ item.title }}
-          </router-link>
-        </template>
-      </KTabs>
+      <NavTabs :tabs="tabs" />
```
johncowen, line 4: Sorry just looking at this again following the last lot of changes plus our GUI sync.

Am I right in thinking that in the case of Mesh Tabs `NavTabs` is only replacing:

`(items.find(item => children.includes(data.get(item.hash)?.module as string)) ?? items[0]).hash`

?

Now I can see that we have left more of the "50% towards doing this automatically" in here I'm not as concerned.

johncowen (approved):

src/app/meshes/views/MeshView.vue
kleinfreund: Yes, in essence, the changes to MeshView are relatively small now. Some changes in generating the tabs array were necessary as a result but the approach is almost the same

PR #966, opened 2023-05-29
docs(zone): added small docs change

lukidzi:


src/app/zones/locales/en-us/index.yaml
```diff
@@ -92,6 +88,8 @@ zones:
               enabled: {zoneIngressEnabled}
           egress:
               enabled: {zoneEgressEnabled}
```
kleinfreund, line 90: **Suggestion**: While we’re here, would you mind fixing these, too? They’re not technically broken but still indented one level too far.

```suggestion
            enabled: {zoneIngressEnabled}
          egress:
            enabled: {zoneEgressEnabled}
```

PR #977, opened 2023-05-30
fix: not determining Zone ingress/egress correctly

kleinfreund:

### DIFF
--- a/src/app/zones/views/ZoneListView.vue
+++ b/src/app/zones/views/ZoneListView.vue
@@ -184,15 +184,15 @@ async function loadData(offset: number) {
     ])
 
     nextUrl.value = next
-    tableData.value.data = transformToTableData(items ?? [])
     zonesWithIngress.value = new Set(zoneIngressOverviews.map((zoneIngressOverview) => zoneIngressOverview.zoneIngress.zone))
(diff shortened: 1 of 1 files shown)


johncowen (approved):
So the fix here is due to use accessing `zonesWithIngress` within the `transformToTableData` function, but previously we were calling `transformToTableData` before `zoneWithIngress` was set/updated?

I think it would be better to try to favour pure functions with less side effects when doing this sort of stuff. But as a fix for the moment this LGTM if I understood correctly.

kleinfreund:
> I think it would be better to try to favour pure functions with less side effects when doing this sort of stuff. But as a fix for the moment this LGTM if I understood correctly.

Yeah, totally and in this case, the data isn’t used outside that function so we could easily move it in. I’ll make a change for this.

kleinfreund:
See https://github.com/kumahq/kuma-gui/pull/977/commits/a0fcede8fefa7b1a83f298672540f426e9619595

kleinfreund:
@johncowen Are you okay with https://github.com/kumahq/kuma-gui/pull/977/commits/a0fcede8fefa7b1a83f298672540f426e9619595?

johncowen (approved):
Oh sorry, yeah sure I'm good either way rn, the additional thing you added seems better also ty!

PR #978, opened 2023-05-30
fix: reading gateway listeners unsafely

### LINKED ISSUE #964: Gateway policies tab shows error when listeners is empty because listeners can be null
### What happened?

`src/app/data-planes/components/DataplanePolicies.vue:135` assumes `listeners` to be array based on type `MeshGatewayDataplane` in `src/types/index.d.ts:338`, but apparently it’s `null` when it’s empty.

Either the type and code should be updated if this is expected or backend should be updated to return empty array instead of `null`.

kleinfreund:
Resolves #964.


johncowen (approved):
Another instance where we would benefit from a centralized data layer plus auto generated types. 🤔 

Fix LGTM

PR #989, opened 2023-05-31
fix: inadequate error handling of Zone creation flow

kleinfreund:
Adds an error block at the start of the form content area that’s populated if an error occurs during Zone provisioning.


johncowen (commented):
Wondered if the error might go better underneath the input field? LMK, otherwise I think this is pretty much good to go!

kleinfreund:
This is meant to be a band aid to the error being not exposed to users at all. I mean to change this to more appropriate form validation based on error responses of the respective endpoints, but I’d like to do so when/once we rework the older wizards and possibly when we make changes to the API error format (I can’t find the ticket for that right now).

johncowen (approved):
LGTM!

PR #990, opened 2023-05-31
chore: regenerates snapshots for Kongponents

kleinfreund:


johncowen (approved):
Looks like theres some issues in kongponents with duplicated classes 😬 

But, this LG for us!

kleinfreund:
@johncowen Yeah I looked at it for a while and I suspect the missing `inheritAttrs: false` in the latest KIcon implementation causes this, but I stopped looking after realizing that this is likely a timing artifact in the test: in the actual code, I can’t see these duplicate classes.

PR #991, opened 2023-05-31
chore: Removes link underlines and uses pseudo-standard 16px font-size

johncowen:
This PR removes underlines from links across the entire app, but does not go through and remove all the places where we then remove the `underlines` (this can be a follow up PR, or if folks think its better to do that here, I'm happy to add)

We also start using the pseudo-standard `font-size: 16px` on our `body` element.

Overall this should mean less overwriting, and things should automatically start to look similar/consistent across environments.

I also added a root font weight here using a `--font-weight-normal` which we don't actually set as yet, but this is more for consistency as I'd seen this being used elsewhere.

There's maybe some more things we could remove, but I didn't want to go there just yet.

For checking/testing I've gone over most of the app and things look fine, but I may have missed somewhere I wasn't aware of so extra 👀 would be much appreciated.

Example before/after (note breadcrumb underlines and (Disabled) badges font size):

## Before

Screenshot 2023-05-31 at 11 57 50

## After

Screenshot 2023-05-31 at 11 57 37

### DIFF
--- a/src/assets/styles/_base.scss
+++ b/src/assets/styles/_base.scss
@@ -8,8 +8,9 @@ body {
   margin: 0;
   tab-size: 2;
   color: var(--black-500);
-  font-size: 1em;
   font-family: var(--font-family-sans);
+  font-weight: var(--font-weight-normal, 400);
+  font-size: 16px;
   line-height: 1.5;
 }
 
@@ -40,7 +41,7 @@ h6 {
 }
 
 a {
-  text-decoration: underline;
+  text-decoration: none;
 }
 
 a:link,
--- a/src/assets/styles/_variables.scss
+++ b/src/assets/styles/_variables.scss
@@ -30,8 +30,6 @@
   // KONGPONENTS
   // Reduces the excessive vertical padding.
   --KPopPaddingY: var(--spacing-xs);
-  // Sets the font size to the size of body copy as the default of 12 pixels is way too small.
-  --KBadgeFontSize: 1rem;
   // Removes the default max width of 200 pixels because badges should never be cut-off.
   --KBadgeMaxWidth: auto;
   // Sets a default border style for commonly used component frames/panels.


src/assets/styles/_base.scss
```diff
@@ -8,8 +8,9 @@ body {
   margin: 0;
   tab-size: 2;
   color: var(--black-500);
-  font-size: 1em;
   font-family: var(--font-family-sans);
+  font-weight: var(--font-weight-normal, 400);
```
kleinfreund, line 12: **Suggestion**: This should be `--font-weight-regular` from `src/assets/styles/_variables.scss`.

```suggestion
  font-weight: var(--font-weight-regular, 400);
```

src/assets/styles/_base.scss
```diff
@@ -8,8 +8,9 @@ body {
...
+  font-size: 16px;
```
kleinfreund, line 13: **Issue**: This should remain `1em`. As pointed out, `16px` is the de facto default font size set by browser user agent styles and `1em` will be computed to `16px` with the benefit that changing the default font size in your browser will be reflected by that. Using a non-relative unit will break this.

src/assets/styles/_base.scss
```diff
@@ -40,7 +41,7 @@ h6 {
 }
 
 a {
-  text-decoration: underline;
+  text-decoration: none;
```
kleinfreund, line 44: **Question**: Have you talked to design about whether the choice to not underline links can be reconsidered?

It’s a terrible practice and we really shouldn’t be doing it.

src/assets/styles/_variables.scss
```diff
@@ -30,8 +30,6 @@
   // KONGPONENTS
   // Reduces the excessive vertical padding.
   --KPopPaddingY: var(--spacing-xs);
-  // Sets the font size to the size of body copy as the default of 12 pixels is way too small.
-  --KBadgeFontSize: 1rem;
```
kleinfreund, line 34: **Issue**: Following the comment about not using non-relative units for the body font size, this shouldn’t be removed.

kleinfreund (changes requested):

src/assets/styles/_base.scss
johncowen: This was set to `--font-weight-normal` on purpose, at the moment it fallsback to `400` as we don't currently use `-normal`.

I want to try to align to more pseudo-standard things so that things "Just Work", see another project below:

Screenshot 2023-06-01 at 10 12 44

src/assets/styles/_base.scss
johncowen: Same reasoning as above/previous comment. Notice this doing this solves the majority of our weird font-sizings issues. to be fair, if anything this should be on `html` but I'd rather stick to the `pseudo-standard` of other projects where they use `body`

src/assets/styles/_base.scss
johncowen: I think we all talked to design a while ago. This was the team decision https://github.com/kumahq/kuma-gui/issues/738 as linked to in the description above.

There're follow up PRs to come that add slightly different treatment on hover and removing all the `text-decoration: none`s we have throughout the codebase, I wasn't sure whether to do that here or in subsequent PRs. I'm happy to do follows ups but lmk if you prefer otherwise.

Apart from that, AFAIK this is the decision as its stood for a while now and hasn't changed from what I understand. If you still disagree then feel free to bring it up in sync with the team.

> It’s a terrible practice and we really shouldn’t be doing it.

^ in your opinion?

src/assets/styles/_variables.scss
johncowen: See above

src/assets/styles/_base.scss
kleinfreund: We have those variables specifically for that purpose. Changing it should also affect the font-weight here.

Why should it be a new, different variable?

src/assets/styles/_base.scss
kleinfreund: We made that decision as a team (which did not involve design) because we wanted to follow what other projects were doing, but we did note that this wasn’t ideal and that a conversation with design should take place to reconsider that.

> ^ in your opinion?

Not opinion but a matter of fact. Links need to be underlined as otherwise the sole distinguishing visual element is its color which isn’t enough for some visually-impaired people. Adding a differentiation on hover does absolutely nothing to help that as by the point you hover over something you already harbor a guess as it being interactive.

src/assets/styles/_base.scss
kleinfreund: We shouldn’t sacrifice this feature just so we can drop `--KBadgeFontSize: 1rem;`.

src/assets/styles/_base.scss
johncowen: This is how it is in Konnect. We are aiming for consistency rather than things being different due to differences of opinion. If Konnect want to change their opinion then we stick to that, and we change as/when/if they do. As of right now this is how it is.

src/assets/styles/_base.scss
johncowen: Please see screengrabs for what this fixes for font-sizing. Its also exactly the same as Konnect.

kleinfreund: Whatever this change fixes needs to be accomplished in another way because this introduces the bug of not being able to change the browser font size.

johncowen: Why does it needs to be accomplished in another way, this code here is exactly the same as Konnect.

> introduces the bug of not being able to change the browser font size.

Could you explain a little more. I can still change my browser font size, maybe I'm misunderstanding what you mean?

src/assets/styles/_base.scss
johncowen: Because this is what they do in Konnect, I would like our CSS to be as close as possible to Konnect and avoid overwriting things as much as possible.

I think if anything we should change our existing variable name to be exactly the same as Konnect/Kongponents. 

Honestly this one isn't doing anything at the moment, but the point is, if the variable gets changed by the Konnect team in Konnect, then our will just automatically update to use the same setting without us doing anything, and that's the main point.

src/assets/styles/_base.scss
johncowen: Are we good to resolve this one?

src/assets/styles/_base.scss
kleinfreund: > Why does it needs to be accomplished in another way, this code here is exactly the same as Konnect.

Because it introduces a bug and there is nothing wrong with `font-size: 1em` on `body` to begin with.

> Could you explain a little more. I can still change my browser font size, maybe I'm misunderstanding what you mean?

Changing the font size in your browser (e.g. in Chrome from Medium to Large) changes what the browser puts into its user agent styles. As a result, `font-size: 1em` on `body` will compute to whatever that default is. Medium happens to be 16px while Large works out to 20px. And this works in Kuma GUI, but it breaks when using non-relative units.

johncowen: How do you change the font size in your browser?

kleinfreund: I don’t know which browser you’re using, but usually they have search features nowadays. A query for "font size" should lead you to the right place.

src/assets/styles/_base.scss
kleinfreund: Yes, but I had hoped for this conversation to be brought to design which is actually what we agreed as a team.

src/assets/styles/_base.scss
johncowen: Well I was asking how _you_ do it. Do you _ever_ do it yourself?

As mentioned above, I did it how I would usually do it and my font size change perfectly fine, so I don't see what the problem is? Which is why I thought I may have misunderstood 🤷‍♂️ 

Screenshot 2023-06-01 at 16 06 30

johncowen: Yeah right, that's kinda why I was asking how _you_ do it, I never use this for example. How many folks do you think use this when "Page Zoom" is available over this font size preference?

johncowen: Also note that most browsers have this in the top level menus, and is even referred to as you can see in my above browser grab as "Make Text Larger"

johncowen: I think the thing that trumps all this is, if this is how it is in Konnect, and somebody somewhere did actually use what you outlined in your screengrabs as opposed to just using the "Page Zoom" to set their font size. Nothing else in Konnect would change, which I would consider a bug also as its inconsistent.

So to sum up here, if you have a larger issue with not being able to use "Page Zoom" maybe you should raise it in Konnect-land. If they change, we will to.

src/assets/styles/_base.scss
johncowen: Awesome. If you still feel strongly about this feel free to take that to design or bring it with you to GUI or larger team sync and I personally would be happy to chat further. In the meantime https://github.com/kumahq/kuma-gui/issues/738 is what has been agreed as a team. If Konnect changes look and feel as a result of that, then we can too.

src/assets/styles/_base.scss
johncowen: For a bit more context, I used my browser preferences instead of "Page Zoom / Make Text Larger" to set the font size (my browser only gives me a very small range of sizes it seems). I then tried out a couple of popular website, namely Google and Github (this very PR). Github doesn't change at all (but does if I use Page Zoom). Google does have some text that changes, but only 1 piece of text, noticed the button text doesn't change at all.

Considering this and even more specifically our tech audience, how likely is it that somebody is using this font-size preference over something like "Page Zoom"?

Screenshot 2023-06-01 at 16 36 16

Honestly I don't see me making any changes here, so I think we should either resolve, or raise it with the rest of the team? Lmk

kleinfreund: Just because other implementations make the same mistake shouldn't mean need to make it, too, _especially_ if we have ways to avoid it. We can easily address a Kongponents component using relative font sizes and becoming too small/big by setting its variables (which is very much the intended way Kongponents favors for controlling theming properties).

I'm not okay with this change because of the bug I mentioned. Also, it's really not necessary.

johncowen: Bit more information here:

Here is our app on `master` (i.e. previous to this PR)

Screenshot 2023-06-02 at 09 21 06

That looks pretty broken currently to me if you are using the browser font-size preference to make the text bigger.

Here's the application with this PR applied and using relative units for font sizing if you are using the browser font-size preference to make the text bigger:

Screenshot 2023-06-02 at 09 23 50

Note I've only used relative font sizes on the menu items here.

Using a base font of 16px for a font size and then using relative units is pretty standard practice is quite a common almost "pseudo standard" approach in quite a few things I've been involved with including Konnect.

To finish up here:

1. Konnect uses a base font of 16px
2. As far as I'm aware most folks use "Page Zoom" nowadays if they want to increase their text size.
3. Some of the most popular sites on the internet do not increase their font size when using the font size browser preference.
4. This PR fixes a good few issues we have with messy looking font-sizing
5. If you want our application to increase font size for you when you increase the text size using your font size browser preference, then we should use a 16px base size (for the additional reasons as explained above), and then specify our font-size in relative units. As far as I am aware this is the most common approach.

Each one of those above reasons would be enough for me to apply this PR, but when taken al together I can't see I reason why we should not do this.

Let me know what you think.

kleinfreund: All of this works just as well with using `1em` instead of `16px` as the base font size **and** it doesn’t break the browser font size setting.

It may well be common practice but it’s certainly more problematic. We can do better and should. We shouldn’t assume what browser features people make use of to access a page. Since it’s easy for us to not break this feature, we should just do it.

> If you want our application to increase font size for you when you increase the text size using your font size browser preference, then we should use a 16px base size (for the additional reasons as explained above), and then specify our font-size in relative units. As far as I am aware this is the most common approach.

That’s not the best approach though. You _can totally_ use a relative base font size **and** relative font sizes for everything else and get the same benefit **without** breaking the browser font size setting.

This may be a common approach, but I pointed out how flawed it is.

All the things fixed in this PR can be fixed without breaking this.

johncowen: @lahabana following the chat we've all just had could you confirm the outcome of this conversation for me? The PR is still in "Changes Requested" and I'm unsure if there is anything else to be done here? lmk

kleinfreund: As per this conversation it was decided to use `font-size: 16px` and so I’m resolving this thread.

kleinfreund (approved):

PR #992, opened 2023-05-31
ci: backports workflow changes (release-2.2)

kleinfreund:
Backports changes to workflow files from `master` into `release-2.2`.


johncowen (approved):
Seems fine to me, the GH GUIs branch-protection rule needs updating after this also right?

kleinfreund:
Actually, no. I thought so, too, but: The branch protection rule here is using "Tests" because the name of the test job here is still "Tests" not "branch-protection". Should’ve maybe updated this first.

PR #997, opened 2023-06-01
refactor: internalizes data management for overview charts

kleinfreund:
Moves all the data fetching and chart data computation out of storeConfig and into OverviewCharts as there is no advantage in having that handled via the Vuex store.


johncowen (approved):
This one is quite dense for a code review, but I get why its like this. I tried it out and everything seems fine to me.

I noticed in the description that this does fix a couple of bugs, is there somewhere specific in this PR where you could point to where the fixes where?

Otherwise I'm good to 👍 this.

kleinfreund:
> I noticed in the description that this does fix a couple of bugs, is there somewhere specific in this PR where you could point to where the fixes where?

> Some charts sometimes animating in twice with potentially conflicting data

This is fixed by the loading of data now being central to the component and only setting data once per load cycle rather than kicking off a couple of independent store actions that all populate independent state.

> Some charts sometimes not showing a "circle" (arc) when they should show, for example, "0 Meshes"

When populating the charts, it’s important to _not_ add data points when the chart is empty. There was a case where we added data points like this:

```
const dataPoints: ChartDataPoint[] = [
  {
    title: i18n.t('common.charts.meshes.meshLabel'),
    data: meshStatus.value.total,
  }
]
```

And in that case, if `meshStatus.value.total` is `0`, you get the "no circle" effect because there is _one data point of zero size_ instead of _no data points_.

PR #998, opened 2023-06-01
chore: switches Datadog instances

kleinfreund:
**refactor: uses the correct types for DatadogLogger**

**chore: switches Datadog instances**

Updates the Datadog Logs initialization code to use a new client token associated with a different Datadog instance.

Also makes sure the environment will always be set correctly.

### DIFF
--- a/src/services/logger/DatadogLogger.ts
+++ b/src/services/logger/DatadogLogger.ts
@@ -12,14 +12,17 @@ export default class DatadogLogger {
 
   setup(config: ClientConfigInterface) {
     if (config.reports.enabled) {
+      const service = this.env('KUMA_PRODUCT_NAME').toLowerCase().replaceAll(' ', '-') + '-ui'
+
       datadogLogs.init({
-        // This token is called “kuma-gui” in Datadog
-        clientToken: 'pub94a0029259f79f29a5d881a06d1e9653',
+        // The current client token is called “kuma-ui” in the Datadog UI.
+        // Previous versions of the application used a different client token attached to a different Datadog instance. This client token was called “kuma-gui” in the Datadog UI.
+        clientToken: 'pub1aadd2cab84c05bf9959e00a35b213a1',
         site: 'datadoghq.com',
         forwardErrorsToLogs: true,
-        service: this.env('KUMA_PRODUCT_NAME'),
         sampleRate: 100,
-        env: 'production', // logging is currently disabled in anything other than production
+        service,
+        env: import.meta.env.MODE,
       })
     }
   }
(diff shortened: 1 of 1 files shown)


src/services/logger/DatadogLogger.ts
```diff
@@ -1,28 +1,28 @@
...
-  constructor(env: Env) {
+  env: Env['var']
+
+  constructor(env: Env['var']) {
     this.env = env
   }
 
   setup(config: ClientConfigInterface) {
     if (config.reports.enabled) {
+      const service = this.env('KUMA_PRODUCT_NAME').toLowerCase().replaceAll(' ', '-') + '-ui'
```
kleinfreund, line 15: **Comment**: I made sure this will end in Datadog as `kuma-ui` so it can be filtered from things related to Kuma itself which might be using the same Datadog instance (or not, still it’s nicer this way).

src/services/logger/DatadogLogger.ts
```diff
@@ -1,28 +1,28 @@
...
+        // The current client token is called “kuma-ui” in the Datadog UI.
+        // Previous versions of the application used a different client token attached to a different Datadog instance. This client token was called “kuma-gui” in the Datadog UI.
+        clientToken: 'pub1aadd2cab84c05bf9959e00a35b213a1',
         site: 'datadoghq.com',
         forwardErrorsToLogs: true,
-        service: this.env('KUMA_PRODUCT_NAME'),
         sampleRate: 100,
-        env: 'production', // logging is currently disabled in anything other than production
+        service,
+        env: import.meta.env.MODE,
```
kleinfreund, line 25: **Comment**: We’re not actually setting `NODE_ENV` anymore so I’m using `MODE` from Vite here.

src/services/logger/DatadogLogger.ts
johncowen: should we use `kuma-gui` not `-ui`. I actually prefer `ui` over `gui` (gooey), but seems like its a pattern we should stick with.

Also, if we are making a change here anyway can we use backticks/templates here instead of addition?

kleinfreund: In isolation, I would’ve used `-gui`, but I found that UI apps in Datadog were suffixed with `-ui` so I thought it’s better to have the consistency there.

src/services/logger/DatadogLogger.ts
kleinfreund: Yeah, I think it’s more safe to use a variable here:

- When you test out such an integration locally, it’s nicer to label events sent to Datadog with a dev environment
- If we were to change our approach to also log events from dev environments, we wouldn’t need to remember changing this

kleinfreund:
I’m closing this PR because we decided to remove Datadog Logs from the project as we aren’t really looking at the data.

PR #999, opened 2023-06-01
chore: Makes mesh tabs autogenerated

johncowen:
This PR makes it so the new tabs in a Mesh view are auto generated from the routes, this means that as we add new routes they will automatically be added to the tab bar.

---

Notes:

The routing changes here are just as a consequence of making it easier to get closer to a more generic "find the routes we want" i.e. I just removed an empty nest. I 'think' this was here from https://github.com/kumahq/kuma-gui/pull/922 where I found it really difficult to get our current breadcrumbs setup to work:

> One thing I will note is that it took me quite a while to organize the routing configurations to get the breadcrumbs correct, for a long while I could get it to link to meshes properly (the URL would look literally like /mesh/:mesh). I don't know how I managed to get it to print it properly, but it seems to be fine now, but I think there is a degree of unnecessary nesting in the route configuration just to get the breadcrumbs correct. I would like to change how we do breadcrumbing, but again that is something I'm punting for the minute.

^ from https://github.com/kumahq/kuma-gui/pull/922

Apart from that this just adds some sort of "find the root Mesh route and its children" that we want and use those for the tabs, it also uses our new i18n support to give the tabs human readable titles that are configurable from The Outside 🎉 

Longer term, once we've addressed the issue with our current breadcrumbing configuration, I'd like to revisit our routing configs to make sure they are only nested as much as required, and I would still like to make all the naming consistent. I would have much rather had some sort of function here where I just go `route('meshes.item').children.map()` to be able to retrieve the tabs from the routing. So all in all this is more of a lets move forwards with the feature rather than re-arrange everything now, I would guess this will be made more re-usable once we have more instances of doing this.

### DIFF
--- a/src/app/meshes/views/MeshView.vue
+++ b/src/app/meshes/views/MeshView.vue
@@ -32,18 +32,12 @@
 import { KTabs } from '@kong/kongponents'
 import { useRouter } from 'vue-router'
 
-import { useMeshRoutes, useI18n } from '@/utilities'
+import { useI18n } from '@/utilities'
 const { t } = useI18n()
 
 const router = useRouter()
 
-// dig down into where our mesh tabs are and automatically
-// convert the routes to tabs
-const meshRoutes = useMeshRoutes()
-  .find(item => item.name === 'mesh-index-view')?.children
-  ?.find(item => item.name === 'mesh-detail-view')?.children
-  ?.find(item => item.name === 'mesh-abstract-view')?.children ?? []
-
+const meshRoutes = router.getRoutes().find((route) => route.name === 'mesh-abstract-view')?.children ?? []
 const items = meshRoutes.map((item) => {
   if (typeof item.name === 'undefined') {
     const route = item.children?.[0]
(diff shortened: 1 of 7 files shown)


src/app/meshes/views/MeshView.vue
```diff
@@ -32,29 +32,35 @@
...
-    title: 'Policies',
-  },
-]
+
+// dig down into where our mesh tabs are and automatically
+// convert the routes to tabs
+const meshRoutes = useMeshRoutes()
+  .find(item => item.name === 'mesh-index-view')?.children
+  ?.find(item => item.name === 'mesh-detail-view')?.children
+  ?.find(item => item.name === 'mesh-abstract-view')?.children ?? []
```
kleinfreund, line 45: **Suggestion**: This really seems quite brittle and reliant to a _very_ specific route nesting and naming.

Wouldn’t something like this be more robust?

```ts
const meshRoutes = router.getRoutes().find((route) => route.name === 'mesh-abstract-view') ?? []
```

johncowen: Oh definitely yes, exactly the kind of thing I was after! ty!

kleinfreund: (As an aside, we should probably start putting route names in constants in the routes config file and export them and never use string literals for them.)

kleinfreund (approved):

PR #1003, opened 2023-06-02
refactor: abstracts away log calls

kleinfreund:
Adds an `info` method to our logger service and uses it throughout the code base instead of calling Datadog’s logger directly. The `info` method supports the same signature as `datadogLogs.logger.info`.

Also, removes Datadog Logs as we decided to not use it for the time being.


src/services/logger/Logger.ts
```diff
@@ -0,0 +1,28 @@
...
+  setup() {
+    // Currently, there is no setup code here. This could contain Datadog Logs setup code, for example.
+  }
+
+  info(...args: LogArgs) {
+    this._log('info', ...args)
+  }
+
+  _log(type: 'info', ...args: LogArgs) {
+    if (this.isEnabled) {
```
kleinfreund, line 24: **Comment**: It looks like Datagog’s log calls are automatically noops if it’s not initialized which is probably why we didn’t need something like that `this.isEnabled` flag. I was wondering why my logs were still visible even though `DisabledLogger` was in place, but I hadn’t added this, yet.

src/services/logger/Logger.ts
```diff
@@ -0,0 +1,28 @@
...
+  SIDEBAR_ITEM_CLICKED: 'sidebar-item-clicked',
+  TABLE_REFRESH_BUTTON_CLICKED: 'table-refresh-button-clicked',
+  TABS_TAB_CHANGE: 'tabs-tab-change',
+  CREATE_DATA_PLANE_PROXY_CLICKED: 'create-data-plane-proxy-clicked',
+}
+
+type LogArgs = [string] | [string, object | undefined] | [string, object | undefined, Error | undefined]
+
+export default class Logger {
+  isEnabled: boolean = true
```
lahabana, line 13: I don't understand why this is here when you have a specialised class `DisabledLogger` to me `DisabledLogger` would override the `_log()` method with an empty body and you wouldn't need this here right? If not what's the point of `DisabledLogger` ?

src/services/production.ts
```diff
@@ -223,7 +220,6 @@ export const services: ServiceConfigurator<SupportedTokens> = ($) => [
     service: useBootstrap,
     arguments: [
       $.logger,
-      $.api,
```
lahabana, line 226: Why is this change needed here?

src/services/logger/Logger.ts
kleinfreund: That’s also an option, but if we add more methods to `Logger`, you’d always need to override them all in `DisabledLogger`.

src/services/production.ts
kleinfreund: Because the bootstrap function no longer uses the API service and so doesn’t need it. The reason it was using it was that we initialized the logger with the result to an API call to `/config` which we no longer do with this change.

src/services/logger/Logger.ts
lahabana: The way these log classes are typically implemented is by having 1 internal methods and all the others just call this 1 method.
To me either you delete the `DisabledLogger` class because it's not useful or you make it as mentioned previously. This seems to be a little confusing...

kleinfreund: That works for me. Overriding `_log` in the disabled logger should be enough as well.

kleinfreund: This was changed in https://github.com/kumahq/kuma-gui/pull/1003/commits/ee9829d1de23ca8b2a8c7c1722398565796a9cc1.

lahabana (approved):

PR #1006, opened 2023-06-05
fix: certificates not showing correctly

kleinfreund:
Fixes the issue of not showing the mTLS certificates data correctly and also takes that moment to move the label text into the i18n layer.


kleinfreund:
Yeah, good idea. I added some to see this work.

johncowen (approved):
LGTM! Thanks for switching this out to use our i18n 👏

PR #1007, opened 2023-06-05
refactor: logger setup

kleinfreund:
**refactor: moves logger setup into main.ts**

Moves the logger setup call to a place where it can’t happen multiple times.

**chore: removes logger.log method**

Removes the log method as it’s not very useful with an `info` method already present.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -17,9 +17,6 @@ async function mountVueApplication() {
       : [],
   )
 
-  const logger = get($.logger)
-  logger.setup()
-
   const app = await get($.app)((await import('./app/App.vue')).default)
   app.mount('#app')
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -16,6 +16,10 @@ async function mountVueApplication() {
       })()
       : [],
   )
+
+  const logger = get($.logger)
+  logger.setup()
+
   const app = await get($.app)((await import('./app/App.vue')).default)
   app.mount('#app')
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,7 +3,6 @@ import { RouteRecordRaw } from 'vue-router'
 import { Store, storeKey } from 'vuex'
 
 import { createRouter } from './router/router'
-import type Logger from './services/logger/Logger'
 import type { EnvVars } from '@/services/env/Env'
 import type { State } from '@/store/storeConfig'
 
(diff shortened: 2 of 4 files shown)


src/main.ts
```diff
@@ -16,6 +16,10 @@ async function mountVueApplication() {
       })()
       : [],
   )
+
+  const logger = get($.logger)
+  logger.setup()
```
johncowen, line 21: I know this work is spread over a few different PRs in different places, so I could be wrong here. But if we no longer need information from the API in order to `setup` the logging, could we just use the constructor to setup?

Alternatively just as a general FYI service instantiation is always a good place to perform post-constructor setup, but in this case I think just in the constructor is fine.

Just to underline, I might have misunderstood the reasons behind keeping `setup` (maybe we want it there incase we need to pass API info in again at some point?)

Another point I'd like to make here is that we really need to keep the `main` files to a bare minimum. It should just be build the container and instantiate the application from it. Nothing more.

There is already some extra bits in here that I didn't pick on just because I thought it was a quick fix for a previous bug fix/elastoplast. but I really want to avoid this practice spreading.

TLDR; Only put extra things in `main` as a last resort, when you've absolutely exhausted every other option.

kleinfreund: This is changed now.

johncowen: ok cool, so just to check. Am I right in thinking that the thinking is that the logger constructor will only ever be instantiated when we need to use it right (via `useLogger`)? Just checking that there is nothing in the constructor that we need to `await` for before we call `logger.info` i.e.:


```javascript
const logger = useLogger() // this is the first time I call useLogger so instantiation/construction is done at this point
logger.info('hi') // will this call go through correctly without having awaited on something?
```

If anything in the above example isn't guaranteed to work then we need to cover that one way or another (there are options). If it is guaranteed to work then fine I can probably approve this in that case.

kleinfreund: Currently, there isn’t anything that can’t be covered by a constructor and so this fine. If we find a need for that, the will have to reintroduce a setup method as constructors can’t be async, but that’s a hypothetical we shouldn’t account for right now.

johncowen: Ok gotcha, sounds good then, was just to check as I wasn't totally sure.

johncowen (approved):

PR #1017, opened 2023-06-06
feat: moves zones into tabs

kleinfreund:
**feat: moves zones into tabs**:

Renames the routes associated to the Zone CP views to include the `-cp` bit to make it more clear this is a view of Zone CPs.

Moves the Zone sidebar items into a Zones page with navigational tabs which shows the Zone CPs view by default (i.e. like we’ve done for mesh-related items).

**chore: removes sidebar item counts**:

Removes the sidebar item counts as they’re really not useful anymore.

### DIFF
--- a/features/zones/Index.feature
+++ b/features/zones/Index.feature
@@ -8,7 +8,7 @@ Feature: The create Zone flow works
       | environment-kubernetes-radio-button | [data-testid='environment-kubernetes-radio-button'] |
       | ingress-input-switch                | [data-testid='ingress-input-switch']                |
       | egress-input-switch                 | [data-testid='egress-input-switch']                 |
-    When I visit the "/zones/-create" URL
+    When I visit the "/zone-cps/-create" URL
 
   Scenario: The form shows only the initial elements
     Then the "$name-input" element exists
--- a/features/zones/Index.feature
+++ b/features/zones/Index.feature
@@ -8,7 +8,7 @@ Feature: The create Zone flow works
       | environment-kubernetes-radio-button | [data-testid='environment-kubernetes-radio-button'] |
       | ingress-input-switch                | [data-testid='ingress-input-switch']                |
       | egress-input-switch                 | [data-testid='egress-input-switch']                 |
-    When I visit the "/zone-cps/-create" URL
+    When I visit the "/zones/-create" URL
 
   Scenario: The form shows only the initial elements
     Then the "$name-input" element exists
--- a/src/app/zones/routes.ts
+++ b/src/app/zones/routes.ts
@@ -29,7 +29,7 @@ export const routes = (
       redirect: () => ({ name: 'zone-cp-list-view' }),
       children: [
         {
-          path: '/zone-cps',
+          path: 'zone-cps',
           name: 'zone-cp-abstract-view',
           meta: {
             title: 'Zone CPs',
@@ -61,7 +61,7 @@ export const routes = (
           ],
         },
         {
-          path: '/zone-ingresses',
+          path: 'zone-ingresses',
           name: 'zone-ingress-abstract-view',
           meta: {
             title: 'Zone Ingresses',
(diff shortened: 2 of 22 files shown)


src/app/zones/routes.ts
```diff
@@ -16,22 +17,28 @@ export const routes = (
   actions: RouteRecordRaw[],
 ): RouteRecordRaw[] => {
   return [
+    ...actions,
     {
       path: '/zones',
-      name: 'zone-abstract-view',
+      name: 'zone-index-view',
```
johncowen, line 23: What's the desired URL structure here?

`/gui/zones/zone-cps` etc etc or `/gui/zone-cps` etc etc ?

(I maybe have some more questions, but not totally sure as I'm still looking but wanted to get this question posted)

kleinfreund: The desired structure is `/zone-cps`, `/zone-ingresses`, etc. (hence these paths _also_ start with `/`). However, to make sure we’re not accidentally adding an empty path home route I’ve given that one the path `/zones` because it would otherwise be matched *instead* of the home route.

This is a little weird, but I didn’t want to repeat using `ZoneIndexView` for each of the three zone route groups.

johncowen: I am right in thinking you can't actually get to `/zones` right?

I think we should make the URL structure be `/zones/zone-cps` etc, wdyt?

johncowen:
I was also looking at what happens in standalone. I see we change the sidebar item to change the link to `Zone Egresses` but still take you to a page where there are three tabs for each of `Zone CPs`, `Zone Egresses` and `Zone Ingresses`.

Not sure if this is exactly what the ideal would be 🤔 umm, maybe:

- If we are changing the side menu item, then we shouldn't show the extra tabs
- If we are leaving all the tabs visible, then we should keep the side menu item the same but make it take you to the Egresses by default.

wdyt?

src/app/zones/routes.ts
kleinfreund: Alright, I changed this , too.

src/app/zones/routes.ts
```diff
@@ -1,8 +1,9 @@
 import { getLastNumberParameter } from '@/router/getLastParameter'
 import type { RouteRecordRaw } from 'vue-router'
+
 export const actions = (): RouteRecordRaw[] => {
   return [{
-    path: '-create',
+    path: '/zone-cps/-create',
```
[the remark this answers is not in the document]
kleinfreund: It would get tabs and it shouldn’t have tabs.

features/zones/Index.feature
```diff
@@ -8,7 +8,7 @@ Feature: The create Zone flow works
       | environment-kubernetes-radio-button | [data-testid='environment-kubernetes-radio-button'] |
       | ingress-input-switch                | [data-testid='ingress-input-switch']                |
       | egress-input-switch                 | [data-testid='egress-input-switch']                 |
-    When I visit the "/zones/-create" URL
+    When I visit the "/zone-cps/-create" URL
```
johncowen, line 11: This one doesn't seem to be inside `zones`, what are we actually creating here a `zone-cp` or a `zone`? (also see below comment on route nesting that could be related)

kleinfreund: I think this is actually for _Zones_ rather than Zone CPs. I’m not totally sure, but I think that’s accurate. I’ve moved the URL to `/zones`.

src/app/zones/routes.ts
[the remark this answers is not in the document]
johncowen: Ah of course! 🤔 Ok taking another look at this, thanks!

kleinfreund:
On the other hand, it’s kind of expected.

Making the redirect conditional isn’t straightforward because we can’t access the store. I think I’m fine with this. We could also not make Zones a breadcrumb item.

kleinfreund:
> Is linking the breadcrumb straight to zone-egress if we are in standalone an option?

Currently, it’s not, no.

johncowen:
Ok just taking one last look at this then I should be able to 👍

johncowen (approved):
I'm happy for this to go in as is. The "Clicking the Zones breadcrumb" thing in standalone is a little weird, but especially what we decided in our chat I think this can be approached in a follow up.

PR #1026, opened 2023-06-07
feat: change wording of status not available

### LINKED ISSUE #643: feat: Change wording of Status: not available
### Description

Sometimes we display `Status: not available` which sounds a bit like the service itself is not available, not the fact that the status information is not available

We should word this a little better like maybe `Status: Information not available`

kleinfreund:
**fix: not showing status in services list view**

Fixes not showing a service’s status in the service list view at all.

**feat: changes wording for "not_available" status**

Changes the wording for the `not_available` status from "not available" to "information not available". (I’ve added both camel-cased and snake-cased variants to the messages file as we sometimes use one, sometimes the other. If we change this, we can drop one.)

Resolves #643.

### DIFF
--- a/src/app/common/StatusBadge.vue
+++ b/src/app/common/StatusBadge.vue
@@ -17,24 +17,15 @@
 import { computed, PropType } from 'vue'
 
 import { StatusKeyword } from '@/types/index.d'
+import { useI18n } from '@/utilities'
 
-const STATUS: Record<StatusKeyword, { title: string, appearance: string }> = {
-  not_available: {
-    title: 'not available',
-    appearance: 'warning',
-  },
-  partially_degraded: {
-    title: 'partially degraded',
-    appearance: 'warning',
-  },
-  offline: {
-    title: 'offline',
-    appearance: 'danger',
-  },
-  online: {
-    title: 'online',
-    appearance: 'success',
-  },
+const i18n = useI18n()
+
+const STATUS: Record<StatusKeyword, string> = {
+  not_available: 'not-available',
+  partially_degraded: 'warning',
+  offline: 'danger',
+  online: 'success',
 }
 
 const props = defineProps({
@@ -50,7 +41,10 @@ const props = defineProps({
   },
 })
 
-const statusObject = computed(() => STATUS[props.status])
+const statusObject = computed(() => ({
+  title: i18n.t(`http.api.property.${props.status}`),
+  appearance: STATUS[props.status],
+}))
 </script>
 
 <style lang="scss" scoped>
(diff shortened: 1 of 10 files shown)


src/locales/en-us/common/index.yaml
```diff
@@ -3,3 +3,10 @@ common:
   copySuccessText: 'Copied!'
   copyKubernetesText: 'Copy as Kubernetes'
   copyKubernetesShortText: 'as k8s'
+  status:
+    online: 'Online'
+    offline: 'Offline'
+    partiallyDegraded: 'Partially degraded'
+    partially_degraded: 'Partially degraded'
+    notAvailable: 'Information not available'
+    not_available: 'Information not available'
```
kleinfreund, line 12: **Question**: Should these start with a capital letter? We currently do a mix of both.

src/locales/en-us/http/index.yaml
```diff
@@ -17,3 +17,9 @@ http:
       certificateExpirationTime: Expiration time
       lastCertificateRegeneration: Last generated
       certificateRegenerations: Regenerations
+      online: 'online'
+      offline: 'offline'
+      partiallyDegraded: 'partially degraded'
+      partially_degraded: 'partially degraded'
+      notAvailable: 'information not available'
+      not_available: 'information not available'
```
johncowen, line 25: nit: The thinking here was we would have `http.api.property` and `http.api.value`, is that ok to change up? I don't ever think there would be an instance were we don't know in code whether we are after a `property` or `value` and it could be vaguely useful to have the as separate things.

kleinfreund: Sure, I changed it.

src/app/common/StatusBadge.vue
```diff
@@ -50,7 +41,10 @@ const props = defineProps({
   },
 })
 
-const statusObject = computed(() => STATUS[props.status])
+const statusObject = computed(() => ({
+  title: i18n.t(`http.api.value.${props.status}`),
+  appearance: STATUS[props.status],
+}))
```
johncowen, line 47: Could we just use the http property as the classname i.e. `not_available` vs `not-available` (i.e. use `props.status`) then we can get rid of this extra code (and the above `STATUS`)? In fact this would almost be a pure javascript-less presentational component then

kleinfreund: Fine by me. I’ve changed this now.

src/locales/en-us/common/index.yaml
johncowen: Yeah I also wondered, me I'd say yes but not knowing exactly where they are used, but I figured I'd now just leave that for i18n folks to change in the YAML files if they want to, kinda not our decision anymore 🎉 . I'd do whatever feels right to you right now and then people who decide these things can change.

johncowen (approved):
Thanks for this!

PR #1031, opened 2023-06-08
chore: hides copy as kubernetes button for internal service insights

kleinfreund:
See https://github.com/kumahq/kuma/issues/6780#issuecomment-1582657955

### DIFF
--- a/src/app/common/ResourceCodeBlock.vue
+++ b/src/app/common/ResourceCodeBlock.vue
@@ -11,7 +11,10 @@
       :query-key="props.id"
       :code-max-height="props.codeMaxHeight"
     >
-      <template #secondary-actions>
+      <template
+        v-if="showCopyAsKubernetesButton"
+        #secondary-actions
+      >
         <CopyButton
           :get-text="getYamlAsKubernetes"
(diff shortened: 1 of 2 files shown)


johncowen (commented):
Just a question more than anything, but instead of making another type of switch for something from what I understand are already switchable (slots). Could we just override the default slot with an empty slot? Half asking as I'm not 100% sure this would be as straight forwards as I thought, maybe I missed something

kleinfreund:
> Just a question more than anything, but instead of making another type of switch for something from what I understand are already switchable (slots). Could we just override the default slot with an empty slot? Half asking as I'm not 100% sure this would be as straight forwards as I thought, maybe I missed something

While that’s technically possible, I don’t want to remove content by overriding it with ... empty content. Using a prop is much more clear in the intention and easier to use.

johncowen (approved):
ok, I don't mind as of this moment, but just saying I prefer smaller interfaces and more slot usage as they are then more flexible/reusable and easier to use.

Adding a switch for something that is already switchable is a bit weird, plus if anything it should be called something like `:has-secondary-actions`.

No action required here for the moment.

PR #1034, opened 2023-06-09
fix: zones not having correct breadcrumb title

kleinfreund:
**chore: fix providing superfluous route parameters**

Specifies the exact route parameters in the new breadcrumbs to avoid the vue-router warning about discarding invalid parameters.

**fix: zone egress breadcrumb using wrong route name**

**fix: zones not having correct breadcrumb title**

Adds the missing messages that are used zone detail view.

### DIFF
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -18,7 +18,11 @@ zones:
   routes:
     create:
       title: 'Create & connect Zone'
+    item:
+      title: Zone CPs
+      breadcrumbs: Zone CPs
     items:
+      title: Zones
       breadcrumbs: Zones
       navigation:
         zone-cp-list-view: Zone CPs
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -19,8 +19,6 @@ zones:
     create:
       title: 'Create & connect Zone'
     items:
-      title: Zones
-      breadcrumbs: Zones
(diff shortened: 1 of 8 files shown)


src/app/zones/locales/en-us/index.yaml
```diff
@@ -18,7 +18,11 @@ zones:
   routes:
     create:
       title: 'Create & connect Zone'
+    item:
+      title: Zone CPs
+      breadcrumbs: Zone CPs
```
lahabana, line 23: Does this even exist? Isn't it `zone-cps.routes.item` and there are already i18n entries for this? Same for items.title below.

src/app/zones/views/ZoneEgressDetailView.vue
```diff
@@ -1,16 +1,13 @@
...
     <RouteTitle
       :title="t('zone-egresses.routes.item.title')"
     />
     <AppView
       :breadcrumbs="[
         {
           to: {
-            name: 'zone-cp-list-view',
-            params: _route.params
+            name: 'zone-egress-list-view',
```
lahabana, line 10: Good catch!

src/app/zones/locales/en-us/index.yaml
kleinfreund: Oh, I didn’t notice the `zone-cps` entry existed. I’ve removed the unnecessary entries here and used the CP entry instead (for breadcrumbs).

johncowen (approved):

PR #1037, opened 2023-06-09
chore: Hard errors on missing i18n key, tmp adds tests for all titles

johncowen:
Firstly, the tests in this PR should fail without https://github.com/kumahq/kuma-gui/pull/1032 and therefore requires that in first.

I've added a hard error for when we are missing translation keys here which will fail the tests (by default formatjs only `console.error`s), this is possibly temporary as I'd rather not do this in our own i18n service but instead rely on configuring formatjs. I'm looking into that separately and I'd like to get these tests in.

Additionally the tests here are probably temporary or at least temporary in this file. There are several reasons why this is so:

1. Ideally we would have separate files for absolutely every page, with way more e2e tests in them. Right now its easy to bung all these in one file and loop through them, eventually we will have all the separate files and these can be moved over there.
2. We are doing lots of testing on human strings. This makes editing i18n a lot harder as you also need to change the test. Right now we just need a fairly good coverage of these things, so hitting every page and checking the title is a fairly good way to test all this out temporarily. I would imagine further down the line we will have a better approach so we avoid asserting on human text. Right now, this gives us what we want, and for the moment editing the titles also means editing the tests.

As an aside this also gives us a very good reference to look at when deciding what every page title should be instead of a mish-mash of different things (capitalisation/style etc) depending on who wrote the code/when the code was written.

### DIFF
--- a/features/application/Titles.feature
+++ b/features/application/Titles.feature
@@ -4,16 +4,43 @@ Feature: The HTML title is correct on each page
     Then the page title contains "<Title>"
 
     Examples:
-      | URL                                     | Title              |
-      | /                                       | Overview           |
-      | /diagnostics                            | Diagnostics        |
-      | /zones/zone-cps                         | Zone CPs           |
-      | /zones/zone-ingresses                   | Zone Ingresses     |
-      | /zones/zone-egresses                    | Zone Egresses      |
+      | URL                                                 | Title                         |
+      | /                                                   | Overview                      |
+      | /diagnostics                                        | Diagnostics                   |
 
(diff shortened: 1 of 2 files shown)


johncowen:
Following https://github.com/kumahq/kuma-gui/pull/1034/files#diff-b74e52e340de40ac28d9f03911b3150dd6e2a60629b55351d4d160735bda2fcbL22 this also requires https://github.com/kumahq/kuma-gui/pull/1039

johncowen:
@kleinfreund this is now green

kleinfreund (approved):

Issue #1029, opened 2023-06-07
Move meshes detail view inside tabs


lahabana:
### Description

Currently details pages are not embedded inside tabs. We should change things so that it works this way

PR #1040, opened 2023-06-09
chore: Moves Mesh detail pages back underneath tabs

johncowen:
Ideally I would have split this up into 2 separate PRs as there are two things here:

1. We originally decided that detail pages would not have tabs, we've since decided that they should have tabs, so this PR puts the Mesh Service, Dataplane, Gateway, Policy detail pages back underneath the tabs.
2. Instead of hardcoding more and more potential "this route means that this tabs should be selected". I'd rather do something more automatic, i.e. because of nested routing/route views we know that this thing is related to this tab (its a child of the route the tab is linked to), so select it. <= This is the bit I would rather have worked on first, and is "this is fine for now" so we can move on with this. e.g. child removal probably needs adding plus some other things. The eventual plan is this can be easily used for not just tabs but for other navigation elements also.

Eventually the extra code in `MeshView.vue` will be reduced to nothing.

I may add some inline comments

Closes https://github.com/kumahq/kuma-gui/issues/1029

### DIFF
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -86,13 +86,12 @@ if (!hasParent) {
   provide('route-view-parent', routeView)
 }
 
-const iParent: ImmediateParent | undefined = inject('route-view-immediate-parent', undefined)
+const iParent = inject<ImmediateParent | undefined>('route-view-immediate-parent', undefined)
 
 const immediateParent: ImmediateParent = {
   addChild: (module, sym) => {
     children.value.push(module)
     if (typeof iParent !== 'undefined') {
-      // @ts-ignore
       iParent.addChild(module, sym)
     }
   },
@@ -106,7 +105,6 @@ watch(() => props.module, (module = '') => {
     typeof iParent !== 'undefined' &&
     module.length > 0
   ) {
-    // @ts-ignore
     iParent.addChild(module, sym)
   }
 }, { immediate: true })
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -20,11 +20,12 @@
           ]
         }))
       }"
+      :children="children"
     />
   </div>
 </template>
 <script lang="ts" setup>
-import { provide, inject, ref } from 'vue'
+import { provide, inject, ref, watch } from 'vue'
 import { useRoute } from 'vue-router'
 
 import { useI18n } from '@/utilities'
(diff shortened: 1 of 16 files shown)


src/app/application/components/route-view/RouteView.vue
```diff
@@ -71,6 +86,31 @@ if (!hasParent) {
...
 }
 
+const iParent: ImmediateParent | undefined = inject('route-view-immediate-parent', undefined)
+
+const immediateParent: ImmediateParent = {
+  addChild: (module, sym) => {
+    children.value.push(module)
+    if (typeof iParent !== 'undefined') {
+      // @ts-ignore
+      iParent.addChild(module, sym)
```
johncowen, line 96: Typescript tells me this can `never` happen, javascript tells me otherwise.

Maybe I'm doing something wrong, but I didn't want to spend to much time looking into it right now, seeing as I'm coming back here after the PR is merged, but if its a simple "oh you just need to...." lmk

src/app/meshes/routes.ts
```diff
@@ -25,22 +25,26 @@ export const routes = (
     {
       path: '/mesh',
       name: 'mesh-index-view',
+      component: () => import('@/app/meshes/views/MeshView.vue'),
```
johncowen, line 28: This moves the tabs ontop of details as well as lists.

src/app/meshes/views/MeshOverviewView.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <RouteView>
+  <RouteView
+    module="meshes"
```
johncowen, line 3: Eventually these won't be hardcoded

src/app/meshes/views/MeshView.vue
```diff
@@ -35,31 +39,28 @@
...
+const data: Map<string, {module: string | undefined}> = new Map()
+const items: Tab[] = meshRoutes.map((item) => {
   const name = String(item.name)
+  data.set(`#${name}`, {
+    module: item?.meta?.module,
+  })
   return {
     title: t(`meshes.routes.item.navigation.${name}`),
-    hash: name,
+    hash: `#${name}`,
```
johncowen, line 63: Eventually I should be able to remove all this code from this file and just end up with the KTabs.

johncowen (commented):
Adding inline notes.

src/app/application/components/route-view/RouteView.vue
```diff
@@ -71,6 +86,31 @@ if (!hasParent) {
   provide('route-view-parent', routeView)
 }
 
+const iParent: ImmediateParent | undefined = inject('route-view-immediate-parent', undefined)
```
kleinfreund, line 89: **Suggestion**: This avoids the two TypeScript errors ignored in this file.

```suggestion
const iParent = inject<ImmediateParent | undefined>('route-view-immediate-parent', undefined)
```

kleinfreund (approved):

src/app/application/components/route-view/RouteView.vue
johncowen: >> Typescript tells me this can never happen, javascript tells me otherwise.

>>  Maybe I'm doing something wrong, but I didn't want to spend to much time looking into it right now, seeing as I'm coming back here after the PR is merged, but if its a simple "oh you just need to...." lmk

> Suggestion: This avoids the two TypeScript errors ignored in this file.

Ah of course! ty! Weird it only needs it here an not the one above, which I'm assuming would be better using the same form 🤔 . Lemme change this one up before merging.

PR #1046, opened 2023-06-12
refactor: internalizes data management for mesh charts

kleinfreund:
refactor: internalizes data management for mesh charts

Moves all the data fetching and chart data computation out of storeConfig and into MeshCharts as there is no advantage in having handled via the Vuex store.

Depends on (and includes) https://github.com/kumahq/kuma-gui/pull/997.

### DIFF
--- a/src/app/meshes/views/MeshOverviewView.vue
+++ b/src/app/meshes/views/MeshOverviewView.vue
@@ -122,6 +122,7 @@ import MeshResources from '@/app/common/MeshResources.vue'
 import ResourceCodeBlock from '@/app/common/ResourceCodeBlock.vue'
 import StatusInfo from '@/app/common/StatusInfo.vue'
 import TextWithCopyButton from '@/app/common/TextWithCopyButton.vue'
+import { mergeInsightsReducer } from '@/store/reducers/mesh-insights'
 import { useStore } from '@/store/store'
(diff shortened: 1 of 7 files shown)


src/app/meshes/views/MeshOverviewView.vue
```diff
@@ -202,6 +203,8 @@ async function loadMesh(): Promise<void> {
   try {
     mesh.value = await kumaApi.getMesh({ name })
     meshInsights.value = await kumaApi.getMeshInsights({ name })
+    const meshInsight = mergeInsightsReducer([meshInsights.value])
+    store.state.currentMeshPolicies = meshInsight.policies
```
kleinfreund, line 207: **Comment**: As far as I can tell, this is the last use of always loaded mesh insights data. It’s used in the NotificationManager and thus always needed as it’s in our top bar. I tried to keep the change here to a minimum to make it clear exactly _what_ part of mesh insights we’re using (i.e. the `policies` field).

johncowen (approved):
Similarly to https://github.com/kumahq/kuma-gui/pull/997 tough one to code review but I get why, I tried it out and everything seems fine. Thanks for adding that comment also, I can see where you are going.

As long as there is nothing here you want me to specifically give some extra 🦅 👀 , I'm good to 👍 this. I think reducing store usage for this is a good outcome.

PR #1048, opened 2023-06-12
chore: update zone creation instructions for secrets

kleinfreund:
**chore: removes token from k8s instructions**

**chore: change universal instructions to use token from file**

Changes the universal instructions for the Zone creation flow to store the secret in a file and to refer to that file in the CP config.

### DIFF
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -44,9 +44,13 @@ zones:
       errorTitle: 'Could not connect Zone'
       errorDescription: 'We were unable to connect Zone.'
     universal:
-      copySaveToken:
-        title: 'Copy and save token'
-        alertMessage: 'You won’t see the token again! Make sure to save it somewhere now.'
+      saveToken:
+        title: 'Save token'
+        saveTokenDescription: 'Save the token to a file.'
+        saveTokenCommand: |
+          mkdir -p ~/kuma-cp \
+            && echo {token} > ~/kuma-cp/cpTokenFile \
+            && chmod 600 ~/kuma-cp/cpTokenFile
       connectZone:
         title: 'Connect Zone'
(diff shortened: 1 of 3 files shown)


src/app/zones/locales/en-us/index.yaml
```diff
@@ -62,8 +66,8 @@ zones:
             kdsDeltaEnabled: true
         connectDescription: 'Next, download Kuma and connect the Zone'
         connectCommand: |
-          (export VERSION=0.0.0-preview.vabf3cdb72; curl -L https://kuma.io/installer.sh | sh - \
-            && kuma-$VERSION/bin/kuma-cp run -c config.yaml)
+          curl -L https://kuma.io/installer.sh | VERSION=0.0.0-preview.vabf3cdb72 sh - \
+            && kuma-0.0.0-preview.vabf3cdb72/bin/kuma-cp run -c config.yaml
```
johncowen, line 70: Not sure how worth it it would be but should we be passing these in as a variable sourced from either `env` or `t`, do we have these values already in `env`? I think we have version, but maybe not the hash over here. The hash might be worth adding over here also? We'd probably use it for other things.

Anyway no action required, just half thinking out loud half asking what you think.

johncowen (approved):
LGTM, I had a question but definitely not blocking in any way.

src/app/zones/locales/en-us/index.yaml
kleinfreund: That’s not worth the effort because this is a temporary measure as far as I understand it. The long term variant shouldn’t include a specific version at all.

For the time being, this being here is extra beneficial because it means other folks can update it in the language files.

johncowen: Gotcha, 

> For the time being, this being here is extra beneficial because it means other folks can update it in the language files.

Well ideally they wouldn't have to, it would automatically change.

Anyway, all good, this is tmp so 🤷‍♂️

PR #1063, opened 2023-06-13
chore(e2e): expands Zone creation test

kleinfreund:
**chore(e2e): renames zone test file**

**chore(e2e): expands Zone creation test**

Adds assertions around the entity scanner content.


johncowen (commented):
Am I right in thinking some of this stuff might be useful here:

https://github.com/kumahq/kuma-gui/blob/83ad80d1ca3596fbea7037bc9b92488f8baf596c/features/application/Loading.feature#L10-L21

Also could we find a way to not assert on human text if possible?

kleinfreund:
> Am I right in thinking some of this stuff might be useful here:

We should really avoid using this in tests as much as possible because it will make the tests actually go through the added latency (as it uses `setTimeout`). In the case of the additions in this PR, just switching the response to return the right thing works and is fast so as far as I understand, introducing latency here would just make the test slower than it needs to be.

johncowen:
> We should really avoid using this in tests as much as possible because it will make the tests actually go through

Hmm yeah, I think the loading one linked requires the latency to be able to test the loading state reliably (we want to make sure it doesn't immediately load).

Let me look at this a little more, theres something about it that doesn't sit well with me, its around that fact that the state changes from:

```cucumber
Then the "$zone-connected-scanner" element contains "Waiting for Zone to be connected"
```

to:

```cucumber
Then the "$zone-connected-scanner" element contains "The Zone “test” is now connected"
```

without you actually interacting with anything.

If possible could we still look at not asserting on human text in the meantime? I also left a couple of little nits seeing as we tweaking a little bit anyway.

johncowen:
Yeah exactly I'm not sure if that sits with me well because that isn't clear from the test, wondering if there is an improvement.

It looks like its a 1 second poll, so this only works as a test as we have a default wait of 4 seconds for something to pass/fail an assertion, is that right? Otherwise you would have to add the 1 - 2 second wait into the test?

Hmm yeah I just added a KUMA_LATENCY of 5000 and it failed, I know we aren't adding a latency here, but could we use `the wait for` step just before your last assertion? That means we are Arranging, Acting (waiting) and then Asserting.

If you are super strong you don't want to do this then fine, but I think it would make the test read better and would have prevented me from going "what!? this doesn't make sense!", and I don't think it has any consequences seeing as we secretly wait for 4000 in Cypress anyway.

johncowen (approved):
Lets get this in, I'm gonna be working around here soon so I'd rather it was in and I know you've moved off of tests for a little while

(just noticed the 2 nits here were hanging out in Pending as I hadn't done a proper review, doesn't matter if you address them here or not)

PR #1064, opened 2023-06-13
chore: Set top-level classes from RouteView

johncowen:
Adds functionality to be able to set top-level class names from our `RouteView` component

Previously we were doing this on a per-class basis i.e. `setWizardPageClass`. This PR does pretty much the same thing just at a more generic/reusable level and handily from our `RouteView` component which is available in every single view/route component.

At an application level, I went with `is-fullscreen` which better describes what the required CSS here does.

Notes:

- Whilst the plan is to be able to set any attrs on the root node, I haven't done this yet as we don't currently need it. This only currently works for `class`
- As all this code is centralized in `RouteView` I went with "I'm happy with the interface, lets get the thing working" whilst not worrying to much about the fact that RouteView is getting a little messier than I would like. There is still one more thing to finish off in `RouteView` and I'd rather finish that up before a cleanup/de-dupe and make the code in RouteView more consistent.

### DIFF
--- a/src/assets/styles/_utilities.scss
+++ b/src/assets/styles/_utilities.scss
@@ -31,3 +31,13 @@
   border: none;
   cursor: pointer;
 }
+html:not(.is-fullscreen) .app-content-container {
+  padding-top: var(--AppHeaderHeight, initial);
+  display: grid;
+  grid-template-columns: var(--AppSidebarWidth) 1fr;
+}
+
+html:not(.is-fullscreen) .app-main-content {
+  padding: var(--AppGap);
+}
+
--- a/src/assets/styles/_utilities.scss
+++ b/src/assets/styles/_utilities.scss
@@ -31,13 +31,3 @@
   border: none;
   cursor: pointer;
 }
-html:not(.is-fullscreen) .app-content-container {
-  padding-top: var(--AppHeaderHeight, initial);
-  display: grid;
-  grid-template-columns: var(--AppSidebarWidth) 1fr;
-}
-
-html:not(.is-fullscreen) .app-main-content {
-  padding: var(--AppGap);
-}
-
--- a/src/assets/styles/_variables.scss
+++ b/src/assets/styles/_variables.scss
@@ -10,6 +10,7 @@
   --AppHeaderHeight: 60px;
   --AppSidebarWidth: 280px;
   --AppGap: var(--spacing-lg);
+  --AppDisplay: grid;
 
   // ONBOARDING
   --onboarding-accent: #822dc5;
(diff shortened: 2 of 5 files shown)


src/assets/styles/_utilities.scss
```diff
@@ -31,3 +31,13 @@
...
 }
+html:not(.is-fullscreen) .app-content-container {
+  padding-top: var(--AppHeaderHeight, initial);
+  display: grid;
+  grid-template-columns: var(--AppSidebarWidth) 1fr;
+}
+
+html:not(.is-fullscreen) .app-main-content {
+  padding: var(--AppGap);
+}
```
kleinfreund, line 42: **Suggestion**: I think these styles should be defined as close as possible to where they’re being used and that’s App.vue and AppView.vue.

johncowen: Hmm yeah not totally sure, they aren't being used in either of those components anymore, plus if we decided we wanted to put them into `RouteView` that is putting our specific application concerns into "framework components", and if we need to do this more and more, then `RouteView` is only going to get full of unrelated CSS.

I did wonder where was the best place to put them and saw that we had utility CSS things in this file, this thing being a 'utility'. Do you have any other ideas where this could go?

kleinfreund: > Hmm yeah not totally sure, they aren't being used in either of those components anymore

But ... they are? App.vue uses `app-content-container` on line 14 and AppView.vue uses `app-main-content` on line 4.

And those respectively are the key elements of the selectors in play here. Those components are exactly where those rules should live. The `html:not(.is-fullscreen)` is irrelevant here as this CSS is for the elements targeted by the key selector. `html:not(.is-fullscreen)` is just modifying when they apply.

Also, these are really not utility rules. This file is for re-usable utilities.

johncowen: Oh I see what you mean now! Sorry I thought you were talking about `is-fullscreen`, one sec lemme have another think.

johncowen:
I also just realized that none of this was necessary here seeing as Zone creation is feature flagged... Ah well at least we have this more generic "set the root class" util now.

src/assets/styles/_variables.scss
```diff
@@ -39,3 +40,11 @@
   --KCardBackground: var(--white);
   --KCardBorder: 1px solid var(--grey-300);
 }
+:root.is-fullscreen {
+  --AppHeaderHeight: 0;
+  --AppSidebarWidth: 0;
+  --AppGap: 0;
```
kleinfreund, line 46: **Suggestion**: I suggest to introduce a new property `--AppContentPadding` specifically for the use on `.app-main-content`. In hindsight, I think it was a mistake to use `--AppGap` for this element in particular. The reason for that is that it’s used in a couple of other places to, for example, space out cards.

So then we have:

- `--AppContentPadding` for `.app-main-content`
- `--AppGap` for the current uses

johncowen: Oh hey, sorry for the delay, totally didn't realize this was here! This has been changed now, lmk!

kleinfreund (approved):

PR #1073, opened 2023-06-16
fix(charts): avoid NaN error on hover

johnharris85:
Handle division by zero cases for doughnut chart hover tooltip


lahabana:
Should this be backported in 2.3?

kleinfreund:
@lahabana Yes. I've created a PR for that here: https://github.com/kumahq/kuma-gui/pull/1084

PR #1089, opened 2023-06-20
chore: expands test helpers

kleinfreund:
**chore(e2e): adds "requested N times" helper**

**chore(e2e): support clearing input elements**

### DIFF
--- a/cypress/support/step_definitions/index.ts
+++ b/cypress/support/step_definitions/index.ts
@@ -77,10 +77,12 @@ When(/^I click the "(.*)" element(?: and select "(.*)")?$/, (selector: string, v
 })
 
 When('I {string} {string} into the {string} element', (event: string, text: string, selector: string) => {
+  const method = text === '' ? 'clear' : 'type'
+
   switch (event) {
     case 'input':
     case 'type':
-      $(selector).type(text)
+      $(selector)[method](text)
       break
   }
 })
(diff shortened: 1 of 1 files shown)


kleinfreund:
I’ve removed `refactor(e2e): adds ability to pass config object to mocks` for the time being as I need the other two additions for tests I’m writing.

cypress/support/step_definitions/index.ts
```diff
@@ -77,10 +77,12 @@ When(/^I click the "(.*)" element(?: and select "(.*)")?$/, (selector: string, v
 })
 
 When('I {string} {string} into the {string} element', (event: string, text: string, selector: string) => {
+  const method = text === '' ? 'clear' : 'type'
```
johncowen, line 80: nit:

`And I clear {string} into the "input" element`

Sounds a bit weird, also at second glance is that even right? Anyway, is there anyway we can make this read better, either regex acrobatics or a separate step would be fine.

kleinfreund: Oh, no. The intended usage of this would be:

> And I type "" into the "$input" element

But I now realize that that’s ambiguous because it could mean "I add the empty string to the end of input". Maybe it should just be a separate step.

kleinfreund: I’ve changed this to a separate step now.

johncowen (approved):
Looks good to me! Thanks for holding off on the ability to pass config object to mocks for the time being 🙏 . Do bring it up again if we absolutely need it in the short term though, will maybe look at moving the Cypress mocking to reuse the same code as MSW sooner rather than later.

PR #1098, opened 2023-06-21
chore: expands zone tests to assert content

kleinfreund:
Creates two new test suites for zone list and detail views that assess their content. Some of the content of the page navigation test was integrated into them and removed from that test.

Removes all snapshot-based tests for zones.


src/app/zones/views/ZoneListView.spec.ts
```diff
@@ -15,137 +15,6 @@ describe('ZoneListView', () => {
...
-
-    expect(wrapper.html()).toContain('dpToken')
-
-    await wrapper.find('#insights-tab').trigger('click')
-    expect(wrapper.find('[data-testid="tab-container"]').element).toMatchSnapshot()
-  })
-
   test('updates list when deleting Zone', async () => {
     const firstZoneName = 'zone-1'
     mock('/zones+insights', {
```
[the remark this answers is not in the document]
kleinfreund: Not sure, but it's perfectly fine to keep this as a unit test

features/zones/Create.feature
```diff
@@ -68,5 +68,6 @@ Feature: The create Zone flow works
         zoneInsight:
           subscriptions:
             - connectTime: '2020-07-28T16:18:09.743141Z'
+              status: {}
```
johncowen, line 71: How comes we added this here? Is `status` always an empty object immediately on zone creation? I also noticed we do this through this PR, does doing that have any consequences for what we are covering in the test?

johncowen (approved):
LGTM!

I had a question just out of curiosity, but all looks fine to me.

src/app/zones/views/ZoneListView.spec.ts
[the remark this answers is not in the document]
johncowen: Well I think my point is its not really a unit test is it? Its not really testing a single isolated unit of source code, rather the a single feature of the entire application's code base.

I looked at it and didn't see why it wouldn't be possible, so just wondered, we can move this to where it should be in a follow up though.

features/zones/Create.feature
kleinfreund: There some pages that access this object (via `SubscriptionDetails`) and setting this as an empty object is the most minimal way of getting the errors out of those tests.

johncowen: So when you say "getting the errors out of those tests" do you mean "getting errors out" as in making the thing error, or making sure it doesn't error? I think its the latter but just checking

kleinfreund: Making sure the application code doesn't throw an exception on account of accessing `status.property` (as `status` would be wholly absent from the mock data without this change).

johncowen: Hmm I might catch up with you elsewhere on this one, just so I follow completely.

PR #1099, opened 2023-06-21
chore: removes mesh pro tip banner

### LINKED ISSUE #1044: remove mesh pro tip banner
### What happened?

It doesn't really display at the right places and also doesn't bring a lot of info.

Let's remove and we'll rethink later about what we want to do here.

kleinfreund:
Removes all code related to the mesh pro tip banner.

Resolves #1044.


johncowen (commented):
Maybe I misunderstand this PR/the scope of this PR but aren't we only supposed to be remove the Pro tip banner:

Screenshot 2023-06-22 at 09 57 36

Lemme know, maybe I misunderstood

kleinfreund:
@johncowen Oh dang, of course. I completely unified those two things into one when doing this PR. I’ll change this to _only_ remove the mesh pro tip alert. 😬

kleinfreund:
Sorry, my bad. Updated now.

src/app/notification-manager/components/NotificationManager.spec.ts
```diff
@@ -44,9 +21,7 @@ describe('NotificationManager.vue', () => {
       },
     })
 
-    const openModalButton = wrapper.find('[data-testid="open-modal-button"]')
-    expect(openModalButton.html()).toContain('Check your mesh!')
-    await openModalButton.trigger('click')
+    await store.dispatch('notifications/openModal')
```
johncowen, line 24: Would be nice to have an e2e test to ensure clicking the little bell shows this modal. Maybe a follow up.

johncowen (approved):
LGTM, left one comment but would be a follow up if anything.

PR #1101, opened 2023-06-22
fix: not cards missing border radius

kleinfreund:
## Changes

Fixes cards often missing the intended border radius on account of a custom property being set incorrectly.

Also adds a bottom margin to `NavTabs` as the double-border issue would now look particularly bad.

## Screenshots

Before:

Screenshot from 2023-06-22 09-56-13

After:

Screenshot from 2023-06-22 09-56-03


johncowen (approved):
I think this is fine for now. Good spot on the missed `var(` although weirdly I thought this went in a while back 🤷

PR #1102, opened 2023-06-22
chore: disables logger in prod, enables console logger in dev

kleinfreund:
Disables the logger service in prod. Enables the logger service in dev.


johncowen (approved):
Confirmed that I no longer see the `console.logs` when running our Cypress tests (which use a production build)

PR #1105, opened 2023-06-22
chore: Removes last vestiges of old store based titling

johncowen:
I saw some bits of the old store based titling still existed but didn't really do anything anymore, so I went through and cleaned up.

While I was there I made it so all "detail" pages use the form `name-of-the-specific-thing Type of thing` (e.g. `program-0 Data plane proxy`) consistently. The shouldn't/doesn't require test amends seeing as we check titles contain the text rather than equals the text.

Just to note I much like the fact that we are getting rid of more code here and its much easier and more obvious how to set the title of the page, a small thing but all these little small improvements will gradually add up.

### DIFF
--- a/src/app/data-planes/locales/en-us/index.yaml
+++ b/src/app/data-planes/locales/en-us/index.yaml
@@ -1,7 +1,7 @@
 data-planes:
   routes:
     item:
-      title: Data plane proxy
+      title: "{name} Data plane proxy"
       breadcrumbs: Data plane proxies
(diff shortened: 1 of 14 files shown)


kleinfreund:
**Suggestion**: Perhaps we should also add "Mesh" to the mesh name in the document title.

johncowen:
👍 added!

kleinfreund (approved):

PR #1108, opened 2023-06-23
chore: disables logger in test setup

kleinfreund:

### DIFF
--- a/src/services/testing.ts
+++ b/src/services/testing.ts
@@ -1,7 +1,9 @@
 import { setupServer } from 'msw/node'
 
+import createDisabledLogger from './logger/DisabledLogger'
 import { Alias, ServiceConfigurator, token } from './utils'
 import CliEnv from '@/services/env/CliEnv'
+import Logger from '@/services/logger/Logger'
 import { mocker, fakeApi, FS } from '@/test-support'
 import type { Mocker } from '@/test-support'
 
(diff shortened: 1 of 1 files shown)


src/services/testing.ts
```diff
@@ -10,6 +12,11 @@ const $ = {
 }
 
 export const services: ServiceConfigurator = (app) => [
+  [token<Logger>('logger'), {
+    service: createDisabledLogger,
+    decorates: app.logger,
```
johncowen, line 17: Just wanted to note, that ideally decorators should decorate the original service, i.e. augment it or in this case disable just the bits we need to disable.

If we were to add functionality to the original service we would lose it here (or have to add it in multiple places).

At the moment we `extend Logger` which means we _do_ get any extra functionality from `Logger`, which happens to be fine for this case. But, if we needed to keep functionality from `Logger` and elsewhere use say `DatadogLogger` then this would begin to fall apart as we are using inheritance over composition and the disabled logger elsewhere would extend `Logger` not `DatadogLogger`.

If we decorated _whatever logger is_, so here it would be `Logger` and elsewhere it would be `DatadogLogger`, disabledlogger would still be an instance of `DatadogLogger`

Just to be clear, in this specific case it doesn't really matter (at the moment), but in others it does, and has, which is why we added decorators originally.

Also not a blocker just a note!

johncowen (approved):
Thanks!

src/services/testing.ts
kleinfreund: In other words, `createDisabledLogger` would take the underlying logger instance as an argument to create a disabled one for whatever is in use, right?

johncowen: Yeah something like that, there are a few ways to go about it I guess, but yeah basically decorate the original thing rather than completely replace.

PR #1109, opened 2023-06-23
chore(e2e): adds assertion option for exact request contents

kleinfreund:
Adds an optional "only" keyword to allow the `the URL "..." was requested with` step to assert exact request contents (e.g. when utilizing PATCH requests).


johncowen (approved):
Nice 🚢

PR #1110, opened 2023-06-23
refactor: types service tokens

kleinfreund:
## Changes

Exports a type for a service’s _specific_ tokens so that other services can easily create a `ServiceConfigurator<SupportedTokens>` type for their own service definition. This makes it much easier to work with composed services and stops us from needing to add artificial `service: Token` types just to satisfy the TypeScript compiler.

## Notes

- Initially I tried adding the production service to the Cypress `build` call, but that leads to errors related to stuff Cypress can’t handle presently (i.e. in our setup: the presence of .vue files in the router). However, now that the `build` calls will have stricter type checking for the services and their tokens, leaving it out causes errors. For now I `@ts-ignore`d the two erros in the Cypress entrypoint. Not great, might work.

### DIFF
--- a/src/services/development.ts
+++ b/src/services/development.ts
@@ -4,8 +4,9 @@ import Logger from './logger/Logger'
 import cookied from '@/services/env/CookiedEnv'
 import type Env from '@/services/env/Env'
 import debugI18n from '@/services/i18n/DebugI18n'
+import type { ProductionTokens } from '@/services/production'
 import { token, get } from '@/services/utils'
-import type { ServiceConfigurator, ReturnDecorated, Decorator, Alias, Token, TokenType } from '@/services/utils'
+import type { ServiceConfigurator, ReturnDecorated, Decorator, Alias, TokenType } from '@/services/utils'
 import type { FS } from '@/test-support'
 import { fakeApi } from '@/test-support'
 import { fs } from '@/test-support/mocks/fs'
@@ -23,17 +24,11 @@ const $ = {
   fakeFS: token<FS>('fake.fs'),
   kumaFS: token<FS>('fake.fs.kuma'),
 }
-type I18n = ReturnType<typeof debugI18n>
-type SupportedTokens = {
-  Env: Token
-  EnvVars: Token
-  i18n: Token
-  logger: Token
-  msw: Token
-  bootstrap: Token
-  env: Token<Alias<Env['var']>>
-}
+export type DevelopmentTokens = typeof $
 
+type I18n = ReturnType<typeof debugI18n>
+
+type SupportedTokens = ProductionTokens & DevelopmentTokens
 export const services: ServiceConfigurator<SupportedTokens> = (app) => [
 
   [token<Decorator<typeof app.bootstrap>>('bootstrap.with.mockServer'), {
--- a/cypress/main.ts
+++ b/cypress/main.ts
@@ -1,16 +1,20 @@
 import { TOKENS as DEV_TOKENS, services as development } from '@/services/development'
-import { TOKENS, services as e2e } from '@/services/e2e'
-import { build, token } from '@/services/utils'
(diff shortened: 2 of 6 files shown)


src/services/development.ts
```diff
@@ -23,17 +24,11 @@ const $ = {
...
-  Env: Token
-  EnvVars: Token
-  i18n: Token
-  logger: Token
-  msw: Token
-  bootstrap: Token
-  env: Token<Alias<Env['var']>>
-}
 
+type SupportedTokens = ProductionTokens & DevelopmentTokens
```
johncowen, line 31: I know that this is in draft, but wanted to point something out.

This approach now means that all these files depend on each other in something similar to an inheritance chain, something we are trying to avoid.

I know in this case its 'only' the types but still, these files are supposed to be as independent as possible apart from what is passed/injected into them.

I also find the SupportedTokens thing slightly tedious and I'd rather not have to do that, but 🤷 

Just wanted to mention my thoughts here incase it saves unnecessary work. Open to hearing other opinion also.

johncowen: P.S.

> needing to add artificial service: Token types just to satisfy the TypeScript compiler.

This is the thing I also find tedious ^ so I get you, just I'm cautious of getting into the non-injected dependency thing

kleinfreund: Okay, but the service definitions routinely and purposefully access services from other services that they depend on. For example, the development service depends on the production service and uses (via decorators, for example) its services to extend them. So that means `app` in those service definitions needs the type of the "previous" services.

Right now, the services are just really not independent at all.

So how else can we tell the services that services exist that they can rely on without constructing the types as I’m doing it here?

The alternative would be to completely redefining what’s in the previous services.

johncowen: Ok let me consider this a bit more, I more wanted to raise a hand to give my opinion on this sooner rather than later.

---

I think the main thing to think about is these files should be able to exist on their own as 'plugins' for example development  doesn't have to support all the tokens in production and it could also support tokens that aren't in production, but they are elsewhere.

As long as you know now rather than later that this isn't as straightforwards for me as it seems. It's _so_ not straightforwards that I'm unsure how strong I am on this. If you are happy to keep going with it with that in mind thats fine, I'm not ruling it out here and now.

I think the other thing to consider is. This is all pretty much done, all our development/testing stuff is setup now and there won't be much to add but tweaks moving forwards so value-wise its not adding much apart from itching a scratch.

TLDR; If you are keen on this then happy to consider it properly and possibly approve once you are done, not ruling it out and I might have my mind completely changed.

kleinfreund: The question essentially is this: how can a service container definition file know about the types it may access?

It needs to be able to do that and so far we’ve coped by manually defining the types with "anonymous" tokens in those files. This is really brittle because it easily becomes out of date and doesn’t relate to the actual source of their definition (unless we always, _always_) keep them up-to-date.

Now by looking at what our containers are today, they’re effectively a merge of a couple of service containers where the container that is defined later overrides or appends the tokens and services of those which are defined earlier.

I generally agree with the notion that these containers should be able to exist on their own; however, I don’t agree with this detail:

> for example development doesn't have to support all the tokens in production

Specifically for the relationship production-other container, I think this cannot be true. Everything that exists in the production container is used by the application (otherwise it can be removed). Therefore, containers constructed with it **must** include it. This currently happens based on how we merge service containers. And as far as I can tell it’s currently not even possible to delete/undefine a service coming from another container.

So, going back to the root question I posed: how can a service container definition file know about the types it may access?

There are pretty much two options: (1) the container does not even access other containers. That doesn't seem feasible. Decorating services relies on this being possible. (2) the container knows which services it may decorate.

For the latter, we need those types to come from somewhere.

I’m not sure how I address this in a different way. It’s definitely true that what I propose here ties the containers together based on their order of merging that should be unknown inside the definition file, but ... it does provide the type safety and addresses the maintenance issue.

PR #1122, opened 2023-06-26
feature: Adds DataSource component and uses it in PolicyListView

johncowen:
An alternative approach to externally sourced data.

Supports:

- 'background reloading' a.k.a Stale-while-revalidate (implemented here)
- Request reuse/sharing. Only make HTTP requests once when required.(implemented here)
- Response sharing. Update stale data in one area of the page if a new request is made for another area of the page.(implemented here)
- Polling (implemented here), and/or other live data approaches such as long polling or Server Sent Events
- Error backoff/retry
- Lazy loading. Only load data when the UI for it comes into view i.e. during scrolling
- Running in a WebWorker. Optionally re-shape massive amounts of data without blocking the main thread.

Probably other things not at the top of my head rn. Note I've pulled something out to make a more minimal PR for the moment, so when I say "support" not all of the above is implemented but the ground work is there to do it when/if we ever need.

I redid the Policy Listing page for a PoC/test as it has reasonably common application functionality, I've also implemented this slightly elsewhere where we have issues with syncing data in difference parts of the page using on one HTTP request.


src/app/meshes/sources.ts
```diff
@@ -0,0 +1,17 @@
...
+} from '@/types/index.d'
+type MeshParams = {
+  mesh: string
+}
+export type MeshInsightSource = DataSourceResponse<MeshInsight>
+export const sources = (api: KumaApi) => {
+  return {
+    '/:mesh/insights': async (params: MeshParams, source: {close: () => void}) => {
+      source.close()
+      return api.getMeshInsights({ name: params.mesh })
```
johncowen, line 14: Wanted to note that to avoid too much change in this PR I'm just reusing `KumaApi` as a HTTP client. In the future we can slowly move away from a large single API client and towards 'module' specific "sources of data". We can eventually pass in a lighter `request` HTTP client into where we are currently passing `api: KumaApi`, but we can do that incrementally as tech debt work.

All that is required for a `Source` _right now_ is that it returns a Promise. That Promise could be generated by anything, `fetch`, a change to `localStorage` value, or a change to a users session. Currently, if you want it to poll, you just remove the `source.close` and let the source stay open keep 'requesting' (this is potentially a temporary way of polling to avoid adding too much at once)

src/app/application/services/data-source/DataSource.ts
```diff
@@ -0,0 +1,83 @@
...
+      // latency
+      // await new Promise(resolve => setTimeout(resolve, 5000))
+      yield route.route({
+        ...{
+          offset: parseInt(queryParams.get('offset') || '0'),
+        },
+        ...route.params,
+      }, this)
+      if (!isClosed(this)) {
+        await new Promise(resolve => setTimeout(resolve, 5000))
```
johncowen, line 25: If we aren't closed after a `Source` resolves its promise, then we are polling, at the moment the poll interval is hardcoded to 5s.

src/app/application/services/data-source/DataSource.ts
```diff
@@ -0,0 +1,83 @@
...
+    const requestRouter: Router<Source> = new Router(requests)
+
+    this.pool = new SharedPool<string, EventSource>(
+      (src: string) => {
+        return get(src, requestRouter)
+      },
+      (src: string, source: EventSource) => {
+        return destroy(src, source)
+      },
+    )
```
johncowen, line 50: The SharedPool manages request sharing and therefore response sharing and knows how to acquire (and initiate, if its the first instance) and teardown (if its the last instance) connections/Sources. SharedPool keeps a record of how many things are currently using a particular `Source`. Once nothing is using a `Source` anymore it gets torn down/destroyed.

src/app/application/services/data-source/Router.ts
```diff
@@ -0,0 +1,21 @@
...
+    for (const [pattern, route] of this.routes) {
+      const _url = `data:${path}`
+      if (pattern.test(_url)) {
+        const args = pattern.exec(_url)
+        return { route, params: args?.pathname.groups }
+      }
+    }
+    throw new Error(`Matching route for '${path}' not found`)
+  }
+}
```
johncowen, line 21: We use this class currently in testing, and I might have edited it slightly since I wrote this. They'll both need matching up and reusing instead of having two separate Router classes.

src/app/meshes/index.ts
```diff
@@ -1 +1,25 @@
...
+      arguments: [
+        app.api,
+      ],
+      labels: [
+        app.sources,
+      ],
+    }],
+  ]
+}
+export const TOKENS = $
```
johncowen, line 25: `app/*` things are now becoming proper 'modules'/'plugins' so you can just `build()` up service configuration for infinite amounts of plugins. Importantly you shouldn't have to satisfy/use all 'external' tokens, `undefined` tokens will eventually mean more or less a noop.

src/app/policies/sources.ts
```diff
@@ -0,0 +1,44 @@
...
+import type KumaApi from '@/services/kuma-api/KumaApi'
+import type {
+  PolicyEntity as Policy,
+  PolicyType,
+} from '@/types/index.d'
+export interface CollectionResponse<T = Record<string, unknown>> {
+  total: number,
+  items: T[]
+  next: string | null
+}
```
johncowen, line 11: I think we might have a thing for this already, we should try and merge/reuse if we do.

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,78 @@
...
+}
+</script>
+
+<template>
+  <slot
+    :data="message as any"
+    :error="error"
+    :refresh="refresh"
+  />
+  <data />
```
johncowen, line 77: This `data` HTML element will allow us to do lazy loading.

johncowen (commented):
I added some inline notes considering its in draft but still looking for some initial 👀 , hope they help

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,78 @@
...
+const data = useDataSource()
+
+const props = defineProps({
+  src: {
+    type: String,
+    required: true,
+  },
+})
+
+const message = ref<unknown | undefined>(undefined)
```
kleinfreund, line 16: **Suggestion**: This shouldn’t include the type `undefined` as `unknown | undefined` is just `unknown`.

```suggestion
const message = ref<unknown>(undefined)
```

src/app/application/services/data-source/CallableEventSource.ts
```diff
@@ -0,0 +1,54 @@
...
+          }
+        }
+        self.readyState = 2
+      } catch (e) {
+        console.error(e)
+        self.dispatchEvent(new ErrorEvent('error', {
+          error: e,
+        }))
+      }
+    })(this)
```
kleinfreund, line 42: **Question**: Why is this wrapped in an IIFE? As far as I can tell this would work just the same without it.

src/app/application/services/data-source/DataSource.ts
```diff
@@ -0,0 +1,83 @@
...
+      // await new Promise(resolve => setTimeout(resolve, 5000))
+      yield route.route({
+        ...{
+          offset: parseInt(queryParams.get('offset') || '0'),
+        },
+        ...route.params,
+      }, this)
+      if (!isClosed(this)) {
+        await new Promise(resolve => setTimeout(resolve, 5000))
+      }
```
kleinfreund, line 26: **Issue**: This duplicates a polling implementation. Can you please look into whether you can reuse `src/utilities/poll.ts`?

src/app/policies/sources.ts
kleinfreund: Yes, we do. Please reuse the types from `src/types/api.d.ts`.

src/app/policies/sources.ts
```diff
@@ -0,0 +1,44 @@
...
+type PaginationParams = {
+  offset: number
+}
+
+type PolicyTypeParams = {
+  policyType: string
+}
+// type PolicyParams = {
+//   policy: string
+// }
```
kleinfreund, line 25: **Suggestion**: Please move types like this into `src/types/api.d.ts`.

kleinfreund:
> * Request reuse/sharing. Only make HTTP requests once when required.(implemented here)

Is it controllable on the level of usage where this logic is side-stepped? It’s likely that there are cases where we know we don’t want to re-use a request and not employ this logic.

> * Response sharing. Update stale data in one area of the page if a new request is made for another area of the page.(implemented here)

How can we control whether data is shared? I can think of use cases where updating data across the application in the background would be completely unexpected (e.g. in a form context where existing resources are being loaded. Those being updated in the background should not affect the current context).

> * Polling (implemented here), and/or other live data approaches such as long polling or Server Sent Events

We should reuse our existing poll implementation.

src/app/application/components/data-source/DataSource.vue
johncowen: Oh weird I thought I'd added an inline comment here, it's so we can do lazy loading i.e. so we know when the DataSource is in the viewport. Can be any HTML element really, but I've always considered populating the `data` element with the actual data, then it would work with native `select`/`data` interactions (not saying I'm gonna do that at all.

kleinfreund: So the intention is to not actually populate this element whatsoever but rather just make it "browser visible" so that, potentially, mechanisms like IntersectionObserver can do visibility checks?

If so, then this should definitely just be a `<div />` I think.

src/app/application/services/data-source/DataSource.ts
johncowen: I'd rather use generators. That means we can check for whether it should stop in between individual HTTP requests within a single poll loop if we need to (this probably includes backoff/retries etc)

This will replace `src/utilities/poll.ts`, not the other way around.

kleinfreund: If so, then please externalize it to not prevent other code from using it.

src/app/policies/sources.ts
johncowen: Yeah for sure. Certain things will be moved as part of the WIP/before the end of the WIP. But certain single use types won't be. We also need to consider the modular nature of things. Eventually lot of this will be autogenerated into a specific place anyway so we don't have to hand write them.

src/app/application/components/data-source/DataSource.vue
johncowen: `div` s are more likely to get styled by accident. `data` not so much. I actually don't think it matters that much apart from that, plus the fact that its in one place and takes all of 2 seconds to change with not a lot of impact on the codebase.

All in all, unless you can think of any other downsides to using `data` I'd rather keep it as that if only to be less likely to be accidentally styled.

johncowen: >So the intention is to not actually populate this element whatsoever but rather just make it "browser visible" so that, potentially, mechanisms like IntersectionObserver can do visibility checks?

At the moment, yes we won't be populating it right now. But if we want to we can, as I said it might be useful for native `select/data` and other interactions. but yes, IntersectionObserver for the moment (which we don't have here but probably will add later when we first need it)

kleinfreund: > `div` s are more likely to get styled by accident. `data` not so much.

How? We control that and we don’t have a single instance of selecting a descendant `div` in our code base.

It should just be a `div` because it’s completely non-semantic, never will carry any actual data and we also won’t set the `data` element’s `value` attribute. There is no reason to use `data` here.

johncowen:
> Is it controllable on the level of usage where this logic is side-stepped? It’s likely that there are cases where we know we don’t want to re-use a request and not employ this logic.

No but we can make it so if we ever need to. I can't think of a case where we would want to re-request an endpoint when we haven't yet received a response from a previous request. Could you give me an example of a likely use case?

> How can we control whether data is shared? I can think of use cases where updating data across the application in the background would be completely unexpected (e.g. in a form context where existing resources are being loaded. Those being updated in the background should not affect the current context).

Aren't there Vue things that allow us to do this? Apart from that you could make a separate uri for the DataSource, that way the data isn't shared

> (e.g. in a form context where existing resources are being loaded. Those being updated in the background should not affect the current context).

If you are loading data for a form, where else would you be sharing the data? Not sure I follow the entire use case? Usually there are only a few examples of where the sharing of data is useful.

> We should reuse our existing poll implementation.

Why?

src/app/application/services/data-source/CallableEventSource.ts
johncowen: I don't want `_open` to return anything and I need `async`.

kleinfreund: Why?

src/app/application/components/data-source/DataSource.vue
johncowen: Ah I think this was when I was messing around with generic components, I've taken the `undefined` out for now.

kleinfreund:
> Why?

Because we shouldn’t have two poll implementations in the code base.

src/app/application/components/data-source/DataSource.vue
johncowen: Couple of things here. `DataSource` is a representation of data and div is a representation of nothing/everything, so something that says 'data' makes more sense semantically.

Secondly I think I actually meant to use `datalist` here 😆 🤷 . Honestly don't really care right now, its in one place so its easy to change later if we want than spend time discussing semantics here.

Just want to point out there is no technical reason that I'm aware of as to why using `data` is detrimental (but I've given a reason why a div may _become_ one), tbh feels like opinion/bikeshedding more than anything 🤷 

All in all I changed this to a `div`

johncowen:
> Because we shouldn’t have two poll implementations in the code base.

Agree, we are both on the same page then (just kinda in reverse). If we are happy with this PR and its merged then we can eventually remove the old one.

Importantly, I think the reasons for our current polling are completely removed by the extra features we get in this PR. From what I understood we were polling 2 endpoints every 10 seconds mainly for local data syncing reasons i.e. so two pieces of the same data on the same page stayed in sync. This PR approaches local data syncing differently.

On the other hand if we want to poll in order to keep in sync with remote data, we can do that with DataSource instead.

src/app/application/services/data-source/CallableEventSource.ts
johncowen: Because I don't want anyone thinking they need to `await` it.

src/app/application/services/data-source/DataSource.ts
johncowen: Other code could just use an async generator? Apart from that I'm not sure what other use cases we are going to have for polling apart from HTTP requesting?

src/app/policies/sources.ts
johncowen: This is done/in-progress

src/app/policies/sources.ts
johncowen: See https://github.com/kumahq/kuma-gui/pull/1132 for the removal of the `null` on `items`

johncowen:
Taking this out of draft, feels like at this point it may aswell be a proper PR. Wanted to note that there are a couple of commented in things that are likely to be uncommented in follow up PRs.

src/app/application/services/data-source/CallableEventSource.ts
```diff
@@ -0,0 +1,54 @@
...
+  _open() {
+    (async function (self) {
+      try {
+        self.readyState = 0
+        const source = self.source()
+        for await (const res of source) {
+          self.dispatchEvent(new MessageEvent('message', {
+            data: res,
+          }))
+          if (self.readyState === 2) {
```
kleinfreund, line 31: **Suggestion**: Please use constants in the places you check against `readyState`. It’s too easy to forget what’s "0", "1", or "2".

src/app/application/services/data-source/DataSource.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  open(src: string, ref: symbol): EventSource | undefined {
+    const _source = this.pool.acquire(src, ref)
+    if (typeof _source !== 'undefined') {
+      // _source.open();
+      _source.addEventListener('message', (e: Event) => {
+        // always fill the cache on a successful response
+        this.cache.set(src, (e as MessageEvent).data)
+      })
```
kleinfreund, line 68: **Suggestion**: Functions defined directly at the callsite of a properly typed function should typically not have or need type annotations. In this case, typing the `event` of the listener as `Error` prevents TypeScript from inferring it as `MessageEvent`.

```suggestion
      _source.addEventListener('message', (e) => {
        // always fill the cache on a successful response
        this.cache.set(src, e.data)
      })
```

src/app/policies/sources.ts
kleinfreund: I mostly want to avoid the common "now we have two places for this" problem. We already have a central place for this that adequately covers our needs for storage of types like this.

I also really disagree that these types should live here at all. Our APIs may be called from arbitrary points of our application and so arbitrary places may want to reference the parameter types (etc.).

kleinfreund (commented):
So, I’ve reviewed the bigger part of this now, and in general, I think the concept of data sources as presented can allow us to get a hold of request duplication and data sharing. If you think the trade off between that end and the added complexity is worth it, then that’s fine by me. I will say that the added logic for this mechanism is not simple. I would like to see unit tests for specifically the DataSource component and its auxiliary parts.

One thing I want to suggest is to consider lazy loading separately (i.e. not as part of this PR). I’m not convinced we will need to make use of lazy loading techniques or at least we have no concrete plans to do so any time so. I’m not a fan of prematurely adding mechanisms that we’re not employing, yet.

src/app/application/services/data-source/DataSource.ts
johncowen: Ah I think this was from an earlier noodle, seeing as I typed everything to native EventSource pretty late on. I've also noticed that this is required for the firing and ErrorEvent as it seems like EventSource doesn't fire those. I guess probably because ErrorEvents didn't exist when EventSource started out maybe 🤔 

Anyway, whilst this seems like an easy fix, I want to think about it first. Considering this has very little impact I believe, and is required in the current form the case of ErrorEvent I'm not too concerned rn. I might change the MessageEvent one but I kinda like the consistency.

Considering this is a Suggestion I may or may not just resolve this, maybe adding a comment to the code or something.

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,78 @@
...
+})
+const refresh = () => {
+  close(controller, _src)
+  open(props.src)
+}
+</script>
+
+<template>
+  <slot
+    :data="message as any"
```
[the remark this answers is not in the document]
johncowen: Yeah seems like it is, if I just remove this I get complaints elsewhere. I think if we also change the definition of the ref for message to `any` it should be fine, but I'm unsure of making these changes on a whim when there are probably reasons for me doing it this way. I think a lot of this is centralized away from the application itself (a lot of the point of this PR) so if these things cause issue we can fix up behind the DataSource.

For now, unless anyone is super strong about me tweaking these things (considering perhaps unintended effects), I'd probably leave as is for the moment. Whilst I appreciate you checking these sorts of things I'm not too concerned about leaving as is for the moment, especially considering the type of PR this is. If there is anything you feel strongly we should change please feel free to nudge more.

src/app/application/services/data-source/CallableEventSource.ts
johncowen: > Just because something is a promise doesn’t mean it must be awaited for use

No but somebody might think that when their editor tells them it returns a promise. This is a potential negative consequence, and I can't think of any potential negative consequences of leaving as is. If you can think of one let me know and I'll gladly change it up, otherwise it feels like opinion.

If a negative consequence arises at a later date we can also change.

lahabana:
I do think that if Datasource is meant to be a standalone component there should be unit tests for it. integration tests are good but the feedback loop and ability to trigger complex code path is tough in these high level tests

src/app/application/services/data-source/CallableEventSource.ts
johncowen: Resolving this, I don't think its a major thing but please let me know if I should reopen

src/app/application/services/data-source/DataSource.ts
johncowen: At the moment we only use polling for HTML requests, and as there aren't very many other reasons for polling apart from HTTP requests and we are now going to be using DataSource for all our HTTP requests, as mentioned below

> I’m not a fan of prematurely adding mechanisms that we’re not employing

If for whatever reason we need polling for anything other than HTTP requests, and we can't just use a quick async generator, then we can expose something else. Until then I don't think this is a big deal.

Lastly, I don't want anyone trying to bypass DataSource to do HTTP polling, and giving them an exposed poll function would make it more likely that they try and do that.

src/app/policies/sources.ts
johncowen: The `sources` files will become the central place for anything to do with HTTP. This should not leak through in to the application code as it currently does, therefore we shouldn't need these types anywhere but here. Moreover here when I say 'central' I mean central to the module, not the application.

Take into account I pulled back a lot of this to keep using KumaApi _temporarily_. This PR is a path to removing all of that and ensuring there is a good boundary between the data layer and the view layer i.e.:

> Our APIs may be called from arbitrary points of our application and so arbitrary places may want to reference the parameter types

We should avoid using the HTTP parameter types in arbitrary places in the application.

src/app/policies/views/PolicyListView.vue
```diff
@@ -1,93 +1,170 @@
...
-                  appearance="select"
-                  :enable-filtering="true"
-                  @selected="changePolicyType"
-                >
-                  <template #item-template="{ item }">
-                    <span
-                      :class="{
-                        'policy-type-empty': policyTypeNamesWithNoPolicies.includes(item.label)
+        <template
+          v-for="selected in [policies.policies.find((item: any) => item.path === _route.params.policyPath) ?? policies.policies[0]]"
```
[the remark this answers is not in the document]
johncowen: This is done now (the `any`), thanks again!

src/app/application/components/data-source/DataSource.vue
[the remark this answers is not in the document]
johncowen: Resolving this one

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,78 @@
...
+      { signal: controller.signal },
+    )
+    source.addEventListener(
+      'error',
+      (e: Event) => {
+        error.value = (e as ErrorEvent).error as Error
+        emit('error', error.value)
+      },
+      { signal: controller.signal },
+    )
```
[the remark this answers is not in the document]
johncowen: This is done

src/app/application/services/data-source/DataSource.ts
johncowen: I think this is covered above, but whichever it done now as far as I understand.

johncowen:
> I do think that if Datasource is meant to be a standalone component there should be unit tests for it. integration tests are good but the feedback loop and ability to trigger complex code path is tough in these high level tests

I think we are maybe all on the same page but using different words. You want me to test `<DataSource />` on its own in isolation to make sure it all works by being able to trigger paths to test all its functionality, rather than in the application right?

johncowen:
I've added some initial tests here which I think is what people where most after. I would like to continue expanding on those, but probably after this PR is approved.

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+const close = (controller: AbortController | undefined, src: string) => {
+  if (typeof controller !== 'undefined') {
+    controller.abort()
+  }
+  if (typeof src !== 'undefined') {
+    data.close(src, sym)
+  }
+  controller = undefined
```
lahabana, line 62: either cleanup both `controller` and `_src` or neither. Doing only half is confusing

lahabana: You are shadowing global state with parameter names this makes it confusing and was the original reason for the comment above. Probably change the param name and remove the line

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,81 @@
...
+const message = ref<unknown>(undefined)
+const error = ref<Error | undefined>(undefined)
+
+const emit = defineEmits<{
+  (e: 'change', value: any): void
+  (e: 'error', error: Error): void
+}>()
+
+let controller: AbortController
+let _src: string
```
lahabana, line 25: how about having an object like:

```js
{
  controller: AbortController
  src: string
}
```

src/app/application/services/data-source/SharedPool.ts
```diff
@@ -0,0 +1,45 @@
+type Init<K, T> = (key: K) => T | undefined
+type Del<K, T> = (key: K, item: T) => void
+export default class SharedPool<K, T> {
+  pool: Map<K, T> = new Map()
+  usage: Map<T, symbol[]> = new Map()
```
lahabana, line 5: Should this be a set of symbol?

src/app/application/services/data-source/SharedPool.ts
```diff
@@ -0,0 +1,45 @@
...
+  init: Init<K, T>
+  del: (key: K, item: T) => void
+  constructor(init: Init<K, T>, del: Del<K, T>) {
+    this.init = init
+    this.del = del
+  }
+
+  // getter, not init
+  acquire(key: K, ref: symbol) {
```
lahabana, line 14: missing return type here

src/app/application/services/data-source/SharedPool.ts
lahabana: Try doing: `Map<K, CountingReference[T, symbol]>`

src/app/application/services/data-source/SharedPool.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!this.pool.has(key)) {
+      this.pool.set(key, this.init(key) as T)
+    }
+    const item = this.pool.get(key) as T
+    if (typeof this.usage.get(item) === 'undefined') {
+      this.usage.set(item, [])
```
lahabana, line 20: This array should be a set.

src/app/application/services/data-source/DataSource.ts
```diff
@@ -0,0 +1,83 @@
...
+  })
+  return _source as EventSource
+}
+const destroy = (_src: string, source: EventSource) => {
+  if (source) {
+    source.close()
+  }
+}
+
+export class DataSource {
```
lahabana, line 37: DataSourcePool?

src/app/application/components/data-source/DataSource.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      props: {
+        src: '',
+      },
+    })
+
+    await flushPromises()
+
+    expect(wrapper.emitted('change')).toBeFalsy()
+  })
+  test('cached repsonses and errors work', async () => {
```
lahabana, line 19:
```suggestion
  test('cached responses and errors work', async () => {
```

src/app/application/components/data-source/DataSource.vue
johncowen: I change this to use a "component global" state object with two props, rather than two separate "component global" vars. I then pass that around instead of the two vars.

src/app/application/components/data-source/DataSource.vue
johncowen: This is done

src/app/application/services/data-source/SharedPool.ts
johncowen: I kept the two separate `pool`, `usage` as I 'think' we still need that, but I changed things to use a Set rather than just an array 👍 , I'm just checking with you offline to see if we definitely need the two separate `pool`, `usage`

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,84 @@
...
+    state.controller.abort()
+  }
+  if (typeof state.src !== 'undefined') {
+    data.close(state.src, sym)
+  }
+  return {}
+}
+watch(() => props.src, (src) => open(src), { immediate: true })
+onBeforeUnmount(() => {
+  close(state)
```
lahabana, line 69:
```suggestion
  state = close(state)
```
no?

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,84 @@
...
+})
+const refresh = () => {
+  close(state)
+  open(props.src)
```
lahabana, line 73: open calls `close` so why do close here?

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,84 @@
...
+      { signal: state.controller.signal },
+    )
+  }
+}
+const close = (state: State) => {
+  if (typeof state.controller !== 'undefined') {
+    state.controller.abort()
+  }
+  if (typeof state.src !== 'undefined') {
+    data.close(state.src, sym)
```
lahabana, line 63: Is this actually just `release` and not `close` ?

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,84 @@
...
+const sym = Symbol('')
+const open = async (src: string) => {
+  state = close(state)
+  state.src = src
+  if (src === '') {
+    return
+  }
+  state.controller = new AbortController()
+  // this should emit proper events
+  const source = data.source(src, sym)
```
lahabana, line 38: Maybe this is `acquire`? Or `getOrCreate`

src/app/application/components/data-source/DataSource.vue
```diff
@@ -0,0 +1,84 @@
...
+  if (typeof source !== 'undefined') {
```
lahabana, line 39: should the check be on src so it's symetric with your close function?

lahabana: What are the cases where source is 'undefined' ?

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,83 @@
+import CallableEventSource, { isClosed } from './CallableEventSource'
+import Router from './Router'
+import SharedPool from './SharedPool'
+
+export type DataSourceResponse<T> = {data: T | undefined, error: Error | undefined, refresh: () => void}
+export type Source = (params: Record<string, unknown>, source: {close: () => void}) => Promise<unknown>
+export type Sources = Record<string, Source>
+
+const get = (src: string, router: Router<Source>) => {
```
lahabana, line 9: I get `get` because is an HTTP `GET` but maybe this would be better as a `create` which would be symetric with `destroy`

lahabana: Missing return type btw and if you look at it `undefined` is impossible which makes a lot of `undefined` handling in SharedPool and DataSourcePool unnecessary

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,83 @@
...
+    return this.open(src, ref)
+  }
+
+  cached(src: string) {
+    return this.cache.get(src)
+  }
+
+  open(src: string, ref: symbol): EventSource | undefined {
+    const _source = this.pool.acquire(src, ref)
+    if (typeof _source !== 'undefined') {
```
lahabana, line 63: Here's an example of undefined handling that's impossible I believe

src/app/application/components/data-source/DataSource.vue
johncowen: I wanted it to read like `data source` ie. "give me a data source please, I don't care how just go get it for me".

The fact that there is a pool underneath is purposefully hidden from the interface of the datasource itself, hence why I much prefer `data.source` rather than things like `acquire`. Then `source.open` and `source.close` which sticks closer to the semantics of `EventSource` which is what this is modelled on.

johncowen:
I think I got all those last comments, I much prefer the `create/destroy` consistency for the shared pool, anyway lemme know. Its a lot tighter now I think.

src/app/application/services/data-source/DataSourcePool.ts
johncowen: I removed a lot of the `| undefined` elsewhere, this one doesn't seem to require a return type, but I added for readability anyway.

I also moved everything to use `CallableEventSource` rather than just `EventSource` (still later on I would like it to all use `EventSource` as a type)

src/app/application/services/data-source/SharedPool.ts
johncowen: I think we settled on this one as it is now, lemme know if not

src/app/application/services/data-source/CallableEventSource.ts
```diff
@@ -0,0 +1,54 @@
+export const isClosed = (source: { readyState: number }) => source.readyState === 2
+export default class CallableEventSource extends EventTarget {
```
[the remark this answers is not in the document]
lahabana: I think this should just be a comment in the code

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,80 @@
...
+  const queryParams = new URLSearchParams(query)
+  const route = router.match(path)
+  const _source = new CallableEventSource(async function * (this: CallableEventSource) {
+    while (true) {
+      this.readyState = 1
+      // latency
+      // await new Promise(resolve => setTimeout(resolve, 5000))
+      yield route.route({
+        ...{
+          offset: parseInt(queryParams.get('offset') || '0'),
```
lahabana, line 20: Why is offset receiving special treatment here?

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+    }
+  })
+  return _source
+}
+const destroy = (_src: string, source: CallableEventSource) => {
+  if (source) {
+    source.close()
+  }
+}
```
lahabana, line 35: _src is not used. Why do we have this method should we just do the suggestion below.

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,80 @@
...
+  pool: SharedPool<string, CallableEventSource>
+  constructor(requests: Sources) {
+    const requestRouter: Router<Source> = new Router(requests)
+
+    this.pool = new SharedPool<string, CallableEventSource>(
+      (src: string) => {
+        return create(src, requestRouter)
+      },
+      (src: string, source: CallableEventSource) => {
+        return destroy(src, source)
```
lahabana, line 48:
```suggestion
        return source.close()
```

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,80 @@
...
+    )
+  }
+
+  source(src: string, ref: symbol) {
+    return this.open(src, ref)
+  }
+
+  cached(src: string) {
+    return this.cache.get(src)
+  }
```
lahabana, line 59: I don't think this method is used anywhere we should just remove it

src/app/application/services/data-source/Router.ts
lahabana: Maybe a unit test of this would be useful. This code is a little hairy and ensuring it works as expected will make things easier to debug

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,80 @@
...
+      },
+      (src: string, source: CallableEventSource) => {
+        return destroy(src, source)
+      },
+    )
+  }
+
+  source(src: string, ref: symbol) {
+    return this.open(src, ref)
+  }
```
lahabana, line 55: Should this method go away if it's just a passthrough to `open` (or inline open in here)?

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,80 @@
...
+  return _source
+}
+const destroy = (_src: string, source: CallableEventSource) => {
+  if (source) {
+    source.close()
+  }
+}
+
+export class DataSourcePool {
+  cache: Map<string, unknown> = new Map()
```
lahabana, line 38: Worth adding a comment saying that this cache is unbounded but it's just a buffer to limit the use of loaders.

src/app/policies/views/PolicyListView.vue
```diff
@@ -1,93 +1,170 @@
...
-                      :class="{
-                        'policy-type-empty': policyTypeNamesWithNoPolicies.includes(item.label)
+        <template
+          v-for="selected in [policies.policies.find(item => item.path === _route.params.policyPath) ?? policies.policies[0]]"
+          :key="selected.path"
+        >
+          <DataSource
+            v-slot="{data, error, refresh}: PolicyCollectionSource"
+            :src="`/${_route.params.mesh}/policy-type/${selected.path}?offset=${route.query.offset || '0'}`"
+            @change="change"
```
lahabana, line 26: I don't understand why we react to these change? Is it to navigate to the first policy type as soon as it's loaded?

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+  return _source
+}
+const destroy = (_src: string, source: CallableEventSource) => {
+  if (source) {
+    source.close()
+  }
+}
+
+export class DataSourcePool {
```
lahabana, line 37: This class now does so little that it could almost be collapsed with SharedPool. But that's really up to you :)

lahabana (commented):
We're getting there. 1 question which is more showing my lack of JS understanding. We never remove the eventListener is this ok? Are they removed when the object is GCd?

johncowen:
Hey @lahabana 

I'm going to start going over these inlines now but to come back to you on the main comment:

> We're getting there. 1 question which is more showing my lack of JS understanding. We never remove the eventListener is this ok? Are they removed when the object is GCd?

These should be getting cleared up via the AbortController which gets aborted whenever we close the DataSource (i.e. whenever the component is unmounted) this is in `DataSource.vue`. Do take another look for me though and see if that all makes sense to you.

Looking at the other comments now.

src/app/application/services/data-source/DataSourcePool.ts
johncowen: If you mean the `parseInt`, it's just to turn it into a number for the argument for the function call/HTTP API request in the `sources` file. I was thinking by the time you get to the function call offset should be a number.

Slightly more longer term I think I would rather specify a `page` and then calculate the correct offset behind the scenes (this is how I do it in the original code for this). But I didn't want to bring to much in here rn.

Actually you've just reminded me, I think we said we would rename `route.route` to something else so its clearer that we are calling the `'/uri/:arg' => Promise` here.

--

(Let me know if I misunderstood the question here)

src/app/application/services/data-source/DataSourcePool.ts
johncowen: Yeah sure, I don't think it here will remove

src/app/application/services/data-source/DataSourcePool.ts
johncowen: I like this as then you can get an instance of a DataSource as `data` and then call `data.source(...)` on it, so it reads nice, so it's meant to only ever be an alias of `open`. I can remove if you don't like that, but thats exactly what its supposed to be a passthrough and nothing more. lmk

src/app/application/services/data-source/DataSourcePool.ts
johncowen: Good call, we should talk about this fact deeper at a later date as I definitely would like to add something to address this. I'll add something in comments for now.

src/app/application/services/data-source/DataSourcePool.ts
johncowen: Let me take another look today. Initially I am still "I like that fact that the pool of instances and the cache of responses are separate things" but I get why you like collapsing it. Let me look again today.

src/app/application/services/data-source/DataSourcePool.ts
johncowen: I suppose it was just because "as `create` is separate, lets make `destroy` separate also".

---

Ah, I think also I would really like to consider getting shared pool instantiation out of the constructor, in fact its more I would really like to get both `create` and `destroy` functions out of this file entirely, so you can define them externally to DataSource and pass them in if you want to. Right now I'm restricted by the `router` also being instantiated in the constructor and I don't want to have to pass the `router`, you can see I need that in the `create` function.

🤔 actually what I want to do here is a bit difficult to explain here, might be better to explain next time we pair on this.

For the moment I'm going to do the suggestion you added here, but would be good to get your input on what I would like to do here, right now what I would like to do isn't urgent at all though.

src/app/policies/views/PolicyListView.vue
johncowen: I think this one is good to resolve

src/app/application/services/data-source/DataSourcePool.ts
lahabana: Why not remove open then? It's exactly the same function

lahabana (approved):
LGTM I only have something about this useless "passthrough" method. But up to you

src/app/application/services/data-source/DataSourcePool.ts
johncowen: Lemme remove `open`, 2 tics

johncowen:
K I'm merging this, thanks all for the walking through it with me

PR #1137, opened 2023-06-28
fix: Avoids throwing an error with subscription-less zone-cps

johncowen:
I noticed in some logs that we were throwing an error here depending on whether subscriptions are defined or not.

I fixed the bug and added a more user facing graceful warning in the application code, but also made a few updates to some tests related files while I was there.


features/zones/DetailViewContent.feature
```diff
@@ -4,38 +4,6 @@ Feature: Zones: Detail view content
...
-    Then the "$details" element contains "ZoneOverview"
-    Then the "$details" element contains "online"
-    Then the "$details" element contains "dpToken"
-
-    When I click the "#insights-tab" element
-    Then the "$details" element contains "Connect time: July 28, 2020 at 4:18:09 PM"
-
-    When I click the "#config-tab" element
-    Then the "$details" element contains "adminAccessLogPath"
-
```
johncowen, line 38: I moved this into `features/zones/zone-cps/Item.feature` while I was here (we should also move the others to `features/zones/zone-ingresses` and `features/zones/zone-egresses` at some point)

src/app/zones/components/ZoneDetails.vue
```diff
@@ -168,10 +177,12 @@ const warnings = computed<ZoneCompatibility[]>(() => {
 
 const codeOutput = computed(() => {
   const subscriptions = props.zoneOverview.zoneInsight?.subscriptions ?? []
-  const lastSubscription = subscriptions[subscriptions.length - 1]
+  if (subscriptions.length > 0) {
```
johncowen, line 180: This conditional/guard is the bug fix

src/app/zones/locales/en-us/index.yaml
```diff
@@ -3,6 +3,8 @@ zone-cps:
     item:
       title: "{name} Zone CP"
       breadcrumbs: Zone CPs
+      config:
+        no-subscriptions: This zone has no subscriptions
```
johncowen, line 7: If someone knows some better text we can amend that here or later.

johncowen (commented):
Adding inline notes

lahabana (approved):

PR #1143, opened 2023-06-30
refactor: Cleans up RouteView code and adds basic docs

johncowen:
The `RouteView`/`RouteTitle` and `AppView` components were initially made in a closed manner to PoC the different concept and then roll it out across the applications.

I'd always planned on coming back to tidy things up inside the components and add some documentation to help any external contributors or anyone unfamiliar with our codebase. This PR does that.

There was the initial concept of automatic navigation building/navigation selecting in `RouteView`, which we were using but pulled out and therefore we aren't using just yet. I decided to remove this entirely in this the clean up and add afresh at a later date. In the PR you will see things related to `children` and `module=""` being removed, this deletion is related to this.

### DIFF
--- a/src/app/application/utilities/index.ts
+++ b/src/app/application/utilities/index.ts
@@ -15,7 +15,7 @@ export const createTitleSetter = ($doc = document) => {
     $doc.title = title
   })
 }
-export const createAttrsSetter = ($el = document.querySelector('html')) => {
+export const createAttrsSetter = ($el = document.documentElement) => {
   if (!$el) {
     return () => {}
   }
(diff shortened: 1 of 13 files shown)


src/app/application/utilities/index.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+export const urlParam = function <T extends string | null> (param: T | T[]): string {
+  return (Array.isArray(param) ? param[0] : param) ?? ''
+}
+export const createTitleSetter = ($doc = document) => {
+  return beforePaint((title) => {
+    $doc.title = title
+  })
+}
+export const createAttrsSetter = ($el = document.querySelector('html')) => {
```
kleinfreund, line 18: **Suggestion**: No need to query this one

```suggestion
export const createAttrsSetter = ($el = document.documentElement) => {
```

kleinfreund (approved):

PR #1156, opened 2023-07-04
feat(docs): use i18n for docs links

lahabana:
Always use this utility function to create doc links.

Also use the name for policy docs so that it's more systematic (follow up website docs to make it all work well)


lahabana:
Ok there's a level of typescript I'm not getting. I've already tried to move `docLink` to be a method on env but didn't get any success. So I guess @johncowen I'll need your help tomorrow

johncowen (approved):
LGTM, glad we are using `t` over `env` more and more for our strings 🎉 , ty!

PR #1157, opened 2023-07-04
chore: Removes last HTTP call to `/` and moves bootstrapping

johncowen:
This PR began as removing the last HTTP API call to `/` that we have. Whilst doing this I noticing that we just invisibly failed on any errors in our 'bootstrapping' HTTP calls. I decided it was finally time to move our bootstrapping functionality into `App.vue` where we could catch any errors and show a proper error state.

---

### Removing the last HTTP API call to `/`

Previous to this PR we would make a call to `/` to see if that call was successful, if it wasn't successful we would just do nothing and miss out a bunch of HTTP calls we depend on further into the application, not great.

If we were successful, we wouldn't use any of the data from the API call, but instead make another bunch of API calls. Then if any of those failed, do nothing, and therefore be missing a bunch of data we then depend on further into the application, not great.

I don't quite understand why we would make a call to an endpoint just to see if its successful even though we don't use any of its data, when it would make more sense to use an endpoint that we actually use data from.

All the loading and error problems here led me to also...

### Move our 'bootstrapping' HTTP calls to App.vue

As we've been wanting to do this for quite a while, I figured if I did this now I could also use all our existing loading and error states to fail gracefully. 

---

### Lastly:

There is still so much stuff we can remove here now, but as its really intertwined throughout the app it takes quite a bit of picking apart. i.e. we still get a list of meshes at the start of the application, when we shouldn't need to; we still get a load of policy information at the start of the application when we shouldn't need to. There is ongoing work to change things so that we only load information for the page we are on directly from the backend every single time we load a page (as we want our data to be up-to-date and not stale), so hopefully we can pick this stuff out once thats there.

Our `bootstrap` function now does nothing in production, it is decorated with MSW functionality in development. Whilst this looks a little weird having an empty function, we can at some point noodle around with this as well, just I didn't want to do too much more here.

### DIFF
--- a/features/application/Loading.feature
+++ b/features/application/Loading.feature
@@ -19,21 +19,16 @@ Feature: index
     And I wait for 2000 milliseconds
     Then the "$loading" element doesn't exist
     And the "$logo" element exists
-
-  # TODO: This test needs fixing it currently console.errors
-  @skip
   Scenario: Application errors
     Given the environment
       """
       KUMA_LATENCY: 1000
       """
-    And the URL "/" responds with
+    And the URL "/config" responds with
       """
       headers:
         Status-Code: 500
       """
     When I visit the "/" URL
-    Then the "$loading" element exists
-    And I wait for 2000 milliseconds
     Then the "$loading" element doesn't exist
     Then the "$error" element exists
--- a/src/app/App.vue
+++ b/src/app/App.vue
@@ -1,6 +1,8 @@
 <template>
   <RouteView>
-    <AppLoadingBar v-if="store.state.globalLoading || route.name === undefined" />
+    <AppLoadingBar
+      v-if="loading || route.name === undefined"
+    />
 
     <template v-else>
       <AppHeader v-if="!isWizard" />
@@ -17,31 +19,34 @@
 
         <AppView>
           <AppErrorMessage
-            v-if="shouldShowAppError"
+            v-if="error"
             data-testid="app-error"
           />
-
-          <AppOnboardingNotification v-if="!isWizard && shouldShowOnboardingNotification" />
-
-          <router-view
-            :key="routeKey"
-            v-slot="{ Component }"
+          <template
+            v-else
           >
-            <transition
-              mode="out-in"
-              name="fade"
+            <AppOnboardingNotification v-if="!isWizard && shouldShowOnboardingNotification" />
+
+            <router-view
+              :key="routeKey"
+              v-slot="{ Component }"
             >
-              <div
-                :key="(route.name as string)"
-                class="transition-root"
+              <transition
+                mode="out-in"
+                name="fade"
               >
-                <component
-                  :is="Component"
-                  :data="props.data"
-                />
-              </div>
-            </transition>
-          </router-view>
+                <div
+                  :key="(route.name as string)"
+                  class="transition-root"
+                >
+                  <component
+                    :is="Component"
+                    :data="props.data"
+                  />
+                </div>
+              </transition>
+            </router-view>
+          </template>
         </AppView>
       </div>
     </template>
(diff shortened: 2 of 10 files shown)


src/app/App.vue
```diff
@@ -99,8 +104,28 @@ const route = useRoute()
  */
 const routeKey = computed(() => route.path)
 const isWizard = computed(() => route.meta.isWizard === true)
-const shouldShowAppError = computed(() => store.getters.shouldShowAppError)
```
johncowen, line 102: Again a fairly local piece of state, so I switched this out for a local piece of state.

src/app/App.vue
```diff
@@ -99,8 +104,28 @@ const route = useRoute()
...
+        // information used for and policy lookups in the app.
+        store.dispatch('fetchPolicyTypes'),
+      ])
+    } catch (e) {
+      error.value = e as Error
+    } finally {
+      loading.value = false
+    }
+  }
+)()
```
johncowen, line 128: This is the stuff I moved into `App.vue`, now with loading and error integration.

src/app/AppHeader.vue
```diff
@@ -119,7 +118,6 @@ const [
 const store = useStore()
 const env = useEnv()
 
-const shouldShowAppError = computed(() => store.getters.shouldShowAppError)
```
johncowen, line 122: As our version is now burnt into the kuma binary/GUI, whether we get an error from the API has no bearing on whether we show you your current version or not.

In fact I just realized a UX advantage that if there is an error, you probably want to know what version of kuma you are running.

src/store/storeConfig.ts
```diff
@@ -122,42 +117,36 @@ export const storeConfig = (kumaApi: KumaApi): StoreOptions<State> => {
...
-            await dispatch('updateSelectedMesh', null)
+          if (mesh === undefined) {
+            // If the stored mesh doesn’t exist, use the first mesh instead.
+            mesh = state.meshes.items[0]
           }
+
+          await dispatch('updateSelectedMesh', mesh.name)
+          await dispatch('sidebar/getInsights')
+        } else {
+          await dispatch('updateSelectedMesh', null)
```
johncowen, line 149: All this green in this file is just removing the wrapping conditional checking to make sure that `status === 'OK'`

features/application/Loading.feature
```diff
@@ -19,21 +19,16 @@ Feature: index
     And I wait for 2000 milliseconds
     Then the "$loading" element doesn't exist
     And the "$logo" element exists
-
-  # TODO: This test needs fixing it currently console.errors
-  @skip
```
johncowen, line 24: Another benefit here is that this test now passes, and whilst it did also pass in its previous incarnation in the CLI tests, when I tried to recreate it in a real browser it failed.

johncowen (commented):
Inline notes

src/index.ts
```diff
@@ -25,20 +25,7 @@ export function useApp(
...
-        // onboarding flow in the appropriate scenarios.
-        store.dispatch('bootstrap'),
-        // Loads available policy types in order to populate the necessary
-        // information used for and policy lookups in the app.
-        store.dispatch('fetchPolicyTypes'),
-      ])
-    } else {
-      store.state.defaultVisibility.appError = false
-    }
+export function useBootstrap() {
```
lahabana, line 28: Why is this still around btw?

lahabana (approved):

src/index.ts
johncowen: Yeah not super happy about it either, but didn't want to rearrange too much here.

> Why is this still around btw?

In production its a noop, but in development this empty function gets decorated with MSW functionality. Longer term we will probably move MSW elsewhere and then we can get rid of this entirely.

lahabana: Could you make this explanation a comment in the code so that someone doesn't remove what seems to be deadcode?

lahabana (approved):

src/index.ts
johncowen: Oh yeah sure makes sense, sorry just saw this! 2 tics

PR #1158, opened 2023-07-05
fix: Revert "chore(deps): Bump @kong/kongponents from 8.83.3 to 8.84.0 (#154)

johncowen:
See https://github.com/Kong/kongponents/pull/1500

TLDR; 8.84.0 is missing a dependency when installed as an indirect dependency.

Reverting for this moment

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@kong-ui-public/i18n": "^0.6.1",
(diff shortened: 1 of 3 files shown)


johncowen:
This was eventually fixed in https://github.com/Kong/kongponents/pull/1501 and we have a PR up for the upgrade here https://github.com/kumahq/kuma-gui/pull/1159 . Closing this revert.

PR #1162, opened 2023-07-10
fix: DataOverview error states

johncowen:
The Error/Loading states in our very very commonly used DataOverview component don't seem to work as intended.

Currently, if it is in a loading state and you pass the component an error, the component remains in its loading state. This sounds wrong to me. If we have an error then surely we are not loading any more. Consequently I have seen this behaviour resulting a surprising issue in code I have been working on (DataSource work).

I am unsure if the change here has an effect elsewhere, or whether we are relying on this what I see as strange behaviour (if we have an error we shouldn't be loading anymore). But I am little cautious about this as I still don't know whether we are relying on this elsewhere.

If component should be in a loading state even if you pass it an error, then please let me know and I'd be happy to close the PR.

### DIFF
--- a/src/app/common/DataOverview.vue
+++ b/src/app/common/DataOverview.vue
@@ -18,12 +18,11 @@
       </KButton>
     </div>
 
-    <LoadingBlock v-if="isLoading" />
-
     <ErrorBlock
-      v-else-if="error !== null"
+      v-if="error"
       :error="error"
     />
+    <LoadingBlock v-else-if="isLoading" />
 
     <EmptyBlock v-else-if="isEmpty" />
 


kleinfreund:
> Currently, if it is in a loading state and you pass the component an error, the component remains in its loading state. This sounds wrong to me.

This may seem weird at first, but this is the correct order in my opinion: our list views are generally set-up to support loading data at arbitrary points (e.g. initial load, refresh, pagination interactions, item deletion). Also, it’s the responsibility of the loading routine to set the error states appropriately. Therefore, if the list view is in a loading state, this should take priority over an already discovered error state because at that point it is unknown if the loading routine will succeed (i.e. no more error) or fail (i.e. old or new error).

johncowen:
Hey @kleinfreund thanks for taking a look at this I was thinking of closing up this one anyway as I've had a change of heart on it.

> Therefore, if the list view is in a loading state, this should take priority over an already discovered error state because at that point it is unknown if the loading routine will succeed

FWIW, and I might have misunderstood what you are saying but, if we have to perform a refresh for some reason, then the state should be reset back to it's initial state, and if it was originally in an error state then that would therefore be reset also, meaning if the refresh then errored again, error would take precedence over loading.

I still do think overall its weird how all this works, but happy to leave this one for the minute, hence I'm closing this one for now.

PR #1175, opened 2023-07-13
chore: removes charts being clickable

kleinfreund:
Removes the feature making charts link to related pages which people were generally surprised existed in the first place and doesn't seem to serve much purpose.

Note: This removes one of the last uses of `store.state.selectedMesh`. The last one will be removed in a follow up in which I’m dropping the sidebar store module (its main purpose no longer exists and we only use a bit of its state).


lahabana (approved):
great idea!

johncowen:
Was it intentional to remove the clicks from the charts here also? They shouldn't require the `selectedMesh` right?

Screenshot 2023-07-13 at 17 47 16

johncowen (approved):

PR #1177, opened 2023-07-14
ci: adds cypress-fail-fast plugin

kleinfreund:
Adds the cypress-fail-fast plugin which will cause the entire Cypress test run to fail fast after 1 failed retry.

image


kleinfreund:
**Issue**: Huh, strange. It just doesn’t work?

PR #1178, opened 2023-07-14
ci: uploads Cypress screenshots and videos on failure

kleinfreund:
Uploads Cypress screenshots and videos on failure of the test run.

Workflow run summary in failure case (has uploaded artifacts): https://github.com/kumahq/kuma-gui/actions/runs/5552202267

Workflow run summary in success case (doesn't have uploaded artifacts): https://github.com/kumahq/kuma-gui/actions/runs/5552247266?pr=1178

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -87,7 +87,15 @@ jobs:
           yarn build:staging
       - run: |
           yarn preview:built &
-          yarn run test:browser
+          CYPRESS_video=true yarn run test:browser
+      - uses: actions/upload-artifact@v3
+        if: failure()
+        with:
+          name: cypress-artifacts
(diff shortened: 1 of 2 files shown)


johncowen (commented):
This LGTM!

Just a tiny thing if you could take a quick look, is there a way to open the web inspector so we can see that also. I guess not? But might be worth taking a quick look incase? lmk

(oh and thanks for posting a link to the fails 👍 )

kleinfreund:
@johncowen From looking at https://github.com/cypress-io/cypress/issues/2024, this seems technically possible but has issues with actually capturing the content on the dev tools.

johncowen (approved):

johncowen:
> This seems technically possible but has issues with actually capturing the content on the dev tools.

Would be good to sort this at some point, or keep an eye on the issue if its an upstream problem

PR #1180, opened 2023-07-14
chore: Add title and actions slots to AppView

johncowen:
Add title and actions slots to `AppView`

The vast majority of `AppView`s are going to be displayed with a title and probably a set of actions in either a dropdown or just a single button. This PR therefore adds title and actions slots to `AppView` so we can add additional markup/styling for that in one place.

I've used `section` for section content, which along with the `header`s and `h*`, will should give us more correct structuring/document outline.

`h*` tags are defined per route component so we can choose which level we need depending on the level of route/header nesting.

I've also added `render=true` to `RouteTitle` so you can use it within a `h*` to provide the header and the HTML title using a single element. Later we will probably default to true and turn it off for the rarer time when you don't want it to render. Right now I would rather gradually roll this out across the application and then switch in one go.

Lastly, I removed a `kcard-stack` CSS rule as we don't seem to be using it (see convo offline).

**Note: The change in the application code here is for PR example only, and it will be removed post approval before merging, and initial usage or this functionality will be done elsewhere.** If this approach seems fine I will also work on integration elsewhere and update the `AppView` docs.

---

Ok this grew a tiny bit bigger than I'd originally planned as I noticed we could reuse `AppView#title` and `AppView#actions` instead of the extra `WizardTitleBar` component. Less code, more reuse, less to learn.


src/app/services/views/ServiceListView.vue
```diff
@@ -34,6 +46,7 @@
 </template>
 
 <script lang="ts" setup>
+import { KButton } from '@kong/kongponents'
```
johncowen, line 49: All this change will be undone before merge.

src/app/meshes/views/MeshView.vue
```diff
@@ -1,6 +1,11 @@
 <template>
-  <RouteView>
+  <RouteView
+    v-slot="{route}"
+  >
     <AppView>
+      <template #title>
+        <h1>{{ route.params.mesh }}</h1>
+      </template>
```
johncowen, line 8: This change will also be removed before merge.

johncowen (commented):
Inlines:

kleinfreund:
> Lastly, I removed a `kcard-stack` CSS rule as we don't seem to be using it (see convo offline).

I specifically mentioned I intend to use it. Please don’t remove it.

johncowen:
I added it back in along with a comment to say its currently unused here and should be renamed to a less misleading name when we come to use it.

src/assets/styles/_base.scss
```diff
@@ -34,10 +34,14 @@ h2,
...
 h5,
-h6 {
+h6,
+section h1 {
   line-height: 1.3;
   font-weight: var(--font-weight-semi-bold);
 }
+section h1 {
+  font-size: var(--type-xxxl, 32px);
+}
```
kleinfreund, line 44: **Issue**: This and the above shouldn’t be a base style. It’s specific to the component that uses these section elements and should be defined there.

src/app/application/components/app-view/AppView.vue
```diff
@@ -82,3 +140,45 @@ onBeforeUnmount(() => {
...
+  line-height: 36px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+header > :is(h1) {
+  font-size: var(--type-xxxl, 32px);
+}
+header > :is(h2) {
+  font-size: var(--type-xl, 22px);
+}
```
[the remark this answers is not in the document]
kleinfreund: Same for instance below.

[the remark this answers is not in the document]
johncowen: Unless I'm mistaken doing this would mean that the styles are not applied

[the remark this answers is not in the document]
johncowen: Because they are scoped styles and we need to cross the slot.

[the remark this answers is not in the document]
kleinfreund: `:is()` is counted by the highest specificity in its selector list which is a tag selector in this case and so contributes a specificity of 0 0 1 to the 0 0 1 of `header` which makes it 0 0 2, the exact same as `header h1`.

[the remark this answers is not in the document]
johncowen: Why not try by removing the `is` and adding a `color: red` to the rule and see if it gets applied.

src/assets/styles/_base.scss
johncowen: If we overwriting `h1` (as we are), then we should also overwrite `section h1` here:

Screenshot 2023-07-17 at 17 15 51

They are both styles that are applied by the browser

src/app/application/components/app-view/AppView.vue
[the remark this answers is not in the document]
johncowen: I've removed the `>`s here, but not the `is()`, lmk

src/assets/styles/_base.scss
kleinfreund: What user agent styles exist does not determine for what we need to add base styles. `section` is functionally equivalent with a `div` besides what’s specified in the HTML standards, I don’t see any reason at all that we ever want a base style for `section h1`.

Besides the use introduced here, we have none that need it. And the use here should be handled by component styles, not base styles.

src/app/application/components/app-view/AppView.vue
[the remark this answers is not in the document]
johncowen: I changed this to use a `.app-view-title-bar h*` in an unscoped block

src/assets/styles/_base.scss
johncowen: I've moved these into the component

kleinfreund (approved):

src/app/meshes/views/MeshView.vue
johncowen: I've decided to leave this here as I am currently working on these views and I may as well leave it in here instead of remove here and re-add back in again.

src/app/services/views/ServiceListView.vue
johncowen: I've decided to leave this here as I am currently working on these views and I may as well leave it in here instead of remove here and re-add back in again.

PR #1192, opened 2023-07-17
refactor: removes sidebar module

kleinfreund:
Removes the entirety of the sidebar store module as its initial purpose no longer exists (storing the data needed to display resource counts in the sidebar). There is only one bit of data from this module that’s still needed: the total policy type counts used in the policy pages. This data is now stored in the general store as `policyTypeTotals` and populated in MeshView.vue.

Removes a couple of unused bits from storeConfig.ts: `menu` (state), `itemQueryNamespace` (state), `currentMeshPolicies` (state), `version` (state), `globalLoading` (getter), `getMeshList` (getter), `getItemQueryNamespace` (getter). This also removes the last uses of `selectedMesh` (state).


src/app/meshes/views/MeshOverviewView.vue
```diff
@@ -169,13 +167,33 @@ const extendedMesh = computed(() => {
...
+        return {
+          name: policyType.name,
+          path: policyType.path,
+          total: stat.total,
+        }
+      }
+
+      return null
+    })
+    .filter(notEmpty)
```
johncowen, line 196: What does `notEmpty` offer over using `Boolean`?

johncowen (commented):
Glad we are getting rid of lots of Vuex code.

I left one question, but otherwise should be good, lmk on the q.

src/app/meshes/views/MeshOverviewView.vue
kleinfreund: Type safety. With `.filter(Boolean)`, the resulting list still has values that can be `null`. `notEmpty` returning `value is TValue` prevents this.

PR #1196, opened 2023-07-17
chore(e2e): removes unnecessary cy.wait calls

kleinfreund:
Removes a couple of calls to `cy.wait` from our tests.


johncowen (approved):
Just to check, you mentioned a "couple", but there was only one removed just infront of checking the title, which I guess is now unnecessary?

If there wasn't another then this LGTM!

kleinfreund:
> Just to check, you mentioned a "couple", but there was only one removed just infront of checking the title, which I guess is now unnecessary?

The other removals are in the form of commands in feature files which are also `cy.wait` calls.

johncowen:
Ah gotcha 🤔

PR #1197, opened 2023-07-17
fix: AppCollection -  defaults total to zero

johncowen:
Defaults page total to zero in AppCollection so we don't have to deal with that everywhere.

### DIFF
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -2,7 +2,7 @@
   <KTable
     class="app-collection"
     :has-error="(typeof props.error !== 'undefined')"
-    :pagination-total-items="props.total"
+    :pagination-total-items="props.total || 0"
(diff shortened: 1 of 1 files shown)


src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -51,7 +51,7 @@ type ChangeValue = {
 }
 
 const props = defineProps<{
-  total: number,
+  total: number | undefined,
```
kleinfreund, line 54: **Comment**: It’s better to define props using an object literal because then you can actually define this default and don’t have to handle it in code again.

The type parameter style of defining props doesn’t allow for default values to be defined (unless you use that cumbersome `withDefaults` helper https://vuejs.org/api/sfc-script-setup.html#defineprops-defineemits)

johncowen: this is done

kleinfreund (approved):

PR #1203, opened 2023-07-18
feat: expand app collection

kleinfreund:
## Changes

**feat: adds custom toolbar styles**

Adds custom toolbar styles so that toolbar entries are automatically placed and spaced in the right position above the table.

**refactor: moves cell attributes function into script block**

**chore: aid syntax highlighting by wrapping `as string` in template**

**chore: marks `total` as optional prop**

### DIFF
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -13,13 +13,7 @@
       })
       return {data: items}
     }"
-    :cell-attrs="({ headerKey }: any) => {
-      return {
-        class: {
-          [`${headerKey}-column`]: true,
-        },
-      }
-    }"
+    :cell-attrs="getCellAttributes"
     empty-state-icon-size="96"
     disable-sorting
     hide-pagination-when-optional
@@ -82,6 +76,14 @@ const click = (e: MouseEvent) => {
     if ($a !== null) {
       $a.click()
     }
+  }
+}
+
+function getCellAttributes({ headerKey }: any) {
+  return {
+    class: {
+      [`${headerKey}-column`]: true,
+    },
   }
 }
 </script>
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -102,7 +102,7 @@ const click = (e: MouseEvent) => {
 }
 </script>
 <style type="scss" scoped>
-.app-collection :is(td) > :is(a) {
+.app-collection :deep(td:first-of-type > a) {
   color: inherit;
   font-weight: var(--font-weight-semi-bold);
   text-decoration: none;
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -13,13 +13,9 @@
       })
       return {data: items}
     }"
-    :cell-attrs="({ headerKey }: any) => {
-      return {
-        class: {
-          [`${headerKey}-column`]: true,
-        },
-      }
-    }"
+    :cell-attrs="({ headerKey }: CellAttrParams) => ({
+      class: `${headerKey}-column`
+    })"
     empty-state-icon-size="96"
     disable-sorting
     hide-pagination-when-optional
@@ -53,6 +49,12 @@ import {
 } from '@kong/kongponents'
 import { computed, useSlots, ref, watch } from 'vue'
 
+type CellAttrParams = {
+  headerKey: string
+  row: any
+  rowIndex: number
+  colIndex: number
+}
 type FetcherParams = {
   page: number,
   pageSize: number,
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -22,23 +22,25 @@
     @row:click="click"
   >
     <template
-      v-if="slots.toolbar"
-      #toolbar
+      v-for="key in Object.keys(slots)"
+      :key="key"
+      #[key]="{ row, rowValue }"
     >
-      <div class="app-collection-toolbar">
-        <slot name="toolbar" />
-      </div>
-    </template>
+      <template
+        v-if="key === 'toolbar'"
+      >
+        <div class="app-collection-toolbar">
+          <slot name="toolbar" />
+        </div>
+      </template>
 
-    <template
-      v-for="key in Object.keys(remainingSlots)"
-      #[`${key}`]="{ row, rowValue }"
-    >
-      <slot
-        :name="key"
-        :row="row"
-        :row-value="rowValue"
-      />
+      <template v-else>
+        <slot
+          :name="key"
+          :row="row"
+          :row-value="rowValue"
+        />
+      </template>
     </template>
   </KTable>
 </template>
@@ -47,7 +49,7 @@
 import {
   KTable,
 } from '@kong/kongponents'
-import { computed, useSlots, ref, watch } from 'vue'
+import { useSlots, ref, watch } from 'vue'
 
 type CellAttrParams = {
   headerKey: string
@@ -84,12 +86,6 @@ const slots = useSlots()
 
 const items = ref<unknown[] | undefined>(props.items)
 const cacheKey = ref<number>(0)
-
-const remainingSlots = computed(() => {
-  const { toolbar, ...remainingSlots } = slots
-
-  return remainingSlots
-})
 
 watch(() => props.items, () => {
   cacheKey.value++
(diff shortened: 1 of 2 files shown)


src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -54,41 +59,117 @@ type ChangeValue = {
...
+      const hasItem = items.find((item) => item && typeof item === 'object' && 'id' in item && item.id === props.selectedRowId)
+
+      if (hasItem) {
+        currentSelectedRowId.value = props.selectedRowId
+      }
+    } else {
+      // Use the first item as a fallback.
+      const item = items[0]
+
+      if (item && typeof item === 'object' && 'id' in item && typeof item.id === 'string') {
```
kleinfreund, line 113: **Comment**: Not in love with the necessity for that degree of type-checking this value.

I’d rather type `props.items` as a list of objects that at the very least have an `id` property (so we can standardize on _one_ behavior for selecting an active row). But I didn’t want to add this here because it would increase the scope.

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -54,41 +59,117 @@ type ChangeValue = {
...
+  if ('id' in row && row.id === currentSelectedRowId.value) {
+    attributes.class = 'is-selected'
+  }
+
+  return attributes
+}
+
+function handleRowClickEvent(_event: any, row: any) {
+  currentSelectedRowId.value = row.id
+  emit('row:click', row)
```
kleinfreund, line 150: **Comment**: `AppCollection` should, at the very least currently, not contain the _behavior_ attached to row clicks. Specifically, navigating to the detail view on row click currently doesn’t make sense because almost none of our list views behave that way: instead, they select the item to show in the summary.

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -13,34 +13,39 @@
...
+      #toolbar
+    >
+      <div class="app-collection-toolbar">
+        <slot name="toolbar" />
+      </div>
+    </template>
+
+    <template
+      v-for="key in Object.keys(remainingSlots)"
+      #[`${key}`]="{ row, rowValue }"
```
johncowen, line 33: Could we just check if `key` equals `toolbar` and if so add the wrapping div? In fact from what I can see the KTable toolbar already has a wrapping `k-table-toolbar`, do we need to add yet another one?

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -71,6 +76,13 @@ const slots = useSlots()
 
 const items = ref<unknown[] | undefined>(props.items)
 const cacheKey = ref<number>(0)
+
+const remainingSlots = computed(() => {
+  const { toolbar, ...remainingSlots } = slots
+
+  return remainingSlots
+})
+
```
johncowen, line 85: If we can either do the `key === 'toolbar'`, or even better just reuse the `.k-table-toolbar` selector, when we can get rid of this code.

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -84,11 +96,28 @@ const click = (e: MouseEvent) => {
...
   }
 }
+
+function getCellAttributes({ headerKey }: any) {
+  return {
+    class: {
+      [`${headerKey}-column`]: true,
+    },
+  }
+}
```
johncowen, line 106: I would rather keep the `class` inline please.

src/app/application/components/app-collection/AppCollection.vue
kleinfreund: > Could we just check if key equals toolbar and if so add the wrapping div?

Technically, yes, but I don’t see how that’s better. Leaving this as is.

> In fact from what I can see the KTable toolbar already has a wrapping k-table-toolbar, do we need to add yet another one?

`.k-table-toolbar` has styles attached to it that I don’t want to re-use. Specifically that every element in toolbar should have `display: flex` is not desirable in my book. Having a separate element here makes the layout of toolbar items robust and independent from changes in Kongponents.

src/app/application/components/app-collection/AppCollection.vue
kleinfreund: See https://github.com/kumahq/kuma-gui/pull/1203#discussion_r1267649374 for why I’m not using the existing `.k-table-toolbar` selector for these styles.

src/app/application/components/app-collection/AppCollection.vue
kleinfreund: Why? Right now, we keep moving more and more logic into templates where it doesn’t belong. It’s already hard enough to parse what’s going on with a `KTable` that is fed a bunch of props, but adding inline functions to that makes it even worse.

src/app/application/components/app-collection/AppCollection.vue
johncowen: > .k-table-toolbar has styles attached to it that I don’t want to re-use. Specifically that every element in toolbar should have display: flex is not desirable in my book.

I'm not seeing this could you post a screengrab or something showing me these?

Screenshot 2023-07-19 at 08 57 11

lmk

kleinfreund: You have access to https://github.com/Kong/kongponents, right?

https://github.com/Kong/kongponents/blob/main/src/components/KTable/KTable.vue#L986-L992

src/app/application/components/app-collection/AppCollection.vue
johncowen: AppCollection.vue is 94 lines long. The template/html in there is 36 lines long, the vast majority of those 36 lines are just setting an attribute.

How hard is that to parse? Harder to parse than our listing components used to be?

src/app/application/components/app-collection/AppCollection.vue
johncowen: Just to note for history that we are continuing this conversation in the single place that @kleinfreund referred to.

src/app/application/components/app-collection/AppCollection.vue
kleinfreund: Templates should have hardly any logic in them at all. I’m just putting this function where it belongs.

If there is nothing particularly wrong with this, I’m leaving this as is.

src/app/application/components/app-collection/AppCollection.vue
johncowen: Yes I do, but thanks for linking thats super helpful

src/app/application/components/app-collection/AppCollection.vue
johncowen: Where is the logic?

kleinfreund: There isn’t really a point arguing this. Tell if this is a blocker for you and I’ll move it back.

johncowen: This time around its a blocker for me.

The reasons you have given for doing this:

1. It's hard to follow. Its not.
2. It's logic and therefore shouldn't be in the template. There's no logic.
3. There's nothing wrong with it. There was nothing wrong with it in the first place and no reason to change (see points 1 and 2)

Happy to hear more opinion, but as of right now, this is a pointless conversation, I think we can agree on that?

Happy to revisit once we have this stream of work complete. The code we are talking about is in a centralized place and not spread out in multiple places in the application.

kleinfreund: It’s not useful to argue this. I’ve moved it back.

johncowen:
I would really like to approve this and get it merged, I thought it might be helpful to make a suggestion before doing that in the form of a commit, mostly amends to the my own code:

https://github.com/johncowen/kuma-gui/commit/a31752ee9d237a8405b8757d60d2d2e421419993

Changes:

1. FetcherProps > FetcherParams. As they are params not props
2. CellAttrParams, so we know what we are getting here.  I think this is something you would prefer also? `:cell-attrs` remains inlined for the moment.
3. Use `v-if` to do the slot dance instead of `computed` plus destructuring.
4. Apparently, from looking at the link you provided `:deep` seems to be the Vue sanctioned way of what I had been using `:is()` for (crossing the slot), so I've also made that change.

I would really love it if you could work with me to make those ^ changes here also. It achieves what we want functionally and I would be super happy to then approve/merge this.

They are small things and I would much prefer be best to do them here if you could (I want to avoid yet another PR after this, ideally I would have done them in the original PR for AppCollection) and then leave this as is until we get the stream of work complete. At which point we can revisit this component if we find other improvements can be made.

Let me know.

johncowen (approved):
Awesome thankyou!

PR #1204, opened 2023-07-18
chore: Uses DataSource/AppCollection for Mesh listing

johncowen:
Integrates DataSource/AppCollection for Mesh listing

### DIFF
--- a/src/app/meshes/sources.ts
+++ b/src/app/meshes/sources.ts
@@ -26,7 +26,7 @@ export type MeshInsightSource = DataSourceResponse<MeshInsight>
 
 export const sources = (api: KumaApi) => {
   return {
-    '/:mesh/meshes': async (params: MeshParams & PaginationParams, source: Closeable) => {
+    '/meshes': async (params: MeshParams & PaginationParams, source: Closeable) => {
       source.close()
       const offset = params.size * (params.page - 1)
       return api.getAllMeshes({
@@ -40,11 +40,7 @@ export const sources = (api: KumaApi) => {
     },
     '/:mesh/insights': async (params: MeshParams, source: Closeable) => {
       source.close()
-      if (params.mesh === '*') {
-        return api.getAllMeshInsights()
-      } else {
-        return api.getMeshInsights({ name: params.mesh })
-      }
+      return api.getMeshInsights({ name: params.mesh })
     },
 
   }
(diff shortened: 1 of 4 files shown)


src/app/meshes/sources.ts
```diff
@@ -1,18 +1,51 @@
...
+
 export type MeshInsightSource = DataSourceResponse<MeshInsight>
+
 export const sources = (api: KumaApi) => {
   return {
-    '/:mesh/insights': async (params: MeshParams, source: {close: () => void}) => {
+    '/:mesh/meshes': async (params: MeshParams & PaginationParams, source: Closeable) => {
       source.close()
-      return api.getMeshInsights({ name: params.mesh })
+      const offset = params.size * params.page
```
kleinfreund, line 31: **Issue**: This needs to subtract one page count before multiplying as the first offset must be 0 (`size * (1 - 1)`).

```suggestion
      const offset = params.size * (params.page - 1)
```

src/app/meshes/sources.ts
```diff
@@ -1,18 +1,51 @@
...
+        offset,
+      })
     },
+    '/:mesh/mesh': (params: MeshParams, source: Closeable) => {
+      source.close()
+      return api.getMesh({ name: params.mesh })
+    },
+    '/:mesh/insights': async (params: MeshParams, source: Closeable) => {
+      source.close()
+      if (params.mesh === '*') {
```
lahabana, line 43: Why is mesh with 2 endpoints 1 for list and one for get single item and insights is not list this? Isn't this going to be confusing too because depending on the param different response types are going to be removed.

I'd be very much for having 2 sources for insights.

src/app/meshes/sources.ts
```diff
@@ -1,18 +1,51 @@
...
+export type MeshSource = DataSourceResponse<Mesh>
+export type MeshCollection = CollectionResponse<Mesh>
+export type MeshCollectionSource = DataSourceResponse<MeshCollection>
+
 export type MeshInsightSource = DataSourceResponse<MeshInsight>
+
 export const sources = (api: KumaApi) => {
   return {
-    '/:mesh/insights': async (params: MeshParams, source: {close: () => void}) => {
+    '/:mesh/meshes': async (params: MeshParams & PaginationParams, source: Closeable) => {
```
lahabana, line 29: I don't understand the ':mesh' param here. We're listing all meshes right? Why is there a Mesh param?

lahabana (commented):
Confused about your sources. Quite happy about the rest :)

src/app/meshes/sources.ts
johncowen: I changed this up, at this moment (at least) we only need the `/:mesh/insights` source.

src/app/meshes/sources.ts
johncowen: I've change this now also, now its just `/meshes`

PR #1205, opened 2023-07-18
chore: Move PolicyListing to use AppCollection

johncowen:
Move PolicyListing to use AppCollection

Note summary views have been removed here

### DIFF
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -21,6 +21,9 @@
             }
           )
         },
+        replace: (...args: Parameters<typeof router['push']>) => {
+          router.push(...args)
+        },
         params: Object.fromEntries(Object.entries(route.params).map(([prop, value]) => {
           return [
             prop,
--- a/src/app/policies/views/PolicyListView.vue
+++ b/src/app/policies/views/PolicyListView.vue
@@ -177,7 +177,7 @@
                                       policy: item.name,
                                     },
                                   },
-                                  label: 'View',
+                                  label: t('common.collection.actions.view'),
                                 }"
                               />
                             </template>
--- a/src/app/policies/views/PolicyListView.vue
+++ b/src/app/policies/views/PolicyListView.vue
@@ -85,51 +85,47 @@
                         @change="route.update"
                       >
                         <template #toolbar>
-                          <div
-                            class="toolbar"
-                          >
-                            <!--
+                          <!--
                               Load in all the totals for the policies so we can show them in the dropdown menu
                             -->
-                            <DataSource
-                              v-slot="{data: insights}: MeshInsightSource"
-                              :src="`/${route.params.mesh}/insights`"
+                          <DataSource
+                            v-slot="{data: insights}: MeshInsightSource"
+                            :src="`/${route.params.mesh}/insights`"
+                          >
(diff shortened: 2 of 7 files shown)


lahabana (commented):
Great to see things making progress towards standardization!

src/app/policies/views/PolicyListView.vue
```diff
@@ -219,8 +235,30 @@ const change = function (e: PolicyCollection) {
...
-  font-weight: var(--font-weight-regular);
+.toolbar {
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  flex-wrap: wrap;
+  gap: var(--spacing-md);
+  padding: var(--spacing-sm) var(--spacing-sm) 0 var(--spacing-sm);
+
+}
```
kleinfreund, line 246: **Suggestion**: These styles shouldn’t be own a user of AppCollection but rather AppCollection itself.

Can we remove them in favor of https://github.com/kumahq/kuma-gui/pull/1203?

src/app/application/components/route-view/RouteView.vue
```diff
@@ -21,6 +21,9 @@
             }
           )
         },
+        replace: (...args: Parameters<typeof router['push']>) => {
```
kleinfreund, line 24: **Suggestion**: This should be `replace`, right?

```suggestion
        replace: (...args: Parameters<typeof router['replace']>) => {
```

src/app/policies/views/PolicyListView.vue
johncowen: Yes exactly 👍

src/app/application/components/route-view/RouteView.vue
johncowen: Nope, we are calling `router.push` i.e. replace the current route entirely

kleinfreund: Ah, I got confused by the name of the method then. It should be renamed to `push` to avoid that.

johncowen: Its not push you are not pushing to a `route` you are replacing the `route` with a new one. You push to the `router`. This is the `route`.

johncowen: > Its not push you are not pushing to a route you are replacing the route with a new one. You push to the router. This is the route.

I don't think it's the same thing. The router and the route are completely different things, you push a new route to the router to make it the current route, you replace the current route with a new one to replace the current route.

kleinfreund: Alright. Resolving this as it’s not worth spending time on.

johncowen: Yeah my thoughts exactly.

johncowen: Is there anything else here that blocks an approval?

kleinfreund: There is an outstanding thread that hasn’t been addressed.

src/app/policies/views/PolicyListView.vue
johncowen: Ok this is also now done.

kleinfreund (approved):

PR #1207, opened 2023-07-18
chore: Convert Service listing to DataSource / AppCollection

johncowen:
Converts Service Listing to use DataSource and AppCollection

### DIFF
--- a/src/app/services/views/ServiceListView.vue
+++ b/src/app/services/views/ServiceListView.vue
@@ -1,6 +1,7 @@
 <template>
   <RouteView
     v-slot="{ route }"
+    name="services-list-view"
   >
     <DataSource
       v-slot="{data, error}: ServiceInsightCollectionSource"
@@ -8,12 +9,12 @@
     >
       <AppView>
         <template #title>
-          <h1>
+          <h2>
             <RouteTitle
               :title="t('services.routes.items.title')"
               :render="true"
             />
-          </h1>
+          </h2>
         </template>
         <KCard>
           <template #body>
--- a/src/app/services/views/ServiceListView.vue
+++ b/src/app/services/views/ServiceListView.vue
@@ -1,176 +1,138 @@
 <template>
-  <RouteView>
-    <RouteTitle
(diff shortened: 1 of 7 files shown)


lahabana (commented):
Sweet!

src/app/services/views/ServiceListView.vue
```diff
@@ -1,176 +1,146 @@
...
+                { label: 'Actions', key: 'actions', hideLabel: true },
+              ]"
+              :page-number="props.page"
+              :page-size="props.size"
+              :total="data?.total"
+              :items="data?.items"
+              :error="error"
+              @change="route.update"
+            >
+              <template #name="{ row: item }">
```
kleinfreund, line 40: **Question**: Why all this renaming to `item`?

src/app/services/views/ServiceListView.vue
```diff
@@ -1,176 +1,146 @@
...
+                  <template #items>
+                    <KDropdownItem
+                      :item="{
+                        to: {
+                          name: 'service-detail-view',
+                          params: {
+                            service: item.name,
+                          },
+                        },
+                        label: 'View',
```
kleinfreund, line 105: **Suggestion**: Please move this into the message file. Also, it should be "View details" to be consistent with other views.

src/app/services/views/ServiceListView.vue
johncowen: Because its more generic than `row`, ie. if I want to copy/paste something from inside the slot elsewhere (i.e. to somewhere where it isn't a `row`) the word `item` still works. I don't have to copy/paste and then rename the variable from row to something else.

src/app/services/views/ServiceListView.vue
johncowen: Moved

src/app/services/views/ServiceListView.vue
johncowen: Is this a **Question** or a blocker?

kleinfreund (approved):

PR #1209, opened 2023-07-19
refactor: migrates zone lists to use DataSource

kleinfreund:
**refactor: migrates Zone CP list to use DataSource**

Migrates Zone CP list view to use `DataSource` and `AppCollection`.

**refactor: migrates Zone Ingress list to use DataSource**

Migrates Zone Ingress list view to use `DataSource` and `AppCollection`.

**refactor: migrates Zone Egress list to use DataSource**

Migrates Zone Egress list view to use `DataSource` and `AppCollection`.

### DIFF
--- a/features/zones/PageNavigation.feature
+++ b/features/zones/PageNavigation.feature
@@ -39,7 +39,7 @@ Feature: Zones: Page navigation
     Then the page title contains "Control Planes"
 
     Examples:
-      | RouteName              | TableSelector                      | DetailViewTitle | ListViewTitle  |
-      | zone-cp-list-view      | [data-testid='zone-cp-table']      | Control Plane   | Control Planes |
-      | zone-ingress-list-view | [data-testid='zone-ingress-table'] | Ingress         | Ingresses      |
-      | zone-egress-list-view  | [data-testid='zone-egress-table']  | Egress          | Egresses       |
+      | RouteName              | TableSelector                           | DetailViewTitle | ListViewTitle  |
+      | zone-cp-list-view      | [data-testid='zone-cp-collection']      | Control Plane   | Control Planes |
+      | zone-ingress-list-view | [data-testid='zone-ingress-collection'] | Ingress         | Ingresses      |
+      | zone-egress-list-view  | [data-testid='zone-egress-collection']  | Egress          | Egresses       |
--- a/features/zones/PageNavigation.feature
+++ b/features/zones/PageNavigation.feature
@@ -5,7 +5,6 @@ Feature: Zones: Page navigation
       | breadcrumbs | .k-breadcrumbs                      |
       | main-nav    | .app-sidebar                        |
       | nav-tabs    | [data-testid='nav-tabs']            |
-      | summary     | [data-testid='list-view-summary']   |
       | table       | [data-testid='data-overview-table'] |
       | table-row   | $table tbody tr                     |
 
@@ -22,25 +21,25 @@ Feature: Zones: Page navigation
         mode: global
       """
 
-    When I visit the "/" URL
-    And I click the "$main-nav .nav-item-zone-cp-list-view > a" element
+    When I visit the "/zones" URL
+    # This `cy.wait` stabilizes the test significantly. For some reason, it can happen that the subsequent navigation to via nav tab is never triggered.
+    Then I wait for 1000 milliseconds
 
     When I click the "$nav-tabs #<RouteName>-tab a" element
     Then the page title contains "<ListViewTitle>"
 
-    When I click the "$table-row:nth-child(1) [data-testid='detail-view-link']" element
+    When I click the "<TableSelector> tbody tr:nth-child(1) [data-testid='detail-view-link']" element
     Then the page title contains "<DetailViewTitle>"
 
     When I click the "$nav-tabs #<RouteName>-tab a" element
     Then the page title contains "<ListViewTitle>"
-    Then the "$table" element exists
-    And the "$summary" element exists
+    Then the "<TableSelector>" element exists
 
     When I click the "$breadcrumbs > .k-breadcrumbs-item:nth-child(1) > a" element
     Then the page title contains "Zone CPs"
 
     Examples:
-      | RouteName              | DetailViewTitle | ListViewTitle  |
-      | zone-cp-list-view      | Zone CP         | Zone CPs       |
-      | zone-ingress-list-view | Zone Ingress    | Zone Ingresses |
-      | zone-egress-list-view  | Zone Egress     | Zone Egresses  |
+      | RouteName              | TableSelector                      | DetailViewTitle | ListViewTitle  |
+      | zone-cp-list-view      | [data-testid='zone-cp-table']      | Zone CP         | Zone CPs       |
+      | zone-ingress-list-view | [data-testid='zone-ingress-table'] | Zone Ingress    | Zone Ingresses |
+      | zone-egress-list-view  | [data-testid='zone-egress-table']  | Zone Egress     | Zone Egresses  |
--- a/features/zones/PageNavigation.feature
+++ b/features/zones/PageNavigation.feature
@@ -36,10 +36,10 @@ Feature: Zones: Page navigation
     Then the "<TableSelector>" element exists
 
     When I click the "$breadcrumbs > .k-breadcrumbs-item:nth-child(1) > a" element
-    Then the page title contains "Zone CPs"
+    Then the page title contains "Control Planes"
 
     Examples:
       | RouteName              | TableSelector                      | DetailViewTitle | ListViewTitle  |
-      | zone-cp-list-view      | [data-testid='zone-cp-table']      | Zone CP         | Zone CPs       |
-      | zone-ingress-list-view | [data-testid='zone-ingress-table'] | Zone Ingress    | Zone Ingresses |
-      | zone-egress-list-view  | [data-testid='zone-egress-table']  | Zone Egress     | Zone Egresses  |
+      | zone-cp-list-view      | [data-testid='zone-cp-table']      | Control Plane   | Control Planes |
+      | zone-ingress-list-view | [data-testid='zone-ingress-table'] | Ingress         | Ingresses      |
+      | zone-egress-list-view  | [data-testid='zone-egress-table']  | Egress          | Egresses       |
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -10,8 +10,8 @@ zone-cps:
       breadcrumbs: Control Planes
       headers:
         name: 'Name'
-        zoneCpVersion: 'Zone CP Version'
-        storeType: 'Storage type'
+        zonecpversion: 'Zone CP Version'
+        storetype: 'Storage type'
         status: 'Status'
         warnings: 'Warnings'
         actions: 'Actions'
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -8,13 +8,6 @@ zone-cps:
     items:
       title: Control Planes
       breadcrumbs: Control Planes
-      headers:
-        name: 'Name'
-        zonecpversion: 'Zone CP Version'
-        storetype: 'Storage type'
-        status: 'Status'
-        warnings: 'Warnings'
-        actions: 'Actions'
 
 zone-ingresses:
   routes:
@@ -24,10 +17,6 @@ zone-ingresses:
     items:
       title: Ingresses
       breadcrumbs: Ingresses
-      headers:
-        name: 'Name'
-        status: 'Status'
-        actions: 'Actions'
 
 zone-egresses:
   routes:
@@ -37,10 +26,6 @@ zone-egresses:
     items:
       title: Egresses
       breadcrumbs: Egresses
-      headers:
-        name: 'Name'
-        status: 'Status'
-        actions: 'Actions'
 
 zones:
   href:
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -8,6 +8,15 @@ zone-cps:
     items:
       title: Control Planes
       breadcrumbs: Control Planes
+  list:
+    tableHeaders:
+      name: 'Name'
+      zoneCpVersion: 'Zone CP Version'
+      storeType: 'Storage type'
+      status: 'Status'
+      warnings: 'Warnings'
+      actions: 'Actions'
+
 zone-ingresses:
   routes:
     item:
@@ -16,6 +25,12 @@ zone-ingresses:
     items:
       title: Ingresses
       breadcrumbs: Ingresses
+  list:
+    tableHeaders:
+      name: 'Name'
+      status: 'Status'
+      actions: 'Actions'
+
 zone-egresses:
   routes:
     item:
@@ -24,6 +39,12 @@ zone-egresses:
     items:
       title: Egresses
       breadcrumbs: Egresses
+  list:
+    tableHeaders:
+      name: 'Name'
+      status: 'Status'
+      actions: 'Actions'
+
 zones:
   href:
     docs:
@@ -156,6 +177,3 @@ zones:
       proceedText: 'Yes, delete'
       title: 'Delete Zone'
       errorText: 'An unexpected error occurred'
-  list:
-    viewDetailsActionText: 'View details'
-    deleteActionText: 'Delete'
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -2,28 +2,28 @@ zone-cps:
   routes:
     item:
       title: "{name} Zone CP"
-      breadcrumbs: Zone CPs
+      breadcrumbs: Control Planes
       config:
         no-subscriptions: This zone has no subscriptions
     items:
-      title: Zone CPs
-      breadcrumbs: Zone CPs
+      title: Control Planes
+      breadcrumbs: Control Planes
 zone-ingresses:
   routes:
     item:
       title: "{name} Zone Ingress"
-      breadcrumbs: Zone Ingresses
+      breadcrumbs: Ingresses
     items:
-      title: Zone Ingresses
-      breadcrumbs: Zone Ingresses
+      title: Ingresses
+      breadcrumbs: Ingresses
 zone-egresses:
   routes:
     item:
       title: "{name} Zone Egress"
-      breadcrumbs: Zone Egresses
+      breadcrumbs: Egresses
     items:
-      title: Zone Egresses
-      breadcrumbs: Zone Egresses
+      title: Egresses
+      breadcrumbs: Egresses
 zones:
   href:
     docs:
@@ -32,11 +32,12 @@ zones:
     create:
       title: 'Create & connect Zone'
     items:
+      title: Zones
       breadcrumbs: Zones
       navigation:
-        zone-cp-list-view: Zone CPs
-        zone-ingress-list-view: Zone Ingresses
-        zone-egress-list-view: Zone Egresses
+        zone-cp-list-view: Control Planes
+        zone-ingress-list-view: Ingresses
+        zone-egress-list-view: Egresses
   form:
     exit: 'Exit'
     nameLabel: 'Name'
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -1,7 +1,7 @@
 zone-cps:
   routes:
     item:
-      title: "{name} Zone CP"
+      title: "{name} Control Plane"
       breadcrumbs: Control Planes
       config:
         no-subscriptions: This zone has no subscriptions
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -8,14 +8,13 @@ zone-cps:
     items:
       title: Control Planes
       breadcrumbs: Control Planes
-  list:
-    tableHeaders:
-      name: 'Name'
-      zoneCpVersion: 'Zone CP Version'
-      storeType: 'Storage type'
-      status: 'Status'
-      warnings: 'Warnings'
-      actions: 'Actions'
+      headers:
+        name: 'Name'
+        zoneCpVersion: 'Zone CP Version'
+        storeType: 'Storage type'
+        status: 'Status'
+        warnings: 'Warnings'
+        actions: 'Actions'
 
 zone-ingresses:
   routes:
@@ -25,11 +24,10 @@ zone-ingresses:
     items:
       title: Ingresses
       breadcrumbs: Ingresses
-  list:
-    tableHeaders:
-      name: 'Name'
-      status: 'Status'
-      actions: 'Actions'
+      headers:
+        name: 'Name'
+        status: 'Status'
+        actions: 'Actions'
 
 zone-egresses:
   routes:
@@ -39,11 +37,10 @@ zone-egresses:
     items:
       title: Egresses
       breadcrumbs: Egresses
-  list:
-    tableHeaders:
-      name: 'Name'
-      status: 'Status'
-      actions: 'Actions'
+      headers:
+        name: 'Name'
+        status: 'Status'
+        actions: 'Actions'
 
 zones:
   href:
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -155,3 +155,6 @@ zones:
       proceedText: 'Yes, delete'
       title: 'Delete Zone'
       errorText: 'An unexpected error occurred'
+  list:
+    viewDetailsActionText: 'View details'
+    deleteActionText: 'Delete'
--- a/src/app/zones/sources.ts
+++ b/src/app/zones/sources.ts
@@ -1,0 +1,47 @@
+import { DataSourceResponse } from '@/app/application/services/data-source/DataSourcePool'
+import type KumaApi from '@/services/kuma-api/KumaApi'
+import type { PaginatedApiListResponse as CollectionResponse } from '@/types/api.d'
+import type { ZoneOverview, ZoneIngressOverview, ZoneEgressOverview } from '@/types/index.d'
+import { fetchAllResources } from '@/utilities/helpers'
+
+type PaginationParams = {
+  size: number
+  page: number
+}
+
+export type ZoneOverviewCollection = CollectionResponse<ZoneOverview>
+export type ZoneOverviewSource = DataSourceResponse<ZoneOverview>
+export type ZoneOverviewCollectionSource = DataSourceResponse<ZoneOverviewCollection>
+
+export type AllZoneIngressOverviewCollection = CollectionResponse<ZoneIngressOverview>
+export type AllZoneIngressOverviewSource = DataSourceResponse<ZoneIngressOverview>
+export type AllZoneIngressOverviewCollectionSource = DataSourceResponse<AllZoneIngressOverviewCollection>
+
+export type AllZoneEgressOverviewCollection = CollectionResponse<ZoneEgressOverview>
+export type AllZoneEgressOverviewSource = DataSourceResponse<ZoneEgressOverview>
+export type AllZoneEgressOverviewCollectionSource = DataSourceResponse<AllZoneEgressOverviewCollection>
+
+export const sources = (api: KumaApi) => {
+  return {
+    '/zones/zone-cps': async (params: PaginationParams, source: { close: () => void }) => {
+      source.close()
+
+      const size = params.size
+      const offset = params.size * (params.page - 1)
+
+      return await api.getAllZoneOverviews({ size, offset })
+    },
+
+    '/zones/all-zone-ingresses': async (_params: unknown, source: { close: () => void }) => {
+      source.close()
+
+      return await fetchAllResources(api.getAllZoneIngressOverviews.bind(api))
+    },
+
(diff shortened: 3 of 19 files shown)


src/app/zones/views/ZoneListView.vue
```diff
@@ -13,103 +15,180 @@
...
-              @table-action="loadEntity"
-              @load-data="loadData"
+          <DataSource
+            v-slot="{ data: allZoneIngressOverviewData }: AllZoneIngressOverviewCollectionSource"
+            :src="`/zones/all-zone-ingresses`"
+          >
+            <DataSource
+              v-slot="{ data: allZoneEgressOverviewData }: AllZoneEgressOverviewCollectionSource"
+              :src="`/zones/all-zone-egresses`"
             >
```
kleinfreund, line 32: **Comment**: This is not great. Ideally, we would be able to get multiple data sources out of one `DataSource` usage to mirror a `Promise.all([...])` use case (or is that somehow already possible?).

One problem here, besides it being annoying to write down like this in the template, is that we only use the `error` from the first block.

features/zones/PageNavigation.feature
```diff
@@ -28,19 +27,18 @@ Feature: Zones: Page navigation
     When I click the "$nav-tabs #<RouteName>-tab a" element
     Then the page title contains "<ListViewTitle>"
```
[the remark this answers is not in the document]
kleinfreund: I found that adding a short `cy.wait` in between the two navigations here largely stabilizes this test.

On a side note, should we explicitly set `--browser chrome` for the Cypress commands in our package.json so we always run tests in Chrome rather than Electron?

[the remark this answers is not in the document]
kleinfreund: Actually, the test is flaky in both headed and headless mode, but seems much more severe in headless mode (where the test passes most of the time).

Now that I think more about this difference, I’ve retried running the tests locally in headless mode but without video turned on and that seems to really impact the flakyness. I found _some_ indication that this could indeed be related (https://stackoverflow.com/questions/71872119/cypress-fails-tests-randomly-in-headless-mode-but-all-passes-all-tests-in-real-b), but it’s not very conclusive.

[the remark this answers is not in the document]
kleinfreund: Looking at the CI test runs and its videos, it looks like whenever we hit the home page with the charts, the page kind of freezes for a moment.

I’m changing this test to go to `/zones` instead of `/` to rule out this isn’t related.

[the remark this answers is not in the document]
kleinfreund: 🤷🏻 

image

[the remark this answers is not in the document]
kleinfreund: Another angle seems to be using Cypress video recordings with GitHub Actions runners: https://github.com/cypress-io/cypress/issues/8210

One piece of advice I’ve seen is switching to use Google Chrome as the browser.

If nothing works, I’ll use https://docs.cypress.io/guides/references/troubleshooting#Log-memory-and-CPU-usage to profile hardware usage.

src/app/zones/sources.ts
```diff
@@ -0,0 +1,73 @@
...
+      return await api.getAllZoneOverviews({ size, offset })
+    },
+
+    '/zones/all-zone-ingresses': async (_params: unknown, source: { close: () => void }) => {
+      source.close()
+
+      return await fetchAllResources(api.getAllZoneIngressOverviews.bind(api))
+    },
+
+    '/zones/all-zone-egresses': async (_params: unknown, source: { close: () => void }) => {
```
lahabana, line 49: Would naming this zones/zone-egresses/overviews help?

lahabana: Also why do we want this kind of calls?

kleinfreund: > Would naming this zones/zone-egresses/overviews help?

I chose `all-` here because of what I’m explaining below. Technically, the `/zones/zone-ingresses` and `/zones/zone-egresses` entries are also for their respective *Overview representation, but there we’re actually talking about paginated results.

> Also why do we want this kind of calls?

This looks pretty much like a hack from old times.

In short: We fetch _all_ ingresses and egresses just to be able to associated them with the Zone CPs (based on `zoneIngress.zone`/`zoneEgress.zone`) and we use this for exactly one thing: showing Yes or No in the respective column in the table. We could drop all of this if a `ZoneOverview` had something like `hasIngress: boolean` and `hasEgress: boolean`.

lahabana: Yes please drop show ZoneIngress/ZoneEgress for the moment please :) We'll reintroduce this later. (We were actually talking about showing ingress/egress counts with health instead of yes/no

kleinfreund: I’ve removed them in https://github.com/kumahq/kuma-gui/pull/1209/commits/23562e577553fe088f50a30d1ad8112bb3587057.

src/app/zones/views/ZoneListView.vue
kleinfreund: Resolving this as it doesn’t apply to these components anymore as of https://github.com/kumahq/kuma-gui/pull/1209#discussion_r1268097990.

features/zones/PageNavigation.feature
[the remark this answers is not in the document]
kleinfreund: Tests passed a few times now so I’m resolving this for the time being.

src/app/zones/sources.ts
```diff
@@ -0,0 +1,52 @@
...
+    '/zones/zone-egresses': async (params: PaginationParams, source: { close: () => void }) => {
+      source.close()
+
+      const size = params.size
+      const offset = params.size * (params.page - 1)
+
+      return await api.getAllZoneEgressOverviews({ size, offset })
+    },
+  }
+}
```
johncowen, line 52: There is no need to prefix these with `/zones`. I know they look like our URLs but they aren't.

To help to explain, we are currently prefixing twice here `/zones/zone-ingresses`, we may as well use `/zone-ingresses`

src/app/zones/views/ZoneEgressListView.vue
```diff
@@ -13,128 +15,94 @@
...
-              :is-loading="isLoading"
+      <DataSource
+        v-slot="{ data: zoneEgressOverviewData, error }: ZoneEgressOverviewCollectionSource"
+        :src="`/zones/zone-egresses?size=${props.size}&page=${props.page}`"
+      >
+        <KCard>
+          <template #body>
+            <AppCollection
+              data-testid="zone-egress-table"
+              :headers="HEADERS"
```
johncowen, line 26: Can we just put the headers in here please, it makes it so much easier to see what is going on.

src/app/zones/views/ZoneEgressListView.vue
```diff
@@ -13,128 +15,94 @@
...
     >
-      <div class="zoneegresses">
-        <div class="stack">
-          <div class="kcard-border">
-            <DataOverview
-              :selected-entity-name="entity?.name"
-              :page-size="PAGE_SIZE_DEFAULT"
-              :is-loading="isLoading"
+      <DataSource
+        v-slot="{ data: zoneEgressOverviewData, error }: ZoneEgressOverviewCollectionSource"
```
johncowen, line 19: What is the reason for renaming `data`?

src/app/zones/views/ZoneIngressListView.vue
```diff
@@ -13,142 +15,101 @@
...
-        <MultizoneInfo v-if="store.getters['config/getMulticlusterStatus'] === false" />
+      <MultizoneInfo v-if="store.getters['config/getMulticlusterStatus'] === false" />
 
-        <!-- Zone CPs information for when Multicluster is enabled -->
-        <div
-          v-else
-          class="stack"
+      <template v-else>
+        <DataSource
+          v-slot="{ data: zoneIngressOverviewData, error }: ZoneIngressOverviewCollectionSource"
```
johncowen, line 22: Same comment here on the renaming, can we just stick with `data` without the rename?

src/app/zones/views/ZoneEgressListView.vue
```diff
@@ -13,128 +15,94 @@
...
-          >
-            <ZoneEgressDetails :zone-egress-overview="entity" />
-          </div>
-        </div>
-      </div>
+              @change="route.update"
+            >
+              <template #name="{ row }">
+                <RouterLink
+                  :to="row.detailViewRoute"
```
johncowen, line 36: Can we just add this stuff inline?

johncowen (commented):
I made some initial comments that apply in several places, I think you get the idea of where I'm going with those.

I haven't finished looking at this yet but wanted to post these.

src/app/zones/views/ZoneEgressListView.vue
kleinfreund: I changed this, but I don’t understand what this brings you. But I don’t argue this again.

src/app/zones/views/ZoneEgressListView.vue
kleinfreund: No. I’ve explained several times now that I really don’t want to have all of this in the template and it worked like this before as well. I see no reason to move this into the template.

src/app/zones/sources.ts
kleinfreund: Changed this.

johncowen:
Oh I was also gonna say, I think we need some of the titling changes here also?

kleinfreund:
> Oh I was also gonna say, I think we need some of the titling changes here also?

Hm, I just noticed something: this title should be rendered above the nav tabs, right? Adding them for Zones looks like this:

image

johncowen:
I don't think so, the titles for the sub tabs should be underneath the tabs, from what I understand from the pattern is that above the tabs it should probably say "Zones" and below the tabs depending on which tab you are in it should say "Zone CPs" etc.

kleinfreund:
@johncowen I’ve added https://github.com/kumahq/kuma-gui/pull/1209/commits/b5300c884141ff10282a8fd6b5f5a1b3cdc3e3eb now. This adds both h1 and h2 level headings to the Zone views and aligns all the titles with the design (e.g. replacing `Zone CPs` with `Control Planes`, etc.). Does that make sense?

johncowen:
Could you change this up to use `t('common.collection.actions.view')` please (from the other PR you reviewed, you'll need to rebase). Also can we avoid using camelCase in the translation files where possible. ty!

kleinfreund:
I’ve changed these keys again.

johncowen:
There seems to be a lot of changes in here that aren't related to the PR, could you remove these changes?

Also:

https://github.com/kumahq/kuma-gui/pull/1209/files#diff-c5427ff78517c84149faacfc941a64032025a794d9e61c4fc6c08695a7a7a70bL105

It looks like since my last comment, you've changed things to do the opposite of what I was asking?

> Could you change this up to use t('common.collection.actions.view') please (from the other PR you reviewed, you'll need to rebase). Also can we avoid using camelCase in the translation files where possible. ty!

Please can you change it to use `t('common.collection.actions.view')` and avoid camelCase generally.

kleinfreund:
> There seems to be a lot of changes in here that aren't related to the PR, could you remove these changes?

Which? Review for this has dragged out quite a bit by this point and unfortunately some changes were necessary (specifically, ensuring column styles are scoped to the particular table).

> It looks like since my last comment, you've changed things to do the opposite of what I was asking?

It’s hardly "the opposite". I used the new key (collection) that you introduced and added my additions there. I don’t intend to rename the key from "viewDetails" to "view". The former is perfectly adequate and already in use.

Also, I don’t intend to avoid camel-cased keys. I see no benefit in doing so but rather downsides. It promotes short names which aren’t always adequate for a given context.

kleinfreund:
I’ve updated this once more to extract the following two change sets:

- https://github.com/kumahq/kuma-gui/pull/1225
- https://github.com/kumahq/kuma-gui/pull/1224

Please finish reviewing this.

johncowen:
I'm having another look at this now, thankyou for changing things to use `-collection` over `-table`

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,29 +1,50 @@
...
+      breadcrumbs: Control Planes
       config:
         no-subscriptions: This zone has no subscriptions
     items:
-      title: Zone CPs
-      breadcrumbs: Zone CPs
+      title: Control Planes
+      breadcrumbs: Control Planes
+  list:
+    tableHeaders:
```
johncowen, line 12:
```yaml
zone-cps.route.items.headers:
  version:
  type:
  status:
  warnings:
# or (up to you)
zone-cps.route.items.collection.headers:
  version:
  type:
  status:
  warnings:
```

```yaml
common.collection.headers:
  actions: Actions
```

Alternatively, as I know this is all about rolling out DataSource so we shouldn't really be doing i18n at the same time, I'd be fine if you removed this and come back later when we can go through and i18n all the collections together in one go to prevent going over this all over again.

Lmk

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,29 +1,50 @@
...
-      title: Zone Ingresses
-      breadcrumbs: Zone Ingresses
+      title: Ingresses
+      breadcrumbs: Ingresses
+  list:
+    tableHeaders:
+      name: 'Name'
+      status: 'Status'
+      actions: 'Actions'
+
```
johncowen, line 33: As above

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,29 +1,50 @@
...
-      title: Zone Egresses
-      breadcrumbs: Zone Egresses
+      title: Egresses
+      breadcrumbs: Egresses
+  list:
+    tableHeaders:
+      name: 'Name'
+      status: 'Status'
+      actions: 'Actions'
+
```
johncowen, line 47: As above

src/app/zones/locales/en-us/index.yaml
kleinfreund: Done.

src/app/zones/locales/en-us/index.yaml
kleinfreund: Done.

src/app/zones/locales/en-us/index.yaml
kleinfreund: Done.

src/app/zones/views/ZoneEgressListView.vue
johncowen: Please can you see if you can inline this, looking at it myself it looks like we can. There's no point reshaping all this data in some function down the bottom of the page when you can just do it inline.

src/app/zones/locales/en-us/index.yaml
kleinfreund: Done.

src/app/zones/views/ZoneEgressListView.vue
kleinfreund: I’m not doing this. There is no point in reshaping all this data in the template when we can do so how it is done right now.

johncowen: Same goes for any other places in this PR were we aren't inlining this sort of simple stuff. If its not simple stuff please let me know.

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,29 +1,47 @@
...
         no-subscriptions: This zone has no subscriptions
     items:
-      title: Zone CPs
-      breadcrumbs: Zone CPs
+      title: Control Planes
+      breadcrumbs: Control Planes
+      headers:
+        name: 'Name'
+        zoneCpVersion: 'Zone CP Version'
+        storeType: 'Storage type'
```
johncowen, line 14: Please can you stop using camelCasing for i18n keys, I would imagine you can change these to be

```yaml
  version:
#and
  type:
#or
  storage:
```

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,29 +1,47 @@
...
+        status: 'Status'
+        warnings: 'Warnings'
+        actions: 'Actions'
```
johncowen, line 17: `actions:` should come from `common`

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,29 +1,47 @@
...
+      breadcrumbs: Ingresses
     items:
-      title: Zone Ingresses
-      breadcrumbs: Zone Ingresses
+      title: Ingresses
+      breadcrumbs: Ingresses
+      headers:
+        name: 'Name'
+        status: 'Status'
+        actions: 'Actions'
```
johncowen, line 30: As above re: `actions`

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,29 +1,47 @@
...
+      breadcrumbs: Egresses
     items:
-      title: Zone Egresses
-      breadcrumbs: Zone Egresses
+      title: Egresses
+      breadcrumbs: Egresses
+      headers:
+        name: 'Name'
+        status: 'Status'
+        actions: 'Actions'
```
johncowen, line 43: As above re: actions

src/app/zones/locales/en-us/index.yaml
johncowen: I've unresolved this as I spotted a few places where we are still using camelCasing and not pulling `actions` from `common`

kleinfreund: I have no intention of going about finding all instances of camel-cased message keys. Also, the table headers are pulled from the zone message file. I won’t move them into common because they don’t belong there. Aside from that, the _actions_ actually all reference the common message file.

Resolving this again.

src/app/zones/locales/en-us/index.yaml
kleinfreund: I’ve changed this to be lowercase. I won’t address anymore nitpicks related to message key casing.

src/app/zones/locales/en-us/index.yaml
kleinfreund: Not moving this. Table headers should come from one place even though this might be duplicated.

src/app/zones/locales/en-us/index.yaml
kleinfreund: Not moving this. Table headers should come from one place even though this might be duplicated.

src/app/zones/locales/en-us/index.yaml
kleinfreund: Not moving this. Table headers should come from one place even though this might be duplicated.

src/app/zones/locales/en-us/index.yaml
johncowen: Ok so it looks like you've changed some of these now, let me take another look.

> I have no intention of going about finding all instances of camel-cased message keys

I don't want you to do that, only stop adding new ones, there were plenty of new ones in this PR.

>  Also, the table headers are pulled from the zone message file. I won’t move them into common because they don’t belong there. Aside from that, the actions actually all reference the common message file.

This sounds like what I would like, let me look again now you've made these other changes.

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,29 +1,47 @@
...
         no-subscriptions: This zone has no subscriptions
     items:
-      title: Zone CPs
-      breadcrumbs: Zone CPs
+      title: Control Planes
+      breadcrumbs: Control Planes
+      headers:
+        name: 'Name'
+        zonecpversion: 'Zone CP Version'
+        storetype: 'Storage type'
```
johncowen, line 14: Why can't these just be `version` and `storage`? What am I missing here?

src/app/zones/locales/en-us/index.yaml
johncowen: All actions columns for all collections should reference the same piece of text. We don't want to have potentially 10s of places in the i18n files that just repeat `Actions`. Please pull this from common.

src/app/zones/locales/en-us/index.yaml
kleinfreund: It’s not important that they are those particular values.

src/app/zones/views/ZoneIngressListView.vue
```diff
@@ -13,142 +13,144 @@
...
-      </div>
+          <KCard>
+            <template #body>
+              <AppCollection
+                class="zone-ingress-collection"
+                data-testid="zone-ingress-collection"
+                :headers="[
+                  { label: 'Name', key: 'name' },
+                  { label: 'Status', key: 'status' },
+                  { label: 'Actions', key: 'actions', hideLabel: true },
```
johncowen, line 40: Why do we have the labels here and i18n? Either one or the other

src/app/zones/locales/en-us/index.yaml
kleinfreund: I don’t want to fragment the source of column names so I’m not moving them. Feel free to file a follow up PR if you want to do this differently.

src/app/zones/locales/en-us/index.yaml
johncowen: Ok then if its not important to you can you change it please?

`version` and `storage` would be ideal (unless I'm missing something), partly so we aren't namespacing with `zonecp` twice. We know its a zonecp already.

src/app/zones/views/ZoneIngressListView.vue
kleinfreund: I forgot to remove them. I’ve done so now.

src/app/zones/views/ZoneListView.vue
```diff
@@ -226,20 +226,20 @@ function transformToTableData(zoneOverviews: ZoneOverview[]): ZoneOverviewTableR
         zone: name,
       },
     }
-    let zoneCpVersion = '-'
```
johncowen, line 223: Why is all this so tightly coupled to the name of the i18n key?

johncowen:
Ok looking now

src/app/zones/views/ZoneEgressListView.vue
```diff
@@ -144,25 +153,32 @@ function transformToTableData(zoneEgressOverviews: ZoneEgressOverview[]): ZoneEg
...
     return {
-      entity,
+      id,
       detailViewRoute,
+      name,
       status,
     }
   })
 }
+</script>
```
johncowen, line 163: I think I commented on this before but I can no longer see the comment on GH. Is there any reason not to just do this inline?

kleinfreund: One of the other threads is here: https://github.com/kumahq/kuma-gui/pull/1209#discussion_r1268155151

> I’m not doing this. There is no point in reshaping all this data in the template when we can do it how we do it right now.

src/services/production.ts
```diff
@@ -239,8 +239,11 @@ export const services: ServiceConfigurator<SupportedTokens> = ($) => [
       $.store,
     ],
   }],
+
+  // Modules
+  ...zoneModule($),
   ...meshes($),
-  ...servicesModule($),
+  ...serviceModule($),
```
johncowen, line 246: Could you put this back please?

They are all currently plural, like the folder names, I would also prefer it if you used `zones($)` like the rest.

src/app/zones/views/ZoneEgressListView.vue
johncowen: Half the point of this stream of work is to reduce all the complexity in just rendering some JSON to a list. I posted the other day on Slack about the "Small functions considered Harmful".

If there is something here that is overly-complex to do inline please point it out to me.

src/services/production.ts
kleinfreund: I’ve made them both use the plural.

johncowen:
https://github.com/kumahq/kuma-gui/pull/1209#pullrequestreview-1543430085

> Why is all this so tightly coupled to the name of the i18n key?

Thanks for undoing the i18n changes, I wanted to point out and underline that we shouldn't be tightly coupling this sort of code to the name of the i18n key, and I'm guessing you've come across the reasons why we shoudn't. No further action on this one.

src/app/zones/views/ZoneEgressListView.vue
kleinfreund: It doesn’t need changing. It’s fine. A function transforming data between formats isn’t increasing the complexity of a component.

I really, really, really, really don’t want to change this. There is no need and I personally think the suggested alternative is worse (something which I’ve pointed out numerous times by now).

src/app/zones/views/ZoneEgressListView.vue
```diff
@@ -144,25 +153,32 @@ function transformToTableData(zoneEgressOverviews: ZoneEgressOverview[]): ZoneEg
     const status = getItemStatusFromInsight(entity.zoneEgressInsight ?? {})
 
     return {
-      entity,
+      id,
```
johncowen, line 156: What is `id` for?

kleinfreund: I want to start providing IDs for all tables so that we have an easier time later on implementing mechanisms relying on the ability for identifying rows again. If every table already has an `id` (which we can ensure unlike using `name` which may not be unique within a table), adding mechanisms like a `row:select` event are easier.

kleinfreund: I’ve removed id.

kleinfreund: We do: service list view lists both services and external services and I’m pretty sure uniqueness between both resource types isn’t guaranteed.

Also, for resources that do have their own IDs, their names aren’t necessarily unique either.

johncowen: OK so we just checked with backend and it sounds like most things are unique within a mesh. There are potentially outliers where it "technically" could be an issue, in which case instead of having a unique ID we should be able to qualify it with something else:

In the case of:

1. Service/ExternalService we could use the `serviceType`
2. Anything else generally the name of the mesh.

Important to note we mostly view things by mesh anyway.

It doesn't sounds like this is anything pressing, and from the sounds of it we may have something added in the backend to gives us something unique. So I would say this is up in the air for the moment at least.

Lastly, we don't use unique ids anywhere for the moment, and for the purpose of this PR the `id` var wasn't being used at all anyway, and I don't think that previously it was being used to provide some sort of unique or qualified id anyway.

Just to be clear, no further action required from me here, just noting for historical purposes.

src/app/zones/views/ZoneEgressListView.vue
```diff
@@ -13,125 +13,132 @@
...
+              data-testid="zone-egress-collection"
+              :headers="[
+                { label: 'Name', key: 'name' },
+                { label: 'Status', key: 'status' },
+                { label: 'Actions', key: 'actions', hideLabel: true },
+              ]"
+              :page-number="props.page"
+              :page-size="props.size"
+              :total="data?.total"
+              :items="data ? transformToTableData(data.items) : []"
```
johncowen, line 42: If data isn't set yet it means `:items` should be empty as in "undefined empty", rather than an empty array, which would translate to "this has loaded but there are zero items". TLDR; having `[]` as the default here breaks the loading skeleton.

kleinfreund: Fixed in https://github.com/kumahq/kuma-gui/pull/1209/commits/d04c1e36a7a72697a79010fc619f3e6c9bb5841b

kleinfreund:
> Deleting a Zone seems to be hanging, could you take a look at that?

Good catch. I forgot to close the modal.

> Also would be good to add a test for it so it fails if we regress again. Might be worth doing the same as where we have implemented this flow elsewhere.

Done.

johncowen:
I also noticed we aren't specifying the slot for every single table cell in ZoneListView (etc).

Could you add a slot for every single table cell in the *ListView.vue file, and preferably in the order that they appear (at the very least `actions` should be at the end).

This makes it much easier for folks not familiar with the project figure out where they need to add/remove/change things and prevents people from diving into `DataOverview` type components to overload/over-complicate the single component with fields for the entire application such as:

https://github.com/kumahq/kuma-gui/blob/b8a19684d7ff1c2edcacc500ddcd7a9761ae91f6/src/app/common/DataOverview.vue#L62-L168

I'm guessing this needs doing in all Zone listings

johncowen:
https://github.com/kumahq/kuma-gui/pull/1209#issuecomment-1649316783

Thanks! (crossed messages!)

kleinfreund:
> Could you add a slot for every single table cell in the *ListView.vue file, and preferably in the order that they appear (at the very least `actions` should be at the end).

This is also done.

johncowen:
Ok looking now

johncowen:
Ah sorry, I should have mentioned in my previous comment, can we use `row` instead of `rowValue`?

Reason:

```javascript
{{ row.theNameOfTheProperty}}
```
is much much clearer than:
```javascript
{{ rowValue}}
```

You can probably guess I would prefer the `row: item` rename also, but I'll leave that one to you for the moment.

kleinfreund:
> Ah sorry, I should have mentioned in my previous comment, can we use `row` instead of `rowValue`?

It’s not any more clear, really. Not changing this _again_.

I’m not addressing anymore stylistic nitpicks on this PR.

johncowen:
Ok let me double check the functionality again then. I would guess this is fine and ready to approve.

I just wanted to point out that part of the reason for this stream of work was to make the entire application more approachable to external contributors or other members of the team. So these are not necessarily stylistic changes, they are about making it easy to see what is going on so its straightforwards to convert JSON into a list.

johncowen:
> Also would be good to add a test for it so it fails if we regress again.

I don't think we added this ^?

kleinfreund:
> I don't think we added this ^?

Yet I did.

johncowen:
Ah I see it, but not as an e2e test, 🤔 ...

johncowen (approved):
Thanks for all the work on this!

PR #1219, opened 2023-07-21
chore: Convert Dataplane/Gateway Listings to DataSource/AppCollection

johncowen:
There is still quite a bit of work to be done here, but I wanted to get far enough to convert everything to use DataSource/AppCollection, and as soon as I got there I stopped to submit a PR. There are likely to be follow up PRs to come to improve various things here, some of them pre-existing (I don't think we should always mention zones in the filter when we are in standalone more for example).

Notable significant changes:

- We now have a specific route component for the GatewayListing tab
- We no longer have 2 components to show the summary detail page (SummaryDetail has been deleted)
- The KFilterBar usage has been moved to the top i.e. its used in the route component rather than inside nested components.


src/app/common/KFilterBar.vue
```diff
@@ -149,7 +149,8 @@ const props = defineProps({
    */
   id: {
     type: String,
-    required: true,
+    required: false,
+    default: 'data-plane-proxy-filter',
```
kleinfreund, line 153: **Suggestion**: This shouldn’t be here. This component is completely agnostic to its usage context and should remain so. It’s no problem setting this on use.

src/app/common/KFilterBar.vue
```diff
@@ -443,6 +444,7 @@ function areFieldsSemanticallyIdentical(fieldsA: Fields, fieldB: Fields): boolea
   flex-grow: 1;
   width: 100%;
   border: none;
+  padding: 12px;
```
kleinfreund, line 447: **Question**: Why was this added?

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -102,7 +102,7 @@ const click = (e: MouseEvent) => {
 }
 </script>
 <style type="scss" scoped>
-.app-collection :deep(td:first-of-type > a) {
+.app-collection :is(td) > :is(a) {
```
kleinfreund, line 105: **Question**: Why was this changed back? The styles work without this change.

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -1,114 +1,141 @@
...
 
-          <KDropdownMenu
-            label="Columns"
-            icon="cogwheel"
-            button-appearance="outline"
+      <template v-else>
+        —
+      </template>
+    </template>
+    <template #actions="{ row: item }">
```
kleinfreund, line 83: **Issue**: The actions are missing the columns dropdown.

johncowen (commented):
I left an inline note

johncowen:
GH is having issues at the minute so just waiting for that to clear up

src/app/common/KFilterBar.vue
kleinfreund: This should be handled where KFilter is used to avoid the component to _always_ have an arbitrary padding.

johncowen: I _always_ want it to have this padding, so I want to add it here.

johncowen: We only ever use this in a KTable toolbar, all things in the toolbar should have the same height so they all line up nicely. We also use KSelect in the toolbar. Giving KFilterBar this padding makes it the same height as the KSelect.

i.e. I _always_ want this to be this height as this is only ever used in the toolbar alongside a KSelect

johncowen: Again, my thoughts exactly. Is there anything else outstanding here?

src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -1,157 +1,89 @@
...
-.table-header-selector-item .k-dropdown-item-trigger {
-  // Removes the paddings from the dropdown items so that the items are interactive for their full visual size.
-  padding: 0 !important;
+<style lang="scss" scoped>
+.data-plane-proxy-filter {
+  flex-basis: 350px;
+  flex-grow: 1;
+  margin-right: auto;
 }
 </style>
```
kleinfreund, line 89: **Suggestion**: Can you make sure to add the same kinds of styles I added in the already converted list views for column sizing?

- Would need the following styles for three tables
- Would need the class name for the three tables

```suggestion
.actions-dropdown {
  display: inline-block;
}
</style>

<style lang="scss">
.data-plane-proxy-collection {
  .actions-column {
    width: 5%;
    min-width: 80px;
    text-align: end;
  }

  .status-column {
    width: 10%;
    min-width: 200px;
  }
}
</style>
```

src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -1,157 +1,89 @@
...
+            <RouteTitle
+              :title="t('data-planes.routes.items.title')"
+              :render="true"
+            />
+          </h2>
+        </template>
+        <KCard>
+          <template #body>
+            <DataPlaneList
+              data-testid="data-overview-table"
```
kleinfreund, line 21: **Suggestion**: So the column style suggestion works and to be consistent with the other list views.

```suggestion
              class="data-plane-proxy-collection"
              data-testid="data-plane-proxy-collection"
```

src/app/data-planes/views/DataPlaneListView.vue
johncowen: Oh yes good call absolutely! Will look now

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -1,114 +1,141 @@
...
+      { label: 'Last Updated', key: 'lastUpdated' },
+      { label: 'Kuma DP version', key: 'dpVersion' },
+      { label: 'Actions', key: 'actions', hideLabel: true },
+    ].filter(Boolean)"
+    :page-number="props.pageNumber"
+    :page-size="props.pageSize"
+    :total="props.total"
+    :items="props.items ? transformToTableData(props.items) : []"
+    :error="props.error"
+    @change="(val: ChangeValue) => emit('change', val)"
```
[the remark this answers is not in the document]
johncowen: Hmm let me look at this, I would also like to avoid needing the type parameter here seeing as we don't access the data at all. Little cautious about the weird global $event thing though (thats what it is is it?). Let me take a look at this one though.

johncowen:
Ok thanks @kleinfreund I've addressed your last round of feedback, I think I got everything lmk if not

kleinfreund (approved):

johncowen:
Awesome thanks for the reviews on this one. I was cautious of making sure I covered the existing functionality of the things we are keeping, and it took me a good while to simplify it all out. Just to mention as we discussed in yesterdays sync, there will be PRs coming to turn this into 3 separate views:

1. Dataplane Proxy Listing (only columns for proxies)
2. Gateway Listing (only columns for gateways)
3. Dataplanes for a Service listing (columns for both proxies and gateways)

From what I remember from working on this, the main reason we have a separate `DataplaneList` is to have the data re-shaping in one place, we then use awkward conditionals inside the component which are going to break down now we need 3 different setups. For that reason, once we have moved the data reshaping to the data layer so it can be reused anywhere, I would guess I will just delete the `DataplaneList` component and just use 3 separate `AppCollection` components configured with the correct columns for each view. I will be able to do this as the data will already have been reshaped for me for all 3 views in the data layer. I would guess the "move data reshaping to the data layer" PR will come first.

PR #1221, opened 2023-07-24
chore: Make router injectable and pullable baseURL 

johncowen:
This PR does 2 things:

---

### Injectable router

Back in https://github.com/kumahq/kuma-gui/pull/765 I added an injectable `$.router` specifically for our CLI tests, but I'd always planned to then use that for our main application see (https://github.com/kumahq/kuma-gui/pull/765/files#r1151912562).

I also corrected it here to use our base path, hence the snapshot update.

We now feed the `$.router` into `useApp` just like the other Vue plugins, eventually we'll be able to add these sorts of plugins with some DI labelling.

---

### BaseUrl changes

I made baseUrl just pull straight from `env('KUMA_API_URL')` meaning we don't have to keep setting it in certain places, meaning we can remove a bunch of code elsewhere.

### DIFF
--- a/src/app/__snapshots__/AppOnboardingNotification.spec.ts.snap
+++ b/src/app/__snapshots__/AppOnboardingNotification.spec.ts.snap
@@ -30,7 +30,7 @@ exports[`AppOnboardingNotification renders snapshot 1`] = `
           <div>
             <a
               class="k-button small rounded primary action-button"
-              href="/onboarding/welcome"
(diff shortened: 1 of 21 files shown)


kleinfreund:
This needs a rebase.

johncowen:
Oh yes it does thanks! Rebased. Seeing as it was a one line conflict, I take it there are no further comments on the change set then?

kleinfreund (approved):

PR #1222, opened 2023-07-24
feat: stores response on ApiError

kleinfreund:
Stores the full fetch `Response` object on `ApiError` instances. This allows us to reacting to more than just the response status (e.g. for implementing response header-based retrying logic).

### DIFF
--- a/src/app/common/ErrorBlock.vue
+++ b/src/app/common/ErrorBlock.vue
@@ -60,7 +60,7 @@
       </KBadge>
 
       <KBadge :appearance="props.badgeAppearance">
-        {{ error.status }}
+        {{ error.response.status }}
       </KBadge>
     </div>
   </div>
--- a/src/app/zones/views/ZoneCreateView.vue
+++ b/src/app/zones/views/ZoneCreateView.vue
@@ -277,13 +277,13 @@ async function createZone() {
   try {
     zone.value = await kumaApi.createZone({ name: name.value })
   } catch (err) {
-    if (err instanceof ApiError && HANDLED_STATUS_CODES.includes(err.status)) {
+    if (err instanceof ApiError && HANDLED_STATUS_CODES.includes(err.response.status)) {
       errorState.value = {
         error: err,
-        title: t(`zones.create.statusError.${err.status}.title`, { zoneName: name.value }),
-        description: t(`zones.create.statusError.${err.status}.description`).trim(),
-        icon: err.status === 500 ? 'warning' : 'errorFilled',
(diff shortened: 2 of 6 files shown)


johncowen (changes requested):
> Stores the full fetch Response object on ApiError instances. This allows us to reacting to more than just the response status (e.g. for implementing response header-based retrying logic).

What else do we need apart from the http.status to react to 5xx errors?

I would really rather close this for the moment and revisit in the not too distant future when we want to build this out rather than change a bunch of code now that might not even need changing.

kleinfreund:
> What else do we need apart from the http.status to react to 5xx errors?

`response.headers`

johncowen:
Could you explain to me why we need the `response.headers`?

kleinfreund:
> Could you explain to me why we need the `response.headers`?

We have tentative plans to allow failing requests to be retried based on a response header (e.g. "retry-after").

johncowen:
Ok so we don't need to yet? But there is maybe a plan to do this?

kleinfreund:
I do think it makes general sense to store the response on the error object like done in this PR because you’d otherwise never have access to its data.

johncowen:
I wanted to add a bit more detail here.

Most "pseudo-standard" HTTPErrors include some sort of `status` property. So at the very least I don't think we should remove `status`.

I am also cautious to put something like a full response object on an Error, seeing as errors can then be exposed/bubbled up to almost every area of the application. Judging by the potential use case you have explained, we could just use a `.retryAfter`  property, or even a `headers` property which is just a `Record<string, string>` which would be exposing less throughout the application.

Whether we think we should use `.retryAfter`, `.headers` (or possibly `.meta`), or `.response` can be decided if/when we actually come to implement the tentative feature you have explained.

In the meantime, I think `status` is just fine.

Lmk your thoughts.

kleinfreund:
Alright, I’m fine with leaving the status property even though it’s completely redundant if we expose the `response`.

Also, I don’t see a reason not to expose the full response. Where’s the issue?

Anyway, would you be fine with exposing `response.headers` as `error.headers` as a compromise?

But still I’m not sure why this is contentious.

johncowen:
Ok, I think the main issue is as far as I understand it **we don't have a need for this** as yet. And if/when we do, I'd like to make sure we think about it properly. For example, at some point we may also want to access successful response headers also.

> Alright, I’m fine with leaving the status property even though it’s completely redundant if we expose the response.

>> Most "pseudo-standard" HTTPErrors include some sort of status property. So at the very least I don't think we should remove status.

This and my previous comment, and the fact that we don't need this right now, are all the issues together.

PR #1225, opened 2023-07-24
fix: table column styles leaking to other tables

kleinfreund:
**chore(meshes): address linter warning**

**fix(services): table column styles leaking to other tables**

**fix(meshes): table column styles leaking to other tables**

**fix(policies): table column styles leaking to other tables**

### DIFF
--- a/features/mesh/Index.feature
+++ b/features/mesh/Index.feature
@@ -2,7 +2,7 @@ Feature: mesh / index
   Background:
     Given the CSS selectors
       | Alias          | Selector                                     |
-      | items          | [data-testid='mesh-collection']              |
+      | items          | [data-testid='mesh-table']                   |
       | item           | $items tbody tr                              |
       | breadcrumb     | .k-breadcrumbs                               |
       | button-refresh | [data-testid='data-overview-refresh-button'] |
@@ -38,7 +38,6 @@ Feature: mesh / index
     Then the "$item" element exists 2 times
 
     Examples:
-      | Mesh         | Selector |
+      | Mesh         | Selector                              |
       | another-mesh | $item:nth-child(2) td:first-of-type a |
       | default      | $item:nth-child(1) td:first-of-type a |
-
--- a/features/mesh/Index.feature
+++ b/features/mesh/Index.feature
@@ -2,7 +2,7 @@ Feature: mesh / index
   Background:
     Given the CSS selectors
       | Alias          | Selector                                     |
-      | items          | [data-testid='mesh-table']                   |
+      | items          | [data-testid='mesh-collection']              |
       | item           | $items tbody tr                              |
       | breadcrumb     | .k-breadcrumbs                               |
       | button-refresh | [data-testid='data-overview-refresh-button'] |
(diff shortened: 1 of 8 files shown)


features/mesh/Index.feature
```diff
@@ -2,7 +2,7 @@ Feature: mesh / index
   Background:
     Given the CSS selectors
       | Alias          | Selector                                     |
-      | items          | [data-testid='mesh-collection']              |
```
johncowen, line 5: Please could you leave this as collection

features/mesh/services/Index.feature
```diff
@@ -2,7 +2,7 @@ Feature: mesh / services / index
   Background:
     Given the CSS selectors
       | Alias               | Selector                            |
-      | items               | [data-testid='service-collection']  |
```
johncowen, line 5: Please could you leave this as collection.

src/app/meshes/views/MeshListView.vue
```diff
@@ -29,7 +29,8 @@
           <KCard>
             <template #body>
               <AppCollection
-                data-testid="mesh-collection"
```
johncowen, line 32: See above

src/app/meshes/views/MeshListView.vue
```diff
@@ -29,7 +29,8 @@
...
+                class="mesh-table"
+                data-testid="mesh-table"
```
johncowen, line 33: See above

src/app/services/views/ServiceListView.vue
```diff
@@ -19,7 +19,8 @@
         <KCard>
           <template #body>
             <AppCollection
-              data-testid="service-collection"
+              class="service-table"
+              data-testid="service-table"
```
johncowen, line 23: See above

johncowen (changes requested):
Thanks for doing this! Couple of naming things that there is no need to change, in fact I would rather you didn't.

features/mesh/Index.feature
kleinfreund: We already had *-table selectors in two other places so I stuck with that to not further have different patterns for this.

johncowen: Everything has moved to use `AppCollection` please could you use the term 'collection' for multiple things. You've actually changed them back to what we don't want here. Please could you use `collection` terminology, or at the very least leave where we are using `collection` terminology as it is.

src/app/meshes/views/MeshListView.vue
```diff
@@ -29,7 +29,8 @@
           <KCard>
             <template #body>
               <AppCollection
-                data-testid="mesh-collection"
+                class="mesh-collections"
```
johncowen, line 32: Singular please.

features/mesh/Index.feature
kleinfreund: > to what we don't want here.

You. I was perfectly happy with calling those tables tables.

I changed those things now.

johncowen (approved):

PR #1233, opened 2023-07-25
chore: regenerates snapshots

kleinfreund:


johncowen (approved):
At this point I think we should just remove all the snapshotting entirely (as a new task of course), I'll make an issue.

kleinfreund:
Note that we’re currently seeing more snapshot changes on account of ongoing work in Kongponents that replaces/removes a lot of utility classes in favor of regular styles.

PR #1235, opened 2023-07-25
chore: Expose some commonly used utilities/components

johncowen:
Exposes `t` and `env` from RouteView as we pretty much always need these in Route components and we want to default to 'generally' not using them anywhere but Route components (we will need to occasionally use them in non-route components, so want to stress the 'generally')

Also installs Kongponents as a Vue plugin.

### DIFF
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -12,6 +12,8 @@
     </div>
     <slot
       name="default"
+      :t="t"
+      :env="env"
       :route="{
         update: (params: Record<string, string | undefined>) => {
           router.push(
@@ -45,8 +47,9 @@ import {
   createAttrsSetter,
   createTitleSetter,
 } from '../../utilities'
-import { useI18n } from '@/utilities'
+import { useI18n, useEnv } from '@/utilities'
 
+const env = useEnv()
 const { t } = useI18n()
 const route = useRoute()
 const router = useRouter()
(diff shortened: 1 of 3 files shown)


src/app/application/components/route-view/RouteView.vue
```diff
@@ -12,6 +12,8 @@
     </div>
     <slot
       name="default"
+      :t="t"
```
kleinfreund, line 15: **Suggestion**: We should really expose the full `i18n` object here. Then we can freely decide what to use and don’t need to amend this every time we find out we now need, for example, access to `i18n.formatIsoDate` or one of the other translation functions.

johncowen: Yeah I did consider this, but figured we use this far far less. Also, I actually wondered why we don't integrate with FormatJS/ICU standard date formats https://formatjs.io/docs/core-concepts/icu-syntax#date-type for all our date formatting.

johncowen: After asking around and looking at some things, I'm fairly sure we can install `i18n.formatIsoDate` as our default datetime formatter within ICU (and maybe remove the others) this would mean that using:

```vuw
`t('a.key.somewhere', {value: date})
```

and:

```yaml
a.key.somewhere: "{value, date }"
```

would use `i18n.formatIsoDate`, but also

```yaml
a.key.somewhere: "Hey there todays date is {value, date}"
```

Would also use `i18n.formatIsoDate` i.e. we pretty much make it impossible to use a different format.

All in all, that would mean we don't need to expose this JS function anywhere. If I'm wrong about the above we can just import it as we usually would in the few places we use it (or even add your suggestion of exporting the entire `i18n` here.

For the moment tho, I'd rather just keep `t`

kleinfreund (approved):

PR #1237, opened 2023-07-25
chore: moves status column towards the end

kleinfreund:
Moves the status column in list views (if present) towards the end so it is the last column, but before warnings and actions columns.


johncowen (approved):
ty!

PR #1238, opened 2023-07-25
chore: Add configurable exponential error backoff to all read requests

johncowen:
Closes all sources after the response has been received, also adds a configurable exponential error 50x backoff to all read HTTP requests.

### DIFF
--- a/features/mesh/policies/Data.feature
+++ b/features/mesh/policies/Data.feature
@@ -45,7 +45,7 @@ Feature: mesh / policies / data
     Given the URL "/meshes/default/circuit-breakers" responds with
       """
       headers:
-        Status-Code: '503'
+        Status-Code: '520'
       """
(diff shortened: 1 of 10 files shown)


lahabana:
Do we need to get that in now? Can we focus on the immediate effort? (There's a lot of reviews outstanding so trying to prioritise things :) )

johncowen:
Nope not at all, I had it hanging around and I saw that we might be changing how APIError works and figured it would be better to PR this sooner rather than later. Happy to close and keep hold of the branch, or just leave hanging until someone gets a sec to review the code. lmk

Oh actually, moving `source.close` to be after the response is required. If you don't think we should add the backoff here, I can remove that from here and keep a branch with it in.

lahabana:
No it's fine I just want to make sure @kleinfreund knows what needs to be reviewed in priority before you go

PR #1241, opened 2023-07-25
chore: Adds AppCollectionActions

johncowen:
This is to avoid too much repeated setup of KDropdownMenu specifically for all our many AppCollections. I've tested this locally but wanted to get a draft of this up before rolling out everywhere incase there are any review comments.

I've included an example in the AppCollection docs to show the reduction.

### DIFF
--- a/src/app/application/components/app-collection/AppCollectionAction.vue
+++ b/src/app/application/components/app-collection/AppCollectionAction.vue
@@ -1,8 +1,37 @@
 <template>
-  <KDropdownItem />
+  <KDropdownItem
+    v-bind="props"
+  />
 </template>
 <script lang="ts" setup>
 import {
   KDropdownItem,
 } from '@kong/kongponents'
+
+import type { DropdownItem } from '@kong/kongponents'
+import type { PropType } from 'vue'
+
+const props = defineProps({
+  item: {
+    type: Object as PropType<DropdownItem>,
+    default: null,
+    validator: (item: DropdownItem) => item.label !== undefined,
+  },
+  hasDivider: {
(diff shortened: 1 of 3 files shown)


src/app/application/components/app-collection/AppCollectionActions.vue
```diff
@@ -0,0 +1,33 @@
+<template>
```
lahabana, line 1: Any reason not to inline this inside AppCollection?

johncowen: I suppose because I figure that sometimes an AppCollection might not have any actions at all.

i.e. I don't want to overload AppCollections too much. What if we have an AppCollection that doesn't have a dropdown, but instead has a single button? Or the actions has a dropdown and a checkbox, or possibly other things I haven't thought of?

lahabana:
Btw this would replace things already done in the new lists right? So I think this is also lower priority than some of the other changes no?

src/app/application/components/app-collection/AppCollectionAction.vue
```diff
@@ -0,0 +1,8 @@
+<template>
+  <KDropdownItem />
```
kleinfreund, line 2: **Issue**: For this to work well with the tooling around Vue and type checking of props, props need to be passed explicitly. This is in fact already a problem with `AppCollection` where the props we pass to it aren’t any longer type-checked against the `KTable` props (e.g. you can pass `:headers="1"` and nothing will alight with errors).

kleinfreund (changes requested):

PR #1243, opened 2023-07-26
fix(zones): zone name not being quoted in YAML

kleinfreund:
Fixes the zone name not being quoted in certain YAML files leading to numerical values being treated as numbers.

### DIFF
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -77,7 +77,7 @@ zones:
           mode: zone
           multizone:
             zone:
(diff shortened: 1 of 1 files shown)


johncowen (commented):
Just out of interest is this causing any user facing issues?

kleinfreund:
> Just out of interest is this causing any user facing issues?

Yes

johncowen:
Ah, ok could you explain how I can reproduce, I'm not getting any errors here and that was surprising. Myabe its a smaller issue that I'm not aware of?

johncowen (approved):

PR #1246, opened 2023-07-26
fix: handling of empty table cells

kleinfreund:
**fix(zones): handling of empty table cells**

Fixes empty cells not rendering the right content.

Fixes the warnings cell showing `false` instead of nothing.

**fix(services): handling of empty table cells**

Randomizes presence of certain fields on service mocks and only sets dataplanes if the service is an external one.

Makes sure to always show a status badge.

### DIFF
--- a/src/app/services/views/ServiceListView.vue
+++ b/src/app/services/views/ServiceListView.vue
@@ -51,6 +51,26 @@
                 </RouterLink>
               </template>
 
+              <template #serviceType="{ rowValue }">
+                <template v-if="rowValue">
+                  {{ rowValue }}
+                </template>
+
+                <template v-else>
+                  —
+                </template>
+              </template>
+
+              <template #addressPort="{ rowValue }">
+                <template v-if="rowValue">
+                  {{ rowValue }}
+                </template>
+
+                <template v-else>
+                  —
+                </template>
+              </template>
+
               <template #online="{ row: item }">
                 <template
                   v-if="item.dataplanes"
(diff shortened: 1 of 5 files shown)


src/app/services/views/ServiceListView.vue
```diff
@@ -65,14 +85,7 @@
...
               <template #status="{ row: item }">
-                <StatusBadge
-                  v-if="item.status"
-                  :status="item.status"
-                />
-
-                <template v-else>
-                  —
-                </template>
+                <StatusBadge :status="item.status || 'not_available'" />
```
lahabana, line 88: Would it make sense to have this done in your `transfromToTableData` ?

kleinfreund: Technically, yes, but it looks very much like this (these transform call) is all going to be thrown out in favor of a different approach.

lahabana: Yes but the transform call will likely be the base for the code of the new approach no?

src/app/zones/views/ZoneListView.vue
```diff
@@ -72,11 +72,23 @@
...
 
                 <template #zoneCpVersion="{ rowValue }">
-                  {{ rowValue }}
+                  <template v-if="rowValue">
+                    {{ rowValue }}
+                  </template>
+
+                  <template v-else>
+                    —
+                  </template>
```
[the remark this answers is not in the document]
kleinfreund: That works, too. I’ve changed the instance in the two list views I’m changing in this PR.

lahabana (approved):

PR #1247, opened 2023-07-26
chore: Remove dead code

johncowen:
We've rebuilt quite a lot of the application lately, and therefore there is quite a bit of dead code to delete.

This PR deletes some obvious "we don't need these now" files. There might be some other things we don't need still but we can iteratively delete those bits over time.


kleinfreund (approved):

lahabana:
🥳

PR #1248, opened 2023-07-27
chore(e2e): adds mockMergeStyle option

kleinfreund:
**refactor(e2e): allows mock options to be expanded beyond env**

Changes the `Options` type and its use to allow for additional properties beyond just env.

**chore(e2e): adds mockMergeStyle option**

Adds a new mock option `mockMergeStyle` that controls the merging behavior for mocks. It adds support for a "replace" option opting out of merging mocks entirely so that tests can specify a full mock without interference of the default mocks.

### DIFF
--- a/cypress/support/step_definitions/index.ts
+++ b/cypress/support/step_definitions/index.ts
@@ -56,9 +56,10 @@ Given('the environment', (yaml: string) => {
     cy.setCookie(key, String(value))
   })
 })
-Given('the URL {string} responds with', (url: string, yaml: string) => {
+Given(/^the URL "(.*)" responds with( exactly)?$/, (url: string, exactly: string, yaml: string) => {
   const options: Options = {
     env,
+    mockMergeStyle: exactly ? 'replace' : 'deep',
   }
   const now = new Date().getTime()
   const mock = useMock()
--- a/cypress/support/step_definitions/index.ts
+++ b/cypress/support/step_definitions/index.ts
@@ -2,6 +2,7 @@ import { When, Then, Before, Given, DataTable } from '@badeball/cypress-cucumber
 import YAML from 'js-yaml'
 
 import { useServer, useMock } from '@/services/e2e'
+import type { Options } from '@/test-support'
 
 const console = {
   log: (message: unknown) => Cypress.log({ displayName: 'LOG', message: JSON.stringify(message) }),
@@ -56,10 +57,13 @@ Given('the environment', (yaml: string) => {
   })
 })
 Given('the URL {string} responds with', (url: string, yaml: string) => {
+  const options: Options = {
+    env,
+  }
   const now = new Date().getTime()
   const mock = useMock()
   urls.set(url, `spy-${now}`)
-  mock(url, env, (respond) => {
+  mock(url, options, (respond) => {
     const response = respond((YAML.load(yaml) || {}) as { headers?: Record<string, string>, body?: Record<string, unknown> })
     return response
   }).as(urls.get(url))
--- a/src/app/onboarding/views/DataplanesOverview.spec.ts
+++ b/src/app/onboarding/views/DataplanesOverview.spec.ts
@@ -11,7 +11,7 @@ function renderComponent() {
 describe('DataplanesOverview.vue', () => {
   const mock = useMock()
   test('renders snapshot', async () => {
-    mock('/dataplanes', { KUMA_DATAPLANE_COUNT: '1' }, (merge) => {
+    mock('/dataplanes', { env: { KUMA_DATAPLANE_COUNT: '1' } }, (merge) => {
       return merge({
         body: {
           items: [
@@ -22,7 +22,7 @@ describe('DataplanesOverview.vue', () => {
         },
       })
     })
-    mock('/meshes/:mesh/dataplanes+insights/:name', { KUMA_SUBSCRIPTION_COUNT: '1' }, (merge) => {
+    mock('/meshes/:mesh/dataplanes+insights/:name', { env: { KUMA_SUBSCRIPTION_COUNT: '1' } }, (merge) => {
       return merge({
         body: {
           name: 'dataplane-test-456',
--- a/src/app/zones/views/ZoneListView.spec.ts
+++ b/src/app/zones/views/ZoneListView.spec.ts
@@ -23,7 +23,9 @@ describe('ZoneListView', () => {
   test('updates list when deleting Zone', async () => {
     const firstZoneName = 'zone-1'
     mock('/zones+insights', {
-      KUMA_ZONE_COUNT: '3',
+      env: {
+        KUMA_ZONE_COUNT: '3',
+      },
     }, (merge) => {
       return merge({
         body: {
--- a/src/services/e2e.ts
+++ b/src/services/e2e.ts
@@ -12,7 +12,7 @@ const env = (
 type AEnv = ReturnType<typeof env>
 type Server = typeof cy
 // temporary intercept returning Mocker
-type Mocker = (route: string, opts?: Options, cb?: Callback) => ReturnType<typeof cy['intercept']>
+type Mocker = (route: string, opts?: Partial<Options>, cb?: Callback) => ReturnType<typeof cy['intercept']>
 const $ = {
   EnvVars: token<EnvVars>('EnvVars'),
   env: token<AEnv>('env'),
--- a/src/test-support/index.ts
+++ b/src/test-support/index.ts
@@ -10,8 +10,16 @@ export type { FS, EndpointDependencies, MockResponder } from './fake'
 
 export type Merge = (obj: Partial<MockResponse>) => MockResponse
 export type Callback = (merge: Merge, req: RestRequest, response: MockResponse) => MockResponse
-export type Options = Record<string, string>
+export type Options = {
+  env: Record<string, string>
+}
 type Server = ReturnType<typeof setupServer>
+
+export function getOptions(options: Partial<Options>): Options {
+  return {
+    env: options.env ?? {},
+  }
+}
 
 // merges objects in array positions rather than replacing
 const combineMerge = (target: object[], source: object[], options: ArrayMergeOptions): object[] => {
@@ -33,10 +41,12 @@ const noop: Callback = (_merge, _req, response) => response
 export const createMerge = (response: MockResponse): Merge => (obj) => deepmerge(response, obj, { arrayMerge: combineMerge })
 
 const useResponder = <T extends RestRequest>(fs: FS, env: AEnv) => {
-  return (route: string, opts: Options = {}, cb: Callback = noop) => {
-    const mockEnv: Env = (key, d = '') => (opts[key as MockEnvKeys] ?? '') || env(key as AppEnvKeys, d)
+  return (route: string, opts: Partial<Options> = {}, cb: Callback = noop) => {
+    const options = getOptions(opts)
+
+    const mockEnv: Env = (key, d = '') => (options.env[key as MockEnvKeys] ?? '') || env(key as AppEnvKeys, d)
     if (route !== '*') {
-      dependencies.fake.seed(typeof opts.FAKE_SEED !== 'undefined' ? parseInt(typeof opts.FAKE_SEED) : 1)
+      dependencies.fake.seed(typeof options.env.FAKE_SEED !== 'undefined' ? parseInt(typeof options.env.FAKE_SEED) : 1)
     }
     const endpoint = fs[route]
     const fetch = endpoint({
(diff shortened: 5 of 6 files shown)


johncowen:
I think we've had discussion elsewhere on this. As far as I understand it right now, this PR was because:

1. The KUMA_ env vars weren't being used to set the amount of items coming back from the API during tests.
2. There is the odd occasion where ideally we need to omit/undefine a property.

In the code where point 1 was an issue, I think thats been amended to use KUMA_ env vars, and I've PRed https://github.com/kumahq/kuma-gui/pull/1452 which addresses point 2.

If there are outstanding points that I'm not aware of please let me know, I'm happy to understand potential shortcomings of the current approach. Otherwise as discussed elsewhere I'm still unsure of the need for this and I don't want to encourage folks adding entire API responses to tests, we should only be arranging/in-test mocking the data that we test on.

kleinfreund:
I just want to be able to use regular standard mocking and not care about other source of data (like our default mocks) influencing a test. That’s a perfectly reasonable expectation. If we were writing plain Cypress tests, this would simply be possible because I could call `cy.intercept` as I see fit.

A lot of test writing has become more difficult because I constantly need to keep track of which data I’m expecting despite the mock I provide in a test. On top of that, I manually need to control the amount of elements in arrays which in a lot of cases would be much more straightforward if I could provide the exact mock.

> I don't want to encourage folks adding entire API responses to tests, we should only be arranging/in-test mocking the data that we test on.

There’s no good reason to flat out disallow this. Sometimes, mocking the entire and exact API response _is_ the best approach because it guarantees that no change in the default mocks can break your test.

johncowen:
> I manually need to control the amount of elements in arrays

Aren't you suggesting here that we _manually_ provide the entire exact API response, instead of just saying "just give me 5 of these things with any old correct data"

> guarantees that no change in the default mocks can break your test.

If you tests break because of the default mocks, that means the test isn't mocking things out that it is testing for, or there is something that hasn't catered for in the application code i.e. it gives the chance to understand what the problem is, and either alter the application code to cope with that or stabilise the piece of API data in your test by mocking that piece of data in your test i.e. the randomness is there for a reason.

All in all we shouldn't need to mock out the entire API response. As I explained above, if we need to mock out the entire API response then we need to change the test framework so you don't have to (as was the case with `undefined`).

> There’s no good reason to flat out disallow this.

The thing I'm trying to explain is, I'm yet to see a good reason to flat out allow this, and I and others have given several reasons why we don't want to allow this. If I can see an example of where we absolutely have to add entire API responses to tests I'd happily reconsider.

I think to prevent this from going on for a long time, I think if you want to continue with this then it would be good if the next step here would be to provide an example.

If you can give me an example of something that you absolutely cannot mock without providing the entire API response, **for a test that is providing a valuable e2e test for the application** then I'm happy to discuss that and then perhaps consider this PR. As far as I'm aware for the two examples given, one was questionable/user-omission and there is a PR up for the other that addresses the problem in the correct place (adding the ability to undefine properties), which I think is a very good improvement and I'm grateful you pointed that one out.

kleinfreund:
The examples in context of which I’ve argued for this before are all still examples where I vastly prefer mocking the entire API response instead of providing `KUMA_*_COUNT` config and a partial mock.

Just for the sake of naming them again: Subscriptions inside DataplaneOverview, subscriptions inside ZoneInsight, and generally instances of array inside collection resources.

johncowen:
> I vastly prefer mocking

Ok so from that, it sounds like its a preference, not an inability?

It's been said elsewhere I think, but we should encourage using these KUMA_*_COUNT variables and adding them should be part of writing mocks/tests as these then help setting up scenarios when previewing the application.

I think its also worth noting that, once we have any data reshaping code isolated in a data layer outside of Vue components, any data reshaping code would probably be tested using unit tests, not e2e.

kleinfreund:
> Ok so from that, it sounds like its a preference, not an inability?

Without this change, I **cannot** mock data the way I want to so, no, it’s not just a preference.

I’m fine with the `KUMA_*_COUNT` mechanism _existing_ but I’m _not_ fine with it being the only mechanism to control mock data. It’s unnecessarily limiting. At this point I would much rather write plain Cypress tests because I wouldn’t need to worry which features of it I can access.

johncowen:
> I think to prevent this from going on for a long time, I think if you want to continue with this then it would be good if the next step here would be to provide an example.

kleinfreund:
I have:

> Just for the sake of naming them again: Subscriptions inside DataplaneOverview, subscriptions inside ZoneInsight, and generally instances of arrays inside collection resources.

johncowen:
I mean an example test

kleinfreund:
Any test that uses subscriptions is a good example.

johncowen:
> I think to prevent this from going on for a long time, I think if you want to continue with this then it would be good if the next step here would be to **provide an example**. (edit: an example test)

kleinfreund:
You can search the project for `subscriptions:` to find the following tests:

- features/main-overview/DetailViewContent.feature
- features/mesh/dataplanes/DetailViewContent.feature
- features/mesh/dataplanes/Index.feature
- features/zones/Create.feature
- features/zones/DetailViewContent.feature
- features/zones/ListViewContent.feature
- features/zones/zone-cps/Item.feature

Alternatively, you can search for `KUMA_SUBSCRIPTION_COUNT`.

For a lot of these tests I want to use exact mocking for reasons I explained a couple of times now.

johncowen:
Ok so I just looked at the first one in that llist:

> features/main-overview/DetailViewContent.feature

Maybe I don't follow? But as far as I can see we were able to write these tests without needing this PR? Could you explain why this PR is needed for `features/main-overview/DetailViewContent.feature` to work?

kleinfreund:
But only by manually mutilating the response using _COUNT variables. It’s very cumbersome and intransparent to do that when I could use standard Cypress mocking instead.

johncowen:
> But only by manually mutilating the response using _COUNT variables. It’s very cumbersome and intransparent to do that when I could use standard Cypress mocking instead.

Ok so from that I understand that these aren't examples where you cannot mock out your test using the existing approach then, its just that you personally don't want to use/add _COUNT variables.

Let me know if you can give an example of a test where you cannot mock out your test using the existing approach.

johncowen:
P.S. I also noticed that in `features/main-overview/DetailViewContent.feature` you aren't even "manually mutilating the response using _COUNT variables", not for subscriptions anyway.

kleinfreund:
There are also cases where using _COUNT variables is just simply not good enough. They control the amount of elements in a list, but the merging behavior of mock data also breaks the data _inside_ of lists.

johncowen:
> There are also cases where using _COUNT variables is just simply not good enough. They control the amount of elements in a list, but the merging behavior of mock data also breaks the data inside of lists.

Thats fine, but could you give me an example test of where you can't write a valuable e2e test due to our current approach to mocking so I can see.

> I think to prevent this from going on for a long time, I think if you want to continue with this then it would be good if the next step here would be to provide an example.

kleinfreund:
These two are very good examples:

- features/zones/Create.feature
- features/mesh/dataplanes/DetailViewContent.feature

In both cases we set `disconnectTime` to a value that _happens_ to prevent the flakiness that’s caused by the random (i.e. faker.js random) appearance of mock data to hack around the mock merging behavior. This is a poor approach because the values we use for this can’t actually appear on the mocked API. A complete mock is a much better solution to such a scenario.

johncowen:
I think I mentioned this PR further up https://github.com/kumahq/kuma-gui/pull/1452, which if I understand correctly addresses this by letting us specify properties as undefined/omitted

kleinfreund:
> I think I mentioned this PR further up #1452, which if I understand correctly addresses this by letting us specify properties as undefined/omitted

Providing more ways to express mocks in that way are piece meal workarounds to a problem that doesn’t exist with the default mocking behavior. We don’t _need_ extra workarounds or solutions because the solution already exists and it’s the default mocking behavior.

johncowen:
Surely it's better to specifically mock the property out as "This property should be `undefined` for the test"? i.e. We are explicitly testing what happens when this property is undefined, rather than adding the entire API response to the test and it not being super clear what bits of the response are necessary (or not) to assert what we are testing.

kleinfreund:
> Surely it's better to specifically mock the property out as "This property should be `undefined` for the test"? i.e. We are explicitly testing what happens when this property is undefined, rather than adding the entire API response to the test and it not being super clear what bits of the response are necessary (or not) to assert what we are testing.

I disagree with the assessment that that’s simpler. Not even close. It’s so much more intransparent and confusion to track all the logic attached to our default mocks, the merging algorithm and the special rules you’re proposing in https://github.com/kumahq/kuma-gui/pull/1452.

It doesn’t need to be that complicated because the basic mocking already handles all of this.

lahabana:
I'm not sure I follow the resistance to using the existing thing when every issue @kleinfreund points out has a working solution as demonstrated by @johncowen.

Also with full request/response API mocks. What happens when you change the API significantly? Don't you end up needing to update a lot of tests?

These are also a few things I see as a benefit to not have the full response:

1. You only mock what you need to mock (And that's a best practice of mocking in general)
2. It's clear in the test what's api case being tested (with a full response it's not very clear what part of the response actually needs to be different).

As we don't have any obvious blocker with the existing stuff. Can we only do this once we realise the existing thing doesn't work?

kleinfreund:
> I'm not sure I follow the resistance to using the existing thing when every issue @kleinfreund points out has a working solution as demonstrated by @johncowen.

There are working workarounds for _some_ of the things I’m pointing out.

We _can_ use `*_COUNT` variables to limit the number of elements in an array.

We _can_ set `null` or empty string values on properties that we want to be absent in a mocked response so they don’t get accidentally merged with a non-null/non-empty string value (but this just happens to work because our application code is graceful enough to accept it; it can just as easily break and is really rather a hack).

We _cannot_ model omitted values because there is always a risk that changed default mocks suddenly introduce fields that you want to omit.

We _cannot_ model empty arrays by just setting `array: []` in a test because the array will potentially be merged with default mock data. Even _if_ you introduce a fitting `*_COUNT` variable, the previous point will prevent you from making sure that its elements will appear as you mocked them so you would again need to resort to specifying explicit `null` or empty string values that can’t actually appear on the API like that.

While those things have workarounds, they’re not great when you want to control and understand a test's data because they make that process very intransparent and convoluted. Being able to provide a full response completely removes the complexity from such scenarios. It’s simpler and predictable. It does have the downside that you would potentially need to update more mock data when you make bigger API changes, but we’re very conservative with breaking API changes anyway so that should be an exceptional case.

From an idealistic point of view I understand it’s nice to only mock data that you require for your test scenario. Practically, it’s really tedious if we achieve that by flat-out disallowing access to the standard mocking utility of our testing framework and instead requiring all these workarounds and mechanisms to control test data instead.

> 2\. It's clear in the test what's api case being tested (with a full response it's not very clear what part of the response actually needs to be different).

That’s true, but in my opinion this is a much smaller problem than not being able to know how a mocked response will actually look like by only looking at a test. You currently need to look at the mock data in the test, the `*_COUNT` variables, the environment block, the default mock data, and the merging algorithm to understand what the actual result of a mock API call will be.

> As we don't have any obvious blocker with the existing stuff. Can we only do this once we realise the existing thing doesn't work?

I had several occasions in the past and in my current tasks where I was not able to write out the test scenario I had in mind because I didn’t have access to the default mocking functionality.

I see some value in the current approach, but it has made modelling some test scenarios far more difficult than it needs to be.

johncowen:
> We can set null or empty string values on properties that we want to be absent in a mocked response so they don’t get accidentally merged with a non-null/non-empty string value (but this just happens to work because our application code is graceful enough to accept it; it can just as easily break and is really rather a hack).

> We cannot model omitted values because there is always a risk that changed default mocks suddenly introduce fields that you want to omit.

Please could you give https://github.com/kumahq/kuma-gui/pull/1452 a review?

> We cannot model empty arrays by just setting array: [] in a test because the array will potentially be merged with default mock data. Even if you introduce a fitting *_COUNT variable, the previous point will prevent you from making sure that its elements will appear as you mocked them so you would again need to resort to specifying explicit null or empty string values that can’t actually appear on the API like that.

`*_COUNT=0` is the same as `array: []`, and has the additional benefit of being able to do that whilst previewing/developing.

> I had several occasions in the past and in my current tasks where I was not able to write out the test scenario I had in mind because I didn’t have access to the default mocking functionality.

As I've asked a few times up above, could you provide these as examples so I can see? You never know I might be like "Oh I see what you are talking about now, lets consider this PR then". The issue at the minute is, despite all this back and forth, I still can't see what your issue is. If you provide an example, maybe I will?

I think for me to continue on this thread I would like to:

1. See a review of https://github.com/kumahq/kuma-gui/pull/1452
2. See an example test that is impossible to mock out using our existing approach (assuming 1452 gets merged)

Otherwise we are just going around in circles.

kleinfreund:
I’m closing this PR because it seems not possible to find consensus on the need for the default mocking functionality being available.

PR #1260, opened 2023-07-27
chore: Don't request all meshes when we don't need to

johncowen:
Completely removes the onboarding route guard and begins to move the functionality into the AppOnboardingNotification. This AppOnboardingNotification component is only loaded in certain circumstances, so it feels like a better place to make a HTTP request to find out how many meshes we have installed for the purposes of onboarding.

Note: There are some rendering issues currently with the onboarding screens that also seem to be present on `master`, these will be addressed in a separate PR.

~As a follow up here, we also make a HTTP request in order to find out how many dataplanes we have, I plan to also move this in a follow up PR.~

I added the dataplane check here also now.

Probably worth noting that as we are using DataSource for this now, if you happen to land on either the Mesh listing page or the default Dataplane proxy page, then we re-use the single request for figuring out whether to show the notification or not. Previously we would make two requests no matter what.


src/router/router.ts
```diff
@@ -32,22 +30,3 @@ const redirectOldHashHistoryUrlPaths = (): NavigationGuard => (to, _from, next)
...
- *
- * Redirects the user to the home view if they’re navigating to an onboarding route while having already completed onboarding. An exception is made when we suggest onboarding for users who don’t have data plane proxies, yet (we show an alert suggesting it and allow going to the onboarding again).
- */
-export const onboardingRouteGuard = (store: Store<State>): NavigationGuard => (to, _from, next) => {
-  const isOnboardingCompleted = store.state.onboarding.isCompleted
-  const isOnboardingRoute = to.meta.onboardingProcess
-  const shouldSuggestOnboarding = store.getters.shouldShowOnboardingNotification
-
-  if (isOnboardingCompleted && isOnboardingRoute && !shouldSuggestOnboarding) {
-    next({ name: 'home' })
```
kleinfreund, line 47: **Issue**: There is a regression here. Having completed onboarding no longer redirects to "home" when trying to access an onboarding route.

kleinfreund (changes requested):

src/router/router.ts
johncowen: Having looked at this in quite a bit of detail, I really don't think we should have this redirect, in fact as it stands before this PR its a little buggy:

1. I think the intention here was to load the onboarding the first time you hit the application. But it looks like sometimes the information needed to know whether we should show you the onboarding or not hasn't loaded yet, this means we don't then show the onboarding as you hit the application. Then unfortunately when you then click on any link on the application, the data by this time has loaded, so instead of taking you to where you wanted to go, we throw you into the onboarding screen.
2. Previous to this PR, even if I had completed onboarding, you could still visit the onboarding by clicking the 'Get started' button in the onboarding notification. The thing is, if I then happened to refresh my page whilst going through the onboarding, it would throw me back out to the home page again.

There are a few more edge cases I can think of where the previous behaviour is at least problematic, and I think this is a case of us trying to do too much, and considering that even if we did want to do this, the value it adds is very very small.

Because of this I have deleted the route guard entirely. If you now go to one of the onboarding screens again by maybe using a link I had saved previously, or clicking on a link shared by somebody else, then I won't be unexpectedly redirected to the home page.

All in all, I actually think the behaviour without the route guard is an improvement.

johncowen:
@kleinfreund this one was in draft for a while while I was on PTO, just wanted to ping re quick to say I've taken it out of Draft.

src/app/App.vue
```diff
@@ -21,7 +21,7 @@
             data-testid="app-error"
           />
 
-          <AppOnboardingNotification v-if="!isWizard && shouldShowOnboardingNotification" />
+          <AppOnboardingNotification v-if="!isWizard && route.name" />
```
kleinfreund, line 24: **Question**: Why is the addition of `route.name` needed?

johncowen: Pretty cure its not necessary, I've removed it, thanks for the spot!

kleinfreund (approved):

PR #1262, opened 2023-07-28
chore: replaces docs link button with book icon link

kleinfreund:
**chore: wraps DataSource in toolbar**

Wraps the `DataSource` element in the toolbar to avoid the empty `span` element it creates from being spaced out with the visible toolbar items. This otherwise creates a gap twice as big with the next following toolbar item.

**chore: replaces docs link button with book icon link**

Replaces the documentation link button with a regular link and a book icon, visually hiding the link text.


src/app/common/DocumentationLink.vue
```diff
@@ -1,28 +1,29 @@
...
     target="_blank"
-    :to="props.href"
   >
-    <KIcon
-      icon="externalLink"
-      color="currentColor"
-      size="16"
-      hide-title
-    />
+    <img src="@/assets/images/book.svg?url">
```
lahabana, line 7: Shouldn't that be: https://kongponents.konghq.com/components/icon.html#icon-1 ?

lahabana (commented):
I don't understand the div addition. I've looked with a previous version and don't see the gap you are mentionning

kleinfreund:
> I don't understand the div addition. I've looked with a previous version and don't see the gap you are mentionning

image

It’s there, but invisible as the `span` added by `DataSource` is empty by default.

The gap between both items here is twice as big as it should be.

src/app/common/DocumentationLink.vue
kleinfreund: I’ve updated this now.

lahabana:
Can you put the doclink in the datasource and avoid this? random divs are always unfortunate :(
Also I'm curious about this cause @johncowen used a span so that it didn't have styling implications

kleinfreund:
> Can you put the doclink in the datasource and avoid this? random divs are always unfortunate :(

That wouldn’t avoid the gap and would require additional styling.

I think the wrapping `div` is the best compromise because then we don’t have to move `DataSource` into a less adequate place.

kleinfreund:
> What I meant is moving `DocumentationLink` inside the `DataSource`. To me it makes sense cause without the data you are not going to have a selected item so no reason to display this.

Ah, I see. Well, it’s a little bit more complex: `selected` comes from line 25 and relates to data from the most top-level `DataSource` in `src/app/policies/views/PolicyListView.vue`. The `DataSource` that’s inside the actions toolbar is only providing the policy dropdown with the data for policy counts from mesh insights and is only related to `selected` (and with that the docs link) insofar as it triggers a change of the route parameters which in turn will cause `selected` to be different.

So, _structurally_, the documentation link is in the right place.

> With this change the span is after the elements and renders differently but actually better (notice the alignment with the details button):

As for the design aspect of your comment, I don’t think it makes sense to use the action dropdown in the table as an alignment point for the toolbar items.

johncowen:
Not really had time to read through a lot of this, but feel free to wrap the `<span />` in DataSource in an `v-if="false"`, we don't use it as yet, and that conditional will be eventually converted into a `v-if="props.loading='lazy'"` anyway.

kleinfreund:
> But feel free to wrap the `<span />` in DataSource in an `v-if="false"`

I like that better and changed that now and removed the wrapping `div` around `DataSource`.

johncowen:
Again looking briefly, I think you could also move the DataSource to wrap AppCollection instead of wrapping the Select directly?

lahabana:
DataSource to wrap AppCollection makes even more sense. If you didn't get the list of PolicyType you for sure don't have any policies to render.

johncowen:
> If you didn't get the list of PolicyType you for sure don't have any policies to render.

Oh yeah! Excellent!

lahabana (approved):
wasn't there also a point about aria-hidden or somehting?

kleinfreund:
> wasn't there also a point about aria-hidden or somehting?

Yes, but I’m not sure what’s the intention here: both `aria-hidden="true"` and `class="visually-hidden"` on an element are effectively hiding it from all users. In short, one might as well use `display: none` or not render the element at all. I suspect the intention is to still give IntersectionObserver a way to find the element which it won’t be able to if it has `display: none`. Is that what you had in mind, @johncowen?

To be honest, as we’re not using this currently, I’d prefer to not render this element at all.

PR #1270, opened 2023-07-31
chore: clean up DPP list in service view

kleinfreund:
**feat(services): adds copy button to address cells**

**fix: data plane list toolbar items misalignment**

**fix(services): data plane proxy list layout**

Fixes the DPP list within the service detail view for external services missing its wrapping KCard.

Before:

image

After:

image

### DIFF
--- a/src/app/common/KFilterBar.vue
+++ b/src/app/common/KFilterBar.vue
@@ -444,7 +444,6 @@ function areFieldsSemanticallyIdentical(fieldsA: Fields, fieldB: Fields): boolea
   flex-grow: 1;
   width: 100%;
   border: none;
-  padding: 12px;
 }
 
 .k-filter-bar-input:focus {
--- a/src/app/services/views/ServiceDetailView.vue
+++ b/src/app/services/views/ServiceDetailView.vue
@@ -30,92 +30,97 @@
         <template
           v-else
         >
-          <ServiceSummary
-            :service="item"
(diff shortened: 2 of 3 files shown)


johncowen (commented):
I was confused about this one when we saw it and wondered when it got like that.

> fix: data plane list toolbar items misalignment

It got to this state here right?

https://github.com/kumahq/kuma-gui/pull/1244

?

kleinfreund:
@johncowen There were a couple of influences breaking this layout. One was the missing KCard. In its absence, different font properties applied sizing the various toolbar items out of proportion. After fixing that, the DPP list in the service detail view and the gateway list view looked mostly fine, but the `padding: 12px` rule still caused the items to be of a different height because KSelect doesn’t grow in a stretch-aligned flex container. Removing the padding rule fixes this on our end.

johncowen:
Right but my question was, where did it get into this state? It seemed to me to be #1244, which I'm pretty sure it was, I just wanted to see if I could get confirmation of that in case I was wrong, no problem! Oh I'd meant to approve here as well 1 sec.

johncowen (approved):

kleinfreund:
> Right but my question was, where did it get into this state?

It got into this state through the removal of KCard and the addition of `padding: 12px` on the KFilter input. https://github.com/kumahq/kuma-gui/pull/1244 was pretty much unrelated to this.

PR #1272, opened 2023-07-31
fix(policy): support link not using feedback URL env

kleinfreund:
**fix(policy): support link not using feedback URL env**

**chore: updates /policies mock**

**fix(policy): policy type in docs link not being lowercased**


src/app/policies/views/PolicyListView.vue
```diff
@@ -56,7 +56,7 @@
                           <p>
                             <strong>Warning</strong> This policy is experimental. If you encountered any problem please open an
                             <a
-                              href="https://github.com/kumahq/kuma/issues/new/choose"
+                              :href="env('KUMA_FEEDBACK_URL')"
```
johncowen, line 59: Could you change this to use `t`. We are moving away from using `env()` for these types of things now that we have a proper i18n layer.

src/app/policies/views/PolicyListView.vue
```diff
@@ -117,7 +117,7 @@
                             </KSelect>
 
                             <DocumentationLink
-                              :href="t('policies.href.docs', {'name': selected.name})"
+                              :href="t('policies.href.docs', { name: selected.name.toLowerCase() })"
```
johncowen, line 120: In relation to the above comment. You can see here how we are now using `t` for hrefs/URLs.

Also, I'm pretty sure @lahabana wanted the URLs here to use the exact unmodified name (i.e. without the lowercasing). I might have misunderstood or things may have changed since then, but it's probably a good idea to double check with him.

johncowen (commented):
Left a couple of comments inline

src/app/policies/views/PolicyListView.vue
kleinfreund: This breaks the URLs though. The redirects on the docs website don’t work when using the plain policy type names.

src/app/policies/views/PolicyListView.vue
kleinfreund: Happy to do this in a follow up, but I don’t have the time for this currently.

src/app/policies/views/PolicyListView.vue
johncowen: Yes I know, I asked that also when this was done. I'd understood they were going to be fixed on the docs site. I might be completely wrong but I think we should at least check before changing this as I think it was done like this on purpose.

src/app/policies/views/PolicyListView.vue
johncowen: How long is this going to take you? Longer than 2 minutes?

src/app/policies/views/PolicyListView.vue
kleinfreund: I have checked this (hence my addition of this fix) and it’s not working like that.

Navigating to a docs URL without the lowercasing which also relies on a redirect will currently 404 because the redirect won’t match against the non-lowercased value.

johncowen: >I have checked this (hence my addition of this fix) and it’s not working like that.

ok well as I said that was the case when @lahabana added this and we discussed it at the time, so I'll have to leave it to him as I just don't know either way. Maybe since then it was agreed to lowercase these?

kleinfreund: There was no agreement of that kind, but it was observed in a couple of places that the documentation URLs are broken and that’s one of the issues. If we decide to make the redirects more robust to work without the lowercasing, we can easily remove this here.

Until then, I much rather have the documentation links working.

johncowen (approved):

PR #1273, opened 2023-07-31
fix(zones): create zone button being visible in standalone mode

kleinfreund:

### DIFF
--- a/src/app/zones/views/ZoneIndexView.vue
+++ b/src/app/zones/views/ZoneIndexView.vue
@@ -20,7 +20,7 @@
       </template>
 
       <template
-        v-if="env('KUMA_ZONE_CREATION_FLOW') === 'enabled'"
+        v-if="env('KUMA_ZONE_CREATION_FLOW') === 'enabled' && isGlobalMode"
         #actions
       >
         <KButton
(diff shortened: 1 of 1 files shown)


slonka (approved):

src/app/zones/views/ZoneIndexView.vue
```diff
@@ -20,7 +20,7 @@
       </template>
 
       <template
-        v-if="env('KUMA_ZONE_CREATION_FLOW') === 'enabled'"
+        v-if="env('KUMA_ZONE_CREATION_FLOW') === 'enabled' && isGlobalMode"
```
johncowen, line 23: Just incase I misunderstood this, as it stands right now do we ever enable this when we aren't in "global mode"? (a.k.a. `isMultizoneMode`)

kleinfreund: Only when the environment is in global model does it make sense to expose the "Create Zone" flow. Does that answer your question?

johncowen: > Do we ever enable this when we aren't in "global mode"?

^ was my question, so I don't think it does no.

I have another question that might help me to understand also.

Guessing here, but is this for when we you can only see Zone Egresses. Does the button show up then if we were to have this enabled in zone sort of non-zone environment?

johncowen: Ah thankyou that makes sense now. Let me take another pass.

PR #1274, opened 2023-07-31
chore: removes nav tabs from detail views

kleinfreund:
**chore(meshes): removes nav tabs from detail views**

**chore(zones): removes nav tabs from detail views**


kleinfreund:
I specifically asked this question to design to have clarity on what we should do with breadcrumbs and that was the outcome:

- Generally, don’t include the current view in the breadcrumbs
- Specifically for Zones and RBAC, don’t include the top-level Zones and RBAC items. They don’t serve a clear purpose

johncowen:
Ok gotcha, so we don't have two levels for Zones and RBAC.

So at the moment we are good with:

```
1. Zone Control Planes
2. Ingresses
3. Egresses
```

?

kleinfreund:
> So at the moment we are good with:

Yes, no more two levels like that and this should be fine. As far as I know, the terms "Ingress" and "Egress" are not ambiguous without the "Zone" prefix. Easy enough to add them again. Now that we drop this level, I probably also want them back.

Overall, I think also the question of what titles should appear where will influence this so I’m sure this is not the last time we’ll pass over this.

PR #1279, opened 2023-08-02
fix: premature access of mesh insight policies field

kleinfreund:


johncowen (approved):
Good spot!

PR #1281, opened 2023-08-02
refactor: use DataSource for diagnostics view

kleinfreund:

### DIFF
--- a/features/diagnostics/DetailViewContent.feature
+++ b/features/diagnostics/DetailViewContent.feature
@@ -1,0 +1,17 @@
+Feature: Diagnostics: Detail view content
+  Background:
+    Given the CSS selectors
+      | Alias   | Selector                               |
+      | details | [data-testid='code-block-diagnostics'] |
+
+  Scenario: Diagnostics detail view has expected content
+    Given the URL "/config" responds with
+      """
+      body:
+        mode: global
+      """
+
+    When I visit the "/diagnostics" URL
+    Then the page title contains "Diagnostics"
+    Then the "$details" element contains ""mode": "global""
+    Then the "$details" element contains ""environment": "universal""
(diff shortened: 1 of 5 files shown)


johncowen (approved):

johncowen:
Oh, do we have a tests for this?

kleinfreund:
> Oh, do we have a tests for this?

I’ll add some tests for this.

features/diagnostics/DetailViewContent.feature
```diff
@@ -0,0 +1,17 @@
...
+    Given the URL "/config" responds with
+      """
+      body:
+        mode: global
+      """
+
+    When I visit the "/diagnostics" URL
+    Then the page title contains "Diagnostics"
+    Then the "$details" element contains ""mode": "global""
+    Then the "$details" element contains ""environment": "universal""
```
slonka, line 17: That's not always the case but I guess that's true for the fixtures that we have, right?

src/app/diagnostics/views/DiagnosticsView.vue
```diff
@@ -1,58 +1,65 @@
...
+
+            <ErrorBlock
+              v-else-if="error"
+              :error="error"
+            />
+
+            <CodeBlock
+              v-else
+              id="code-block-diagnostics"
+              data-testid="code-block-diagnostics"
```
slonka, line 38: just out of curiocity why do we need this with the same value as `id`?

slonka (approved):
Looks fine, just 2 Qs as before

features/diagnostics/DetailViewContent.feature
kleinfreund: That is true. Our fixture currently statically states environment to be universal. We could mock this out here, too, but it’s the same thing in the end.

src/app/diagnostics/views/DiagnosticsView.vue
kleinfreund: It could be a different value. I added the `data-testid` with that value because it doesn’t _need_ to be different and because I don’t want to use `id`s in tests if I can avoid it. This way, tests don’t depend on IDs and classes being the same.

PR #1282, opened 2023-08-02
chore(e2e): adds title assertion wait again

kleinfreund:
Adds the `cy.wait` call we had in the title assertion step back in as tests are still notably flaky without it.

### DIFF
--- a/cypress/support/step_definitions/index.ts
+++ b/cypress/support/step_definitions/index.ts
@@ -213,6 +213,7 @@ Then(/^the "(.*)" element contains "(.*)"$/, (selector: string, value: string) =
 })
 
 Then('the page title contains {string}', function (title: string) {
+  cy.wait(1000)
   cy.title().should('contain', title)
 })
 


cypress/support/step_definitions/index.ts
```diff
@@ -213,6 +213,7 @@ Then(/^the "(.*)" element contains "(.*)"$/, (selector: string, value: string) =
 })
 
 Then('the page title contains {string}', function (title: string) {
+  cy.wait(1000)
```
slonka, line 216: do we have like "waitUntil(title.contain(...))"?

kleinfreund: No, but we do have command timeouts. Unfortunately, in my experience, tests that are flaky related to this step definition don’t become less flaky with long command timeouts.

slonka: I was more thinking about speeding this up not making it less flaky. Something like https://onsi.github.io/gomega/#eventually

kleinfreund: Cypress only works works under that semantic already. Each Cypress command has a built-in timeout (by default 4 seconds). So all our assertions _already_ "wait until assertion is true for 4  seconds".

johncowen: This will hopefully eventually be replaced by request counting i.e. after visiting a page don't proceed until all of our requests have responded i.e. when the client is idle. I 'think' this would remove the need for dropping waits everywhere.

PR #1284, opened 2023-08-03
refactor: use data source for zone views

kleinfreund:
**feat: adds isLoading state to DataSource**

**refactor: use DataSource for Zone CP detail view**

**refactor: use DataSource for Zone Egress detail view**

**refactor: use DataSource for Zone Ingress detail view**

### DIFF
--- a/src/app/application/components/data-source/DataSource.vue
+++ b/src/app/application/components/data-source/DataSource.vue
@@ -1,5 +1,15 @@
-<script setup lang="ts">
+<template>
+  <slot
+    :data="(message as any)"
+    :is-loading="isLoading"
+    :error="error"
+    :refresh="refresh"
+  />
 
+  <span class="visually-hidden" />
+</template>
+
+<script lang="ts" setup>
 import { watch, ref, onBeforeUnmount } from 'vue'
 
 import { useDataSourcePool } from '@/utilities'
@@ -14,6 +24,7 @@ const props = defineProps({
 })
 
 const message = ref<unknown>(undefined)
+const isLoading = ref(false)
 const error = ref<Error | undefined>(undefined)
 
 const emit = defineEmits<{
@@ -34,23 +45,30 @@ const open = async (src: string) => {
   if (src === '') {
     return
   }
+  isLoading.value = true
   state.controller = new AbortController()
   // this should emit proper events
   const source = data.source(src, sym)
+
   source.addEventListener(
     'message',
     (e) => {
       message.value = (e as MessageEvent).data
       // if we got a message we are no longer erroneous
       error.value = undefined
+      isLoading.value = false
+
       emit('change', message.value)
     },
     { signal: state.controller.signal },
   )
+
   source.addEventListener(
     'error',
     (e) => {
       error.value = (e as ErrorEvent).error as Error
+      isLoading.value = false
+
       emit('error', error.value)
     },
     { signal: state.controller.signal },
@@ -73,12 +91,3 @@ const refresh = () => {
   open(props.src)
 }
 </script>
-
-<template>
-  <slot
-    :data="(message as any)"
-    :error="error"
-    :refresh="refresh"
-  />
-  <span class="visually-hidden" />
-</template>
--- a/src/app/application/services/data-source/DataSourcePool.ts
+++ b/src/app/application/services/data-source/DataSourcePool.ts
@@ -6,7 +6,12 @@ import SharedPool from './SharedPool'
 export type Source = (params: Record<string, unknown>, source: {close: () => void}) => Promise<unknown>
 export type Sources = Record<string, Source>
 // reusable Type Utility for easy to use Types within Vue templates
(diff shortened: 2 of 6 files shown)


src/app/application/components/data-source/DataSource.vue
```diff
@@ -1,5 +1,15 @@
-<script setup lang="ts">
+<template>
+  <slot
+    :data="(message as any)"
```
slonka, line 3: I'm sorry but I'm going to be tedious for the first couple of ones :(

why is this `any`? can't we have a concrete type there? or does it not give us anything?

src/app/zones/sources.ts
```diff
@@ -22,7 +28,7 @@ export type ZoneEgressOverviewCollectionSource = DataSourceResponse<ZoneEgressOv
 
 export const sources = (api: KumaApi) => {
   return {
-    '/zone-cps': async (params: PaginationParams, source: { close: () => void }) => {
+    '/zone-cps': async (params: PaginationParams, source: Closeable) => {
```
slonka, line 31: it's not strictly to this PR but seems like we're mixing up singular plural in our URLs, we have `.../mesh/...` but we have `.../policy/circuit-breakers/...` - do you know the reason for this? should we standardize on one? I think we could do what github does with `/pulls` for listing and `/pull/x` for single PR.

slonka (approved):
in general ok, just 2 questions

src/app/application/components/data-source/DataSource.vue
kleinfreund: Note that this bit was just moved around from the bottom of the file (all our components start with the template).

Typing this component is hard because what we effectively need is _the component_ (i.e. not some JavaScript function) to be typed with a generic parameter. Then, `data` should be something like `DataSource<T>` instead of `any`.

Vue (as of _very_ recently) supports typing components as generics, but when I tried it locally, I couldn’t avoid some TypeScript compilation errors that were introduced by doing so.

In short, doing this better is currently not possible.

src/app/zones/sources.ts
slonka: > there, we have no control

we do have control on the back-end right? maybe we need to do a better job on the back-end

kleinfreund: Oh, sure. I was just talking about the frontend perspective. But then again, these are APIs consumed by other people and software, too. Harder to change than what’s isolated to the UI (the other two types).

johncowen: Just to note:

If I understand @slonka correctly here this is exactly what we should be doing for DataSource URIs

> I think we could do what github does with /pulls for listing and /pull/x for single PR.


i.e.

```
/zone-cps => list
/zone-cp/:name => item
```

Listings should use plurals, 'singles' should not.

Not super related to this but for browser URLs we should use plural if we are looking at a "summary view" i.e. listing plus detail, and single if we are looking at a detail view (which has no listing)

i.e. 

```
/mesh/default/services/service-1  => listing plus drawer (we don't have views like this yet)
/mesh/default/service/service-1 => just the detail view for service-1 i.e. no listing on the page
```

lahabana: In the backend we always use plural for resources regardless if it's a list or an item. Because Datasource separates things from the backend you can do whatever but it should at least be coherent inside the component

src/app/application/services/data-source/DataSourcePool.ts
```diff
@@ -6,7 +6,12 @@ import SharedPool from './SharedPool'
 export type Source = (params: Record<string, unknown>, source: {close: () => void}) => Promise<unknown>
 export type Sources = Record<string, Source>
 // reusable Type Utility for easy to use Types within Vue templates
-export type DataSourceResponse<T> = {data: T | undefined, error: Error | undefined, refresh: () => void}
+export type DataSourceResponse<T> = {
+  data: T | undefined
+  isLoading: boolean
```
lahabana, line 11: Did we never need this for table views?

kleinfreund: We did and we had this for all of them prior to migrating to DataSource/AppCollection. Instead, some of the views now use `data` being `undefined` as the signal for loading.

I think the loading state of data fetching should be indicated separately than via the variable that holds the data.

johncowen: > We did 

No we didn't.

We used the simple rules of:

- If we have no data then we are not 'connected' ...yet
- If we have no error then we are not 'erroneous' ...yet

Just to note, there will be a separate component that deals with data loading and showing visuals depending on the state depending on the above rules. DataSource is supposed to be very low level and with a very small API. For example here, with a polling or EventSource based DataSource `isLoading` is never reset to `false` meaning at the very least this state is confusingly named.

Another sidenote here is that as DataSources are theoretically never completely loaded they (or rather the component that will deal with those states) should have at least 4 states probably named something like `connecting`, `connected`, `error`, `disconnected`.

PR #1285, opened 2023-08-03
chore: handles /provision-zone validation errors

kleinfreund:
**chore(zones): handles validation errors on /provision-zone**

**chore(zones): adds tooltip detailing naming rules**

### DIFF
--- a/src/app/zones/views/ZoneCreateView.vue
+++ b/src/app/zones/views/ZoneCreateView.vue
@@ -31,8 +31,16 @@
 
         <div class="form-wrapper mt-4">
           <div>
-            <KLabel for="zone-name">
-              {{ t('zones.form.nameLabel') }} *
+            <KLabel
+              for="zone-name"
+              required
+              :tooltip-attributes="{ placement: 'right'}"
+            >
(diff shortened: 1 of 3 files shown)


src/app/zones/views/ZoneCreateView.vue
```diff
@@ -277,7 +283,7 @@ async function createZone() {
   try {
     zone.value = await kumaApi.createZone({ name: name.value })
   } catch (err) {
-    if (err instanceof ApiError && HANDLED_STATUS_CODES.includes(err.status)) {
+    if (err instanceof ApiError && [409, 500].includes(err.status)) {
```
slonka, line 286: nitpick: why did you inline it? could've just been changed at the 241 line.

kleinfreund: I figured it’s better to have this inline as the constant wasn’t re-used and it doesn’t really aid in the readability here. It’s only two values now.

slonka (approved):
High priority enough even thought it's not strictly related to current effort.

PR #1286, opened 2023-08-03
refactor: uses DataSource for mesh detail view

kleinfreund:
Changes MeshOverviewView to use DataSource and to solely be responsible for fetching data. All further presentational purposes where moved into the new MeshDetails and the existing MeshCharts components.

Adds the new MeshDetails component for the presentational purposes of the mesh detail view. Its pretty much a copy of what used to be in MeshOverviewView.

Changes MeshCharts to receive `meshInsight` as a prop instead of fetching the data itself. The mesh detail view always loads that data anyway, so this removes an unnecessary request from this view.

### DIFF
--- a/src/app/meshes/components/MeshDetails.vue
+++ b/src/app/meshes/components/MeshDetails.vue
@@ -1,0 +1,186 @@
+<template>
+  <div class="stack">
+    <KCard>
+      <template #body>
+        <MeshCharts :mesh-insight="meshInsight" />
+      </template>
+    </KCard>
+
+    <KCard>
+      <template #body>
+        <div class="columns">
+          <DefinitionList>
+            <DefinitionListItem
+              v-for="(value, property) in basicMesh"
+              :key="property"
+              :term="t(`http.api.property.${property}`)"
+            >
+              <KBadge
+                v-if="typeof value === 'boolean'"
+                :appearance="value ? 'success' : 'danger'"
+              >
+                {{ value ? 'Enabled' : 'Disabled' }}
+              </KBadge>
+
+              <template v-else-if="property === 'name' && typeof value === 'string'">
+                <TextWithCopyButton :text="value" />
+              </template>
+
+              <template v-else>
+                {{ value }}
+              </template>
+            </DefinitionListItem>
+          </DefinitionList>
+
+          <DefinitionList>
+            <DefinitionListItem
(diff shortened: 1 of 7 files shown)


src/app/meshes/components/MeshDetails.vue
```diff
@@ -0,0 +1,186 @@
...
+      </template>
+    </KCard>
+
+    <KCard>
+      <template #body>
+        <ResourceCodeBlock
+          id="code-block-mesh"
+          :resource-fetcher="fetchMesh"
+          :resource-fetcher-watch-key="mesh?.name || null"
+        />
```
slonka, line 85: github syntax highlighting turned blue here, it's probably their bug but maybe we can just do `>` like on line 68 with `router-link` so it just works.

src/app/meshes/components/MeshDetails.vue
```diff
@@ -0,0 +1,186 @@
...
+          total: stat.total,
+        }
+      }
+
+      return null
+    })
+    .filter(notEmpty)
+})
+
+function getBackendData(mesh: Mesh, field: 'mtls' | 'logging' | 'metrics' | 'tracing'): string | boolean {
```
slonka, line 171: isn't `string or boolean` a bit weird of a type?

slonka (approved):
mostly looks good, just the "string | boolean" return type looks kinda werid

src/app/meshes/components/MeshDetails.vue
kleinfreund: The syntax highlighting GitHub uses is unfortunately quite out-of-date for .vue files.

Somewhat modern syntax like `mesh?.name` trips up the syntax highlighter. We _could_ "fix" this by wrapping such expressions in parenthesis which has proven to work (and we do it sometimes), but really, we shouldn’t alter our code to satisfy the GH diff highlighting.

And yeah, the reason is not the self-closing syntax on line 85.

src/app/meshes/components/MeshDetails.vue
kleinfreund: You’re right. I’ve rewrote this a bit here: https://github.com/kumahq/kuma-gui/pull/1286/commits/37f0444d9b01c967e72e0bc2ce0aeba3b336c5b1

PR #1287, opened 2023-08-03
refactor: uses DataSource for service detail view

kleinfreund:

### DIFF
--- a/src/app/services/sources.ts
+++ b/src/app/services/sources.ts
@@ -1,34 +1,57 @@
 import { DataSourceResponse } from '@/app/application/services/data-source/DataSourcePool'
 import type KumaApi from '@/services/kuma-api/KumaApi'
-import type {
-  PaginatedApiListResponse as CollectionResponse,
-} from '@/types/api.d'
-import type {
-  ServiceInsight,
-} from '@/types/index.d'
+import type { PaginatedApiListResponse as CollectionResponse } from '@/types/api.d'
+import type { ExternalService, ServiceInsight } from '@/types/index.d'
 
-type MeshParams = {
+type CollectionParams = {
   mesh: string
 }
+
+type DetailParams = CollectionParams & {
+  name: string
+}
+
 type PaginationParams = {
   size: number
   page: number
 }
+
 type Closeable = { close: () => void }
+
+export type ServiceInsightSource = DataSourceResponse<ServiceInsight>
 export type ServiceInsightCollection = CollectionResponse<ServiceInsight>
 export type ServiceInsightCollectionSource = DataSourceResponse<ServiceInsightCollection>
 
+export type ExternalServiceSource = DataSourceResponse<ExternalService>
+
 export const sources = (api: KumaApi) => {
   return {
-    '/:mesh/services': async (params: MeshParams & PaginationParams, source: Closeable) => {
+    '/:mesh/service-insights': async (params: CollectionParams & PaginationParams, source: Closeable) => {
       source.close()
+
+      const mesh = params.mesh
+      const size = params.size
       const offset = params.size * (params.page - 1)
-      return api.getAllServiceInsightsFromMesh({
-        mesh: params.mesh,
-      }, {
-        size: params.size,
-        offset,
-      })
+
+      return api.getAllServiceInsightsFromMesh({ mesh }, { size, offset })
+    },
+
+    '/:mesh/service-insights/:name': (params: DetailParams, source: Closeable) => {
+      source.close()
+
+      const mesh = params.mesh
+      const name = params.name
+
+      return api.getServiceInsight({ mesh, name })
+    },
+
+    '/:mesh/external-services/:name': (params: DetailParams, source: Closeable) => {
+      source.close()
+
+      const mesh = params.mesh
+      const name = params.name
+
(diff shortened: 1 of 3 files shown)


src/app/services/sources.ts
```diff
@@ -1,34 +1,57 @@
...
+      const name = params.name
+
+      return api.getServiceInsight({ mesh, name })
+    },
+
+    '/:mesh/external-services/:name': (params: DetailParams, source: Closeable) => {
+      source.close()
+
+      const mesh = params.mesh
+      const name = params.name
```
slonka, line 52: nitpick: could be `const { mesh, name } = { params.mesh, params.name }` right?

src/app/services/views/ServiceDetailView.vue
```diff
@@ -18,120 +18,127 @@
...
+          :src="`/${route.params.mesh}/external-services/${route.params.service}`"
+        >
+          <LoadingBlock v-if="isLoading || isLoadingExternalService" />
 
-        <ErrorBlock
-          v-else-if="error !== null"
-          :error="error"
-        />
+          <ErrorBlock
+            v-else-if="error ?? externalServiceError"
```
slonka, line 32: what is this operator? :O https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing wow

slonka: oh, so you can use this instead of `x === null ? [] : x` thing that I mentioned in the previous PR

slonka (approved):
just nitpicks and questions

src/app/services/sources.ts
kleinfreund: Almost. It could use destructuring to look like this:

```js
const { mesh, name } = params
```

Let me change this when I work on the next pass over the detail views so I save myself the CI pipeline here.

src/app/services/views/ServiceDetailView.vue
kleinfreund: Nullish in JavaScript are values that are either `null` or `undefined`. Generally, I prefer very explicit conditionals (e.g. `array.length === 0` instead of `!array.length`, `error !== undefined` instead of `error`), but in the interest of making these expressions shorter, I opted for this instead.

So, here, I would’ve otherwise written (`null` isn’t an expected value here):

```vue
            v-else-if="error !== undefined ?? externalServiceError !== undefined"
```

src/app/services/sources.ts
```diff
@@ -1,34 +1,57 @@
...
+export type ServiceInsightSource = DataSourceResponse<ServiceInsight>
 export type ServiceInsightCollection = CollectionResponse<ServiceInsight>
 export type ServiceInsightCollectionSource = DataSourceResponse<ServiceInsightCollection>
 
+export type ExternalServiceSource = DataSourceResponse<ExternalService>
+
 export const sources = (api: KumaApi) => {
   return {
-    '/:mesh/services': async (params: MeshParams & PaginationParams, source: Closeable) => {
+    '/:mesh/service-insights': async (params: CollectionParams & PaginationParams, source: Closeable) => {
```
lahabana, line 29: Why change this name?

src/app/services/views/ServiceDetailView.vue
```diff
@@ -18,120 +18,127 @@
...
-      <div class="service-details">
-        <LoadingBlock v-if="isLoading" />
+      <DataSource
+        v-slot="{ data, isLoading, error }: ServiceInsightSource"
+        :src="`/${route.params.mesh}/service-insights/${route.params.service}`"
+      >
+        <DataSource
+          v-slot="{ data: externalService, isLoading: isLoadingExternalService, error: externalServiceError }: ExternalServiceSource"
+          :src="`/${route.params.mesh}/external-services/${route.params.service}`"
+        >
```
lahabana, line 28: Seems like this complexity should be pushed in the `sources` part of your datasource instead of here no?

src/app/services/sources.ts
```diff
@@ -1,34 +1,57 @@
...
+    '/:mesh/service-insights/:name': (params: DetailParams, source: Closeable) => {
+      source.close()
+
+      const mesh = params.mesh
+      const name = params.name
+
+      return api.getServiceInsight({ mesh, name })
+    },
+
+    '/:mesh/external-services/:name': (params: DetailParams, source: Closeable) => {
```
lahabana, line 48: Sounds like making this `/:mesh/service/:name` and hide here pulling from both external service and service-insights would make the component simpler. I think that's actually what @johncowen is hoping to achieve with the datasource: moving API complexity somewhere and keep in the component only the presentation

src/app/services/sources.ts
kleinfreund: Because the services term is pretty overloaded in the UI application due to there being services in the context of our DI system, services in the context of the application domain (`ServiceInsight`, `ExternalService`). Personally, I found it easiest to formulate these paths roughly after the names of the types/resources we access through them.

src/app/services/sources.ts
kleinfreund: Yeah, I agree. I’ll be moving this around in https://github.com/kumahq/kuma-gui/pull/1345.

PR #1288, opened 2023-08-03
refactor: uses DataSource for DPP detail view

kleinfreund:
Uses DataSource for data fetching purposes in the DPP detail view.

Also changes the child components slightly to no longer require a `Dataplane` object. All the data we need is already available on the `DataplaneOverview` object so we can save a request here.

### DIFF
--- a/src/app/data-planes/components/DataPlaneDetails.vue
+++ b/src/app/data-planes/components/DataPlaneDetails.vue
@@ -4,9 +4,9 @@
       <h1 class="entity-heading">
         DPP:
 
-        <TextWithCopyButton :text="dataPlane.name">
+        <TextWithCopyButton :text="props.dataplaneOverview.name">
           <router-link :to="detailViewRoute">
-            {{ dataPlane.name }}
+            {{ props.dataplaneOverview.name }}
           </router-link>
         </TextWithCopyButton>
       </h1>
@@ -61,7 +61,7 @@
         id="code-block-data-plane"
         class="mt-4"
         :resource-fetcher="fetchDataPlaneProxy"
-        :resource-fetcher-watch-key="props.dataPlane.name"
+        :resource-fetcher-watch-key="props.dataplaneOverview.name"
         is-searchable
       />
     </template>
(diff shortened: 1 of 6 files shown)


src/app/data-planes/components/DataPlaneDetails.vue
```diff
@@ -4,9 +4,9 @@
       <h1 class="entity-heading">
         DPP:
 
-        <TextWithCopyButton :text="dataPlane.name">
+        <TextWithCopyButton :text="props.dataplaneOverview.name">
```
slonka, line 7: nitpick: sometimes is `dataPlaneOverview` and sometimes `dataplaneOverview`

src/app/data-planes/sources.ts
```diff
@@ -1,72 +1,89 @@
...
       const offset = params.size * (params.page - 1)
+
       // here 'all' means both proxies/sidecars and gateways this currently fits
       // our usecases but we should probably include `gateway | sidecar` or
       // similar
-      const search = Object.fromEntries(normalizeFilterFields(JSON.parse(params.search || '[]')))
-      if (typeof search.tag === 'undefined') {
-        search.tag = []
+      const filterParams = Object.fromEntries(normalizeFilterFields(JSON.parse(params.search || '[]')))
+      if (typeof filterParams.tag === 'undefined') {
```
slonka, line 74: why is this a typeof? shouldn't this just be 

```
filterParams.tag === undefined
```
?

slonka (approved):
LGTM just a couple of Qs

src/app/data-planes/components/DataPlaneDetails.vue
kleinfreund: I’ve slowly started to name things using the `dataplane` rather than `dataPlane` because the resource names from the API are `Dataplane` and `DataplaneOverview` even though the `p` should technically be capitalized. I only did this for things here that I actually needed to touch to avoid making the diff bigger than it needed to be.

src/app/data-planes/sources.ts
kleinfreund: I prefer the condition style to be `filterParams.tag === undefined` as well. As far as I know, it makes no difference. I could change this, but it’s just a difference in "programmer preference" that I don’t feel too strongly about, so I’m rather leaving it as is.

> filterParams.tag = filterParams.tag === undefined ? [] : filterParams.tag

As for this, I do prefer the style of using an if statement as I’m not unnecessarily assigning a property (even though here it would be assigning its original value).

src/app/data-planes/sources.ts
```diff
@@ -1,72 +1,89 @@
...
+        filterParams.tag = []
       }
-      search.tag = search.tag.filter((item) => !item.startsWith('kuma.io/service:'))
-      search.tag.push(`kuma.io/service:${params.service}`)
+      filterParams.tag = filterParams.tag.filter((item) => !item.startsWith('kuma.io/service:'))
```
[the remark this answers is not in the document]
kleinfreund: We’re perhaps at a point where that makes sense. We do have a couple of place where we use the annotations in the form of string literals. I’m leaving this as-is for the time being seeing as it’s very easy to search the project for instances of it and move them into constants.

PR #1289, opened 2023-08-03
refactor: use DataSource for policy detail view

kleinfreund:
**refactor: uses DataSource for policy connections**

**chore(policies): cleans up list view**

**chore: cleans up policy detail view, adds name to route view**

### DIFF
--- a/src/app/policies/components/PolicyConnections.spec.ts
+++ b/src/app/policies/components/PolicyConnections.spec.ts
@@ -100,7 +100,7 @@ describe('PolicyConnections.vue', () => {
   test('renders no item', async () => {
     const server = useServer()
     server.use(
-      rest.get(import.meta.env.VITE_KUMA_API_SERVER_URL + '/meshes/:mesh/:policyPath/:policyName/dataplanes', (req, res, ctx) =>
+      rest.get(import.meta.env.VITE_KUMA_API_SERVER_URL + '/meshes/:mesh/:policyPath/:policyName/dataplanes', (_req, res, ctx) =>
         res(ctx.status(200), ctx.json({ total: 0, items: [] })),
       ),
     )
--- a/src/app/policies/components/PolicyConnections.vue
+++ b/src/app/policies/components/PolicyConnections.vue
@@ -1,49 +1,60 @@
 <template>
-  <StatusInfo
-    :has-error="hasError"
-    :is-loading="isLoading"
-    :is-empty="!hasDataplanes"
+  <h2>Dataplanes</h2>
+
+  <input
+    id="dataplane-search"
+    v-model="searchInput"
+    type="text"
+    class="k-input mt-4"
+    placeholder="Filter by name"
+    required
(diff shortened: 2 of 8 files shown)


src/app/policies/components/PolicyConnections.spec.ts
```diff
@@ -100,7 +100,7 @@ describe('PolicyConnections.vue', () => {
   test('renders no item', async () => {
     const server = useServer()
     server.use(
-      rest.get(import.meta.env.VITE_KUMA_API_SERVER_URL + '/meshes/:mesh/:policyPath/:policyName/dataplanes', (req, res, ctx) =>
+      rest.get(import.meta.env.VITE_KUMA_API_SERVER_URL + '/meshes/:mesh/:policyPath/:policyName/dataplanes', (_req, res, ctx) =>
```
slonka, line 103: nitpick: _req seems not used so why the rename?

```suggestion
      rest.get(import.meta.env.VITE_KUMA_API_SERVER_URL + '/meshes/:mesh/:policyPath/:policyName/dataplanes', (req, res, ctx) =>
```

slonka (approved):
LGTM

src/app/policies/components/PolicyConnections.spec.ts
kleinfreund: Yeah, I now realize we should leave this as `req` to trigger the linter error for unused variables should this be refactored to no longer access the other positional arguments. I’ll change it back in a follow up.

src/app/policies/components/PolicyConnections.vue
```diff
@@ -1,49 +1,60 @@
...
-    >
-      <router-link
-        :to="{
-          name: 'data-plane-detail-view',
-          params: {
-            mesh: dataplane.dataplane.mesh,
-            dataPlane: dataplane.dataplane.name,
-          },
-        }"
+    <EmptyBlock v-else-if="data === undefined || data.items.length === 0" />
```
johncowen, line 25: If I'm not mistaken these kinds of extra conditionals shouldn't be needed, if data is `undefined` then we are in a loading state. This fact is now hidden by the addition of `isLoading` whereas if we always use `typeof data === undefined` to check whether we are waiting for our initial load or not it becomes much clearer that these sorts of double conditionals aren't required.

src/app/policies/sources.ts
```diff
@@ -1,50 +1,56 @@
...
   return {
-    '/*/policy-types': async (_params: {}, source: {close: () => void}) => {
+    '/*/policy-types': (_params: {}, source: Closeable) => {
       source.close()
+
       return api.getPolicyTypes()
     },
-    '/:mesh/policy-type/:policyType': async (params: MeshParams & PolicyTypeParams & PaginationParams, source: {close: () => void}) => {
+
+    '/:mesh/:path': (params: CollectionParams & PaginationParams, source: Closeable) => {
```
lahabana, line 39: The version above was more readable why change?

PR #1292, opened 2023-08-07
refactor: removes tailwind

kleinfreund:
## Changes

Removes Tailwind and replaces/adjusts all uses of remaining utility classes. The utility classes `text-right` and `text-center` were added to our own as we’re using them.

Removes PostCSS as we are not using it anymore.

Removes the CSS rule that reset list styles from `ul` elements and adjusts affected views accordingly (i.e. resetting the style where necessary, e.g. in AccordionList).

Adds a handful of default styles to things like headings, paragraphs, and lists to reduce the need to manually add margin utilities to them. Components that did set custom styles for such purposes were updated to no longer do so.

## Notes

* This PR is a preparation for changes coming to the `@kong/kongponents` package soon. It will require us to replace all uses of utility classes, colors, etc. that come from Kongponents with either our own or ones coming from `@kong/design-tokens`. That process will be made easier if we don’t have two sources of such utilities: Kongponents _and_ Tailwind. We don’t want to use Tailwind anyway.


johncowen (approved):
I think this looks fine to me, I took it for a brief spin and didn't notice anything drastic, plus I think this will come together more once we start using the new kong module.

PR #1293, opened 2023-08-07
refactor: integrates config module into root state

kleinfreund:
**refactor: integrates config module into root store**

Integrates the config module into the root store as there isn’t a good reason to keep it separate. With this, accessing its state has become easier (`store.getters['config/getMulticlusterStatus']` → `store.getters.isGlobalMode`, `store.getters['config/getEnvironment']` → `store.state.environment`, `store.getters['config/getConfigurationType']` → `store.state.configurationType`). This also makes the store’s `bootstrap` action only interact with and rely on data in `storeConfig` which should make any future refactors related to it simpler. Furthermore, we only store things we need from the `getConfig` call in `storeConfig` to make it easier to remove/replace them in the future.

**refactor: renames former config store uses**

Renames all instances of former config store uses to their newer variants.


johncowen:
Just to note, as we briefly spoke about this offline.

There is already work underway (including https://github.com/kumahq/kuma-gui/pull/1261 plus more unPRed) which moves all the config things to use the variables form the index.html file. This will mean we no longer have to make a call to `/config` upfront and store things in the store.

I hope to PR more over the next day or so and it that work will supercede anything that uses the store for this.

PR #1294, opened 2023-08-07
fix: non-unique DataSource paths

kleinfreund:

### DIFF
--- a/src/app/data-planes/sources.ts
+++ b/src/app/data-planes/sources.ts
@@ -34,7 +34,7 @@ export type DataPlaneCollectionSource = DataSourceResponse<DataPlaneCollection>
 
 export const sources = (api: KumaApi) => {
   return {
-    '/:mesh/dataplanes': async (params: CollectionParams & PaginationParams, source: Closeable) => {
+    '/meshes/:mesh/dataplanes': async (params: CollectionParams & PaginationParams, source: Closeable) => {
       source.close()
 
       const { mesh, size } = params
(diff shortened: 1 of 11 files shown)


slonka (approved):

src/app/data-planes/sources.ts
```diff
@@ -34,7 +34,7 @@ export type DataPlaneCollectionSource = DataSourceResponse<DataPlaneCollection>
 
 export const sources = (api: KumaApi) => {
   return {
-    '/:mesh/dataplanes': async (params: CollectionParams & PaginationParams, source: Closeable) => {
+    '/meshes/:mesh/dataplanes': async (params: CollectionParams & PaginationParams, source: Closeable) => {
```
lahabana, line 37: Why do we need to also add this `/meshes` prefix ?

kleinfreund: I found that if you’re not careful, you can easily create non-unique paths so I’m prefixing every dynamic path segment with the corresponding term.

So, for example, if you define a source with path `/:mesh/:path` and one with `/:id/test`, depending on which is found first, you might get the wrong one. Prefixing the dynamic segments makes them unambiguous.

johncowen: Yeah everything should be prefixed to prevent accidental clashes (ideally with the name of the 'module'). This was the reasoning for the weird `/*/policy-types` and `/*/zones` whereever it was I did that. But unfortunately I forgot that was the reason 🤦

PR #1295, opened 2023-08-07
fix: incorrect DataSource path

kleinfreund:


slonka (approved):

lahabana:
How was this not caught in tests?

kleinfreund:
> How was this not caught in tests?

@lahabana This never broke anything in Kuma GUI because we didn’t happen to have ambiguity in the paths.

PR #1297, opened 2023-08-07
feat: handles Zone Control Plane `zone.enabled` field

kleinfreund:
Treats a Zone Control Plane's `zone.enabled` field as a signal to show its status as "disabled" when its value is `false`.

### DIFF
--- a/src/app/main-overview/components/ZoneControlPlanesDetails.vue
+++ b/src/app/main-overview/components/ZoneControlPlanesDetails.vue
@@ -39,7 +39,7 @@ import { KTable } from '@kong/kongponents'
 import { PropType, computed } from 'vue'
 
 import StatusBadge from '@/app/common/StatusBadge.vue'
-import type { ZoneOverview } from '@/types/index.d'
+import type { StatusKeyword, ZoneOverview } from '@/types/index.d'
 import { useI18n } from '@/utilities'
 import { getItemStatusFromInsight } from '@/utilities/dataplane'
 
@@ -55,7 +55,13 @@ const props = defineProps({
 const tableData = computed(() => props.zoneOverviews.map((zoneOverview) => {
   const { name } = zoneOverview
 
-  const status = getItemStatusFromInsight(zoneOverview.zoneInsight)
+  let status: StatusKeyword | 'disabled'
+
+  if (zoneOverview.zone.enabled === false) {
+    status = 'disabled'
+  } else {
+    status = getItemStatusFromInsight(zoneOverview.zoneInsight)
+  }
 
   return {
     name,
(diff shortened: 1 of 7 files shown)


src/app/main-overview/components/ZoneControlPlanesDetails.vue
```diff
@@ -55,7 +55,13 @@ const props = defineProps({
...
   const { name } = zoneOverview
 
-  const status = getItemStatusFromInsight(zoneOverview.zoneInsight)
+  let status: StatusKeyword | 'disabled'
+
+  if (zoneOverview.zone.enabled === false) {
+    status = 'disabled'
+  } else {
+    status = getItemStatusFromInsight(zoneOverview.zoneInsight)
+  }
```
slonka, line 64: nitpick: if I read it correctly and by default it's disabled then why do we need to set it to disabled one more time? can we just do:

```suggestion
  if (zoneOverview.zone.enabled === true) {
    status = getItemStatusFromInsight(zoneOverview.zoneInsight)
  }
```

src/app/zones/views/ZoneListView.vue
```diff
@@ -252,7 +252,13 @@ function transformToTableData(zoneOverviews: ZoneOverview[]): ZoneOverviewTableR
...
     })
 
-    const status = getItemStatusFromInsight(zoneOverview.zoneInsight)
+    let status: StatusKeyword | 'disabled'
+
+    if (zoneOverview.zone.enabled === false) {
+      status = 'disabled'
+    } else {
+      status = getItemStatusFromInsight(zoneOverview.zoneInsight)
+    }
```
slonka, line 261: same as above

slonka (approved):
just 2 nits

src/app/main-overview/components/ZoneControlPlanesDetails.vue
kleinfreund: There is no default value in this context currently. Perhaps you read the type annotation above wrong?

Current:

```ts
  let status: StatusKeyword | 'disabled'

  if (zoneOverview.zone.enabled === false) {
    status = 'disabled'
  } else {
    status = getItemStatusFromInsight(zoneOverview.zoneInsight)
  }
```

We can however do what you suggest:

```ts
  let status: StatusKeyword | 'disabled' = 'disabled'

  if (zoneOverview.zone.enabled === true) {
    status = getItemStatusFromInsight(zoneOverview.zoneInsight)
  }
```

I made those changes in the two relevant places now.

src/app/zones/views/ZoneListView.vue
kleinfreund: Addressed in https://github.com/kumahq/kuma-gui/pull/1297#discussion_r1289659869.

PR #1308, opened 2023-08-08
refactor: standardizes detail view titles

kleinfreund:
Uses the same style of titles for all detail views.

Removes the top-level title from Zones.


slonka (approved):
LGTM

src/app/policies/views/PolicyDetailView.vue
```diff
@@ -19,6 +18,18 @@
         },
       ]"
     >
+      <template #title>
+        <h2>
+          <RouteTitle
+            :title="t('policies.routes.item.title', {
+              name: route.params.policy,
+              type: store.state.policyTypesByPath[route.params.policyPath]?.name ?? route.params.policyPath,
```
johncowen, line 26: Little note: Unless I'm mistaken, we should be able to get the name and type from the response from the DataSource and therefore not rely on cached store data.

kleinfreund: That’s correct, but it would also mean that we would need to delay setting the document title until this data has arrived. We would need to wrap this file’s DataSource _around_ AppView so the data is even in scope and then we’ll possibly end up having a flash of a changing document title that I really want to avoid.

johncowen: I don't think any of that would be an issue. We take this approach pretty much everywhere else.

PR #1309, opened 2023-08-09
fix: browser back navigation not working

kleinfreund:
## Changes

**fix: browser back navigation not working**

## To reproduce

Navigate from the home page to a list view (e.g. via one of the "view all" links) and press the Back button in the browser. You will stay in the list view when you should move back to the initial page.

### DIFF
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -16,12 +16,10 @@
       :env="env"
       :route="{
         update: (params: Record<string, string | undefined>) => {
-          router.push(
-            {
-              name: props.name,
-              query: cleanQuery(params, route.query),
-            }
-          )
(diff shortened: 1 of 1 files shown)


slonka (approved):

johncowen:
If I'm not mistaken  it looks like if I click to page 2 in the listing, and then click the back button, I do not go back to page 1? I go back to wherever I was before going to the listing page.

johncowen:
Ok thanks for taking a look, kinda crossing threads here and offline, but I think by default all updates to query params should be added to the history (apart from the odd outlier as explained offline). I think my point here is that `update` should `push` not `replace`.

Do you think w could approach this differently by avoiding the redirect or only making the redirect use a `replace`?

Overall lets leave this for the moment until we hear input from others.

PR #1310, opened 2023-08-09
chore: cleans up some dead code

kleinfreund:
Left this in by accident.


slonka (approved):
Approved but shouldn't linter catch this?

kleinfreund:
> Approved but shouldn't linter catch this?

That’s not possible. The code removed here was used and Vue won’t be able to know that the value I’m passing to the component isn’t just a plain attribute. Much like it can’t know whether I actually use a `class="certainly-not-used"`.

PR #1311, opened 2023-08-09
refactor: standardizes TabsWidget usage

kleinfreund:
## Changes

Standardizes the usage of the TabsWidget component to not include a KCard as the new detail view designs don’t have a card around all of their content (though sometimes parts of it).

Removes a bunch of code from the TabsWidget component that we were already not using anymore.

## Notes

- The biggest contributor to the number of lines changed is a regenerated snapshot for a snapshot test. When I’ll work on implementing a new design for Data Plane Proxies, I’ll add some tests and remove the snapshot test for that component.

### DIFF
--- a/src/app/common/TabsWidget.vue
+++ b/src/app/common/TabsWidget.vue
@@ -1,66 +1,35 @@
 <template>
-  <div
-    class="tab-container"
-    data-testid="tab-container"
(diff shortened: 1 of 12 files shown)


slonka (approved):
LGTM

PR #1312, opened 2023-08-09
feat: implements new zone details designs

kleinfreund:
## Changes

Moves the config and warnings tab content of the Zone Control Plane detail view into the overview tab.

## Screenshots

**Before**:

image

**After**:

image

### DIFF
--- a/src/app/zones/components/ZoneEgressDetails.vue
+++ b/src/app/zones/components/ZoneEgressDetails.vue
@@ -3,13 +3,23 @@
     <template #overview>
       <KCard>
         <template #body>
-          <DefinitionList>
-            <DefinitionListItem
-              v-for="(value, property) in processedZoneEgressOverview"
-              :key="property"
-              :term="t(`http.api.property.${property}`)"
-            >
-              <template v-if="property === 'name'">
+          <div class="variable-columns">
+            <DefinitionCard>
+              <template #title>
+                {{ t('http.api.property.status') }}
+              </template>
+
+              <template #body>
+                <StatusBadge :status="status" />
+              </template>
+            </DefinitionCard>
+
+            <DefinitionCard>
+              <template #title>
+                {{ t('http.api.property.name') }}
+              </template>
+
+              <template #body>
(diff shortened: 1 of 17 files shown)


src/app/zones/components/ZoneEgressDetails.vue
```diff
@@ -85,38 +101,39 @@ import { computed, PropType } from 'vue'
...
   },
   {
     hash: '#insights',
-    title: 'Zone Egress Insights',
+    title: t('zone-egresses.routes.item.tabs.insights'),
   },
   {
     hash: '#xds-configuration',
-    title: 'XDS Configuration',
+    title: t('zone-egresses.routes.item.tabs.xds_configuration'),
```
slonka, line 128: nitpick: why do we mix `-` and `_`? shouldn't it both be `zone_egresses` and `xds_configuration` or `zone-egresses` and `xds-configuration`

src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,30 +1,48 @@
 zone-cps:
   routes:
     item:
-      title: "{name} Zone Control Plane"
+      title: '{name} Zone Control Plane'
```
slonka, line 4: nitpick: just to minimise changes like this, can we easily have a linter to enforce this?

slonka (approved):
just nits

src/app/zones/components/ZoneEgressDetails.vue
kleinfreund: I’m starting to only use snakecase as that will be what we’ll standardize on. However, to minimize changes, I’m not changing existing keys because that would require also changing all their uses.

src/app/zones/locales/en-us/index.yaml
kleinfreund: You’re right, we should lint this. FWIW the inconsistencies in our YAML file (aside from the casing/naming question) are fairly minimal so we aren’t hard-pressed to do so soon.

PR #1315, opened 2023-08-10
feat: updates global control plane health

### LINKED ISSUE #1346: Update overview (and mesh detail view) with online figures
### Description

Follow up work for https://github.com/kumahq/kuma-gui/issues/1145.

Once we have the right API for this, we should update those views with the online figures for Zone CPs, services, and DPPs.

kleinfreund:
Adds online figures to some of the stats shown in the global control plane health section.

Resolves #1346.

Depends on https://github.com/kumahq/kuma/pull/7775.


src/app/common/ResourceStatus.vue
```diff
@@ -30,15 +38,41 @@ const props = defineProps({
...
+
+const statusAppearance = computed(() => {
+  if (props.online !== null) {
+    const ratio = props.online / props.total
+
+    if (ratio <= 0.5) {
+      return 'danger'
+    } else if (ratio <= 0.85) {
+      return 'warning'
+    }
```
lahabana, line 54: How was this picked? To me I feel like we should show warning as soon as `props.online !== props.total`

src/app/main-overview/components/ControlPlaneDetails.vue
```diff
@@ -13,7 +13,8 @@
           style="--columns: 4;"
         >
           <ResourceStatus
-            :total="store.getters['config/getMulticlusterStatus'] ? props.zoneOverviews.length : 1"
+            :total="store.getters['config/getMulticlusterStatus'] ? props.globalInsights.zones.cps.total : 1"
+            :online="store.getters['config/getMulticlusterStatus'] ? props.globalInsights.zones.cps.online : 1"
             data-testid="zone-control-planes-status"
           >
```
lahabana, line 19: This shows zones when in standalone:
image


Can we just remove the zones entry altogether by doing `<ResourceStatus v-if="store.getters['config/getMulticlusterStatus']" ... ` ?

Also could we make the mesh list take a the page in standalone?

src/store/reducers/mesh-insights.ts
```diff
@@ -86,6 +98,7 @@ const sumVersions = (curr: DpVersions = { kumaDp: {}, envoy: {} }, next: DpVersi
 })
 
 export function mergeInsightsReducer(insights: MeshInsight[]): MergedMeshInsights {
+  console.log(insights)
```
lahabana, line 101: left over console.log

src/app/main-overview/components/MainOverview.vue
```diff
@@ -1,36 +1,44 @@
 <template>
   <DataSource
-    v-slot="{ data: meshInsightsData, isLoading: meshInsightsIsLoading, error: meshInsightsError }: MeshInsightCollectionSource"
-    src="/all-mesh-insights"
+    v-slot="{ data, isLoading, error }: GlobalInsightsSource"
+    src="/global-insights"
   >
```
lahabana, line 5: Did we say that in the future this will become a single call to `/global-insights` ?

src/app/common/ResourceStatus.vue
kleinfreund: Arbitrarily (and I don’t remember). I changed that to check for `ratio < 1` now.

src/app/main-overview/components/ControlPlaneDetails.vue
kleinfreund: > Can we just remove the zones entry altogether by doing `<ResourceStatus v-if="store.getters['config/getMulticlusterStatus']" ... ` ?

I’ve changed this.

> Also could we make the mesh list take a the page in standalone?

Fixed.

src/store/reducers/mesh-insights.ts
kleinfreund: Fixed.

src/app/main-overview/components/MainOverview.vue
kleinfreund: With what we show on this page, we definitely need all three sources. Global insights is the data source for the top card containing the online figures. Zone insights and mesh insights are for the table respectively. Note how for both, we only request at most 10 (i.e. the `size` query parameter is `10`).

lahabana: Yeah I agree. What I was hoping is that all 3 Datasource could be declared close to their usage and then all 3 requests would be done at once.

kleinfreund: They will be done in parallel still. Nesting DataSources like that doesn’t mean they are processed in series. That would only be the case if we enforce that via `v-if` conditions on them.

lahabana: Is that true for the rendering as well? And do you not think that would improve readability?

kleinfreund: > Is that true for the rendering as well?

Pretty much, yes.

> And do you not think that would improve readability?

To me it doesn’t really help with readability moving the DataSources closer to where their data is used, no. I find it rather a distraction.

Regardless, I’ve made a change that does so in https://github.com/kumahq/kuma-gui/pull/1315/commits/24df7dd2f0e1811858ca80b6e7aa1c05fc19042a.

lahabana (approved):
Unrelated to this PR but isn't this allignment wrong:

image


?

lahabana:
Yes

johncowen (commented):
Looks good, thanks for filling out out the mocks with randomness! I left a couple of comments one was a q, lmk other than that should be a 👍 from me.

src/app/main-overview/components/MainOverview.vue
```diff
@@ -1,41 +1,215 @@
...
+                <div class="card-title">
+                  <h2>{{ t('main-overview.detail.zone_control_planes.title') }}</h2>
+
+                  <RouterLink :to="{ name: 'zone-cp-list-view' }">
+                    {{ t('main-overview.detail.health.view_all') }}
+                  </RouterLink>
+                </div>
+
+                <div
+                  v-if="env('KUMA_ZONE_CREATION_FLOW') === 'enabled' && zoneOverviewsData.items.length > 0"
```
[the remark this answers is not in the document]
kleinfreund: I’ve replaced this and the other three instances of it along with it.

kleinfreund:
Also removed fetchAllResources as we are no longer using it. https://github.com/kumahq/kuma-gui/pull/1315/commits/7a7d6355123e62274c661167e055c3214d6dea33

src/utilities/helpers.ts
```diff
@@ -1,41 +1,7 @@
...
 
-export async function fetchAllResources<T = Object>(endpoint: (params: Object) => Promise<PaginatedApiListResponse<T>>): Promise<{ items: T[]; total: number }> {
-  try {
-    let allTotal = null
-    let offset = 0
-    let allItems: T[] = []
-    let previousNext: string | null = null
-
-    while (true) {
-      const params = { size: PAGE_REQUEST_SIZE_DEFAULT, offset }
```
johncowen, line 14: I think this constant from can be removed the constants file also, up to you if you wanted to do it here or not. I reckon that entire file will be deleted soon anyway. One sec I think I can approve this now with the changes you've made.

johncowen (approved):
Much nicer!

PR #1316, opened 2023-08-10
feat: implements new DPP detail view design

kleinfreund:
## Changes

**feat: implements new insights tab design** (specifically overview and insights, but not policies)

Implements the new design for the insights tab of the zone and DPP detail views.

**feat: implements new DPP detail view design**

Fixes a bug where some data in DPP views was calculated based on the first instead of the last subscription.

## Screenshots

**Before**:

image

image

image

**After**:

image

image

image


features/mesh/dataplanes/DetailViewContent.feature
```diff
@@ -0,0 +1,101 @@
...
+    And the environment
+      """
+      KUMA_SUBSCRIPTION_COUNT: 2
+      """
+    And the URL "/config" responds with
+      """
+      body:
+        mode: global
+      """
+    # TODO: Use exact mocking here. It’s almost impossible to control the test data without exact mocking. For example, I need to set `disconnectTime: ''` just so that I don’t risk a faker-random disconnect time appearing in the last subscription.
```
johncowen, line 24: Why don't you only set the disconnectTime to ''? You shouldn't need to set all the other data.

Just so I understand, for the test, why is it necessary to have a disconnectTime of '' on the last subscription? Is it so a warning doesn't show up?

src/app/common/subscriptions/SubscriptionDetails.vue
```diff
@@ -1,97 +1,69 @@
...
       appearance="info"
-      class="mt-4"
     >
       <template #alertIcon>
         <KIcon icon="portal" />
       </template>
 
       <template #alertMessage>
-        There are no subscription statistics for <strong>{{ details.id }}</strong>
+        {{ t('common.detail.subscriptions.no_stats', { id: props.subscription.id }) }}
```
johncowen, line 12: Nit: I'm not sure that this should go in `common`. I'm guessing this should be on the i18n for the component (ideally the route component but I don't think we are there yet). If we are not careful `common` is just going to be a massive bucket full of all sorts.

I think the fact that we further namespace this under `.subscriptions` means its not "common"

src/app/common/subscriptions/SubscriptionDetails.vue
```diff
@@ -101,32 +73,38 @@ const props = defineProps({
...
+    const { responsesSent = '0', responsesAcknowledged = '0', responsesRejected = '0' } = stat
+
+    return {
+      type,
+      responsesSent: parseInt(responsesSent),
+      responsesAcknowledged: parseInt(responsesAcknowledged),
+      responsesRejected: parseInt(responsesRejected),
+    }
+  })
+})
```
johncowen, line 107: Will be lovely once we have our data layer in here so we don't have to do this all the time. I was wondering why these need to be integers? I can't see any math in here, looks like we are just printing them to the page as strings?

src/app/data-planes/components/DataPlaneDetails.vue
```diff
@@ -1,81 +1,105 @@
 <template>
-  <TabsWidget :tabs="filteredTabs">
+  <TabsWidget :tabs="TABS">
```
johncowen, line 2: Same comment as the previous PR, should this use nested routing? (at the very least in a later PR)

src/app/zones/components/ZoneIngressDetails.vue
```diff
@@ -54,16 +54,16 @@
...
             >
               <template #accordion-header>
-                <SubscriptionHeader :details="value" />
+                <SubscriptionHeader :subscription="subscription" />
               </template>
 
               <template #accordion-content>
                 <SubscriptionDetails
-                  :details="value"
+                  :subscription="subscription"
```
johncowen, line 66: nit:

Wondering why sometimes these are:

`:subscription="subscription"`, sometimes `:subscription="description"` and sometimes `:subscription="insight"`.

Might it be better to use something more generic like:

`:subscription="item"`

The `item` is then described by the name of the attribute, in this case a subscription

src/test-support/mocks/src/meshes/_/dataplanes/_.ts
```diff
@@ -31,6 +31,9 @@ export default ({ fake }: EndpointDependencies): MockResponder => (req) => {
         }),
         inbound: [
           {
+            health: {
+              ready: fake.datatype.boolean(),
+            },
```
johncowen, line 36: Am I right in thinking that sometimes this isn't set? If so this should only randomly be set.

features/mesh/dataplanes/DetailViewContent.feature
kleinfreund: Having `disconnectTime` set to `''` is a hack that avoids the default mock data randomly setting it. It ensures that I can control that this DPP always shows as offline.

I also need pretty much all the data here as it’s under test.

src/app/common/subscriptions/SubscriptionDetails.vue
kleinfreund: But it’s also not specific to this module as both zone and DPP views use this component.

src/app/common/subscriptions/SubscriptionDetails.vue
kleinfreund: Ah, that’s fair enough. I changed this to keep strings here (though not sure why they’re strings in the API to begin with).

features/mesh/dataplanes/DetailViewContent.feature
johncowen: Ah ok I see, so you are setting the subscription count to 2 to only get 2 subscriptions and then mocking out the final `disconnectTime` to be empty. Should it be null or undefined? Is that why you are saying it is a hack?

kleinfreund: I’m saying it’s a hack because what I would want is there to _not_ be a `disconnectTime` field to begin with because that’s the actual application scenario I want to test. We omit the field rather than it having a null value.

So, what I want is to leave it off and not have the mock merging mechanism populate it. That’s a really good example for when exact mocking is much, much easier to deal with.

src/app/common/subscriptions/SubscriptionDetails.vue
johncowen: Right so if should go on the component i18n keys

i.e. `components.subscription_details.stuff`. 

The larger question about both zone and DPP views using this component, to me it would make sense that this component should live in the dataplane module and reused from zones, but lets leave this for now, we can circle back later.

Feel free to resolve.

features/mesh/dataplanes/DetailViewContent.feature
johncowen: Right so you are saying it should be undefined? Not null or an empty ''?

kleinfreund: No, the field should be _absent_. Also, I shouldn’t need to know the default condition of the mock If I’m setting an array.

src/test-support/mocks/src/meshes/_/dataplanes/_.ts
kleinfreund: I’m not sure. I _think_ `ready` can indeed be absent, but I’m not sure `health` can.

features/mesh/dataplanes/DetailViewContent.feature
johncowen: > Also, I shouldn’t need to know the default condition of the mock If I’m setting an array.

No you shouldn't. That is why you should mock everything you test for instead of just relying on the default state like we used to. This is why the mocks try to give you random data. so you cannot depend on that.

src/test-support/mocks/src/meshes/_/dataplanes/_.ts
johncowen: Right yeah, we should make sure we provide mocks that replicate that

features/mesh/dataplanes/DetailViewContent.feature
kleinfreund: Yeah, but that’s precisely what’s not possible with the mock merging setup we have.

I cannot express "do not add a field to this object or array that’s not in my mock". I can only set it to something that has the same effect, but that’s hacky because it doesn’t reflect how our application actually works.

johncowen: We either need to add a way to set a field as undefined and therefore remove it, we should be able to add a YAML tag to do that.

 I think as it stands now you could set the count to 1 and then mock out 2 items, the last one having only the fields you want, which from what I remember from the code should also work.

It sounds like either way there are a few options here.

src/test-support/mocks/src/meshes/_/dataplanes/_.ts
kleinfreund: I’ve updated the mocking of inbounds so they have the same source and randomized this part.

johncowen (approved):

PR #1334, opened 2023-08-14
feature: Adds `can` utility for querying user access/ability.

johncowen:
Adds a `can` utility for querying user access (potentially based on permissions) or user abilities, i.e. can a user `use zones`.

---

- `can` can only ever return a boolean (unlike `env` which should only ever return strings)
- `can` can centralize these sorts of decisions within the 'module' meaning we don't mistakenly have these sorts of decisions made differently in different places by accident.
- `can` can have "features" injected from The Outside and "features" are composable so we can switch things in and out if we need to.

We purposefully return `false` if a "feature" is not defined i.e. we don't throw an error. Therefore identifiers are purposefully arbitrary strings and human readable/declarative-like. We should stick to `use/create/read/write/update/delete moduleName` semantics i.e. `can('delete meshes')`. This means as we now have `t` for printing human strings (using centralized i18n files), `can` for making decisions (using centralized feature files) and DataSource for accessing external data (using centralized sources files). As a sidenote `env` can be used a lot less within the application and reserved for actual environment variable usecases (of which we have very few), I guess we will still use env quite a bit within the `can` features.

Important note: We general use service injection to do these sorts of forks and prefer that approach over using conditional forks deep in the codebase. We should continue to prefer this, but! There are places/times where we may want to use this approach either temporarily or based on data that we only know at run-time, therefore:

```html
<DataSource
  v-slot="{data}"
>
<div v-if="can('use thing', data)"></div>
</DataSource>
```
```javascript
'use thing' => (can, data) => {
  return env('KUMA_THINGS') && data.permissions.for.user && data.enabled.at.runtime
}

```

The above may depend on an env var plus something that is only known via an API request.

I've included one instance of usage within the application to give an example of the usecase plus usage. We don't currently use service injection for many zone related things.

`can` will mostly be used in Route components (and its decision passed down into other components), hence its added as an export of `RouteView`, but it can also be injected into places like routing config if required.

### DIFF
--- a/src/app/zones/features.ts
+++ b/src/app/zones/features.ts
@@ -1,7 +1,7 @@
-import type { Can } from '@/app/application/services/can'
+import type { Can, Features } from '@/app/application/services/can'
 import type Env from '@/services/env/Env'
-export const features = (env: Env['var']) => {
-  const features = {
+export const features = (env: Env['var']): Features => {
+  return {
     'use zones': () => {
       return env('KUMA_MODE') === 'global'
     },
@@ -9,5 +9,4 @@ export const features = (env: Env['var']) => {
       return can('use zones') && env('KUMA_ZONE_CREATION_FLOW') === 'enabled'
     },
   }
-  return features
 }
--- a/src/app/zones/features.ts
+++ b/src/app/zones/features.ts
@@ -1,0 +1,13 @@
+import type { Can } from '@/app/application/services/can'
+import type Env from '@/services/env/Env'
+export const features = (env: Env['var']) => {
+  const features = {
+    'use zones': () => {
+      return env('KUMA_MODE') === 'global'
+    },
+    'create zones': (can: Can) => {
+      return can('use zones') && env('KUMA_ZONE_CREATION_FLOW') === 'enabled'
+    },
+  }
+  return features
+}
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -14,6 +14,7 @@
       name="default"
       :t="t"
       :env="env"
+      :can="can"
       :route="{
         update: (params: Record<string, string | undefined>) => {
           router.replace({
@@ -39,6 +40,7 @@ import { provide, inject, ref, watch, onBeforeUnmount } from 'vue'
 import { useRoute, useRouter } from 'vue-router'
 
 import { ROUTE_VIEW_PARENT } from '.'
+import { useCan } from '../../index'
 import {
   urlParam,
   cleanQuery,
@@ -48,6 +50,7 @@ import {
 import { useI18n, useEnv } from '@/utilities'
 
 const env = useEnv()
+const can = useCan()
 const { t } = useI18n()
 const route = useRoute()
 const router = useRouter()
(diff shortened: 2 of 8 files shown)


src/app/zones/features.ts
```diff
@@ -0,0 +1,12 @@
+import type { Can, Features } from '@/app/application/services/can'
+import type Env from '@/services/env/Env'
```
kleinfreund, line 2: **Suggestion**: I think we should make this a more type-safe so it’s not possible to introduce `can` checks against features we don’t know about/don’t define (e.g. through typos).

This can be accomplished like that (though it would be nicer to derive the feature keys from the features.ts files; regardless it’s an improvement):

```suggestion
type FeatureKey =
  'create zones' |
  'use zones'
export type Can = (str: FeatureKey, obj?: any) => boolean
export type Features = Record<FeatureKey, (can: Can, obj?: any) => boolean>
```

johncowen: > it’s not possible to introduce can checks against features we don’t know about/don’t define (e.g. through typos)

This is a feature not a bug (the unknown features not the typo bit). i.e. you should be able to introduce features we don't know about/don't define, and these default to `false` i.e. the thing you are guarding doesn't show.

Whilst I agree it would be nice to protect against typos I don't think your suggestion will work with the composable/modular nature of this (the type definition you've suggested here I don't think applies to the `can` function that we use?), so I'm happy to leave that as something we can figure a solution for later. That being said I am looking at a potential solution for this as we speak, but not saying that will happen in this PR. Maybe think of these identifiers like route names in the meantime which are also prone to typos

All in all if you use `can` I'm sure you'd check what you have written works and/or maybe provide a test for it?

kleinfreund: > the type definition you've suggested here I don't think applies to the `can` function that we use?

I tried it and verified it’s working locally so I would think it does apply.

> This is a feature not a bug (the unknown features not the typo bit). i.e. you should be able to introduce features we don't know about/don't define, and these default to false i.e. the thing you are guarding doesn't show.

The unknown features part of your comment is handled by this suggestion as well: You can update the type and implement a feature function that just returns `true`.

> That being said I am looking at a potential solution for this as we speak, but not saying that will happen in this PR. Maybe think of these identifiers like route names in the meantime which are also prone to typos
> 
> All in all if you use `can` I'm sure you'd check what you have written works and/or maybe provide a test for it?

But for the `can` service, it’s easily done so we don’t even need to rely on us double-checking the spelling. This would be harder (as in a lot of work) for route names for example.

johncowen: > I tried it and verified it’s working locally so I would think it does apply.

Oh I see, sorry I was looking at the wrong file.

Ok so doing this here means we can't inject features from elsewhere, so its a no go.

> The unknown features part of your comment is handled by this suggestion as well: You can update the type and implement a feature function that just returns true.

Well yes you could, (returning false)

> it’s easily done

See above comment on injection, the can service shouldn't be aware of the individual modules.

kleinfreund: Alright then. I think it would be a significant improvement for the cost of not upholding the ideal that the keywords (mind you, they’re just a type, not JS values) are defined in their respective modules.

If you can think of a way to make this work, I think that would be super helpful.

kleinfreund (approved):

src/app/zones/features.ts
johncowen: > the cost of not upholding the ideal

Its not an "ideal", its a necessity, we have other modules elsewhere as you know, hence we can't add them here. "Ideally" we'd have typo protection, but just make sure you check what you build.

> I think that would be super helpful.

Agree

PR #1339, opened 2023-08-15
chore: replaces css custom properties import

kleinfreund:
## Changes

chore: adds @kong/design-tokens Sass variables import

Replaces the import of custom CSS properties with one for Sass variables. We should be using those instead.

chore: addresses Sass @use error

As a result of adding an import to all Sass files, this error pops up in one component that uses a Sass `@use` statement which we unfortunately need to work around like this.

chore: adds deprecated utility classes lint rule

Adds an ESLint rule to disallow the use of Kongponents utility classes. It comments out those entries for which we define our own classes.


dev-utilities/deprecatedUtilityClasses.js
```diff
@@ -0,0 +1,296 @@
...
+  'style-body-md',
+  'style-body-md-bold',
+  'style-body-sm',
+  'style-body-sm-bold',
+  'style-body-tiny',
+  'style-body-link',
+  'style-body-bc',
+  'style-body-code',
+
+  /* We define the following classes.
```
slonka, line 90: why is this commented out? should we just remove it then?

dev-utilities/hoistUseStatements.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Moves `@use` statements before data injected via vite’s `css.preprocessorOptions.scss.additionalData` option which avoids the following Sass error:
+ *
+ * ```
+ * @use rules must be written before any other rules.
+ * ```
+ *
+ * Source: https://github.com/vitejs/vite/discussions/6361#discussioncomment-6577205
```
slonka, line 8: do we have an issue to track this? so we can remove this hack when it's fixed upstream?

slonka (approved):

dev-utilities/deprecatedUtilityClasses.js
kleinfreund: I wanted to keep the list complete for reference. Once Kongponents is updated to its next major version, these utility classes will be removed. Once we updated Kongponents, we will also remove the added linter and those rules again as there is no benefit in disallowing them then anymore.

In short: this will be short-lived.

dev-utilities/hoistUseStatements.ts
kleinfreund: This isn’t really a bug though.

From the perspective of Sass, it’s probably a reasonable design decision to force `@use` statements to be the first in a file.

From the perspective of a build tool, it’s reasonable to not mess with preprocessing other than giving users the option to alter files on their own responsibility (i.e. what we’re making use of here).

So, this is an unfortunate compromise. The alternative would be to not prefix all Sass files with the variables import but that would mean that we would need to manually import the variables in all our `<style lang="scss">` blocks. A hassle I’d like to avoid.

PR #1341, opened 2023-08-15
fix: browser navigation through history

kleinfreund:
## Changes

Fixes a few issues related to browser navigation through history. Specifically, it implements a hack around a shortcoming of KTable to observe external page number changes.

## Steps to reproduce

In the UI, navigate to a page with a list view and use its pagination. Then use the browser’s back button to navigate back. This will behave incorrectly (i.e. it will skip a page and navigate back two steps)

### DIFF
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -97,6 +97,8 @@ const cacheKey = ref<number>(0)
  * This is a hack around the fact that there is no other way to tell KTable if the current page of a paginated view has changed. KTable assumes it’s the *sole* owner/maintainer of pagination-related state, but that’s a design flaw that doesn’t account for browser history navigation (e.g. pressing back to go from page 2 to page 1).
  *
  * Is triggered via a watcher whenever `props.pageNumber` changes while the last page number that was emitted by KTable (via calls to its `fetcher` prop) is different (i.e. the page number has changed independently of a KTable mechanism).
+ *
+ * TODO: If https://github.com/Kong/kongponents/pull/1631 is accepted, this hack can be removed in favor of setting the new prop whenever the page number changes externally.
  */
 const kTableMountKey = ref(0)
 const lastPageNumber = ref(props.pageNumber)
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -1,5 +1,6 @@
 <template>
   <KTable
+    :key="kTableMountKey"
     class="app-collection"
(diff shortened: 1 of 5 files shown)


src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -89,6 +91,15 @@ const slots = useSlots()
...
 const cacheKey = ref<number>(0)
+/**
+ * Used as a means to instruct KTable to re-mount.
+ *
+ * This is a hack around the fact that there is no other way to tell KTable if the current page of a paginated view has changed. KTable assumes it’s the *sole* owner/maintainer of pagination-related state, but that’s a design flaw that doesn’t account for browser history navigation (e.g. pressing back to go from page 2 to page 1).
+ *
+ * Is triggered via a watcher whenever `props.pageNumber` changes while the last page number that was emitted by KTable (via calls to its `fetcher` prop) is different (i.e. the page number has changed independently of a KTable mechanism).
+ */
+const kTableMountKey = ref(0)
+const lastPageNumber = ref(props.pageNumber)
```
kleinfreund, line 102: **Comment**: Personally, I think the better solution to this would be supporting a new `props.pageNumber` in KTable. Adding a watcher to it and updating the internal `page` variable whenever it is set _to a different value than `page`_ would trigger the same mechanisms attached to using the built-in pagination and consequently trigger `props.fetcher` with the correct parameters.

KTable is currently the sole owner of all matters of pagination and I think that’s a design flaw that should be considered a bug rather than a lack of a feature as browser history navigation is a core browser feature that it currently can’t handle (except for via the hack employed here).

slonka: so why don't we just fix it in KTable?

slonka (commented):
Not approving because I think this should be fixed in KTable

src/app/application/components/app-collection/AppCollection.vue
slonka: @kleinfreund to link a PR here until it's merged we will use this workaround

slonka (approved):

kleinfreund:
@slonka I’ve updated the in-code comment with the link in https://github.com/kumahq/kuma-gui/pull/1341/commits/35eafe0a6a1678ecf3d71d99f0a9ff61821076cc

src/app/application/components/app-collection/AppCollection.vue
kleinfreund: I’ve updated the in-code comment with the link in https://github.com/kumahq/kuma-gui/commit/35eafe0a6a1678ecf3d71d99f0a9ff61821076cc

PR #1342, opened 2023-08-16
feat: implements new mesh detail view design

kleinfreund:
## Changes

Implements the new mesh detail view design.

## Screenshots

**Before**:

image

**After**:

image


src/app/meshes/components/MeshDetails.vue
```diff
@@ -2,95 +2,156 @@
...
-                <TextWithCopyButton :text="value" />
+              <template #title>
+                {{ t('meshes.detail.services') }}
               </template>
+            </ResourceStatus>
 
-              <template v-else>
-                {{ value }}
+            <ResourceStatus
+              :total="props.meshInsight.dataplanesByType.standard.total ?? 0"
```
slonka, line 41: can any of these be null when a dp type does not exist (e.g. gateway)? should we do optional chaining?

src/app/meshes/components/MeshDetails.vue
```diff
@@ -127,53 +185,26 @@ const props = defineProps({
...
-
-      return null
-    })
-    .filter(notEmpty)
-})
 
-function getBackendData(mesh: Mesh, field: 'mtls' | 'logging' | 'metrics' | 'tracing') {
-  if (mesh[field] === undefined) {
+function getBackendData(meshBackend?: MeshBackend) {
+  if (!meshBackend || !Array.isArray(meshBackend.backends) || meshBackend.backends.length === 0) {
```
slonka, line 198: nit: we could extract it to `meshBackendEmptyOrNull` just like couple of lines below for enabledBackendName.

src/app/meshes/components/MeshDetails.vue
```diff
@@ -127,53 +185,26 @@ const props = defineProps({
...
-      }
-
-      return null
-    })
-    .filter(notEmpty)
-})
 
-function getBackendData(mesh: Mesh, field: 'mtls' | 'logging' | 'metrics' | 'tracing') {
-  if (mesh[field] === undefined) {
+function getBackendData(meshBackend?: MeshBackend) {
```
slonka, line 197: why don't we have return type here? and if it's really data if it retuns `type / name`?

slonka (approved):
mostly nitpicks

src/app/meshes/components/MeshDetails.vue
kleinfreund: Yes. Specifically, `total` can be absent here. So while `dataplanesByType` always exists and always has the properties `standard` and `gateway`, these objects may be completely empty.

src/app/meshes/components/MeshDetails.vue
kleinfreund: If it’s not a big improvement in readability I prefer to leave this in the if statement because the evaluation can then short-circuit on the first truthy result. Depending on how this code will be minified (i.e. will the expression be inlined because it’s clearly only in function scope and thus safe to do?), this won’t be possible.

src/app/meshes/components/MeshDetails.vue
kleinfreund: I’ve changed both aspects now. There’s some value here in having the return type be annotated indeed.

src/app/meshes/components/MeshDetails.vue
kleinfreund: I misread your question. You’re asking if `gateway` and/or `standard` can be absent, and I don’t think they can.

PR #1345, opened 2023-08-16
refactor: moves service loading complexity into source

kleinfreund:


src/app/services/views/ServiceDetailView.vue
```diff
@@ -39,125 +39,109 @@
...
-          v-slot="{ data: externalService, isLoading: isLoadingExternalService, error: externalServiceError }: ExternalServiceSource"
-          :should-make-request="data.serviceType === 'external'"
-          :src="`/meshes/${route.params.mesh}/external-services/${route.params.service}`"
+          :tabs="TABS.filter((tab) => {
+            if (tab.hash === '#overview') {
+              return true
+            } else {
+              return data.serviceInsight.serviceType !== 'external'
+            }
+          })"
```
lahabana, line 50: Does it make sense to inlide TABS here? I find it would be easier to read:

```typescript
:tabs="(() => {
  var r = [{
    hash: '#overview',
    title: t('services.routes.item.tabs.overview'),
  }]
  if (data.serviceInsights.serviceType !== 'external') {
    r.push({
      hash: '#dataPlaneProxies',
      title: t('services.routes.item.tabs.data_plane_proxies'),
    })
  }
  return r
})()"
```

lahabana (commented):
Looking at this some more it seems like moving the datasource into (or around) the ServiceSummary would make more sense no? It's the only place where the data is needed and the goal of datasource is to get data close to where it's used.

Also as summary is likely to always diverge more and more between gateways/external and internal services would it make sense to just have a dedicated component for each?

kleinfreund:
> Looking at this some more it seems like moving the datasource into (or around) the ServiceSummary would make more sense no? It's the only place where the data is needed and the goal of datasource is to get data close to where it's used.

This DataSource can’t be moved further in because the result is used for (1) determining which tabs to show and (2) whether to render the content related to the DPP table (and thus preventing it making an API call when not needed)

kleinfreund:
> Also as summary is likely to always diverge more and more between gateways/external and internal services would it make sense to just have a dedicated component for each?

Yeah, I agree. I can handle this in a follow up at some point.

src/app/services/views/ServiceDetailView.vue
kleinfreund: I really don’t like this style of inlining logic in the template. I would actually rather move the filtering I added here into the script and have something like this:

```ts
function getTabs(serviceInsight: ServiceInsight) {
  return [
    {
      hash: '#overview',
      title: t('services.routes.item.tabs.overview'),
    },
    {
      hash: '#dataPlaneProxies',
      title: t('services.routes.item.tabs.data_plane_proxies'),
    },
  ].filter((tab) => {
    if (tab.hash === '#overview') {
      return true
    } else {
      return serviceInsight.serviceType !== 'external'
    }
  })
}
```

There are cases when it’s just very convenient to inline a simple expression, but generally, it makes templates much harder to read for me.

lahabana: Abstract it as a function if you want. However, doing a filter rather than just building the right array is just making things harder to read.

kleinfreund:
> I don't think you need the externalService data to figure if you need to show the tab or not. You only need the serviceType which is part of service Insights. So to me looks like the datasource for fetching the externalService could be around the template (or even this new component we're talking about).

I think [the initial suggestion of combining the data source for ServiceInsight and ExternalService](https://github.com/kumahq/kuma-gui/pull/1287#discussion_r1295682832) makes more sense to me. I could split them up again and indeed move only the ExternalService one and have it wrap solely around the summary component, but managing error and loading states is actually easier like this now.

src/app/services/views/ServiceDetailView.vue
kleinfreund: I’ve moved this logic into a function and turned into direct array construction here: https://github.com/kumahq/kuma-gui/pull/1345/commits/dd7c78ac21d3762bc64260656ec4a86792302371

lahabana:
But then you'll have to undo it as soon as you have gateways and other things coming into play no?

lahabana:
When you look at service that's a gateway we'd likely want to show listeners and routes in the future more datasources that you'll need to handle errors etc for.

kleinfreund:
I’ve decided to split the ServiceSummary component into two in https://github.com/kumahq/kuma-gui/pull/1345/commits/de7ffd221d680e245cea497124a993b6277e63ce.

With this, I also separated the data source for ServiceInsight and ExternalService again and move the latter source around its dedicated component.

src/app/services/views/ServiceDetailView.vue
```diff
@@ -41,17 +41,38 @@
...
+              :src="`/meshes/${route.params.mesh}/external-services/${route.params.service}`"
+            >
+              <LoadingBlock v-if="externalServiceIsLoading" />
+
+              <ErrorBlock
+                v-else-if="externalServiceError"
+                :error="externalServiceError"
+              />
+
+              <EmptyBlock v-else-if="externalService === undefined" />
```
lahabana, line 59: Is that even needed? Seems like it can't happen with the above check no?

src/app/services/components/ExternalServiceDetails.vue
```diff
@@ -0,0 +1,97 @@
...
+            <template #title>
+              {{ t('http.api.property.address') }}
+            </template>
+
+            <template #body>
+              {{ props.externalService.networking.address }}
+            </template>
+          </DefinitionCard>
+
+          <DefinitionCard>
```
lahabana, line 41: I don't think this is useful to show

src/app/services/views/ServiceDetailView.vue
kleinfreund: It is, currently.

It’s not the ideal way of handling this. A better approach would be turning the `data` object into an object following the following type spec:

```ts
interface DataSourceLoaded<T> {
  status: 'loaded'
  value: T
}

interface DataSourceLoading<T> {
  status: 'loading'
  value: undefined
}

interface DataSourceError<T> {
  status: 'loading'
  value: ApiError
}
```

And defining `data` as a union of those interfaces.

Then, `LoadingBlock` would be used if `data.status === 'loading'` and `ErrorBlock` would be used if `data.status === 'error'` (at which point TS would know `value` can be used as an instance of `ApiError`).

`EmptyBlock` wouldn’t come into play here because after checking for the loading and error status, we would know `value` is of type `T` (here `ExternalService`). For views where that makes sense (and we don’t want to rely on something else like KTable’s own empty state), we can use some custom check and `EmptyBlock`.

src/app/services/components/ServiceInsightDetails.vue
```diff
@@ -0,0 +1,80 @@
+<template>
+  <div class="stack">
+    <KCard>
+      <template #body>
+        <div
+          class="columns"
+          style="--columns: 4;"
+        >
```
lahabana, line 8: On a 15" screen one of the column moves to the second row:

image

It's strange as it looks like there's a lot of real estate to put everything on the same row. After more digging it seems like the sidebar on Kuma is 280px whereas it's 240px on Konnect. I'm going to submit a PR to fix this as it also seems to solve this line break.

src/app/services/locales/en-us/index.yaml
```diff
@@ -8,3 +8,6 @@ services:
         data_plane_proxies: 'Data Plane Proxies'
     items:
       title: Services
+  detail:
+    enabled: 'Enabled'
+    disabled: 'Disabled'
```
lahabana, line 13: I think these go away if you remove the TLS from externalServiceDetails

src/app/services/views/ServiceDetailView.vue
```diff
@@ -39,125 +39,124 @@
...
+                      data-testid="data-plane-collection"
+                      class="data-plane-collection"
+                      :page-number="props.page"
+                      :page-size="props.size"
+                      :total="dataplanesData?.total"
+                      :items="dataplanesData?.items"
+                      :error="dataplanesError"
+                      :gateways="gateways"
+                      @change="({page, size}) => {
+                        // @TODO: Should we remove s: undefined?
```
lahabana, line 98: What is this TODO for? Should it be removed?

lahabana (commented):
Some nits but looking good!

src/app/services/components/ServiceInsightDetails.vue
lahabana: https://github.com/kumahq/kuma-gui/pull/1354

src/app/services/views/ServiceDetailView.vue
kleinfreund: Alright, I’m changing this. I’m not a fan of conflating the data-carrying variable to also imply the loading state even though it happens to be enough.

src/app/services/views/ServiceDetailView.vue
kleinfreund: I don’t know.

src/app/services/components/ExternalServiceDetails.vue
kleinfreund: Removed it.

src/app/services/locales/en-us/index.yaml
kleinfreund: Removed it.

src/app/services/views/ServiceDetailView.vue
lahabana: @johncowen also added more info about the motivation of this recently: https://github.com/kumahq/kuma-gui/pull/1284#discussion_r1298149625

johncowen: > Didn't @johncowen point out in another PR that up until now the loading state was handled as data === undefined? Seems like this would solve this ambiguity without introducing further types.

Yes definitely, kind of half catching up here so not entirely read right through this looked at this code 😅 . But I think we should keep things as simple as possible, we should just be implementing all the detail views using DataSource as it stands and as we did for the list views.

The codebase is generally looking a lot simpler, welcoming and approachable now we have added DataSource and inlined a lot of the simpler code. I want to avoid overcomplicating things where we don't need to.

src/app/services/components/ExternalServiceDetails.vue
```diff
@@ -0,0 +1,87 @@
...
+        </div>
+      </template>
+    </KCard>
+
+    <ResourceCodeBlock
+      id="code-block-service"
+      :resource="props.externalService"
+      :resource-fetcher="fetchExternalService"
+      is-searchable
+      code-max-height="250px"
```
lahabana, line 59: Why have a max height btw?

src/app/services/views/ServiceDetailView.vue
lahabana: Should we just remove it then?

src/app/services/components/ExternalServiceDetails.vue
kleinfreund: I’m pretty sure we added this when the code block used to be next to the details. Now that it’s separate, this should actually be removed. Removing it.

src/app/services/views/ServiceDetailView.vue
kleinfreund: @johncowen Fine by you? We have this in three places. Could we just do `@change="route.update"` like in other places?

johncowen: Yeah sure fine by me to remove this comment.

lahabana (approved):

PR #1349, opened 2023-08-17
chore(e2e): fixes flaky dataplanes index test

kleinfreund:
Fixes a test that was flaky due to incomplete mock data.

### DIFF
--- a/features/mesh/dataplanes/Index.feature
+++ b/features/mesh/dataplanes/Index.feature
@@ -25,11 +25,21 @@ Feature: mesh / dataplanes / index
             subscriptions:
             - status:
                 lastUpdateTime: 2021-02-16T08:33:36.442044+01:00
+                total: {}
+                cds: {}
+                eds: {}
+                lds: {}
+                rds: {}
               version:
                 kumaDp:
                   version: 1.0.7
             - status:
                 lastUpdateTime: 2021-02-18T08:33:36.442044+01:00
+                total: {}
+                cds: {}
+                eds: {}
+                lds: {}
+                rds: {}
               version:
                 kumaDp:
                   version: 1.0.8


kleinfreund:
> Could you give some more detail into how overwriting populated data with empty data makes a test less flakey, I don't see in the test where we are depending on the data being empty?

It’s strange that this test is flaky and not constantly flaky because I verified locally that this test definitely fails on the absence of this data and the source is this mock.

johncowen:
>  I verified locally that this test definitely fails on the absence of this data

So the test fails/can flake if you don't overwrite these values with empty data?

kleinfreund:
Yes. To be exact: what I added here is the minimum to prevent the access of data to cause a JS exception. I could’ve also set values in the objects I added.

johncowen:
And when you see a JS exception what does the data equal?

kleinfreund:
The data in the case of the test failing is exactly the data set in the test file (i.e. missing those fields).

johncowen:
So are you saying that the data we are setting in the test file is not getting merged in with the mock data?

kleinfreund:
> So are you saying that the data we are setting in the test file is not getting merged in with the mock data?

I don’t know. You can try it out.

johncowen:
Well yes I could, but that would be the actual bug if that was the case, and not only that but it it was a bug and we fixed that bug in the correct place then it would fix it for the entire project. If you are working on this, could you check please? If that is a bug I would rather fix it in the right place.

kleinfreund:
I don’t have the time for this now. If you want to fix this now, please look into it yourself. This test is really, really flaky right now and it’s more important to address the immediate issue than spend the time fixing the underlying mechanism.

Besides that, the mechanism of aggressively merging mock data within arrays has proven more and more problematic the more randomization enters our mock data.

johncowen:
> Besides that, the mechanism of aggressively merging mock data within arrays has proven more and more problematic the more randomization enters our mock data.

The "aggressive merging of mock data" means you should not have to do what you are doing here. If the merging isn't working correctly in this instance then it should be fixed in the correct place so it applies everywhere. I can't remember in the past an instance where the "aggressive merging of mock data" has been broken, and I can't remember this test ever being flakey so I am cautious that we are missing something else here. I would really like to understand what the problem actually is if you "don't know" whether the merging is happening or not.

Additionally the randomization is specifically there so you can't just depend on the dev time mocks. You have to mock out what you test for in the tests i.e. you _have to_ arrange, act, then assert. We would previously break things in dev by changing our fixtures just for tests when we should have been mocking the data out in tests correctly. The randomization prevents you from being able to do that. i.e. **if it flakes due to the randomization then there is a problem with the test mocking**.

> I don’t have the time for this now.

Not quite sure what you have on your plate right now so I can't comment on that.

johncowen:
I think this can be closed now due to https://github.com/kumahq/kuma-gui/pull/1380

PR #1350, opened 2023-08-17
refactor: loads policy type in components instead of globally

kleinfreund:
Replaces the route parameter `policyPath` with `policyType` for policy routes. This is much simpler to manage in the application because in many places, we already know the policy type name and don’t need anything else. Only in the policy views do we need to know the corresponding API path, but we can now isolate the mapping of policy type name to policy type path to those components. This in turn allows us to get rid of all store-kept policy type data and one app boot request.

Changes how navigating to the policy list view works to use an arbitrary (i.e. CircuitBreaker) policy type as the initially selected one instead of determining the first populated type on the route level.


src/app/policies/routes.ts
```diff
@@ -40,38 +40,22 @@ export const routes = (store: Store<State>) => {
...
-                })
-
-                if (item === undefined) {
-                  item = store.state.policyTypes[0]
-                }
-
-                if (item === undefined) {
-                  return { name: 'home' }
-                }
-
```
kleinfreund, line 58: **Comment**: This was the logic that made us consider to just use a more or less arbitrary policy type as the default target as we would otherwise not be able to get rid of the global policy type state. We picked CircuitBreaker as most users will have a policy of this type via the default mesh policies.

features/application/Titles.feature
```diff
@@ -38,5 +38,5 @@ Feature: The HTML title is correct on each page
       | /mesh/default/data-planes                | Data Plane Proxies |
       | /mesh/default/data-plane/data-plane-name | Data Plane Proxy   |
 
-      | /mesh/default/policies/circuit-breakers         | CircuitBreaker |
-      | /mesh/default/policy/circuit-breakers/program-0 | CircuitBreaker |
+      | /mesh/default/policies/CircuitBreaker         | CircuitBreaker |
```
johncowen, line 41: Has there been any discussion elsewhere on changing these to be camelCase?

kleinfreund: They’re changing specifically to exact name of policy _types_ rather than using the API path. That makes quite a few things easier across the application and eventually this is necessary to allow us to get rid of the global policy types state (as you would otherwise often need to map from policy type to policy path).

kleinfreund: @lahabana and I talked about this yesterday. There otherwise isn’t a more formal discussion to refer to, no.

src/app/common/PolicyTypeTag.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <span class="policy-type-tag">
     <img
-      v-if="policy.iconUrl !== null"
+      v-if="POLICY_ICON[props.policyType]"
       class="policy-type-tag-icon"
-      :src="policy.iconUrl"
+      :src="POLICY_ICON[props.policyType]"
```
lahabana, line 6: What if there isn't one?

features/application/Titles.feature
lahabana: Wait... the discussion we had yesterday was around picking the right default in the dropdown. We didn't talk about changing url paths.

lahabana (commented):
There's a thing I don't get here.
Policy type to me is like an instance of a dataplane. We don't need to define a route per type but can define a route where type is a parameter and then inside it have a dataSource to get the policyTypes and then the policies for this type. This means that you don't need to call the api to build your router and that paths can stay the same right?

Updating the url from `/policies` to `/policies/<whateverIsSelected>` is similar to what we do with query parameters for page changes right?

johncowen:
Kinda trying to play (or guess at) catchup here, but if there has been discussion on needing data in the router in order to do the redirect. You can do either one or both of the following:

- Wrap the Policies tab button itself in the DataSource, then the href on the Policies tab is correct. You could even pass this data down through the routing hierarchy so you don't have to request it again when you then click on the tab. (of course if you want super fresh data you can also just request it again when you visit the page, this will also have the benefit of updating the tab href if it has changed due to DataSource's syncing)
- The policies route `policies` could just contain the DataSources required to figure out the redirect. The first time, yes there will be a little pause whilst it gets this data, but the second time it will redirect immediately due to DataSource's caching

If you do point 1, point 2 is hardly necessary which means the little pause to figure out the redirect isn't even worth worrying about.

---

I might be missing the point of the PR so please ignore if its not the point, but I wanted to at least leave what I had planned here.

src/app/policies/views/PolicyListView.vue
```diff
@@ -11,45 +11,45 @@
...
       <EmptyBlock v-else-if="policies.policies.length === 0" />
 
-      <template
-        v-else
-      >
+      <template v-else>
         <template
-          v-for="selected in [policies.policies.find(item => item.path === route.params.policyPath) ?? policies.policies[0]]"
-          :key="selected.path"
+          v-for="currentPolicyType in [policies.policies.find((item) => item.name === route.params.policyType) ?? policies.policies[0]]"
```
johncowen, line 25: Is `currentPolicyType` the policyType from the current URL, or it is the one I've `selected`?

i.e. I'm not sure of the need to change this.

src/app/common/PolicyTypeTag.vue
kleinfreund: If there isn’t one, we don’t show an icon at all.

features/application/Titles.feature
kleinfreund: If we don’t change the URL paths, we cannot avoid the global knowledge of `/policies` because that’s what allows us to map from policy type name to policy type API path.

The whole point of that change was to avoid that dependency.

kleinfreund:
> There's a thing I don't get here.
> Policy type to me is like an instance of a dataplane. We don't need to define a route per type but can define a route where type is a parameter and then inside it have a dataSource to get the policyTypes and then the policies for this type.

That’s exactly what I did in this PR. I change the dynamic path segment in the URL to be a policy type name.

> This means that you don't need to call the api to build your router and that paths can stay the same right?

I don’t see how. In the places where we link to a policy view, we know of a policy type name but not of the corresponding API path. If the URLs have API paths as their parameter, we need to know those at the point of defining the link target (i.e. we need to map from policy type name to policy type API path right then).

src/app/policies/views/PolicyListView.vue
kleinfreund: It’s the same value with a better variable name.

lahabana:
> There's a thing I don't get here.
>Policy type to me is like an instance of a dataplane. We don't need to define a route per type but can define a route where >type is a parameter and then inside it have a dataSource to get the policyTypes and then the policies for this type.

>> That’s exactly what I did in this PR. I change the dynamic path segment in the URL to be a policy type name.

Then I don't get this:

> If we don’t change the URL paths, we cannot avoid the global knowledge of /policies because that’s what allows us to map from policy type name to policy type API path.
The whole point of that change was to avoid that dependency.

Where do you need the API path where you don't have the policy list?

kleinfreund:
> Where do you need the API path where you don't have the policy list?

Every time you want to construct the UI URL of the form `/meshes/:mesh/policies/:policyApiPath`. That’s what makes the URLs have `/circuit-breakers` in them and requires this knowledge.

lahabana:
> Every time you want to construct the UI URL of the form /meshes/:mesh/policies/:policyApiPath. That’s what makes the URLs have /circuit-breakers in them and requires this knowledge.

Do we agree that the 2 cases where this happens are:

- policy list (in which case we already have the list)
- dataplane policy matching view?

Can we just add a Datasource call in the policy matching view and have this solved? Btw how come the whole dataplane/policy list component is not using Datasource?

kleinfreund:
> Do we agree that the 2 cases where this happens are:
> 
> * policy list (in which case we already have the list)
> * dataplane policy matching view?

Yes, that’s accurate.

> Can we just add a Datasource call in the policy matching view and have this solved?

That works, yes.

> Btw how come the whole dataplane/policy list component is not using Datasource?

That’s just one of the (few) places where I haven’t gotten around to implement this, yet.

src/app/data-planes/components/DataplanePolicies.vue
```diff
@@ -221,10 +224,9 @@ function getPolicyTypeEntries(sidecarDataplanes: SidecarDataplane[]): PolicyType
       }
 
       const policyTypeEntry = policyTypeEntriesByType.get(policyTypeName) as PolicyTypeEntry
-      const policyType = store.state.policyTypesByName[policyTypeName] as PolicyType
 
       for (const policy of policies) {
-        const connections = getPolicyTypeEntryConnections(policy, policyType, sidecarDataplane, destinationTags, name)
+        const connections = getPolicyTypeEntryConnections(policy, sidecarDataplane, destinationTags, name)
```
lahabana, line 229: Why change this to now rely on global state instead?

lahabana: Especially as in the future this function won't be called as part of a watch and will just be called directly from the template and passed down the data from Datasources

kleinfreund: > Especially as in the future this function won't be called as part of a watch and will just be called directly from the template and passed down the data from Datasources

Ah, this comment just appeared.

So, you would like to make this a pure function so it can be moved more easily?

src/app/common/PolicyTypeTag.vue
```diff
@@ -35,36 +34,30 @@ import TrafficPermissionIconUrl from '@/assets/images/policies/TrafficPermission
...
+  ProxyTemplate: ProxyTemplateIconUrl,
+  RateLimit: RateLimitIconUrl,
+  Retry: RetryIconUrl,
+  Timeout: TimeoutIconUrl,
+  TrafficLog: TrafficLogIconUrl,
+  TrafficPermission: TrafficPermissionIconUrl,
+  TrafficRoute: TrafficRouteIconUrl,
+  TrafficTrace: TrafficTraceIconUrl,
+  VirtualOutbound: VirtualOutboundIconUrl,
 }
```
lahabana, line 61: In the future these icons, should  be the sort of thing we do in the layer between the API and the actual data (as introduced in https://github.com/kumahq/kuma-gui/pull/1242)?

src/app/data-planes/components/DataplanePolicies.vue
lahabana: Yeah. Moreover, I don't see how changing this makes it any better.

src/app/common/PolicyTypeTag.vue
kleinfreund: Possibly, yes. That could be attached to loading `/policies`, but the bigger question would be whether we could get a complete set of icons for the current set of policies as with the advent of new policy types especially, we’re missing quite a few.

lahabana: We'd likely have a generic icon for new and plugin types.

src/app/data-planes/components/DataplanePolicies.vue
kleinfreund: I’ve changed this back.

lahabana (approved):

features/application/Titles.feature
johncowen: Thats sounds better 👍

PR #1353, opened 2023-08-18
chore: removes lowercasing of policy docs URL

kleinfreund:


lahabana:
Hmm seems like case are ignored for redirects going to do another fix on the docs...

kleinfreund:
> Hmm seems like case are ignored for redirects going to do another fix on the docs...

I’m noticing that, too. I’m going through all variants now and making a list.

lahabana (approved):
Just checked with updated Kuma-website and it works well!

PR #1356, opened 2023-08-21
feat: changes policy type dropdown to in-page list

### LINKED ISSUE #1144: rework policy picker to not be a dropdown
### Description

The idea explored was a side list:

https://www.figma.com/file/Q5z8lPJeU26763q30S9GTy/MinK?type=design&node-id=1473-35106&mode=design&t=bPAKe928kR4PCSt0-0

kleinfreund:
## Changes

Changes the policy type dropdown to an in-page list.

Splits out the presentational part of the policy list view into its own PolicyList component.

Resolves #1144.

## Screenshots

**Before**:

image

**After**:

image


src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,253 @@
...
+
+const sortedPolicyTypeWrappers = computed(() => {
+  const policyTypeWrappers = props.policyTypes.map((policyType) => ({
+    policyType,
+    numberOfPolicies: props.meshInsight?.policies?.[policyType.name]?.total ?? 0,
+  }))
+
+  policyTypeWrappers.sort((policyTypeWrapperA, policyTypeWrapperB) => {
+    return policyTypeWrapperA.policyType.name.localeCompare(policyTypeWrapperB.policyType.name)
+  })
```
lahabana, line 207: I know this used to be the existing sorting. However, Does it make sense? Seems like this would change the order of items in the sidebar and make the UX poor as it would prevent folks from developing muscle memory no?

kleinfreund: Since this sorts by name, shouldn’t this always be a stable and predictable order of items until we introduce new policy types?

Would you prefer to not sort them here and rely on whatever order `/policies` returns?

lahabana (commented):
There are a few odd things here.

1. The actual implementation looks very different from the figma (missing top card with short desc and link to docs), Side selection shouldn't have borders, title should be just policies.
2. We seem to be requesting the list of policies and the stats on each change of policy. This seems unnecessary
3. For new policies we mentioned showing the `kind` of the top level targetRef to make it more useful (the current type column is useless as it's already selected on the left).
4. Not found show errors (I don't know if it's a mock issue) but it should show the empty state.

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,253 @@
...
+                <p>
+                  <strong>Warning</strong> This policy is experimental. If you encountered any problem please open an
+                  <a
+                    :href="env('KUMA_FEEDBACK_URL')"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                  >issue</a>
+                </p>
+              </template>
+            </KAlert>
```
lahabana, line 61: Should the instead of being an alert could be a badge next to the docs button?

kleinfreund:
This PR’s focus was to implement the change from the policy type dropdown to a sidebar.

> * The actual implementation looks very different from the figma (missing top card with short desc and link to docs),

Where do I get the descriptions from?

> Side selection shouldn't have borders

We can’t currently avoid using cards around content like that. The designs assume that the background color of the page is a plain white, but that’s not always the case. On grey background, the content of the policy type list, for example, would look really bad (barely legible grey text on grey background). That’s why I asked about the designs being done on a grey background so this would be taken into account automatically.

> title should be just policies.

I’ve changed that.

> * We seem to be requesting the list of policies and the stats on each change of policy. This seems unnecessary

I can try and think of a way to avoid this.

Both currently and in this PR, selecting a policy type is a navigation to the policy list view with the appropriate parameter representing the policy type. That will always trigger the `DataSource`s again.

> * For new policies we mentioned showing the `kind` of the top level targetRef to make it more useful (the current type column is useless as it's already selected on the left).

I have a branch prepared that implements this which is waiting for https://github.com/kumahq/kuma/issues/7485 to be implemented. We need this so we know when to show the column.

> * Not found show errors (I don't know if it's a mock issue) but it should show the empty state.

Yeah, that’s a mock issue. We haven’t mocked out the data for all policy types (especially newer types) and so landing on such a view shows (correctly) the error state because with the mock data, the request does actually result in a 404. With the real API, this wouldn’t happen unless the policy type actually doesn’t exist.

src/app/policies/components/PolicyList.vue
kleinfreund: This no longer applies sorting.

src/app/policies/components/PolicyList.vue
kleinfreund: For the length of the content, I really prefer this being a KAlert. We can move it into the toolbar so it’s next to the docs button though.

lahabana: This badge content could be just `beta`

kleinfreund: Done and moved.

lahabana:
>> The actual implementation looks very different from the figma (missing top card with short desc and link to docs),

> Where do I get the descriptions from?

For the description let's start with a generic text: 

> Use policies to apply filters to incoming or outgoing traffic of data plane proxies. To generate the Envoy configuration of a proxy the control-plane uses its data-plane configuration with the policies matching it.

Make sure there's a i18n entry so we can easily change it in the future.


Opening a PR for https://github.com/kumahq/kuma/issues/7485 in a sec

kleinfreund:
- Added the "About" section for policy types
- Added inbound and outbound badges to the "About" section (used badges for now because we already have the beta badge now)
- Pulled in the work for adding the target ref column because it will be available with https://github.com/kumahq/kuma/pull/7561
- Removed the type column
- Updated the /policies mock

kleinfreund:
If the top-level target ref has a name and its kind is `MeshService`, does that represent a service we can link to (i.e. to the service detail view)?

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+                      policy: rowValue,
+                    },
+                  }"
+                >
+                  {{ rowValue }}
+                </RouterLink>
+              </template>
+
+              <template #targetRef="{ row }">
+                <template v-if="row.spec?.targetRef">
```
lahabana, line 118:
```suggestion
                <template v-if="policyType.isTargetRefBase">
```

This looks more straight forward to me.

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+                  <KBadge appearance="neutral">
+                    {{ row.spec?.targetRef?.kind }}
+                  </KBadge>
+
+                  {{ row.spec?.targetRef?.name ?? t('common.detail.none') }}
```
lahabana, line 123: No need for all these `?` they are going to be set.
As we mentioned on the call let's have the name inside the KBadge like for tags.

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+const emit = defineEmits<{
+  (event: 'change', value: ChangeValue): void
+}>()
+
+const policyType = computed(() => props.policyTypes.find((policyType) => policyType.path === props.currentPolicyTypePath) ?? props.policyTypes[0])
+
+const sortedPolicyTypeWrappers = computed(() => props.policyTypes.map((policyType) => ({
+  policyType,
+  numberOfPolicies: props.meshInsight?.policies?.[policyType.name]?.total ?? 0,
+})))
```
lahabana, line 226: AS this computed prop is only used once I'd inline it above.  like commented previously

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
+<template>
+  <div class="policy-list-content">
+    <KCard
+      class="policy-type-list"
+      data-testid="policy-type-list"
+    >
+      <template #body>
+        <div
+          v-for="(policyTypeWrapper, index) in sortedPolicyTypeWrappers"
```
lahabana, line 9:
```suggestion
          v-for="(policyType, index) in props.policyTypes"
```

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+          :key="index"
+          class="policy-type-link-wrapper"
+          :class="{
+            'policy-type-link-wrapper--is-active': policyTypeWrapper.policyType.path === props.currentPolicyTypePath,
```
lahabana, line 13:
```suggestion
            'policy-type-link-wrapper--is-active': policyType.path === props.currentPolicyTypePath,
```

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+          }"
+        >
+          <RouterLink
+            class="policy-type-link"
+            :to="{
+              name: 'policies-list-view',
+              params: {
+                mesh: route.params.mesh,
+                policyPath: policyTypeWrapper.policyType.path,
```
lahabana, line 22:
```suggestion
                policyPath: policyType.path,
```

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+              },
+            }"
+            :data-testid="`policy-type-link-${policyTypeWrapper.policyType.name}`"
```
lahabana, line 25:
```suggestion
            :data-testid="`policy-type-link-${policyType.name}`"
```

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+          >
+            {{ policyTypeWrapper.policyType.name }}
```
lahabana, line 27:
```suggestion
            {{ policyType.name }}
```

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+  policyCollection: PolicyCollection | undefined
+  policyError: Error | undefined
+  meshInsight: MeshInsight | undefined
+}>()
+
+const emit = defineEmits<{
+  (event: 'change', value: ChangeValue): void
+}>()
+
+const policyType = computed(() => props.policyTypes.find((policyType) => policyType.path === props.currentPolicyTypePath) ?? props.policyTypes[0])
```
lahabana, line 221: As this is used in a logically hierarchical manner would it make sense to use the vue template entry: `set="policyType = props.policyTypes.find((policyType) => policyType.path === props.currentPolicyTypePath) ?? props.policyTypes[0])"`

src/app/policies/views/PolicyDetailView.vue
```diff
@@ -5,21 +5,21 @@
...
 
-      <EmptyBlock v-else-if="policiesData === undefined || policiesData.policies.length === 0" />
+      <LoadingBlock v-else-if="policyTypesData === undefined" />
+
+      <EmptyBlock v-else-if="policyTypesData.policies.length === 0" />
 
       <template v-else>
         <template
-          v-for="currentPolicyType in [policiesData.policies.find((item) => item.path === route.params.policyPath) ?? policiesData.policies[0]]"
+          v-for="currentPolicyType in [policyTypesData.policies.find((item) => item.path === route.params.policyPath) ?? policyTypesData.policies[0]]"
```
lahabana, line 22: It's odd we're using a for when there's always a single element no?

lahabana (approved):
LGTM other than the few comments pointed out

src/app/policies/components/PolicyList.vue
kleinfreund: Changed.

src/app/policies/components/PolicyList.vue
kleinfreund: Was removed with the change to include the name inside the badge.

src/app/policies/components/PolicyList.vue
kleinfreund: Done.

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,288 @@
...
+                policyPath: policyTypeWrapper.policyType.path,
+              },
+            }"
+            :data-testid="`policy-type-link-${policyTypeWrapper.policyType.name}`"
+          >
+            {{ policyTypeWrapper.policyType.name }}
+          </RouterLink>
+
+          <div class="policy-count">
+            {{ policyTypeWrapper.numberOfPolicies }}
```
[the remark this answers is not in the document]
kleinfreund: I applied the suggestion around iterating over policy types directly and will resolve the following related comments.

src/app/policies/components/PolicyList.vue
kleinfreund: I did something differently. I moved this definition into the list view when passing the prop.

src/app/policies/views/PolicyDetailView.vue
kleinfreund: Yeah, I’m really not a fan of this style of setting a variable in a template through the mechanism of the `v-for` directive. It’s basically a workaround around the fact that with DataSource, data is always scoped to templates and not available in a Vue SFC script section at all (except via function calls). This also makes it impossible to create computed properties from this data which is what one would use here.

lahabana: Can't you use the `:set` mentioned previously no?

kleinfreund: Unless I’m out of touch with Vue templating, this is not a thing. You can’t declare variables inside templates and the use of the `v-for` loop here is really a workaround around this.

src/app/policies/components/PolicyList.vue
```diff
@@ -0,0 +1,279 @@
+<template>
```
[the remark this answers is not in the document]
kleinfreund: I think that separating the data fetching from the presentational responsibilities using *View and non-*View component is very helpful in working on these components over time. On top of that, I think that inlining expressions for things like the current policy type makes the components overall harder to read.

src/app/policies/views/PolicyDetailView.vue
lahabana: I got it from: https://github.com/vuejs/core/issues/1172#issuecomment-1035939776 it sort of makes sense to me.

Looking further down this hole I also noticed: https://dev.to/loilo/an-approach-to-vuejs-template-variables-5aik which does what we do. 

> It’s basically a workaround around the fact that with DataSource, data is always scoped to templates and not available in a Vue SFC script section at all (except via function calls). This also makes it impossible to create computed properties from this data which is what one would use here.

So it seems that this is workaround around what DataSource was made for: getting the data close to where you actually use it (remember the `` analogy). I think there's a @change callback or equivalent that @johncowen created to still be able to have computed properties if it were absolutely required. To me most of the time computed properties are better avoiding because they are implicit unscoped, global variables.

src/app/policies/components/PolicyList.vue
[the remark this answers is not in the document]
lahabana: We can debate about this for a while so I'll stop here, but `inlining expressions` is precisely what happens with `v-if` and `v-for`.

PR #1357, opened 2023-08-21
refactor: uses DataSource for envoy data

kleinfreund:
Uses DataSource for the `EnvoyData` component everywhere.


src/app/data-planes/components/DataPlaneDetails.vue
```diff
@@ -158,38 +158,80 @@
...
+
+            <EmptyBlock v-else-if="data === ''" />
+
+            <EnvoyData
+              v-else
+              :content="data"
+              query-key="envoy-data-xds-data-plane"
+              @refresh="refresh"
+            />
+          </DataSource>
```
lahabana, line 180: Why do we need to extract the data fetching from the `EnvoyData` component. It just seems very verbose and redundant. Seems like we could achieve the same with just passing the :src as a prop to `EnvoyData` no?

kleinfreund: That’s possible. I’ve made a change for this. The downside to that is how we source the `EnvoyDataSource` type. I currently picked one at random because they’re virtually the same between dataplanes and zones, but it’s weird to import either inside this generic component.

johncowen (approved):
I think this is fine considering we don't have a DataSource related loader component as yet. It is a bit weird that you can only feed this component data with a datasource that it has internally, but it looks like we are avoiding a bit of repetition here, so mostly LGTM!

PR #1373, opened 2023-08-22
chore: expands standalone test coverage

kleinfreund:
chore(zones): adds test for lack of create zone button

chore(dataplanes): adds test for Zone column

### DIFF
--- a/features/diagnostics/DetailViewContent.feature
+++ b/features/diagnostics/DetailViewContent.feature
@@ -5,10 +5,9 @@ Feature: Diagnostics: Detail view content
       | details | [data-testid='code-block-diagnostics'] |
 
   Scenario: Diagnostics detail view has expected content
-    Given the URL "/config" responds with
+    Given the environment
       """
-      body:
-        mode: global
+      KUMA_MODE: global
       """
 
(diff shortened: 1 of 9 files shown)


johncowen (commented):
`global/standalone` should no longer use the `/config` endpoint.

My last 2 PRs you reviewed make this available via env vars/`can`

- https://github.com/kumahq/kuma-gui/pull/1261
- https://github.com/kumahq/kuma-gui/pull/1334

Whilst this work is not yet done, as the underlying work _is_ done, we shouldn't be writing tests using the old API approach at this point. We'll only have to come back in here and undo it in the near future.

kleinfreund:
As far as I can tell, writing tests using `KUMA_MODE` doesn’t work. Trying it now just makes these tests fail.

Besides that, we should write tests against the mechanisms we’re actually using (which is API calls to `/config`) not what we will be using in the future. Once we’re actually using this mechanism, it should be trivial to update the testing approach.

johncowen (approved):
Assuming https://github.com/kumahq/kuma-gui/pull/1293 is no longer required, this LGTM!

PR #1374, opened 2023-08-22
chore(policies): moves affected DPPs into overview

kleinfreund:
Moves the affected DPPs tab content into the overview sections and removes the tabs.


src/app/policies/components/PolicyConnections.vue
```diff
@@ -12,17 +12,17 @@
...
       :error="error"
     />
 
-    <EmptyBlock v-else-if="data === undefined || data.items.length === 0" />
+    <LoadingBlock v-else-if="data === undefined" />
+
+    <EmptyBlock v-else-if="data.items.length === 0" />
 
     <template v-else>
       <p
```
lahabana, line 28: Can we change this to at least an `<ul>` or something like this while we're touching this?

kleinfreund: Done.

src/app/policies/components/PolicyConnections.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <h2>Dataplanes</h2>
+  <h3>{{ t('policies.detail.affected_dpps') }}</h3>
```
johncowen, line 2: I think these things should be move into the `*View.vue`. We should try to never have `h*`s in components as it makes them not reusable. I'm pretty sure we could get rid of the `PolicyDetails` component entirely?

kleinfreund: I moved the heading into the PolicyDetails component, but I want to keep this split of *View and (mostly) presentational components.

lahabana (approved):
Feels like the PolicyConnections component could be inlined.

PR #1376, opened 2023-08-23
chore: updates breadcrumbs and page titles

kleinfreund:
**chore: removes suffixes from page titles in detail views**

Removes the suffixes from page titles in detail views (e.g. the service detail view will now only include the service name and not "{name} Service").

**chore(meshes): updates breadcrumbs and heading levels**

Removes the Mesh name from the breadcrumbs for the list views Services, Gateways, Data Plane Proxies, and Policies.

Removes the Mesh name heading from the detail views Service, Gateway, Data Plane Proxy, and Policy. Instead, their respective resource name is now the level 1 heading (used to be level 2).


johncowen (approved):
LGTM!

PR #1381, opened 2023-08-23
chore: Uses can for zone abilities/features in all *View.vue files

johncowen:
Uses `can` for checking whether we can do zone type things in all `*View.vue` files (where we export `can` from RouteView).

There will be a follow up for the components. I've split this into separate PRs as I would rather look to see how straightforwards it is to pass down whether we can do zone type things into the components rather than using can in the component directly.

### DIFF
--- a/src/app/onboarding/views/WelcomeView.spec.ts
+++ b/src/app/onboarding/views/WelcomeView.spec.ts
@@ -17,7 +17,6 @@ describe('WelcomeView.vue', () => {
       return merge({
         body: {
           environment: 'universal',
-          mode: 'standalone',
         },
       })
     })
--- a/src/app/onboarding/views/WelcomeView.vue
+++ b/src/app/onboarding/views/WelcomeView.vue
@@ -1,5 +1,7 @@
 <template>
-  <RouteView>
+  <RouteView
+    v-slot="{ t, can }"
+  >
     <RouteTitle
       :title="t('onboarding.routes.welcome.title', {name: t('common.product.name')})"
     />
@@ -32,7 +34,33 @@
             <div class="item-status-list-wrapper">
               <ul class="item-status-list">
                 <li
-                  v-for="item in statuses"
+                  v-for="item in [
+                    {
+                      name: `Run ${t('common.product.name')} control plane`,
+                      status: true,
+                    },
+                    {
+                      name: 'Learn about deployments',
+                      status: false,
+                    },
+                    {
+                      name: 'Learn about configuration storage',
+                      status: false,
+                    },
+                    ...can('use zones') ? [{ name: 'Add zones', status: false }] : [],
+                    {
+                      name: 'Create the mesh',
+                      status: false,
+                    },
+                    {
+                      name: 'Add services',
+                      status: false,
+                    },
+                    {
+                      name: 'Go to the dashboard',
+                      status: false,
+                    },
+                  ]"
                   :key="item.name"
                 >
                   <span class="circle mr-2">
@@ -55,7 +83,7 @@
           </template>
         </OnboardingPage>
 
-        <WelcomeAnimationSvg :longer="isMulticluster" />
+        <WelcomeAnimationSvg :longer="can('use zones')" />
       </div>
     </AppView>
   </RouteView>
@@ -74,44 +102,14 @@ import AppView from '@/app/application/components/app-view/AppView.vue'
 import RouteTitle from '@/app/application/components/route-view/RouteTitle.vue'
 import RouteView from '@/app/application/components/route-view/RouteView.vue'
 import { useStore } from '@/store/store'
-import { useI18n } from '@/utilities'
 
 const store = useStore()
-const { t } = useI18n()
 
 const enviromentFormatted = computed(() => {
   const environment = store.getters['config/getEnvironment']
   return environment.charAt(0).toUpperCase() + environment.slice(1)
 })
 
-const isMulticluster = computed(() => store.getters['config/getMulticlusterStatus'])
-const statuses = computed(() => [
-  {
-    name: `Run ${t('common.product.name')} control plane`,
-    status: true,
-  },
-  {
-    name: 'Learn about deployments',
-    status: false,
-  },
-  {
-    name: 'Learn about configuration storage',
-    status: false,
-  },
-  ...isMulticluster.value ? [{ name: 'Add zones', status: false }] : [],
-  {
-    name: 'Create the mesh',
-    status: false,
-  },
-  {
-    name: 'Add services',
-    status: false,
-  },
-  {
-    name: 'Go to the dashboard',
-    status: false,
-  },
-])
 </script>
 
 <style lang="scss" scoped>
--- a/src/app/onboarding/views/__snapshots__/WelcomeView.spec.ts.snap
+++ b/src/app/onboarding/views/__snapshots__/WelcomeView.spec.ts.snap
@@ -137,6 +137,14 @@ exports[`WelcomeView.vue renders snapshot 1`] = `
                       >
                         <!--v-if-->
                       </span>
+                       Add zones
+                    </li>
+                    <li>
+                      <span
+                        class="circle mr-2"
+                      >
+                        <!--v-if-->
+                      </span>
                        Create the mesh
                     </li>
                     <li>
@@ -211,7 +219,7 @@ exports[`WelcomeView.vue renders snapshot 1`] = `
         </div>
         <svg
           class="background svg"
-          longer="false"
+          longer="true"
           viewBox="0 0 1920 1080"
           xmlns="http://www.w3.org/2000/svg"
         >
--- a/src/app/zones/views/ZoneListView.vue
+++ b/src/app/zones/views/ZoneListView.vue
@@ -1,6 +1,6 @@
 <template>
   <RouteView
-    v-slot="{ route }"
+    v-slot="{ route, t, can, env }"
     name="zone-cp-list-view"
   >
     <AppView>
@@ -14,7 +14,7 @@
       </template>
 
       <template
-        v-if="env('KUMA_ZONE_CREATION_FLOW') === 'enabled' && store.getters['config/getMulticlusterStatus'] && isCreateZoneButtonVisible"
+        v-if="can('create zones') && isCreateZoneButtonVisible"
         #actions
       >
         <KButton
@@ -27,7 +27,7 @@
         </KButton>
       </template>
 
-      <MultizoneInfo v-if="store.getters['config/getMulticlusterStatus'] === false" />
+      <MultizoneInfo v-if="!can('use zones')" />
 
       <template v-else>
         <DataSource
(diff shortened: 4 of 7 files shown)


src/app/onboarding/views/WelcomeView.spec.ts
```diff
@@ -17,7 +17,6 @@ describe('WelcomeView.vue', () => {
       return merge({
         body: {
           environment: 'universal',
-          mode: 'standalone',
```
kleinfreund, line 20: **Issue**: Until we actually use state like mode from the index.html file, I’m not okay with removing the test coverage for the API-based source of data. We can always do that after we switched (or while we do so) to that approach.

src/app/onboarding/views/__snapshots__/WelcomeView.spec.ts.snap
```diff
@@ -211,7 +219,7 @@ exports[`WelcomeView.vue renders snapshot 1`] = `
         </div>
         <svg
           class="background svg"
-          longer="false"
+          longer="true"
```
kleinfreund, line 222: **Question**: How come this changed snapshot changed?

src/app/zones/views/ZoneListView.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <RouteView
-    v-slot="{ route }"
+    v-slot="{ route, t, can }"
```
kleinfreund, line 3: **Question**: Remind me, why do you access `t` and `can` via `RouteView`? Both of them should be available as utilities via a `const { t } = useI18n()` and `const can = useCan()`, no?

src/app/zones/views/ZoneTabsView.vue
```diff
@@ -1,11 +1,33 @@
+<script lang="ts" setup>
+import AppView from '@/app/application/components/app-view/AppView.vue'
+import RouteView from '@/app/application/components/route-view/RouteView.vue'
+import NavTabs from '@/app/common/NavTabs.vue'
+</script>
```
kleinfreund, line 5: **Comment**: I really don’t think we spent our time well by moving the script section to the top of files. In a component, the template should have the priority as that’s what we’re building, that’s the result. What a component imports is at best secondary and can be looked up easily enough.

vite.config.preview.ts
```diff
@@ -46,6 +46,10 @@ export const config: (context: PreviewConfigContext) => UserConfigFn = ({
                 baseGuiPath: base,
                 apiUrl: api,
                 version,
+                product: 'Kuma',
+                mode: 'global',
+                environment: 'universal',
+                apiReadOnly: false,
```
kleinfreund, line 52: **Question**: Out of curiosity, can actually rely on this data being present in the index.html file already? I.e. is the backend work for this complete?

kleinfreund (changes requested):

src/app/onboarding/views/WelcomeView.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <RouteView>
+  <RouteView
+    v-slot="{ t, can }"
```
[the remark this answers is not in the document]
johncowen: I can jump to the definition from the reference on line 3 above. As for docs there is an initial stub for it here

https://github.com/kumahq/kuma-gui/blob/master/src/app/application/services/can/README.md

Which will be expanded upon in the future. There's an initial PR to help me to do this here https://github.com/kumahq/kuma-gui/pull/1138, but ideally I would like to add automatically derived args/slots/props documentation so I don't have to write it all out twice.

src/app/zones/views/ZoneListView.vue
johncowen: They are things you will pretty much _always_ need in a route component, ideally they should never be used outside of a route component.

kleinfreund: Why? There doesn’t seem to be a good reason to impose such a limitation and exposing this utility in the script just seems easier. Also, it’s quite likely that these utilities will be necessary in non-component code (for example on the route level).

src/app/zones/views/ZoneTabsView.vue
johncowen: We discussed this offline, as far as I remember the decision was to move the headers to the top of the file throughout the project.

---

The above decision was based on the fact that all that is left in the majority of files now are the JS/TS 'headers', and they should be at the 'head' of the file. In the past I've found it really weird having to find the headers in the middle of the file and having to guess where they are to take a quick look at them instead of jumping to the 'head' of the file. 

>  I really don’t think we spent our time well by moving the script section

From what I remember I think in our previous iteration of the project you spent a lot of time moving these around, I think you mentioned this in a PR a while back. The project has vastly changed since then and as mentioned above, pretty much all that is left in terms of JS/TS are the 'headers'

There is a tool to do this plus an eslint rule for the future, I plan to PR that change soon, (also see https://twitter.com/youyuxi/status/1427982990957846529) so comparatively there will far less time spent moving these to the 'head' of the file.

Let me know if you still have any concerns.

vite.config.preview.ts
johncowen: I ran `kuma` and took these values from the HTML file it serves. The PR for this was done quite a while ago I believe, let me know if you want me to find the link for it.

src/app/onboarding/views/WelcomeView.vue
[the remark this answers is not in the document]
lahabana: I wasn't asking for more docs :) I was just saying that I can't find the definition myself and it would be great if  you held my hand here :)

src/app/zones/views/ZoneListView.vue
johncowen: > Also, it’s quite likely that these utilities will be necessary in non-component code (for example on the route level).

It's fine to use these in non-component code, just inject the service where necessary.

When I said "ideally they should never be used outside of a route component", I meant "ideally they shouldn't be used in non-route components" i.e. I was only considering components here, and when I say "ideally" I mean there are obviously times when you do want to use these in non-route components, but its far less common, and using them in the route component and passing the data down should be the 'default'.

src/app/onboarding/views/WelcomeView.vue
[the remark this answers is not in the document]
johncowen: Oh I see! Sorry, umm here is `can`:

https://github.com/kumahq/kuma-gui/blob/master/src/app/application/services/can/index.ts

`can` itself is configured with composable 'features' these `features.ts` will all live in the root of each 'module' that we have, i.e. https://github.com/kumahq/kuma-gui/blob/master/src/app/zones/features.ts

src/app/onboarding/views/__snapshots__/WelcomeView.spec.ts.snap
johncowen: Because the snapshots are no longer taken in standalone mode

src/app/onboarding/views/WelcomeView.spec.ts
johncowen: We barely have any tests over these values at the moment, and the test here only tests a design element for the Welcome screen in onboarding, which in the scheme of things is a bit 🤷 .

Considering there was no test to check what the snapshot was when this was in non-standalone mode, I'm not totally sure that this is worth blocking on, especially considering the importance of getting this work done.

I noticed you've started expanded on these here https://github.com/kumahq/kuma-gui/pull/1373 but using the old API based approach that I've been working on removing for quite a while now, and considering:

> Once we’re actually using this mechanism, it should be trivial to update the testing approach.

1. We are using this mechanism.
2. You said it in the above PR it should be trivial to update.

Maybe it's best to update it over there in https://github.com/kumahq/kuma-gui/pull/1373?

I agree that testing these things in other parts of the code base is important, so let me know what you think.

johncowen: P.S. I also noticed today whilst reviewing PRs that where merged whilst I was away that quite a lot of test coverage for things like loading and error states where removed in a few of the PRs that went in other the past few weeks. I would be more "not ok" with us removing those, are we definitely sure all our loading/error states are fine?

kleinfreund: My point was really more general: Since we’re not using the non-API-based approach yet, the tests are more valuable as they’re currently written because they match the reality of our application in use.

I just don’t think we should switch tests to a new approach that isn’t actually in use, yet. That should happen when the change is made, not before. That’s also why I don’t want to update https://github.com/kumahq/kuma-gui/pull/1373 to use a different approach because that wouldn’t test what we’re actually using.

What’s the problem with updating the tests when we actually start using this different approach?

johncowen: > My point was really more general

If its more general then should it be blocking this PR, especially considering the triviality of the test.

> What’s the problem with updating the tests when we actually start using this different approach?

Well you already have that underway right? You said it was trivial to update to use the new approach, seeing as you've known for a while now that we are using this new approach following all the discussions we've had and the PRs that have been merged, I think it would be better to continue and build on the work you have over in #1373 especially considering you think it will be trivial to add.

kleinfreund: I think expecting our tests to run against the mechanism we actually use is a pretty sensible one to make. This test isn’t relevant enough to block this PR over it so I will resolve this conversation, but I will not change my approach in https://github.com/kumahq/kuma-gui/pull/1373 (considering also that the tests break once I try to switch the approach).

> Do you have any comment on:

Not really, it’s not related to this conversation and our coverage for the tests that I removed is good.

src/app/zones/views/ZoneListView.vue
kleinfreund: So then why don’t we just always access `can` via the script section?

src/app/onboarding/views/WelcomeView.vue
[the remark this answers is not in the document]
lahabana: Looks cool! Will we have auto completion of the features in `can()` ?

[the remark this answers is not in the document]
johncowen: Eventually yes (I hope)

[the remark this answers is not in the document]
kleinfreund: See also https://github.com/kumahq/kuma-gui/pull/1334#discussion_r1294298195

src/app/zones/views/ZoneListView.vue
johncowen: We can't we just always access `can` via the template section? i.e. why have an entire script section plus `import`, plus `useCan` if we don't even need a script section otherwise?

Umm, could you explain more what your issue is with using:

```javascript
RouteView = (can) => {
  can('')
}
```

instead of:

```javascript
import '@/app/.../can'
const can = useCan()
RouteView = () => {
  can('')
}
```

src/app/zones/views/ZoneTabsView.vue
lahabana: We've now spent more effort discussing this than any other option. This is really getting into a tabs vs space argument here.

src/app/zones/views/ZoneListView.vue
kleinfreund: I just don’t understand why `RouteView` has anything to do with `can`. It’s incidental that  that component appears in commonly.

kleinfreund (approved):

PR #1387, opened 2023-08-25
chore: standardizes empty state usage

kleinfreund:
## Changes

**chore: standardizes empty state usage**

Makes AppCollection accept our standard empty state component and passes it on to KTable. This way, each list view can easily customize the empty state with enriched UI elements like a "Create Zone" button. By default, our own empty state component will also be used (again) in list views.

**chore: uses primary appearance for creation buttons**

**chore: adds Create Zone button to overview**

## Screenshots

**Before**:

image

**After**:

image

image

### DIFF
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -20,11 +20,17 @@
     :cell-attrs="({ headerKey }: CellAttrParams) => ({
       class: `${headerKey}-column`
     })"
-    empty-state-icon-size="96"
     disable-sorting
     hide-pagination-when-optional
     @row:click="click"
   >
+    <template
+      v-if="$slots.emptyState || props.items?.length === 0"
+      #empty-state
+    >
+      <EmptyBlock />
+    </template>
+
     <template
       v-for="key in Object.keys(slots)"
       :key="key"
@@ -52,6 +58,8 @@
 <script lang="ts" setup>
 import { KTable, TableHeader } from '@kong/kongponents'
 import { useSlots, ref, watch, computed } from 'vue'
+
+import EmptyBlock from '@/app/common/EmptyBlock.vue'
 
 type CellAttrParams = {
   headerKey: string
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -42,7 +42,15 @@
           v-if="props.emptyStateCtaTo"
           #cta
         >
+          <DocumentationLink
+            v-if="typeof props.emptyStateCtaTo === 'string'"
+            :href="props.emptyStateCtaTo"
+          >
+            {{ props.emptyStateCtaText }}
+          </DocumentationLink>
+
           <KButton
+            v-else
             appearance="primary"
             icon="plus"
             :to="props.emptyStateCtaTo"
@@ -82,6 +90,7 @@ import { KButton, KTable, TableHeader } from '@kong/kongponents'
 import { useSlots, ref, watch, computed } from 'vue'
 import { RouteLocationRaw } from 'vue-router'
 
+import DocumentationLink from '@/app/common/DocumentationLink.vue'
 import EmptyBlock from '@/app/common/EmptyBlock.vue'
 import { PAGE_SIZE_DEFAULT } from '@/constants'
 import { useI18n } from '@/utilities'
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -5,7 +5,7 @@
     :style="`--column-width: ${columnWidth}; --special-column-width: ${SPECIAL_COLUMN_WIDTH}%;`"
     :has-error="(typeof props.error !== 'undefined')"
     :pagination-total-items="props.total"
-    :initial-fetcher-params="{ page: props.pageNumber, pageSize: props.pageSize }"
+    :initial-fetcher-params="{ page: props.pageNumber, pageSize: props.pageSize ?? PAGE_SIZE_DEFAULT }"
     :headers="props.headers"
     :fetcher-cache-key="String(cacheKey)"
     :fetcher="({ page, pageSize, query }: FetcherParams) => {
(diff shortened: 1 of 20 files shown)


lahabana:
@kleinfreund what's the hack to have the mocks return 0 data?

kleinfreund:
> what's the hack to have the mocks return 0 data?

You can open dev tools and go to application, cookies and set variables like `KUMA_ZONE_COUNT`/`KUMA_MESH_COUNT` to `0`.

src/app/main-overview/components/ZoneControlPlanesDetails.vue
```diff
@@ -8,6 +8,35 @@
     disable-sorting
     hide-pagination-when-optional
   >
+    <template
+      v-if="props.zoneOverviews.length === 0"
+      #empty-state
+    >
+      <EmptyBlock>
```
lahabana, line 15: As CTA are meant to be systematic. Would make sense to always have this in the AppCollection?

You'd only take conf for:

- title
- desc
- button content
- where the button gets us

I feel like it'd be easier to adopt than all of this for each of the lists

kleinfreund: I agree. I turned empty state usage for AppCollection into props and changed the two tables on the overview to use AppCollection so they can make use of the same mechanism.

src/app/zones/views/ZoneListView.vue
```diff
@@ -14,11 +14,11 @@
       </template>
 
       <template
```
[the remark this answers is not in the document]
kleinfreund: I’ll remove what you’re mentioning (the blue alert with the "Create Zone" button). What you commented on is the Create Zone button in the Zone CP list view that (now) only shows up if the list is not empty. That, we shouldn’t remove.

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -20,11 +20,39 @@
...
+          {{ props.emptyStateMessage }}
+        </template>
+
+        <template
+          v-if="props.emptyStateCtaTo"
+          #cta
+        >
+          <KButton
+            appearance="primary"
+            icon="plus"
```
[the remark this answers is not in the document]
kleinfreund: Yep, we can easily customize this. Do you want me to add documentation links for empty states of list views (where we don’t have a creation flow)?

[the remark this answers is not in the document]
kleinfreund: I’ve added docs link to empty states of the ingress, egress, mesh, DPP, and policy list views. Couldn’t think of a fitting docs page for the service list view.

src/app/main-overview/components/ZoneControlPlanesDetails.vue
```diff
@@ -1,12 +1,17 @@
...
       { label: t('main-overview.detail.zone_control_planes.table.status'), key: 'status'},
     ]"
-    :fetcher="() => ({ data: tableData })"
-    disable-sorting
-    hide-pagination-when-optional
+    :items="tableData"
+    :total="tableData.length"
+    :empty-state-title="t('zone-cps.empty_state.title')"
+    :empty-state-message="env('KUMA_ZONE_CREATION_FLOW') === 'enabled' ? t('zone-cps.empty_state.message') : t('common.emptyState.message', { type: 'Zones' })"
+    :empty-state-cta-to="env('KUMA_ZONE_CREATION_FLOW') === 'enabled' ? { name: 'zone-create-view' } : undefined"
```
lahabana, line 13: Something seems odd I have a "create zone button" when using Kuma:

image

Also can we had an icon that is not Kong here? Clearly Kuma shouldn't show Kong logos.

kleinfreund: > Something seems odd I have a "create zone button" when using Kuma:

What’s odd here? This looks expected to me.

> Also can we had an icon that is not Kong here? Clearly Kuma shouldn't show Kong logos.

I’ve changed that for a more generic "no data" icon from the list here: https://kongponents.konghq.com/components/icon.html#state-icons

src/app/main-overview/components/ControlPlaneDetails.vue
```diff
@@ -127,10 +115,7 @@
             </div>
           </div>
 
-          <EmptyBlock v-if="props.meshInsights.length === 0" />
-
           <MeshesDetails
-            v-else
             data-testid="meshes-details"
             :mesh-insights="props.meshInsights.slice(0, 10)"
```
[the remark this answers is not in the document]
kleinfreund: Had missed this one. Added it now.

src/app/zones/views/ZoneListView.vue
```diff
@@ -33,6 +33,7 @@
         <DataSource
           v-slot="{ data, error, refresh }: ZoneOverviewCollectionSource"
           :src="`/zone-cps?size=${props.size}&page=${props.page}`"
+          @change="setIsCreateZoneButtonVisible"
```
lahabana, line 36: I don't get why the button doesn't move in the v-else and inside this DataSource.
This fixes:

- need to repeat checks to `store.getters`
- need to add global state
- readability (the component is poorly named so I had to check the code to see that the `MultizoneInfo` component is actually the component that shows that we're in standalone).

I've just looked and zoneIngressList does this and it's a lot easier to follow.
Btw I also just noticed that the conditional is missing from zoneEgressList

kleinfreund: This button, by design, is placed in the same row as the heading "Zone Control Planes". To do this, we use `AppView`’s `actions` slot. That has the consequence that the button can’t move further into the tree.

The next alternative is moving the `DataSource` further out (specifically around `AppView`) so its data can be used for this element without setting the `isCreateZoneButtonVisible` state on the component. That would however cause the `DataSource` component to trigger a request without the check attached to the `MultizoneInfo` component preventing it in standalone mode. We could instruct `DataSource` to only make the request conditionally (using its `shouldMakeRequest` prop) to address this.

However, it would be easier to drop the `MultizoneInfo` component instead. We don’t show the Zone & Ingress views in standalone mode (though you can access them via URL) so its information is unlikely to be seen by users anyway. Then, moving the `DataSource` out around `AppView` provides the right things with the necessary data access.

kleinfreund: Because we want to show:

- if no Zones: empty state with create button
- if Zones: create button above

(I.e. following what we do in overview)

lahabana: You could use the `if=getMultiCluster` in the datasource no? In any case it doesn't matter too much.

PR #1400, opened 2023-08-28
chore: adds stylelint

kleinfreund:
Adds stylelint and configures the rule sets plugins for HTML, SCSS, and Vue.

Adds the `@kong/design-tokens/stylelint-plugin` plugin.


johncowen (commented):
There seem to be a few other fixes in here, is that right?

kleinfreund:
> There seem to be a few other fixes in here, is that right?

No.

johncowen:
What about the `$kui` variable removals? Did you remove those due to the calc?

kleinfreund:
> What about the `$kui` variable removals? Did you remove those due to the calc?

See https://github.com/Kong/design-tokens/tree/main/stylelint-plugin. TL;DR: The stylelint plugin disallows using KUI variables for unintended properties.

johncowen:
Ah I see, it because we are using space variables for positioning, cool, lgtm then, 1 tic

PR #1413, opened 2023-08-29
refactor: turns regular tabs into nav tabs based on routes

kleinfreund:
**refactor(data-planes): turns tabs into routes**

Turns the tabs of the Data Plane Proxy detail view into nav tabs based on actual routes.

**refactor(zones): turns tabs into routes**

Turns the tabs of the Zone detail views into nav tabs based on actual routes.

**refactor(services): turns tabs into routes**

Turns the tabs of the Service detail view into nav tabs based on actual routes.

**chore**: moves copy button into page title

Moves the copy button into the page title for all detail views.

Removes the name card from all detail views.


johncowen:
I know you are still working on this, but I wanted to check something.

If I'm not mistaken, with the tabs changes we now load the `dataplane+insights` endpoint every time we click on a tab, whereas previously we would load it once for the entire page.

There are pros and cons to each approach, but wanted to check what we actually want to build here.

TLDR;

Do we want it so:

1. We only request `dataplane+insights` once when we load the detail page and not when clicking through each tab (or)
2. We re-request `dataplane+insights` every time we click on a tab

?

kleinfreund:
> I know you are still working on this, but I wanted to check something.
> 
> If I'm not mistaken, with the tabs changes we now load the `dataplane+insights` endpoint every time we click on a tab, whereas previously we would load it once for the entire page.
> 
> There are pros and cons to each approach, but wanted to check what we actually want to build here.
> 
> TLDR;
> 
> Do we want it so:
> 
> 1. We only request `dataplane+insights` once when we load the detail page and not when clicking through each tab (or)
> 2. We re-request `dataplane+insights` every time we click on a tab
> 
> ?

This is indeed a problem and I’m not sure it’s easily avoided with the current setup.

Moving the data sourcing into `src/app/data-planes/views/DataPlaneDetailTabsView.vue` doesn’t help (even though it saves on code repetition).

We can’t really avoid this save for actively caching results on the data source level across page views because moving between tabs will now behave like a navigation (to the router anyway).

johncowen:
So from that I understand that we want option 1? i.e. We load it once?

If so we should use the DataSource in the upper View component (the one that holds the tabs) as you've suggested above, and then pass that data down into each tab via the RouterView component. Doing that should mean that we only call that endpoint once for the page and not every time we click on the tab.

johncowen:
P.S. if you are looking at implementing the above suggestion, and its easier to split the other things in this PR off into their own PRs so we can get them moving separately/quicker let me know.

kleinfreund:
> So from that I understand that we want option 1? i.e. We load it once?
> 
> If so we should use the DataSource in the upper View component (the one that holds the tabs) as you've suggested above, and then pass that data down into each tab via the RouterView component. Doing that should mean that we only call that endpoint once for the page and not every time we click on the tab.

That’s exactly what I had tried and it didn’t work right away. The issue was (and I’ve addressed this now) is that we had a couple of places in our application where we force re-rendering of `RouterView` content on page transition regardless of whether we need that to happen. As a consequence, even the tabs components would re-render completely (not just remount) and would impose the loading again.

I addressed this now and it seems to work. The loading of the "main" resource in all tabbed views now happens in the TabsView component and is passed down.

johncowen:
> The issue was (and I’ve addressed this now) is that we had a couple of places in our application where we force re-rendering of RouterView content on page transition regardless of whether we need that to happen. As a consequence, even the tabs components would re-render completely (not just remount) and would impose the loading again.

Oh perfect! I'd noticed we were re-rendering the entire application on every single route change a few times, so I'm glad we aren't doing that anymore.

I noticed we added that here, but I don't see why?

https://github.com/kumahq/kuma-gui/commit/33a7b40f9693813f1fa3937aa0992cb609168bca#diff-7a7a37b12ee1265d7e27577ec286120d2cbc0940908635e264a2be44ccb9a8a0

There were a few things when I was doing the titling and breadcrumbing that this issue affected and I tried to build those in such a way that ignored that we were re-rendering everything on every route change (so that when that was removed, things should still work). So I suppose my biggest question is, do you know why that was added?

kleinfreund:
> I noticed we added that here, but I don't see why?
> 
> [33a7b40#diff-7a7a37b12ee1265d7e27577ec286120d2cbc0940908635e264a2be44ccb9a8a0](https://github.com/kumahq/kuma-gui/commit/33a7b40f9693813f1fa3937aa0992cb609168bca#diff-7a7a37b12ee1265d7e27577ec286120d2cbc0940908635e264a2be44ccb9a8a0)
> 
> There were a few things when I was doing the titling and breadcrumbing that this issue affected and I tried to build those in such a way that ignored that we were re-rendering everything on every route change (so that when that was removed, things should still work). So I suppose my biggest question is, do you know why that was added?

Yes, I remember that. Back then, we had a few interactions on a quasi global level that changed route parameters in response to which we expected things to update. The biggest example is the mesh selector. Switching the mesh there used to change the mesh parameter and as a result, it was expected that you stayed, for example on the service list view - just in the context of a different mesh. To make sure this worked, this "route view keying" was introduced and back then it was a bit more specific: only certain routes triggered it.

We do have something similar in the app currently and that’s the policy list view. Clicking on a policy type has a similar effect (but it’s much more local). For that reason, I added a `key` prop to the `PolicyList` element to make sure the table updates correctly (it doesn’t without).

src/app/services/views/ServiceDetailView.vue
```diff
@@ -4,222 +4,60 @@
...
-              v-slot="{ data: externalService, error: externalServiceError }: ExternalServiceSource"
-              :src="`/meshes/${route.params.mesh}/external-services/${route.params.service}`"
-            >
-              <LoadingBlock v-if="externalService === undefined" />
-
-              <ErrorBlock
-                v-else-if="externalServiceError"
-                :error="externalServiceError"
-              />
+          <LoadingBlock v-if="externalService === undefined" />
```
johncowen, line 22: What happens if externalService is undefined (i.e. still loading) and we get an error?

src/app/services/views/ServiceListView.vue
```diff
@@ -57,7 +57,7 @@
               </template>
 
               <template #serviceType="{ rowValue }">
-                {{ rowValue || t('common.collection.none') }}
+                {{ rowValue || 'internal' }}
```
johncowen, line 60: Should this use `t`, seeing as it was using `t` previously? Isn't this a `http.api.value`?

johncowen (commented):
I'm kinda half way through a code review and I wanted to ask if it is a major hassle to split these up so this PR is only about redoing the tabs and not all the other design changes that we said had a higher priority. There is lots of change in this PR (possibly some due to just moving stuff around) so its not straightforwards to look at. If the design related changes were in their own PR then it would much quicker to get them approved and merged.

Anyway let me know, if its a massive hassle don't worry though.

kleinfreund:
> I'm kinda half way through a code review and I wanted to ask if it is a major hassle to split these up so this PR is only about redoing the tabs and not all the other design changes that we said had a higher priority. There is lots of change in this PR (possibly some due to just moving stuff around) so its not straightforwards to look at. If the design related changes were in their own PR then it would much quicker to get them approved and merged.
> 
> Anyway let me know, if its a massive hassle don't worry though.

That would honestly be quite a big hassle. I wouldn’t have conflated the changes to the tabs with the changes to titles/copy buttons had I known this before starting work on the former. With the tabs changes splitting up the views into different files, it would mean doing the changes to titles and copy buttons twice.

I will have a look at what it would take to just do the title changes and copy button changes separately from this PR, but it would inadvertently lead to quite a lot of conflicts with this PR. It’s unfortunate that the work on tabs happens to touch exactly the same files as the ones for titles.

I’ll have a look.

src/app/services/views/ServiceListView.vue
kleinfreund: This should definitely not use t. This is a value that may never be different.

kleinfreund:
I filed https://github.com/kumahq/kuma-gui/pull/1419 now. This will be a best effort type PR as A LOT of its changes will simply be undone by this PR.

src/app/services/views/ServiceListView.vue
johncowen: Yeah gotcha, but its a value that is like this because it comes from the `http.api` right? All the others like this use `t` (well all the others up until now)

kleinfreund: This is an API value that should be only every represented as the literal string `'internal'`. In other cases, changing the value for aesthetic reasons might be fine. Here, I really think this *must* not happen. This should only ever be allowed to be exactly `'internal'`.

src/app/services/views/ServiceDetailView.vue
kleinfreund: Then this stays in a loading state. I’ve changed the order for this instance now.

PR #1236, opened 2023-07-25
chore: Make route guards injectable

johncowen:
Takes 1 of a few steps to enable us to disable onboarding entirely i.e. not even compile it in if we wanted to do that.

~Currently awaiting merge of some other PRs (specifically https://github.com/kumahq/kuma-gui/pull/1221) but I wanted to get this up for visibility for the moment, it's likely to change a little.~

Signed-off-by: John Cowen <john.cowen@konghq.com>

Issue #1332, opened 2023-08-14
improvement: Implement request counting for Cypress


johncowen:
### Description

After visiting a page in our e2e tests we should wait for all in-flight HTTP requests to respond before moving on.

We should be able to do this via request counting in `intercept`.

I would guess this could solve some of our flake and reduce our reliance on `wait`ing.

If it does help, then we should think about using it for `click` steps also.

Additionally, whilst most Cypress commands should wait until the assertion succeeds (and timeout if not), look to see if there are places where this isn't happening.

johncowen:
Closing this one at least for the moment see https://github.com/kumahq/kuma-gui/pull/1482

PR #1414, opened 2023-08-29
chore(onboarding): Makes onboarding route guards injectable/modular

johncowen:
Continuation of https://github.com/kumahq/kuma-gui/pull/1236

Moves onboarding route guards into their own 'module'

Also:

- Adds exact URL asserting so we can test for `/` (`contains` isn't enough for `/`)
- Temporarily moves our custom `createRouter` into `production.ts`, so we can then delete `router.ts` (this will be moved elsewhere later). We no longer have our own distinct createRouter/router setup function.
- Permanently deletes the old `#/url` style route guard. I don't think nowadays this is needed anymore (happy to put back if need be)

### DIFF
--- a/features/onboarding/Index.feature
+++ b/features/onboarding/Index.feature
@@ -1,0 +1,16 @@
+Feature: onboarding / index
+  Background:
+    Given the CSS selectors
+      | Alias           | Selector                              |
+      | skip-button           | [data-testid='onboarding-skip-button'] |
+      | table-header    | $table th                             |
+      | table-row       | $table tbody tr                       |
+      | dataplane-title | [data-testid='data-plane-collection'] |
+  Scenario: Visiting onboarding redirects to either welcome or root
+    When I visit the "/onboarding" URL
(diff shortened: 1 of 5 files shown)


features/onboarding/Index.feature
```diff
@@ -0,0 +1,16 @@
...
+      | Alias           | Selector                              |
+      | skip-button           | [data-testid='onboarding-skip-button'] |
+      | table-header    | $table th                             |
+      | table-row       | $table tbody tr                       |
+      | dataplane-title | [data-testid='data-plane-collection'] |
+  Scenario: Visiting onboarding redirects to either welcome or root
+    When I visit the "/onboarding" URL
+    Then the URL is "/onboarding/welcome"
+    Then I click the "$skip-button" element
+    And I wait for 300 milliseconds
```
johncowen, line 13: I'm almost certain https://github.com/kumahq/kuma-gui/issues/1332 will avoid all of this stuff.

kleinfreund (approved):

PR #1420, opened 2023-08-31
feat: moves configs into dedicated tabs

kleinfreund:
**feat(services): moves config into dedicated tab**

Moves the External Service config shown in the Service detail view into a new YAML tab.

**feat(dataplanes): moves subscriptions into DPP detail view**

**feat(dataplanes): moves config into dedicated tab**

Moves the config shown in the DPP detail view into a new YAML tab.

**feat(zones): moves subscriptions into Zone detail view**

**chore(subscriptions): always shows subscriptions as closed by default**

**feat(zones): moves configs into dedicated tabs**

Moves the config shown in the Zone detail view into a new Config tab.

Adds a new YAML to the Zone Ingress/Egress detail views showing their respective configs. With this, I also changes the data source keys that fetch `ZoneIngressOverview`s and `ZoneEgressOverview`s to make clear when one is loading the non-Overview resource and when one is loading an Overview.


src/app/services/views/ServiceDetailTabsView.vue
```diff
@@ -89,6 +89,13 @@ function getNavTabs(serviceInsight: ServiceInsight): NavTab[] {
 
   return routes
     .filter((route) => {
+      if (
+        (serviceInsight.serviceType === undefined || serviceInsight.serviceType === 'internal') &&
```
johncowen, line 93: If I understand this `internal` thing correctly. It would be good to just set `serviceType` to equal `internal` as a default when it comes in the the HTTP API, then we would know that it is always a `string` not `string | undefined` and not have to worry about these things in the rest of the application.

src/app/zones/sources.ts
```diff
@@ -76,6 +78,14 @@ export const sources = (api: KumaApi) => {
       return api.getZoneIngressData({ zoneIngressName: name, dataPath })
     },
 
+    '/zone-ingress-overviews/:name': async (params: DetailParams, source: Closeable) => {
```
johncowen, line 81: How comes the listing URIs for these are just `/zone-ingresses` and the details ones are `/zone-ingress-overviews`.

Is there a reason why the following would work out?

```
`/zone-ingresses`
`/zone-ingress/:name`
```

i.e. use the following everywhere

```
/things
/thing/:name
```

Maybe I'm not following what is being returned in these sources? lmk

src/services/kuma-api/KumaApi.ts
```diff
@@ -102,6 +108,10 @@ export default class KumaApi extends Api {
     return this.client.get(`/zoneingresses+insights/${name}`, { params })
   }
 
+  getZoneEgress({ name }: { name: string }, params?: any): Promise<ZoneEgress> {
+    return this.client.get(`/zoneegresses/${name}`, { params })
+  }
```
johncowen, line 113: Are we retrieving new data with this PR that we weren't retrieving before?

johncowen (commented):
Quick code review with some question, gonna take this for a spin now.

src/app/zones/sources.ts
johncowen: Ah I should read the description more carefully 🤦

src/app/policies/components/PolicyDetails.vue
```diff
@@ -15,7 +15,7 @@
     <ResourceCodeBlock
       id="code-block-policy"
       :resource="props.policy"
-      :resource-fetcher="fetchPolicy"
+      :resource-fetcher="(params) => kumaApi.getSinglePolicyEntity({ name: props.policy.name, mesh: props.policy.mesh, path: props.path }, params)"
```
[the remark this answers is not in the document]
kleinfreund: These can’t be switched out to DataSource easily because they’re on-demand interactions.

src/app/services/views/ServiceDetailTabsView.vue
kleinfreund: This is a case were really this shouldn’t even be possible to be omitted on the API in the first place, but as far as I’m aware it can’t be changed there because of our use of protobuf.

Not a big point in changing this now if we’re eventually streamlining this sort of data on the DataSource level.

src/app/services/views/ServiceConfigView.vue
```diff
@@ -0,0 +1,57 @@
...
+          <DataSource
+            v-slot="{ data: externalService, error: externalServiceError }: ExternalServiceSource"
+            :src="`/meshes/${route.params.mesh}/external-services/${route.params.service}`"
+          >
+            <LoadingBlock v-if="externalService === undefined" />
+
+            <ErrorBlock
+              v-else-if="externalServiceError"
+              :error="externalServiceError"
+            />
```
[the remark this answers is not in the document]
kleinfreund: Fixed.

src/app/zones/sources.ts
kleinfreund: I updated the list entries to also include `-overviews` to make this unambiguous.

src/services/kuma-api/KumaApi.ts
kleinfreund: Yes, we are. I did check these APIs exist and work as expected.

src/app/policies/components/PolicyDetails.vue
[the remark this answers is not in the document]
johncowen: We should be able to use `KToggle` for doing this (and if not we need our own disclosure component). No action here rn.

src/app/services/views/ServiceDetailTabsView.vue
johncowen: Yeah gotcha, no action here rn. but in relation to the ongoing data layer discussions, if we had our own data layer to smooth these things out/hide this sharp edge from our application we could until it was resolve in the backend

src/services/kuma-api/KumaApi.ts
johncowen: Cool ty

src/app/meshes/views/MeshConfigView.vue
```diff
@@ -0,0 +1,59 @@
...
+            :render="true"
+          />
+        </h2>
+      </template>
+
+      <KCard class="mt-4">
+        <template #body>
+          <DataSource
+            v-slot="{ data, error }: MeshSource"
+            :src="`/meshes/${route.params.mesh}`"
```
johncowen, line 21: Is this another place where we are loading the same data everytime we click a tab?

Anywhere we use the same data across different tabs, the DataSource should be added in the topmost View (where the tabs are) and then passed down using `RouterView`.

This means we make far fewer HTTP requests and only refresh the data when you first land on the 'detail/tabbed' page.

Are there other places where we are loading the same data over and over everytime you click a tab?

kleinfreund: It’s not easy to actually have such a clear-cut rule because it’s not always the case that all tabs in such a view need access to the same data. In this case, the Mesh resource is only needed when accessing the Overview or YAML tabs. Moving this data source into the tabs view component would cause it to be loaded if you, for example, access services directly without even going to the Mesh overview.

johncowen: Oh gotcha, interesting one, I guess we could only add the datasource to the page if certain tabs are active?

kleinfreund: Yeah, we could build something like that, but it’s kind of strange to transfer the knowledge of which "deep" view needs which data into a parent view (i.e. the tabs view). Not really sure how to handle this better.

johncowen: Not sure how thats strange, we quite often pass data down from parent components into deeper components

johncowen: Is this something you are looking at, or do we want to do that in a separate PR?

kleinfreund: Not looking at this, no. I don’t think in this case, moving the data source into the tabs is appropriate so I also don’t plan on following up on this immediately. This is fine.

johncowen: Ok I think we are good with this for the minute then, one sec

johncowen (approved):
👍 , some potential "circle backs" in here, but good to go!

PR #1422, opened 2023-08-31
fix: order of loading and error blocks

kleinfreund:
Fixes the order of loading and error blocks and makes them consistently check for `data` being `undefined`.

### DIFF
--- a/src/app/application/components/data-source/DataSource.vue
+++ b/src/app/application/components/data-source/DataSource.vue
@@ -1,7 +1,6 @@
 <template>
   <slot
     :data="(message as any)"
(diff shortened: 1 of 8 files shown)


johncowen (approved):
Thanks for sorting this!

I get why you added the `isLoading`, but as explained elsewhere in another PR somewhere, it felt like more of quick fix rather than anything, plus it wasn't used consistently everywhere.

I think there's a better thing coming in the form of a specific `Loader` type component that will do all this for us, with more explicit states than just loading and error, and also hide both `Error` and `Loading` blocks completely (and also centralize them). Until then I'm good with doing it like this.

PR #1440, opened 2023-09-05
chore(onboarding): Move getConfigurationType to use DataSource

johncowen:
Moves `getConfigurationType` to use `DataSource`.

Ideally the property this uses should probably be added to our `index.html` vars (cc @lahabana ), for now this PR just uses DataSource and `/config` seeing as we only use it in one place at present.


kleinfreund (approved):

lahabana:
@johncowen what's the config you need in index.html?

johncowen:
`config.store.type`

https://github.com/kumahq/kuma-gui/blob/master/src/test-support/mocks/src/config.ts#L245

lahabana:
Yes it's feasible

PR #1445, opened 2023-09-06
chore: tweaks copy as k8s button appearance

kleinfreund:
Adds a bordered style to the "Copy as Kubernetes" button to visually separate it from the default copy button a bit better.

Adds a tooltip to the "Copy as Kubernetes" button to explain its purpose.

### DIFF
--- a/src/app/common/ResourceCodeBlock.vue
+++ b/src/app/common/ResourceCodeBlock.vue
@@ -9,16 +9,33 @@
   >
     <template #secondary-actions>
       <CopyButton
+        class="kubernetes-copy-button"
         :get-text="getYamlAsKubernetes"
-        :copy-text="i18n.t('common.copyKubernetesText')"
+        :copy-text="t('common.copyKubernetesText')"
+        hide-title
       >
-        {{ i18n.t('common.copyKubernetesShortText') }}
+        {{ t('common.copyKubernetesShortText') }}
+
+        <KTooltip
+          :label="t('common.copyKubernetesTooltip')"
+          placement="bottomEnd"
+          max-width="200"
+        >
+          <KIcon
+            icon="info"
+            color="currentColor"
+            :size="KUI_ICON_SIZE_30"
+            hide-title
+          />
+        </KTooltip>
       </CopyButton>
     </template>
   </CodeBlock>
 </template>
 
 <script lang="ts" setup>
+import { KUI_ICON_SIZE_30 } from '@kong/design-tokens'
+import { KIcon, KTooltip } from '@kong/kongponents'
 import { PropType, computed } from 'vue'
 
 import CodeBlock from './CodeBlock.vue'
@@ -28,7 +45,7 @@ import type { Entity } from '@/types/index.d'
 import { useI18n } from '@/utilities'
 import { toYaml } from '@/utilities/toYaml'
 
-const i18n = useI18n()
+const { t } = useI18n()
 
 const props = defineProps({
   id: {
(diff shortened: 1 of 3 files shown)


src/app/common/ResourceCodeBlock.vue
```diff
@@ -76,3 +93,11 @@ function toYamlRepresentation(resource: Entity): string {
   return toYaml(resourceWithoutTimes)
 }
 </script>
+
+<style lang="scss">
+.kubernetes-copy-button:not(.increase-specificity.increase-specificity) {
```
johncowen, line 98: 😬 , what's the background here? I assume something else is already overwriting with `.increase-specificity` somehow?

src/locales/en-us/common/index.yaml
```diff
@@ -14,6 +14,7 @@ common:
   copySuccessText: 'Copied!'
   copyKubernetesText: 'Copy as Kubernetes'
   copyKubernetesShortText: 'as k8s'
+  copyKubernetesTooltip: 'Copy as Kubernetes'
```
johncowen, line 17: Is this needed? I would have thought that whatever text we use in the tooltip should be the text that we visually hide? Actually how accessible are our tooltips, do you know?

johncowen (commented):
I left a couple of questions, also I noticed two things:

1. Looks like all our copy buttons now have a little border? I quickly checked here also and I don't see it in my local build but I do on the preview here, let me know if you can't see it and I can check again

Screenshot 2023-09-06 at 18 34 23

2. Do we need another icon just for the tooltip? Shouldn't the toooltip just be on the button?

Screenshot 2023-09-06 at 18 35 16

src/locales/en-us/common/index.yaml
kleinfreund: What’s the point of the tooltip if it’s the exact same content as the button text?

kleinfreund:
> 1\. Looks like all our copy buttons now have a little border? I quickly checked here also and I don't see it in my local build but I do on the preview here, let me know if you can't see it and I can check again

I’ve fixed that.

> 2\. Do we need another icon just for the tooltip? Shouldn't the toooltip just be on the button?

I’ve changed this so that the tooltip wraps around the button. This is a little nicer, I think.

src/locales/en-us/common/index.yaml
johncowen: > What’s the point of the tooltip if it’s the exact same content as the button text?

Maybe I'd misunderstood but isn't the `copyKubernetesText` visually hidden? This is also why I was asking about the accessibility of the tooltip, if its accessible enough theres an argument to say we should try to avoid it saying "Copy as Kubernetes Copy as Kubernetes" (i.e. repeating the same text twice).

I know we aren't super hot on accessibility at the moment, but I think we are starting to get to a point where we can begin to pay more attention to these sorts of things.

I think for the moment, is it worth just reusing the `copyKubernetesText` for the text for the toolip? I suppose that is the root of my querying.

kleinfreund: I’ve changed this around.

johncowen:
> > 1. Looks like all our copy buttons now have a little border? I quickly checked here also and I don't see it in my local build but I do on the preview here, let me know if you can't see it and I can check again

> I’ve fixed that.

Did you push all this? I'm still seeing it on the preview site as it per the screengrabs above?

kleinfreund:
Huh. I’ve pushed again and I’m still not seeing it here. 🤔

johncowen:
Yeah lets wait a bit, GH took a very long time this morning when I did a `git pull` on something else, I think theres weirdness happening, I'll come back in a bit, or ping me if you notice its here.

kleinfreund:
Oh, I see. I accidentally created this PR on `origin` and I’m pushing to my fork. 😓

kleinfreund:
Closing this in favor of https://github.com/kumahq/kuma-gui/pull/1449.

PR #1449, opened 2023-09-07
chore: tweaks copy as k8s button appearance

kleinfreund:
Adds a bordered style to the "Copy as Kubernetes" button to visually separate it from the default copy button a bit better.

Adds a tooltip to the "Copy as Kubernetes" button to explain its purpose.

### DIFF
--- a/src/app/common/ResourceCodeBlock.vue
+++ b/src/app/common/ResourceCodeBlock.vue
@@ -8,17 +8,28 @@
     :code-max-height="props.codeMaxHeight"
   >
     <template #secondary-actions>
-      <CopyButton
-        :get-text="getYamlAsKubernetes"
-        :copy-text="i18n.t('common.copyKubernetesText')"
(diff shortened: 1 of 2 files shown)


src/app/common/ResourceCodeBlock.vue
```diff
@@ -76,3 +87,15 @@ function toYamlRepresentation(resource: Entity): string {
   return toYaml(resourceWithoutTimes)
 }
 </script>
+
+<style lang="scss">
```
johncowen, line 91: If these aren't scoped, should we manually scope them by giving this component a class of `resource-code-block`?

kleinfreund: I don’t see a need to make the scoping of the selectors in use here more tight. They’re specific enough to not risk conflicts.

johncowen (approved):

PR #1450, opened 2023-09-07
feat: updates create zone appearance

kleinfreund:
**feat(zones): shows confirm modal when exiting zone creation mid process**

Shows a confirmation modal when exiting the Zone creation process while (1) a Zone with a token was already create and (2) the Zone wasn’t connected, yet.

**feat: updates Create Zone wizard layout**

Updates the layout of the Create Zone wizard to split the form into sections for Zone name, configuration, connecting a Zone, and waiting for a Zone to be connected.

### DIFF
--- a/features/zones/Create.feature
+++ b/features/zones/Create.feature
@@ -3,6 +3,9 @@ Feature: Zones: Create Zone flow
     Given the CSS selectors
       | Alias                               | Selector                                             |
       | zone-nav-item                       | .app-sidebar > .nav-item:nth-child(2) > a            |
+      | exit-button                         | [data-testid='exit-button']                          |
+      | confirm-exit-modal                  | [data-testid='confirm-exit-modal']                   |
+      | confirm-exit-button                 | [data-testid='confirm-exit-button']                  |
       | name-input                          | [data-testid='name-input']                           |
       | name-input-invalid-dns-name         | $name-input[data-test-error-type='invalid-dns-name'] |
       | create-zone-button                  | [data-testid='create-zone-button']                   |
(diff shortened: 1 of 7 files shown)


src/app/zones/components/ZoneCreateKubernetesInstructions.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div>
-    <h3>1. {{ i18n.t('zones.form.kubernetes.prerequisites.title') }}</h3>
+    <h3>
+      <span class="step-number">1</span>
```
johncowen, line 4: Probably not for this PR, but it feels like this should be an order list instead of headers?

src/app/zones/components/EntityScanner.vue
```diff
@@ -3,7 +3,7 @@
     <div class="scanner-content">
       <KEmptyState cta-is-hidden>
         <template #title>
-          <div class="mb-2">
+          <span class="mr-1">
```
johncowen, line 6: Can we start to try to avoid using these helper classes? At the very least it seems strange to me that sometimes we use these and sometimes we use the style node  (for example the margin-right below https://github.com/kumahq/kuma-gui/pull/1450/files#diff-3bc3c5b294b392b43bea19c02637aecd234549d4ce0dc204cb399f291eb33c3eR178).

features/zones/Create.feature
```diff
@@ -148,3 +151,68 @@ Feature: Zones: Create Zone flow
...
+        zoneInsight:
+          subscriptions: []
+      """
+
+    When I visit the "/zones/create" URL
+    And I "type" "test" into the "$name-input" element
+    And I click the "$create-zone-button" element
+
+    Then the "$instructions" element exists
+    Then the "$zone-connected-scanner" element contains "Waiting for Zone to be connected"
```
johncowen, line 209: I'm kinda assuming this will never change due to subscriptions being zero length, is that right?

johncowen (approved):
Looks good! I left a couple of comments, I think only the stye on is potentially actionable but up to you.

src/app/zones/components/EntityScanner.vue
kleinfreund: Why? They’re convenient. What’s strange about using the utility for situations where nothing else is required and a property declaration for when there already happens to be a CSS rule set?

features/zones/Create.feature
kleinfreund: No, the scanner can also be in the waiting state.

I’m using these assertions as a kind of insurance for the test having enough time to actually see this UI before trying to click Exit. When working on this, I occasionally had scenarios where these tests failed because the test clicked Exit too early (i.e. in the short gap between clicking the create button and the conditions for changing the exit button interaction). If the scanner shows the waiting state, these conditions are up-to-date, too.

src/app/zones/components/EntityScanner.vue
johncowen: Well you've almost pointed out one reason yourself. If I want to add more styles, then there will already be a CSS ruleset to add to.

Also, it means styles are always defined in the same place, in style nodes not bits in classes and bits in style nodes.

Don't worry about it for now.

features/zones/Create.feature
johncowen: Not sure I follow but I think from what you said the assumption was right, subscriptions=0 seems to mean that the zone will never be connected and therefore when the text clicks the exit button we are guaranteed to trigger the warning. Feel free to resolve.

kleinfreund: That’s not the case: the scanner isn’t immediately in the complete/success state but first in the loading state. At that point, the condition for triggering the confirm modal is _also_ not in effect yet and we would exit directly without the confirmation.

johncowen: I think we are talking about different things here. I am talking about after this message is visible. You are talking about before the message it visible.

kleinfreund: Okay so once it switched to the waiting state, it will remain in that state on account of the zero subscriptions.

src/app/zones/components/ZoneCreateKubernetesInstructions.vue
kleinfreund: Not sure if there’s much gained of markup up a big list of instructions as an `ol`, but I feel indifferent about it so we could change this.

johncowen: Yeah sure, not for this PR. But more generally, if you are keen on avoiding things like `[role=button]` then we should be keen on doing things like this also.

PR #1452, opened 2023-09-07
chore: Adds YAML undefined type to aid testing

johncowen:
See https://github.com/kumahq/kuma-gui/pull/1316#discussion_r1293326627

We had a bit of discussion a while back on the inability to omit a property from a merge (i.e. mark it as undefined). This adds a `!!js/undefined` YAML tag to let you do that.

I don't think we'll ever need the ability to specify between omitted vs set to `undefined` but if we do we can just add a new tag to cope for that.


kleinfreund (approved):
I’m approving this, but I think it’s the wrong approach still. Tests that need this degree of control over their test data should use complete mock responses that aren’t merged with a default mock response. It shouldn’t be required to have knowledge of the merging algorithm to understand how a test’s data will actually look like.

PR #1455, opened 2023-09-08
chore: Remove store/localStorage usage from onboarding

johncowen:
Removes store/localStorage usage from onboarding.


johncowen:
Ah thanks for the nudge @kleinfreund , rebased!

kleinfreund (approved):

PR #1459, opened 2023-09-11
refactor: uses plural resource names in URL paths

### LINKED ISSUE #1355: Standardize URL paths
### Description

We should standardize our URL paths across the UI to follow one pattern as we’re currently having a bit of a mix.

Generally, there are two options:

- Always use plural. For example:
  - `/zone-cps`
  - `/zone-cps/:mesh`
  - `/meshes`
  - `/meshes/:mesh`
  - `/meshes/:mesh/dataplanes`
  - `/meshes/:mesh/dataplanes/:name`
  - `/meshes/:mesh/services`
  - `/meshes/:mesh/services/:name`
  - etc.
- Use plural for list views; singular for detail views. For example:
  - `/zone-cps`
  - `/zone-cp/:mesh`
  - `/meshes`
  - `/mesh/:mesh`
  - `/mesh/:mesh/dataplanes`
  - `/mesh/:mesh/dataplane/:name`
  - `/mesh/:mesh/services`
  - `/mesh/:mesh/services/:name`
  - etc.

kleinfreund:
## Introduction

In order to verify the approach of using plural resource names in URL paths together with the idea for using summary drawers in list views that are hooked up to their own route, I did the following:

- Added a change that changes all routes to use plural resource names in the URL paths. The changed paths are listen below. I also wanted to change `/data-planes` to `/dataplanes` because its more consistent with established names, but I left this out for now.
- Added a _temporary_ change that changes the behavior for clicking a table row to no longer navigate to the detail view but emit a `row:click` event. This will be used to demonstrate that selecting a row to navigate to the summary drawer route is viable.
- Added a _temporary_ change that adds a prototypical implementation of a summary drawer component for the service list view. This isn’t feature-complete. For example, using a table’s pagination won’t currently clear the selected item.

If we’re happy with how this works, the commits labeled as _temporary_ above will be removed again as they relate to implementing the summary drawer and not the plural resource names in route paths.

Resolves #1355.

## Changes

**refactor: uses plural resource names in URL paths**

Uses plural resource names in URL paths and uses `/overview` as the final path segment for detail views. The following URL paths were changed:

| Before                                  | Now                                                  |
| --- | --- |
| `mesh`                                  | `meshes`                                             |
| `mesh/:mesh/data-plane/:dataplane`      | `meshes/:mesh/data-planes/:dataplane/overview`       |
| `mesh/:mesh/gateway/:dataplane`         | `meshes/:mesh/gateways/:dataplane/overview`          |
| `mesh/:mesh/policy/:policyPath/:policy` | `meshes/:mesh/policies/:policyPath/:policy/overview` |
| `mesh/:mesh/service/:service`           | `meshes/:mesh/services/:service/overview`            |
| `zones/zone-cps/:zone`                  | `zones/zone-cps/:zone/overview`                      |

**feat: adds support for marking a selected row** (**REMOVED**, was only present during verification of approach)

Adds support for marking a selected row with `AppCollection` based on a row’s `id` field.

Adds a new event `row:selected` to `AppCollection` that is triggered both on row clicks and when a default selected row is determined.

**chore(services): adds prototype for summary route** (**REMOVED**, was only present during verification of approach)

### DIFF
--- a/features/application/MainNavigation.feature
+++ b/features/application/MainNavigation.feature
@@ -56,7 +56,7 @@ Feature: application / MainNavigation
     And the URL contains "page=2&size=50"
     And the "[data-testid='page-2-btn'].active" element exists
 
-    When I visit the "/mesh/default/data-planes" URL
+    When I visit the "/meshes/default/data-planes" URL
     And the URL contains "page=1&size=50"
     And the URL doesn't contain "mesh=default"
 
--- a/features/application/Titles.feature
+++ b/features/application/Titles.feature
@@ -18,25 +18,25 @@ Feature: application / titles
       | /onboarding/dataplanes-overview | Data plane overview |
       | /onboarding/completed           | Completed           |
 
-      | /zones/create                           | Create & connect Zone |
-      | /zones/zone-cps                         | Zone Control Planes   |
-      | /zones/zone-cps/zone-cp-name            | zone-cp-name          |
-      | /zones/zone-ingresses                   | Ingresses             |
-      | /zones/zone-ingresses/zone-ingress-name | zone-ingress-name     |
-      | /zones/zone-egresses                    | Egresses              |
-      | /zones/zone-egresses/zone-egress-name   | zone-egress-name      |
+      | /zones/-create                                   | Create & connect Zone |
+      | /zones/zone-cps                                  | Zone Control Planes   |
+      | /zones/zone-cps/zone-cp-name/overview            | zone-cp-name          |
+      | /zones/zone-ingresses                            | Ingresses             |
+      | /zones/zone-ingresses/zone-ingress-name/overview | zone-ingress-name     |
+      | /zones/zone-egresses                             | Egresses              |
+      | /zones/zone-egresses/zone-egress-name/overview   | zone-egress-name      |
 
-      | /mesh         | Meshes        |
-      | /mesh/default | Mesh overview |
+      | /meshes         | Meshes        |
+      | /meshes/default | Mesh overview |
 
-      | /mesh/default/services             | Services     |
-      | /mesh/default/service/service-name | service-name |
+      | /meshes/default/services                       | Services     |
+      | /meshes/default/services/service-name/overview | service-name |
 
-      | /mesh/default/gateways             | Gateways     |
-      | /mesh/default/gateway/gateway-name | gateway-name |
+      | /meshes/default/gateways                       | Gateways     |
+      | /meshes/default/gateways/gateway-name/overview | gateway-name |
 
-      | /mesh/default/data-planes                | Data Plane Proxies |
-      | /mesh/default/data-plane/data-plane-name | data-plane-name    |
+      | /meshes/default/data-planes                          | Data Plane Proxies |
+      | /meshes/default/data-planes/data-plane-name/overview | data-plane-name    |
 
-      | /mesh/default/policies/circuit-breakers         | Policies  |
-      | /mesh/default/policy/circuit-breakers/program-0 | program-0 |
+      | /meshes/default/policies/circuit-breakers                    | Policies  |
+      | /meshes/default/policies/circuit-breakers/program-0/overview | program-0 |
--- a/features/mesh/Index.feature
+++ b/features/mesh/Index.feature
@@ -24,7 +24,7 @@ Feature: mesh / index
   Scenario: Clicking a mesh and back again for <Mesh>
     Then the "$item" element exists 2 times
     Then I click the "<Selector>" element
-    Then the URL contains "/mesh/<Mesh>"
+    Then the URL contains "/meshes/<Mesh>"
     And the "$breadcrumbs" element contains "Meshes"
 
     Then I click the "$navigation li:nth-child(2) a" element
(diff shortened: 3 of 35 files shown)


kleinfreund:
I’m opening this op for review so we can have a look at whether this approach is viable/satisfying. See PR description for more information.

johncowen (commented):
Just wanted to leave a quick message here as I looked at this the other day from a user perspective and it all looked great 👏 . I can't see why we can't use this approach, I did think that putting the tabs somewhere could have been an issue but it doesn't look like it.

I'm yet to take a deeper look, but wanted to leave some initial comment

johncowen:
Wanted to leave a quick note on this one that generally this approach looks really good, adn we are likely to go forwards with this approach, wanted to say that this isn't a priority right now and thats why we are leaving this hanging for the moment, it will be picked back up soonish.

johncowen:
Something additional I thought of here, which we should at least recognise.

If we had extra actions for an "entity" what would we do there, for example:

If we can register/create something from the UI:

```
/meshes/:mesh/some-things/....
```

we can't use 

```
/meshes/:mesh/some-things/create
```

We could do the DNS trick, if that is ok for the entity we are concerned with

```
/meshes/:mesh/some-things/-create
```

But we don't know if counting on DNS will be ok, plus ideally I think we wanted to avoid this trick.

wdyt?

kleinfreund:
It would probably be the easiest (read: least ambiguous) if we created such routes _next_ to the list ones:

- /services (List view)
- /services/:service (List view with summary)
- /services/:service/overview etc. (Detail views)
- /create-service (Create view)
- /services/:service/update (Update view)

Having `/services/create` would be nice, but I don’t really mind sacrificing this.

johncowen:
On this one:

```
/services/:service/update (Update view)
```

I would have assumed that update views would be the same as the "view" views, not sure if we'd ever need a "view" separate from an "update", they'd probably just use the same I think.

Me personally this is my least favourite option:

```
/create-service (Create view)
```

My preference ideally would be:

1. `/services`, `/services/create`, `/service/:name` (I know the point of this is to see if we can avoid the mix of plural singular without any issues)
2. `/services`, `/services/-create`, `/services/:name` (has the "what if we can't rely on the DNS trick)
3. `/services`, `/create-service`, `/service/:name`. For me this kinda spoils the point of trying to not use both `services` and `service` (i.e. singular plus plural). I wonder what would happen if we needed more than 1 'action' (not just create)

I get there are many reasons for wanting to do this, so this is just IMO rather than anything.

kleinfreund:
> I would have assumed that update views would be the same as the "view" views, not sure if we'd ever need a "view" separate from an "update", they'd probably just use the same I think.

They will definitely be separate views. An update view for something (and we might not have something like this in Kuma at all) would have a form for changing all changeable things of a resource. The detail view doesn’t need any of that. They shouldn’t mix.

kleinfreund:
I’ve removed the two commits implementing the prototypical approach to summary drawers so that this PR can be properly reviewed now.

johncowen:
Hey, just to confirm here, this is based on using option 2 from above?

> 2. `/services`, `/services/-create`, `/services/:name` (has the "what if we can't rely on the DNS trick?" problem)

With any "non-tab-like" "actions" being hyphen prefixed such as `/services/-create`, `/services/-action` etc?

kleinfreund:
> With any "non-tab-like" "actions" being hyphen prefixed such as `/services/-create`, `/services/-action` etc?

No, I haven’t changed this in this PR, yet. I can do that for zones now. There aren’t any other action routes.

kleinfreund:
> > With any "non-tab-like" "actions" being hyphen prefixed such as `/services/-create`, `/services/-action` etc?
> 
> No, I haven’t changed this in this PR, yet. I can do that for zones now. There aren’t any other action routes.

I’ve added this now.

johncowen:
One other thing I just realized here (which I don't think matters as I'm pretty sure its a nitpick, but good to be mindful of).

```
/mesh/default/service/firewall-0_external/config <-- detail only view with config tab
# and
/mesh/default/service/firewall-0_external/config <-- list plus summary view plus sub config tab/view
```

Are we likely to want a "summary drawer with a config tab"? Even if we do, do we mind calling it something different to get around this?

I'm guessing that if we did want sub views in a summary view (I know, unlikely!) we would still be fine as long as we didn't use the same name for an already existing "detail only" sub/tab view? (guessing the answer here is "yes, its fine", but just checking)

Other than that I think this is good to go! lmk

kleinfreund:
We absolutely should avoid sub-classing list views with even more sub routes beyond the summary. Even the summary route should really use a query parameter rather than a path segment. The only reason we need this complex routing scheme in the first place is the inability to easily update query parameters in vue-router without triggering a navigation.

slonka:
@johncowen - when you get a chance can you review this? Do you need any help around this?

johncowen:
Unless I need the work here for the current thing I'm looking at (not totally sure right now), I was just gonna leave this for next week. Also It needs a bunch of rebasing, and if it can wait until next week, ideally the original author would best doing that. I can approve in the meantime I suppose, 1 sec.

johncowen (approved):

PR #1476, opened 2023-09-13
chore(zones): adds more content hooks for zone creation

kleinfreund:
**chore: adds text gradient utility**

Adds a text-gradient utility class and uses it for the Zone creation title.

**chore(zones): adds content hooks for descriptions and facts to creation view**

### DIFF
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -74,15 +74,22 @@ zones:
     create: 'Create Zone'
   form:
     exit: 'Exit'
+    title: 'Create & connect Zone'
+    description: ' '
+    facts: []
     section:
       name:
         title: 'Zone name'
+        description: ' '
       configuration:
         title: 'Configuration'
+        description: ' '
       connect_zone:
         title: 'Connect Zone'
+        description: ' '
       scanner:
         title: 'Waiting for Zone to be connected …'
+        description: ' '
     nameLabel: 'Name'
     name_tooltip: "The name must be a valid RFC 1035 DNS name, which means it must start with a letter, be less than 64 characters long, and only contain lowercase letters, numbers, and '-'."
     createZoneButtonLabel: 'Create Zone & generate token'
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -76,7 +76,7 @@ zones:
     exit: 'Exit'
     title: 'Create & connect Zone'
     description: ' '
-    facts: ['One', 'Two', 'Three', 'Four']
+    facts: []
     section:
       name:
         title: 'Zone name'
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -76,7 +76,7 @@ zones:
     exit: 'Exit'
     title: 'Create & connect Zone'
     description: ' '
-    facts: []
+    facts: ['One', 'Two', 'Three', 'Four']
     section:
       name:
         title: 'Zone name'
--- a/src/app/zones/views/ZoneCreateView.vue
+++ b/src/app/zones/views/ZoneCreateView.vue
@@ -41,10 +41,34 @@
         <KCard>
           <template #body>
             <div class="form">
-              <div>
-                <h1 class="form-title">
-                  <span class="text-gradient">{{ t('zones.routes.create.title') }}</span>
-                </h1>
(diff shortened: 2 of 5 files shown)


src/app/zones/views/ZoneCreateView.vue
```diff
@@ -41,17 +41,45 @@
...
-                <h1 class="form-title">
-                  {{ t('zones.routes.create.title') }}
-                </h1>
+              <div class="form-header">
+                <div>
+                  <h1 class="form-title">
+                    <span class="text-gradient">{{ t('zones.form.title') }}</span>
+                  </h1>
+
+                  <p v-if="t('zones.form.description') !== ' '">
```
johncowen, line 50: Would `''` and a length check work here?

src/app/zones/views/ZoneCreateView.vue
```diff
@@ -41,17 +41,45 @@
...
+                    class="fact-list__item"
+                  >
+                    <KIcon
+                      icon="check"
+                      :color="KUI_COLOR_TEXT_SUCCESS"
+                    />
+
+                    {{ fact }}
+                  </div>
+                </div>
```
johncowen, line 71: Bit of a nit but should we use a `ul/li` here instead of `div`s?

src/app/zones/views/ZoneCreateView.vue
```diff
@@ -234,6 +270,10 @@
                     <h2 class="form-section-title">
                       {{ t('zones.form.section.scanner.title') }}
                     </h2>
```
johncowen, line 272: I know its not part of this PR but I noticed that when the zone has been found, it looks really strange that it still says "Waiting for Zone to be connected...", should we just remove this header?

johncowen (commented):
Left a few inlines

src/app/zones/views/ZoneCreateView.vue
kleinfreund: It’s better than not arbitrary having a form section without a header. It’s also unambiguous that the Zone is connected when the widget says so so I don’t see this as a conflict.

src/app/zones/views/ZoneCreateView.vue
kleinfreund: Yep, still the case:

```
Error: [@formatjs/intl Error MISSING_TRANSLATION] Missing message: "zones.form.section.name.description" for locale "en-us", using id as fallback.
```

src/app/zones/views/ZoneCreateView.vue
kleinfreund: It doesn’t look broken to me, no.

It doesn’t make sense to change the wording of a section heading conditionally to some state inside the section.

I’m happy to change the title to something not related to the state, but I’m not removing it because that’s inconsistent with the form layout.

src/app/zones/views/ZoneCreateView.vue
kleinfreund: Changed this.

src/app/zones/views/ZoneCreateView.vue
johncowen: Hmm thats weird we do an `undefined` check for this, oh this one is coming directly from formatjs.

Umm ok. I'd like to be able to specify `''` when we need to if possible. I'd actually like to use the error config that was added to `@kong-ui-public/i18n` a while back, oh and actually a couple of other things now come to think of it. I'll make an issue

Thanks for trying it out tho!

src/app/zones/views/ZoneCreateView.vue
johncowen: Ok maybe lets bring it up with @johnharris85 ? Feel free to resolve for now, we can add an issue for this if folks want to tweak.

src/app/zones/locales/en-us/index.yaml
```diff
@@ -74,15 +74,22 @@ zones:
     create: 'Create Zone'
   form:
     exit: 'Exit'
+    title: 'Create & connect Zone'
+    description: ' '
+    facts: ['One', 'Two', 'Three', 'Four']
```
johncowen, line 79: Not sure if you meant to leave these in or not, will approve anyway, but just incase you didn't notice these

johncowen (approved):
LGTM!

src/app/zones/locales/en-us/index.yaml
kleinfreund: Thanks, removed those.

Issue #1142, opened 2023-06-30
fix(e2e): e2e URL assertions don't seem to be working


johncowen:
### What happened?

I think the regex added here https://github.com/kumahq/kuma-gui/pull/1109 is problematic. Unsure right now whether this just affects the `only` wording, but I think it affects all of them not just `only`.

We should when fixing the regex we should add something to catch if no assertions are made in this step (which is what I think is happening now) so it doesn't go unnoticed in future.

johncowen:
Note: I've since added the ability to use `was not requested with`, so maybe we should remove the `only` step which I'm pretty sure its not used anywhere anymore? The only place we were using it is now commented out, so if we do remove it we can just replace this one place with the `was not requested with` alternative

johncowen:
See https://github.com/kumahq/kuma-gui/pull/1482, which removes this step but does not remove/replace the commented out usage yet.

johncowen:
This is all done and in now

PR #1482, opened 2023-09-14
chore(e2e): Maintain history of requests instead of Cypress request spies

johncowen:
Back when I started adding our Cypress support I spent a lot of time trying to get some sort of request assertion working in a nice way but didn't get there, but I wanted to come back and make it nicer at some point.

> After working with Cypress's intercept and request aliasing this PR I realized that request asserting ideally needs intercept to maintain a history of all requested API calls...
> This PR doesn't take this approach yet and will require a bit of reworking. The method I used here gets us to a point where we can write many more tests and come back to change this up. I want to add more tests and get this into CI before I come back to that though.

https://github.com/kumahq/kuma-gui/pull/828

Up until now occasionally you sometimes have to add extra steps to "tag" a request so Cypress can wait on it. Ideally, as we automatically mock out every single URL so you don't have to, we should just be able to say "tell me this request has happened" without have to tag it first. There were other little issues with this that I can't quite remember the details of.

This PR, instead of using Cypress, maintains a separate history of API requests, and then checks that history or waits on additional entries to that history in order to then perform an assertion as to whether it happened or not (timing out after 4000)

I did a few checks here first to see if the previous request asserting was causing the odd flake or difficulty and it looks like it was, you can see in the last lot of commits where I've added a quick "check for a DOM element first" and then remove that again once this was in. I also checked this was working by dropping a setTimeout in a `KumaApi` call before it made the actual HTTP request, and the new approach would wait until that request had happened before the step would move on, I also made the setTimeout really big to make sure it failed.

I've PRed this now even though I would like to rearrange some things here generally, just as I have in mind that this PR might be immediately beneficial as is, so if we get an approval here and merge I can work on moving things around slightly as we are behind the gherkin layer so my fiddling around in an upcoming PR won't affect anything.

Umm some of the inlines down below now might be out of date as I've gone back and forth thinking exactly what I wanted to do here, I'll resolve any that are out of date and keep unresolved any that are still relevant.

Closes https://github.com/kumahq/kuma-gui/issues/1332


cypress/support/step_definitions/index.ts
```diff
@@ -24,11 +25,11 @@ const YAML = {
 
 let env = {}
 let selectors: Record<string, string> = {}
-let urls = new Map()
+const client = useClient()
 Before(() => {
   env = {}
   selectors = {}
-  urls = new Map()
+  client.history = []
```
johncowen, line 32: I'll likely add a `client.clear()` in a soon to come PR, if we are to use request counting (or anything else to do with a 'client' that we need to reset/clear) we'll need to clear/reset other things too.

cypress/support/step_definitions/index.ts
```diff
@@ -129,84 +128,57 @@ Then('the URL contains {string}', (str: string) => {
...
-            if (exact) {
-              expect(Object.keys(xhr.request.query[key]).length).to.equal(searchParamsEntries.length)
-            }
-
-            break
           }
-        }
-      },
-    )
-  })
```
johncowen, line 209: See https://github.com/kumahq/kuma-gui/issues/1142. The places where did use this where broken due to this step not working, and therefore commented out. These currently commented out tests will be replaced with the `was not request with` variant that was added recently.

features/mesh/gateways/Index.feature
```diff
@@ -59,6 +59,7 @@ Feature: mesh / gateways / index
   Rule: The Gateway listing can filter gateways by type
 
     Scenario: No filtering
+      Then the "$item" element exists 2 times
```
johncowen, line 62: This ensures that that we wait until enough rendering has finished (but no longer) to assert whether the below request was called.

features/mesh/services/Item.feature
```diff
@@ -96,40 +90,26 @@ Feature: mesh / services / item
       Then the "$input-search" element isn't disabled
       And I wait for 500 ms
       When I "type" "name:a-service protocol:tcp" into the "$input-search" element
-      And the URL "/meshes/default/dataplanes+insights" responds with
-        """
-        """
```
johncowen, line 101: These weird empty "set up the Cypress spy" steps are no longer needed as we maintain a full history of requests made during the test and use that for looking for the request we are interested in.

features/mesh/services/Item.feature
```diff
@@ -141,9 +121,6 @@ Feature: mesh / services / item
           offset: 0
           size: 50
         """
-      And the URL "/meshes/default/dataplanes+insights" responds with
-        """
-        """
       And I click the "$button-clear-search" element
       Then the URL "/meshes/default/dataplanes+insights" was requested with
         """
```
johncowen, line 127: The changes above here look weird in the diff, almost like tests were removed, but if you look at the files we still cover the same thing.

src/services/e2e.ts
```diff
@@ -21,6 +21,9 @@ const $ = {
   mockServer: token('mockServer'),
   mock: token<Mocker>('mocker'),
   Env: token('Env'),
+  client: token<{
+    history: HistoryEntry[]
+  }>('client'),
```
johncowen, line 26: This will become a more formally defined 'client' in a soon to come PR, I want to investigate a few things before I do that tho.

src/services/e2e.ts
```diff
@@ -41,6 +44,15 @@ export const services = <T extends Record<string, Token>>(app: T): ServiceDefini
...
   }],
+  [$.client, {
+    service: () => {
+      return {
+        history: [],
+      }
+    },
+    arguments: [
+    ],
+  }],
```
johncowen, line 55: Again this is likely to be moved to a more formal 'client' once the "is request counting advantageous or not" is clearer.

src/test-support/intercept.ts
```diff
@@ -44,6 +48,7 @@ export const mocker = (env: (key: AppEnvKeys, d?: string) => string, cy: Server,
     // changing to `/`
     path = path === '*' ? '/' : path
     const baseUrl = env('KUMA_API_URL')
+    const client = get(TOKENS.client)
```
johncowen, line 51: Again in the "much spoken about soon to come PR" this is likely to be injected properly instead of using `get`

johncowen (commented):
Just left some inline notes while I'm experimenting with this in case folks look whilst in draft.

Top-level thinking here is in this case I don't think I want to use Cypress `should` for request asserting. So the removal of that is on purpose whilst I'm experimenting. My thinking is either:

1. We should always assert for something that results from the HTTP request first, like a DOM element that can only be there is the request has happened, then we only use the more fine grained request assertions to check for query params etc. (see last commit here)
2. I'm not totally sure request counting is going to work out, and I would like to look into it a bit more before I discount it. If it does work out, I would like to be able to say "just wait until the page has stopped loading.

Extra bit of info here, the `wasn't` assertions are the ones we need to think more about. If we know when the page is idle we can do a proper check to ensure a request **wasn't** made. Until the page is idle I'm not sure if we could reliable use a Cypress `should` for that without using a timeout.

Anyway, just using this PR to get flakey CI tests while I think about it/look into it (its harder to produce flakes locally)

cypress/support/step_definitions/index.ts
johncowen: This one is done

features/mesh/gateways/Index.feature
johncowen: Once I'd reimplemented the request asserting this was one I could remove.

src/services/e2e.ts
johncowen: I've now made a more formal `client`

src/test-support/intercept.ts
johncowen: Ah not done this yet, but will probably do in the follow up rather than here.

src/test-support/client.ts
```diff
@@ -0,0 +1,78 @@
...
+type Request = {
+  url: string,
+  method: string,
+  searchParams: Record<string, string>,
+  body: Record<string, unknown>
+}
+type Listener = {
+  request: Request
+  resolve: (bool: boolean) => void
+}
```
johncowen, line 17: Some of these types are a bit awkward and I might be able to combine two of them, I'll look at that in a follow up.

kleinfreund (commented):
How good do think are the chances that request counting will do what we expect?

johncowen:
> How good do think are the chances that request counting will do what we expect?

We aren't doing request counting (I think we've spoken about "why not" previously a bit) we are waiting for the specific request.

> How good do think are the chances <this PR> will do what we expect?

Well I dunno, its seems to be an improvement over what was there previously. What do you think?

src/test-support/client.ts
```diff
@@ -0,0 +1,78 @@
...
+export default () => {
+  let history: HistoryEntry[] = []
+  let listeners: Listener[] = []
+  return {
+    reset: () => {
+      history = []
+      listeners = []
+    },
+    request: (request: HistoryEntry) => {
+      const found = listeners.filter((item: Listener) => equals(item.request, request))
```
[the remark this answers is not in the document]
johncowen: Ah yes, sorry I typed this up at the end and moved things around a bit afterwards. This isn't needed at all now, I'll remove it, thanks!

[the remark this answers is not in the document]
johncowen: This is done here.

kleinfreund (approved):

johncowen:
@kleinfreund FYI I made a little tweak in https://github.com/kumahq/kuma-gui/pull/1482/commits/f56a7173dcf423c2787309715c9c836325eda997 for potential deep HTTP bodys, if you wanted to look at that.

johncowen:
Gonna merge this one

PR #1483, opened 2023-09-15
chore(zones): tweaks create UI content formatting

kleinfreund:
## Changes

Changes the base ul and ol padding so that it doesn’t overflow its content container narrowly.

Overrides the KCard body font size specifically for forms where we expect general body text to have the same font size (`$kui-font-size-40`).

Tweaks the gaps between various form elements and groups to make it more clear which elements belong together in some way or another.

Stops using `k-input-label` as a styling utility as it will be removed in a future version of Kongponents.

## Screenshots

**Before**:

image

**After**:

image

### DIFF
--- a/src/app/zones/components/ZoneCreateKubernetesInstructions.vue
+++ b/src/app/zones/components/ZoneCreateKubernetesInstructions.vue
@@ -1,11 +1,11 @@
 <template>
   <div>
-    <h3>
-      <span class="step-number">1</span>
+    <h3 class="form-step-title">
+      <span class="form-step-number">1</span>
       {{ i18n.t('zones.form.kubernetes.prerequisites.title') }}
     </h3>
 
-    <ul>
+    <ul class="instruction-list">
       <li>
         <b>{{ i18n.t('zones.form.kubernetes.prerequisites.step1Label') }}{{ props.zoneIngressEnabled ? ' ' + i18n.t('zones.form.kubernetes.prerequisites.step1LabelAddendum') : '' }}</b>:
         {{ i18n.t('zones.form.kubernetes.prerequisites.step1Description', { productName: i18n.t('common.product.name') }) }}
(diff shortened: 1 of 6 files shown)


src/assets/styles/_forms.scss
```diff
@@ -58,10 +63,47 @@
   font-size: $kui-font-size-50;
 }
 
+.field-group-list > *+* {
```
johncowen, line 66: nit/question: Do you know if there a lint rule to always give us spaces around these things?

```css
.field-group-list > * + * {
}

```

src/assets/styles/_forms.scss
```diff
@@ -58,10 +63,47 @@
...
+
+.form-step-number {
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  width: 30px;
+  height: 30px;
+  margin-right: $kui-space-20;
+  color: $kui-color-text-inverse;
+  background-color: #169fcc;
```
johncowen, line 106: I know we kinda spoke about this before, but what would you think of having `--kuma-blue-???` etc etc and using those for any random hex codes we have. Even if they are just one off instances to my eyes it makes things easier, plus it means we would have a central list of these random colors in a single place.

johncowen (commented):
LGTM, mainly questions and nits, lmk either way and I can 👍

src/assets/styles/_forms.scss
kleinfreund: I don’t.

src/assets/styles/_forms.scss
kleinfreund: I’ve moved this into a variable.

src/app/zones/components/ZoneCreateKubernetesInstructions.vue
```diff
@@ -77,14 +77,14 @@
...
-      <span class="step-number">4</span>
+    <h3 class="form-step-title">
+      <span class="form-step-number">4</span>
       {{ i18n.t('zones.form.kubernetes.connectZone.title') }}
     </h3>
 
     <p>{{ i18n.t('zones.form.kubernetes.connectZone.configDescription') }}</p>
 
-    <span class="k-input-label mt-4">
+    <span class="field-group-label mt-4">
```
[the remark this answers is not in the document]
johncowen: Well you are already kind of doing this coupling by prefixing everything with `.form-`, or `.field-`

`.form-something` is that not that far from `.form { .something {}}` > `.form { ul {}}`

But no matter, it was just a suggestion at present.

src/assets/styles/_forms.scss
johncowen: I made an issue https://github.com/kumahq/kuma-gui/issues/1484

src/assets/styles/_variables.scss
```diff
@@ -31,6 +31,7 @@
 
   // MISC
   --TextGradientBackground: linear-gradient(90deg, #473cfb 0%, #a300bd 33.17%);
+  --StepBackground: #169fcc;
```
johncowen, line 34: I just noticed this, but how come we are using CapitalCasing for these? Traditionally CSS vars are kebab-case and the majority of them above are. Kinda looks like this CapitalCasing is influenced by old Kongponents, but it seems like even they are moving towards `kui-kebab-case`?

Anyway just a comment here that is a super nit we should probably go back to kebab-case for CSS vars. I guess there might be a lint rule for that if we wanted to be lint-y about it.

1 sec gonna approve this now, no action here.

johncowen (approved):

src/assets/styles/_variables.scss
kleinfreund: Yeah, I think we can rename these at some point.

PR #1485, opened 2023-09-15
feat(zones): adds action menu to detail view

kleinfreund:
Adds an action menu to the Zone detail view that gives access to deleting a Zone.

### DIFF
--- a/src/app/zones/components/ZoneActionMenu.vue
+++ b/src/app/zones/components/ZoneActionMenu.vue
@@ -1,0 +1,86 @@
+<template>
+  <div>
+    <KDropdownMenu
+      button-appearance="creation"
+      :kpop-attributes="props.kpopAttributes"
+      :label="t('zones.action_menu.toggle_button')"
+      show-caret
+      width="280"
+    >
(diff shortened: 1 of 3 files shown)


src/app/zones/components/ZoneActionMenu.vue
```diff
@@ -0,0 +1,86 @@
...
+}
+
+async function deleteZone() {
+  // Intentionally not wrapped in a try-catch block so that the DeleteResourceModal can discover when the operation failed.
+  await kumaApi.deleteZone({ name: props.zoneOverview.name })
+}
+
+function navigateToListView() {
+  router.push({ name: 'zone-cp-list-view' })
+}
```
johncowen, line 85: Just wanted to note that `router` and `kumaApi` access would be better happening in the route component rather than these menu components. I think we've mentioned before that we should try and keep `router` and `data-source` access in the top level route components (I know this isn't a data source but its very similar). Not blocking here, but just wanted to mention.

johncowen (approved):

PR #1486, opened 2023-09-15
tests(zones): Adds e2e Zone deletion test

johncowen:
I noticed that this Jest based test was hiding out in the Zones view folder, which makes it hard to see at a glance that we are testing Zone deletion (and it would be better to test this in a browser rather than a CLI tool), so this PR adds a Zone e2e test instead of a Jest based one.


kleinfreund:
> I noticed that this Jest based test was hiding out in the Zones view folder, which makes it hard to see at a glance that we are testing Zone deletion

Zone deletion does happen from the Zone list view so it seems like the test location is exactly as one would expect, no?

kleinfreund (approved):

johncowen:
> Zone deletion does happen from the Zone list view so it seems like the test location is exactly as one would expect, no?

Maybe I didn't explain properly. Basically I was looking through the `features/` folder to start to get an idea of things we aren't e2e testing, and as this one test isn't in `features/` I was like "oh also remember there is also that odd Zone Deletion test still". So I basically moved this file to `features/` (but change the format also, and made it test in a browser vs CLI)

PR #1488, opened 2023-09-15
tests(onboarding): Moves onboarding tests to browser based e2e tests

johncowen:
Moves onboarding tests to in-browser e2e tests instead of CLI tests (and moves them alongside the rest of our e2e tests)

There are a few additional things that I found here, which I've covered inline.


src/app/onboarding/views/DataplanesOverview.vue
```diff
@@ -1,23 +1,31 @@
...
+                {{ t(`onboarding.routes.dataplanes-overview.header.${item}.title`) }}
+              </template>
+
+              <template
+                #description
+              >
+                <p>{{ t(`onboarding.routes.dataplanes-overview.header.${item}.description`) }}</p>
+              </template>
+            </OnboardingHeading>
+          </template>
```
johncowen, line 28: I made quite a bit of change here so I could test state rather than human text

src/app/onboarding/views/DataplanesOverview.vue
```diff
@@ -134,6 +138,8 @@ async function getAllDataplanes() {
           mesh,
         })
       }
+    } else {
+      shouldRefetch = true
```
johncowen, line 142: I think this was theoretically a bug

src/services/i18n/I18n.ts
```diff
@@ -28,6 +28,9 @@ export default <T extends I18nRecord>(strs: T, env: Env['var']) => {
             throw new Error(`Missing message: "${key}" for locale "en-us", using id as fallback`)
           }
         }
+        if (get(strs, key).length === 0) {
+          return ''
+        }
```
johncowen, line 33: This lets us use `''` in locales files.

johncowen (commented):
Inlines:

src/assets/styles/_base.scss
```diff
@@ -83,6 +83,9 @@ ol {
     margin-top: $kui-space-50;
   }
 }
+p:empty {
+  display: none;
+}
```
kleinfreund, line 88: **Question**: Why was this added? We shouldn’t have empty paragraphs anywhere. If we do, we’re doing something wrong there.

As an aside, this will only work for `<p></p>` but not `<p> </p>` (i.e. paragraphs with a text node).

kleinfreund (changes requested):

src/app/onboarding/views/DataplanesOverview.vue
```diff
@@ -1,23 +1,31 @@
...
-              v-if="description !== null"
-              #description
-            >
-              <p>{{ description }}</p>
-            </template>
-          </onboardingheading>
+          <template
+            v-for="item in [
+              tableData.data.length > 0 ? 'success' : 'waiting'
+            ]"
```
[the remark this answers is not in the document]
johncowen: Stylistically, pretty sure we decided that this style is fine, for the moment at least. There are plenty of places where we've used this style, and its not just code I've added/changed. I think the phrase was "having the data where its used", or similar to that. I'm resolving this one, unless I hear more.

src/assets/styles/_base.scss
johncowen: This is why I made our i18n service allow empty strings `''` instead of `' '`.

Now, if you don't set a i18n string for a paragraph, we don't display the paragraph. Previously we had JS conditionals checking to see if the string was set or not to more or less do the same thing. I can't think of an occasion where you'd want a completely empty `p` tag to be displayed.

src/app/onboarding/views/DataplanesOverview.vue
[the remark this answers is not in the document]
kleinfreund: For the sake of reference: Computed properties are in every way better to handle this.

The other cases where we use this style aren’t comparable because they’re ones where we use DataSource which introduces `data` in a template-scoped manner (i.e. we would need to set a `ref` using `update` and add a computed property on top).

I don’t think it’s really an advantage to have this variable be more narrowly scoped to where it’s used. The component being the scoping boundary for computed properties is almost always completely fine and I’m not aware of any bugs that were introduced by us being able to read a component-scoped variable when we shouldn’t have.

I don’t understand the reluctance to use the core building blocks of Vue (specifically `computed`).

src/assets/styles/_base.scss
kleinfreund: We should definitely use `v-if` directives to avoid rendering empty paragraphs instead.

> I can't think of an occasion where you'd want a completely empty p tag to be displayed.

Yes, so let’s not render them.

src/app/onboarding/views/DataplanesOverview.vue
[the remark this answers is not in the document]
johncowen: There's no reluctance, I don't have a problem using it where it helps, and whilst I don't want to speak for others I think this is echoed by other folks on the team.

I think we've discussed at length that in these simple cases its probably not helping, especially when we are using lots for very simple view logic that in some cases barely changes.

I think you can probably equate this to an argument between:

```javascript
const { title, description, message} = { title: 'Title', description: 'Description', message: 'Message'}
console.log(title)
// vs
const title = 'Title'
const description = 'Description'
const message = 'Message'
console.log(title)
```

Apart from the second example would probably be more a realistic comparison like:

```javascript
console.log(title)

// loads of code
// loads of code
const title = 'Title'
// loads of code
const description = 'Description'
// loads of code
const message = 'Message'
// loads of code
// loads of code
```

So there's one way at least where I think it's worse

src/assets/styles/_base.scss
johncowen: We aren't rendering them, it has `display: none`. They literally don't have any display. It also means we reduce the amount of javascript code/conditional forking code we have to write as this little one-liner works for the entire project from The Outside. Set and forget.

kleinfreund: As long as they are in the DOM, they are rendered. They might not be painted, but rendered they are.

Sure, it would be nice to not need to write conditionals for optional text, but this comes at the cost of leaving empty DOM nodes around. That can come with problems in CSS where, for example, hidden elements participate in flex and grid contexts as items (and so they influence layout).

johncowen: I'm pretty sure whilst that might be the case for non-p tags, for p tags specifically I don't think this is an issue. I also think the cases where we actually specify `''` for a i18n string are incredibly slim.

If it does cause an issue with CSS flex there are other alternatives remove them, worst case, just add the `v-if` for that case. To be honest, we probably shouldn't be using flex layouts with `p` tags anyway, I can't think of a place where we already do or would want to.

As for the "cost" of of leaving the odd (probably single digit across the entire application) HTML tag I don't think that's an issue.

Lastly, this gives people editing i18n strings the opportunity to remove a piece of text (and paragraph) rather than leave a screen-rendered/displayed empty tag that possibly has padding/margins etc.

All in all the benefits of this outweigh any cost.

kleinfreund (approved):

PR #1491, opened 2023-09-18
fix: zone ingress single resource API endpoint being wrong

kleinfreund:

Note: Zone Egress looks correct according to https://kuma.io/docs/2.4.x/reference/http-api/#get-zone-egress (i.e. `/zoneegresses/:zoneEgress`)

### DIFF
--- a/features/zones/DetailViewContent.feature
+++ b/features/zones/DetailViewContent.feature
@@ -3,15 +3,19 @@ Feature: Zones: Detail view content
     Given the CSS selectors
       | Alias                    | Selector                                      |
       | ingress-detail-view      | [data-testid='zone-ingress-detail-view']      |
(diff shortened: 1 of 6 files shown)


johncowen:
Once you have the confirmation on the correct endpoints (https://github.com/Kong/kong-mesh-gui/issues/1097#issuecomment-1722923358) could you check we have a mock files reflecting the correct endpoints (and if not add them), plus add a e2e test for this, ty! I'm gonna approve this, so up to you if you want to merge this and then follow up with those two things in a separate PR.

(Little confused as to what was used to build this out if the endpoints were wrong. It doesn't look like mock endpoints exist for either the correct or incorrect endpoints as yet so its not like we mocked it out wrong)

johncowen (approved):

johncowen (approved):
Looks good, thanks for adding those extra bits!

PR #1492, opened 2023-09-18
chore: tweaks error block appearance

kleinfreund:
Changes the badge list to no longer be absolutely position in order to avoid it overlapping other content.

Puts icon and error title in one row.

Removes `details` element when showing invalid parameters.


johncowen (approved):
LGTM, but probably get @johnharris85 to take a look also.

kleinfreund:
@johncowen Even though we’re not gonna use this component for the Zone create view anymore, it’ll be used for the general error case of most of our views. I’d still like to merge this because I think the layout is a little cleaner. What do you think?

johncowen:
Yeah sounds good to me! Agree the layout like this looks cleaner 👍

PR #1493, opened 2023-09-18
fix(routing): Fix routing so we can replace/redirect queryParams without adding history

johncowen:
We had/have some strange race-y route/URL interactions that are causing confusing issues during testing where user interactions can happen at a much faster rate than usual. This combined with how paging works in AppCollection, meant that the click on a tab could be immediately "overwritten" by AppCollection, and the navigation would be lost.

The fix seems relatively straight forwards at first, but then this broke a little fix we had in place for our history interactions, and therefore led to a different approach for solving that.

I've left a few inline comments that explain further, but let me know if you have any additional questions.

This is only currently half done in order to get feedback on the approach, test won't currently pass for some listings until this is integrated throughout. Right now the MeshListing is the place to see this working.


features/zones/PageNavigation.feature
```diff
@@ -33,8 +33,6 @@ Feature: Zones: Page navigation
       """
 
     When I visit the "/zones" URL
-    # This `cy.wait` stabilizes the test significantly. For some reason, it can happen that the subsequent navigation to via nav tab is never triggered.
-    Then I wait for 1000 milliseconds
```
johncowen, line 37: This `wait` is required because KTable:fetcher will overwrite this click (and subsequent router push) when we pass it our data. Avoiding this overwrite meant that the history e2e started failing. Which led me to fix that our history interactions. i.e. instead of avoiding adding possibly important history entries, we instead avoid adding non-important ones.

src/app/App.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-  <RouteView>
+  <RouteView
+    name="app"
+  >
```
johncowen, line 4: I filled out all the RouteView#names here and made the prop required.

src/app/application/components/route-view/RouteView.vue
```diff
@@ -91,17 +79,72 @@ export type RouteView = typeof routeView
...
+  })
+  newParams = {}
+})
+const routeUpdate = (params: Record<string, string | undefined>) => {
+  newParams = {
+    ...newParams,
+    ...params,
+  }
+  routerPush(newParams)
+}
```
johncowen, line 135: We "save" any attempts to `router.push` here and call the actual `router.push` in a raf-like thing so we don't lose any updates. I guess we'll want to wipe some out/unset things at somepoint, if we do we can use `undefined` to unset things.

src/app/application/components/route-view/RouteView.vue
```diff
@@ -91,17 +79,72 @@ export type RouteView = typeof routeView
...
+  router.push(...args)
+}
+watch(() => props.name, (name) => {
+  if (name !== '' && Object.keys(props.params).length > 0) {
+    router.replace({
+      name,
+      query: cleanQuery(props.params, route.query),
+    })
+  }
+}, { immediate: true })
```
johncowen, line 146: This immediately does a `route.replace` with "default" values if we are missing any of the params defined on `<RouteView :params="">`

src/app/meshes/views/MeshListView.vue
```diff
@@ -1,124 +1,123 @@
...
-      v-slot="{data, error}: MeshCollectionSource"
-      :src="`/meshes?page=${props.page}&size=${props.size}`"
+    <RouteView
+      v-if="me"
+      v-slot="{ route, t }"
+      name="mesh-list-view"
+      :params="{
+        page: 1,
+        size: me.pageSize
+      }"
```
johncowen, line 13: This is the best place to have a look at what this looks like at an application level.

Here we:

1. Load in any user values (eventually from local storage) to use as defaults.
2. Set some default params on `RouteView#params`. If the URL doesn't have things then we `route.replace` them onto it (i.e. we don't add that navigation to history)
3. Under the hood if the URL has these params but they aren't numerical (I'll be adding more simple validation for other things later), then we again fallback to the default.
4. Below this comment we swap out the previous `props.page` for `route.params.page` to take advantage of all this. Most of the green below is because i wrapped the entire thing in a DataSource

src/app/me/sources.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type MeSource = DataSourceResponse<{
+  pageSize: number
+}>
+
+export const sources = () => {
+  return {
+    '/me': async (_params = {}, source: Closeable) => {
+      source.close()
+      return { pageSize: 50 }
```
johncowen, line 13: This will eventually come from localStorage, and/or default to `50`. We'll only use the default `50` in one place, so I left the CONSTANT behind.

johncowen (commented):
Inlines:

src/app/application/components/route-view/RouteView.vue
```diff
@@ -1,5 +1,6 @@
 <template>
```
kleinfreund, line 1: **Issue**: (leaving this here just so we have a resolvable thread instead of plain comments on the PR)

History navigation is breaking in some cases:

1. Go to https://deploy-preview-1493--kuma-gui.netlify.app/gui/zones/zone-cps
2. Navigate to page 2
3. Use browser history back
4. Use browser history forward

Expected: you end up on page 2

Actual: you stay on page 1 (after step 3)

Note: Repeating steps 3 and 4 after the first style will work, presumably because the history entries has `?page=1` already.

johncowen: Please only look at Mesh Listings, nothing else is complete. If we are ok with MeshListings I can go an roll it out everywhere.

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -146,17 +148,26 @@ const slots = useSlots()
...
- * Is triggered via a watcher whenever `props.pageNumber` changes while the last page number that was emitted by KTable (via calls to its `fetcher` prop) is different (i.e. the page number has changed independently of a KTable mechanism).
+ * Is triggered via a watcher whenever `props.pageNumber` changes while the
+ * last page number that was emitted by KTable (via calls to its `fetcher`
+ * prop) is different (i.e. the page number has changed independently of a
+ * KTable mechanism).
  *
- * TODO: If https://github.com/Kong/kongponents/pull/1631 is accepted, this hack can be removed in favor of setting the new prop whenever the page number changes externally.
+ * TODO: If https://github.com/Kong/kongponents/pull/1631 is accepted, this
+ * hack can be removed in favor of setting the new prop whenever the page
+ * number changes externally.
```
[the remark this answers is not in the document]
johncowen: Comments are easier to read like this, a bit like git message wrapping and quite a few other types of plain text tech documents.

src/app/application/components/route-view/RouteView.vue
```diff
@@ -1,5 +1,6 @@
...
   <div
+    v-if="props.name.length > 0"
```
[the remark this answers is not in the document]
johncowen: Ah I think this was left in from when I was trying this out before setting all the `name` for all the routes. I've removed this from here and the similarly below in a watch.

johncowen:
Thanks for taking a look @kleinfreund I addresses those little comments. Do you have any feedback on the overall approach? Do you think the approach is an improvement over the previous one, if so I can continue on the work left here. Seems like the intial comments here are minor?

src/app/me/sources.ts
```diff
@@ -0,0 +1,16 @@
+import { DataSourceResponse } from '@/app/application/services/data-source/DataSourcePool'
+
+type Closeable = { close: () => void }
+
+export type MeSource = DataSourceResponse<{
```
kleinfreund, line 5: **Question**: Why the name? What does it mean?

src/app/meshes/views/MeshListView.vue
```diff
@@ -1,124 +1,123 @@
 <template>
-  <RouteView
-    v-slot="{ route }"
-    name="mesh-list-view"
+  <DataSource
+    v-slot="{ data: me }: MeSource"
+    src="/me"
   >
```
kleinfreund, line 5: **Question**: I’m not sure I follow why we need this. Why can’t we just use the `pageSize` from the route query parameters (and default to a constant)? Why does this need data-sourcing?

src/app/meshes/views/MeshListView.vue
```diff
@@ -1,124 +1,123 @@
...
+                <AppCollection
+                  v-else
+                  class="mesh-collection"
+                  data-testid="mesh-collection"
+                  :headers="[
+                    { label: 'Name', key: 'name' },
+                    { label: 'Actions', key: 'actions', hideLabel: true },
+                  ]"
+                  :page-number="parseInt(route.params.page)"
+                  :page-size="parseInt(route.params.size)"
```
kleinfreund, line 45: **Issue**: This will yield to `NaN` values in the case of parameter absence.

src/app/meshes/views/MeshListView.vue
johncowen: Because we are going to persist this value to localStorage when the user selects it from the menu they don't have to keep doing it every time they come to the page.

src/app/meshes/views/MeshListView.vue
johncowen: The idea is these values should never be absent, if they are then its a bug in RouteView and should be fixed there.

src/app/me/sources.ts
johncowen: Good/commonly used name for profile/user setting type things. Pretty much guaranteed not to clash with anything else that may be added to to kuma in the future (such as `config`, `settings`). I think most dev folks would think "`/me` oh, this is probably about user profile/settings"?

src/app/meshes/views/MeshListView.vue
kleinfreund: But we already know that they will always be absent initially. Easy enough to pass in sensible values from the start.

src/app/me/sources.ts
kleinfreund: I’d prefer something like `UserConfigSource` to make this unambiguous.

src/app/meshes/views/MeshListView.vue
johncowen: Hmm maybe I'm not following or missing something here, the default values are here:

https://github.com/kumahq/kuma-gui/pull/1493/files#diff-0bd076fe74003e778a25db73b4e23416ba30f352d6e97fe4894acad30d018772R10-R13

```javascript
<RouteView
      v-if="me"
      v-slot="{ route, t }"
      name="mesh-list-view"
      :params="{
        page: 1,
        size: me.pageSize
      }"
>
```

`page` defaults to 1 and `size` defaults to whatever is return by `/me` (either a localStorage value or `50`). Thats the idea anyway, let me know if that isn't working for you.

kleinfreund: Any router navigation will go to a URL without query parameters unless we explicitly specify them (which we won’t) so there is always a window where those values can’t be set on the route level.

johncowen: > Any router navigation will go to a URL without query parameters

Yes, and the point of this PR is to redirect/replace that URL (without adding to history) with a URL with the query parameters. Can you explain where the window is when these values can't be set?

src/app/me/sources.ts
johncowen: What if its not configurable, say we want the browser the user is using? `UserConfigSource` breaks down. What if we end up with a concept of multiple users? `UserConfigSource` breaks down.

`/me` is generic enough to cover anything we need this for without naming issues, plus I'd say its pretty recognisable to most folks at a glance.

src/app/meshes/views/MeshListView.vue
kleinfreund: > Can you explain where the window is when these values can't be set?

I did already:

> Any router navigation will go to a URL without query parameters unless we explicitly specify them (which we won’t) so there is always a window where those values can’t be set on the route level.

If you navigate to `/meshes`, for example, then that’s a case where that route won’t have these parameters.

johncowen: Right and then they are immediately set with the defaults? I put a console log in the routeParams computed function in RouteView and it does seem to me like they are always set for the `mesh-list-view` route? Are you not seeing the same?

kleinfreund: Okay, so when we use `route` from RouteView, as long as `props.params` has those defaults, that would work. How can we guarantee `props.params` has those values for list views? Looks like if that’s not set, those values will come through as `NaN`.

Also, this means we _can’t_ use vue-router’s `route` for this data the same way.

johncowen: > Okay, so when we use route from RouteView, as long as props.params has those defaults, that would work.

Cool ok I think you get this now. Glad you think this idea "has legs"/would work.

> How can we guarantee props.params has those values for list views? Looks like if that’s not set, those values will come through as NaN.

As of right now our tests should catch this, I'm pretty sure we never visit a page using `?queryParams=`. But if we think this PR is a good approach, then `route.params.page` wouldn't ever be set unless you gave it a default on the RouteView#params (even if a user decided to just add `?page=something` to the URL). If we do think this is a good approach then this PR (or a quickish follow up) will add the necessary default params to every route, and then add the work to not even set the `route.param` if a default isn't set. Then you would immediately see if you hadn't set the default by accident.

kleinfreund:
From what I can see so far, this approach can work as well. It remains to be seen how it affects the stability of tests once this works for all views.

I’m somewhat concerned about the amount of custom code we find ourselves adding on top of the default routing mechanisms.

johncowen:
> From what I can see so far, this approach can work as well. It remains to be seen how it affects the stability of tests once this works for all views.

I think the most important thing here is that history now works correctly, even for URLs/links containing query parameters. The problematic test that is also fixed here only led me to this.

> I’m somewhat concerned about the amount of custom code we find ourselves adding on top of the default routing mechanisms.

I agree, I'm disappointed that we have to build things like this, these things should "Just Work" or there should be documented mechanisms to allow you to add this easily. Considering that this is 10-20 lines of additional code to add this centrally hidden from the application, I'm not too concerned.

Ok, it sounds like I should go ahead and roll this out?

johncowen:
@kleinfreund and I briefly spoke about this offline, we are gonna roll this out in this PR.

src/app/application/components/route-view/RouteView.vue
```diff
@@ -16,29 +16,15 @@
       :env="env"
       :can="can"
       :route="{
-        update: (params: Record<string, string | undefined>) => {
```
kleinfreund, line 19: **Issue**: (thread unrelated to line of code) Reloading a list view while on page >1 doesn’t work correctly. It always loads page 1.

johncowen: I added a fix and a couple more e2e tests to cover this, lmk!

kleinfreund (approved):

johncowen:
@kleinfreund I need to do a tiny bit more here, see last commit for a little issue I found (https://github.com/kumahq/kuma-gui/pull/1493/commits/3dcecacbacd60ff7ed1cbe147262cfbcaa540d3c), currently can add non-specified path segment route params as query params. I added a broken test for it and I'm going to look at fixing it now.

johncowen:
@kleinfreund this should be better now, you will probably want to look at https://github.com/kumahq/kuma-gui/pull/1493/commits/b2e2e38162ea7a9fd9978e37e3bf4d7592f3c671 onwards (so the last three commits)

Things I added:

1. For parameters to be available on `RouteView route.params` you also need to specify them in `RouteView::params`.
2. I threaded the `params` through `RouteView` > `route.params` so that we get typo checking on `route.params` depending on what you set as `params`.
3. I added a `data-testid` to all `RouteViews` internally using the `RouteView::name`
4. I went through every single RouteView we have and made sure we were using the above things.

There's a fairly significant amount of change here even though its all the same thing. So could you give it another good look over (I know you've approved already but I noticed that extra issue and added a good amount since then, ty!)

kleinfreund (approved):

PR #1494, opened 2023-09-18
feat(zones): add step descriptions

johnharris85:
Adds step descriptions for the zone flow.

### DIFF
--- a/src/app/zones/locales/en-us/index.yaml
+++ b/src/app/zones/locales/en-us/index.yaml
@@ -77,19 +77,19 @@ zones:
     delete_button: 'Delete'
   form:
     exit: 'Exit'
-    title: 'Create & connect Zone'
-    description: ' '
-    facts: []
+    title: 'Create & Connect Zone'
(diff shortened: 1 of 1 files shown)


kleinfreund (approved):

src/app/zones/locales/en-us/index.yaml
```diff
@@ -77,19 +77,19 @@ zones:
     delete_button: 'Delete'
   form:
     exit: 'Exit'
-    title: 'Create & connect Zone'
-    description: ' '
-    facts: []
+    title: 'Create & Connect Zone'
+    description: 'Provision and connect a remote zone to your Kuma Global Control Plane.'
+    facts: ['Observe and manage zone workloads', 'Gain insight into service configuration and stats', 'View and modify mesh policies', '']
```
kleinfreund, line 82: **Issue** (minor): This will render an empty fact item otherwise

```suggestion
    facts: ['Observe and manage zone workloads', 'Gain insight into service configuration and stats', 'View and modify mesh policies']
```

PR #1505, opened 2023-09-19
chore(policies): allows policy type-specific descriptions

kleinfreund:
chore: fix incorrect prop definition

chore(policies): allows policy type-specific descriptions

### DIFF
--- a/src/app/policies/components/PolicyList.vue
+++ b/src/app/policies/components/PolicyList.vue
@@ -46,7 +46,7 @@
                   </PolicyTypeTag>
                 </h3>
 
-                <p>{{ t('policies.collection.description') }}</p>
+                <p>{{ t(`policies.type.${props.currentPolicyType.name}.description`, undefined, { fallback: t('policies.collection.description') }) }}</p>
               </div>
 
               <div class="description-actions">
--- a/src/services/i18n/I18n.ts
+++ b/src/services/i18n/I18n.ts
@@ -4,6 +4,12 @@ import type Env from '@/services/env/Env'
 import { camelCaseToWords } from '@/utilities/camelCaseToWords'
 import { get } from '@/utilities/get'
 
+declare module 'intl-messageformat' {
+  interface Options {
+    fallback: string
+  }
+}
+
 interface I18nRecord {
   [key: string]: I18nRecord | string
 }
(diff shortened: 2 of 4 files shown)


src/app/policies/components/PolicyList.vue
```diff
@@ -46,7 +46,7 @@
                   </PolicyTypeTag>
                 </h3>
 
-                <p>{{ t('policies.collection.description') }}</p>
+                <p>{{ t(`policies.type.${props.currentPolicyType.name}.description`, undefined, { fallback: t('policies.collection.description') }) }}</p>
```
johncowen, line 49: The `undefined` here are the interpolated props right? Does it accept `{}`? Kinda looks better if so.

src/services/i18n/I18n.ts
```diff
@@ -4,6 +4,12 @@ import type Env from '@/services/env/Env'
 import { camelCaseToWords } from '@/utilities/camelCaseToWords'
 import { get } from '@/utilities/get'
 
+declare module 'intl-messageformat' {
+  interface Options {
+    fallback: string
+  }
```
johncowen, line 10: I keep seeing the term `defaultMessage` being used for this, maybe we should use that name?

https://formatjs.io/docs/react-intl/components/#message-descriptor
https://formatjs.io/docs/vue-intl/#using-icu-in-vue-sfc

johncowen (commented):
Couple of minor questions/comment other than that LGTM should be a quick follow up 👍

src/app/policies/components/PolicyList.vue
kleinfreund: It does accept it, but `undefined` is cleaner because it wouldn’t trip a check like `if (values)` and the signature explicitly allows `undefined`.

johncowen: Interested as to why there would be a "trippable" check for `values` somewhere? I take it you mean `Object.values()`? Let me approve this as its not important for the PR.

johncowen (approved):

src/app/policies/components/PolicyList.vue
kleinfreund: I rather meant `values` as in the name of the argument. Libraries often implement checks on which argument to skip and will `{}` would very likely work, why even let it consider it if it understands `undefined`.

johncowen: > Kinda looks better if so.

Generally, as with `null`, the less I see `undefined` (as opposed to typeof ... "undefined") in an application the better.

kleinfreund: But why? `typeof thing === 'undefined'` is just more verbose but functionally the same as `thing === undefined`. Same with `null`.

johncowen: I'm not talking about verbosity, I'm not looking for a discussion between `typeof thing === 'undefined'` and `thing === undefined`. I'm talking about setting things to `null` and `undefined`, I just wanted to point out that I wasn't talking about the string `"undefined"` but the value `undefined`

PR #1521, opened 2023-09-20
chore: improves DeleteResourceModal

kleinfreund:
Adds some additional error information to the DeleteResourceModal.

### DIFF
--- a/src/app/common/DeleteResourceModal.vue
+++ b/src/app/common/DeleteResourceModal.vue
@@ -3,10 +3,8 @@
     :action-button-text="props.actionButtonText"
     :confirmation-text="props.confirmationText"
     :is-visible="props.isVisible"
-    :modal-id="props.modalId"
     :title="props.title"
     type="danger"
-    data-testid="delete-resource-modal"
     @canceled="emit('cancel')"
     @proceed="deleteResource"
   >
@@ -14,13 +12,31 @@
       <slot name="body-content" />
 
       <KAlert
-        v-if="hasError"
+        v-if="error !== null"
         class="mt-4"
         appearance="danger"
         is-dismissible
       >
         <template #alertMessage>
-          <slot name="error" />
+          <template v-if="(error instanceof ApiError)">
+            <p>{{ t('common.error_state.api_error', { status: error.status, title: error.title }) }}</p>
+
+            <ul
+              v-if="error.invalidParameters.length > 0"
+              :data-testid="`error-${error.status}`"
+            >
+              <li
+                v-for="(parameter, index) in error.invalidParameters"
+                :key="index"
+              >
+                <b><code>{{ parameter.field }}</code></b>: {{ parameter.reason }}
+              </li>
+            </ul>
+          </template>
+
+          <template v-else>
+            <p>{{ t('common.error_state.default_error') }}</p>
+          </template>
         </template>
       </KAlert>
     </template>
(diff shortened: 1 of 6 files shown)


src/app/common/DeleteResourceModal.vue
```diff
@@ -6,21 +6,39 @@
     :modal-id="props.modalId"
     :title="props.title"
     type="danger"
-    data-testid="delete-resource-modal"
+    :data-testid="props.modalId"
```
johncowen, line 9: What is `:modal-id` used for?

src/app/common/DeleteResourceModal.vue
```diff
@@ -76,16 +99,20 @@ const emit = defineEmits<{
   (event: 'delete'): void
 }>()
 
-const hasError = ref(false)
+const error = ref<Error | null>(null)
 
 async function deleteResource() {
-  hasError.value = false
+  error.value = null
```
johncowen, line 105: I'm glad `hasError` things are being removed.

src/app/zones/locales/en-us/index.yaml
```diff
@@ -205,9 +205,8 @@ zones:
...
         description: 'You may retry this operation.'
   delete:
-    confirmModal:
+    confirm_modal:
+      proceed_button: 'Yes, delete'
       text1: 'Are you sure you want to delete the Zone {zoneName}?'
       text2: 'This action cannot be reversed.'
-      proceedText: 'Yes, delete'
       title: 'Delete Zone'
-      errorText: 'An unexpected error occurred'
```
johncowen, line 213: I feel like a lot more of this "delete modal" text could just go elsewhere (in common or component.delete-resource-modal)? `Delete {type}`, `Are you sure you want to delete the {type} {name}?` etc.

I would think we want this kinda stuff to be consistent between "entities"?

src/locales/en-us/common/index.yaml
```diff
@@ -19,6 +19,9 @@ common:
     title: 'An error has occurred while trying to load this data.'
     details: 'Details'
     field: 'Field'
+  delete_modal:
+    api_error: 'Error {status}: {title}'
+    default_error: 'An unexpected error occurred'
```
johncowen, line 24: Why does `delete_modal` has specifically defined error messages, surely we want to define this sort of stuff in one place so its consistent?

johncowen (commented):
Mostly looks good to me, I had a question around modal id, and also questions/suggestions around starting to think a little more on reusing common terms so these sorts of messages are consistent, lmk.

src/app/common/DeleteResourceModal.vue
kleinfreund: Not totally sure if we ever used `modalId` on KPrompt (it’s set a few lines above). Perhaps we don’t need this. Let me check.

johncowen: Sorry yeah thats kinda the one I was referring to a few lines up. I was wondering why we can't just pass `data-testid` in from above, rather than have a `props.modalId` but I wasn't sure why the KPrompt needed `:modal-id`

kleinfreund: I’ve removed this and just set `data-testid` as usual.

src/locales/en-us/common/index.yaml
kleinfreund: Not sure I’m following. What do you mean by this?

Note that this object now also contains the messages from https://github.com/kumahq/kuma-gui/pull/1521#discussion_r1331569390.

johncowen: Well aren't these generic/common error messages/text? Or do we only use them in the delete modal?

kleinfreund: These are currently only used in the delete modal. They _are_ pretty generic and could be used elsewhere.

johncowen (approved):

PR #1523, opened 2023-09-20
chore(zones): switches to alert for error in create view

kleinfreund:
## Changes

Switches to using a KAlert at the top of the page in the Zone Create View instead of ErrorBlock.

## Screenshots

**Before**:

image

image

**After**:

image

image


johncowen (approved):
Nice! LGTM! Only thought was is this covered by a test? Looks like theres a test for the text under the input field, do we want to add one for the top alert? (could be a follow up also, but I think its a one-liner)

kleinfreund:
> Looks like theres a test for the text under the input field, do we want to add one for the top alert? (could be a follow up also, but I think its a one-liner)

There are tests covering most of this. Specifically the ones referring to the 409/500 scenarios.

PR #1525, opened 2023-09-20
chore: Add automatic typing to source internals

johncowen:
~I've had various iterations of this sat around for a little while now, and I think this is one I could be happy merging, so whilst I'm still not 100% I want to I wanted to put it up for review to see what feedback there was. So, this is currently in draft but~ looking for feedback.

---

This PR adds automatic types to the `"/services/:name": (params) => { params.name }` sources. The one little tiny downside is that currently you have to use `defineSources`, but you do maintain the key names (or the URIs) in the result type, which is important and why a simpler type definition wouldn't have been good enough for me.

Notes:

- We maintain the keys in the resulting definition, I don't want to lose these by doing something simpler like `Record<string,...>`
- `params` are now automatically inferred to strings based on the URI
- We don't assume things like `'all' | 'delegated' | 'builtin'` for a string that could come from the browser URL. i.e. `params.*` are always just strings not unions. I add a little runtime validation here to use a default if somehow something else ends up in the URL.
- `PaginationParams` are still typed as before, I'll potentially also make these strings, just not thought past that yet.

The plan is to eventually have a similar approach on The Other Side (i.e. the view layer), but both of these approaches should maintain some sort of separation similar to what we currently have, and should be as automatic as possible.

~Anyway as mentioned, I'm only 90% sure I want to merge this,~ but would like to get some other thoughts on it to help me make my mind up (and if we go with it, before I roll it out everywhere)

### DIFF
--- a/src/app/application/services/data-source/index.ts
+++ b/src/app/application/services/data-source/index.ts
@@ -2,6 +2,29 @@ import CallableEventSource from './CallableEventSource'
 import type { Creator, Destroyer } from './DataSourcePool'
 export type { DataSourceResponse } from './DataSourcePool'
 
+type ExtractRouteParams<T extends string> =
+  string extends T
+    ? Record<string, string>
+    : T extends `${infer _Start}:${infer Param}/${infer Rest}`
+      ? {[k in Param | keyof ExtractRouteParams<Rest>]: string}
+      : T extends `${infer _Start}:${infer Param}`
+        ? {[k in Param]: string}
+        : {};
+
+type PaginationParams = {
+  size: number
+  page: number
+  search: string
+}
+
+type ExtractSources<T extends string, K> = {
+  [Route in T]: (params: ExtractRouteParams<Route> & K, source: { close: () => void }) => void
+}
+
+export const defineSources = <T extends string>(sources: ExtractSources<T, PaginationParams>) => {
+  return sources
+}
+
 type Configuration = {
   interval?: number
   retry?: (e: unknown) => Promise<void> | undefined
--- a/src/app/data-planes/sources.ts
+++ b/src/app/data-planes/sources.ts
@@ -88,7 +88,7 @@ export const sources = (api: KumaApi) => {
       filterParams.tag = filterParams.tag.filter((item) => !item.startsWith('kuma.io/service:'))
       filterParams.tag.push(`kuma.io/service:${params.service}`)
 
-      // we use `all` in code to mean "don't specify ?gateway=0 at all" i.e.
+      // we use `all` in code to mean "don't specify `?gateway` in the API URL" i.e.
       // both gateway and sidecars
       const type = params.type === 'standard' ? 'false' : params.type
       const gatewayParams = includes(['delegated', 'builtin', 'false'] as const, type)
--- a/src/app/data-planes/sources.ts
+++ b/src/app/data-planes/sources.ts
@@ -1,36 +1,9 @@
+import { defineSources } from '@/app/application/services/data-source'
 import { DataSourceResponse } from '@/app/application/services/data-source/DataSourcePool'
 import { normalizeFilterFields } from '@/app/common/filter-bar/normalizeFilterFields'
 import type KumaApi from '@/services/kuma-api/KumaApi'
-import type { PaginatedApiListResponse as CollectionResponse, DataPlaneOverviewParameters, ApiKindListResponse as KindCollectionResponse } from '@/types/api.d'
+import type { PaginatedApiListResponse as CollectionResponse, ApiKindListResponse as KindCollectionResponse } from '@/types/api.d'
 import type { DataPlane, DataPlaneOverview as DataplaneOverview, DataplaneRule, MeshGatewayDataplane, SidecarDataplane } from '@/types/index.d'
-
-type CollectionParams = {
-  mesh: string
-}
-
-type DetailParams = CollectionParams & {
-  name: string
-}
-
-type EnvoyDataParams = DetailParams & {
-  dataPath: 'xds' | 'clusters' | 'stats'
-}
-
-type PaginationParams = {
-  size: number
-  page: number
-  search: string
-}
-
-type ServiceParams = {
-  service: string
-}
-
-type DataplaneTypeParams = {
-  type: 'all' | 'standard' | 'delegated' | 'builtin'
-}
-
-type Closeable = { close: () => void }
 
 export type DataplaneSource = DataSourceResponse<DataPlane>
 export type DataplaneOverviewSource = DataSourceResponse<DataplaneOverview>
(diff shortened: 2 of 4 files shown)


src/app/application/services/data-source/index.ts
```diff
@@ -1,12 +1,35 @@
...
+        : {};
+
+type PaginationParams = {
+  size: number
+  page: number
+  search: string
+}
+
+type ExtractSources<T extends string, K> = {
+  [Route in T]: (params: ExtractRouteParams<Route> & K, source: { close: () => void }) => void
```
kleinfreund, line 20: **Comment**: It’s a little unfortunate that this implies that `params` will always have `PaginationParams`.

johncowen: Yeah for sure defo agree there, that being said, theoretically they always could potentially have these I suppose. I think this was half my thinking with not putting these things in the URI for the DataSource in the first place. Overall still thinking about this one, but I get what you are saying here.

(btw, I'll probably change the name of PaginationParams at some point)

johncowen: Wanted to check if you had any more feedback on this one, if not we can decide whether we want to roll this out or not.

kleinfreund: Indifferent about this for now. In general we of course shouldn’t say `params` can potentially have these fields because that might not always be correct, but it might be good enough for now.

johncowen:
@kleinfreund putting this up for proper review now, I think the existing inline comment here can be resolved, but let me know if not.

src/app/data-planes/sources.ts
```diff
@@ -47,70 +20,53 @@ export type MeshGatewayDataplaneSource = DataSourceResponse<MeshGatewayDataplane
...
-      const gatewayParams: DataPlaneOverviewParameters = {}
-      if (type === 'standard') {
-        gatewayParams.gateway = 'false'
-      } else if (type !== 'all') {
-        gatewayParams.gateway = type
-      }
+      const type = params.type === 'standard' ? 'false' : params.type
+      const gatewayParams = includes(['delegated', 'builtin', 'false'] as const, type)
+        ? { gateway: type }
+        : {}
```
kleinfreund, line 69: **Question**: Why does the implementation of this change? Is this a rebase mistake?

src/app/data-planes/sources.ts
```diff
@@ -120,28 +76,24 @@ export const sources = (api: KumaApi) => {
...
-        gatewayParams.gateway = 'false'
-      } else if (type !== 'all') {
-        gatewayParams.gateway = type
-      }
+      // we use `all` in code to mean "don't specify ?gateway=0 at all" i.e.
+      // both gateway and sidecars
+      const type = params.type === 'standard' ? 'false' : params.type
+      const gatewayParams = includes(['delegated', 'builtin', 'false'] as const, type)
+        ? { gateway: type }
+        : {}
```
kleinfreund, line 96: **Question**: Why does the implementation of this change? Is this a rebase mistake?

johncowen: Its not a rebase mistake

The reason behind this is: this is what this PR did previous to the change over in the red half of the diff (remember this was PRed way before now). I updated the code to take into account the changes that were made since September when I originally PRed this.

I did spend a tiny bit more time to think about it during rebase to try to make sure it was still correct, which I think it is? No tests blew up at least, and if I change `standard` to `standrd` for example, the tests do blow up.

Here's my thinking:

---

- `?gateway` can only be `delegated`, `builtin`, `false` (or `true` which I don't think we need)
- `standard` means `false` (i.e. don't give me gateways only 'sidecars')
- Anything else (including `all` which is the string we use to say don't add the query param), then don't add `?gateway` at all.

---

Does that sound right to you? Do you think this is correct?

kleinfreund: I’m following, yes. So in terms of behavior it’s the same as the current implementation only now `gatewayParams` is no longer type-checked against `DataPlaneOverviewParameters`.

Can you please add this again in both places?

johncowen: Here's the type check:

https://github.com/kumahq/kuma-gui/blob/6b453886e8eec507c171134429f5654b01b6865d/src/services/kuma-api/KumaApi.ts#L148

Which checks `params` against `DataPlaneOverviewParameters`

DataSource URLs are just like browser URLs, you can narrow the type check to only certain strings all you like, it doesn't prevent anyone from typing a different thing in the URL apart from the 3 or 4 options you expect. We have this issue URLs in quite a few places still but I'm gradually reducing this issue as I go.

For example with this PR, the following URL no longer sends `?gateway=stadard` to the backend

`/gui/meshes/default/data-planes?page=1&size=50&dataplaneType=stadard`

This is because we do a runtime check on the value first, if its none of the values we expect we just don't pass the `?gateway`. Well, thats the thinking behind it.

Does that sounds right to you? Do you think this is correct?

kleinfreund: That’s right, but this type check is still valuable here because it prevents us from making mistakes like trying to assign `{ gateways: '...' }` to `gatewayParams`. Without the type check, this would just work.

johncowen: Ok great thanks. So wait, where do you want me to add the additional type check?

johncowen: Right, but `DataPlaneOverviewParameters` includes `true`

> `?gateway` can only be `delegated`, `builtin`, `false` (or `true` which I don't think we need)

> `true` which I don't think we need

We don't ever want to pass `?gateway=true` as far as I'm aware?

So I think this is probably more correct?

Screenshot 2023-11-24 at 11 16 53

If you add something to the `includes` that `DataPlaneOverviewParameters` doesn't have it will be type checked by the code I linked to above I think?

Plus as far as I'm concerned its an awful lot clearer and less error prone (see above), obvs my own opinion on the clarity.

I could also just inline the includes if you'd prefer, something like:

```js
      return api.getAllDataplaneOverviewsFromMesh({ mesh }, {
        ...filterParams,
        ...includes(['delegated', 'builtin', 'false'] as const, type) ? { gateway: type } : {},
        offset,
        size,
      })
```

Let me know, maybe I'm missing something else here?

kleinfreund: Well, alright, let’s leave it as in the PR I guess.

We could change `DataPlaneOverviewParameters` to not include `'true'` if we don’t ever care about it and still get the type checking benefits within the source definition.

The type check will only work in some cases. It won’t work, for example, when spreading the includes result into the params object:

```ts
type Params = {
  field?: string
}

function fn(obj: Params) {}

// Case 1
fn({
  field: '',
  unexpectedField: '', // <-- ✅ Error
})

// Case 2
fn({
  ...{
    field: '',
    unexpectedField: '', // <-- ❌ No error
  },
})

// Case 3
const b: Params = {
  field: '',
  unexpectedField: '', // <-- ✅ Error
}

// Case 4
const c = {
  field: '',
  unexpectedField: '', // <-- ❌ No error
}
fn(c)
```

johncowen: > Well, alright, let’s leave it as in the PR I guess.

Resolving this one.

johncowen: P.S. Let me know on the rest.

kleinfreund (approved):

PR #1530, opened 2023-09-21
chore(e2e): Adds a negative timeout defaulting to 4000

johncowen:
Mostly makes any "negative" assertions use `4000` for a timeout, for if we extend "positive" assertions to a `10000` timeout. This will mean "positive" will pass as soon as possible, but "negative" ones won't wait a full `10000` until they fail.

(plus a few tiny amends detailed in the inlines)

### DIFF
--- a/cypress/support/step_definitions/visit.ts
+++ b/cypress/support/step_definitions/visit.ts
@@ -1,17 +1,12 @@
 import { When } from '@badeball/cypress-cucumber-preprocessor'
 
-const config = {
-  BASE_URL: Cypress.env('KUMA_BASE_URL') || 'http://localhost:5681/gui',
-}
-
 When('I visit the {string} URL', function (path: string) {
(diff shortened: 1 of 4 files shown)


cypress/support/step_definitions/visit.ts
```diff
@@ -35,6 +30,5 @@ When('I visit the {string} URL', function (path: string) {
   cy.get('.app-main-content', { timeout: 10000 }).should('be.visible')
 })
 When('I load the {string} URL', function (path: string) {
-  cy.viewport(1366, 768)
-  cy.visit(`${config.BASE_URL}${path}`)
+  cy.visit(`${path}`)
```
johncowen, line 33: Don't think we need to do this anymore.

johncowen (commented):
Inlines:

kleinfreund (approved):

PR #1532, opened 2023-09-22
refactor: switches to kongponents v9 alpha

kleinfreund:
Kongponents v9 migration guide: https://alpha--kongponents.netlify.app/guide/migrating-to-version-9.html

## Changes

- **chore(deps): updates @kong/kongponents to v9.0.0-alpha.N**

- **refactor(codeblock): switches to @kong/icons**

  Replaces all the copy KIcon.

- **chore: addresses KBadge changes**

- **chore: addresses KButton changes**

  Appearance: renames outline to tertiary.

  Size: removes some instances of `size="small"` as we don’t need it anymore in most cases.

  Adds `icon-only` to icon-only buttons.

- **chore: addresses KCard changes**

- **chore: addresses KCodeBlock changes**

- **chore: addresses KDropdownMenu changes**

  Removes unnecessary `props.buttonAppearance`.

  Renames `props.label` to `props.triggerText`.

  Updates related test selector usage.

- **chore: addresses KInput changes**

- **chore: addresses KPop changes**

- **chore: addresses KRadio changes**

  Makes sure radio button groups in Zone create view are stacked vertically.

  Adjusts test selector to account for breaking change in KRadio markup (label is no longer a sibling of the input).

- **chore: addresses KSelect changes**

  Removes `props.overlayLabel`. The removal of the floating label breaks the layout of the gateway list view toolbar. To account for this, I’ve added a few styles to position the label before the select element instead of above it.

- **chore: addresses KSlideout changes**

  Uses `props.offsetTop` instead of overriding `top` using CSS.

- **chore: addresses KTabs changes**

  Adds a Kongponents override to NavTabs to make the component’s border bottom have the same color as KCard.

- **chore: makes non-visual-button use !important**

- **chore: changes upgrade check appearance to info**

## Notes

**Known/outstanding issues**:

- **Icons**: No Kong/icons replacement for the "stateNoData" KIcon
- **KSlideout**: has a too dark left border color (an override was added in this PR)
- **KTabs**: has a too dark bottom border color (an override was added in this PR)
- **KTabs**: inactive tabs have the same text color as active tabs

**Related**:

- https://github.com/kumahq/kuma-gui/pull/1527
- https://github.com/kumahq/kuma-gui/pull/1563

### DIFF
--- a/features/zones/Create.feature
+++ b/features/zones/Create.feature
@@ -86,7 +86,7 @@ Feature: zones / create
     Then the "$ingress-input-switch input" element isn't checked
     And the "$egress-input-switch input" element isn't checked
 
-    When I click the "$environment-universal-radio-button + label" element
+    When I click the "$environment-universal-radio-button" element
     Then the "$ingress-input-switch input" element doesn't exist
     And the "$egress-input-switch input" element doesn't exist
     And the "$environment-universal-config" element contains "globalAddress: grpcs://<global-kds-address>:5685"
--- a/features/zones/Delete.feature
+++ b/features/zones/Delete.feature
@@ -4,8 +4,8 @@ Feature: zones / delete
       | Alias          | Selector                                                       |
       | items          | [data-testid="zone-cp-collection"]                             |
       | item           | $items tbody tr                                                |
-      | actions-button | $item:nth-child(1) [data-testid="k-dropdown-trigger"] button   |
-      | delete-button  | $item:nth-child(1) [data-testid="dropdown-delete-item"] button |
+      | actions-button | $item:nth-child(1) [data-testid='dropdown-trigger'] button     |
+      | delete-button  | $item:nth-child(1) [data-testid='dropdown-delete-item'] button |
       | delete-prompt  | [data-testid="delete-zone-modal"]                              |
       | confirm-button | $delete-prompt .k-prompt-proceed                               |
       | confirm-input  | $delete-prompt .k-prompt-confirm-text .k-input                 |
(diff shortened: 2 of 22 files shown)


johncowen (approved):
This looks fine to me. I noticed something that doesn't need to be fixed here necessarily.

Screenshot 2023-11-20 at 15 04 05

The text here to me looks like a clickable link, but its not clickable.

One of those, if you want to fix it here and have time then go for it otherwise we can always fix it after this PR.

---

Other than that, would be good to get a kongponents person to cast their eye over this also.

Oh one last question, do you know if there will be a final bump to take us out of an alpha version?

kleinfreund:
> Oh one last question, do you know if there will be a final bump to take us out of an alpha version?

There _won’t_ be a bump out of alpha for Kongponents v9 _now_. The process of updating all components to the v9 appearance isn’t complete, yet, and for the remainder of this process, Kongponents’ alpha branch will be used. @adamdehaven, is that correct?

adamdehaven:
Correct, you will be on the alpha for a while.

We will continue updating components in "groups" and then we'll ask you to update your app(s) in followup PRs, e.g. `KInputSwitch, KSelect, KMultiselect, KCard` etc.

kleinfreund:
I updated this PR one more time to update the Kong-specific packages in https://github.com/kumahq/kuma-gui/pull/1532/commits/cbf167ac858a96382f8a2037b9d1e32086ec5482.

kleinfreund:
I’ve converted this PR to a draft to signal that it shouldn’t be merged today. We’ve postponed this switch to tomorrow.

PR #1533, opened 2023-09-22
chore: fixes vitepress build

kleinfreund:
**chore: fixes vitepress build**

Renames config.ts to config.mts to address `"vitepress" resolved to an ESM file.` error.

Adds `ignoreDeadLinks` config for the dead link checker matching on localhost references.

**chore: address vitepress config type issues**

### DIFF
--- a/.vitepress/config.mts
+++ b/.vitepress/config.mts
@@ -18,6 +18,10 @@ export default defineConfig({
   title: 'kuma-gui',
   description: '',
   cleanUrls: true,
+  ignoreDeadLinks: [
+    // ignore all localhost links
+    /^https?:\/\/localhost/,
+  ],
   themeConfig: {
     // https://vitepress.dev/reference/default-theme-config
     nav: [
--- a/.vitepress/config.mts
+++ b/.vitepress/config.mts
@@ -11,7 +11,11 @@ const files = globSync('./src/**/README.md').map(item => {
     text: name,
     link: item
   }
-}).filter(Boolean)
+}).filter(notEmpty)
+
+function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
+  return value !== null && value !== undefined
+}
 
 // https://vitepress.dev/reference/site-config
 export default defineConfig({
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,7 @@ yarn-error.log*
 
 # Yarn
 .yarn/*
+
+# Vitepress
 .vitepress/cache
+.vitepress/dist


.vitepress/config.mts
```diff
@@ -11,13 +11,21 @@ const files = globSync('./src/**/README.md').map(item => {
...
 
 // https://vitepress.dev/reference/site-config
 export default defineConfig({
   title: 'kuma-gui',
   description: '',
   cleanUrls: true,
+  ignoreDeadLinks: [
+    // ignore all localhost links
+    /^https?:\/\/localhost/,
+  ],
```
johncowen, line 28: Do you know what links are causing this not to build?

I don't think we would ever want to build this, and I was going to say we may as well just remove that script, but! the build script would be good to use to check that there aren't any broken links (unless you can get it to do the same check on `docs:dev`?).

.vitepress/config.mts
```diff
@@ -11,13 +11,21 @@ const files = globSync('./src/**/README.md').map(item => {
     text: name,
     link: item
   }
-}).filter(Boolean)
+}).filter(notEmpty)
+
+function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
+  return value !== null && value !== undefined
+}
```
johncowen, line 18: Any reason not to use

`.filter(<T,>(val?: T): val is T => Boolean(val))`

here?

It's closer to the JS and to me says "I'm only doing this cos Typescript can't figure this out on its own"

Lastly, could you make sure this file is picked up by `vue-tsc` if we are doing these sorts of things here? ty!

johncowen (commented):
Related: https://github.com/kumahq/kuma-gui/pull/1467 (that's when the .mts change was required)

If you are trying to remove warnings here could you have a quick look and see if you can add the config so it can consume our components? 

When I added this I was like "well I've got as far as installing vitepress in a place that can work for us and I can write basic docs, I can sort this later" but if you are looking at these things now it would be good if we can get those warning sorted also.

Don't spend too long on it but that would also be helpful here. lmk!

.vitepress/config.mts
kleinfreund: It’s the references in DEVELOPER.md and they’re perfectly fine there _but_ will of course fail any link check test.

.vitepress/config.mts
kleinfreund: > Any reason not to use
>
> `.filter(<T,>(val?: T): val is T => Boolean(val))`
>
> here?

I don’t see how that’s any better.

> Lastly, could you make sure this file is picked up by vue-tsc if we are doing these sorts of things here?

This should definitely not be picked up by our `lint:ts` script because that should only evaluate source code. We could use a separate vue-tsc script, but this feels like over-engineering. Not sure what the value of that would be.

johncowen:
Don't worry I was only asking if you could take a quick look while you were here

PR #1555, opened 2023-09-26
chore(deps): pins all dependency versions

kleinfreund:
**chore(deps): pins all dependency versions**

Pins the versions of all dependencies (i.e. removes all `^` and `~` characters from version ranges).

**chore(deps): regenerates yarn.lock**


johncowen:
Looks good to me, looks like we could also add something like this?

https://github.com/kellyselden/eslint-plugin-json-files/blob/master/docs/rules/restrict-ranges.md

wdyt?

kleinfreund:
@johncowen I added a `.yarnrc` config that makes sure `yarn add` only adds exact versions in https://github.com/kumahq/kuma-gui/pull/1555/commits/7302fb922c10bd805334bff1bec1116d8cb5cfd3. I don’t think we need to lint for this.

johncowen:
> @johncowen I added a .yarnrc config that makes sure yarn add only adds exact versions

Ah perfect yes, much better, 2 tics

johncowen (approved):

PR #1558, opened 2023-09-26
chore(e2e): general e2e test improvements

johncowen:
This began as a clean up of e2e test code following improvements in:

- https://github.com/kumahq/kuma-gui/pull/1482
- https://github.com/kumahq/kuma-gui/pull/1417
- https://github.com/kumahq/kuma-gui/pull/1452

I also noticed some areas mainly around subscription counts that felt a little strange, and began to tweak those a little, and then as I delved deeper I started to go down a bit of a rabbit hole:

- I started to make all mock instances of subscriptions used the `KUMA_SUBSCRIPTION_COUNT` variable.
- Started making all mock instances of subscriptions as consistent as possible, and decided to make a `fake.kuma.connection` for providing `connectTime`/`disconnectTime`, which I _think_ is more realistic, would be good for a second pair of eyes on that. The intention is that only the last subscription possibly won't have a disconnectTime and it will always be after the connectTime. I just padded out the difference in golang date formats with a random number.
- The above led me to looking at our types around subscriptions and then zone overview type things, and this is where I felt like I was going too far into a rabbit hole for this PR. For example, one of the type corrections I made (`zoneEgress` here https://github.com/kumahq/kuma-gui/pull/1558/commits/14dd7105ecc328ae0c0a9635074f5e8c845854c2 seems to be possibly `undefined`) started breaking things deeper in the app, so if this type is incorrect there are more potential undefined access bugs to be had. Instead of fixing things here I undid that change (https://github.com/kumahq/kuma-gui/pull/1558/commits/db2cb1f7a3def1359ab2bf15f5697fb9c78f9abc) for a future piece of work.

### DIFF
--- a/src/test-support/FakeKuma.ts
+++ b/src/test-support/FakeKuma.ts
@@ -61,6 +61,25 @@ export class KumaModule {
     return `${this.faker.hacker.noun()}-${deploymentId}-${podId}.${namespace}`
   }
 
+  connection<T>(_: T, i: number, arr: T[]) {
+    const connected = this.faker.date.past()
+    const times: {
+      connectTime: string,
+      disconnectTime?: string
+    } = {
+      connectTime: `${connected.toISOString().slice(0, -1)}${this.faker.number.int({ min: 1000, max: 9999 })}Z`,
+      // connectTime: '2021-07-13T08:41:04.556796688Z',
+    }
+    if ((arr.length > 1 && i !== arr.length - 1) || this.faker.datatype.boolean({ probability: 0.3 })) {
+      times.disconnectTime = `${this.faker.date.between({
+        from: connected,
+        to: new Date(),
+      }).toISOString().slice(0, -1)}${this.faker.number.int({ min: 1000, max: 9999 })}Z`
+      // times.disconnectTime = '2021-02-17T07:33:36.412683Z'
+    }
+    return times
+  }
+
   status() {
     return this.faker.helpers.arrayElement(
       [
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -545,7 +545,7 @@ export interface ZoneEgress extends MeshEntity {
 
 export interface ZoneEgressOverview extends MeshEntity {
   type: 'ZoneEgressOverview'
-  zoneEgress?: {
+  zoneEgress: {
     zone?: string
     networking?: ZoneEgressNetworking
   }
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -549,9 +549,7 @@ export interface ZoneEgressOverview extends MeshEntity {
     zone?: string
     networking?: ZoneEgressNetworking
   }
-  zoneEgressInsight?: {
-    subscriptions: DiscoverySubscription[]
-  }
+  zoneEgressInsight: any
 }
 
 export interface Backend {
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -545,11 +545,13 @@ export interface ZoneEgress extends MeshEntity {
 
 export interface ZoneEgressOverview extends MeshEntity {
   type: 'ZoneEgressOverview'
-  zoneEgress: {
+  zoneEgress?: {
     zone?: string
     networking?: ZoneEgressNetworking
   }
-  zoneEgressInsight: any
+  zoneEgressInsight?: {
+    subscriptions: DiscoverySubscription[]
+  }
 }
 
 export interface Backend {
--- a/features/application/Loading.feature
+++ b/features/application/Loading.feature
@@ -11,9 +11,6 @@ Feature: application / loading
       """
       KUMA_LATENCY: 1000
       """
-    And the URL "/meshes" responds with
-      """
-      """
     When I load the "/" URL
     Then the "$loading" element exists
     Then the "$loading" element doesn't exist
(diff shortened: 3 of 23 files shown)


kleinfreund:
**Comment**: I’m almost certain that the change in https://github.com/kumahq/kuma-gui/commit/14dd7105ecc328ae0c0a9635074f5e8c845854c2 would be incorrect anyway. I don’t think `zoneEgress` or `zoneEgressInsight` can actually be absent. Would be good to verify that, but this seems to be a pretty stable pattern: it seems that an object that has properties that themselves refer to objects are never omitted. Another example of this I can think of is `dataplanesByType` on `MeshInsight`. It can never be omitted and neither can its properties.

I don’t however know how general this rule is. It possibly has something to do with our use of protobuf? Don’t know.

src/test-support/FakeKuma.ts
```diff
@@ -61,6 +61,25 @@ export class KumaModule {
...
   }
 
+  connection<T>(_: T, i: number, arr: T[]) {
+    const connected = this.faker.date.past()
+    const times: {
+      connectTime: string,
+      disconnectTime?: string
+    } = {
+      connectTime: `${connected.toISOString().slice(0, -1)}${this.faker.number.int({ min: 1000, max: 9999 })}Z`,
+      // connectTime: '2021-07-13T08:41:04.556796688Z',
```
kleinfreund, line 71: **Suggestion**: Development remnant
```suggestion
```

src/test-support/FakeKuma.ts
```diff
@@ -61,6 +61,25 @@ export class KumaModule {
...
+    }
+    if ((arr.length > 1 && i !== arr.length - 1) || this.faker.datatype.boolean({ probability: 0.3 })) {
+      times.disconnectTime = `${this.faker.date.between({
+        from: connected,
+        to: new Date(),
+      }).toISOString().slice(0, -1)}${this.faker.number.int({ min: 1000, max: 9999 })}Z`
+      // times.disconnectTime = '2021-02-17T07:33:36.412683Z'
```
kleinfreund, line 78: **Suggestion**: Development remnant
```suggestion
```

src/types/index.d.ts
```diff
@@ -562,7 +562,9 @@ export interface ZoneEgressOverview extends MeshEntity {
     zone?: string
     networking?: ZoneEgressNetworking
   }
-  zoneEgressInsight: any
+  zoneEgressInsight?: {
+    subscriptions: DiscoverySubscription[]
+  }
```
kleinfreund, line 567: **Issue**: I don’t thin k`zoneEgressInsight` can be omitted, but `subscriptions` can.

```suggestion
  zoneEgressInsight: {
    subscriptions?: DiscoverySubscription[]
  }
```

kleinfreund (changes requested):

src/test-support/FakeKuma.ts
johncowen: I left these in on purpose so you could see the format and understand what all the fake number padding was about.

src/test-support/FakeKuma.ts
johncowen: See above

src/types/index.d.ts
johncowen: Well I'm not totally sure these are the generated types:

Screenshot 2023-09-28 at 10 08 38
Screenshot 2023-09-28 at 10 10 19

?

johncowen: but then again this is DataplaneInsights, which we know, I guess without the extra protections that the generated types give, can have `subscriptions` as `undefined`

Screenshot 2023-09-28 at 10 14 52

This is were I felt I was going down a large rabbit hole, so I'm more inlined to remove this change entirely. Hopefully we'll be sorting this whole issue out once and for all soonish.

Give me a bit let me take a better look at this from here. brb

kleinfreund (approved):

src/types/index.d.ts
johncowen: Well the point is, we shouldn't have to keep that in mind as these generated types make sure that subscriptions is always an array, which is what we want.

Screenshot 2023-09-28 at 13 18 41

>  It definitely looks there is a knowledge gap somewhere we because we do indeed know subscriptions can be undefined.

Well, I'm not sure if its a "knowledge gap" thing. With those generated types `subscriptions` is always an array, which then prevents all the undefined access errors we have without having to keep that in mind or having to have the knowledge about it. I think that's part of the point.

1 sec I think I'm going to take these type changes out as I'm guessing they could have a bigger impact that I don't want to have to look at in this PR and this was part of the rabbit hole I got the feeling I was getting into here.

johncowen: OK I removed the type changes here, I have an approval already so I'll wait a bit before merging just incase.

johncowen:
> Comment: I’m almost certain that the change in https://github.com/kumahq/kuma-gui/commit/14dd7105ecc328ae0c0a9635074f5e8c845854c2 would be incorrect anyway. I don’t think zoneEgress or zoneEgressInsight can actually be absent. Would be good to verify that, but this seems to be a pretty stable pattern: it seems that an object that has properties that themselves refer to objects are never omitted. Another example of this I can think of is dataplanesByType on MeshInsight. It can never be omitted and neither can its properties.

> I don’t however know how general this rule is. It possibly has something to do with our use of protobuf? Don’t know.

I think I missed this comment yesterday and I've just seen it now, just for note cos I've already mentioned it in the other inline, but I removed the type changes here anyway - it was those that made me feel like I was getting started into the rabbit hole

PR #1569, opened 2023-09-29
chore: Reorganizes `production.ts` wiring root into micro-services/frontends

johncowen:
TBD


jest/jest-setup-after-env.ts
```diff
@@ -34,11 +32,8 @@ const $ = {
     development($),
     testing($),
   )
-
-  /**
-  * Adds the application’s router to vue test utils. This way tests don’t have to set-up a new router instance on their own.
-  */
-  config.global.plugins.push(get($.router))
+  // initializes vue-test-utils with any global components and/or plugins etc
+  get($.app)
```
johncowen, line 36: This now shares the same `$.components` and `$.plugins` tokens as the main application, i.e. whatever you install in the application, also gets installed for the tests. You don't have to add the things in two places anymore.

src/app/AppNotFoundView.vue
```diff
@@ -1,30 +1,37 @@
...
+              appearance="primary"
+              :to="{ name: 'home' }"
+            >
+              Go Home
+            </KButton>
+          </template>
+        </KEmptyState>
+      </div>
+    </AppView>
+  </RouteView>
```
johncowen, line 33: I noticed out Page Not Found view was missing RouteView/Title and AppView, kinda not related to this PR, but as I was there I added it here.

src/app/application/index.ts
```diff
@@ -48,6 +67,19 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
...
       ],
     }],
+
+    [token('application.routes'), {
+      service: routes,
+      arguments: [
+        $.notFoundView,
+      ],
+      labels: [
+        app.routes,
```
johncowen, line 77: For anything that is labelled, there's currently not a lot of point typing the token (ideally it would be typed using the type of the label anyway).

Plus we want to make sure these tokens can't be exported, i.e. they are kinda like "private tokens". If you need to overwrite one of these you'd need to re-expose it like it was before.

In the future, labels will themselves be inspectable, so you'll be able to get a specific token/service by inspecting the public label token.

src/app/application/routes.ts
```diff
@@ -0,0 +1,16 @@
...
+export const routes = (notFoundViews: (() => Promise<Component>)[]): RouteRecordRaw[] => {
+  return [
+    {
+      path: '/404',
+      name: 'not-found',
+      alias: '/:pathMatch(.*)*',
+      meta: {
+        title: 'Item not found',
+      },
+      component: () => notFoundViews[notFoundViews.length - 1](),
```
johncowen, line 13: I wanted to point this one out. Its not super important as its a non important 404 page, but I wanted to using labelling to specify the component for this from The Outside, as labels are mostly arrays, this is why we do the weird "get the last one in the array" thing. If we want to overwrite the 404 page at anypoint this "ge tthe last one of the labelled things" should"Just Work(tm)" for us.

src/app/data-planes/index.ts
```diff
@@ -4,14 +4,10 @@ import { token } from '@/services/utils'
...
 type Token = ReturnType<typeof token>
-type Sources = ReturnType<typeof sources>
 
-const $ = {
-  sources: token<Sources>('mesh.sources'),
-}
 export const services = (app: Record<string, Token>): ServiceDefinition[] => {
   return [
-    [$.sources, {
+    [token('data-planes.sources'), {
```
johncowen, line 10: Again, all of these (and further down) are because:

1. I don't want to export these tokens
2. Its almost pointless typing them as they should use the type of the label (they don't, yet)

src/app/meshes/index.ts
```diff
@@ -20,6 +52,9 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
         app.sources,
       ],
     }],
+    ...servicesModule(app),
+    ...dataplanes(app),
+    ...gateways(app),
+    ...policies(app),
```
johncowen, line 58: The mesh "micro-frontend"/plugin is the weirdest one in that it has sub dependencies inside it. I wanted to do this so you can just add "mesh" on its own and you get all the sub dependencies automatically.

src/app/vue/index.ts
```diff
@@ -0,0 +1,123 @@
...
+    }],
+  ]
+}
+
+export const TOKENS = $
+export const [
+  useRouter,
+] = createInjections(
+  TOKENS.router,
+)
```
johncowen, line 123: A new 'vue' "plugin" for the essential vue-type stuff.

src/services/production.ts
```diff
@@ -149,73 +104,28 @@ export const services: ServiceConfigurator<SupportedTokens> = ($) => [
...
+    routes: $.routesLabel,
+  }),
+  ...zones({
+    ...$,
+    routes: $.routesLabel,
+  }),
+  ...meshes({
+    ...$,
+    routes: $.routesLabel,
+  }),
```
johncowen, line 127: I think this could end up being a `serviceMesh` plugin/module/micro-frontend.

That would mean that in each app, you would only add the single `serviceMesh` plugin to get all of this, then you can pick and choose what extras you want ontop of that.

src/services/testing.ts
```diff
@@ -1,23 +1,49 @@
...
+      return async () => {
+        throw new Error('You shouldn\'t be calling $.app during testing, just get(\'$.app\') is fine')
+      }
+    },
+    arguments: [
+      app.components,
+      app.plugins,
+    ],
+  }],
+
```
johncowen, line 46: CLI testing now uses the exact same approach as the application for adding global components and or plugins (or directives or whatever else we might need to add at a global level).

johncowen (commented):
Inlines:

kleinfreund (approved):

Issue #1480, opened 2023-09-13
For Zones that use in memory, show warning that this is not ideal in list and detail view


kleinfreund:
### Description

See title

jakubdyszkiewicz:
https://kuma.io/docs/2.4.x/documentation/configuration/#memory

PR #1571, opened 2023-10-02
chore: remove WarningsWidget and add new warnings

johncowen:
This PR:

- Adds a 'Certificate Expiry' column to `DataPlaneList.vue` but does nothing else but add the information (there will be a further PR incoming to do more with that)
- Refactors how we do warnings:
  - Removes `WarningsWidget.vue` and its sub components.
  - Properly enables markdown in our YAML/i18n files (we need to use v-html to prevent re-encoding the HTML, considering we assume the YAML is safe)
  - Adds a `notification` slot to `AppView` to situate the notifications (currently notifications are only of the warning 'type')
- Adds a warning in ZoneCP Listings and ZoneCPDetails if you are using the memory store type.
- Adds e2e tests for the warnings (both new and previous ones)

Notes:

- I've left in the old `compatibility` things plus the `WARNING_CONSTANTS` that we used to use. When we have a data layer this will probably be done in the data layer instead of the view, and I guess we'll fiddle more with that then.
- Warnings now use one single 'box' with a `ol` of the multiple warnings in it. Previously it would stack/use separate 'boxes'.
- I also left a load of inline PR comments to help review.

Closes https://github.com/kumahq/kuma-gui/issues/1480


lahabana:
https://github.com/kumahq/kuma/issues/2689

lahabana:
In mocks how can I force MTLS?

features/zones/ListViewContent.feature
```diff
@@ -2,79 +2,9 @@ Feature: Zones: List view content
...
-    Then the "$zone-cp-table-row:nth-child(1) .type-column" element contains "universal"
-
-    Then the "$zone-cp-table-row:nth-child(2) .status-column" element contains "offline"
-    Then the "$zone-cp-table-row:nth-child(2) .name-column" element contains "zone-cp-2"
-    Then the "$zone-cp-table-row:nth-child(2) .zoneCpVersion-column" element contains "1.0.0-rc2-211-g823fe8ce"
-    Then the "$zone-cp-table-row:nth-child(2) .type-column" element contains "kubernetes"
-
-    Then the "$zone-cp-table-row:nth-child(3) .status-column" element contains "disabled"
-    Then the "$zone-cp-table-row:nth-child(3) .name-column" element contains "zone-cp-3"
-
```
johncowen, line 77: This red and the below green is just file moving from `zones/ListViewContent` to `zones/zone-cps/Index`

features/zones/zone-cps/Item.feature
```diff
@@ -44,15 +43,3 @@ Feature: zones / zone-cps / item
...
-    And the URL "/zones+insights/zone-cp-1" responds with
-      """
-      body:
-        name: zone-cp-1
-        zoneInsight:
-          subscriptions: ~
-      """
-    When I visit the "/zones/zone-cps/zone-cp-1" URL
-    And I click the "#zone-cp-config-view-tab a" element
-    Then the "$warning-no-subscriptions" element exists
```
johncowen, line 58: This was moved into the Warnings.feature file below (and added to)

jest/yaml-transformer.js
```diff
@@ -1,10 +1,30 @@
...
+            // things, plus ICU '{' and '}', hence this replace.
+            // If we ever need !!text/markdown for anything else we should do something like !!text/icu+markdown
+            return marked(data).replace(/</g, "'<'")
+              .replace(/%7B/g, '{')
+              .replace(/%7D/g, '}')
+          },
+        }),
+      ),
+    })
+    const json = JSON.stringify(result, undefined, '\t')
```
johncowen, line 25: makes jest support our in-yaml markdown

src/app/application/services/i18n/I18n.ts
```diff
@@ -44,7 +49,10 @@ export default <T extends I18nRecord>(strs: T, env: Env['var']) => {
         if (get(strs, key).length === 0) {
           return ''
         }
-        rest[1] = { KUMA_DOCS_URL: env('KUMA_DOCS_URL'), KUMA_UTM_QUERY_PARAMS: env('KUMA_UTM_QUERY_PARAMS'), ...rest[1] }
+        rest[1] = {
+          ...globals,
+          ...rest[1],
+        }
```
johncowen, line 55: These 'globals; never change so I just hoisted them to only retrieve the once rather than everytime we call `t`

src/app/data-planes/data/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  const rawExpDate = new Date(dataPlaneOverview.dataplaneInsight.mTLS.certificateExpirationTime)
+  // this prevents any weird date shifting
+  const fixedExpDate = new Date(rawExpDate.getTime() + rawExpDate.getTimezoneOffset() * 60000)
+
+  return {
+    ...dataPlaneOverview.dataplaneInsight.mTLS,
+    certificateExpirationTime: fixedExpDate.toISOString(),
+  }
+}
```
johncowen, line 18: This was moved out of `utilities/dataplane.ts`, but I also removed the formatting that it used to have here and moved that formatting to the template instead.

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -1,21 +1,313 @@
...
+  gap: $kui-space-50;
+}
+</style>
+
+<style lang="scss">
+.reason-tooltip .kong-icon {
+  display: flex;
+  align-items: center;
+}
+</style>
```
johncowen, line 313: All of this green is because I removed the `DataPlaneDetail.vue` component entirely and copied its source directly into `DataPlaneDetailView.vue`

src/app/zones/data/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  const subscriptions = zoneOverview.zoneInsight?.subscriptions ?? []
+  if (subscriptions.length === 0) {
+    return 'offline'
+  }
+
+  const lastSubscription = subscriptions[subscriptions.length - 1]
+  return lastSubscription.connectTime?.length && !lastSubscription.disconnectTime ? 'online' : 'offline'
+}
```
johncowen, line 28: This has just been moved from elsewhere into the `zones/data` folder

src/app/zones/views/ZoneDetailView.vue
```diff
@@ -1,21 +1,135 @@
...
+        payload: {
+          zoneCpVersion: kumaCpVersion,
+        },
+      })
+    }
+  }
+
+  return warnings
+})
+
```
johncowen, line 134: This green is just because I removed the `ZoneDetail.vue` and copy/pasta'ed its source directly into `ZoneDetailView.vue`

src/locales/en-us/common/index.yaml
```diff
@@ -10,6 +10,19 @@ common:
...
+    STORE_TYPE_MEMORY: !!text/markdown |
+      When using the `memory` store type the state is not persisted. This store should only be used when trying things out. <a target="_blank" href="{KUMA_DOCS_URL}/documentation/configuration/#store">Read more about store types</a>
+    INCOMPATIBLE_UNSUPPORTED_ENVOY: !!text/markdown |
+      Envoy (**{ envoy }**) is unsupported by the current version of Kuma DP (**{ kumaDp }**)
+    INCOMPATIBLE_UNSUPPORTED_KUMA_DP: !!text/markdown |
+      Unsupported version of Kuma DP (**{ kumaDp }**)
+    INCOMPATIBLE_ZONE_CP_AND_KUMA_DP_VERSIONS: !!text/markdown |
+      There is a mismatch between versions of Kuma DP (**{ kumaDp }**) and the Zone Control Plane.
+    INCOMPATIBLE_ZONE_AND_GLOBAL_CPS_VERSIONS: !!text/markdown |
+      There is mismatch between versions of Zone Control Plane (**{ zoneCpVersion }**) and the Global Control Plane (**{ KUMA_VERSION }**)
```
johncowen, line 25: The uppercase here is a little awkward, but its just a styling thing, and I've already changed quite a bit. I might alter that at some point in the future.

johncowen (commented):
Inlines:

lahabana:
image

Which zone is it? Is it for the global? Can we have a similar warning for each zones like the version mismatch?

features/mesh/dataplanes/Index.feature
```diff
@@ -39,16 +39,17 @@ Feature: mesh / dataplanes / index
...
-      | Last Updated |
-      | Status       |
-      | Warnings     |
+      | Value              |
+      | Name               |
+      | Service            |
+      | Protocol           |
+      | Zone               |
+      | Last Updated       |
+      | Certificate Expiry |
```
lahabana, line 50: I'd call this `Certificate Info` as you are going to add more info here.

lahabana: Also when there's no cert maybe it's worth filling the column with `No certificate` or something like this.

johncowen:
> Which zone is it? Is it for the global? Can we have a similar warning for each zones like the version mismatch?

Currently its from the global config, but I guess I can store type for the zone from the `zoneInsight.subscription[].config` can I? Ah yes, cool, let me amend that. I can put this in the zone detail page instead.

johncowen:
@lahabana I did everything we spoke about here _apart from_ the new warnings that would happen on a global `store.type="memory"`.

I'll probably do this outstanding work in a separate PR as it now depends on https://github.com/kumahq/kuma/pull/7965 and it will need a tiny bit more rewiring in the GUI also.

Lmk!

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -1,21 +1,290 @@
...
+    if (compatibility.kind !== COMPATIBLE && compatibility.kind !== INCOMPATIBLE_WRONG_FORMAT) {
+      warnings.push(compatibility)
+    }
+  }
+  if (
+    mtlsData.value &&
+    (Date.now() > new Date(mtlsData.value?.certificateExpirationTime).getTime())
+  ) {
+    warnings.push({
+      kind: 'CA_EXPIRED',
```
lahabana, line 254: It's not the CA (certificate authority) that is expired it's the certificate

src/locales/en-us/common/index.yaml
```diff
@@ -10,6 +12,21 @@ common:
     mode:
       standalone: Standalone
       global: Multi-Zone
+  warnings:
+    CA_EXPIRED: !!text/markdown |
+      The CA for this dataplane has expired
```
lahabana, line 17:
```suggestion
      The certificate for this dataplane has expired
```

features/mesh/dataplanes/Warnings.feature
```diff
@@ -0,0 +1,105 @@
+Feature: mesh / dataplanes / warnings
+  Background:
+    Given the CSS selectors
+      | Alias                     | Selector                                                          |
+      | expired-ca-warning        | [data-testid='warning-CA_EXPIRED']                                |
+      | unsupported-kuma-warning  | [data-testid='warning-INCOMPATIBLE_UNSUPPORTED_KUMA_DP']          |
+      | unsupported-envoy-warning | [data-testid='warning-INCOMPATIBLE_UNSUPPORTED_ENVOY']            |
+      | unsupported-zone-warning  | [data-testid='warning-INCOMPATIBLE_ZONE_CP_AND_KUMA_DP_VERSIONS'] |
+
+  Scenario: With an expired CA a CA warning is shown
```
lahabana, line 10: Cert :)

src/app/data-planes/data/index.ts
```diff
@@ -0,0 +1,18 @@
...
+} from '@/types/index.d'
+
+export function parseMTLSData(dataPlaneOverview: DataPlaneOverview) {
+  if (dataPlaneOverview.dataplaneInsight === undefined || dataPlaneOverview.dataplaneInsight.mTLS === undefined) {
+    return null
+  }
+
+  const rawExpDate = new Date(dataPlaneOverview.dataplaneInsight.mTLS.certificateExpirationTime)
+  // this prevents any weird date shifting
+  const fixedExpDate = new Date(rawExpDate.getTime() + rawExpDate.getTimezoneOffset() * 60000)
```
lahabana, line 12: Times should always be returned in UTC so I'm not sure that's useful

src/app/data-planes/data/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  return {
+    ...dataPlaneOverview.dataplaneInsight.mTLS,
+    certificateExpirationTime: fixedExpDate.toISOString(),
```
lahabana, line 16: I thought there was a t funciton or something like this for date display no?

src/app/zones/data/index.ts
```diff
@@ -0,0 +1,28 @@
...
+export function getZoneDpServerAuthType(zone: ZoneOverview): string {
+  const subscriptions = zone.zoneInsight?.subscriptions ?? []
+  if (subscriptions.length > 0) {
+    const lastSubscription = subscriptions[subscriptions.length - 1]
+    if (lastSubscription.config) {
+      const parsedConfig = JSON.parse(lastSubscription.config)
+      return get(parsedConfig, 'dpServer.auth.type', 'disabled')
+    }
+  }
+  return 'disabled'
```
lahabana, line 13: It's more `N/A` than disabled right?

lahabana (approved):
Just a confusion between C/A and Cert and other nits but otherwise LGTM

src/app/data-planes/data/index.ts
johncowen: Yeah that used in the template now as its a view/format thing, ideally we would be using `Date`s here instead of string also 🤔 let me look at that actually, I might move the `toISOString` to the view so we are dealing with correct data types in the data layer.

src/app/zones/data/index.ts
johncowen: umm dunno, let me look this is another one that comes from previous code and I just moved it around a bit.

src/app/data-planes/data/index.ts
johncowen: Just took a quick look at this, and I had second thoughts, the rest of the dates coming from the API are all in a string format, so will leave them all the same.

src/app/zones/data/index.ts
johncowen: I changed this to read `N/A` for the authenticationType if its empty/non-existent

johncowen:
Wow so in addressing the review feedback here, I noticed how we could simplify/improve this even more (the last 4 commits). Including I spotted more things that organisationally did not make sense to me, which led me to move the remaining data getter thing (`type`) in ZoneDetailView to the `data/index.ts` file also so its all consistent.

@lahabana I wanted to ask just incase:

Previously if we could not find the `environment` for the Zone by looking at its config, we would just default to show/assume `kubernetes`.

- I've changed this here to show `-` instead, in the I guess unlikely case that we don't know the environment.
- I also capitalised `Universal` / `Kubernetes` when we do know.

If you could let me know if those ^ 2 things are ok, that'd be great, if not I'll re-tweak.

lahabana:
sounds good

PR #1586, opened 2023-10-06
chore(zones): Nest ZoneIngresses and ZoneEgresses into their Zone

johncowen:
This PR nests ZoneIngresses and ZoneEgresses inside of their Zones, so you see each listing of the 'gresses inside of it parent ZoneCP.

---

The majority of this PR is just moving things around, so there is lots of green and red with no real change. The `@/app/zones` module was becoming increasingly hard to stare at (i.e. the `views/` folder difficult to figure out where I wanted to be at a glance).

Here, I've created 3 distinct modules instead of the 1 zones module:

1. `zones` (should really be `zone-cps`, but no biggie for now)
2. `zone-ingresses`
3. `zone-egresses`

I've purposefully in this PR not used our usual micro-frontend installing that we do with all our other modules, partly as I have https://github.com/kumahq/kuma-gui/pull/1569 up, and partly because these 3 modules mostly fit together, but there is an exception to this "fitting together".

When zones are enabled, they all "fit together" and it kinda makes sense to just "install" the `zones` micro-frontend. But, when zones aren't enabled, we only want `zone-egresses` which should live/be installed completely on its own. Therefore eventually these modules will be "installed" similarly to our other "micro-frontends", probably after https://github.com/kumahq/kuma-gui/pull/1569 has gone in.

Other Notes:

1. I deleted the `MultizoneInfo` component, this was just a notification thing if you visited say a zone-cps URL when you had zones disabled. With this PR these routes don't even exist anymore so you just get our generic 404 if you manually go to these routes when zone aren't enabled.
2. Zone'gress filtering by the parent zone is done on the frontend, I would imagine this will be replaced by a change to the backend API to allow us to filter by zone. Therefore I've avoided changing too much to do this, and left something commented in for when we have this so we can just undo the comments to put things back.

I'll try and add some inlines showing the more meaningful changes.

Oh lastly, I noticed there was a lot of repetition in the application code between Ingress and Egresses, they are almost the same...but not quite. But seeing as I'm in 'feature' mode rather than 'refactor/tech debt' mode, I tried not to fiddle with this too much as yet.


features/application/Titles.feature
```diff
@@ -18,13 +22,13 @@ Feature: application / titles
...
-      | /zones/zone-ingresses/zone-ingress-name | zone-ingress-name     |
-      | /zones/zone-egresses                    | Egresses              |
-      | /zones/zone-egresses/zone-egress-name   | zone-egress-name      |
+      | /zones/create                                            | Create & connect Zone |
+      | /zones/zone-cps                                          | Zone Control Planes   |
+      | /zones/zone-cps/zone-cp-name                             | zone-cp-name          |
+      | /zones/zone-cps/zone-cp-name/ingresses                   | Ingresses             |
+      | /zones/zone-cps/zone-cp-name/ingresses/zone-ingress-name | zone-ingress-name     |
+      | /zones/zone-cps/zone-cp-name/egresses                    | Egresses              |
+      | /zones/zone-cps/zone-cp-name/egresses/zone-egress-name   | zone-egress-name      |
```
johncowen, line 31: This is what the new URL structure looks like.

features/application/Titles.feature
```diff
@@ -40,3 +44,17 @@ Feature: application / titles
...
+      """
+      KUMA_MODE: standalone
+      """
+    When I visit the "<URL>" URL
+    Then the page title contains "<Title>"
+
+    Examples:
+      | URL                              | Title            |
+      | /zones/egresses                  | Egresses         |
+      | /zones/egresses/zone-egress-name | zone-egress-name |
```
johncowen, line 59: This is what the URLs look like with zones disabled.

src/app/zone-egresses/sources.ts
```diff
@@ -0,0 +1,78 @@
...
+      const offset = size * (page - 1)
+
+      const res = await api.getAllZoneEgressOverviews({ size, offset })
+      if (name !== '*') {
+        res.items = res.items.filter((item) => {
+          return item.zoneEgress.zone === name
+        })
+        res.total = res.items.length
+      }
+      return res
```
johncowen, line 42: This extra bit does the frontend filtering, we just ask for 100 ZoneEgresses and then only use the ones with the correct parent zone name. The `*` here means "gimme all of them", in the case of zones being disabled.

src/app/zone-egresses/views/IndexView.vue
```diff
@@ -8,23 +8,24 @@
...
               :render="true"
             />
-          </h1>
+          </h2>
         </template>
 
         <DataSource
           v-slot="{ data, error }: ZoneEgressOverviewCollectionSource"
-          :src="`/zone-egress-overviews?page=${route.params.page}&size=${route.params.size}`"
+          :src="`/zone-cps/${route.params.zone || '*'}/egresses?page=${'1'}&size=${'100'}`"
```
johncowen, line 28: route.params.zone will be empty when zones are disabled (`zones/egresses` vs `zones/zone-cps/name-of-parent/egresses)` so this is where we use `*`/"gimme all of them"

src/app/zone-egresses/views/IndexView.vue
```diff
@@ -8,23 +8,24 @@
       v-slot="{ route, t }"
       name="zone-egress-list-view"
       :params="{
-        page: 1,
-        size: me.pageSize
+        /* page: 1, */
+        /* size: me.pageSize, */
```
johncowen, line 12: Once we have backend filtering we'll put all this back, for now we just hardcode `page=1` and `size=100` in the URI for the DataSource.

src/app/zone-egresses/views/item/DetailView.vue
```diff
@@ -43,15 +43,24 @@
...
-        </div>
+            <KCard class="mt-4">
+              <template #body>
+                <SubscriptionList
+                  :subscriptions="subscriptions"
+                />
+              </template>
+            </KCard>
+          </div>
+        </template>
```
johncowen, line 63: I simplified this for Dataplane subscriptions in a previous PR, so I did this here also for zone'gresses

src/app/zone-egresses/views/item/IndexView.vue
```diff
@@ -1,13 +1,19 @@
...
     }"
   >
     <AppView
       :breadcrumbs="[
+        ...(can('use zones') ? [{
+          to: {
+            name: 'zone-cp-list-view',
+          },
+          text: t('zone-cps.routes.item.breadcrumbs')
+        }] : []),
```
johncowen, line 16: Only show the `Zone Control Planes >` breadcrumb in the Egress listing if zones are enabled.

src/app/zone-ingresses/sources.ts
```diff
@@ -0,0 +1,78 @@
...
+      source.close()
+
+      const { name, size, page } = params
+      const offset = size * (page - 1)
+
+      const res = await api.getAllZoneIngressOverviews({ size, offset })
+      res.items = res.items.filter((item) => {
+        return item.zoneIngress.zone === name
+      })
+      res.total = res.items.length
```
johncowen, line 41: Frontend filtering again, but no `*` this time as we never need "gimme all of them" for ingresses.

src/app/zone-ingresses/views/item/DetailView.vue
```diff
@@ -59,15 +59,24 @@
...
-        </div>
+            <KCard class="mt-4">
+              <template #body>
+                <SubscriptionList
+                  :subscriptions="subscriptions"
+                />
+              </template>
+            </KCard>
+          </div>
+        </template>
```
johncowen, line 79: Same simplification for ingress subscriptions.

src/app/zones/routes.ts
```diff
@@ -7,165 +10,72 @@ export const actions = (): RouteRecordRaw[] => {
...
+                ...egresses().item(),
               ],
             },
           ],
         },
-      ],
-    },
+
+      ]
+      : []),
```
johncowen, line 79: The entire green above here in this file is a little weird as this is were we use `can('use zones')` to give us different configurations of routing depending on whether zones are enabled or not. Probably later at some point, we'll probably do this higher up in our wiring root. I just did it here for now.

src/app/zones/sources.ts
```diff
@@ -13,30 +14,19 @@ type DetailParams = {
...
-export type ZoneEgressOverviewCollection = CollectionResponse<ZoneEgressOverview>
-export type ZoneEgressOverviewSource = DataSourceResponse<ZoneEgressOverview>
-export type ZoneEgressOverviewCollectionSource = DataSourceResponse<ZoneEgressOverviewCollection>
-
 export type EnvoyDataSource = DataSourceResponse<object | string>
 
 export const sources = (api: KumaApi) => {
   return {
+    ...zoneIngresses(api),
+    ...zoneEgresses(api),
```
johncowen, line 28: We'll eventually do this by installing these as proper "micro-frontends" at a later date. These "destructures" just contain the red below on the left here, I just moved them into sub micro-frontends/modules.

src/test-support/fake.ts
```diff
@@ -49,6 +49,7 @@ export type MockEnvKeys = keyof {
   KUMA_ZONEEGRESS_COUNT: string
   KUMA_ZONEINGRESS_COUNT: string
   KUMA_ZONE_COUNT: string
+  KUMA_ZONE_NAME: string
```
johncowen, line 52: Ah, awkwardly at the moment, if you want to see items in the list when preview you have to set a dev cookie with the name of the zone you are looking at, e.g. if you are looking at:

```
/gui/zones/zone-cps/port-1/ingresses?page=1&size=50
```

You need to set:

```
KUMA_ZONE_NAME=port-1
```

I didn't think about this too much as once the backend API is there to allow us to filter by zone, we'll have to pass the zone name back to the API/mocks anyway, so once we do that we'll be able to pick up the zone name in the mocks and not have to do this.

johncowen (commented):
Inlines, just to re-stress, the vaaaaaast majority of the change here is just moving things around.

features/application/Titles.feature
```diff
@@ -18,13 +22,13 @@ Feature: application / titles
...
-      | /zones/create                           | Create & connect Zone |
-      | /zones/zone-cps                         | Zone Control Planes   |
-      | /zones/zone-cps/zone-cp-name            | zone-cp-name          |
-      | /zones/zone-ingresses                   | Ingresses             |
-      | /zones/zone-ingresses/zone-ingress-name | zone-ingress-name     |
-      | /zones/zone-egresses                    | Egresses              |
-      | /zones/zone-egresses/zone-egress-name   | zone-egress-name      |
+      | /zones/create                                            | Create & connect Zone |
+      | /zones/zone-cps                                          | Zone Control Planes   |
+      | /zones/zone-cps/zone-cp-name                             | zone-cp-name          |
```
lahabana, line 27: Should this be `overview`?

johncowen: You mean the URL here right? (lmk if not)

To be exact yes, but! `/zones/zone-cps/zone-cp-name` should redirect to `/zones/zone-cps/zone-cp-name/overview` (which it does), so these types of 'incomplete/redirect URL' tests are kinda nice to have.

e.g. In a slightly different test, when I thought I had finished this piece of work and then ran the tests a bunch failed because I hadn't added redirects in certain places

johncowen: Yeah gimme 1 sec I'm just rebasing this now that https://github.com/kumahq/kuma-gui/pull/1459 has gone in

johncowen: K thats rebased, gonna resolve this convo now as I think the rebase resolved it, let me know on the rest @kleinfreund

src/app/zone-egresses/sources.ts
kleinfreund: Do we have a ticket for doing this on the API instead? I *really* want to avoid using client-side filtering like that. It’s unreliable when people have more than 100 resources (even though that’s unlikely, it _can_ happen).

src/app/zone-ingresses/sources.ts
kleinfreund: Same question here: do we have a ticket for filtering this on the API?

src/app/zone-egresses/sources.ts
johncowen: Oh not sure if theres a ticket or not for this, I spoke about this with @lahabana I can check in with him.

src/app/zone-ingresses/views/item/DetailView.vue
kleinfreund: I’m a little confused. What’s the simplification here? Adding this "v-for to introduce template-scoped variable" hack is something we should really not use that liberally.

src/app/zone-ingresses/sources.ts
johncowen: See above

src/app/zone-ingresses/views/item/DetailView.vue
johncowen: Yeah if you look at the removed code here, we were doing the same multiple existence checks twice, now we just do it once.

features/application/Titles.feature
```diff
@@ -18,13 +22,13 @@ Feature: application / titles
...
-      | /zones/zone-ingresses/zone-ingress-name/overview | zone-ingress-name     |
-      | /zones/zone-egresses                             | Egresses              |
-      | /zones/zone-egresses/zone-egress-name/overview   | zone-egress-name      |
+      | /zones/-create                                                    | Create & connect Zone |
+      | /zones/zone-cps                                                   | Zone Control Planes   |
+      | /zones/zone-cps/zone-cp-name/overview                             | zone-cp-name          |
+      | /zones/zone-cps/zone-cp-name/ingresses                            | Ingresses             |
+      | /zones/zone-cps/zone-cp-name/ingresses/zone-ingress-name/overview | zone-ingress-name     |
+      | /zones/zone-cps/zone-cp-name/egresses                             | Egresses              |
+      | /zones/zone-cps/zone-cp-name/egresses/zone-egress-name/overview   | zone-egress-name      |
```
lahabana, line 31: I don't understand why we still have such a nesting.

Shouldn't it be:

```
/zones
/zones/-create
/zones/<zone-name>/overview
/zones/<zone-name>/config
/zones/<zone-name>/ingresses
/zones/<zone-name>/ingresses/<ingressName>/overview
/zones/<zone-name>/egresses
/zones/<zone-name>/egresses/<egressName>/overview
```

johncowen: Yeah sure we can do that I think, I was just following what was there already I suppose.

Ah, also when I did this work it was  before we added back the `-create` (DNS hyphen trick), without that we wouldn't be able to have a URL like `/zone/create` (due to maybe having a zone called `create`).

Lemme have a look now this has been rebased with the `-create` thing.

johncowen: Ah one other things I've spotted, I doubt its an issue, but thought I'd best mention incase people have opinions on it:

If I remove the `/zones/zone-cps` and use `/zones` instead.

1. When Zones are disabled you have: `/zones/egresses` (the egress listing)
2. When Zones are enabled, I can have a zone called `egresses` at `/zones/egresses` (the zone called `egresses` details page), and then `/zones/egresses/egresses` would be its egresses.

That would mean that 1 unique URL can point to 2 different 'documents' depending on whether you have zones enabled or not.

I don't think it has any other consequences, but would rather point it out incase

Lemme know, in the meantime I'm changing all this to remove the `zone-cps` segment as I'm guessing we are fine with this.

johncowen: K `/zone-cps/` has gone, we are now like this:

```
/zones
/zones/-create
/zones/<zone-name>/overview
/zones/<zone-name>/config
/zones/<zone-name>/ingresses
/zones/<zone-name>/ingresses/<ingressName>/overview
/zones/<zone-name>/egresses
/zones/<zone-name>/egresses/<egressName>/overview
```

will resolve this one.

src/app/zone-egresses/sources.ts
johncowen: There's work happening towards this here https://github.com/kumahq/kuma/pull/7999, sounds like that PR is a step one towards this, in the meantime I think we are ok for the moment with the 100 limit, considering the #7999 will be coming soon.

kleinfreund (approved):
**Suggestion**: Please try and avoid moving around/renaming files so liberally. Most of the names and locations that were changed here were perfectly fine. These operations make reviews harder (if they’re necessary and excessive, they should be separate PRs). They also tend to break tools that work with the git history to show a file’s changes (e.g. VS Code’s timeline feature).

Issue #1587, opened 2023-10-09
Show warning when using Global or Standalone in in-memory


lahabana:
### Description

This is not for each zone it's for the main state (`/config`). We should show this always everywhere this can be potentially shown in the top bar next to the upgrade warning.

PR #1589, opened 2023-10-09
chore: adds memory warning persistently if storeType=memory 

johncowen:
Adds/uses new index.html `storeType` variable and shows a persistent warning if that equals `memory`

I also moved `pathConfigDefault` over into `src/services/env` as we only use it for `env` type things so that seemed to make sense.

Lastly, I noticed a CSS issue in onboarding with the radio groups:

Screenshot 2023-10-09 at 11 30 49

 so I fixed that up while I was here.

Closes https://github.com/kumahq/kuma-gui/issues/1587


src/app/zones/views/ZoneCreateView.vue
```diff
@@ -498,3 +498,8 @@ function toggleConfirmModal() {
   isConfirmModalVisible.value = !isConfirmModalVisible.value
 }
 </script>
+<style lang="scss" scoped>
+.radio-button-group > * + * {
+  margin-block-start: $kui-space-40;
+}
+</style>
```
kleinfreund, line 505: **Question**: This shouldn’t be necessary as this is available as a global style. Can you check again whether this is still necessary (could be related to the stylesheet ordering issue I fixed earlier perhaps)?

kleinfreund: I just noticed it’s actually removed from the global styles. Why?

kleinfreund (approved):

PR #1604, opened 2023-10-10
refactor: avoids duplicating title bar template

kleinfreund:
**chore: replaces deprecated faker methods**

**refactor: avoids duplicating title bar template**

Changes AppView so that the title bar template doesn’t have to be duplicated.


johncowen (approved):
Thanks!

PR #1605, opened 2023-10-10
chore: Use mesh-insights for main Meshes listing

johncowen:
TBD


src/app/meshes/components/MeshInsightsList.vue
```diff
@@ -0,0 +1,45 @@
...
+import AppCollection from '@/app/application/components/app-collection/AppCollection.vue'
+import type { MeshInsight } from '@/types/index.d'
+import { useI18n } from '@/utilities'
+
+const { t } = useI18n()
+
+const props = defineProps<{
+  items?: MeshInsight[]
+}>()
+</script>
```
kleinfreund, line 45: **Question**: Why are this and the ZoneDetails component moved into meshes and zones respectively when they’re not used there?

johncowen: Because they are components for rendering Meshes and Zones, you can use them where ever you like i.e. its what they deal with not where they are used which is important. The `control-planes`/`main-overview` module would depend on the `meshes` and `zones` modules to be able to access/use these components.

(I also very much toyed with re-using them for the main Zone and Mesh listings but still not totally sure I want to do that)

lahabana: Should there be a top level `app/components` folder for these components that are used in multiple places?

kleinfreund: There is such a folder: app/common

However, these components _aren’t_ used in multiple places. They’re only used in the main overview hence that’s where they should live also.

lahabana: That's odd I thought the point of extracting the list was to reuse in in `ZoneControlPlanesList` and `MeshInsightsList` though it doesn't seem to be the case. I guess @johncowen needs to explain to us a little more what he's doing here :)

johncowen: We pretty much have an entire application (well multiple entire applications) made up of just micro services/plugins/packages.

You make a different application just by moving around/omitting or including the pieces of those micro services/plugins/packages. You can use those pieces in multiple different places or just one place. Whether you use them in multiple places or a single place doesn't matter. But we don't really want to be moving things around depending on the amount of times we use them.

If you want to build something that has a mesh listing in it, say a "main-overview" you would install the "mesh" package and use the mesh listing component from there, you also use the mesh data source component from there. Then in another (or maybe the same) application you wanted to build a "diagnostics" area that happened to have a list of meshes in it, you wouldn't then install the "main-overview" thing, you'd still install the "mesh" package as before and use the mesh listing component from there. I am guessing we might say, "well, if we use it in main-overview and diagnostics, then its ok for it to be in meshes now" - but again I'd don't want to be moving things around depending on the amount of times we use them, better to put the component in a place that makes sense depending on what it is rather than where we use it.

The important thing is, "what it is, not where it is used". I can use the thing in multiple places, or move where it is used around, but it is always a mesh listing, so it belongs in the "mesh" package.

I can't think of any other instances where a package and then things its contains being named by where the end user could potentially use it rather than what the package contains/what it is.

kleinfreund: While this sounds theoretically nice and desirable, it doesn’t really represent the practical reality of our application. Our modules aren’t independent little packages that we can install at will. The main overview is the place where we use those two components. Why should they live anywhere but where the other main overview components live?

johncowen: Because we should organize things based on what they are not where we use them. For example if I decide to reuse the MeshList component for the `/meshes` listing, what do I do then import from 'main-overview'?

kleinfreund: Personally, organizing by where we use something (in terms of the UI) is **exactly** how we should do it. It’s nice to look at something in the UI and be generally right with the guess that its code comes from the associated module. That’s useful.

johncowen: What did you think of this question:

> For example if I decide to reuse the MeshList component for the /meshes listing, what do I do then import from 'main-overview'?

johncowen: @kleinfreund quick ping on this question ^

kleinfreund: Not sure what I’d prefer then, but since that’s not actually what we’re doing right now, it doesn’t matter.

I really don’t think dislocating components from their associated modules makes sense. There is no advantage in it and it’ll make it harder to find things.

johncowen: I think my point is, I don't want to be moving re-usable components around based on the amount of times they are used.

> But we don't really want to be moving things around depending on the amount of times we use them.

If I were to re-use MeshList for the `/meshes` listing, would that mean its ok to put in `meshes` seeing as I would be using it more than once?

kleinfreund: > I think my point is, I don't want to be moving re-usable components around based on the amount of times they are used.

And I’m not suggesting that. But that’s also not what’s the case here right now. Right now, these components aren’t re-used and tailored to the use of the main-overview module, so please just leave them there.

Yes, the scenario changes a little bit should these components become re-usable across different modules. However, then it’s still preferable to import them across modules (no reason not to do that) because that avoids moving the file (which should be avoided as it breaks some aspects of git history-based tooling).

johncowen: > Yes, the scenario changes a little bit should these components become re-usable across different modules. However, then it’s still preferable to import them across modules (no reason not to do that) because that avoids moving the file

Ok, glad you can see my point however slightly here. Just so I understand correctly/completely, if we were to reuse MeshList in `/meshes` would I import the MeshList component from the `main-overview` module or from the `meshes` module?

Let me know on that one ^

---

Just a small note on this one:

> (which should be avoided as it breaks some aspects of git history-based tooling).

We shouldn't resist making valid changes because we don't want to add entries to git's history. git and specifically git history are for tracking changes, not prevent people from making them. I don't think we should decide not to move a file because it means it affects the history, even if it might take a tinsy bit more effort because of the file move (and also given we may never need to look in the history for that specific point in time)

kleinfreund: > Ok, glad you can see my point however slightly here. Just so I understand correctly/completely, if we were to reuse MeshList in /meshes would I import the MeshList component from the main-overview module or from the meshes module?

> We shouldn't resist making valid changes

I’m of course not talking about valid changes. 🙃

johncowen: 🤔  you think it makes sense in the MeshListView to use a MeshListComponent that is imported from a completely different module/package, in this case something called 'main-overview' ?

You'd say that is equally as ok as importing it from its own `meshes` module/package? Wouldn't you prefer that it was from `meshes`?

johncowen: Well part of this thread is about putting things in places that makes sense, and then not moving them around based on the amount of times we use them (which doesn't make sense).

From what you have said here, I understand that you agree that we shouldn't be moving things around based on the amount of times we use them, and that it doesn't really matter much to you whether this file is imported from main-overview or meshes? Is that right?

johncowen: > Yes, the scenario changes a little bit should these components become re-usable across different modules.

So here I understood that if we were to re-use them them we should change their location?

Again just to confirm, I understand that we both think it would be silly to move files around based on the amount of times we use them.

> Doesn't really matter much to me whether it’s from either of those modules or common at this point.

From here I understood that it didn't matter much to you whether its from either of those modules, or common at this point. i.e. if I plan on re-using these components in zones and meshes you don't care where they are located, but again we shouldn't be moving things around based on the amount of times we use them.

I'm sorry to be repeating what you said back at you, but that's what I understood, and it seemed pretty clear at the time. If you want to change you mind and/or clarify then thats ok too!

kleinfreund: I’m not going to re-iterate all the same points again.

These files were in the right place and should be moved back.

kleinfreund (approved):

Issue #1388, opened 2023-08-25
Represent available services in Zone Ingress details


kleinfreund:
To do:

- Zone Ingress: Represent available services

Done:

- Zone Ingress: Represent networking information
- Zone Egress: Represent networking information

johncowen:
We had a quick chat about this, for an initial build of this we should probably add a [Services] tab to the Zone ingress detail view and add a table/AppCollection with the information we get from the zone-ingress `availableServices` data.

Later on we could move this around, but this would be a good start.

PR #1607, opened 2023-10-11
feat(zones): adds an available services listing tab for zone ingresses

johncowen:
Adds a listing tab for ZoneIngress AvailableServices.

~I've not added any e2e testing for this as yet, I will do once I get feedback whether this is what we want to show here as a first iteration. lmk and I'll write up some e2e tests for it if its good.~ e2e tests added

https://deploy-preview-1607--kuma-gui.netlify.app/gui/zones/driver-0/ingresses/monitor-0/services

Closes https://github.com/kumahq/kuma-gui/issues/1388


src/app/zone-ingresses/views/item/ServicesView.vue
```diff
@@ -0,0 +1,134 @@
...
+      <template #title>
+        <h2>
+          <RouteTitle
+            :title="t('zone-ingresses.routes.item.navigation.zone-ingress-services-view')"
+            :render="true"
+          />
+        </h2>
+      </template>
+
+      <KCard class="mt-4">
```
kleinfreund, line 16: **Suggestion**: This shouldn’t have a margin.

```suggestion
      <KCard>
```

src/app/zone-ingresses/views/item/ServicesView.vue
```diff
@@ -0,0 +1,134 @@
...
+        <template #body>
+          <template
+            v-for="items in [
+              ...(props.data?.zoneIngress?.availableServices ? [props.data.zoneIngress.availableServices] : [])
```
kleinfreund, line 20: **Suggestion**: Please use a computed property here. They should be preferred over in-template operations like this. For one, they results are cached if their dependencies don’t change and avoid needless Vue rendering cycles. In-template logic breaks that. 

For another, this is just terrible to read.

src/app/zone-ingresses/views/item/ServicesView.vue
```diff
@@ -0,0 +1,134 @@
...
+                { label: 'Protocol', key: 'protocol' },
+                { label: 'No. Instances', key: 'instances' },
+                { label: 'Actions', key: 'actions', hideLabel: true },
+              ]"
+              :items="items"
+            >
+              <template #name="{ row: item }: {row: AvailableService}">
+                <template
+                  v-for="service in [
+                    findTag(item.tags, 'kuma.io/service') ?? t('common.collection.none')
```
kleinfreund, line 38: **Suggestion**: Since we need to process the availableServices data further, we should use a computed property and include the results of `findTag` as part of a new array.

kleinfreund (changes requested):

src/app/zone-ingresses/views/item/ServicesView.vue
johncowen: Oh! I copied/pasta'ed the wrong one to build from 😆 the Yaml/Config view had this also, I've removed it from there, guessing that one doesn't need it either? Let me know if so and I can put it back (also removed the same from in egresses). Will push this change in a sec

kleinfreund: It’s wrong in these views as far as I can tell:

- src/app/meshes/views/MeshConfigView.vue
- src/app/zone-egresses/views/item/ConfigView.vue
- src/app/zone-ingresses/views/item/ConfigView.vue
- src/app/zones/views/item/ConfigView.vue

johncowen: K I removed those other two too, that pesky ConfigView 😅

src/app/zone-ingresses/sources.ts
```diff
@@ -38,8 +39,11 @@ export const sources = (api: KumaApi) => {
       res.items = res.items.filter((item) => {
         return item.zoneIngress.zone === name
       })
-      res.total = res.items.length
-      return res
+      return {
+        ...res,
+        total: res.items.length,
```
kleinfreund, line 44: **Suggestion**: Can we leave a note here saying that this needs to change once we can filter this API by Zone name? Currently, this is (both in `master` and in this PR) incorrect if there is more than one page of results (total != items.length).

src/app/zone-ingresses/views/item/ServicesView.vue
```diff
@@ -15,102 +15,95 @@
...
-              </template>
-
-              <template #mesh="{ row: item }: {row: AvailableService}">
+            <template #name="{ row: item }: {row: AvailableService}">
+              <template
+                v-for="service in [item.tags?.['kuma.io/service'] ?? t('common.collection.none')]"
+                :key="service"
+              >
                 <RouterLink
+                  v-if="service.length > 0"
```
kleinfreund, line 35: **Issue**: This can never be `false` unless the service annotation is actually the empty string. If the purpose of this is to be `true` if the service annotation is there, then line 31 above should use `?? null` instead of a `t(...)` result which can’t be predictable.

kleinfreund (changes requested):

src/app/zone-ingresses/sources.ts
johncowen: I added some comment here and in egresses.

Because we are filtering on the frontend. We kinda accepted that for the moment that is ok seeing as ingresses/egresses for a specific zone is never really gonna get past page 1.

src/app/zone-ingresses/views/item/ServicesView.vue
johncowen: Pretty sure these conditionals would never happen anyway, guessing that in order for one of these to be an `AvailableService` it would need to had a tag `kuma.io/service`. I'm just checking in with backend now, but I removed all this and altered the type to try to reflect this a little better. I'll wait for confirmation before merging this tho.

Oh I also changed `common.none` to be `''` instead of `' '`

johncowen: Just had confirmation that assuming that `kuma.io/service` is always set on `AvailableService` is ok

src/app/zone-ingresses/views/item/ServicesView.vue
johncowen: I removed `findTag` entirely, as I figured we could just access the property here

src/app/zone-ingresses/views/item/ServicesView.vue
johncowen: I swapped this out by guaranteeing `availableServices` is always set in the data layer before it ever gets here. Now I just access the property so no need for extra caching or checking for things

kleinfreund (approved):

johncowen:
I was about to merge but I noticed I'd meant to add some basic e2e tests here so I've done that, I'll leave this hanging for a little while in case but otherwise I'll go ahead and merge

lahabana:
@johncowen how can I get a zone with ingresses? I don't seem to be able to find one...

kleinfreund:
> @johncowen how can I get a zone with ingresses? I don't seem to be able to find one...

I think you’re looking for this information: https://github.com/kumahq/kuma-gui/pull/1586#discussion_r1348567841

TL;DR: Set a cookie `KUMA_ZONE_NAME` to the zone you’re looking at for the mocking stuff to know for which zone to provide mocks.

PR #1610, opened 2023-10-12
refactor: Refactors App.vue to use a single component

johncowen:
This PR collapses several components into one `ApplicationShell.vue` component that is then used in `App.vue`.

---

Again much of this is just moving some things around and collapsing multiple 'App' type components into one single `ApplicationShell` component. There's also corresponding integration work elsewhere.

Notes:

- We are left with a very obvious starting point for the application, `App.vue` (which we kinda had before but...)
- We also have a very obvious way to set the logo, main navigation items etc for the application. Even more importantly this is very similar across all applications and the usage of ApplicationShell within App.vue is very clear instead of spread around across multiple components. This consequently makes other things a lot more straightforwards. Additionally, as an engineer with no knowledge of this project I would open up App.vue and go "oh look this is where I add a new navigation item", just as you would expect.
- Loading state now includes the loading of the Vue/application source code itself. The Loading state is based on CSS and should show as early as possible.

Lastly I would guess that now the "ApplicationShell" is all in one place there is probably some clean up to be done in there (I'm pretty much just copy/pasta'ed the multiple components together into one without thinking too much about the CSS used etc), but we can do that as a separate PR at some point.


index.html
```diff
@@ -12,6 +12,44 @@
 
   <script type="application/json" id="kuma-config"><%- config %></script>
   <script type="module" src="/src/main.ts"></script>
+  <style type="text/css">
```
kleinfreund, line 15: **Question**: Any reason these styles can’t live in a file imported in `main.scss`? Then we’d be able to use Sass variables.

features/application/Loading.feature
```diff
@@ -1,3 +1,4 @@
+@skip
```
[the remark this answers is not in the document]
johncowen: Oh sorry yeah, I need to remove this test entirely now instead of skip. Will sort in a bit

index.html
johncowen: Because we want the loader to appear as soon as possible, actually we probably need to move these above everything else. Let me look at that.

johncowen:
> Issue: Why are you deleting all the App* components only to reimplement them in AppCollection? That seems unnecessary and makes reviewing this PR much harder.

It's the objective of the PR. There's no reason for these all to be separate components. Moving them to be one component makes everything a lot easier to follow. Note the reduction in the amount of components we need to re-inject or otherwise fiddle with.

> Please also don’t dislocate the navigational code into other unrelated modules. They don’t belong there.

These are the "navigation to get to the <insert-module-here> module" so they are very related to the module. In fact if you don't install the module, you don't need the navigation code for the module, which illustrates how "related" these are.

features/zones/Create.feature
```diff
@@ -24,28 +24,10 @@ Feature: zones / create
       KUMA_MODE: global
       """
 
-  Scenario: Create Zone link doesn't exist in standalone mode
```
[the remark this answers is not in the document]
johncowen: Because we never have a create zone button on the Egress listing page. There is no code path that renders that.

features/application/Loading.feature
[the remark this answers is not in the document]
johncowen: This is removed

index.html
johncowen: I've moved things around slightly here so the CSS is ordered before the JS script tag. I even tried inlining the gif, but it was a little cumbersome having it inline, maybe there's a vite plugin we can use to do that? If so that can be a improvement after this PR.

i think this one is ok to resolve, lmk if not and we can reopen.

johncowen:
@kleinfreund apart from the discussion around putting components based on "what it is" instead of "where they are used", is there any other feedback here?

index.html
```diff
@@ -22,6 +59,7 @@
   </noscript>
 
   <div id="app"></div>
+  <script type="module" src="/src/main.ts"></script>
```
kleinfreund, line 62: **Question**: Any particular reason to move this? Have you compared the output between both versions?

src/locales/en-us/components/index.yaml
```diff
@@ -1,4 +1,4 @@
 components:
   route-view:
     title: "{name} Manager"
-    route-announcer: Navigated to {title}
+    route-announcer: Navigated to %s
```
kleinfreund, line 4: **Question**: What’s this change for? I kind of really don’t want to introduce a second interpolation syntax.

index.html
johncowen: So that the loader is loaded/parsed before we make this request also see https://github.com/kumahq/kuma-gui/pull/1610#discussion_r1363558356

src/locales/en-us/components/index.yaml
johncowen: I can't for the life of me remember why I did this, maybe it was left over from a previous experiment or something. I change it to use default `t` interpolation, if something comes up that reminds me why it was like this I can bring it up again.

johncowen:
K I've made that review amend and rebased here @kleinfreund, lemme know if this is good to go

index.html
johncowen: P.S. 

> Have you compared the output between both versions?

I have tried it out yes, if thats what you mean? FYI: Ideally I would like the kuma gif to be inlined also, but I didn't want to straight inline it into the file. I was wondering if there was a vite plugin to do that for me, whichever we can add this at a later date.

johncowen:
Is there any more to be done here @kleinfreund ?

src/app/kuma/components/ApplicationShell.vue
```diff
@@ -0,0 +1,280 @@
...
+  border-right: $kui-border-width-10 solid $kui-color-border;
+  background-color:  $kui-color-background;
+}
+nav > ul > :is(li) {
+  margin-left: $kui-space-40;
+}
+nav > ul > :is(li) + :is(li) {
+  margin-top: $kui-space-20;
+}
+nav > ul > :is(li) > :is(a) {
```
kleinfreund, line 237: **Suggestion**: Please use class names on elements you want to select. There’s no need to tie selectors to the HTML structure unless that is useful to enforce (not the case here).

Something like 

```suggestion
.site-nav :deep(.nav-item) {
```

would be much more robust.

In fact, this type of style illustrates how having a dedicated NavItem component was a good idea: it can own its styles instead of relying on a separate component to own them.

src/app/kuma/components/ApplicationShell.vue
```diff
@@ -0,0 +1,280 @@
...
+  top: var(--AppHeaderHeight);
+  bottom: 0;
+  left: 0;
+  overflow-y: auto;
+  padding-top: $kui-space-40;
+  padding-right: $kui-space-40;
+  border-right: $kui-border-width-10 solid $kui-color-border;
+  background-color:  $kui-color-background;
+}
+nav > ul > :is(li) {
```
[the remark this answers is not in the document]
johncowen: I changed this, but from what I understood its not a bug:

https://github.com/vuejs/core/discussions/8800#discussioncomment-6473724

> :is It is not an entity selector, but a modification selector, which is a supplement to the previous one, so there is no problem with the above compilation.
> Normal logic, because :is is not a entity selector, but a adjective selector. Same as :after

(I google translated the first line)

src/app/kuma/components/ApplicationShell.vue
johncowen: I added an `AppNavigator` component that wraps the slot in a `li` with a class name on it and now use that to internally wrap all the `Navigator`s. I then use that class from ApplicationShell to apply the styles.

src/app/App.vue
```diff
@@ -110,4 +68,7 @@ const isWizard = computed(() => route.meta.isWizard === true)
   // Note: `minmax(0, 1fr)` is used because `1fr` implies `minmax(auto, 1fr)` which will allow grid items to grow beyond their container's size.
   grid-template-columns: var(--AppSidebarWidth) minmax(0, 1fr);
 }
+img {
```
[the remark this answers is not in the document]
johncowen: I change from `img` to `logo` here, resolving

johncowen:
I've addresses this round of review comments @kleinfreund , lmk!

src/app/kuma/components/ApplicationShell.vue
kleinfreund: Sounds good 👍🏻

kleinfreund (approved):

PR #1611, opened 2023-10-13
feat: adds summary tray to service list view

kleinfreund:
**This is a draft PR; do not review**

Depends on https://github.com/kumahq/kuma-gui/pull/1601 and https://github.com/kumahq/kuma-gui/pull/1604.

---

## Changes

**chore: removes unused RouteView slot method**

**chore: adds optional tray slot to AppView**

Adds an optional `tray` slot to AppView.

Moves the `.app-main-content` styles into AppView (where the `.app-main-content` element is).

**refactor: removes name parameter from router.push/replace calls in RouteView**

Removes `name` being passed to calls to `router.push` and `router.replace` in `RouteView`. This otherwise breaks summary trays in that they will cause a view to be re-rendered when closing a tray.

**feat(services): adds summary tray**

Adds a summary tray to the service list view.

## Screenshots


src/app/services/views/ServiceTrayView.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <TrayView @close="emit('tray-close')">
```
johncowen, line 2: I wasn't sure whether to comment here or not, but since https://github.com/kumahq/kuma-gui/pull/1613#discussion_r1361896688 I think you want me to comment here?

All route components should be wrapped with RouteView and AppView components.

kleinfreund: > All route components should be wrapped with RouteView and AppView components.

Why? It needs none of the things provided by those components plus this component isn’t meant as an independent route view. It cannot be used without the corresponding list view for which we have RouteView and AppView.

johncowen (commented):
Couple of inlines/q's:

src/app/services/views/ServiceListView.vue
```diff
@@ -124,13 +133,33 @@
...
+              <component
+                :is="child.Component"
+                :data="selectedService"
+                @tray-close="route.update({
+                  page: route.params.page,
+                  size: route.params.size,
+                })"
+              />
+            </RouterView>
+          </template>
```
johncowen, line 162: All sub views of a listing view will be trays right? Why not just put the Tray component here? (I'd say it almost belongs here in the listing view in the way that they are both super related, i.e. you need the list to see the tray)

I think this would also make the code clearer (at the moment its hard to see that this is a more or less a tray component). You also wouldn't need to do the extra plumbing we have in the ServiceTrayView to pass through `@close`.

Another benefit is that in the cases that some trays might be the same as the detail view, we could potentially avoid multiple/doubling route components for essentially the same content. i.e. we could just reuse the detail view inside of the tray. I think from our sync the other day we said that it is probably quite common that the content in the tray and the detail view is exactly the same.

src/app/application/components/route-view/RouteView.vue
```diff
@@ -140,10 +139,7 @@ const routeUpdate = (params: Record<string, string | undefined>) => {
   }
   routerPush(newParams)
 }
-const routeReplace = (...args: Parameters<typeof router['push']>) => {
-  router.push(...args)
-}
```
johncowen, line 145: I am unsure about removing this, I need to take another look at this again though.

Ideally `RouteView:route:update` should update the parameters of the route you are on as a user, and `RouteView:route:replace` should replace the route entirely with a differently named route. i.e. if as a user I'm on `service-summary-view` but as an engineer coding in the `service-list-view` file then `update(props)` should leave the user on the same URL pathname but update the query params on that URL

With that being said, below you are using `update` to change the route entirely back from the summary-view to the list view, whereas the intention is you should use `replace` to do this, i.e. replace the route with an entirely differently named one.

I think the fact that what you do below works but is a bug in the implementation that we haven't realized just yet because we haven't needed this usecase before now.

Maybe wrote this a bit early, but wanted to at least express a bit of uncertainty about whether we should remove this.

src/app/services/routes.ts
```diff
@@ -38,6 +38,13 @@ export const routes = () => {
             module: 'services',
           },
           component: () => import('@/app/services/views/ServiceListView.vue'),
+          children: [
+            {
+              path: ':service',
+              name: 'service-tray-view',
+              component: () => import('@/app/services/views/ServiceTrayView.vue'),
```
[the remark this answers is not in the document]
kleinfreund: That is the name I had initially, but I changed it halfway through development. I’ll reconsider it again.

src/app/services/views/ServiceListView.vue
kleinfreund: > All sub views of a listing view will be trays right? Why not just put the Tray component here? (I'd say it almost belongs here in the listing view in the way that they are both super related, i.e. you need the list to see the tray)

That’s because we hook up ServiceTrayView via its tray route which means ServiceListView must have a `RouterView` rather than a ServiceTray component directly.

I’m not sure there’s an alternative here.

> Another benefit is that in the cases that some trays might be the same as the detail view, we could potentially avoid multiple/doubling route components for essentially the same content. i.e. we could just reuse the detail view inside of the tray. I think from our sync the other day we said that it is probably quite common that the content in the tray and the detail view is exactly the same.

I don’t know how common this actually will be. Re-using the detail **view** components for the summary trays is problematic because their similarity is incidental and wouldn’t always work. I rather extract non-view components like ExternalServiceDetails and ServiceInsightDetails for easy re-use. Additionally, the layout of the detail view compared to the summary view might be sufficiently different to prevent re-use in the respective other.

src/app/application/components/route-view/RouteView.vue
kleinfreund: How’s what I’m doing behaviorally different?

src/app/services/views/ServiceTrayView.vue
kleinfreund: I’m not sure about this. I think we might already potentially have a bit too much nested titles going on. Personally, I’d prefer titles to at most have two levels of nesting.

On the service list view, we currently have "Services | default | Kuma". I don’t think keeping "| Kuma" makes sense (it does on the home page). Adding the selected service via a ServiceTrayView title would result in "service-1 | Services | default | Kuma".

So maybe doing titling in the list view makes more sense to we would go to "service-1 | default | Kuma" for the tray view and "Services | default | Kuma" for the plain list view?

src/app/services/views/ServiceListView.vue
johncowen: > That’s because we hook up ServiceTrayView via its tray route which means ServiceListView must have a RouterView rather than a ServiceTray component directly.

Yeah I getcha, but can we put the tray inside the RouteView but wrapping the `<component />`? That way all sub views of the listing are trays, and we don't have to repeat the tray for every single sub route in its *View.vue file (and the tray not being in the *View.vue file means its possible to reuse the DetailVue.vue file without much trouble)

> I don’t know how common this actually will be. Re-using the detail view components for the summary trays is problematic because their similarity is incidental and wouldn’t always work. I rather extract non-view components like ExternalServiceDetails and ServiceInsightDetails for easy re-use. Additionally, the layout of the detail view compared to the summary view might be sufficiently different to prevent re-use in the respective other.

Yeah maybe I suppose we will see, but pretty sure we said that them both being the same (or even very similar) would be reasonably common.

Leaving View reuse aside, overall on this point (whether we reuse the DetailView.vue's or not), I think its worth putting the tray inside the `<RouterView>` but wrapping the `<component>`. Does what I'm saying make more sense now? From reading your response, its sounds like I'd not explained properly and you'd understood that I was suggesting that the tray view would be _instead of_ the RouterView, which is not what I meant. lmk

src/app/application/components/route-view/RouteView.vue
johncowen: So here:

```javascript
@tray-close="route.update({
  page: route.params.page,
  size: route.params.size,
  })"
```
Given `update` is meant to "update" the route the user is on with new params/query params (not change the route), the code above is using `update` to change the route from the summary route back to the list view, not solely update the params.

I think the code works now because `update` is slightly wrong (i.e. theres a bug there). I would prefer this code above to use `replace` (to "replace" the summary view with the list view), and then we would fix the bug with `update` to always use the route the user is on rather than the route where you are writing the code.

Going with the intention of how `update` should work, if it was fixed to be like that and you were to call `update` here instead of `replace`, you would stay on the summary view and only update the params, not return to the listing view.

Does that help?

src/app/services/views/ServiceListView.vue
kleinfreund: > Yeah I getcha, but can we put the tray inside the RouteView but wrapping the `<component />`? That way all sub views of the listing are trays, and we don't have to repeat the tray for every single sub route in its *View.vue file

Ah, now I understand. You want me to put the TrayView content except its default slot content into the list view. I’ve done that now.

> (and the tray not being in the *View.vue file means its possible to reuse the DetailVue.vue file without much trouble)

That’s something we just shouldn’t do. The detail views are **generally** not the same as the summary content. I don’t even want to try making them the same by re-using the views even that might be possible in some cases. In fact, it’s already not a 1-to-1 equivalence for services because for external services we show the YAML in the summary but not the detail. Much easier to re-use non-view components here.

src/app/application/components/route-view/RouteView.vue
kleinfreund: So, just so I’m understand what you’d change in RouteView. You would use `route.name` here, right?

```ts
const routerPush = beforePaint((params: Record<string, string | undefined>) => {
  router.push({
    name: route.name as string, // instead of `props.name`
    query: cleanQuery(params, route.query),
  })
  newParams = {}
})
```

johncowen: Umm possibly, if that makes it work as per what I explained before yeah, I think that would make it work as explained. Does that effect anything we've already done?

I think my main thing here is to use `replace` to replace the entire route, the functionality of `update` can be sorted later if need be. I'd be careful tweaking it as part of a feature PR if you were to tweak it here.

kleinfreund: I was just asking what you’d change so I know what its impact is. I don’t know if it breaks something we already built.

I’ve removed the removal of this code and switched to `route.replace`. I’m still not totally convinced this makes sense. I still don’t like the name `replace` for an action that doesn’t do a `router.replace`. It’s just too close.

src/app/services/views/ServiceListView.vue
kleinfreund: I’m resolving this now. The TrayView use was moved to the list view.

src/app/application/components/route-view/RouteView.vue
johncowen: > I was just asking what you’d change so I know what its impact is. I don’t know if it breaks something we already built.

Ah gotcha, we are on the same page then. tbh I'm not super sure either I'd have to take a closer look to be sure.

> I’ve removed the removal of this code and switched to route.replace. I’m still not totally convinced this makes sense. I still don’t like the name replace for an action that doesn’t do a router.replace. It’s just too close.

If there's another word we can use that means the same thing ("replace the current route with an entirely new one") we could change that also.

We can do that separate tho, let me look at the changeset here again now you've made these changes, brb

johncowen:
Ok based on some of the conversations above I had a quick go at a potentially throw away branch/commit ontop of here to satisfy my curiosity:

https://github.com/kumahq/kuma-gui/commit/dfdd7a8bc8a897a52b47d479a6d60fcc2b829b5f

Additions:

- The sub route/view uses all the same patterns we already have in the application for consistency and reuse:
  - RouteView/RouteTitle: route utilities and titling
  - AppView: headers/titling
  - stack and KCard: layout and spacing
- Kinda personal opinion that I'd be happy to change if more people in the team were against, but I would rather that when I click off the tray that it should close. So I removed the `preventCloseOnBlur`. I'm also guessing that it defaults to false because the closing behaviour is probably the preferred option.
- I used `:is` in the CSS to:
  - Make sure we can keep using `<style scoped>`
  - Remove the `!importants`
  - In chatting to you I realized just how coupled the 'tray' was to `AppCollection`, so I renamed it to `AppCollectionSummary`. I'm pretty sure I like this, wdyt?

I think overall the best thing about this is: Codewise, the summary route component looks exactly the same as all our other route components, it uses the same components, the same code and has the same functionality as all our other routes.

(P.S. I did take a quick look at view re-use if we ever wanted to do that, but I didn't bother to PR that here seeing as we are currently unsure if we want to do that)

Let me know what you think.

src/app/services/views/ServiceTrayView.vue
johncowen: The biggest thing is that we should be using RouteView/RouteTitle/AppView in the route component, not only for titling purposes. I don't know why we wouldn't want to use these components, which is half the reason I tried it out in case I was missing something (see comment below with an additional commit)

Is there a reason I'm not aware of why we aren't using these components in the route component?

kleinfreund: The more important questions is: why do we need them here?

If there is a reason to use them, I’m all for it.

johncowen: Sorry missed this bit:

> So that’s a standalone suggestion then? Doesn’t really seem like it has anything to do with RouteView/AppView at the moment as neither component uses either stack or KCard.

Its all about consistency and coherence.

src/app/services/routes.ts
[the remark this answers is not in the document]
kleinfreund: I decided to keep this naming.

src/app/services/views/ServiceTrayView.vue
johncowen: Did you notice that we use AppView (and also RouteView) for "wizards" (e.g. zone creation)

We use AppView there for the header/title, we should also use AppView here for the header/title.

kleinfreund: > We use AppView there for the header/title, we should also use AppView here for the header/title.

That example doesn’t relate to this change set because neither header nor title are used here. Not a bit of functionality from RouteView or AppView is currently needed in ServiceTrayView.

johncowen: I think the issue is that AppView isn't in the changeset, we do use a header title here, but you aren't using AppView to do that (like everywhere else in the application)

Here's some grabs of our headers/titles:

Screenshot 2023-10-20 at 09 49 48


Screenshot 2023-10-20 at 09 50 00

They all should use AppView to do that.

---

Did we get a decision on whether we are changing the unique HTML title when we open the tray?

src/app/services/views/ServiceListView.vue
```diff
@@ -21,11 +22,19 @@
           <template #title>
             <h2>
               <RouteTitle
+                v-if="route.params.service"
+                :title="t('services.routes.items.tray_title', { name: route.params.service })"
+                :render="true"
+              />
+
+              <RouteTitle
+                v-else
```
johncowen, line 31: Why are we change this title based on which service you have in the tray. Its the header for the list so should stay as 'Services'

src/app/services/views/ServiceTrayView.vue
kleinfreund: It would’ve been really helpful to know that you’re talking about the on-page title that we’re currently passing to KSlideout and not the document title.

As we’re using KSlideout for implementing trays, we should use its mechanisms for titling purposes. What you’re proposing in our alternative implementation branch is to avoid using KSlideout’s title entirely.

That difference makes me think more strongly that it would be wrong to use AppView/RouteView here. Using KSlideout’s title mechanisms (hopefully with the availability of a `title` slot in Kongponents v9) will ensure our trays look the same across different uses of KSlideout.

@adamdehaven, do you perhaps know if there are plans for other changes to KSlideout that you already know will happen as part of the development of Kongponents v9?

kleinfreund: > Did we get a decision on whether we are changing the unique HTML title when we open the tray?

I’ll bring that up once we start actually reviewing the concrete service trays. For now, I’m focusing on getting the underlying framework out so following implementations aren’t blocked by it.

src/app/services/views/ServiceListView.vue
kleinfreund: I already explained this here: https://github.com/kumahq/kuma-gui/pull/1611#discussion_r1365099922

Could you please focus on reviewing https://github.com/kumahq/kuma-gui/pull/1613 and not the details of the concrete service tray implementation that are completely unrelated to it?

src/app/services/views/ServiceTrayView.vue
johncowen: Oh well I'm glad those screengrabs helped then! I mentioned this in the original list of things we use AppView for (AppView does headers, not HTML titles)

> The sub route/view uses all the same patterns we already have in the application for consistency and reuse:
> RouteView/RouteTitle: route utilities and titling
> **AppView: headers/titling**
> stack and KCard: layout and spacing

---

> That difference makes me think more strongly that it would be wrong to use AppView/RouteView here. Using KSlideout’s title mechanisms (hopefully with the availability of a title slot in Kongponents v9) will ensure our trays look the same across different uses of KSlideout.

You've already pointed out the issues you were having with using the title in KSlideout, and previous to that you didn't even want to use KSlideout to do this because of the issue you had found you built your own component instead, until I suggested that we should be using KSlideout - https://github.com/kumahq/kuma-gui/pull/1613#discussion_r1360472663

I did mention I while back that we should just put our `h*` headers inside the KSlideout content instead of the title attribute, until such time that the work in KSlideout has settled down.

> [...] I do think we are ok not using this title attribute and just using <h*>s in the default slot for our titles/headers. We can also control any icon from there aswell I think.

We've already documented the issues we have in using the KSlideout title attribute, and you've mentioned quite a few times that you've had to hack around KSlideout in order to implement what we need. 

I would much rather just not use the title attribute in KSlideout rather than hacking around it. Hacking around it means it's much more likely to break as work progresses on KSlideout, and potentially means that the people working on that need to be mindful of any hacks we are using.

> For now, I’m focusing on getting the underlying framework out so following implementations aren’t blocked by it.

As I'm mentioned offline, the implementation in this PR is super important as we want to roll this out everywhere. If we need to set the title in the application when somebody opens a tray (which I would guess we do seeing as we do that when you click on a tab in other areas of the UI), then we need to support that in the implementation here in this PR, and that means using RouteView/RouteTitle.

I'm finding it hard to understand what the big problem is in following the patterns we have in the application. Even if if were the case that offered no additional functionality to what we need to do, I would still argue that for the sake of consistency/coherence they are worth using. By doing this you can make several assumptions about the application as a whole (instead of "oh! great! <insert-solution> will work because we use RouteView in all our routes...oh wait... we didn't use it in the tray 🤦 "). In many case you just automatically do things without having to think knowing that it will be ok as the application all uses the same approach. It also makes it easier to make decisions on how to build things, we just do it the same as the rest of the application.

src/app/services/views/ServiceListView.vue
johncowen: I think this is related to the fact that  we aren't using RouteView and RouteTitle to set the title in the Summary route and trying to do it in the list route instead.

johncowen: As I mentioned offline, #1613 is just a wrapper around KSlideout, there's not a lot to review without seeing it in context. I've left some more comment over there also, but honestly I think we should just close #1613, the code in it is here also, plus all of the additional code in this PR is much more important as that the stuff we want to roll out over the entire application.

src/app/services/views/ServiceTrayView.vue
kleinfreund: I have no problem at all "following the patterns we have in the application". I do so every day. I do so in this PR, too.

This stops when it’s problematic to do so. It’s not striving for consistency to avoid using KSlideout’s title mechanisms but against it.

What you’re suggesting is to only use KSlideout’s default prop so we’re able to leverage AppView and RouteView inside it instead of KSlideout’s built-in title layout (whose problematic aspects I’m in the process of addressing so we can ignore those here).

If we can’t achieve consistent titling with our current patterns while also using KSlideout’s provided title mechanisms and layout, then our current patterns for that purpose aren’t flexible enough to satisfy our use cases.

It’s more important that various uses of KSlideout look the same and we’re unlikely to reliably achieve that by not using its title mechanisms.

johncowen:
Separate to all the conversation here on the code:

Functionally if I enter say the following URL:

https://deploy-preview-1611--kuma-gui.netlify.app/gui/meshes/matrix-0/services/johns-service?page=2&size=50

And johns-service has moved to page 1 because I've removed a bunch of services, I should still see page 2 of the list with the panel opened on john-service, even if johns-service isn't in the list I'm looking at.

src/app/services/views/ServiceListView.vue
kleinfreund: > the code in it is here also

The code is only in here also because we are, as you know, not able to create pull requests against other pull requests coming from a fork.

> I think we should just close #1613 

I’m not closing this PR because it landing will allow us to continue implementing summary trays in the application. It also makes PR review simpler to work in smaller chunks.

johncowen: Yeah I getcha, sorry I'm not complaining about the code being here, in fact is easier for me to review here as its an implementation plus a single example of how we would use it, which in this case is very important.

> I’m not closing this PR because it landing will allow us to continue implementing summary trays in the application. It also makes PR review simpler to work in smaller chunks.

Ok, well the most important thing for me is how we are to use the tray, I don't really mind too much the innards of it as we can always change that in one place. Feel free to leave that one open, but theres already a lot of conversation here (I think you directed me to here a while back), so if its ok with you, considering all the code is here also, can we keep the reviewing in this PR?

kleinfreund:
> And johns-service has moved to page 1 because I've removed a bunch of services, I should still see page 2 of the list

That’s happening.

> with the panel opened on johns-service, even if johns-service isn't in the list I'm looking at.

Let’s discuss with the team whether this should be the case. When I spoke with Charly, we were thinking the URL should change to the plain list view (i.e. without `/johns-service`) or to show some sort of messaging explaining that the service wasn’t found, but this wasn’t completely conclusive.

src/app/services/views/ServiceTrayView.vue
johncowen: > This stops when it’s problematic to do so.

I don't understand why its problematic to use RouteView/RouteTitle and AppView? You argue that there is no need to, but leaving that aside why is it problematic? Does that commit (https://github.com/kumahq/kuma-gui/commit/dfdd7a8bc8a897a52b47d479a6d60fcc2b829b5f) I added on the end here show its not problematic? 

Where is the problem there?

johncowen:
> or to show some sort of messaging explaining that the service wasn’t found

Whats the idea there, then would I have to click through the pages of the list to re-find the service I'm looking for and then re-click on it even though its already in the URL?

src/app/services/views/ServiceListView.vue
kleinfreund: > Ok, well the most important thing for me is how we are to use the tray, I don't really mind too much the innards of it as we can always change that in one place. Feel free to leave that one open, but theres already a lot of conversation here (I think you directed me to here a while back), so if its ok with you, considering all the code is here also, can we keep the reviewing in this PR?

I’m sorry, but no, it isn’t. I’ve been quite clear about what’s up for review and it’s not this PR but #1613. This is indicated by this PR being in draft state and #1613 not being in draft state. It’s also clear from me specifically asking about the review status of #1613 on a few occasions.

I _don’t_ want this PR to be reviewed first because that defeats the purpose of working in smaller chunks of changes. It’s important to see how #1613 is used and that’s why this PR is available, but pretty much all review comments on this PR are now about it and not #1613.

So, can I please ask you to review #1613 so we can both unblock the following summary tray implementation tasks _and_ consider this PR for what it is?

src/app/services/views/ServiceTrayView.vue
kleinfreund: I did previously explain this [here](https://github.com/kumahq/kuma-gui/pull/1611#discussion_r1366698912):

> As we’re using KSlideout for implementing trays, we should use its mechanisms for titling purposes. What you’re proposing in our alternative implementation branch is to avoid using KSlideout’s title entirely.

and [here](https://github.com/kumahq/kuma-gui/pull/1611#discussion_r1366771022) (the message you’re replying to):

>What you’re suggesting is to only use KSlideout’s default prop so we’re able to leverage AppView and RouteView inside it instead of KSlideout’s built-in title layout (whose problematic aspects I’m in the process of addressing so we can ignore those here).

> If we can’t achieve consistent titling with our current patterns while also using KSlideout’s provided title mechanisms and layout, then our current patterns for that purpose aren’t flexible enough to satisfy our use cases.

The technical difficulty stems from the fact that both AppView and KSlideout demand being provided a title for their respective mechanisms. We can’t do both because we’d end up with two titles. We shouldn’t use AppView’s mechanism because that would mean not using KSlideout’s which I consider to be higher priority for the sake of consistent across uses of KSlideout.

So, the problematic aspect here is UI-visible titling using AppView not so much RouteView and RouteTitle.

johncowen: I don't think I'm explaining myself properly. I'm asking what technical problem is caused adding a `AppView` component (whether we use the header/title slot from it or not)?

---

Separately:

> We can’t do both because we’d end up with two titles.

Currently if we use the `title` slot from TrayView and the `:title` attribute KSlideout we would end up with two titles, what you've done here is just not use the `:title` attribute, as its an optional slot thats fine to do that. If we were to use the title slot from AppView, we would just not use the optional `:title` attribute from KSlideout.

johncowen: Worth noting there, that in using the `before-title` slot we don't use the styles specified inside KSlideout title.

So any discussion around not using AppView#title for design consistency reasons is not really a valid reason seeing as we aren't using the styles specified in KSlideout for the title in this PR.

Also worth nothing that we avoid using the title slot in KCard everywhere in our application for the very same reasons and thats never been an issue before.

src/app/services/views/ServiceListView.vue
johncowen: I've added similar concerns over on #1613

kleinfreund:
**Note**: I think we need to add `props.preventCloseOnBlur` after all. I just tried it again and the behavior I described before is there: one interaction to close, one interaction to open a different summary tray.

src/app/services/locales/en-us/index.yaml
```diff
@@ -9,5 +9,8 @@ services:
         service-config-view: 'YAML'
     items:
       title: Services
+      tray_title: "{name}"
   detail:
     no_matching_external_service: 'No matching ExternalService was found for service {name}'
+  overview: 'Overview'
+  config: 'YAML'
```
johncowen, line 16: Could these things be nested somehow, seems strange to me that we put the tray header/title in `services.routes.items.tray_title` but the sub headers/titles in `services.overview/config`.

Honestly looking at the commonalities here, we should probably reuse the terms from `services.routes.item`, theres possibly something to be said in certain cases for sometimes not reusing the route components for both the item and summary view, but for the actual terms we use in both of those, I'm pretty sure in the vast majority of cases we probably want to use the same text.

src/app/services/views/ServiceTrayView.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <AppView>
+    <template #title>
+      <div class="tray-title-wrapper">
+        <img
+          aria-hidden="true"
+          src="@/assets/images/icon-wifi-tethering.svg?url"
+        >
```
johncowen, line 8: Is this icon going to be common to all tray views do you know?

Also given we are using `aria-hidden` would it be better to use CSS to add this icon?

johncowen (commented):
Thanks for using AppView here, I added a few more comments/questions but I'm going to take it for another spin now we've made a few more changes here, in the meantime let me know on those.

src/app/services/views/ServiceTrayView.vue
kleinfreund: > Is this icon going to be common to all tray views do you know?

No. It just so happens that we currently have the same icon (in the main overview) for both services and data plane proxies. Ideally they’d be different, but that’s for another time.

> Also given we are using `aria-hidden` would it be better to use CSS to add this icon?

I find it easier to deal with them like this.

src/app/services/locales/en-us/index.yaml
kleinfreund: I switched the tray’s title out for the item title, but I’m not comfortable using the `navigation` key entries seeing as they’re keyed with route names.

src/app/services/views/ServiceTrayView.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <AppView>
```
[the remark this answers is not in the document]
kleinfreund: I wrapped this in RouteView now.

src/app/services/locales/en-us/index.yaml
johncowen: Hmm yeah thats a good point! Could we maybe move these last ones to `services.routes.item.overview/config` then? That way they are all in the same place. I think it was said a while back that general it didn't make sense for some i18n to come form one places and some from another for no apparent reason.

src/app/services/views/ServiceTrayView.vue
```diff
@@ -0,0 +1,85 @@
...
+</template>
+
+<script lang="ts" setup>
+import ExternalServiceConfig from '../components/ExternalServiceConfig.vue'
+import ExternalServiceDetails from '../components/ExternalServiceDetails.vue'
+import ServiceInsightDetails from '../components/ServiceInsightDetails.vue'
+import { ServiceInsight } from '@/types/index.d'
+import { useI18n } from '@/utilities'
+
+const { t } = useI18n()
```
johncowen, line 68: We may aswell just use `<RouteView v-slot="{ t }" >` if possible? Its what we do everywhere else.

src/app/services/views/ServiceTrayView.vue
```diff
@@ -0,0 +1,85 @@
...
+            </RouterLink>
+          </h2>
+        </div>
+      </template>
+
+      <div class="stack">
+        <div>
+          <h3>{{ t('services.overview') }}</h3>
+
+          <div class="mt-4">
```
johncowen, line 33: How come `stack` spacing isn't being applied here, is it because of that extra div I was asking about?

src/app/services/views/ServiceTrayView.vue
kleinfreund: See https://github.com/kumahq/kuma-gui/pull/1611#discussion_r1367041210.

src/app/services/views/ServiceTrayView.vue
kleinfreund: See https://github.com/kumahq/kuma-gui/pull/1611#discussion_r1367041210.

src/app/services/views/ServiceTrayView.vue
```diff
@@ -0,0 +1,83 @@
...
+      <div v-if="props.data.serviceType === 'external'">
+        <h3>{{ t('services.config') }}</h3>
+
+        <ExternalServiceConfig
+          class="mt-4"
+          :mesh="props.data.mesh"
+          :service="props.data.name"
+        />
+      </div>
+    </div>
```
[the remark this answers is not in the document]
johncowen: Oh sorry just seen this, one sec (sorry for making that other comment, l'll resolve that one and the convo can be here)

src/app/services/locales/en-us/index.yaml
kleinfreund: At this point I’m not sure why we introduced the `routes.item[s]` structures. What’s the point of putting all translations there? Might as well drop them put things on the top-level with minimal nesting. I can see some value in there being those `routes.item[s]` structures, but that vanishes if we just put everything on them that doesn’t relate to routes anymore.

src/app/services/views/ServiceTrayView.vue
kleinfreund: Honestly, I’m still confused why `t` is even provided on RouteView. Stuff like this should just be imported. I have no interest in warping these global components into some kind of import shorthand for utilities. It has no advantage other than saving the import.

src/app/services/locales/en-us/index.yaml
johncowen: You're maybe right, but not sure, maybe here it doesn't seem that necessary but in other places it might. I think its just easier to be like "I'm in the item route, I need the stuff on the item property". I think also there are maybe cases where you'd have like `routes.items.introduction` and `routes.item.introduction` if we were to have like a short one sentence introductory sentence at the top of each view.

I think for the sake of these two bits of text though, shall we just follow whats there for now? We can chat about the i18n at a later date, I think at some point soon we should do a "i18n" effort similar to what we did with out "e2e testing" effort.

I suppose I'm not too precious over this one, so up to you.

kleinfreund: I moved those now.

src/app/services/views/ServiceTrayView.vue
johncowen: One reason is because it would be nice to be able to do `t('.introduction')` and not `t('services.routes.item.introduction')` (and that same thing repeated lots and lots of times in a route file).

If you don't use RouteView::t just in these files, that won't work here.

---

I've mentioned before that Vite's hot reloading works because it knows about imports, therefore if you change something in this file, it can reload everything else imported in this file, that can also then reload the TOKENs which are based on symbols, if the symbols change then that means you can sometimes get errors on a hot reload to do with the DI container not being able to find the right token (because you reloaded only a small part of the application). We mostly change route components (after all route components are mainly made up of pre-made reusable components that we edit rarely)

Again to note, this is only an issue whilst in development and this alone wouldn't lead me to want to do things like this. 

---

I've mentioned quite a few times before that it is good DX to make good choices easy and less good ones harder. Having `t` available in every single route means its more likely you will use it there and pass it into components via slots/attributes. We could potentially restrict this further by making it difficult to import `t` at all without a lint ignore via eslint (not saying I defo want to do that)

---

Separate to that, we do this everywhere else in the entire application in every single one of our RouteView's, why should this route take a different approach?

src/app/services/views/ServiceTrayView.vue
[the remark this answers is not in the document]
johncowen: Cool I just looked at this, I see what you means, ty! resolving this one

src/app/services/views/ServiceTrayView.vue
kleinfreund: I have to be honest, this is sounds very unconvincing. I’m strongly against preferring access of global utilities like i18n or can via slots instead of imports. The latter are vastly simpler and more transparent and there is no reason to impose such a restriction.

I understand that this makes sense if you want to automatically handle something like automatic i18n key trimming, but that sounds unnecessarily complex and not very useful.

Imports are the right tool for this job.

johncowen: They aren't global. The below is pretty much what we are doing and I don't see a problem with this approach. Its what we in do in more "traditional JS" for all our mocks also

```javascript
import RouterView from 'RouterView.vue'
import RouteView from 'RouteView.vue'
RouterView({}, {} => {
  RouteView({ name: 'service-list-view' }, ({can, t, route}) => {
    if(can('use routes') && route.name) {
        return t('services.routes.label', {name})
    }
  })
})
```

Passing the 'tooling' you need specific to the thing you are using as an argument of a callback in JS is an extremely common pattern, not only in our application but in the wider ecosystem. It's one approach to _avoid_ using globals.

Here's the same approach in all our mocks:

```javascript
export default ({env, fake}) => {
   return (params) => {
       return {
          name: params.name,
          createdAt: fake.date()
       }
   }
}
```

---

The reason to not use `t` or `can` inside reusable components, is that passing the data down into the reusable component instead makes it far more reusable.

We discussed this reasoning when we started adding out i18n support a while ago, and we said by default we would use attrs or slots to pass i18n text into components, but it would also be fine to also use `t` inside a reusable component where it made sense. I think this was written into the MADR for that. I think this argument is also very similar for the `h*` discussion we had on this PR somewhere. Its far better to not use `h*` elements inside reusable components as it makes the level incorrect when used in certain places i.e. it makes them less reusable.

---

I've given a good few reasons and advantages with examples and use cases now why we do things like this in this circumstance. I've also explained how this pattern is extremely common, not only in our application but in the wider ecosystem.

As far as I understand it at the moment, your argument for not using them is that you have a personal preference here to use imports. You say things like "its vastly simpler and more transparent" and "there is no reason to impose restrictions" without backing those statements up with any facts.

Considering that currently we use heavily use this approach everywhere, I think if you want to change how we access these utilities, then it would be good to put together/document a convincing argument outside of this PR.

If that argument is convincing enough and you can show its more advantageous then I'd happily change approach. But all that should happen outside of this PR, and I don't think we should change our current approach just in this instance and just in the code added in this PR.

johncowen: @kleinfreund I've just unresolved this again. Let me know.

kleinfreund: > They aren't global.

I mean this in the sense that they're globally-available utilities. Utilities that can be used everywhere unconditionally. Not actual global `window` properties. And they should be globally available. We can use them virtually everywhere in our application. That's a good thing.

> The below is pretty much what we are doing and I don't see a problem with this approach.

It works, sure, but that doesn't mean we should rule out using imports. Imports are by far the preferable mechanism to gain access to things in JavaScript. I fail to see how most of your arguments actually negate that that’s appropriate.

> It's one approach to _avoid_ using globals.

This isn’t about globals. And we shouldn't avoid using those utilities.

> The reason to not use `t` or `can` inside reusable components, is that passing the data down into the reusable component instead makes it far more reusable.
> 
> We discussed this reasoning when we started adding out i18n support a while ago, and we said by default we would use attrs or slots to pass i18n text into components, but it would also be fine to also use `t` inside a reusable component where it made sense. I think this was written into the MADR for that. I think this argument is also very similar for the `h*` discussion we had on this PR somewhere. Its far better to not use `h*` elements inside reusable components as it makes the level incorrect when used in certain places i.e. it makes them less reusable.

I don't understands how that relates to gaining access to i18n/can in the first place. In some components, we will need to get access to i18n/can so that we can pass its results down to "smaller" components. The latter part is not relevant to this conversation. Only the former is. And the former should be accomplished using imports, not `RouteView` slot properties.

> I've given a good few reasons and advantages with examples and use cases now why we do things like this in this circumstance. I've also explained how this pattern is extremely common, not only in our application but in the wider ecosystem.
> 
> As far as I understand it at the moment, your argument for not using them is that you have a personal preference here to use imports. You say things like "its vastly simpler and more transparent" and "there is no reason to impose restrictions" without backing those statements up with any facts.
> 
> Considering that currently we use heavily use this approach everywhere, I think if you want to change how we access these utilities, then it would be good to put together/document a convincing argument outside of this PR.
> 
> If that argument is convincing enough and you can show its more advantageous then I'd happily change approach. But all that should happen outside of this PR, and I don't think we should change our current approach just in this instance and just in the code added in this PR.

I'm not sure why I need to provide an argument for why using JavaScript imports is a good idea. That's just the best mechanism to get access to things in JavaScript we have. It’s tooling and framework-independent and everyone who knows JavaScript knows how to use imports.

Abstracting away importing things into symbols accessed via scoped slots of Vue components we **happen** to use a lot is unnecessary.

Just because we already use this approach for accessing i18n/can doesn't mean it's good or better. As it stands, it's roughly 50/50 between using `useI18n()` from an import and from RouteView. You may also recall that I was not in favor adding it there in the first place because I didn't see and still don't see the point in doing so.

So my argument boils down to "we really don’t need the extra abstraction" and I'm not sure how I can substantiate that with more examples.

More to the point: I don't want to spent time on the suggestion to not use imports for a situation where they’re the right tool.

I’m resolving this thread again because I don’t want to accept the suggestion.

johncowen: > Just because we already use this approach for accessing i18n/can doesn't mean it's good or better. As it stands, it's roughly 50/50 between using useI18n() from an import and from RouteView.

I just went through all our route components, I counted about 3 or 4 that use `import` all the rest (I dunno 50 odd) use `RouteView::t`.

If you want to change this approach, please can you give some reasoning/facts instead of "I don't like it / I'm not in favour / we don't need to", and I'll quite happily reconsider. Outside of this PR.

At the moment this discussion is getting in the way of us progressing here, whereas if you said "ok I'll put something together and we can discuss separately, if we decide the other way we can change the entire application in one go", then it wouldn't be in the way of us progressing, and importantly we wouldn't be ignoring you concerns. We would be prioritising well.

At the moment we can't get this PR through because we have painted the entire house red, and you want to paint one window frame blue. We can't finish the job here until you agree to use red instead of blue. We can always paint over the red with blue entirely at a later date.

---

Just to point out, we don't access t via an `import` we access its getter/token which we access via an import. Then we have to call the getter. Thats two steps, instead of no steps (I'm using RouteView anyway so there are no extra steps)

```
import { useI18n } from '@/utilities'
const { t } = useI18n()
```

johncowen:
What are we doing about what happens when a service isn't on the page I am viewing?

i.e. the 

`/meshes/default/services/johns-service?page=2` (with `johns-service` being on page 1)

src/app/services/views/ServiceTrayView.vue
kleinfreund: I’m not changing any rule-encoded approach or well-established pattern here. There is no rule that says you can’t use imports for i18n when RouteView is in the component.

The reasons for why you want that aren’t convincing me:

You say it’s what we do everywhere. That’s not true in general and in the case of where we use RouteView it’s down to you preferring this and writing it like this and not accepting when I do it differently.

You mention the Vite HMR issue, but as you said, this shouldn’t influence our decision making in general.

You mention that you have ideas around stripping i18n keys in certain places to avoid the repetitive prefixes, but that’s not something we do today and my PRs shouldn’t be the place where I need to do the pre-emptive refactoring work for something we may or may not implement.

Lastly, you talk about how it’s good DX to make good choices easy, and bad ones hard, etc. Yes, but I just disagree that this applies here. If anything, I think it’s a bad choice to restrict access to i18n in the way you mention you want to. We need i18n in a ton of places and enforcing a rule where we need to always pass i18n values down from a route component sounds like bad DX with hardly any active benefits.

**If** we implement something like i18n key stripping leveraging RouteView, **then** we can discuss whether such a rule makes sense.

**If** we agree that we need to enforce not being able to import i18n in components, **then** we can stop using those imports.

Since I’m not actually changing an established rule here, I treat this as a suggestion and I’m not accepting it because, yes, I _prefer_ using the import style.

johncowen: > When writing components we should "default to" avoiding using `t` within
generic/re-usable components. i.e. we should pass translated text into the
components from the outside using slots as much as possible (and occasionally
attributes). Mostly this means `t` would most commonly be used in route/view
components.

That was agreed upon and approved ^

> **If** we agree that we need to enforce not being able to import i18n in components, then we can stop using those imports.

i think this means you should stop using these imports here.

---

> You mention the Vite HMR issue, but as you said, this shouldn’t influence our decision making in general.

I said "that alone"

kleinfreund: > > When writing components we should "default to" avoiding using `t` within
> > generic/re-usable components. i.e. we should pass translated text into the
> > components from the outside using slots as much as possible (and occasionally
> > attributes). Mostly this means `t` would most commonly be used in route/view
> > components.
> 
> That was agreed upon and approved ^

That rule is about generic components and doesn’t apply here because we’re talking about a route component. It also doesn’t necessitate or imply that we shouldn’t use i18n in components other than route components and generic components (i.e. those that fall somewhere in between those categories).

> > **If** we agree that we need to enforce not being able to import i18n in components, then we can stop using those imports.
> 
> i think this means you should stop using these imports here.

But we didn’t. We agreed that should default to avoiding them in _generic_ components.

lahabana:
I'm sorry I don't have much time to look at what actually we're doing so I'll explain 2 theories:

1. We fetch data in the tray view (there's a Datasource in the tray). then let's still render with a list that doesn't contain the item
2. We don't fetch data then let's redirect to the list at the page without the tray (if possible with a banner explaining what's up). It's not amazing but I believe the only cases where I see this happening is  when there has been changes in the list and we're either sharing the url (likely sharing the detail view url is better) or refreshing the page. I think both these cases are rare enough to sacrifice UX for code simplicity.

WDYT?

johncowen:
> We don't fetch data then let's redirect to the list at the page without the tray (if possible with a banner explaining what's up). It's not amazing but I believe the only cases where I see this happening is when there has been changes in the list and we're either sharing the url (likely sharing the detail view url is better) or refreshing the page. I think both these cases are rare enough to sacrifice UX for code simplicity.

We could only use the DataSource if there's not a `props.data`, i.e.

```
<DataSource
  :src="props.data ? `` : `/meshes/${route.params.mesh}/services/${route.params.service}`"
  v-slot="{ data }"
>
<template v-for="service in [props.data || data]">
{{service.name}}
```

but in saying that, I'm happy to leave this for the moment. I think you are right that its rare enough, and without something like a `DataLoader` we would also have to cope with errors and loading here meaning its simpler without. We can always change later.

---

Ok, so just to be clear from me, lets not do anything here on this point.

kleinfreund:
I agree that we should consider how we handle this (both of Charly’s outlines need additional work anyway) outside of this PR. There might be situations where we need to DataSource within the trays (e.g. when the list data isn’t enough for the summary needs), too.

src/app/services/views/ServiceTrayView.vue
johncowen: Ok we spoke about this offline and said lets move forwards with this as is. I'm super happy to resolve now, gimme 2 tics I should be able to approve also.

johncowen (approved):
Lets get this in!

PR #1613, opened 2023-10-16
feat: adds tray view

kleinfreund:
Contains all the non-module specific changes needed for adding service trays (like for services https://github.com/kumahq/kuma-gui/pull/1611).

## Changes

**chore: cleans up AppView**

Moves the `.app-main-content` styles into AppView (where the `.app-main-content` element is).

**feat: adds TrayView component**

**chore: always styles first AppCollection row item**

**chore: removes explicit name parameters from RouteView method**

Removes the `name` parameter from the `router.push` call in RouteView. They’re currently not necessary and cause issues when using tray routes for list views (specifically, they prevent a cold navigation to a summary route).

**refactor: switches to KSlideout**

## Notes

- Follow up to #1601 and #1604.

### DIFF
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -225,7 +225,8 @@ const click = (e: MouseEvent) => {
 </script>
 
 <style lang="scss" scoped>
-.app-collection :deep(td:first-of-type > a) {
+.app-collection :deep(td:first-child),
+.app-collection :deep(td:first-child > *) {
   color: inherit;
   font-weight: $kui-font-weight-semibold;
   text-decoration: none;
--- a/src/app/application/components/app-view/AppView.vue
+++ b/src/app/application/components/app-view/AppView.vue
@@ -52,7 +52,7 @@
 <script lang="ts" setup>
 import { KongIcon } from '@kong/icons'
 import { KBreadcrumbs, BreadcrumbItem } from '@kong/kongponents'
-import { provide, inject, PropType, watch, ref, onBeforeUnmount } from 'vue'
+import { provide, inject, watch, ref, onBeforeUnmount } from 'vue'
 
 import { useMainView } from '@/components'
 
@@ -61,19 +61,15 @@ type AppView = {
   removeBreadcrumbs: (sym: Symbol) => void
 }
 type Breadcrumbs = Map<Symbol, BreadcrumbItem[]>
+
 const MainView = useMainView()
 
-const props = defineProps({
-  breadcrumbs: {
-    type: Array as PropType<BreadcrumbItem[]>,
-    required: false,
-    default: null,
-  },
-  fullscreen: {
-    type: Boolean,
-    required: false,
-    default: false,
-  },
+const props = withDefaults(defineProps<{
+  breadcrumbs?: BreadcrumbItem[] | null
+  fullscreen?: boolean
+}>(), {
+  breadcrumbs: null,
+  fullscreen: false,
 })
 
 const map: Breadcrumbs = new Map()
@@ -153,6 +149,10 @@ onBeforeUnmount(() => {
 </style>
 
 <style lang="scss" scoped>
+.app-main-content {
+  padding: var(--AppContentPadding);
+}
+
 .actions {
   flex-grow: 1;
   display: flex;
--- a/src/app/common/TrayView.vue
+++ b/src/app/common/TrayView.vue
@@ -8,19 +8,6 @@
     data-testid="tray"
     @close="emit('close')"
   >
-    <template #before-title>
-      <div class="tray-title-wrapper">
-        <slot
-          v-if="$slots.icon"
-          name="icon"
-        />
-
-        <h2 class="tray-title">
-          <slot name="title" />
-        </h2>
-      </div>
-    </template>
-
     <slot />
   </KSlideout>
 </template>
@@ -32,17 +19,6 @@ const emit = defineEmits<{
 </script>
 
 <style lang="scss" scoped>
-
-.tray-title-wrapper {
-  display: flex;
-  align-items: center;
-  gap: $kui-space-30;
-}
-
-.tray-title {
-  margin-top: 0;
-}
-
 .tray {
   // TODO: Remove or replace these once we switch to Kongponents v9 which will deprecate these variables.
   // Overrides KSlideout’s override.
--- a/src/app/common/TrayView.vue
+++ b/src/app/common/TrayView.vue
@@ -7,12 +7,12 @@
     data-testid="tray"
     @close="emit('close')"
   >
-    <template
-      v-if="$slots.icon"
-      #before-title
-    >
+    <template #before-title>
       <div class="tray-title-wrapper">
-        <slot name="icon" />
+        <slot
+          v-if="$slots.icon"
+          name="icon"
+        />
 
         <h2 class="tray-title">
           <slot name="title" />
--- a/src/app/common/TrayView.vue
+++ b/src/app/common/TrayView.vue
@@ -4,7 +4,6 @@
     close-button-alignment="end"
     :has-overlay="false"
     is-visible
-    prevent-close-on-blur
     data-testid="tray"
     @close="emit('close')"
   >
--- a/src/app/common/TrayView.vue
+++ b/src/app/common/TrayView.vue
@@ -38,4 +38,18 @@ const emit = defineEmits<{
   padding-right: $kui-space-80;
   padding-left: $kui-space-80;
 }
+
+// Aligns the position of the close button with the tray card’s content box.
+.tray :deep(.close-button-start),
+.tray :deep(.close-button-end) {
+  margin-top: $kui-space-80 !important;
+}
+
+.tray :deep(.close-button-start) {
+  margin-left: $kui-space-80 !important;
+}
+
+.tray :deep(.close-button-end) {
+  margin-right: $kui-space-80 !important;
+}
 </style>
--- a/src/app/common/TrayView.vue
+++ b/src/app/common/TrayView.vue
@@ -1,72 +1,36 @@
 <template>
-  <div
-    class="tray"
+  <KSlideout
+    close-button-alignment="end"
+    :has-overlay="false"
+    is-visible
+    prevent-close-on-blur
     data-testid="tray"
+    @close="emit('close')"
   >
-    <header class="tray-header">
+    <template
+      v-if="$slots.icon"
+      #before-title
+    >
       <div class="tray-title-wrapper">
-        <span
-          v-if="$slots.icon"
-          class="tray-icon"
-        >
-          <slot name="icon" />
-        </span>
+        <slot name="icon" />
 
         <h2 class="tray-title">
           <slot name="title" />
         </h2>
       </div>
-
-      <KButton
-        appearance="btn-link"
-        class="non-visual-button"
-        data-testid="close-tray-button"
-        @click="emit('close')"
-      >
-        <CloseIcon aria-hidden="true" />
-
-        <span class="visually-hidden">{{ t('common.tray.close') }}</span>
-      </KButton>
-    </header>
+    </template>
 
     <slot />
-  </div>
+  </KSlideout>
 </template>
 
 <script lang="ts" setup>
-import { CloseIcon } from '@kong/icons'
-
-import { useI18n } from '@/utilities'
-
-const { t } = useI18n()
-
 const emit = defineEmits<{
   (event: 'close'): void
 }>()
 </script>
 
 <style lang="scss" scoped>
-.tray {
-  position: fixed;
-  z-index: 100;
-  top: var(--AppHeaderHeight);
-  bottom: 0;
-  right: 0;
-  width: 600px;
-  max-width: calc(100% - var(--AppSidebarWidth));
-  padding: $kui-space-70;
-  border-left: $kui-border-width-10 solid $kui-color-border;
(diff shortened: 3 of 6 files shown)


src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -225,7 +225,8 @@ const click = (e: MouseEvent) => {
 </script>
 
 <style lang="scss" scoped>
-.app-collection :deep(td:first-of-type > a) {
+.app-collection :deep(td:first-child),
+.app-collection :deep(td:first-child > *) {
```
johncowen, line 229: Why this change? I'm not sure if we ever had a table that didn't link somewhere we would want it to use the bold style for the first cell?

johncowen (commented):
I had a few questions:

src/app/application/components/app-collection/AppCollection.vue
kleinfreund: That style is meant to highlight the name (i.e. the resource-identifying feature) first. It shouldn’t be a style specific to links in the first place.

src/app/common/TrayView.vue
```diff
@@ -0,0 +1,80 @@
...
+.tray-title-wrapper {
+  display: flex;
+  align-items: center;
+  gap: $kui-space-30;
+}
+
+.tray-title {
+  margin-top: 0;
+}
+</style>
```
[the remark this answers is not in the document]
kleinfreund: I found a few issues with it. Some aren’t immediately fixable with some tactical overrides:

- Has no `title` slot, only a string prop. This way, we can’t use a link in the title (I want to link to the detail view here because the tray will always hide the table’s context menu while it’s open).
- `props.offsetTop` only accepts a number but we want to pass the string`var(--AppHeaderHeight)`. **Workaround**: Override with CSS.
- The component will always trigger a close action when clicking outside of it. That’s inconvenient for us because we would like to let the user select another row without first closing the tray.
- We can add an icon using the before title slot; however, its styles are awkward (no margin the left). **Workaround**: Add it with CSS.
- Uses a KCard with custom `--KCardPaddingX` and `--KCardPaddingY` values that don’t match the KCard default. **Workaround**: Override with CSS.

I added a commit with the KSlideout variant instead of comparison.

Some of the things above would be easy to supply as features or fixes in Kongponents, but those would likely only become available with Kongponents v9.

src/app/application/components/app-view/AppView.vue
```diff
@@ -1,88 +1,62 @@
...
+          </template>
+        </KAlert>
       </aside>
-      <slot name="default" />
+
+      <slot />
+
+      <template v-if="$slots.tray">
+        <slot name="tray" />
+      </template>
```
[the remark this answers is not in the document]
kleinfreund: I had a different initial strategy with the tray framework, but this isn’t actually needed anymore. I dropped it.

src/app/common/TrayView.vue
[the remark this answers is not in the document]
kleinfreund: I’m sticking to KSlideout and filed PRs for all issues mentioned above (I updated https://github.com/kumahq/kuma-gui/pull/1613#discussion_r1360627713 to list which PRs were filed for which problem).

> Then lastly for the purposes of this PR, do you think we need to wrap the kongponent in an extra wrapper component?

Absolutely. I don’t want to leave the matter of which exact props to set how to each implementation. We can abstract this problem away so I will.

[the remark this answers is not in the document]
johncowen: Ok sounds good, just getting in but let me take another look over this now taking this conversation into account.

src/app/common/TrayView.vue
```diff
@@ -0,0 +1,49 @@
...
+    </template>
+
+    <slot />
+  </KSlideout>
+</template>
+
+<script lang="ts" setup>
+const props = defineProps<{
+  title: string
+}>()
```
johncowen, line 24: Are we ever going to need to set a title that gets put inside a p tag inside the tray? Actually its kinda hard to see in this PR how this component would be used, I'd assumed we were gonna but a h* in the default slot. Actually, didn't we say you were gonna add a "usage commit" on the end here that we could just remove once we were happy?

kleinfreund: I said very explicitly that I would keep https://github.com/kumahq/kuma-gui/pull/1611 open for that purpose and is linked from the PR (very first line) and from the ticket in the board.

johncowen: Oh gotcha, sorry misunderstood.

That PR (1611) still has the old custom tray? Can we get it updated to use the stuff form here then so I can see what usage looks like altogether?

kleinfreund: > Are we ever going to need to set a title that gets put inside a p tag inside the tray?

Yes. I plan to change this in KSlideout put for now we have no control over the container element for `props.title`.

kleinfreund: I did update both PRs now

src/app/common/NavTabs.vue
```diff
@@ -73,3 +73,9 @@ const currentTabHash = computed(() => {
   margin-bottom: var(--AppGap);
 }
 </style>
+
+<style lang="scss">
+.nav-tabs > ul {
+  flex-wrap: wrap;
+}
```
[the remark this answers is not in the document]
kleinfreund: I think that works as well, but in try I found a problem with KTabs (which I’m trying to address in https://github.com/Kong/kongponents/pull/1774).

Once KTabs’ items are inside an overflowing container, that container will get a **vertical** scrollbar on account of a pseudo element extending beyond `.k-tabs`’ box.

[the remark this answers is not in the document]
kleinfreund: Addressing this in https://github.com/kumahq/kuma-gui/pull/1632 and taking changes to NavTabs out of this PR.

src/app/common/TrayView.vue
```diff
@@ -0,0 +1,64 @@
...
+      v-if="$slots.icon"
+      #before-title
+    >
+      <div class="tray-title-wrapper">
+        <slot name="icon" />
+
+        <h2 class="tray-title">
+          <slot name="title" />
+        </h2>
+      </div>
```
johncowen, line 20: Why are we putting the title in the "before-title", I get that we can't use `:title` here but can't we just put it in the default slot? I think generally its also a bad idea to put `h*` tags. With reusable components you never know what the level of headers where you are using the component.

johncowen: Separate to this, why are we only showing the title if there is an icon. Could we just put all this in the default slot please?

src/app/common/TrayView.vue
```diff
@@ -0,0 +1,64 @@
...
+.k-slideout .panel-background-transparent {
+  // Overrides KSlideout’s top offset. `props.offsetTop` doesn’t accept plain CSS values.
+  top: var(--AppHeaderHeight) !important;
+}
+
+.k-slideout-header-content {
+  padding-right: $kui-space-80;
+  padding-left: $kui-space-80;
+}
+</style>
```
johncowen, line 64: Could we use a scoped style node for these overwrites please, and ideally is there a way we can avoid using `!important`?

src/app/common/TrayView.vue
kleinfreund: > Why are we putting the title in the "before-title", I get that we can't use `:title` here but can't we just put it in the default slot?

Because that’s not where the title belongs. The KSlideout title should ideally go into a `title` slot. In the absence of such a slot (and the inadequacy of `props.title`), we put the title in the `before-title` slot. That’s not ideal, but something we’ll be able to move away from with changes to KSlideout.

> I think generally its also a bad idea to put h* tags.

In absence of a more flexible pattern, that’s what we have to do. It’s also what we do virtually everywhere in our application. We could provide the level/tag via prop, but I don’t think that’s something we should consider here as part of this work.

> Separate to this, why are we only showing the title if there is an icon. Could we just put all this in the default slot please?

That’s a mistake which I’ve fixed now. The condition was meant to be on the icon slot.

johncowen: >  It’s also what we do virtually everywhere in our application. 

Not sure how factually correct that is ^, but lets not get into that just yet.

> In absence of a more flexible pattern

Putting an AppView inside the default slot and reusing its header/title is a more flexible pattern. This is exactly the same as we do with KCard everywhere in our application. KCard has a header slot but we never use it.

johncowen: Just to be clear, unless things change what I would like to happen here is that:

1. We should avoid using `h*` headers is non-route components.
2. We should use AppView#title for the header, AppView should go in the default slot. The argument about "keeping things consistent with the design" by using the component is not valid as the font styles for the KSlideout title are dependant on you using the `:title` attribute, not the `before-title` slot. There is no difference in "keeping consistency" with using the before-title slot or the default slot. If anything its worse to use the before-title slot as you have solving something with a hack that is probably unexpected by the author of the component, seeing as in this case we have a very simple alternative, we should use that alternative at least in the short term.

Honestly the more I look at this and the more I look at it and think about it, the less happy I am with the approach taken here. Especially considering how straight forwards the alternative I've been proposing is.

kleinfreund: > Putting an AppView inside the default slot and reusing its header/title is a more flexible pattern.

Not very flexible if it prevents the use of titles in things like title slots, is it?

> 1. We should avoid using `h*` headers in non-route components.

That’s a bad rule. Headings need to appear where they need to appear and if that’s in a non-view component it’s in a non-view component.

> The argument about "keeping things consistent with the design" by using the component is not valid as the font styles for the KSlideout title are dependant on you using the `:title` attribute, not the `before-title` slot.

So you agree that if we had access to a title slot (proposed in https://github.com/Kong/kongponents/pull/1770), it would be a valid argument?

> There is no difference in "keeping consistency" with using the before-title slot or the default slot. If anything its worse to use the before-title slot as you are solving something with a hack that is probably unexpected by the author of the component.

You know for a fact that the only reason I’m using the hacky way of populating the title using the `before-title` slot is because there is no `title` slot.

> Honestly the more I look at this and the more I look at it and think about it, the less happy I am with the approach taken here. Especially considering how straight forwards the alternative I've been proposing is.

I have changed this now to your liking, but I will re-visit using KSlideout’s title mechanisms once they have improved. It’s already gotten more problematic to get the alignment of the close button with the content right.

johncowen: > That’s a bad rule. Headings need to appear where they need to appear and if that’s in a non-view component it’s in a non-view component.

We should use a slot so we can specify the level of the heading from The Outside.

> So you agree that if we had access to a title slot (proposed in https://github.com/Kong/kongponents/pull/1770), it would be a valid argument?

I think I saw that it was suggested that those PRs should be closed because other people are working on the component, my comment there suggests the same.

As to whether its a valid argument. I don't know until I see the redesigned/rebuilt component, it might end up being completely different to what we have now. What I do know is, it will probably use the shared styles from the design tokens, which we use everywhere already for our headers, and I also know that we use KCard with headers in its 'content' slot even though they provide a title slot. I should also note the `:title` is an optional attribute.

Once we have a new KSlideout, we can look to see if we can integrate further with it (something which I'm not against doing). In the meantime we aren't confusing matters by adding the title using the before-title slot, even though we don't get any of the designed styles by doing that.

> You know for a fact that the only reason I’m using the hacky way of populating the title using the before-title slot is because there is no title slot.

Yes I know, I get why you are doing it, thats why I was saying if we can't use ':title' we may as well use the approach to headers we already have i.e. AppView#title, and in not using `before-title` we aren't doing something unexpected to the author of the component.

> I have changed this now to your liking, but I will re-visit using KSlideout’s title mechanisms once they have improved. It’s already gotten more problematic to get the alignment of the close button with the content right.

Awesome thankyou! Much appreciated, let me take another look at this now.

Maybe take a look at my commit again, I took extra care there to make sure the baselines where lined up.

Screenshot 2023-10-20 at 12 39 21

src/app/common/TrayView.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+  <KSlideout
+    class="tray"
+    close-button-alignment="end"
+    :has-overlay="false"
+    is-visible
+    prevent-close-on-blur
```
johncowen, line 7: Did we say somewhere we not prevent close on blur? I think I at least asked the question, not sure what the outcome was. I'm gonna approve this anyway as its one of those things we can change later if need be, 2 tics

johncowen (approved):
I just spotted the prevent close on blur thing, up to you whether we want to change that (or not) here or we can also decide in another PR.

kleinfreund:
> I just spotted the prevent close on blur thing, up to you whether we want to change that (or not) here or we can also decide in another PR.

The main reason why I want to keep this as it is is that we can’t otherwise click rows to select a different item (1 interaction) without first closing the tray (2 interactions).

johncowen:
> The main reason why I want to keep this as it is is that we can’t otherwise click rows to select a different item (1 interaction) without first closing the tray (2 interactions).

I'm pretty sure when I tried it without it didn't matter, I could click the rows and change the tray, plus blur and close the tray. Lets get this one in though, we can always change that later if necessary.

src/app/common/TrayView.vue
kleinfreund: > I'm pretty sure when I tried it without it didn't matter, I could click the rows and change the tray, plus blur and close the tray. Lets get this one in though, we can always change that later if necessary.

Hm, you’re right. I’m not sure what’s different now than when I added this. 🤔 

I removed it now as it behaves like you described.

PR #1632, opened 2023-10-17
chore: allows NavTabs to overflow

kleinfreund:
Allows NavTabs to overflow using a scroll container.

Prevents NavTabs items from wrapping.

### DIFF
--- a/src/app/App.vue
+++ b/src/app/App.vue
@@ -107,7 +107,8 @@ const isWizard = computed(() => route.meta.isWizard === true)
 .app-content-container {
   padding-top: var(--AppHeaderHeight, initial);
   display: var(--AppDisplay);
-  grid-template-columns: var(--AppSidebarWidth) 1fr;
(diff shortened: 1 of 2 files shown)


johncowen:
Where's the best place in the App to see this working? I'm trying to find a place where the tabs would scroll so I can see it working.

kleinfreund:
> Where's the best place in the App to see this working? I'm trying to find a place where the tabs would scroll so I can see it working.

This coming into effect is currently not possible because of a small layout bug that I’m addressing in https://github.com/kumahq/kuma-gui/pull/1613/commits/38f5723adad7449f3f080a455507f057eb102754.

This fix addresses how the main content area overflows: it currently triggers a horizontal scroll bar on the entire content area which is problematic.

With both this PR and that change, you can see this easily on any view that shows the meshes tabs while the viewport is small enough.

johncowen:
Should we move that commit over here? Feels like it belongs in this PR? Rather than the summary view one.

kleinfreund:
> Should we move that commit over here?

I did so now.

johncowen (approved):
👏 nice!

LGTM!

Issue #1582, opened 2023-10-04
Disable any HTTP polling when we hide tabs/lose window focus


johncowen:
Disable any polling when we hide tabs/lose window focus.

More specifically we poll on Zone creation whilst waiting for the zone to be created, but we may also do this in other areas, like maybe onboarding.

Issue #1474, opened 2023-09-12
DataSource improvements


johncowen:
### Description

This one is a bit of an umbrella issue, happy to split up if thats what we want to do.

1. Make helper for `source.ts` files such as in order to return 'EventSources' from the `source.ts` files instead of automatically wrapping promises behind the scenes. Example:
```javascript
(params) => {
  return source(() => {
    return anyOldPromiseReturningClient('/api/url')
  }, {
    options
  })
}
```

Ideally we should still be able to return promises to get a simple 1 time message event.

2. Add `src` or `uri` helper instead of JS templates to use in the DataSource component:
```javascript
<DataSource 
  :src="src(
    '/meshes/:mesh/service/:name',
    {
       mesh: route.params.mesh,
       name: route.params.dataPlane
    }
  )"
>
```

`src` should check that the parameters of the 2nd params object aren't `undefined`

lahabana:
1 has been done 2 needs to happen after: [#1817](https://github.com/kumahq/kuma-gui/issues/1817)

johncowen:
1 here was done a while back, I've split point 2 into a separate issue here https://github.com/kumahq/kuma-gui/issues/2432.

Closing this one.

PR #1634, opened 2023-10-19
chore: Use DataSource for entity scanning (with pausing when the window loses focus)

johncowen:
Replaces the EntityScanner with a DataSource.

Makes all polling/retrying DataSources pause when the tab loses focus.

Whilst doing this I made a helper to let us configure the source of the DataSource easier from the sources.ts file so you can opt in to polling rather than opt out.

Closes https://github.com/kumahq/kuma-gui/issues/1582

Also addresses a part of https://github.com/kumahq/kuma-gui/issues/1474

---

I wanted to note that there will be some follow up work here, but I wanted to submit this PR so we could get this in rather than keep working on related things:

- I 'think' its a good point to PR the `DataLoader` component I spoken about in the past.
- The DataSource file/import organization is a bit messy just down to how DataSource came over here, I'd like to make that a bit less messy.
- Probably PR https://github.com/kumahq/kuma-gui/pull/1525 for real.


src/app/application/services/data-source/index.ts
```diff
@@ -1,36 +1,97 @@
...
-      }
+  const params = {
+    ...{
+      offset: parseInt(queryParams.get('offset') || '0'),
+      size: parseInt(queryParams.get('size') || '0'),
+      page: parseInt(queryParams.get('page') || '0'),
+      search: queryParams.get('search') || '',
+    },
+    ...route.params,
+  }
```
johncowen, line 82: note that every single data source has these available in its `params` argument (both before and after this PR)

src/app/application/services/data-source/index.ts
```diff
@@ -1,36 +1,97 @@
...
+  try {
+    // TODO(jc) Once we remove all the source.closes in the sources.ts files the
+    // second argument here can go
+    const init = route.route(params, { close: () => {} })
```
johncowen, line 86: This kind of TODO stuff will probably happen in relative soon follow up PRs that I mention in the PR description. There are some others elsewhere also but I won't comment there.

src/app/zones/data/index.ts
```diff
@@ -12,25 +22,16 @@ export function getZoneDpServerAuthType(zone: ZoneOverview): string {
   }
   return ''
 }
+// TODO(jc): end
 
-export function getZoneControlPlaneStatus(zoneOverview: ZoneOverview): StatusKeyword | 'disabled' {
+export function getZoneControlPlaneStatus(zoneOverview: ZoneOverview): 'online' | 'offline' | 'disabled' {
+  // TODO(jc): should I check disabled for newly created zones for the purposes of this PR?
```
johncowen, line 28: Ah these ones are a real questions for this PR

- Is it ok to use the config from the first subscription we find here? (i'll address that outside this PR if its not ok)
- Should I check disabled for newly created zones for the purposes of this PR? (this one should be done here if necessary)

johncowen (commented):
Inlines:

features/zones/Create.feature
```diff
@@ -110,15 +111,15 @@ Feature: zones / create
       """
       KUMA_SUBSCRIPTION_COUNT: 1
       """
-    When the URL "/zones+insights/test" responds with
+    Then the URL "/zones+insights/test" responds with
```
kleinfreund, line 114: **Question**: Why is this a `Then` instead of a `When`?

src/app/application/services/data-source/index.ts
```diff
@@ -1,36 +1,97 @@
...
+}
+type RetryingEventSource = CallableEventSource<Configuration>
+type Hideable = EventTarget & { hidden: boolean }
+
+export const getSource = (doc: Hideable) => {
+  return (cb: (source: RetryingEventSource) => Promise<unknown>, config: Configuration = {}) => {
+    let attempts = 0
+    let iterations = 0
+    return new CallableEventSource<Configuration>(async function * (this: RetryingEventSource) {
+      const self = this
```
kleinfreund, line 17: **Question**: Why do we need to assign `this` to a local variable here? It doesn’t look like `self` is accessed in a different function body so this shouldn’t be necessary.

src/app/zones/data/index.ts
```diff
@@ -1,6 +1,16 @@
-import type { ZoneOverview, StatusKeyword } from '@/types/index.d'
+import type { ZoneOverview } from '@/types/index.d'
 import { get } from '@/utilities/get'
 
+// TODO(jc): These two are extremely similar, see if we can merge
+export function getZoneControlPlaneEnvironment(zoneOverview: ZoneOverview): string {
+  // TODO(jc): is it ok to use the config from the first subscription we find here?
```
kleinfreund, line 6: **Comment**: I think we should generally only look at the last subscription, period. That reduces the risk of making false assumptions. The last one is the most recent and should be the only thing we care about.

src/app/zones/data/index.ts
```diff
@@ -12,25 +22,16 @@ export function getZoneDpServerAuthType(zone: ZoneOverview): string {
...
-export function getZoneControlPlaneEnvironment(zoneOverview: ZoneOverview): string {
-  for (const subscription of zoneOverview.zoneInsight?.subscriptions ?? []) {
-    if (subscription.config) {
-      return JSON.parse(subscription.config).environment
-    }
-  }
-  return ''
+export function getStatus(subscriptions: {connectTime?: string, disconnectTime?: string}[] | undefined = []): 'online' | 'offline' {
+  const proxyOnline = subscriptions.length > 0 && [subscriptions[subscriptions.length - 1]].every((item) => item.connectTime?.length && !item.disconnectTime)
+  return proxyOnline ? 'online' : 'offline'
```
kleinfreund, line 36: **Issue**: Why is this implementation being changed?

I think this is significantly harder to parse and understand than the previous code:

```ts
  const subscriptions = zoneOverview.zoneInsight?.subscriptions ?? []
  if (subscriptions.length === 0) {
    return 'offline'
  }

  const lastSubscription = subscriptions[subscriptions.length - 1]
  return lastSubscription.connectTime?.length && !lastSubscription.disconnectTime ? 'online' : 'offline'
```

Especially the `[subscriptions[subscriptions.length - 1]].every((item) => item.connectTime?.length && !item.disconnectTime)` bit is very convoluted. Why would we create an array of definite length 1 to use `Array.prototype.every` on it?

Besides this, why are we separating the online/offline check from the disabled check? Seems to me that all three form a Zone’s status.

src/app/application/services/data-source/index.ts
```diff
@@ -1,36 +1,97 @@
-import CallableEventSource, { isClosed } from './CallableEventSource'
+import CallableEventSource from './CallableEventSource'
 import type { Creator, Destroyer } from './DataSourcePool'
+export type { DataSourceResponse } from './DataSourcePool'
+
+type Configuration = {
+  interval?: number
+  retry?: (e: any) => Promise<any> | undefined
```
kleinfreund, line 7: **Issue**: `e` shouldn’t be typed as `any`. We seem to be always passing it a an instance of `ApiError` so why not use that?

Also, it doesn’t look like we ever intent to return data with the retry promise so it should just be using `void`

```suggestion
  retry?: (e: ApiError) => Promise<void> | undefined
```

Also, why can the retry callback return `undefined`?

src/app/zones/sources.ts
```diff
@@ -14,34 +16,54 @@ type DetailParams = {
...
+      }, {
+        retry: (e) => {
+          if (typeof e.status !== 'undefined') {
+            const status = e.status.toString()
+            switch (true) {
+              case status === '404':
+                // wait 2 seconds and try again
+                return new Promise((resolve) => setTimeout(resolve, 2000))
+            }
+          }
```
kleinfreund, line 64: **Suggestion**: With a better type on the `retry` callback, this should be as simple as that:

```suggestion
          if (e.status === 404) {
            // wait 2 seconds and try again
            return new Promise((resolve) => setTimeout(resolve, 2000))
          }
```

Also, why convert the status code to a string first only to then compare it to a string literal?

Why use a switch statement for a one-branch check that can be a simple if statement instead? It seems very unlikely this would need to be extended to any other status code.

src/app/zones/views/CreateView.vue
```diff
@@ -20,19 +21,15 @@
...
-          appearance="outline"
-          data-testid="exit-button"
-          @click="toggleConfirmModal"
+          @click="() => {
+            if(token === '' || isZoneConnected) {
+              route.replace({name: 'zone-cp-list-view'})
+            } else {
+              toggleConfirmModal()
+            }
+          }"
```
kleinfreund, line 32: **Issue**: Please don’t replace the proper link with a click-triggered route navigation. It’s a deliberate choice to have the thing that behaves like a link be a link.

src/app/zones/views/CreateView.vue
```diff
@@ -349,7 +370,7 @@
         :title="t('zones.form.confirm_modal.title')"
         data-testid="confirm-exit-modal"
         @canceled="toggleConfirmModal"
-        @proceed="router.push({ name: 'zone-cp-list-view' })"
+        @proceed="route.replace({ name: 'zone-cp-list-view' })"
```
kleinfreund, line 373: **Question**: Why is this being changed? Doesn’t seem necessary.

We should use the standard route utilities as much as we can so we don’t accidentally build a whole custom router on top of vue-router.

features/zones/Create.feature
johncowen: I guess I read/wrote it like "when I click the create zone button, then the URL `/zones+insights/test` responds with.

Honestly, I don't mind either way as long as it reads fine. I can change it if you like?

src/app/zones/data/index.ts
johncowen: There are quite a few things here so bear with me while I try separate them a little.

> 1. Issue: Why is this implementation being changed?
> Besides this, why are we separating the online/offline check from the disabled check? Seems to me that all three form a Zone’s status.

Previous to this, the code was using the `getStatusFromInsight` function in the `utilities/dataplane` file.

removal: https://github.com/kumahq/kuma-gui/pull/1634/files#diff-9cd2f8af449e6e0679c6c3ec12e5690e7fdc80a964ac738c7626cd5bc0dc6ac7L388

https://github.com/kumahq/kuma-gui/blob/6d6ac8a18925202d728e0300dab2ef5be55b1e23/src/utilities/dataplane.ts#L71-L74

This code is about zones not dataplanes, so copied the code and moved it to zones. When I did that I saw that there was another `getZoneControlPlaneStatus` function that already existed, but I wasn't totally sure what that is used for / why we were using a different dataplane related function in the first place. I also noticed that I could reuse the code that I'd copied from dataplanes (which was essentially a `subscription` util) so I decided to move/merge what I could that made sense, and then stop and ask questions on the rest, hence in this PR:

> Should I check disabled for newly created zones for the purposes of this PR? (this one should be done here if necessary)

This is not what it was doing previously:

https://github.com/kumahq/kuma-gui/commit/1afcaacf6ba11e2eb3195290a61961c77d97a60f

It sounds like from your response here that we should be checking the `disabled` attribute also which makes sense to me and also means I can just reuse.

Does that sound right to you? i.e. we should check disabled for newly created zones for the purposes of this PR?

src/app/zones/views/CreateView.vue
johncowen: We've spoken about this lots of times before. We are replacing the current route here.

> We should use the standard route utilities as much as we can so we don’t accidentally build a whole custom router on top of vue-router.

We aren't building a whole new custom router on top of vue-router. Where do you think we are doing that? We did used to have completely re-implemented query parameter interactions, which I think you yourself added (?) but we've removed all that now purposefully to use vue-router instead of having our own implementation.

What we do have is a thin layer over `vue-router`s `router.push`. This means we can smooth over rough edges (should we find them) in a centralized place should we need to. This means we don't have to re-implement the same fixes over and over again throughout the application. This means that we don't get in a state where some fixes are applied in one place but not another. Something that we still have quite a lot of, but gradually less due to things approaches like this.

We have already used this thin layer over vue-router to fix the interactions between vue-router and pagination history, I would guess at some point in the future we will use it again.

src/app/zones/data/index.ts
johncowen: From reading the review comments here, I think the answers are:

> Is it ok to use the config from the first subscription we find here? (i'll address that outside this PR if its not ok)

Yes

> Should I check disabled for newly created zones for the purposes of this PR? (this one should be done here if necessary)

Yes

features/zones/Create.feature
kleinfreund: I read this as a test prerequisite and not an assertion so I think "When" was correct, but it’s not a big deal.

src/app/zones/data/index.ts
kleinfreund: I think, yes. A disabled Zone is neither offline or online or at least for offline, the difference is philosophical.

src/app/zones/views/CreateView.vue
kleinfreund: We so far haven’t used RouteView’s `route.replace` for something that we couldn’t have accomplished with an **equivalent** `router.push` call. The implementation of the abstraction is calling `router.push` directly without any changes:

```ts
const routeReplace = (...args: Parameters<typeof router['push']>) => {
  router.push(...args)
}
```

This means there isn’t actually a need for a fix or change of `router.push` behavior and we shouldn’t provide the abstraction in the first place.

Abstractions that actually abstract something away can be good. This here is a case of obfuscating a `router.push` call without there being a need for it.

src/app/zones/sources.ts
johncowen: Unless I'm mistaken `e` can be more than just something with a `.status`?

---

I prefer generally using strings for http error code checks as quite quickly, being able to check for things like `4xx` or `5xx`, becomes helpful. In which case you can just use startsWith.

Similarly, for these types of checks it good to start as you mean to go on. `switch` would be much better here (if) it grows, so starting with `switch` means people that come here to add to this are likely to stay with the existing `switch` instead of using lots of `||`s.

Will we need to add things here? Maybe, maybe not. Is is likely someone will copy this paste it elsewhere and amend? Maybe, maybe not. But if any of those cases are a "yes", then we are starting out as we should go on, using `switch`.

johncowen: 1. What's the technical problem with `switch`?
2. I almost used a different error type/error code for the "zone exists but is offline". It may even be better to do that at some point in the future as this state isn't technically 404. In which case the `if` conditional will get more complicated.

I think overall this is more of a general idea of "use strings/switch for forking via HTTP status codes". Its rather like a state machine for which `switch` is extremely common. I pointed out the "string thing (5xx)" and its usefulness before and considering there could be lots of eventualities, it could end up being a complicated conditional if you aren't careful. Lastly as this is likely to be copy pasted elsewhere, then still I would much rather start as I would like it to go on, using `switch`.

I think the above plus, the fact that we already nearly missed that this would throw more than just an ApiError makes me want to keep this as is i.e. logic based on a `.status: String` using the style of a simple state machine, the status being the `state`.

I think largely this isn't something we need to spend much time talking about or fiddling with.

src/app/zones/views/CreateView.vue
johncowen: The important phrase is "so far".

We've spoken about `route.replace` (its not `router.push`) many many times before. I think the bigger thing is...

We've spoken about using the utilities that `RouterView` provides many many times before. We use them as it gives us centralized points to change/fix/amend things, should we need to, at an application global level.

This is what it is doing. This is the "feature" - being able to change/fix/amend things at an application level. We don't yet, but the "feature" is being able to easily.

Consequently, if we do use it to fix things, it means we don't have various fixes dotted all over the application that (try to) achieve the same thing with completely different code. Something I'm trying to reduce.

Please could we stop discussing not using the utilities that `RouteView` provides. Unless there is an actual technical reason not to, we should use them in all Route views.

features/zones/Create.feature
johncowen: I changed this test to read better overall, I used `Given`/`And`s for the arranging, and `Then`/`And`s for the assertions. I also moved some of the arranging to the start of the test. Will resolve, but feel free to reopen if necessary

src/app/application/services/data-source/index.ts
johncowen: Seems to be due to https://github.com/eslint-community/eslint-plugin-n/blob/HEAD/docs/rules/no-callback-literal.md, but there's also a little more `self` usage elsewhere in related code.

I don't think this one is a biggie to be honest but I kinda don't mind either way. But I definitely don't want to be fiddling with removing eslint node rules from browser code in this PR.

FYI: I don't think we should be using this eslint rule here, but I think we should just leave as is at least for the moment.

kleinfreund: Since this lint rule clearly doesn’t apply to our project, we should rather just disable it and replace `self` with `this` here.

johncowen: Yes most probably, but not now. Feel free to make an issue.

src/app/application/services/data-source/index.ts
johncowen: > Issue: e shouldn’t be typed as any. We seem to be always passing it a an instance of ApiError so why not use that?

I think this has been covered in another comment, lmk if not.

> Also, it doesn’t look like we ever intent to return data with the retry promise so it should just be using void

Yeah maybe, maybe not. I'm still not totally sure what this should return longer term.

> Also, why can the retry callback return undefined?

Because if it doesn't return a promise its like saying "in this case, don't retry". Currently `retry` can return `undefined` (not Promise) specifically for this reason.

---

Overall you can probably see from the todo here (and the PR description) I might change the shape of this slightly, but that will become clearer once we are using this more. For the moment, I wanted to just get the feature PRed as I had already spent a bit of time noodling on exactly how `retry` should look, again that will become clearer if/when we use it more.

Is there anything here that is essential to change? I think at a push the only change I'd consider making here is changing `any` to `void` in the Typescript, but even then I'm not too concerned right now.

lmk

src/app/application/services/data-source/index.ts
johncowen: I'm gonna resolve this one, lmk if you need to reopen.

src/app/zones/data/index.ts
johncowen: I changed the callsite in `source.ts` to use `getZoneControlPlaneStatus` instead of just `getStatus`. This means that creation also checks for the zone to be `enabled`. For the moment I left everything else as is, to be amended in a more refactor-ful PR.

I'd probably like to change the naming of `getZoneControlPlaneStatus` seeing as its in the `zones` module, move `getStatus` to a new `subscriptions` module, plus the amends mentioned in the comments and TODOs here.

src/app/zones/views/CreateView.vue
johncowen: I actually think this should use `history.back`. There are two ways to get to this "modal" looking thing. One is from `/` and one is from `/zones`.

Having an 'Exit' button take me to a completely different page (i.e. if I clicked [Create] from `/`) is a little weird, it should take me to the page that I came from.

lmk

---

(Seeing as this is like a modal, normally the idea would be that I haven't actually left the URL that I'm on and the page at `/` should still be underneath the modal thing, this leads into a much larger discussion though, so for the moment we should probably just change this to use `history.back`)

johncowen: FYI I'm gonna look at moving this to use history.back instead, we can always change it back if we spot a reason why we shouldn't do this.

johncowen: I added `route.back` and used that so it now properly 'Exits' you back to where you originally came from. Resolving this, but feel free to reopen if necessary.

src/app/application/services/data-source/index.ts
johncowen: I changed the `Promise<any>` to `Promise<void>` here, I think we are resolved with this now, but let me know if not!

src/app/zones/data/index.ts
johncowen: I think I can resolve this now, but as mentioned there is a little moving around and probably a bit of renaming and reuse to be done in an upcoming PR. but, let me know if this isn't resolved.

src/app/zones/sources.ts
johncowen: I think I can resolve this, but let me know if not.

src/app/zones/views/CreateView.vue
johncowen: I think I can resolve this under "when a RouteView utility is available we should use it", let me know if there is still anything unresolved here. Resolving.

src/app/zones/views/CreateView.vue
kleinfreund: Eh, I really don’t think this should use `history.back`. The last history entry could be anything including a non-Kuma GUI URL.

Can we move the changes to this element back, please? It doesn’t relate to this PR as far as I can tell.

src/app/application/services/data-source/index.ts
kleinfreund: Can we also change `e` to be `unknown` so we’re at least forced to properly access whether it has properties we’re accessing (preferably via an `instanceof` check).

src/app/zones/data/index.ts
kleinfreund: The initial reason why I opened this thread is not addressed, yet.

The logic for the online/offline check was changed from this:

```ts
  if (subscriptions.length === 0) {
    return 'offline'
  }

  const lastSubscription = subscriptions[subscriptions.length - 1]
  return lastSubscription.connectTime?.length && !lastSubscription.disconnectTime ? 'online' : 'offline'
```

to this:

```ts
  const proxyOnline = subscriptions.length > 0 && [subscriptions[subscriptions.length - 1]].every((item) => item.connectTime?.length && !item.disconnectTime)
  return proxyOnline ? 'online' : 'offline'
```

And I think that makes it unnecessarily hard to understand what the logic actually is. Especially the `[subscriptions[subscriptions.length - 1]].every((item) => item.connectTime?.length && !item.disconnectTime)` is really weird to look at.

Can we please move this back?

src/app/zones/sources.ts
kleinfreund: It’s still a premature optimization if you ask me, but so be it.

src/app/zones/views/CreateView.vue
johncowen: The last history entry is where you came from and therefore where you exit back to.

I'm not sure how many people will be adding links to their websites with a link to our zone creation page. If you come directly to it by entering the URL in your location bar it will default to the zone listing page instead. tbh I think even this case is incredibly rare, so I think catering for people how add links to the zone creation page to their own websites is even rarer, and depending where that link is, history.back is probably still the correct action for an `Exit` button

I'd be happy to change it back to how it was before you said this should be a link (and PR this separately), but I disagree that it needs to be a link and I think more than anything here, in this case I really don't think it matters too much, perhaps arguably if we are worried about accessibility we should make exit take you back to where you were in the most common case, not to potentially a completely different page.

src/app/zones/data/index.ts
johncowen: Oh sorry, well I pretty much copied this code from:

https://github.com/kumahq/kuma-gui/blob/master/src/utilities/dataplane.ts#L72

```javascript
const proxyOnline = (insight.subscriptions ?? []).some((subscription) => subscription.connectTime?.length && !subscription.disconnectTime)
```

I've kept it as similar as possible, I've not really not changed it and i would rather keep it as similar as possible, until I get to do the follow up PR here that I keep talking about.

As I keep saying here, I'll be doing a more general cleanup of this zone data layer file once I get this PR merged. That PR will involve me looking at all the code there and reducing the places where it does the same thing but slightly differently and removing all the places that use slightly different functions.

Honestly considering I've been saying I'll doing all this in a follow I don't think this is something to be concerned about, especially considering I'm pretty much using the same code as it was before this PR.

Finally, I think I prefer the code as a one-liner, not that that matters too much at all.

kleinfreund: If you’re changing this back in a follow up, I’m fine with that so feel free to resolve this thread.

One liners are great if they easily convey what’s happening. Neither `[subscriptions[subscriptions.length - 1]].every((item) => item.connectTime?.length && !item.disconnectTime)` nor `(insight.subscriptions ?? []).some((subscription) => subscription.connectTime?.length && !subscription.disconnectTime)` qualify for this, I’m afraid.

kleinfreund:
I’ve looked over this again and am mostly fine with how this works.

Two things I think need to be addressed as part of this PR (linking to the relevant threads) and not only in a follow up:

- https://github.com/kumahq/kuma-gui/pull/1634#discussion_r1375936564
- https://github.com/kumahq/kuma-gui/pull/1634#discussion_r1375930497

Could you also please rebase this PR?

src/app/application/services/data-source/index.ts
johncowen: I changed `e` to `unknown`, a `instanceof` check won't cut it as we aren't checking for an instance we are checking for existence of a `status` property, it can be an instance of anything as long as it has a `status` property.

FYI: this is likely to change entirely pretty soon.

Resolving this convo, lmk if it needs reopening.

src/app/zones/views/CreateView.vue
johncowen: I went back to two buttons here, resolving, lmk if it needs reopening.

kleinfreund (approved):

src/app/zones/views/CreateView.vue
johncowen: See https://github.com/kumahq/kuma-gui/issues/1706

Issue #1443, opened 2023-09-05
Revisit DPP/Gateway listing in the service detail view


johncowen:
### Description

Continuation of https://github.com/kumahq/kuma-gui/issues/1347

Service detail 'proxy listings' should be able to either show:

1. Only DPPs
2. Only Gateways
3. If the end user has configured things incorrectly then it is possible that we can see both DPPs and Gateways in the listing

Whereas currently we assume these are always sidecars/DPPs.

We should try to optimize for the correct configuration, but also show some sort of warning to the user if the user has configured things wrong (i.e. we get a mix of DPPs and gateways in the response)

Options (not exhaustive)

1. Use two tabs, _but generally we only ever show one tab_, but in the case of user configuration error we show both tabs with some sort of warning. As we won't know which type (DPPs/Gateways) is 'unwanted' we just word this as "its incorrect as there is a mix".
2. Make the listing able to "cope" with both types of proxies, again if we detect a mix we show an error.
3. I think there was some chat on knowing what type of proxies to expect using the service insight endpoint. If we can do this then we could choose which type of listing to show, and also only request those specific proxies (using `gateway=true/false`), alternatively if we continue to request all proxies we would know if there is a user error by detecting if there are gateways in the case of a sidecar type service, or vice versa.

Feel free to edit/add if I've missed anything.

kleinfreund:
Just wanted to make note of the fact that we currently have logic in this view that reads `dataplanesData?.items?.[0]?.dataplane?.networking?.gateway` in order to determine whether we’re looking at gateways or DPPs. If that key is present, we treat _all_ entries as gateways; otherwise, DPPs. That’s something we need to take into account and get right when working on this.

johncowen:
This one has been sat here for a while and there have been quite a few related problems that keep appearing that are related to this. My thinking is we should just make this list in Services cope with both types of proxies, they almost have the same properties anyway.

We can then add a warning as a second step as a "nice to have".

lahabana:
Duplicate of #1717

PR #1641, opened 2023-10-23
fix: service dpp list linking to wrong views

kleinfreund:
**fix(services): check for gateway being inverted**

Fixes the service detail view’s DPP list passing `props.gateways` as `true` when the first item **isn’t** a gateway and as `false` when the first item **is** a gateway. This view is still flawed, but it’s less broken now.

**fix(services): dpp list linking to wrong views**

Fixes the service detail view’s DPP list linking to the wrong views depending. This change makes it so that the links in this list are always correct and independent from `props.gateways`.

Addresses some of the issues related to #1443.

### DIFF
--- a/features/mesh/services/Item.feature
+++ b/features/mesh/services/Item.feature
@@ -42,7 +42,7 @@ Feature: mesh / services / item
               - name: fake-dataplane
                 dataplane:
                   networking:
-                    gateway: ~
+                    gateway: !!js/undefined
         """
       When I visit the "/meshes/default/services/system-1/overview" URL
 
--- a/src/app/services/views/ServiceDataPlaneProxiesView.vue
+++ b/src/app/services/views/ServiceDataPlaneProxiesView.vue
@@ -32,7 +32,7 @@
           :src="`/meshes/${route.params.mesh}/dataplanes/for/${route.params.service}/of/${'all'}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
         >
           <template
-            v-for="gateways in [typeof dataplanesData?.items?.[0]?.dataplane?.networking?.gateway === 'undefined']"
+            v-for="gateways in [dataplanesData?.items?.[0]?.dataplane?.networking?.gateway !== undefined]"
             :key="gateways"
           >
             <KCard>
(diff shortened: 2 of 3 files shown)


features/mesh/services/Item.feature
```diff
@@ -42,7 +42,7 @@ Feature: mesh / services / item
               - name: fake-dataplane
                 dataplane:
                   networking:
-                    gateway: ~
+                    gateway: !!js/undefined
```
johncowen, line 45: Did adding this make the tests fail? i.e. Now you've added this would it catch a similar issue in the future?

johncowen: Awesome thankyou!

src/app/services/views/ServiceDataPlaneProxiesView.vue
```diff
@@ -32,7 +32,7 @@
           :src="`/meshes/${route.params.mesh}/dataplanes/for/${route.params.service}/of/${'all'}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
         >
           <template
-            v-for="gateways in [typeof dataplanesData?.items?.[0]?.dataplane?.networking?.gateway === 'undefined']"
+            v-for="gateways in [dataplanesData?.items?.[0]?.dataplane?.networking?.gateway !== undefined]"
```
johncowen, line 35: Feels like this is one of those PRs where we fix the bug (and add tests to catch in future if appropriate, which you've done) and nothing more, we can then push anything else separately.

Just so I'm clear this line here is the bug fix right? Is the rest (in `DataPlaneList.vue`) necessary to fix the bug?

kleinfreund: The other changes are more important in the sense that they’ll allow the links for the rows to always be correct independent from this check.

Removing this dependency from the value created here makes _this_ fix less important, but still relevant. It’s still not great that we decide the entire table is either gateway or DPP by just looking at the first element, but at least it now influences fewer things.

johncowen: ok gotcha, thanks, 2 tics.

johncowen (approved):
Thanks for spotting!

PR #1646, opened 2023-10-23
feat(data-planes): adds summary tray

### LINKED ISSUE #1637: Add summary tray for DPP/Gateway list views
Add a summary tray to the DPP/Gateway list views

Dependencies:

- https://github.com/kumahq/kuma-gui/pull/1613

kleinfreund:
## Changes

Adds a summary tray to the DPP and Gateway list views and the DPP table in the service detail view.

Resolves #1637.

## Notes

- I decided to move the "Last updated" column into the summary trays (and out of the tables). Let’s discuss this.
- There is an issue with KSlideout: content that overflows the tray area is clipped off because the content area has `overflow-y: auto` which implies that `overflow-x` _also_ computes to `auto` (and cannot be changed to `visible`). This can be observed in service trays that have a StatusBadge with a tooltip: the tooltip content will be only partially visible.
- The arrangement of the summary trays in the DPP listing of a service detail view is a little awkward right now. In order to link to the correct summary tray route name withing DataPlaneList, I needed to add `props.trayRouteName`. I’d like to address this as part of https://github.com/kumahq/kuma-gui/issues/1443 and make sure we have dedicated listings for Gateways and DPPs that aren’t mixed in one component (i.e. how we do it today with DataPlaneList), but I think that would really blow up the scope of this PR, too much.

## Screenshots

image

### DIFF
--- a/features/mesh/dataplanes/Index.feature
+++ b/features/mesh/dataplanes/Index.feature
@@ -22,12 +22,6 @@ Feature: mesh / dataplanes / index
               inbound:
               - tags:
                   kuma.io/protocol: http
-          dataplaneInsight:
-            subscriptions:
-            - status:
-                lastUpdateTime: 2021-02-16T08:33:36.442044+01:00
-            - status:
-                lastUpdateTime: 2021-02-18T08:33:36.442044+01:00
         - name: fake-frontend
       """
 
@@ -39,14 +33,13 @@ Feature: mesh / dataplanes / index
 
     When I visit the "/meshes/default/data-planes" URL
 
-    Then the "$table-header" element exists 9 times
+    Then the "$table-header" element exists 8 times
     And the "$table-header" elements contain
       | Value            |
       | Name             |
       | Service          |
       | Protocol         |
       | Zone             |
-      | Last Updated     |
       | Certificate Info |
       | Status           |
       | Warnings         |
@@ -59,13 +52,12 @@ Feature: mesh / dataplanes / index
 
     When I visit the "/meshes/default/data-planes" URL
 
-    Then the "$table-header" element exists 8 times
+    Then the "$table-header" element exists 7 times
     And the "$table-header" elements contain
       | Value            |
       | Name             |
       | Service          |
       | Protocol         |
-      | Last Updated     |
       | Certificate Info |
       | Status           |
       | Warnings         |
@@ -78,4 +70,3 @@ Feature: mesh / dataplanes / index
       | Value        |
       | fake-backend |
       | http         |
-      | Feb 18, 2021 |
--- a/src/app/data-planes/components/DataPlaneList.vue
+++ b/src/app/data-planes/components/DataPlaneList.vue
@@ -9,35 +9,41 @@
       { label: 'Service', key: 'service' },
       ...(!props.gateways ? [{ label: 'Protocol', key: 'protocol' }] : []),
       ...(isMultiZoneMode ? [{ label: 'Zone', key: 'zone' }] : []),
-      { label: 'Last Updated', key: 'lastUpdated' },
       { label: 'Certificate Info', key: 'certificate' },
       { label: 'Status', key: 'status' },
       { label: 'Warnings', key: 'warnings', hideLabel: true },
-      { label: 'Actions', key: 'actions', hideLabel: true },
+      { label: 'Details', key: 'details', hideLabel: true },
     ]"
     :page-number="props.pageNumber"
     :page-size="props.pageSize"
     :total="props.total"
     :items="props.items ? transformToTableData(props.items) : undefined"
     :error="props.error"
+    :is-selected-row="props.isSelectedRow"
     @change="emit('change', $event)"
   >
     <template #toolbar>
       <slot name="toolbar" />
     </template>
+
     <template #name="{ row: item }">
       <RouterLink
         :to="{
-          name: item.isGateway ? 'gateway-detail-view' : 'data-plane-detail-view',
+          name: props.summaryRouteName,
           params: {
+            mesh: item.mesh,
             dataPlane: item.name,
           },
+          query: {
+            page: props.pageNumber,
+            size: props.pageSize,
+          },
         }"
-        data-testid="detail-view-link"
       >
         {{ item.name }}
       </RouterLink>
     </template>
+
     <template #service="{ rowValue }">
       <RouterLink
         v-if="rowValue.route"
@@ -102,56 +108,33 @@
         {{ t('common.collection.none') }}
       </template>
     </template>
-    <template #certificate="{ row: item }">
-      {{
-        item.dataplaneInsight?.mTLS?.certificateExpirationTime ?
-          formatIsoDate(new Date(item.dataplaneInsight.mTLS.certificateExpirationTime).toUTCString()) :
-          t('data-planes.components.data-plane-list.certificate.none')
-      }}
-    </template>
 
-    <template #actions="{ row: item }">
-      <KDropdownMenu
-        class="actions-dropdown"
-        :kpop-attributes="{ placement: 'bottomEnd', popoverClasses: 'mt-5 more-actions-popover' }"
-        width="150"
+    <template #details="{ row }">
+      <RouterLink
+        class="details-link"
+        data-testid="details-link"
+        :to="{
+          name: row.isGateway ? 'gateway-detail-view' : 'data-plane-detail-view',
+          params: {
+            dataPlane: row.name,
+          },
+        }"
       >
-        <template #default>
-          <KButton
-            class="non-visual-button"
-            appearance="secondary"
-            size="small"
-          >
-            <MoreIcon :size="KUI_ICON_SIZE_30" />
-          </KButton>
-        </template>
-        <template #items>
-          <KDropdownItem
-            :item="{
-              to: {
-                name: item.isGateway ? 'gateway-detail-view' : 'data-plane-detail-view',
-                params: {
-                  dataPlane: item.name,
-                },
-              },
-              label: t('common.collection.actions.view'),
-            }"
-          />
-        </template>
-      </KDropdownMenu>
+        {{ t('common.collection.details_link') }}
+
+        <ArrowRightIcon
+          display="inline-block"
+          decorative
+          :size="KUI_ICON_SIZE_30"
+        />
+      </RouterLink>
     </template>
   </AppCollection>
 </template>
 
 <script lang="ts" setup>
 import { KUI_ICON_SIZE_30 } from '@kong/design-tokens'
-import { MoreIcon } from '@kong/icons'
-import {
-  KDropdownItem,
-  KDropdownMenu,
-  KButton,
-  KTooltip,
-} from '@kong/kongponents'
+import { ArrowRightIcon } from '@kong/icons'
 import { type RouteLocationNamedRaw } from 'vue-router'
 
 import { useCan } from '@/app/application'
@@ -188,8 +171,8 @@ type DataPlaneOverviewTableRow = {
     version_mismatch: boolean
     cert_expired: boolean
   }
-  lastUpdated: string
   isGateway: boolean
+  certificate: string
 }
 
 type ChangeValue = {
@@ -204,9 +187,12 @@ const props = withDefaults(defineProps<{
   items: DataPlaneOverview[] | undefined
   error: Error | undefined
   gateways?: boolean
+  isSelectedRow: ((row: any) => boolean) | null
+  summaryRouteName: string
 }>(), {
   total: 0,
   gateways: false,
+  isSelectedRow: null,
 })
 
 const emit = defineEmits<{
@@ -258,31 +244,28 @@ function transformToTableData(dataPlaneOverviews: DataPlaneOverview[]): DataPlan
 
     const initialData: {
       dpVersion: string | null
-      selectedUpdateTime: number
       version: Version | null
     } = {
       dpVersion: null,
-      selectedUpdateTime: NaN,
       version: null,
     }
 
     const summary = subscriptions.reduce(
       (acc, subscription) => {
-        const lastUpdateDate = Date.parse(subscription.status.lastUpdateTime)
-        if (lastUpdateDate) {
-          if (!acc.selectedUpdateTime || lastUpdateDate > acc.selectedUpdateTime) {
-            acc.selectedUpdateTime = lastUpdateDate
-          }
-        }
-
         return {
           dpVersion: subscription.version?.kumaDp.version || acc.dpVersion,
-          selectedUpdateTime: acc.selectedUpdateTime,
           version: subscription.version || acc.version,
         }
       },
       initialData,
     )
+
+    let certificate
+    if (dataPlaneOverview.dataplaneInsight?.mTLS?.certificateExpirationTime) {
+      certificate = formatIsoDate(dataPlaneOverview.dataplaneInsight.mTLS.certificateExpirationTime)
+    } else {
+      certificate = t('data-planes.components.data-plane-list.certificate.none')
+    }
 
     // assemble the table data
     const item: DataPlaneOverviewTableRow = {
@@ -296,8 +279,8 @@ function transformToTableData(dataPlaneOverviews: DataPlaneOverview[]): DataPlan
         version_mismatch: false,
         cert_expired: false,
       },
-      lastUpdated: summary.selectedUpdateTime ? formatIsoDate(new Date(summary.selectedUpdateTime).toUTCString()) : t('common.collection.none'),
       isGateway: dataPlaneOverview.dataplane?.networking?.gateway !== undefined,
+      certificate,
     }
 
     if (summary.version) {
@@ -330,7 +313,9 @@ function transformToTableData(dataPlaneOverviews: DataPlaneOverview[]): DataPlan
 </script>
 
 <style lang="scss" scoped>
-.actions-dropdown {
-  display: inline-block;
+.details-link {
+  display: inline-flex;
+  align-items: center;
+  gap: $kui-space-20;
 }
 </style>
--- a/src/app/data-planes/components/DataPlaneList.vue
+++ b/src/app/data-planes/components/DataPlaneList.vue
@@ -108,6 +108,13 @@
         {{ t('common.collection.none') }}
       </template>
     </template>
+    <template #certificate="{ row: item }">
+      {{
+        item.dataplaneInsight?.mTLS?.certificateExpirationTime ?
+          formatIsoDate(new Date(item.dataplaneInsight.mTLS.certificateExpirationTime).toUTCString()) :
+          t('data-planes.components.data-plane-list.certificate.none')
+      }}
+    </template>
 
     <template #details="{ row }">
       <RouterLink
@@ -172,7 +179,6 @@ type DataPlaneOverviewTableRow = {
     cert_expired: boolean
   }
   isGateway: boolean
-  certificate: string
 }
 
 type ChangeValue = {
@@ -260,13 +266,6 @@ function transformToTableData(dataPlaneOverviews: DataPlaneOverview[]): DataPlan
       initialData,
     )
 
-    let certificate
-    if (dataPlaneOverview.dataplaneInsight?.mTLS?.certificateExpirationTime) {
-      certificate = formatIsoDate(dataPlaneOverview.dataplaneInsight.mTLS.certificateExpirationTime)
-    } else {
-      certificate = t('data-planes.components.data-plane-list.certificate.none')
-    }
-
     // assemble the table data
     const item: DataPlaneOverviewTableRow = {
       name,
@@ -280,7 +279,6 @@ function transformToTableData(dataPlaneOverviews: DataPlaneOverview[]): DataPlan
         cert_expired: false,
       },
       isGateway: dataPlaneOverview.dataplane?.networking?.gateway !== undefined,
-      certificate,
     }
 
     if (summary.version) {
--- a/src/app/data-planes/components/DataPlaneSummary.vue
+++ b/src/app/data-planes/components/DataPlaneSummary.vue
@@ -1,0 +1,139 @@
+<template>
+  <div class="stack">
+    <DefinitionCard>
+      <template #title>
+        {{ t('http.api.property.status') }}
+      </template>
+
+      <template #body>
+        <div class="status-with-reason">
+          <StatusBadge :status="statusWithReason.status" />
+
+          <KTooltip
+            v-if="statusWithReason.reason.length > 0"
+            :label="statusWithReason.reason.join(', ')"
+            class="reason-tooltip"
+          >
+            <InfoIcon
+              :size="KUI_ICON_SIZE_30"
+              hide-title
+            />
+          </KTooltip>
+        </div>
+      </template>
+    </DefinitionCard>
+
+    <DefinitionCard>
+      <template #title>
+        {{ t('http.api.property.tags') }}
+      </template>
+
+      <template #body>
+        <TagList
+          v-if="dataPlaneTags.length > 0"
+          :tags="dataPlaneTags"
+        />
+
+        <template v-else>
+          {{ t('common.detail.none') }}
+        </template>
+      </template>
+    </DefinitionCard>
+
+    <DefinitionCard>
+      <template #title>
+        {{ t('http.api.property.dependencies') }}
+      </template>
+
+      <template #body>
+        <TagList
+          v-if="dataPlaneVersions !== null"
+          :tags="dataPlaneVersions"
+        />
+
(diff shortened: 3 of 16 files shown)


src/app/data-planes/components/DataPlaneConfig.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <div>
+    <DataSource
+      v-slot="{ data, error }: DataplaneSource"
+      :src="`/meshes/${props.mesh}/dataplanes/${props.dataPlane}`"
+    >
```
lahabana, line 6: I think there was a point about not doing `DataSource`  calls in reusable components. The main reason was that as soon as you start doing this you stop controlling what does requests and you slowly fallback to what we used to do (Pull half the world on startup multiple times without much understanding and deduping)

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -205,15 +198,18 @@ type ChangeValue = {
...
-  total?: number,
-  pageNumber: number,
-  pageSize: number,
-  items: DataPlaneOverview[] | undefined,
-  error: Error | undefined,
+  total?: number
+  pageNumber: number
+  pageSize: number
+  items: DataPlaneOverview[] | undefined
+  error: Error | undefined
```
lahabana, line 205: I very often see these `,` changes between PRs which makes me think that @johncowen and you don't have the same editor config. Can we add a linter so we stop changing these back and forth (linter has the advantage that it will work for future folks too)? This would preserve a clean git history and make PR reviews simpler

src/app/data-planes/views/DataPlaneConfigView.vue
```diff
@@ -4,7 +4,7 @@
     name="data-plane-config-view"
     :params="{
       mesh: '',
-      dataPlane: ''
+      dataPlane: '',
```
lahabana, line 7: Same thing about linter

src/app/data-planes/views/DataPlaneTrayView.vue
```diff
@@ -0,0 +1,80 @@
...
+          <h3>{{ t('data-planes.routes.item.overview') }}</h3>
+
+          <div class="mt-4 stack">
+            <DataPlaneDetails :dataplane-overview="props.data" />
+
+            <DataPlaneDateDetails :dataplane-overview="props.data" />
+          </div>
+        </div>
+
+        <div>
```
lahabana, line 40: I think we should include the YAML here. I'd also would like that in 2.5 we don't use any more data in the tray views than the one we already have from the list view.

We can always make it better later and I'd rather keeping it simple with our time constraints.

src/app/data-planes/components/DataPlaneDetails.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+  <div class="columns">
```
lahabana, line 2: I think the assumption that details on the tray and details in the detail view will be the same is wrong. We should probably have different views.

src/app/gateways/views/GatewayTrayView.vue
```diff
@@ -0,0 +1,80 @@
+<template>
```
lahabana, line 1: This is not something to do with this PR as it existed before but I'd like to understand.

Why do we have different lists, details and trays for Gateway and Regular dataplanes? Aren't they the same in the end? At least from a backend POV there are gateway DPPs and regular DPPs.

Looks like clicking on a detail leads me to:
https://deploy-preview-1646--kuma-gui.netlify.app/gui/meshes/port-2/gateways/firewall-185eca577b-cf1ae.card-gateway-0/overview

it feels like it should lead me to:
https://deploy-preview-1646--kuma-gui.netlify.app/gui/meshes/port-2/data-planes/firewall-185eca577b-cf1ae.card-gateway-0/overview

We may render things differently depending on the type of DPP but I don't think these things should be fundamentally different

src/app/data-planes/components/DataPlaneList.vue
johncowen: Ah yeah this is my bad, would definitely be good to get a lint rule in there. I'd prefer type definitions without commas, but habit/muscle memory got the better of me here it seems.

src/app/data-planes/views/DataPlaneConfigView.vue
johncowen: Just to say I'm good with a lint rule to enforce trailing commas in JS

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: We have linting setup for this, but it doesn’t seem to work for JavaScript expressions that are inside Vue templates. Perhaps that’s an easy fix.

kleinfreund: (And this change is in line with our linter configuration)

src/app/data-planes/views/DataPlaneConfigView.vue
kleinfreund: This will be fixed in https://github.com/kumahq/kuma-gui/pull/1672. Apparently this rule needs to be turned on explicitly because it’s not part of the preset we’re using.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: Oh, I realized that I’ve commented on the wrong thread. My two replies here should’ve been made here https://github.com/kumahq/kuma-gui/pull/1646#discussion_r1371682731:

As for this, this specifically happens inside TypeScript types where a delimiter (both comma and semi-colon are allowed) is always optional. We don’t have a linter configuration for this. Not sure if there is a rule. If there is, we can standardize this .

kleinfreund: This issue will be fixed/standardized here: https://github.com/kumahq/kuma-gui/pull/1673

src/app/data-planes/components/DataPlaneConfig.vue
kleinfreund: That’s a fair point, but it has a price, too: this component, as of this PR, is used 5 times. We’d be duplicating this component’s content more or less exactly in those 5 uses. We have to decide what’s better here.

@johncowen Do you have any opinions here?

src/app/data-planes/views/DataPlaneTrayView.vue
kleinfreund: What data should use here then? Exactly what we have in the list view or a subset (I can make an educated guess in that case).

Also, are you fine with _moving_ both "Last updated" and "Certificate info" into the summary and out of the table?

src/app/gateways/views/GatewayTrayView.vue
kleinfreund: They’re very similar, yes, but in the past we had quite a few bugs related to the kinds of conditionals that differentiate between regular DPPs and gateway DPPs. That is still the case in some places. Most notably the DataPlaneList.vue component.

Now I don’t have super strong opinions on whether we do "share the components for regular and gateway DPPs" or "try and keep the separate components to avoid conditionals", so we could change this (in a different PR preferably).

One more thing that might influence that is whether we turn Gateways into their separate resource in which case changing this makes less sense. But of course that might just be too far into the future.

@johncowen Also here, any opinions? I know you were somewhat keen on making this split also?

src/app/data-planes/components/DataPlaneConfig.vue
johncowen: My opinion is we should _try_ not to use DataSource in re-usable/presentational components at all. We should pass the data down. As a general rule DataSource should only appear in route/view components.

Half the reason why DataSource is made to be a component it's then easier to compose it with other reusable components. Just like an `img`.

Overall, I'd want to look at this PR a little better first though. 🤔 But that being said, isn't there a comment in this PR somewhere about starting out our summary views to only use data from the listing response? If so, that would mean this component would go altogether for the moment right? Meaning we don't have to discuss bit here? Removing the component also means we don't have wrap this code up in a further component. For the moment it can stay in the route view, and as @lahabana says that avoids us putting this DataSource in a 'reusable' component.

Let me know if I misunderstood tho

src/app/data-planes/views/DataPlaneTrayView.vue
johncowen: FWIW, I think we added the certificate info into the rows so it was easy to see multiple at a glance. Moving it out means we can't do that, you'd have to click on each row to see it.

src/app/gateways/views/GatewayTrayView.vue
johncowen: I'm not sure if I follow @lahabana s question entirely, but..

If I follow what @kleinfreund is saying, I think in the case of the DataPlaneList component, it at least used to contain a fair amount of code to do something different depending on whether something was a gateway or not. At the time we had discussed using the same DataPlaneList component for yet another use, so dataplanes, gateways _and_ dataplanes plus gateways. So I suggested not having a reusable dataplane list component at all (moving the reused stuff to the data layer), and just using our Table component in the View for all three. That would also have helped with prevent any issues (both technical and UX) around this area.

If it turns out that these aren't fundamentally different, then maybe we just reuse the route, or maybe a component in different routes or maybe just slightly different usages of out table component in each route. Is it worth even having a gateway tab? Maybe we can just filter a list of DPPs by gateway.

I think first we should figure out what we want to see, which I think is what the related issue is about.

@kleinfreund is the recent bugfix related this in this PR preview?

src/app/data-planes/components/DataPlaneConfig.vue
lahabana: That's my understanding as well

src/app/data-planes/views/DataPlaneTrayView.vue
lahabana: Sorry bad wording. What I mean "what we have in the list view" is "what data is returned by the API from a list view". We can show more info that what is in each rows.

src/app/gateways/views/GatewayTrayView.vue
kleinfreund: I’ve been talking with @lahabana about this again.

To summarize, there are some tentative (non-concrete) plans around this that involve two things:

- We should have one listing (i.e. the Data Plane Proxies tab) that shows all resources for its associated endpoint (i.e. including what we currently split-off into the Gateways tab. They are the same resource and only have minor differences (and importantly: much more in common).
- We should have a listing (i.e. the Gateways tab) that shows MeshGateway resources instead (caveat to solve: those would only be for one of the gateway types; I forget whether delegated or provided)

In other words, definitely not something to do anything about here.

src/app/data-planes/views/DataPlaneTrayView.vue
kleinfreund: Oh, I see. I have definitely misunderstood that then.

In that case I’d like consider the following things to show here (if there is enough space I’d like to show all):

- Status (in list as well)
- Last updated (no longer in list, I don’t think it’s important enough)
- Certificate info (in list as well)
- Tags (not in list, only detail)
- Versions for envoy, kumaDp, etc. (not in list, only detail)

Question: Should we show the YAML? That’s not in the list view API data, so by your rule that’d be a No

src/app/data-planes/components/DataPlaneConfig.vue
kleinfreund: I generally moved all the things making up the tray views into those tray views again.

src/app/gateways/views/GatewayTrayView.vue
lahabana: I don't have problems with different lists as indeed the content is quite different.
What I'm more concerned about is Tray and Details because there's just more commonality in these views than there are differences.

I think the more I'm thinking about it the more "list gateway dataplane" is likely to move away. In favour of just "list gateway services". 

Also there are discussions about having "gateway" just a property of an inbound so a dataplane proxy could be both a gateway and a regular dataplane...

src/app/data-planes/views/DataPlaneTrayView.vue
lahabana: Let's not show the YAML

lahabana: I'm happy with the rest of this info

kleinfreund: I removed the YAML now. The tray views now show the data as outlined above.

features/mesh/dataplanes/Index.feature
```diff
@@ -78,4 +70,3 @@ Feature: mesh / dataplanes / index
       | Value        |
       | fake-backend |
       | http         |
-      | Feb 18, 2021 |
```
johncowen, line 81: Has someone asked for this to be removed from the table? I ask because, if you remember in some of our convos, I'm very much "a use case for tables/summary views is to easily compare" and I can imagine a use case where someone might want to know which dataplane was most recently updated.

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -102,56 +109,33 @@
         {{ t('common.collection.none') }}
       </template>
     </template>
-    <template #certificate="{ row: item }">
-      {{
-        item.dataplaneInsight?.mTLS?.certificateExpirationTime ?
-          formatIsoDate(new Date(item.dataplaneInsight.mTLS.certificateExpirationTime).toUTCString()) :
-          t('data-planes.components.data-plane-list.certificate.none')
-      }}
-    </template>
```
johncowen, line 111: Why are we changing this in this PR?

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -9,35 +9,42 @@
...
+            page: props.pageNumber,
+            size: props.pageSize,
+          },
         }"
         data-testid="detail-view-link"
       >
         {{ item.name }}
       </RouterLink>
     </template>
+
```
johncowen, line 47: I still wish that we didn't have this reused `DataplaneList` component and instead just used AppCollection in their respective views. Looks like we are starting to add more props and complicate things, and moreover then use one approach in one place and a different approach in another:

- Why is there a property to specify the summary view route, but not one to specify the detail view route?
- Why does the detail route link use the `isGateway` property whereas the summary one doesn't?

Is there a reason why we use a different approach to render each link?

---

I also noticed that the data-testid here reads 'detail-view-link'. If it was me I'd probably change this to `data-testid="name"` or something. I also just noticed we've added `detail-link` a bit further down. Kinda not sure without fiddling myself but feels like we should name the things here a little differently/generically for when things change.

Just for clarity, I think we should at least change it so this link to the summary view isn't labelled as linking to the detail view. The larger "lets name things more generically for when they change" thing can wait and be done outside of this PR if that fits better with you.

Also just to note that this last point is very related to another conversation re: generic naming of properties elsewhere.

src/app/gateways/routes.ts
```diff
@@ -35,6 +35,11 @@ export const routes = () => {
             name: 'gateway-clusters-view',
             component: () => import('@/app/data-planes/views/DataPlaneClustersView.vue'),
           },
+          {
+            path: 'config',
+            name: 'gateway-config-view',
+            component: () => import('@/app/gateways/views/GatewayConfigView.vue'),
+          },
```
johncowen, line 42: Is this missing currently? Ideally this would be a separate PR if so.

johncowen (commented):
I left a few comments, not super important but at the very least I think we should make the way we decide on a link in the reused DataplaneList the same (unless theres a reason I've missed why they need to be different)

lmk, in the meantime I'm going to pick up another review on one of these PRs.

features/mesh/dataplanes/Index.feature
kleinfreund: This wasn’t requested. However, I noticed that on small screens, the DPP list in particular is too cramped and I wanted to free up some room. I don’t think "Last updated" is very important and I think it’s the least important column in the table.

Also, with the addition of the "Go to details" entry, the table gains more content, so it becomes doubly relevant.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: This shouldn’t been in the `transformToTableData` function from the start. No idea why it was added inline in the first place.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: > * Why is there a property to specify the summary view route, but not one to specify the detail view route?

Because the summary route is needed (because it cannot be determined purely based on whether the entry is a gateway or regular DPP as there are three relevant summary routes at play here) and the detail route isn’t.

> * Why does the detail route link use the `isGateway` property whereas the summary one doesn't?

See previous answer. It doesn’t need to.



> I also noticed that the data-testid here reads 'detail-view-link'. If it was me I'd probably change this to `data-testid="name"` or something. I also just noticed we've added `detail-link` a bit further down. Kinda not sure without fiddling myself but feels like we should name the things here a little differently/generically for when things change.

It’s pretty much the key information that those are detail links because that informs interaction rules under test.

> Just for clarity, I think we should at least change it so this link to the summary view isn't labelled as linking to the detail view. 

Good point. I removed this attribute because it’s not used anywhere.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: > I think it should stay inline. It was probably done inline because as a general rule for the application we prefer to do these relatively simple things inline.

As long as a table has a `transformToTableDatas` that’s where the data transformations should happen. Not in two places. That’s just confusing.

> I think further down the line it will be these `transformToTableData`s that will be removed. We shouldn't need to perform lots of data reshaping here just to print some JSON data into a table.

This table is a really good example for why that’s not true. We do in fact need to do quite some data transformation to render the tables here.

Once we actually have established data transformation to happen on the data sourcing layer, this becomes a different conversation.

So, please don’t change this `transformToTableData` pattern where in use. It’s very useful. If you want to change this, please provide the data transformations on the data layer and specifically **not** inline.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: > But that link is no longer a link to the detail page, it changed to a link to a summary page i.e. the key information that those are detail links is now incorrect key information.

I’m of course referring to the link that does in fact go to the detail page and not the link that now goes to the summary view on which I have removed the attribute.

features/mesh/dataplanes/Index.feature
johncowen: Hmm, not sure we should probably discuss this more widely, I think at the very least it should be in a separate PR.

Overall, we can remove as many columns as we like, but the dataplane names and service names are pretty likely to be longer than we have and therefore push the content out further than we could ever keep up with with just removing more columns.

I kinda think we should split the rows into two separate lines. The dataplane name being on the top, and other information being on the bottom, but thats a whole other story.

---

Umm whilst this is pretty small it ideally should be in another PR so its clear we've done it and not hidden amongst this very large run of PRs for adding summary views. But that being said I'm gonna resolve this.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: > Would it have been possible to just reuse the same summary route for all three, or at least two? Are the summaries all summaries of a dataplane?

I think that’s possible, yes. I’ve added a commit for this.

While doing so I noticed that `RouteView` should probably not expect a `name` prop but just use `route.name` internally. Out of scope of course.

What I don’t think we can get rid of is the `props.summaryRouteName` from DataplaneList.

src/app/data-planes/components/DataPlaneList.vue
johncowen: 1. Could you tell me why this change is necessary for this PR?

> Once we actually have established data transformation to happen on the data sourcing layer, this becomes a different conversation.

We have, this is why we are iteratively moving everything out of data reshaping in .vue files and either inline or into the data layer. There are many reasons for doing this, one of them being we aren't locking this sort of data reshaping away in a single file so it can't be reused across multiple views. But I also think its interesting that in doing this we are also spotting inconsistencies in the application, as was the case with my recent zones PR.

Putting this sort of stuff in a reusable data layer means we don't have different implementations of the same data reshaping dotted around the codebase in unclear places (something I keep pointing out we need to avoid).

The problem we have here is I'm gradually working towards putting things inline (as we keep saying we should be doing) and moving anything else more complicated to a reusable data later, and you are flip-flopping them back.

---

Ideally we'd leave this as is, it has nothing to do with this PR. But if not could you at least put it in a different PR so we can make progress with this line of work without getting caught up in these sorts of discussions again.

src/app/gateways/routes.ts
kleinfreund: Yes. See https://github.com/kumahq/kuma-gui/pull/1701.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: > The problem we have here is I'm gradually working towards putting things inline (as we keep saying we should be doing) and moving anything else more complicated to a reusable data later, and you are flip-flopping them back.

But we really shouldn’t be moving stuff inline _more_ 😬. We should do that less. **Especially** since you want to move things into the data layer you shouldn’t keep putting things inline when there is an existing function that encapsulates completely the table’s data transformation needs. That just makes it harder to move this away iteratively. Not easier. This provides you with an extremely straightforward thing to look at: is it in the transform function? Well then I’ll likely need to move it into the data source layer. By adding it in the template, you created a second place to remember.

I’m basically cleaning up what shouldn’t have gone in the template in the first place.

johncowen: We need to avoid discussing these things in PRs like this. The PR ideally change things within the scope of the PR and nothing more. Apart from that I'm kinda ok doing little clean up things as we go as long as they are small and not controversial.

This is obviously controversial, and we've had the same discussion many times before.

Please undo this change so we can merge this PR. If you want to discuss it separately make the change in another PR.

That means we can move forwards with this work.

---

Just to point out, whether we should be doing this inline or not, the fact of the matter is, this isn't within the scope of this PR. Please can you remove it and put it in another PR if you think we should change it.

If you have an argument that you think is relevant in moving away from keeping PRs within scope then please bring it up offline. As far as I'm aware keeping PRs within scope is an approach recommended and documented by the rest of the team.

kleinfreund: I undid this change now and will file it as a separate PR.

johncowen: Thankyou. Could you put it back in the same place so its not even in the change set please? I know you don't appreciate people unnecessarily changing the git history. I'd prefer it if there was no change here at all.

kleinfreund: ...

kleinfreund: Done. Unsatisfied with anything else?

src/app/data-planes/components/DataPlaneSummary.vue
```diff
@@ -0,0 +1,139 @@
...
+      </template>
+
+      <template #body>
+        <template v-if="props.dataplaneOverview.dataplaneInsight?.mTLS?.certificateExpirationTime">
+          {{ formatIsoDate(props.dataplaneOverview.dataplaneInsight?.mTLS.certificateExpirationTime) }}
+        </template>
+
+        <template v-else>
+          {{ t('data-planes.routes.item.no_certificate') }}
+        </template>
```
johncowen, line 88: Just to note this seems to be almost exactly the same inline code as the code involved in the discussion here:

https://github.com/kumahq/kuma-gui/pull/1646/files#r1377489524

No action here, just a note, but I don't understand why its ok here, but not elsewhere.

src/app/data-planes/components/DataPlaneList.vue
johncowen: Not right now no, I think we are good now, thank you for doing that. Let me take one last look at this incase, I think I can approve this now.

johncowen (approved):

src/app/data-planes/components/DataPlaneList.vue
johncowen: > While doing so I noticed that RouteView should probably not expect a name prop but just use route.name internally. Out of scope of course.

From what I remember for a good while back now, I think that would give you the name of the route you are on, not the name of the route for the file itself.

kleinfreund: Oh right, yeah that makes perfect sense. It’s kinda annoying we have to "repeat" the route name like that. Easy to mess up and prevents reusability (at least if we intend to use this for something more specific).

PR #1671, opened 2023-10-25
fix(zones): Catch zoneEgressInsight/zoneIngressInsight being undefined

johncowen:
See title

Note: I would like to move this code elsewhere (to the `data` folder). I didn't put it there when I originally did it because it was crossing over with other PRs I think, and I didn't do it here, as it would probably appear like lots of change. So I'm PRing this fix, and I'll move the code later.

### DIFF
--- a/src/app/zones/views/IndexView.vue
+++ b/src/app/zones/views/IndexView.vue
@@ -291,7 +291,7 @@ const getIngresses = (data: {items: ZoneIngressOverview[]}) => {
           offline: [],
         }
       }
-      const subscriptions = item[`${prop}Insight`].subscriptions || []
+      const subscriptions = item[`${prop}Insight`]?.subscriptions || []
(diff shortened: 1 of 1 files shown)


kleinfreund:
**Suggestion**: Can you please expand the types for those resources so that `zoneIngressInsight`, etc. aren’t `any`? This could’ve been caught with better types.

johncowen:
I think the idea was to get this into the code first and then circle back to do the DX follow up (as noted here and elsewhere). 

From what I'm aware the amount of types this thing has cropped up has doubled just in the amount of time this PR has been here. So it would be nice to get this in and come back and move things around as a follow up.

FYI: I almost added a correction to the type definition here, but took it back out as I wanted to check in with other people first. I believe in the past we've sometimes preferred to amend this sort of stuff on the backend.

PR #1675, opened 2023-10-25
feat(meshes): adds summary tray

### LINKED ISSUE #1642: Add summary tray for Meshes list view
Add a summary tray to the Meshes list view

Dependencies:

- https://github.com/kumahq/kuma-gui/pull/1613

kleinfreund:
## Changes

Adds a summary tray to the mesh list view.

Resolves #1642.

## Screenshots

image


src/app/meshes/views/MeshTrayView.vue
```diff
@@ -0,0 +1,98 @@
...
+  display: flex;
+  align-items: baseline;
+  gap: $kui-space-30;
+  // Accounts for the absolutely-positioned close button
+  margin-right: calc($kui-space-30 + 24px);
+}
+
+.tray-title {
+  margin-top: 0;
+}
```
johncowen, line 97: We should centralize this sort of code elsewhere (so the CSS and some of the markup used here), preferably before we roll this all out. But again, I'd be fine coming back to sort also if necessary.

Centralizing this will prevent us from having to remember to add these things just in these sorts of routes, and if we need to change them we can do so in one place.

src/locales/en-us/common/index.yaml
```diff
@@ -45,14 +45,15 @@ common:
...
-    close: 'Close'
   emptyState:
     title: 'No data'
     message: 'There are no {type} present'
     icon: 'stateNoData'
   collection:
     none: ''
+    tray:
+      empty_title: 'Couldn’t find the selected item'
+      empty_message: 'The selected item either doesn’t exist or isn’t on the current page.'
```
johncowen, line 56: Feel like we could use `type` here instead of 'item', similar to the code in the `emptyState` above, again small detail, doesn't necessarily need to happen here although it would probably help with the rollout if we did it here.

src/app/meshes/views/MeshListView.vue
```diff
@@ -98,13 +109,38 @@
...
+                  query: {
+                    page: route.params.page,
+                    size: route.params.size,
+                  },
+                })"
+              >
+                <component
+                  :is="child.Component"
+                  :name="route.params.mesh"
+                  :mesh-insight="data?.items.find((item) => item.name === route.params.mesh)"
```
johncowen, line 140: If we just called this `data`/`props.data` it would probably help a little with the rollout. We mostly use `props.data` for other places where we do this sort of thing also.

johncowen (commented):
I left a few comments, lmk

src/app/meshes/views/MeshListView.vue
kleinfreund: `props.data` is worse. I want to know what stuff I’m dealing with and I can by calling it `props.meshInsight`. I want to go back to all non-general examples of this and changes `props.data` there, too because it doesn’t help anything to have that general of a name while it does help to have a more specific name referring to the entity/resource.

features/mesh/MeshTray.feature
```diff
@@ -0,0 +1,43 @@
+Feature: Mesh tray
```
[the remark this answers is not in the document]
kleinfreund: I’ve renamed this everywhere.

src/app/meshes/views/MeshTrayView.vue
kleinfreund: I’ll touch up on this later. Having more uses will provide better constraints for where these styles can and cannot be and for now the duplication is fine. Easy to change later.

src/locales/en-us/common/index.yaml
kleinfreund: I’ve changed this now.

kleinfreund:
I’ve also added the "Go to details" element we discussed to both mesh and service list views in https://github.com/kumahq/kuma-gui/pull/1675/commits/fc2d4d18654e37bdacce7faae217ac22e9853cdd.

I’ll be pulling this kind of redundant AppCollection code out into AppCollection as it should be standardized and not repeated.

features/mesh/MeshTray.feature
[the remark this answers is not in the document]
johncowen: thanks!

src/app/meshes/views/MeshTrayView.vue
johncowen: Sounds good to me!

src/locales/en-us/common/index.yaml
johncowen: 👍

src/app/meshes/views/MeshListView.vue
johncowen: So this will be one of those places where we have completely opposing opinions.

Lets leave this as is for the moment. But I think even in this instance I can see plenty of problems that tell me using `props.data` is better not worse.

Briefly:

- In the case of passing these property down into a dynamic component you don't actually know whether its a MeshTrayView or maybe its something else. It just so happens that in this case it can only be a `MeshTrayView` but in the case of something like a Dataplane/Gateway list (so a list of mixed entities) we wouldn't know, therefore `data` works.
- In the case where the data used in the list can change, say for example at one point we use `meshes` for the listing and then we decided to change it to use `mesh-insights` for the listing (or vice versa) then we have to change all the property names also.

To your point of knowing what stuff is, well you are in the `MeshSummaryView`, so thats a mesh thing. Additionally You have Typescript to tell you what things are, in fact show you want things are, so you don't need to repeat that in the name of the thing.

So to be clear, lets leave this as is, but in my opinion using a `props.data` is definitely not worse, and potentially factually (see above) its not worse either.

johncowen:
> I’ve also added the "Go to details" element we discussed to both mesh and service list views in https://github.com/kumahq/kuma-gui/commit/fc2d4d18654e37bdacce7faae217ac22e9853cdd.

When did we discuss this bit? Or was that with someone else? I think I missed that, and personally I'm not super sure I prefer this.

---

I think overall this is generally good to go though, lmk on the above question and we can see if we can get all these PRs merged this morning/today.

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -175,7 +175,7 @@ const lastPageNumber = ref(props.pageNumber)
 const lastPageSize = ref(props.pageSize)
 
 const columnWidth = computed(() => {
-  const specialColumns = props.headers.filter((header) => ['warnings', 'actions'].includes(header.key))
+  const specialColumns = props.headers.filter((header) => ['details', 'warnings', 'actions'].includes(header.key))
```
johncowen, line 178: I think you mentioned this elsewhere but I also would like to avoid these kinda of special things in a generic component.

johncowen (approved):
I left a couple more comments so share thoughts but nothing blocking.

LGTM!

features/mesh/MeshSummary.feature
```diff
@@ -0,0 +1,43 @@
...
+    Then the "$summary" element doesn't exist
+
+  Scenario: Summary URL goes to page with open summary
+    Given the environment
+      """
+      KUMA_MESH_COUNT: 51
+      """
+
+    When I visit the "/meshes/mesh-1?page=2&size=50" URL
+    Then the "$summary" element exists
```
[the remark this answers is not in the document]
kleinfreund: Seeing as this can break in one view easily without breaking in any other, these more or less duplicated tests are quite valuable in my opinion.

src/app/application/components/app-collection/AppCollection.vue
kleinfreund: It does have its caveats, but it’s better to standardize this in one place than to rely on all users of AppCollection to apply the exact same column sizing styles.

We also wanted to have the behavior that columns, by default, are spread out equally. Together with the desirable behavior of the special columns having a smaller, mostly fixed width, we need to specify columns widths for all columns which brings the need for this kind of logic.

PR #1676, opened 2023-10-25
chore: Adds data-test-id restricted eslint rule

johncowen:
Avoids the mistaken `data-test-id` instead of `data-testid` which I've almost done on more than one occasion.


adamdehaven (approved):
lgtm @kleinfreund

kleinfreund (approved):

PR #1685, opened 2023-10-30
refactor: switches insight endpoints to _overview

### LINKED ISSUE #1683: Use _overview instead +insights for DP, ZoneProxies
### Description

We've added these endpoints to replace the existing one. They support more advanced filtering and live on the endpoints resources which means that they are auto-generated and managed.

The output format hasn't changed, just the URL so it should be straightforward to fix

kleinfreund:
## Changes

Resolves #1683.

## Notes

- Note that I needed to re-arrange some entries in fs.ts so that, for example, `/zones/_overview.ts` matches before `/zones/_.ts`.
- Note that for ZoneIngressOverviews, the resource path segment is in fact `/zone-ingresses` and not `/zoneingresses`. That is **not** the case for ZoneEgressOverviews.


src/services/kuma-api/KumaApi.ts
```diff
@@ -142,15 +142,15 @@ export default class KumaApi extends Api {
...
   }
 
   getAllDataplaneOverviewsFromMesh({ mesh }: { mesh: string }, params?: DataPlaneOverviewParameters): Promise<PaginatedApiListResponse<DataPlaneOverview>> {
-    return this.client.get(`/meshes/${mesh}/dataplanes+insights`, { params })
+    return this.client.get(`/meshes/${mesh}/dataplanes/_overview`, { params })
   }
 
   getDataplaneOverviewFromMesh({ mesh, name }: { mesh: string, name: string }, params?: any): Promise<DataPlaneOverview> {
-    return this.client.get(`/meshes/${mesh}/dataplanes+insights/${name}`, { params })
+    return this.client.get(`/meshes/${mesh}/dataplanes/${name}/_overview`, { params })
```
kleinfreund, line 153: @lahabana Can I ask you to review the changes specifically in this file (i.e. KumaApi.ts)?

I specifically want to double-check that the path change for Zone Ingresses is correct (i.e. the addition of the dash). I did run these APIs locally and I think all of the above are right. Just making sure.

johncowen (approved):
A 👍 from me but I see you had a q on here which was almost the q was gonna ask as well (i.e. did we double check the API with these changes).

Sounds like you did so as soon as you get your answer I'm fine with this.

kleinfreund:
> A 👍 from me but I see you had a q on here which was almost the q was gonna ask as well (i.e. did we double check the API with these changes).
> 
> Sounds like you did so as soon as you get your answer I'm fine with this.

I’m near certain those changes are fine. I checked related PRs and the actual API in a local Kuma environment. Just want to get a second pair of eyes as this has a potentially bigger impact if it goes wrong (cc @lahabana see https://github.com/kumahq/kuma-gui/pull/1685#discussion_r1376145555).

lahabana (approved):
Yeah that's right!

lahabana:
(Might be worth checking the auto-merge PR before deployment to Konnect though to be 100% sure :) )

PR #1700, opened 2023-10-31
chore: removes unused transition

kleinfreund:
Removes the transition on App.vue’s router view. The linter for https://eslint.vuejs.org/rules/require-toggle-inside-transition.html started to go off and that made me wonder if we even really use this transition. I’m not sure what effect it actually has on our pages. Does it have any?

### DIFF
--- a/src/app/App.vue
+++ b/src/app/App.vue
@@ -37,23 +37,13 @@
         </template>
 
         <AppView>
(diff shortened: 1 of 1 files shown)


johncowen:
Also see https://github.com/kumahq/kuma-gui/pull/1687

johncowen (approved):

PR #1701, opened 2023-10-31
fix(gateways): config tab missing

kleinfreund:


johncowen (commented):
I just noticed there is already pretty much a test for these tabs, just it missing this one I think. If it is that straight forwards could you add that in also? I think its just another line in the test (features/mesh/gateways/PageNavigation.feature)

johncowen (approved):
awesome, thanks!

PR #1703, opened 2023-10-31
refactor(services): moves summary content into view

kleinfreund:
Moves the summary content into the view so we’re able to control the layouting mechanisms more precisely. Generally, summary views are different from detail views anyway.


kleinfreund:
> Does this add/correct anything feature wise, or is it just a refactor?

We’ve seen scenarios where the the layout of the summary entered a horizontal list mode rather than being vertically stacked. When I’ve looked at this here, it always looked correct even when using `columns`.

Put another way: this allows us to use `stack` instead of `columns`.

johncowen (approved):
Ah gotcha, thanks. Very interesting.

PR #1709, opened 2023-11-01
chore: disable summary trays for some views

kleinfreund:
Disables the summary trays for Zone CPs, Meshes, and Services.


johncowen (approved):
Code looks fine to me.

The app feels much better like this and I don't think we are losing anything (at least currently) by losing the summaries in these areas.

PR #1712, opened 2023-11-02
fix(zones): wrong egress detail link

kleinfreund:


johncowen (approved):
I think maybe we should add a test for clicking on these rows and back again. I'm unsure whether to do this test in every single one of our Index.feature tests or make a 'application' tests similar to Titles.feature that loops through and does the same one test everywhere. wdyt?

In the meantime this is good to go, lets get it in a follow up with a test PR soonish.

kleinfreund:
@johncowen Yeah, I’ll add a test for this. Not totally sure what’s better.

PR #1715, opened 2023-11-02
chore: add detail link in list views test

kleinfreund:


johncowen (commented):
Thanks for sorting! (just a quick xref https://github.com/kumahq/kuma-gui/pull/1712)

At first glance this kinda looked fine, but then I thought a few things:

- Do we need the Collection column (and related selectors)? We only ever have one collection on these pages, and I would imagine it would be very unlikely we would have two collections on these pages. I think we can also then move the CSS selector to use an alias rather than inlining that one, which will help with test readability.
- Should we check the URLs rather than the breadcrumbs? We are checking that the links work here so maybe we should assert the URLs, not the human text breadcrumbs (which are separately brittle)? I think we can assert part of the URL so that we don't have to mock out a the correct name for everything (but maybe I'm wrong there?). Kind of an aside but currently I think the test is a little confusing to read as the breadcrumb we are asserting on is the part that corresponds to the page you were on before clicking. If we can assert the URL of the detail page, then it will read better.

There might be something I've missed here than mean that the above don't apply, lmk if so!

kleinfreund:
> * Do we need the Collection column (and related selectors)? We only ever have one collection on these pages, and I would imagine it would be very unlikely we would have two collections on these pages. I think we can also then move the CSS selector to use an alias rather than inlining that one, which will help with test readability.

I restructured this a bit. Not completely happy with the brittle data-testid selector, but it’s "fine" for now.

> * Should we check the URLs rather than the breadcrumbs? We are checking that the links work here so maybe we should assert the URLs, not the human text breadcrumbs (which are separately brittle)? I think we can assert part of the URL so that we don't have to mock out a the correct name for everything (but maybe I'm wrong there?).

Yeah, to have a better URL subset match, we would need to mock out the list responses for the first row’s name which I kinda want to avoid in a "global" test like this. Otherwise, the test is probably better placed in the individual list view tests.

I think asserting on the breadcrumb texts is fine. We also assert on titles and that’s kind of useful. This also shouldn’t be the kind of text that’s subject to frequent change.

> Kind of an aside but currently I think the test is a little confusing to read as the breadcrumb we are asserting on is the part that corresponds to the page you were on before clicking. If we can assert the URL of the detail page, then it will read better.

I agree and I did that mainly because it allows me to not mock out the names and it happens to also test the right thing here.

Maybe all things considered, this sort of test should be module-specific instead.

johncowen:
What do you think of this?

https://github.com/kumahq/kuma-gui/compare/master...johncowen:kuma-gui:chore/add-detail-link-in-list-views-test-pt2

It avoids knowing the slug of the entity, but still does a pretty good job of asserting we are going to the correct URL, plus avoids us asserting the human text in the breadcrumbs, something we want to avoid if possible.

We are asserting the URLs are correct, which is what the issue was. There are some edge cases where exactly how it does this could potentially break but its highly unlikely and I think same could be said of the breadcrumb approach, so I think this is just as good a test.

Aside from all this, we could consider adding a `URL matches` step if this start to become a common type of assertion (just to be clear don't need to do this here/now)

kleinfreund:
This has some gaps because there can be an unaccounted for segment between URL and DetailURL:

If I’m on `/zones` and I accidentally created a link to the Zone Egress detail view instead of Zone detail view, the test will still pass because both `/zones` and `/overview` are in the URL, but the complete segment is `/zones/bandwidth-0/egresses/terminal-0/overview` (i.e. the wrong view).

Let me think about this some more.

kleinfreund:
We could use something like this:

```
    Examples:
      | URL                          | DetailViewSelector                       |
      | /zones                       | [data-testid='zone-cp-detail-view']      |
      | /zones/bandwidth-0/egresses  | [data-testid='zone-egress-detail-view']  |
      | /zones/bandwidth-0/ingresses | [data-testid='zone-ingress-detail-view'] |
      | /meshes                      | [data-testid='mesh-detail-view']         |
      | /meshes/default/gateways     | [data-testid='gateway-detail-view']      |
      | /meshes/default/data-planes  | [data-testid='data-plane-detail-view']   |
      | /meshes/default/services     | [data-testid='service-detail-view']      |
      | /meshes/default/policies     | [data-testid='policy-detail-view']       |
```

This is quite robust and direct. Only problem is that this doesn’t work for gateways because we re-use the DPP view component (i.e. its route data-testid attribute is a DPP one too)

johncowen:
> This has some gaps because there can be an unaccounted for segment between URL and DetailURL:

> If I’m on /zones and I accidentally created a link to the Zone Egress detail view instead of Zone detail view, the test will still pass because both /zones and /overview are in the URL, but the complete segment is /zones/bandwidth-0/egresses/terminal-0/overview (i.e. the wrong view).

Yeah gotcha, I think this is what I meant by 

> There are some edge cases where exactly how it does this could potentially break but its highly unlikely and I think same could be said of the breadcrumb approach, so I think this is just as good a test.

I think the zone breadcrumb title can exists on all those pages.

I think a `URL matches` type thing would be good, so we can mark bits we aren't interested in and we solve this little problem.

---

🤔 lets go with your suggestion, I also just thought that URLs are also brittle (but very unlikely to be in our app), and the RouteView names are static so if we ever change them it would make sense to also change the test. Yeah I like that more the more I look at it

> This is quite robust and direct. Only problem is that this doesn’t work for gateways because we re-use the DPP view component (i.e. its route data-testid attribute is a DPP one too)

Could we pull just that one out into a separate non-Outline scenario and also test for the word gateway in the URL? Would that work? (probably comment in there too to explain why its the weird one)

lmk otherwise I can approve this as is.

kleinfreund:
I’ve separated that out. Hopefully we can remove that soon again.

johncowen (approved):
Looks good!

PR #1716, opened 2023-11-02
chore(dataplanes): expand list view coverage

kleinfreund:

### DIFF
--- a/features/mesh/dataplanes/Index.feature
+++ b/features/mesh/dataplanes/Index.feature
@@ -1,11 +1,13 @@
 Feature: mesh / dataplanes / index
   Background:
     Given the CSS selectors
-      | Alias           | Selector                              |
-      | table           | [data-testid='data-plane-collection'] |
-      | table-header    | $table th                             |
-      | table-row       | $table tbody tr                       |
-      | dataplane-title | [data-testid='data-plane-collection'] |
+      | Alias         | Selector                              |
+      | table         | [data-testid='data-plane-collection'] |
+      | table-header  | $table th                             |
+      | item          | $table tbody tr                       |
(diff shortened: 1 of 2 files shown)


features/mesh/dataplanes/Index.feature
```diff
@@ -68,9 +79,52 @@ Feature: mesh / dataplanes / index
   Scenario: The Proxy listing has the expected content and UI elements
     When I visit the "/meshes/default/data-planes" URL
 
-    Then the "$table-row" element exists 9 times
-    Then the "$table-row:nth-child(1)" element contains
+    Then the "$table tbody tr" element exists 9 times
+    Then the "$item" element contains
```
johncowen, line 83: bit of a nit but I'd like to avoid CSS selectors inside the tests where possible and use aliases (our aliases are almost pageObject-y and also/therefore read better)

I think this test was one of those early ones but it looks like in other tests we have

"$item" = tbody tr

```gherkin
Then the "$item" element exists 9 times
Then the "$item:nth-child(1)" element contains
```

Then if I need to I can do:
```gherkin
Then the "$item:nth-child(3)" element contains...
Then the "$item:nth-child(7)" element contains...
Then the "$item:nth-child(100)" element contains
```

..if I ever need to.

Would be good to follow that pattern and change up where we don't as we go, unless there is a reason not to.

johncowen (approved):
LGTM! Thanks for expanding this! I left a non-blocking suggestion/nit

features/mesh/dataplanes/Index.feature
kleinfreund: I’ve changed this.

PR #1719, opened 2023-11-02
fix(services): showing config tab for gateway services

### LINKED ISSUE #1718: /config tries to fetch an externalService when the service is gateway_*
### What happened?

when going to see the yaml it sends request to: `/meshes/default/external-services?tag=kuma.io%2Fservice%3Amy-kong-kong-manager_kong_svc_8002`

service:
```json
{
 "type": "ServiceInsight",
 "mesh": "default",
 "name": "my-kong-kong-manager_kong_svc_8002",
 "creationTime": "2023-11-02T09:09:04.038898Z",
 "modificationTime": "2023-11-02T10:17:54.119392Z",
 "status": "online",
 "dataplanes": {
  "total": 1,
  "online": 1
 },
 "issuedBackends": {
  "ca-1": 1
 },
 "serviceType": "gateway_delegated"
}
```

It shouldn't show the yaml tab I believe.

Same thing happens with provided_gateway

kleinfreund:
Resolves #1718.

### DIFF
--- a/features/mesh/services/Item.feature
+++ b/features/mesh/services/Item.feature
@@ -2,13 +2,14 @@ Feature: mesh / services / item
   Background:
     Given the CSS selectors
       | Alias                  | Selector                                         |
+      | config-tab             | #service-config-view-tab a                       |
       | data-plane-proxies-tab | #service-data-plane-proxies-view-tab a           |
       | item                   | [data-testid='data-plane-collection'] tbody tr   |
       | input-search           | [data-testid='k-filter-bar-filter-input']        |
       | button-search          | [data-testid='k-filter-bar-submit-query-button'] |
       | button-clear-search    | [data-testid="k-filter-bar-clear-query-button"]  |
 
-  Scenario Outline: Shows Data Plane Proxies for service type <ServiceType>
+  Scenario Outline: Shows correct tabs for service type <ServiceType>
     Given the URL "/meshes/default/service-insights/firewall-1" responds with
       """
         body:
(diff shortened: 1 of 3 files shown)


features/mesh/services/Item.feature
```diff
@@ -2,29 +2,31 @@ Feature: mesh / services / item
...
 
     Examples:
-      | ServiceType       | ExistsAssertion |
-      | ~                 | exists          |
-      | internal          | exists          |
-      | gateway_builtin   | exists          |
-      | gateway_delegated | exists          |
-      | external          | doesn't exist   |
+      | ServiceType       | ConfigTabExistsAssertion | DppTabExistsAssertion |
+      | ~                 | doesn't exist            | exists                |
```
johncowen, line 25: Is this a place where we need to change this to use `js/undefined` now we have that?

features/mesh/services/Item.feature
```diff
@@ -2,29 +2,31 @@ Feature: mesh / services / item
...
+      | internal          | doesn't exist            | exists                |
+      | gateway_builtin   | doesn't exist            | exists                |
+      | gateway_delegated | doesn't exist            | exists                |
+      | external          | exists                   | doesn't exist         |
```
johncowen, line 29: I'd really rather avoid these types of "put the assertion in the table" if possible, they take extra to figure out what is being tested, although I can see why you did it here, it wouldn't be great doing a repeated test for each of these cases.

🤔 

What about if we had 2 ScenarioOutlines? If I'm not mistaken there are two groups here? A group where the config tab exists and the dataplane tab doesn't, and another group where its the opposite. Plus there is only one outlier, otherwise they'd all be the same.

Bit of a nit this, but half the reason for the Gherkin tests is to try and write them in a way which makes them super easy to see whats being tested. All in all not super necessary to change here/now, but up to you.

src/test-support/FakeKuma.ts
```diff
@@ -39,8 +39,8 @@ export class KumaModule {
       [
         'internal',
         'external',
-        // 'gateway_delegated',
-        // 'gateway_builtin',
+        'gateway_delegated',
+        'gateway_builtin',
```
johncowen, line 43: I'm trying to remember why I commented these out originally (i.e. I didn't just delete as I was unsure). I think I wasn't sure if there were places where you would guarantee that you wouldn't get the gateway types, in which case we might need to add an arg to specify what you want.

Again bit of a nit so not a blocker here, unless it affects other tests of course, maybe we just need to double check where we already use this, maybe you've done that already, lmk

johncowen (commented):
Fix LGTM! Couple of comments/questions just incase. Lmk and we can probably get this is pretty soon.

features/mesh/services/Item.feature
kleinfreund: Yes. Changed this.

features/mesh/services/Item.feature
kleinfreund: Fair enough. I changed this.

src/test-support/FakeKuma.ts
kleinfreund: It seemed to work fine with our current tests so I think it’s better to put them in. Maybe it exposes similar issues earlier.

johncowen: Ok yeah no worries, I think I half-remembered, maybe there is something we extra we should do/could do here, but we can do it later and I don't think its super important anyway, more of a dev nicety, will resolve.

johncowen (approved):
👍

Issue #1415, opened 2023-08-30
Improve list view filters


kleinfreund:
### Description

Improve the list view filters

- Better error state handling
- Making it more clear how to search for tags (the general syntax is `tag: 'kuma.io/service: service-1'`)
- search DPP by IP

lahabana:
Might be worth following: https://kong-aip.netlify.app/aip/160/

johncowen:
Summary of sync discussion:

As we can 'guess' information based on whether a search term has a `:` in it or not (due to DNS) we can say:

1. If the search term _doesn't_ have a `:` in it, then it is a `name` search.
2. If the search term has a `:` in it then we either:
    - Apply any "special" rules we have hardcoded into the frontend, a.k.a. "shortcut expansion" such as `zone:` > `kuma.io/zone:`
    - Assume that the search term is looking for a tag i.e. `version: 1` (which would be a "shortcut" for `tag: "version: 1"`)
3. If a user wants to specifically search for the tag `zone` (not as a shortcut for `kuma.io/zone:`) they would have to use `tag: "zone: zone.name"` specifically.

---

How this would work in practice:

- `name-of-dpp` would do a search based on name
- `version: 1` would do a search based on the tag `version` equalling `1`
- `kuma.io/zone: zone-name` would do a search based on the tag `kuma.io/zone` equalling `zone-name`
- `zone: zone-name` would be a "shortcut" that would expand to `kuma.io/zone` and use the same `kuma.io/zone: zone-name` tag search as the explained in the previous point.
- `status: degraded` _currently_ would search for a tag called `status` equalling `degraded`
- `status: degraded` _in a possible future_  could search for a dpp with the status _field_ as `degraded`. If you wanted to instead search by a `status` tag you would have to use the long form `tag: "status:degraded"`

---

We also discussed but discarded:

For "shortcut expansion" we could have used both the original tag shortcut and the expanded shortcut, i.e.

- `zone: zone-name` could perform a search for anything with the tags `kuma.io/zone: zone-name` _or_ `zone: zone-name`
- `status: degraded` could perform a search for anything with the tags `status: degraded` _or_ an entity with the field `status` equalling `degraded`

kleinfreund:
> * Apply any "special" rules we have hardcoded into the frontend, a.k.a. "shortcut expansion" such as `zone:` > `kuma.io/zone:`

[...]

PR #1721, opened 2023-11-02
[WIP] refactor: Revisit FilterBar

johncowen:
WIP, Do Not Merge, and all of those caveats 😅 . I've put this up for review as I'd like some help figuring this piece of work out a little more, so I'm not so much looking for code review more like a functionality review.

Please note, importantly, I've removed code here that might be re-added again.

---

I started to re-organize our existing filter bar so its easier to work with so I can change its functionality slightly to work like the discussion here https://github.com/kumahq/kuma-gui/issues/1415#issuecomment-1784978626

Notes:

1. I moved everything I could find to do with the FilterBar into its own component folder so I could isolate the code I need to look at/work with into one place.
2. I deleted a some of this afterwards. I might bring some of this back depending on what I'm missing here.
3. I pulled out some code that didn't immediately look like it was needed, or could be replaced by native functionality. Mostly by using native `form` key events and native `input[type=search]` events and theming. I might need to bring some of this back if the native stuff can't cover what we need.
4. I changed the component to be called FilterBar instead of KFilterBar so we don't use a namespace that doesn't belong to us. I guess the idea was that this would maybe go into kongponents. If we ever wanted to do that, we could add the namespace if/when its moved.

---

The biggest question I have here, to help me continue on this is:

**What functionality have I removed here (if anything) that we need to keep?**

I might have more questions as we go.

Lastly, please only use the dataplane listing for checking as thats where I'm working, I would guess the rest doesn't work at all.


kleinfreund:
Differences to the current implementation:

- Issue: Applying the suggestions via the keyboard (Enter) doesn’t work (submits search instead)
- Issue: Clear button is no longer focusable
- Issue: A query with `name:` expands to a `tag` query (query `name:test`; now `tag=name%3Atest`; before `name=test`)
- Possible issue: Quotation marks are kept and don’t escape spaces (query `name:"test value"`; now `tag=name%3A%22test` and `name=value%22`; before `name=test+value`). Tags never containing spaces is incidental and it could well be we have scenarios where this will become necessary.

johncowen:
Awesome thankyou!

Lemme have a look/think about those

johncowen:
> - Issue: Clear button is no longer focusable

I think this is expected and native behaviour. With this PR hitting Escape clears the search field for you. Anyone used to using this functionality on other websites would expect the same functionality here (you cannot tab to the clear button on other websites). Looking at the state previous to this PR, hitting Escape when the search field is focussed doesn't do anything, so I 'think' this is more of a 'fix'

> - Issue: A query with name: expands to a tag query (query name:test; now tag=name%3Atest; before name=test)

I think this is how it should work i.e. I'm looking for a tag called `name` that equals `test`. Typing the name without the `name:` would search for the title/name/identifier of the thing. I think this is one of the things that jumped out at me when watching someone use this, they didn't use the `name: my-name` form but instead just `my-name`. Alternatively we could turn `name:` into one of our shortcuts, but personally I don't think we should because:

1. Most people probably would just use `my-name` without `name: `
2. `name: my-service` isn't a shortcut as such, its longer. Its easier to just type the name of the thing.

> - Possible issue: Quotation marks are kept and don’t escape spaces (query name:"test value"; now tag=name%3A%22test and name=value%22; before name=test+value). Tags never containing spaces is incidental and it could well be we have scenarios where this will become necessary.

Quotation marks are no longer "special", so if you type them then you are searching for them, therefore they should be kept. From your explanation it sounds like spaces are still encoded but just using `%22` instead of `+`?

> Tags never containing spaces is incidental and it could well be we have scenarios where this will become necessary.

It sounds like we don't have scenarios currently where this is necessary? (unless you know of one) in which case, if/when it becomes necessary we should add it then.

> Issue: Applying the suggestions via the keyboard (Enter) doesn’t work (submits search instead)

Gonna take a look at this ^ one, let me know your thoughts on the others above.

kleinfreund:
> I think this is expected and native behaviour.

One, the native behavior is worse as interactive elements must be focusable and that’s not the case here.

> With this PR hitting Escape clears the search field for you. Anyone used to using this functionality on other websites would expect the same functionality here (you cannot tab to the clear button on other websites). Looking at the state previous to this PR, hitting Escape when the search field is focussed doesn't do anything, so I 'think' this is more of a 'fix'

Two, adding a mechanism to clear on Escape to the existing implementation is possible while adding focusability to the native element is not.

> I think this is how it should work i.e. I'm looking for a tag called `name` that equals `test`. Typing the name without the `name:` would search for the title/name/identifier of the thing. I think this is one of the things that jumped out at me when watching someone use this, they didn't use the `name: my-name` form but instead just `my-name`. Alternatively we could turn `name:` into one of our shortcuts, but personally I don't think we should because:
> 
> 1. Most people probably would just use `my-name` without `name: `
> 2. `name: my-service` isn't a shortcut as such, its longer. Its easier to just type the name of the thing.

I strongly disagree here. It’s fine for the free-form `test` to query a model’s `name` field by default. People defaulting to querying for names with the free-form doesn’t inform how `name: test` should behave. However, we shouldn’t expand _any_ field that exists on the queried model to a tag query because that’s confusing.

The rule should be unambiguous: If a field exists on the model for a given query key, you can query it using `$field: $value`. If such a field doesn’t exist, expand to a tag search. Additionally, free-form queries translate to a `name: $value` query.

> From your explanation it sounds like spaces are still encoded but just using `%22` instead of `+`?

No, that’s not correct. The query `name: "test value"` results in `tag=name%3A%22test&name=value%22` which are two query parameters `tag=name%3A%22test` and `name=value%22`. The part after `test` is basically interpreted as a free-form name query.

> It sounds like we don't have scenarios currently where this is necessary? (unless you know of one) in which case, if/when it becomes necessary we should add it then.

In this case, I don’t agree with this. We should try and build a mostly application-agnostic component here so that future requirement changes don’t cause a need for us to rewrite big chunks of its logic. Persisting spaces in queried value seems like a very sensible precaution to take now rather than later. Especially because we already did take care of that and are now looking to _remove_ this functionality.

More so it doesn’t seem like there is a need to query for values containing quotation marks. I’d argue that’s far, far, far less likely to happen in the future than querying for values containing spaces.

johncowen:
> One, the native behavior is worse as interactive elements must be focusable and that’s not the case here.

Not following native behaviour is confusing for users who might depend on these features. Native behaviour dictates that when you are focussed on a search input, hitting tab should take you away from the search field entirely, including the clear icon. Therefore changing this would _remove_ native behaviour.

Users who depend on these native features would also expect the Escape key to clear the field, which I would guess is why browser vendors decided that the clear icon isn't tabbable as people using keyboard for navigation have no need for it because they would use the Escape key instead.

> No, that’s not correct. The query name: "test value" results in tag=name%3A%22test&name=value%22 which are two query parameters tag=name%3A%22test and name=value%22. The part after test is basically interpreted as a free-form name query.

Oh I see what you are saying. Ok well I think this falls into "none of the things we search on ever have spaces in the name, therefore we use the space to delimit search queries" and I'd say is therefore expected here. Although I think I do need to look at only sending the last name query. Let me look at that some more.

> We should try and build a mostly application-agnostic component here so that future requirement changes don’t cause a need for us to rewrite big chunks of its logic.

None of the code that deals with this is in the component itself (there might be a tiny bit left that I haven't got to yet). So the component _is_ more application agnostic.

If we need to change the logic, which I think is currently pretty unlikely for our application (see below) then the code is currently about 10 lines of code that we would have to change:

https://github.com/kumahq/kuma-gui/pull/1721/files#diff-8c51094d188779df11ec5a418503cf5a77c088bb12581335955952e7deef4c3dR66-R79

I think the only thing I can think of where we might want to have spaces in searches would be if we wanted to search on a description, and I think the usecases there are pretty small. Really we should discourage users depending on searching on terms in descriptions given we have labels and similar specifically for searching for things.

> More so it doesn’t seem like there is a need to query for values containing quotation marks

I completely agree, there is no code here to deal with this, it just happens because we don't try to deal with it.

---

Ok I'll try and distill the discussion into yes/no points/options:

1. Should tab/Escape behaviour of the search input use/follow the native functionality of the browser? I would say yes we should use native browser behaviour, and you would say no we shouldn't.
2. Should we add a `name: ` shortcut for searching on name. I would say no its unnecessary, and you would say yes we should include it.
3. Should we cope with a search term that includes spaces now. I would say no its unnecessary, and you would say yes we should cope with spaces (and I guess quotes and quotes in quotes etc etc.)

I think those summarise the points here. Let me know if I've missed anything.

(just to be clear I haven't yet looked at the Enter functionality when the focus is on the panel, but from what I understand I agree that an Enter on an option in the panel should do the same as clicking an option in the panel)

johncowen:
>> No, that’s not correct. The query name: "test value" results in tag=name%3A%22test&name=value%22 which are two query parameters tag=name%3A%22test and name=value%22. The part after test is basically interpreted as a free-form name query.

> Oh I see what you are saying. Ok well I think this falls into "none of the things we search on ever have spaces in the name, therefore we use the space to delimit search queries" and I'd say is therefore expected here. Although I think I do need to look at only sending the last name query. Let me look at that some more.

I checked this and it looks like I slightly misunderstood what was happening here. I thought I'd remembered making sure we only send `name=` once even if I use `my-name my-service` (i.e. we would only send name=my-service). TLDR; I don't believe a change is required to what is already here due to this.

But! It does still fall into the questions of whether we should 2. Use a `name: ` shortcut, and 3. Cope with spaces/quotes.

So to sum up with this comment. If we address each of the 3 questions either yes or no, then this will either be left as is, or code will need adding if my original assumptions/requirements need changing.

johncowen:
Hey @kleinfreund just incase I'm just waiting to see if the above 3 points are a good summary to capture what we have to decide here, lmk

kleinfreund:
I think it’s a rather dubious claim that people would expect the lack of focusability on a clear button if one is presented. At best, I’d file that under browser bug. It’s definitely not a feature or a fix. It also seems like that’s a non-standard behavior as `input[type=search]` neither has the clear button nor the Escape to clear functionality in Firefox. We’d be better off supplying users with a consistent interaction in all browsers in this case.

---

As for the `name:` question, my take is that the following interaction rules are generally easier to anticipate and learn:

If a model has a field matching a filter query key, using `field: value` allows you to filter for it. If that isn’t the case, but a filter query key is one of the well-known tag symbols then expand the query to a tag filter query behind the scenes.

That’s how filtering for any field other than `name` would work so I don’t see why it shouldn’t _also_ work for `name`.

---

As for spaces in queries, I guess it just comes down to a business decision: are we happy with removing this functionality because we think it’s unlikely any user of FilterBar will ever need it?

---

To summarize, can you please get answers to the following questions:

- In the presence of a field on a model (e.g. `name`), should `field: value` type queries expand to a tag query `tag=field:value` (e.g. `tag=name:test`) or a regular field query `field=value` (e.g. `name=test`)?
- Are we happy with removing the handling of spaces in query values (they are replaced with a `+` character, e.g. `name=test+value`)?

lahabana:
A MADR is pending on this

johncowen:
Closing this for the moment

PR #1722, opened 2023-11-02
fix(gateways): treating delegated gateways as builtin gateways

### LINKED ISSUE #1720: For provided gateway policies tab should request the same as regular DPP
### What happened?

Currently it does the same as builtin gateway. It should be the same as regular DPPs (policies + rules and output of the first is `SidecarDataplane`.

kleinfreund:
Changes the gateway detail view’s policy tab to treat delegated gateways the same as regular Data Plane Proxies (i.e. the same as the DPP detail view’s policy tab).

Resolves #1720.

### DIFF
--- a/src/app/gateways/views/GatewayPoliciesView.vue
+++ b/src/app/gateways/views/GatewayPoliciesView.vue
@@ -19,33 +19,76 @@
 
       <KCard>
         <template #body>
-          <DataSource
-            v-slot="{ data: policyTypesData, error: policyTypesError }: PolicyTypeCollectionSource"
-            :src="`/*/policy-types`"
-          >
+          <template v-if="props.data.dataplane?.networking?.gateway?.type === 'BUILTIN'">
             <DataSource
-              v-slot="{ data, error }: MeshGatewayDataplaneSource"
-              :src="`/meshes/${route.params.mesh}/gateways/${route.params.dataPlane}/policies`"
+              v-slot="{ data: policyTypesData, error: policyTypesError }: PolicyTypeCollectionSource"
+              :src="`/*/policy-types`"
             >
-              <ErrorBlock
-                v-if="policyTypesError"
-                :error="policyTypesError"
-              />
+              <DataSource
+                v-slot="{ data: gatewayDataplane, error }: MeshGatewayDataplaneSource"
+                :src="`/meshes/${route.params.mesh}/gateways/${route.params.dataPlane}/policies`"
+              >
+                <ErrorBlock
+                  v-if="policyTypesError"
+                  :error="policyTypesError"
+                />
 
-              <ErrorBlock
-                v-else-if="error"
-                :error="error"
-              />
+                <ErrorBlock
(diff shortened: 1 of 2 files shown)


src/types/index.d.ts
```diff
@@ -208,7 +208,7 @@ export type DataPlaneNetworking = {
   }[]
   gateway?: {
     tags: Record<string, string>
-    type?: 'builtin' | 'delegated' | undefined
+    type?: 'BUILTIN'
```
lahabana, line 211: I think technically 'DELEGATED' is a type that exists

kleinfreund: I’ve updated this and added a comment:

```ts
/**
 * The presence of the `gateway` field means one of two things:
 *
 * - The resource represents a builtin gateway (when `gateway.type === 'BUILTIN'`)
 * - The resource represents a delegated gateway (when `gateway.type === 'DELEGATED'` or `gateway.type` is omitted)
 *
 * If the `gateway` field is absent, the resource represents a regular Data Plane Proxy.
 */
gateway?: {
  tags: Record<string, string>
  type?: 'BUILTIN' | 'DELEGATED'
}
```

src/app/gateways/views/GatewayPoliciesView.vue
```diff
@@ -19,33 +19,76 @@
...
+
+                  <ErrorBlock
+                    v-else-if="error"
+                    :error="error"
+                  />
+
+                  <ErrorBlock
+                    v-else-if="rulesError"
+                    :error="rulesError"
+                  />
```
johncowen, line 78: Little nit, I 'think' or-ing would work here also 

```html
<ErrorBlock
  v-if="policyTypesError || error || rulesError"
  :error="policyTypesError || error || rulesError"
/>
```

As long as the `v-if/:error` are in the same order the correct error would be passed to the `ErrorBlock` I think.

FYI: I have a first draft `DataLoader` sitting here almost ready to PR, which hopefully will help with all this type of stuff everywhere in the application.

---

I just saw theres another one further up where we use 2 ErrorBlocks also.

These are a non-blocker btw, up to you if you prefer it (and I'm correct that it would be fine)

johncowen (approved):
LGTM! We should probably test this out also, either a follow up here or in a separate is fine by me.

src/app/gateways/views/GatewayPoliciesView.vue
kleinfreund: This is actually necessary because those two variants are not equivalent. The type system doesn’t know that after a check `(T1 | undefined) || (T2 | undefined)` it’s safe to assign to a value expecting `T`.

kleinfreund:
> We should probably test this out also, either a follow up here or in a separate is fine by me.

Yep. But I think we should talk through how we want these views to evolve. From a past conversation with Charly, it seems we might actually go back to the DPP list to handle all types of DataplaneOverview and turn the Gateway list view into something more dedicated.

PR #1725, opened 2023-11-06
chore(e2e): split input and element value assertion commands

kleinfreund:
It looks like it can happen that with the following structure:

```ts
  $(selector).then(($el) => {
    const chainer = $el[0].tagName === 'INPUT' ? 'have.value' : 'contain'
    cy.wrap($el).should(chainer, value)
  })
```

... interactions like browser navigation can happen between getting the element with `$(selector)` and asserting against it inside the `then` callback. This can break tests.

There probably are other solutions to this, but this is the simplest thing I could think of for now.

### DIFF
--- a/cypress/support/step_definitions/index.ts
+++ b/cypress/support/step_definitions/index.ts
@@ -190,11 +190,10 @@ Then(/^the "(.*)" element(s)? contain[s]?$/, (selector: string, multiple = '', t
   }
 })
 Then(/^the "(.*)" element contains "(.*)"$/, (selector: string, value: string) => {
-  $(selector).contains(value)
-})
-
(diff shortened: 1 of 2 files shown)


johncowen (approved):
We chatted a bit offline about this:

"Reselecting" seems to be the thing recommended after performing any actions, whilst there is no action being performed here, we are querying twice, and theoretically the page might not have moved on at the time of the first query, and at the time of the second one, it could have.

Its really unlikely we would use the same selector across two pages that refer to a div on one page and an inout element on another page, and even if we were to do that, I think this assertion would fail anyway, due to either trying to assert a `value` on a div, or trying to assert a `contain` on an input.

All in all I think this feels fine to me, and it means we can continue to abstract the `element contains` for any sort of element from the tests themselves.

PR #1726, opened 2023-11-06
refactor: move filter bar and related code into its own subfolder

johncowen:
Prelude to https://github.com/kumahq/kuma-gui/pull/1721

Moves KFilter related files into a `filter-bar/` folder with only very minimal renaming changes.


src/utilities/tokenizeFieldFilterQuery.spec.ts
```diff
@@ -1,152 +0,0 @@
...
-      'tag:"kuma.io/zone: europe" tag:"version:v1 namePrefix:back',
-      ['tag', 'namePrefix'],
-      new Error('Quote mismatch for field “tag”.'),
-    ],
-  ])('throws expected errors', (query, allowedFields, expectedErrorMessage) => {
-    const call = () => tokenizeFieldFilterQuery(query, allowedFields)
-
-    expect(call).toThrowError(expectedErrorMessage)
-  })
-})
```
kleinfreund, line 152: **Question**: Why is this test being removed here? This should happen in the other PR if it’s no longer necessary/useful.

johncowen: Ah sorry missed the file addition, its here now. It's entirely removed in the other PR also from what I remember

kleinfreund (approved):

PR #1728, opened 2023-11-06
feat(dataplanes): integrates Gateways in Data Plane Proxy list view

### LINKED ISSUE #1407: Data Plane Proxies show only first service/zone from tags
### What happened?

When we show tags for Data Plane Proxies (both list and detail views), we only read the first service and zone from the annotations even though there could be multiple.

How do we want to handle this?

### LINKED ISSUE #1717: Dataplane and gateway simplification
### Description

What we do with gateways and dpp is confusing. It's causing questions in the code and for users as well.

Here's what I suggest:

- Get rid of: /meshes/default/gateways
- Have all dpps irrespective of types in: `/meshes/default/data-planes` remove the column `protocol` and just have `type` similar to what we do in `/meshes/default/services`.
- The tray for a dpp should show each inbound independently which will make it easy to make gateway and dpp render clearly differently
- Then later we can expand each service type view differently depending on the type of service.

I believe this would simplify a lot of things as only the service detail view is really different depending on the service type.

probably superseeds: https://github.com/kumahq/kuma-gui/issues/1443

kleinfreund:
## Changes

**chore(mocks): update mocks to handle gateway types properly**

Updates the dataplane mocks to handle gateway types more properly. The list views now generate names indicating which type of gateway is used so we can properly respond to this in other mocks (e.g. for Dataplane policies, we should only show the gateway policy list for builtin gateways).

**feat(dataplanes): integrates Gateways in Data Plane Proxy list view**

Integrates the Gateway list view in the Data Plane Proxy list view. Adds a type column and the appropriate type filter to the DPP list view and the service detail view’s DPP list.

Removes the protocol and zone columns.

Resolves #1717.

Resolves #1407.

## Notes

- For the type column, how should we display the values:
  - For the two gateway types, should we keep them uppercase (i.e. "BUILTIN" and "DELEGATED")?
  - For regular DPPs, should we show "STANDARD" or omit the value or something else?


src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -23,10 +21,54 @@
...
     <template #toolbar>
-      <slot name="toolbar" />
+      <FilterBar
+        class="data-plane-proxy-filter"
+        :placeholder="`tag: 'kuma.io/service: backend'`"
+        :query="props.query"
+        :fields="{
+          name: { description: 'filter by name or parts of a name' },
+          protocol: { description: 'filter by “kuma.io/protocol” value' },
+          service: { description: 'filter by “kuma.io/service” value' },
```
johncowen, line 31: I would guess that if you are viewing this from the ServiceDetail page then we don't want the `service` shortcut? (as its already being filtered by service)

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -23,10 +21,54 @@
...
+            label: 'Standard',
+            value: 'standard',
+          },
+          {
+            label: 'Builtin',
+            value: 'builtin',
+          },
+          {
+            label: 'Delegated',
+            value: 'delegated',
```
johncowen, line 56: In regards to your question on `STANDARD/BUILTIN/DELEGATED`, whatever is decided there we should also use here.

src/test-support/mocks/src/dataplanes/_overview.ts
```diff
@@ -1,41 +1,64 @@
...
+
 export default ({ fake, pager, env }: EndpointDependencies): MockResponder => (req) => {
+  const query = req.url.searchParams
+  const _gateway = query.get('gateway') ?? ''
+  const _name = query.get('name') ?? ''
+  const _tags = query.get('tags') ?? ''
+
+  const inbounds = parseInt(env('KUMA_DATAPLANEINBOUND_COUNT', `${fake.number.int({ min: 1, max: 3 })}`))
   const subscriptionCount = parseInt(env('KUMA_SUBSCRIPTION_COUNT', `${fake.number.int({ min: 1, max: 10 })}`))
+  const isMtlsEnbaled = env('KUMA_MTLS_ENABLED', fake.helpers.arrayElement(['false', 'true'])) === 'true'
```
johncowen, line 11: nit: Little typo here, in both places, but still a typo

src/test-support/mocks/src/meshes/_/dataplanes/_/_overview.ts
```diff
@@ -1,52 +1,41 @@
 import type { EndpointDependencies, MockResponder } from '@/test-support'
 export default ({ env, fake }: EndpointDependencies): MockResponder => (req) => {
   const { mesh, name } = req.params
+
+  const inbounds = parseInt(env('KUMA_DATAPLANEINBOUND_COUNT', `${fake.number.int({ min: 1, max: 3 })}`))
   const subscriptionCount = parseInt(env('KUMA_SUBSCRIPTION_COUNT', `${fake.number.int({ min: 1, max: 10 })}`))
+  const isMtlsEnbaled = env('KUMA_MTLS_ENABLED', fake.helpers.arrayElement(['false', 'true'])) === 'true'
```
johncowen, line 7: Just commenting here so you don't miss this typo either

src/test-support/mocks/src/meshes/_/dataplanes/_overview.ts
```diff
@@ -1,87 +1,64 @@
...
-  const params = req.params
+  const { mesh } = req.params
   const query = req.url.searchParams
   const _gateway = query.get('gateway') ?? ''
   const _name = query.get('name') ?? ''
   const _tags = query.get('tags') ?? ''
 
+  const inbounds = parseInt(env('KUMA_DATAPLANEINBOUND_COUNT', `${fake.number.int({ min: 1, max: 3 })}`))
   const subscriptionCount = parseInt(env('KUMA_SUBSCRIPTION_COUNT', `${fake.number.int({ min: 1, max: 10 })}`))
+  const isMtlsEnbaled = env('KUMA_MTLS_ENABLED', fake.helpers.arrayElement(['false', 'true'])) === 'true'
```
johncowen, line 12: one more

src/app/data-planes/components/DataPlaneList.vue
johncowen: The more I think about it the more I think we should keep the toolbar slot and just pass slightly differently configured FilterBars and the KSelect into it. I know there is a bit of duplication, but its very simple code to duplicate, plus you save on the plumbing in DataPlaneList in adding the toolbar from The Outside

johncowen (commented):
I had a quick look at this last thing and had some comments, figured I'd submit those before coming back here tomorrow to finish up. Looking great tho!

src/test-support/mocks/src/dataplanes/_overview.ts
kleinfreund: Fixed.

src/test-support/mocks/src/meshes/_/dataplanes/_/_overview.ts
kleinfreund: Fixed.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: I moved the values of the filter and those in the list into the message files so we only need to adjust them there now based on what we decide.

kleinfreund:
> How about "Built-in Gateway", "Delegated Gateway" and "Standard Proxy"?

I moved the values of the filter and those in the list into the message files so we only need to adjust them there now based on what we decide.

I used those terms for now. Will bring it up again when this is close to done.

kleinfreund:
I’ll be adding some tests here that make sure that the DPP detail view’s Policies tab always uses the right "mode" depending on Dataplane type. I noticed I could easily get tests to pass while _not_ handling this correctly.

src/app/data-planes/components/DataPlaneList.vue
kleinfreund: I moved the toolbar back into the views. I’m generally not happy with the amount of duplicated markup for the sake of easier configurability. A lot of the times, it seems like a better idea to pass around data rather than markup.

johncowen: Cool, thanks. I agree that its not always super clear cut, and here it's definitely not.

Couple of points:

> A lot of the times, it seems like a better idea to pass around data rather than markup.

Markup _is_ data, just in a different format, its XML rather than JSON.

---

I think it might be worth making a `DataPlaneFilterBar` component which you would then slot into the toolbar slot in both places. As it stands right now tho, personally I'm still not totally sure I prefer the DataPlaneList component at all and we should just use AppCollection in both places. Just to be clear no action here, kinda talking out loud. I think once we get these description texts into i18n there'll be even less that could diverge.

I'm gonna take another proper look at the changes in a sec.

src/app/data-planes/components/DataPlaneList.vue
johncowen: 👍

johncowen:
> I’ll be adding some tests here that make sure that the DPP detail view’s Policies tab always uses the right "mode" depending on Dataplane type. I noticed I could easily get tests to pass while not handling this correctly.

Awesome, good spot! let me take another look through this.

kleinfreund:
I cleaned up the mocks for dataplanes some more.

I noticed we weren’t correctly reading a `tag`-based filter (we accessed `tags` instead of `tag`). I made it so that now the service detail view’s DPP list actually only shows DPPs with the expected type plus made it so that its service tags will also match the service.

That’s the last of what I wanted to address in this PR.

src/test-support/mocks/src/dataplanes/_overview.ts
```diff
@@ -1,41 +1,66 @@
...
-              outbound: [
-                {
-                  port: 10001,
-                  tags: {
-                    'kuma.io/service': 'frontend',
-                  },
-                },
-              ],
-            },
+            networking: fake.kuma.dataplaneNetworking({ type, inbounds, isMultizone, service }),
```
johncowen, line 60: Again, just a mental note and no action here.

---

I get the reasoning why this is being done, but it does move away from the point of our mocks being as they are. 

The idea is they should look like the API response, and in making reusable functions like this we hide the structure of the JSON away, and then the more that functions depend on each other it gets even more difficult to follow what is going on. In the end if we keep going like this it leads to:

```js
export default ({ fake, pager, env }: EndpointDependencies): MockResponder => (req) => {
   return fake.kuma.dataplane({ fake, pager, env, req }) <== the start of deeply nested functions
}
```

And then trying to see the complete JSON involves figuring out all the nested function calls.

So in removing duplication, we are trading that for less clarity. Its kinda the "small functions considered harmful" thing again.

Ideally, I'd like our mocks to be generated from the OpenAPI schemas, that would keep things clear and duplication would cease to be an issue.

I had a few other ideas on this, but I think it would be better to just wait and get on the OpenAPI auto generation thing when we can.

---

I also think adding the types to all these is just moving the problem of potentially wrong API responses over to potentially wrong types. Our types aren't what the API produces, and we've seen this issue many times, the copy/pasted API responses are.

Having or being able to copy/paste API responses straight from the actual API response of the backend with the odd primitive value replaced `fake` kinda prevents the lack of clarity, its also kind of a poor mans auto-generation.

I think overall its a bit of a balance, the sooner we can get to auto generate these and also the types themselves the better.

johncowen (commented):
I think this is almost good to go

I spotted this:

Screenshot 2023-11-08 at 11 11 11

The filter bar being too small ^

Other than that I think I had one question about `tags: {}` in the inline comments, the rest I think is just talking out loud. Let me know and I think we are good to go with this 👍

features/mesh/dataplanes/Index.feature
```diff
@@ -120,11 +98,112 @@ Feature: mesh / dataplanes / index
...
+      And the URL "/meshes/default/dataplanes/_overview" responds with
+        """
+        body:
+          items:
+            - name: fake-transmitter-gateway_builtin-0
+              dataplane:
+                networking:
+                  gateway:
+                    type: 'BUILTIN'
+                    tags: {}
```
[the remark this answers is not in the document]
kleinfreund: This can’t be filled out by mocks reliably right now. When we generate the mocks for a list, the default mocks will have random Dataplane types. We do mock out the first element here to have the type we expect, but the original mock might not have been a gateway type and therefore doesn’t always have `gateway.tags`. I updated this test to actually set a `kuma.io/service` tag here so it’s not confusing.

features/mesh/dataplanes/Index.feature
```diff
@@ -120,11 +98,112 @@ Feature: mesh / dataplanes / index
...
+      Then the URL "/meshes/default/dataplanes/_overview" was requested with
+        """
+        searchParams:
+          gateway: 'false'
+        """
+      And the "$item" element exists 1 time
+      And the "$item:nth-child(1)" element contains
+        | Value               |
+        | fake-system-proxy-0 |
+        | Standard proxy      |
```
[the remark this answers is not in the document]
kleinfreund: Yeah, I noticed that, too. Something to improve on.

features/mesh/dataplanes/Index.feature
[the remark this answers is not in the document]
kleinfreund: I did update the mocks slightly to always produce service tags in use cases where that’s expected, but that doesn’t actually help here.

johncowen (approved):
> The filter bar being too small ^

Looks to me like the filter bar thing is fixed here now?

In that case this LGTM!

kleinfreund:
> Looks to me like the filter bar thing is fixed here now?

Yep.

PR #1682, opened 2023-10-27
feat(zones): adds summary trays to list views

kleinfreund:
## Changes

Adds a summary trays to the zone list views.

Resolves #1644.

Signed-off-by: Philipp Rudloff <philipp.rudloff@konghq.com>

PR #1743, opened 2023-11-07
chore: Merges `main-overview` module with `control-planes` modules

johncowen:
Merges `main-overview` module with `control-planes` module, so a `main-overview` module no longer exists.

Notes:
- The `control-planes` module is already being added in `main.ts`
- I moved the `MainOverview` component (now `ControlPlaneStatus`) to no longer be injected via the `@/components/index` file. Eventually this file will cease to exist. 
- I didn't rename the i18n keys prefixed with `main-overview` as yet, I'll do that in a follow up.

There were a couple of "while I was here" things, that I'll note inline.


src/app/control-planes/components/ControlPlaneStatus.vue
```diff
@@ -85,21 +85,18 @@
...
-import { useI18n } from '@/utilities'
 
 const { t } = useI18n()
 const can = useCan()
 </script>
 <style lang="scss" scoped>
 .card-header {
   margin-bottom: $kui-space-50;
-  height: 42px;
+  min-height: 42px;
```
johncowen, line 100: Ensures that on very narrow screens that DOM stacking works correctly/fluidly

src/app/control-planes/sources.ts
```diff
@@ -1,20 +1,26 @@
...
 }
 
 export type ControlPlaneAddressesSource = DataSourceResponse<ControlPlaneAddresses>
-export const sources = (env: Env['var']) => {
+export type GlobalInsightSource = DataSourceResponse<GlobalInsight>
+
+export const sources = (env: Env['var'], api: KumaApi) => {
   return {
-    '/control-plane/addresses': async (_params: {}, source: { close: () => void }): Promise<ControlPlaneAddresses> => {
-      source.close()
```
johncowen, line 13: We recently removed the need to close sources if you are returning a straight forward promise, so I removed all this as I was here.

src/app/control-planes/views/item/IndexView.vue
```diff
@@ -107,21 +107,23 @@ import { KUI_ICON_SIZE_30 } from '@kong/design-tokens'
...
 <style lang="scss" scoped>
 .card-header {
   display: flex;
+  flex-wrap: wrap;
   justify-content: space-between;
   align-items: center;
   margin-bottom: $kui-space-50;
   // Makes card header the same height as buttons so that they align with or without buttons present.
-  height: 42px;
+  min-height: 42px;
```
johncowen, line 126: Similar to the above, ensures DOM stacking on narrow screens.

johncowen (commented):
Inlines:

cypress.config.ts
```diff
@@ -43,7 +58,10 @@ export default defineConfig({
       on(
         'file:preprocessor',
         createBundler({
-          plugins: [createEsbuildPlugin(config)],
+          plugins: [
+            createEsbuildPlugin(config),
+            createVuePlugin(),
+          ],
```
johncowen, line 64: This has a little story behind it:

1. Because our e2e.ts 'main' file pulls in `development.ts` in order to know about the mocks for usage with Cypress's `intercept` it also pulls in a `source` that we decorate for development purposes, despite the fact that we don't need that decorated `source` in Cypress.
2. In order to decorate the `source`, we import the `control-planes` module, as that contains the `source` that we want to decorate.
3. Beginning with this PR we import a `.vue` file from the `control-planes` module, which Cypress then chokes on.

The green above here, essentially turn `.vue` files to export `""` when running in Cypress. We don't actually need the vue files/components in Cypress but just because of how things are currently organized we statically import the `vue` files in Cypress.

---

We should think about splitting of a service container definition file that contains only things about mocks that we can share between browser (MSW), Cypress (intercept) and possibly Express-like servers. But for the moment this approach solves the issue in the short term.

kleinfreund (commented):
This mostly looks good to me.

There is one thing that I want to address here: I don’t want to keep naming files `IndexView.vue`. I mentioned multiple times how this impacted my ability to find where stuff is in the code base significantly.

Please stick to the pretty well-established pattern to include the module name in one form or another (e.g. ControlPlaneDetailView.vue).

johncowen:
Just to link to a previous conversation in a _feature PR_ where this came up, it also came up offline previous to that (which is also mentioned in that PR)

https://github.com/kumahq/kuma-gui/pull/1682#pullrequestreview-1706267264

Just to note this is a _refactoring PR_, specifically about renaming things to our newer style, although granted not specifically about the names of `.vue` files.

#1682 decided not to follow the naming convention of the `zones`, `zone-insights` and `zone-egresses` modules. That being - not repeating the namespace of the file in the name of the file. I didn't see a technical reason why there was a problem with not repeating the namespace of the file in the name of the file and I don't think you've given on as yet, but I decided to approve the PR in order to get the _feature_ into the software and then circle back.

---

Here is a screengrab of my editor now:

Screenshot 2023-11-08 at 16 40 29

Note from #1682 I've no idea what the extension is on the `ZoneIngressSummar` name when looking at that. I don't really know whether the name ends in `Summary`. Is this a `.spec` file? Is it a `ZoneIngressSummaryAbstractView`, I've no idea. I've also noticed that this kinda problem is also beginning to happen in the GH web UI in the file explorer during code reviews.

Here's a screengrab of the the `zones` module in my text editor previous to me removing the namespace from the name of the file for the `zones` related modules.

Screenshot 2023-11-08 at 15 08 07

I still find it really hard to look at that screengrab, so I'm really glad I changed it. Both of the above screengrabs show how redundant it is to prefix every file within a folder with the same name of the folder.

Could I widen the sidebar in my editor? Of course I can, but its an extra thing to keep having to do, but I could. Pretty sure I can't on GH though. Would widening my editor sidebar it make that last screengrab any easier to parse quickly, I'd say no.

---

On the other hand, in our past conversations you've pointed out that you have found it hard to find the files in your fuzzy finder with these shortened file names, although I did point out that as the fuzzy finder will also use the name of the folder in its search, there should be no real difference and I didn't get any response back from you on that, so I kinda assumed that was problem solved.

> Does ZoneListView technically contain more characters than IndexView? Yes. Does that make ZoneListView a worse name? No. Doesn’t make it any harder for you to find it and does make it easier for me to find it.

The technical fact there is that is that `IndexView` has less characters than `ZoneListView` and therefore is easier to read in a sidebar. The personal opinions are that ZoneListView isn't a 'worse name' or that it makes it harder or easier for someone to find. And those points are all omitting that fact that its redundant to include the namespace of a thing in the name of the thing that has already been namespaced. Or that with `IndexView` as opposed to `TabView` (the UI might change), or `AbstractView` (abstract generally means its not directly used) is maybe a safer more descriptive choice.

---

> Please stick to the pretty well-established pattern to include the module name in one form or another (e.g. ControlPlaneDetailView.vue).

I wouldn't call `ControlPlaneDetailView` exactly well established, I could probably point to a few repos where they don't repeat the namespace in the name of the file (e.g. https://github.com/Kong/kong-manager/tree/main/src/pages/certificates). In this repository specifically I'd say the split is probably 60/40 amongst our modules, Its mostly only 60/40 due to the fact that we've spent quite a long time organising this repository a little better that it was previously and we are now nearing the end of that, and naming things a little better was one of the last things on my list.

---

After all that being said I'd be fine to change this to use `ControlPlaneDetailView` for the moment if you still have an issue with any of the above, but if so, I think another MADR would then be a good idea instead of flip-flopping backwards and forwards with this. This would give another opportunity to provide some technical reasoning around this rather than just "I prefer it"

Let me know either way and we can either move on here, or I can change the name, move this to a MADR around the value of namespaces and then either change it back or leave as is depending on the outcome of that.

kleinfreund:
Your arguments for why you prefer the index-style file names seem to be:

- 1) preference to defer to the containing directory as the sole name-giving element
- 2) editor configuration not allowing you to see moderately long file names
- 3) short file names being more readable than long ones

Point 1 is something that’s not impacted by different file names so it doesn’t matter if the file is called IndexView or ZoneListView

Point 2 is something that you control. You can configure your editor to show file names longer than X characters adequately. In the GitHub UI for diffs, there are always multiple places to see the file name, so should it be cut-off in some cases, it won’t prevent you from identifying what the file is about.

Point 3 is only really true in the most literal sense of being able to read and parse fewer characters more quickly. But really, a name IndexView that appears perhaps dozens of times in a code base is so much less clear than a name ZoneListView that appears once. Plus ZoneListView has the advantage that you don’t need to look at anything else to know what this file is.

Also, I very much understand that it’s _possible_ to identify an IndexView file in purpose by looking at its containing directory and path. My point is that I shouldn’t need to look at both. Identifying files by name is easy and I don’t want to change this habit unnecessarily. File names are commonly emphasized in UIs like project-wide full-text search and command palettes. File names **should** be the primary identifier of the file. That also makes it much easier to move it around because moving it around shouldn’t change the meaning of its contents.

I’m going to write an MADR for this.

For now, there is no action needed here.

johncowen:
> I’m going to write an MADR for this.

Sounds good! 

Do you want me to change the name so we can merge in the meantime, or shall we wait for the MADR?

I'd prefer to change the name to here `ControlPlaneDetailView` at least temporarily (or maybe not) so I can merge this. I've been waiting for a good time to do this PR and its been kinda hard to find a time when its not gonna clash with other PRs, and i would want it to get in the way of anything else.

Let me know, Im good either way. Separately, ping me when that MADR is there 👍

kleinfreund:
We should generally not wait for an MADR in already on-going work so we can take our time with it.

johncowen:
Ok lemme change the name here so we can merge. Just in case apart from that it sounded like the rest was good to go? Shout me if there was anything else here, otherwise I'll ping you when I've changed up the name.

kleinfreund (approved):

PR #1746, opened 2023-11-08
feat(control-planes): update onboarding alert logic

kleinfreund:
## Changes

**chore(meshes): avoids data source conflict**

Avoids a potential data source conflict with the URI `/meshes` which was recently changed to resolve to a MeshInsight collection response. The existing data source for `/mesh-insights` has the same implementation and is better suited in terms of its name for views accessing `MeshInsight`s. For the MainOverviewView component, the `/meshes` data source was changed to the `/mesh-insights` data source.

**feat(control-planes): update onboarding alert logic**

Changes the logic for showing the onboarding alert to rely on data from the `/global-insight` API and to only show the alert is there is exactly one mesh (we decided to drop the extra conditions for that mesh being called “default” and there being no dataplanes).

Persists the onboarding alert dismiss state in local storage.

Moves the onboarding alert code into the control planes module and adds the OnboardingAlert component as a component in its DI service configuration.

Adds a link to onboarding to the help menu.


src/app/meshes/sources.ts
```diff
@@ -30,7 +30,7 @@ export const sources = (api: KumaApi) => {
       const { size } = params
       const offset = params.size * (params.page - 1)
 
-      return api.getAllMeshInsights({ size, offset })
+      return api.getAllMeshes({ size, offset })
```
kleinfreund, line 33: **Comment**: The `/meshes` data source is used by OnboardingNotification.

johncowen: Q: What is the difference between the `total` property of the `meshes` endpoint and the `total` property of the `mesh-insights` endpoint? Do they both refer to the same piece of data? i.e. the total of meshes?

kleinfreund: Yes.

johncowen: Ok so, with that info, given that you are highly likely to land on the homepage that also uses `getAllMeshInsights`/`/mesh-insights` at the same time that the `OnboardingNotification` component makes its request, I would think it's a good idea to reuse that same single request instead of making two separate ones in order to get the same data? 

That being said, unless I'm missing something, I think the improvement here would be to leave everything as it is, and just make sure that the `size=`'s are the same, i.e. both should have `size=10`.

I think that would mean that in the majority of cases we only make 1 request.

johncowen: P.S. Ahhhh I remember now the OnboardingNotification currently has `size=50` as I wanted to match the it up with the original mesh listing back when I moved all of onboarding to use DataSource. This would have been previous to having `/me` and possibly before the smaller mesh listing on the homepage. 🤔 

Ok, this is starting to be clearer now. This isn't as clear cut as to what `size=` should be on the 3 DataSources involved here, but I still think they should all use the same `/mesh-insights` API endpoint. I'm kinda possibly thinking they should all use the `size` from `/me`.

I'm starting to think we should at least close this PR and possibly reopen another later, probably changing all 3 DataSources use `size=` from `/me`.

---

This is all dependent on me not having missed something here, let me know if so.

kleinfreund: Does DataSource, actually ensure this request is not duplicated? If we can’t ensure that, we’re over-fetching data for nothing.

johncowen: It doesn't 'ensure', it just makes it possible/likely.

It also caches, so if we use `/mesh-insights` on the OnboardingNotification, it means that if you then click on the mesh listing the cached data is already there as the 'loading state'.

kleinfreund: OnboardingNotification will be loaded on every page (i.e. also when someone refreshes the page) whereas the other two components are specific to their respective views. It’s basically chance that we sometimes avoid querying `/mesh-insights` twice plus it’s incidental that the home page happens to query `/mesh-insights`. Yes, it’s the most common home page, but now we started to call `/mesh-insights` on every page load instead of just where we need it.

I think we should load `/meshes` for this purpose. It’s the cheaper endpoint to query.

johncowen: Would it be fair to say that on any page below and `/meshes` if we get a 200 from the API we know that at least 1 mesh exists? (and for `/meshes` itself we query `/mesh-insights` so we can reuse that)

kleinfreund: I don’t understand the question. How is this assumption relevant to the issue at hand?

johncowen: Well in order to show the OnboardingNotification we firstly need to know if a mesh exists, which is why we make that API call that we are talking about. If we already know one exists, then we don't need to make it.

kleinfreund: Well, yes, but this component is also used in non-mesh pages so it doesn’t help us making any assumptions, right?

johncowen: Yeah for sure, but it sounds like a fair assumption to make is that if we land on any page below `/meshes` and its a 200 then we know a mesh exists.

So we are left with anything under browser URL `/zones` I think? I would guess you might be able to have a zone without having a mesh? Kinda unlikely, especially given we are talking about onboarding.. Actually 🤔  I'd say if you have a zone or a mesh then we shouldn't show the onboarding notification?

If that is the case we can probably say if we get a 200 from any page under `/meshes` or `/zones` then we know not to show the OnboardingNotification? If we land on `/gui` then we already use `/mesh-insights`, if we land on `/meshes` then we use already use  `/mesh-insights`. Does that sound correct?

kleinfreund: I think this is very much out of scope for this PR. We can of course consider changing the condition under which we show the onboarding notification, but it was never supposed to use the `/mesh-insights` API.

johncowen: Well I think it is very much in scope, and this is why I suggested closing this, or at the very least think it through, as I think we should think this through properly before making 2 requests for an incredibly common landing page rather than just the 1.

I think its important to understand that when I changed the code that you are changing back here, I did it whilst trying to reduce the amount of HTTP requests we make given we had a big problem with that. So 

>  it was never supposed to use the /mesh-insights API.

Yes, it was supposed to use the `/mesh-insights` API as I wanted to share the requests/cache.

Continuing to use `/mesh-insights` and sharing those requests means less requests for the incredibly common case. For the much less common case (and therefore slightly less important), we could improve in another differently scoped PR if you wanted to do that by omitting the request entirely if we get a 200 on deeper pages because we already know we have a mesh and/or zone.

> We can of course consider changing the condition under which we show the onboarding notification

If you are concerned about the less common landing pages, that would be the better approach I think. That means we don't make the request at all right? So better from a frontend and backend perspective, and for the incredibly common landing pages we continue to reuse the single request, again better for frontend and backend. 

---

On the 'scope' thing. This PR has a title "avoids data source conflict". What's the conflict we are avoiding?

kleinfreund: My main goal here is to avoid calling the data source that retrieves `MeshInsight`s `/meshes`. It’s too easily confused with one that retrieves `Mesh`es.

I still think we rather avoid the request to `/mesh-insights` on every load in favor of using the lighter `/meshes` endpoint.

I find it would be a practice to rely on incidental facts like "we already happen to load mesh insights on the home page so we should also use mesh insights in a component we use on every page".

Again, deciding whether we want to use another metric for whether we show the onboarding notification **is** out of scope of this PR.

johncowen: > My main goal here is to avoid calling the data source that retrieves MeshInsights /meshes. It’s too easily confused with one that retrieves Meshes.

When do we need the API response from `/meshes`?

> I still think we rather avoid the request to /mesh-insights on every load in favor of using the lighter /meshes endpoint.

We can't avoid this if we visit and/or land on the `/gui` or `/gui/meshes` pages, we need it there, but we can reuse it. If we land on deeper pages than those we can potentially avoid it entirely by checking for a 200 on deeper pages. If we do that, would that solve your problem?

kleinfreund: > When do we need the API response from `/meshes`?

For onboarding notifications.

> If we land on deeper pages than those we can potentially avoid it entirely by checking for a 200 on deeper pages. If we do that, would that solve your problem?

We have several pages that aren’t mesh-associated. That’s not a good substitute.

johncowen: > For onboarding notifications.

Previous to this PR, that didn't use `/meshes` and therefore currently in `master` that doesn't use `/meshes`.

Its a little strange using the change you added to justify the reason for making the change. Don't make the change, then there is no justification.


Is there any other place where we need the API response from `/meshes`?

---

> We have several pages that aren’t mesh-associated.

We have meshes (and children), and zones (and children), we also have diagnostics.

I've explained meshes and covered zones (we could reuse the zone listing response, and then use the same 200 check for anything deeper), we are then left with diagnositics. Are there any others we need to think about?

kleinfreund: > Previous to this PR, that didn't use `/meshes` and therefore currently in `master` that doesn't use `/meshes`.

I think I already explained how the prior change to make onboarding notification not use the /meshes API is the source of its removal and that it shouldn’t have been changed there.

So I’m not justifying my change here with something I’m introducing also but rather with the observation that it should’ve stayed like this in the first place.

> I've explained meshes and covered zones (we could reuse the zone listing response, and then use the same 200 check for anything deeper), we are then left with diagnostics. Are there any others we need to think about?

That’s a significantly more complicated logic than what we had originally which is to make a simple call to `/meshes` to find out.

johncowen: > That’s a significantly more complicated logic than what we had originally which is to make a simple call to /meshes to find out.

Maybe it is maybe it isn't, but in an ideal world this would be better for the least common usecase, yes? If it was super straight forwards to do would you prefer to do this?

Thinking about it more, we just don't show the onboarding notification on any subpages, seeing as we assume if you get a 200, then you have a mesh. That shouldn't be too difficult should it?

If you get a 404 (we are now getting into really really uncommon use case territory) we could only then call `/mesh-insights` to get the total of meshes. In fact considering how uncommon this probably is, I'd say if we get a 404 don't do anything, you have a `Go Home` button there to click on, if you click that, you'd go home, and therefore call the `/mesh-insights` endpoint (and hence share that with onboarding)

---

So:

1. If you are on a 'deep' page with a 200, just don't show the onboarding notification.
2. If you are on a deep' page and get a 404, don't do anything new, we expect the user to click the `Go Home` button
3. If you land on a higher page then we re-use `/mesh-insights`, we need to request that anyway.

^ that doesn't sound too complicated? I think its just check whether we are on a 'deep' page before showing the onboarding notification?

lahabana:
Add to the `?` top menu an onboarding entry.
Only show onboarding notification on home page so we don't need to do any extra requests to figure if we need to display this.
Make sure we don't list DPP for the default mesh to check if we need to show this notification

src/app/meshes/sources.ts
kleinfreund: This conversation was resolved through the decision to only use the onboarding notification on the home page and to use `/global-insight` data for it.

See https://github.com/kumahq/kuma-gui/pull/1746#issuecomment-1807882804

src/app/control-planes/views/ControlPlaneDetailView.vue
```diff
@@ -13,109 +13,164 @@
...
+        />
+
+        <LoadingBlock v-else-if="globalInsight === undefined" />
+
+        <template v-else>
+          <div
+            class="stack"
+            data-testid="detail-view-details"
+          >
+            <template v-if="globalInsight.meshes.total <= 1 && globalInsight.dataplanes.standard.total === 0">
```
kleinfreund, line 32: **Question**: Does this condition make sense?

We previously checked for `meshes.total === 1`, but I suppose the onboarding alert is equally is useful when there aren’t any meshes, right?

Also, this is missing the check for whether the one mesh was the default mesh.

kleinfreund:
I updated this PR to incorporate the changes to the onboarding notification we discussed. I think the main thing I want to check again is the condition for showing said notification: https://github.com/kumahq/kuma-gui/pull/1746#discussion_r1391119635

src/app/control-planes/views/ControlPlaneDetailView.vue
johncowen: tbh I'm not sure of the reasoning behind not doing this anymore, it no longer costs us an additional HTTP request anymore, in fact the OnboardingNotification kinda no longer makes HTTP requests anymore (depending on the race for sharing of course).

Is there a case that we want to show the onboarding notification if you have deleted (or otherwise don't have) the `default` mesh, and have possibly replaced it with another one? Maybe we said that in the sync we had, but if we did could you remind me as I don't remember.

Maybe this is related to:

> but I suppose the onboarding alert is equally is useful when there aren’t any meshes, right?

Wouldn't zero meshes mean this isn't a fresh install anymore, and therefore don't show the onboarding? Is there a case where zero meshes means "fresh install"?

kleinfreund: I changed this to check for the default mesh again and to only trigger when there is exactly one mesh.

src/app/control-planes/views/ControlPlaneDetailView.vue
```diff
@@ -13,111 +13,166 @@
...
-              <DataSource
-                v-slot="{ data, error }: ZoneOverviewCollectionSource"
-                src="/zone-cps?page=1&size=10"
+        <DataSource
+          v-slot="{ data: meshInsightsData, error: meshInsightsError }: MeshInsightCollectionSource"
+          src="/mesh-insights?page=1&size=10"
+        >
+          <DataSource
+            v-slot="{ data: zoneOverviewsData, error: zoneOverviewsError }: ZoneOverviewCollectionSource"
+            src="/zone-cps?page=1&size=10"
```
johncowen, line 26: Does this mean we are loading zones data even if they aren't enabled?

src/app/control-planes/views/ControlPlaneDetailView.vue
```diff
@@ -13,111 +13,166 @@
...
+            <template v-else>
+              <div
+                class="stack"
+                data-testid="detail-view-details"
               >
-                <ErrorBlock
-                  v-if="error"
-                  :error="error"
-                />
+                <template v-if="globalInsight.meshes.total === 1 && meshInsightsData?.items[0].name === 'default' && globalInsight.dataplanes.standard.total === 0">
```
johncowen, line 40: What is the difference between the `globalInsight.meshes.total` and the `meshInsightsData.total`?

src/app/control-planes/views/ControlPlaneDetailView.vue
```diff
@@ -13,111 +13,166 @@
...
+                          appearance="primary"
+                          size="small"
+                          class="action-button"
+                          :to="{ name: 'onboarding-welcome' }"
+                        >
+                          {{ t('main-overview.detail.onboarding.get_started_link') }}
+                        </KButton>
+                      </div>
+                    </template>
+                  </KAlert>
```
johncowen, line 63: Considering this file is used in different environments, do we need this here if we don't have onboarding installed?

src/app/control-planes/views/ControlPlaneDetailView.vue
```diff
@@ -13,111 +13,166 @@
...
+import LoadingBlock from '@/app/common/LoadingBlock.vue'
 import { useControlPlaneStatus } from '@/app/control-planes'
 import MeshInsightsList from '@/app/meshes/components/MeshInsightsList.vue'
 import type { MeshInsightCollectionSource } from '@/app/meshes/sources'
 import ZoneControlPlanesList from '@/app/zones/components/ZoneControlPlanesList.vue'
 import type { ZoneOverviewCollectionSource } from '@/app/zones/sources'
 
 const ControlPlaneStatus = useControlPlaneStatus()
+
+const isShowingOnboardingAlert = ref(true)
```
johncowen, line 165: Now that this in in the View and not the never re-rendered ApplicationShell, I think this would mean that after the pressing the cross to close the alert and coming back to the homepage again, the alert would remain?

johncowen (commented):
I left some comments, I'm still not totally sure what this PRs aim is now that https://github.com/kumahq/kuma-gui/pull/1756 resolved the `source` removal issue.

Could you update the description so that following #1756 it describes what the aim is here? Maybe that also might help with the questions I had here.

src/app/control-planes/views/ControlPlaneDetailView.vue
kleinfreund: Hm, yes, that’s correct and a bug. I moved the zone data source up here because having two of them here and the third deeper in was strange. I might just move it back.

src/app/control-planes/views/ControlPlaneDetailView.vue
kleinfreund: There shouldn’t be any.

src/app/control-planes/views/ControlPlaneDetailView.vue
kleinfreund: Yes, but the same happens when reloading the page right now. We don’t really attempt to persist this decision anymore. We should probably do so using localStorage.

src/app/control-planes/views/ControlPlaneDetailView.vue
lahabana: I thought we agreed on just looking at number of meshes == 1 ? This would enable to avoid the triple nesting of datasources and much quicker 1st frame rendering no? It would also make the code a LOT more readable I believe

src/app/control-planes/views/ControlPlaneDetailView.vue
johncowen: Previous to this PR we already do that, and we already check in that component if you are on the homepage

https://github.com/kumahq/kuma-gui/pull/1756/files#diff-c6fc3196605e3b659cfce445494e393e521da228c007c2ae53870e7106da454bR5

src/app/control-planes/views/ControlPlaneDetailView.vue
johncowen: I agree, but in the meantime but at the very least right now, at least we don't continue showing the thing everytime you leave the homepage to say go to the onboarding and come back again.

I suppose this was why I was asking what the aim of this PR is? If we aren't adding the localStorage then this takes away functionality rather than add. Could you update the description of the PR here with the aim of it? I think then it will be easier to understand whats going on here.

src/app/control-planes/views/ControlPlaneDetailView.vue
kleinfreund: I added the component to the control-planes service config so it can be removed this way.  I know the alert was functionally already isolated to the home page, but I think this mechanism should be in the view used on the home page and not in ApplicationShell.

src/app/control-planes/views/ControlPlaneDetailView.vue
kleinfreund: I’ve added the local storage mechanism now. I also updated the PR description to reflect the current state of this PR more closely.

src/app/control-planes/views/ControlPlaneDetailView.vue
kleinfreund: This is fixed now. I moved the data sources back to where they were.

src/app/control-planes/views/ControlPlaneDetailView.vue
kleinfreund: I think checking for there being exactly one mesh is enough. The check for this is now simpler and only requires moving the GlobalInsight data source a bit further up in the relevant component.

src/app/control-planes/components/OnboardingAlert.vue
```diff
@@ -0,0 +1,50 @@
...
+</script>
+
+<style lang="scss" scoped>
+.onboarding-alert-content {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: $kui-space-60;
+}
+</style>
```
johncowen, line 50: If we are maintaining this component as a separate component then this file should go in the onboarding module not the control-planes module because we said we would locate these sorts of file based on "What they are".

I'm just taking another look at this PR as a whole as I see there have been quite a few changes recently, I just wanted to leave this note here before I look at it again.

kleinfreund: I’ve moved the component into the onboarding module now.

src/app/kuma/components/ApplicationShell.vue
```diff
@@ -80,6 +80,13 @@
                   Feedback
                 </a>
               </KDropdownItem>
+
+              <KDropdownItem
+                :item="{
+                  to: { name: 'onboarding-welcome' },
+                  label: 'Onboarding',
+                }"
+              />
```
lahabana, line 89: There should be a conditional or something here so that in Mink you can't do onboarding still no?

johncowen: This ApplicationShell doesn't happen in MinK, theres another implementation in MinK

src/app/control-planes/views/ControlPlaneDetailView.vue
```diff
@@ -108,15 +126,20 @@
...
+import LoadingBlock from '@/app/common/LoadingBlock.vue'
 import { useControlPlaneStatus } from '@/app/control-planes'
 import MeshInsightsList from '@/app/meshes/components/MeshInsightsList.vue'
 import type { MeshInsightCollectionSource } from '@/app/meshes/sources'
+import { useOnboardingAlert } from '@/app/onboarding'
 import ZoneControlPlanesList from '@/app/zones/components/ZoneControlPlanesList.vue'
 import type { ZoneOverviewCollectionSource } from '@/app/zones/sources'
 
 const ControlPlaneStatus = useControlPlaneStatus()
+const OnboardingAlert = useOnboardingAlert()
```
johncowen, line 140: Why don't we try decorating `ControlPlaneStatus` with `OnboardingAlert`?

That way there is zero code to do with Onboarding in the `ControlPlaneDetailView` (which is used for showing control-planes everywhere). Plus if you think about it, it actually works with "the status of the control plane is that it is a fresh install" so it belongs in `ControlPlaneStatus`. This is how we solve other instances of this problem.

That being said if you want to make the additional changes for this elsewhere first then I'd be fine approving this if all is ok and circling back at a later date to see if we can use the above approach instead. There is nothing technically here that wouldn't work, and the objective here is to move the onboarding notification to use `/global-insights`.total rather than `/meshes`.total which this PR does.

Lmk either way, but it would be good to get this PR resolved sooner rather than later.

johncowen (commented):
I left an overall comment in an inline where it was relevant-ish.

src/app/control-planes/views/ControlPlaneDetailView.vue
kleinfreund: I’ve implemented this in https://github.com/kumahq/kuma-gui/pull/1746/commits/892b9c6de9d2e576e05d9b98772cf3ff8a3684f2 now.

Side note: Consumers of ControlPlaneStatus will also need to provide the props now required by this component.

johncowen: > Side note: Consumers of ControlPlaneStatus will also need to provide the props now required by this component.

Yeah thats kinda fine if we are changing the interface, all the props you've added here make sense for everywhere anyway. If not we could have also used DataSource and `can` inside the ControlPlaneStatus component which would avoid the interface change _and_ there would be a reason for diverging from the norm here. (Just incase: don't do this, just go with what you have already done and make the changes to the interface elsewhere where necessary)

Could you confirm that this works everywhere for me? There is multiple levels decoration going on for this (which is great and how it should work), but I haven't tested that personally. I kinda assume everything was fine?

kleinfreund: Yeah, I've checked it for two things decorating the component and saw both additions. I'll reference it in the appropriate place.

johncowen: Awesome 🎉  Sounds good then lets go with this. 1 sec lemme take one last look here

Issue #1749, opened 2023-11-09
investigate zone creation polling


johncowen:
### What happened?

The frontend behaviour and backend behaviour don't quite match up.

Currently the frontend expects the possibility of a 404 to mean "this zone is still waiting to be connected", this "404 while waiting for connection" will never happen. We always get a 200 response whilst waiting for a zone to be connected, unless...

One potential way we would get a 404 would be if we were waiting for a zone to be connected, and then say in another tab, I delete the zone. The "Waiting to be connected" tab would now be receiving 404s and incorrectly continue polling waiting for a connection to a zone that no longer exists.

There are few other details to this interaction that we need to match up better and make sure this interaction is easy to recreate correctly/entirely in mocks and during testing.

I may add some more info here at a later date (also please ask if the above isn't clear)

PR #1754, opened 2023-11-10
fix(zones): Zone creation should error normally on 404 responses

johncowen:
Back in https://github.com/kumahq/kuma-gui/pull/1634 we changed the functionality of checking for a connected zone by additionally catching a 404 error and retrying the request, which I'd wrongly assumed how this functioned.

The endpoint only ever returns a 404 if the zone is deleted (i.e. doesn't exist). It doesn't 404 if the zone exists but is not connected.

Due to the change, it was possible to create a zone, without connecting it, and then in say another tab, delete the zone, and the scanner would continue to spin due to the 404s.

I've removed the 404 retries, and also added further testing for making sure a 404 shows the error page when the scanner is waiting to connect, and I also updated the mocks so that they are closer to the real responses.

We decided that we should check for a `zone.enabled` (https://github.com/kumahq/kuma-gui/pull/1634#discussion_r1376119606) but I don't think we ever need to check for `zone.enabled` for the purposes of this PR (or #1634) when checking for whether the zone is connected. I haven't altered anything there as yet as it doesn't have any effect, but I have updated the mocks here so that we don't get `enabled: true`, only a non-existant `enabled` (i.e. defaults to `true`) or `false`, which I 'think' is more correct.

I re-added some comments explaining this a little better in a couple of places, and I almost renamed and moved around functions to help the code self-comment, but I'm a little way into a longer piece of work of collecting up zone data related functions de-duping them and refactoring them, which I'd rather continue to do separately rather than in this PR. The comments here will help when I continue with that.

There are a couple of other things here that I'll note inline.

Closes https://github.com/kumahq/kuma-gui/issues/1749

### DIFF
--- a/features/zones/Create.feature
+++ b/features/zones/Create.feature
@@ -133,6 +133,11 @@ Feature: zones / create
     When I visit the "/zones/-create" URL
     And I "type" "test" into the "$name-input" element
     And I click the "$create-zone-button" element
+    And the URL "/zones/test/_overview" was requested with
+      """
+      method: GET
+      """
+    And the "$waiting" element exists
     And the URL "/zones/test/_overview" responds with
       """
       headers:
--- a/features/zones/Create.feature
+++ b/features/zones/Create.feature
@@ -16,10 +16,12 @@ Feature: zones / create
       | environment-kubernetes-config       | [data-testid='zone-kubernetes-config']               |
       | ingress-input-switch                | [for='zone-ingress-enabled']                         |
       | egress-input-switch                 | [for='zone-egress-enabled']                          |
+      | create-error                        | [data-testid='create-zone-error']                    |
       | waiting                             | [data-testid='waiting']                              |
       | connected                           | [data-testid='connected']                            |
-      | error                               | [data-testid='create-zone-error']                    |
+      | error                               | [data-testid='error']                                |
       | instructions                        | [data-testid='connect-zone-instructions']            |
+
     And the environment
       """
       KUMA_MODE: global
@@ -89,6 +91,19 @@ Feature: zones / create
     And the "$egress-input-switch input" element doesn't exist
     And the "$environment-universal-config" element contains "globalAddress: grpcs://<global-kds-address>:5685"
 
+  Scenario: The zone get connected successfully and shows the success
+    Given the environment
+      """
+      KUMA_SUBSCRIPTION_COUNT: 0
+      """
+    And the URL "/provision-zone" responds with
+      """
+      body:
+        token: spat_595QOxTSreRmrtdh8ValuoeUAzXMfBmRwYU3V35NQvwgLAWIU
+      """
+    When I visit the "/zones/-create" URL
+    And I "type" "test" into the "$name-input" element
+    And I click the "$create-zone-button" element
     Given the environment
       """
       KUMA_SUBSCRIPTION_COUNT: 1
@@ -97,13 +112,33 @@ Feature: zones / create
       """
       body:
         zone:
-          enabled: true
+          enabled: !!js/undefined
         zoneInsight:
           subscriptions:
             - connectTime: '2020-07-28T16:18:09.743141Z'
               disconnectTime: !!js/undefined
       """
     Then the "$connected" element exists
+
+  Scenario: The zone is deleted whilst waiting to be connected and shows an error
+    Given the environment
+      """
+      KUMA_SUBSCRIPTION_COUNT: 0
+      """
+    And the URL "/provision-zone" responds with
+      """
+      body:
+        token: spat_595QOxTSreRmrtdh8ValuoeUAzXMfBmRwYU3V35NQvwgLAWIU
+      """
+    When I visit the "/zones/-create" URL
+    And I "type" "test" into the "$name-input" element
+    And I click the "$create-zone-button" element
+    And the URL "/zones/test/_overview" responds with
+      """
+      headers:
+        Status-Code: 404
+      """
+    Then the "$error" element exists
 
   Scenario: The form shows expected error for 409 response
     Given the URL "/provision-zone" responds with
@@ -116,7 +151,7 @@ Feature: zones / create
     And I "type" "test" into the "$name-input" element
     And I click the "$create-zone-button" element
 
-    Then the "$error" element exists
+    Then the "$create-error" element exists
     Then the "$instructions" element doesn't exist
 
   Scenario: The form shows expected error for 400 response
@@ -147,7 +182,7 @@ Feature: zones / create
     And I "type" "zone.eu" into the "$name-input" element
     And I click the "$create-zone-button" element
 
-    Then the "$error" element doesn't exist
+    Then the "$create-error" element doesn't exist
     And the "$name-input-invalid-dns-name" element exists
     And the "$instructions" element doesn't exist
 
@@ -155,7 +190,7 @@ Feature: zones / create
     And I "type" "test" into the "$name-input" element
     And I click the "$create-zone-button" element
 
-    Then the "$error" element doesn't exist
+    Then the "$create-error" element doesn't exist
     And the "$instructions" element exists
 
   Scenario: Exiting the form in a safe state without confirm dialog
--- a/src/app/application/services/data-source/CallableEventSource.ts
+++ b/src/app/application/services/data-source/CallableEventSource.ts
@@ -25,11 +25,12 @@ export default class CallableEventSource<T = {}> extends EventTarget {
     this._open()
   }
 
-  _open() {
+  protected _open() {
     (async function (self) {
       try {
-        self.readyState = 0
+        self.readyState = CONNECTING
         const source = self.source()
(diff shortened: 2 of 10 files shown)


features/zones/Create.feature
```diff
@@ -89,6 +91,19 @@ Feature: zones / create
...
+      KUMA_SUBSCRIPTION_COUNT: 0
+      """
+    And the URL "/provision-zone" responds with
+      """
+      body:
+        token: spat_595QOxTSreRmrtdh8ValuoeUAzXMfBmRwYU3V35NQvwgLAWIU
+      """
+    When I visit the "/zones/-create" URL
+    And I "type" "test" into the "$name-input" element
+    And I click the "$create-zone-button" element
```
johncowen, line 106: Previously the below assertion was merged in with a scenario titled `The form interactions behave correctly`. I wanted it to be clearer that we are testing whether we show that the zone connects successfully so I split the test into two. So the green above here is just the same setup copy/pasta'ed

features/zones/Create.feature
```diff
@@ -97,14 +112,34 @@ Feature: zones / create
       """
       body:
         zone:
-          enabled: true
+          enabled: !!js/undefined
```
johncowen, line 115: I 'think' this is more correct

features/zones/Create.feature
```diff
@@ -97,14 +112,34 @@ Feature: zones / create
...
+    When I visit the "/zones/-create" URL
+    And I "type" "test" into the "$name-input" element
+    And I click the "$create-zone-button" element
+    And the URL "/zones/test/_overview" responds with
+      """
+      headers:
+        Status-Code: 404
+      """
+    Then the "$error" element exists
+
```
johncowen, line 142: New test to test that we show an error state on 404. If someone ever makes a similar same mistake again this would fail and cause whoever was doing it to think twice.

src/app/application/services/data-source/CallableEventSource.ts
```diff
@@ -25,11 +25,12 @@ export default class CallableEventSource<T = {}> extends EventTarget {
...
   }
 
-  _open() {
+  protected _open() {
     (async function (self) {
       try {
-        self.readyState = 0
+        self.readyState = CONNECTING
         const source = self.source()
+        self.readyState = OPEN
```
johncowen, line 33: There where several things I did whilst checking in DataSource whilst checking this again and following things around/debugging.

This first one was:

1. We have a `CONNECTING=0` constant a little higher up in this file, I guess this was just missed when this was originally written/iterated on
2. I moved setting the readyState to `OPEN` here
    i. to not expose this kind of stuff to the EventSources `Callable`
    ii. use the OPEN constant which I would rather remain private-ish

src/app/application/services/data-source/index.ts
```diff
@@ -84,11 +83,9 @@ export const create: Creator = (src, router) => {
...
     // second argument here can go
     const init = route.route(params, { close: () => {} })
-    if (init instanceof CallableEventSource) {
-      return init
-    } else {
-      return source(() => Promise.resolve(init))
-    }
+    const eventSource = init instanceof CallableEventSource ? init : source(() => Promise.resolve(init))
+    eventSource.url = src
+    return eventSource
```
johncowen, line 88: These are changes I made whilst exactly tracing/loggin/debugging how things were working, I left them in as they are kinda useful. Little caveat here is that `url` should really be `readonly`, but I didn't want to do too much refactoring here to make this a little more protected. As it stands now sometimes its handy to be able to see the DataSources `src`/`uri` easily.

src/app/zones/locales/en-us/index.yaml
```diff
@@ -76,7 +76,7 @@ zones:
         description: 'Establish a connection to your remote zone by following the instructions to setup dependencies and install the remote zone control plane.'
       scanner:
         title: 'Waiting for Zone to be connected …'
-        description: ' '
+        description: ''
```
johncowen, line 79: I noticed this from a while back, we changed this so you no longer need to use `' '` so I updated this while I was here.

src/app/zones/sources.ts
```diff
@@ -38,30 +38,22 @@ export const sources = (source: Source, api: KumaApi) => {
...
-          const hasStatus = <T extends {status: number}>(e: unknown): e is T => typeof (e as T).status !== 'undefined'
-          if (hasStatus(e)) {
-            const status = e.status.toString()
-            switch (true) {
-              case status === '404':
-                // wait 2 seconds and try again
-                return new Promise((resolve) => setTimeout(resolve, 2000))
-            }
+          if (e instanceof ZoneOfflineError) {
+            return new Promise((resolve) => setTimeout(resolve, 2000))
```
johncowen, line 56: The actual fix is here. We no longer consider a 404 a signal to retry, only a 'zone offline' would trigger a retry.

src/app/zones/views/CreateView.vue
```diff
@@ -284,28 +284,29 @@
...
-                  </div>
-
-                  <div class="form-section__content">
-                    <DataSource
-                      v-slot="{ data, error: scanError }: ZoneOverviewSource"
-                      :src="`/zone-cps/online/${name}?no-cache=${Date.now()}`"
-                      @change="success"
+                  <DataSource
+                    v-slot="{ data, error: scanError }: ZoneOverviewSource"
+                    :src="`/zone-cps/online/${name}?no-cache=${now}`"
```
johncowen, line 289: I moved the DataSOurce up a level so I can stop showing the 'title' if there is an error (the title and the body copy currently say "Waiting etc etc". If there is an error we should no longer say "Waiting" anywhere.

I also moved the poor mans cache bust date to be the same throughout the lifetime of the CreateView route. There'll be a better way to define cache busting at a later date.

src/app/zones/views/CreateView.vue
```diff
@@ -284,28 +284,29 @@
...
+                    @change="success"
+                  >
+                    <div
+                      v-if="!scanError"
+                      class="form-section__header"
                     >
+                      <h2 class="form-section-title">
+                        {{ t('zones.form.section.scanner.title') }}
+                      </h2>
+                      <p>{{ t('zones.form.section.scanner.description') }}</p>
```
johncowen, line 299: `description` no longer needs to check for `' '`

src/test-support/mocks/src/zones/_/_overview.ts
```diff
@@ -12,91 +12,99 @@ export default ({ env, fake }: EndpointDependencies): MockResponder => (req) =>
       creationTime: '2021-02-19T08:06:15.380674+01:00',
       modificationTime: '2021-02-19T08:06:15.380674+01:00',
       zone: {
-        enabled: fake.datatype.boolean(),
+        ...(fake.datatype.boolean()
+          ? {
+            enabled: true,
+          }
+          : {}),
```
johncowen, line 19: The `enabled` change mentioned in the description.

src/test-support/mocks/src/zones/_/_overview.ts
```diff
@@ -12,91 +12,99 @@ export default ({ env, fake }: EndpointDependencies): MockResponder => (req) =>
...
-                  responsesAcknowledged: `${fake.number.int(30)}`,
-                },
-                TrafficPermission: {
-                  responsesSent: `${fake.number.int(30)}`,
-                  responsesAcknowledged: `${fake.number.int(30)}`,
-                },
-                TrafficRoute: {
-                  responsesSent: `${fake.number.int(30)}`,
-                  responsesAcknowledged: `${fake.number.int(30)}`,
+        ...(subscriptionCount !== 0
```
johncowen, line 22: The mocks change to be closer to the real API as mentioned in the description.

johncowen (commented):
Inlines:

features/zones/Create.feature
kleinfreund: How so?

features/zones/Create.feature
```diff
@@ -97,14 +112,34 @@ Feature: zones / create
...
             - connectTime: '2020-07-28T16:18:09.743141Z'
               disconnectTime: !!js/undefined
       """
     Then the "$connected" element exists
 
+  Scenario: The zone is deleted whilst waiting to be connected and shows an error
+    Given the environment
+      """
+      KUMA_SUBSCRIPTION_COUNT: 0
+      """
```
kleinfreund, line 127: **Question**: Why does this need 0 subscriptions?

johncowen: Because the scenario is 'whilst waiting to be connected'. It "should" fire the successful request after `And I click` but before `And the URL: 404`, but thats not clear enough in the test and possibly not guaranteed to happen before the `And the URL`.

I've added a couple more assertions in https://github.com/kumahq/kuma-gui/pull/1754/commits/fb09a7070bda66bd8e1bb23c6377c7ac7d15741d so we can guarantee that at least one successful (but disconnected) request happens.

kleinfreund (approved):

PR #1756, opened 2023-11-13
fix(onboarding): Fixes error with onboarding when detecting new install

johncowen:
The long story:

During https://github.com/kumahq/kuma-gui/pull/1728 a DataSource source was removed (https://github.com/kumahq/kuma-gui/pull/1728/files#diff-8c51094d188779df11ec5a418503cf5a77c088bb12581335955952e7deef4c3dL50-L65), and replaced by https://github.com/kumahq/kuma-gui/pull/1728/files#diff-8c51094d188779df11ec5a418503cf5a77c088bb12581335955952e7deef4c3dR100-R122

Then, as well as this usage:

https://github.com/kumahq/kuma-gui/pull/1728/files#diff-9d4c2f48081cf9b206eb208a8f15ad06ede0411df7c550dbd938e1926d89b2d9L21

... there is another usage of this DataSource for onboarding which, if you only have 1 mesh called `default` would use this DataSource to figure out if you had added any dataplanes yet, if not we would show you the onboarding notification.

Since the source was removed, the DataSource would error with a DataSource not found error.

---

Firstly, I added the source back and then added a test to cover this eventuality, as well as another for the negative.

As we had discussed https://github.com/kumahq/kuma-gui/pull/1746, I also made some small changes here to:

1. Restrict the OnboardingNotification to the homepage by checking for `route.name === 'home'`
2. Use the information already available about dataplanes in the `/mesh-insights` API endpoint, meaning we no longer have to make the additional HTTP request to inspect the amount of dataplanes. This then meant I could re-remove the source whose removal caused the error in the first place.
3. Add a further test to cover the "only show this on the homepage" case
4. As mentioned in https://github.com/kumahq/kuma-gui/pull/1746#discussion_r1387763698 I also made sure that the `?size=`s used for the homepage and the onboarding notification where them same, and could therefore re-use the same request.
5. Note I am still using the `/mesh-insights` API request. I believe it gives us the same information we would get from `/global-insights` plus it also means we can inspect the name of the only mesh. This means if the only mesh is called `default`, then its a possibly a fresh install and we may show the onboarding, if its called `another-mesh` then it's not a fresh install and we shouldn't ever show the onboarding.

Potential further improvement work that is not necessary in this PR:

1. I left a TODO in the mocks to properly use `KUMA_DATAPLANE_COUNT`, which will then need 'partioning' to give us a spread of types from the total requested.
2. Now that onboarding is only on the homepage we should extract it from ApplicationShell, and embed it into ControlPlaneStatus (the 'status' being "is this control plane fresh or not"). Importantly we should try to do this "as a micro service" so that when our onboarding isn't installed at all then we don't include any code whatsover to do with onboarding.

I did not add the help menu item for onboarding here and we should do that as a separate PR aside from the bugfix PR.

### DIFF
--- a/src/app/onboarding/components/OnboardingNotification.vue
+++ b/src/app/onboarding/components/OnboardingNotification.vue
@@ -1,50 +1,48 @@
 <template>
+  <!-- As well as only showing the notification on the homepage -->
+  <!-- We also share the same datasource URL as the mesh listing on the homepage -->
   <DataSource
-    v-if="!route.meta.onboardingProcess"
+    v-if="route.name === 'home'"
     v-slot="{ data }"
-    :src="`/meshes?page=1&size=50`"
+    :src="`/meshes?page=1&size=10`"
   >
     <template
       v-if="data?.items.length === 1 && data.items[0].name === 'default'"
     >
-      <DataSource
-        v-slot="{ data: dataplaneData }"
-        :src="`/meshes/${data.items[0].name}/dataplanes?page=1&size=50&search=`"
+      <template
+        v-if="data.items[0].dataplanes.total === 0"
       >
-        <template
-          v-if="dataplaneData?.total === 0"
+        <div
+          v-if="alertClosed === false"
+          class="onboarding-check"
+          data-testid="onboarding-notification"
         >
-          <div
-            v-if="alertClosed === false"
-            class="onboarding-check"
+          <KAlert
+            appearance="success"
+            dismiss-type="icon"
+            @closed="closeAlert"
           >
-            <KAlert
-              appearance="success"
-              dismiss-type="icon"
-              @closed="closeAlert"
-            >
-              <template #alertMessage>
-                <div class="alert-content">
-                  <div>
-                    <strong>Welcome to {{ t('common.product.name') }}!</strong> We've detected that you don't have any data plane proxies running yet. We've created an onboarding process to help you!
-                  </div>
-
-                  <div>
-                    <KButton
-                      appearance="primary"
-                      size="small"
-                      class="action-button"
-                      :to="{ name: 'onboarding-welcome' }"
-                    >
-                      Get started
-                    </KButton>
-                  </div>
+            <template #alertMessage>
+              <div class="alert-content">
+                <div>
+                  <strong>Welcome to {{ t('common.product.name') }}!</strong> We've detected that you don't have any data plane proxies running yet. We've created an onboarding process to help you!
                 </div>
-              </template>
-            </KAlert>
-          </div>
-        </template>
-      </DataSource>
+
+                <div>
+                  <KButton
+                    appearance="primary"
+                    size="small"
+                    class="action-button"
+                    :to="{ name: 'onboarding-welcome' }"
+                  >
+                    Get started
+                  </KButton>
+                </div>
+              </div>
+            </template>
+          </KAlert>
+        </div>
+      </template>
     </template>
   </DataSource>
 </template>
--- a/features/onboarding/Index.feature
+++ b/features/onboarding/Index.feature
@@ -1,9 +1,37 @@
 Feature: onboarding / index
   Background:
     Given the CSS selectors
-      | Alias            | Selector                               |
-      | skip-button      | [data-testid='onboarding-skip-button'] |
-      | environment-text | [data-testid='kuma-environment']       |
+      | Alias                   | Selector                                |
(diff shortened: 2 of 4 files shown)


src/app/onboarding/components/OnboardingNotification.vue
```diff
@@ -1,50 +1,48 @@
 <template>
+  <!-- As well as only showing the notification on the homepage -->
+  <!-- We also share the same datasource URL as the mesh listing on the homepage -->
   <DataSource
-    v-if="!route.meta.onboardingProcess"
+    v-if="route.name === 'home'"
```
johncowen, line 5: Only show on `home`

src/app/onboarding/components/OnboardingNotification.vue
```diff
@@ -1,50 +1,48 @@
...
     v-slot="{ data }"
-    :src="`/meshes?page=1&size=50`"
+    :src="`/meshes?page=1&size=10`"
```
johncowen, line 7: change to `10` to sync up with the same DataSource used on the homepage.

src/app/onboarding/components/OnboardingNotification.vue
```diff
@@ -1,50 +1,48 @@
...
   >
     <template
       v-if="data?.items.length === 1 && data.items[0].name === 'default'"
     >
-      <DataSource
-        v-slot="{ data: dataplaneData }"
-        :src="`/meshes/${data.items[0].name}/dataplanes?page=1&size=50&search=`"
+      <template
+        v-if="data.items[0].dataplanes.total === 0"
```
johncowen, line 13: Inspect the data from `/mesh-insights` that we already have instead of using an additional DataSource

src/app/onboarding/components/OnboardingNotification.vue
```diff
@@ -1,50 +1,48 @@
...
       >
-        <template
-          v-if="dataplaneData?.total === 0"
+        <div
+          v-if="alertClosed === false"
+          class="onboarding-check"
+          data-testid="onboarding-notification"
```
johncowen, line 18: Added a testid for testing

src/app/onboarding/components/OnboardingNotification.vue
```diff
@@ -1,50 +1,48 @@
...
-              </template>
-            </KAlert>
-          </div>
-        </template>
-      </DataSource>
+              </div>
+            </template>
+          </KAlert>
+        </div>
+      </template>
```
johncowen, line 45: If I haven't commented on a line in this file, the rest of the green is just a consequence of removing the DataSource that we no longer need.

johncowen (commented):
Inlines to help review the very few changes in a sea of green:

kleinfreund:
Why was this PR created? I had already incorporated the changes we discussed in https://github.com/kumahq/kuma-gui/pull/1746.

johncowen:
> Why was this PR created?

Because I noticed the bug and wanted to fix it plus add the tests to make sure that we didn't regress again. https://github.com/kumahq/kuma-gui/pull/1746 doesn't mention anything about a bug or a fix required to fix it. Plus I wanted to get a bugfix PR up sooner rather than later as I know https://github.com/kumahq/kuma-gui/pull/1746 has been up for a while and from looking at the latest changes I think there are still things that would take time to address.

I also wanted to be super clear that there was a bug to fix, and then ensure there are tests to cover it so the same bug doesn't happen again.

This PR:

1. Adds tests to reproduce the bug and prevent future regressions here, then..
2. Fixes the bug
3. Improves the feature by doing what was outlined in https://github.com/kumahq/kuma-gui/pull/1746, and maintains the previous functionality.
4. Adds more tests ontop
5. Uses very little code to do all the above, thus avoiding other problems down the line. This is important as its a bugfix PR rather than a feature or refactor PR.

Just so you know, I spotted this just as I was finishing up last night, so decided to PR my investigation and fix, assuming that we would want it.

All that being said, I wouldn't at all mind closing this in favour of https://github.com/kumahq/kuma-gui/pull/1746, but https://github.com/kumahq/kuma-gui/pull/1746 would have to address the above points also, and I would have to know that it could be potentially merged soonish (this PR has less code to review and should be a straight 👍 )

Let me know either way.

kleinfreund (approved):
To note: This PR does not supersede https://github.com/kumahq/kuma-gui/pull/1746. Please remove the note from the PR description.

PR #1798, opened 2023-11-22
feat: tweak tag list behavior

kleinfreund:
fix: tag list taking up too much width sometimes

chore: truncate TagList only in tables

### DIFF
--- a/src/app/common/TagList.vue
+++ b/src/app/common/TagList.vue
@@ -1,5 +1,11 @@
 <template>
-  <KTruncate width="auto">
+  <component
+    :is="props.shouldTruncate ? 'KTruncate' : 'div'"
+    :width="props.shouldTruncate ? 'auto' : undefined"
+    :class="{
+      'tag-list': !props.shouldTruncate,
+    }"
+  >
     <KBadge
       v-for="(tag, index) in tagList"
       :key="index"
@@ -13,7 +19,7 @@
         {{ tag.label }}:<b>{{ tag.value }}</b>
       </component>
     </KBadge>
-  </KTruncate>
+  </component>
 </template>
 
 <script lang="ts" setup>
@@ -27,9 +33,12 @@ interface LabelValueWithRoute extends LabelValue {
   route: RouteLocationNamedRaw | undefined
 }
 
-const props = defineProps<{
+const props = withDefaults(defineProps<{
   tags: LabelValue[] | Record<string, string> | null | undefined
-}>()
+  shouldTruncate?: boolean
+}>(), {
+  shouldTruncate: false,
+})
 
 const tagList = computed<LabelValueWithRoute[]>(() => {
   const labels = Array.isArray(props.tags) ? props.tags : getLabels(props.tags)
@@ -84,3 +93,11 @@ function getRoute(tag: LabelValue): RouteLocationNamedRaw | undefined {
   }
 }
 </script>
+
+<style lang="scss" scoped>
+.tag-list {
+  display: inline-flex;
+  flex-wrap: wrap;
+  gap: $kui-space-40;
+}
+</style>
--- a/src/app/data-planes/components/PolicyTypeEntryList.vue
+++ b/src/app/data-planes/components/PolicyTypeEntryList.vue
@@ -29,6 +29,7 @@
               <TagList
                 v-if="rowValue.length > 0"
                 class="tag-list"
+                should-truncate
                 :tags="rowValue"
               />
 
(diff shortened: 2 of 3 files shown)


src/app/common/TagList.vue
```diff
@@ -1,5 +1,11 @@
 <template>
-  <KTruncate>
+  <component
+    :is="props.shouldTruncate ? 'KTruncate' : 'div'"
```
johncowen, line 3: Should we use the `AnonymousComponent` here (which is kinda the `absent` of component-land)?

Its an alternative to `if/else` when you don't want anything in the `else` portion, i.e. you really want to use an `if`

src/app/common/TagList.vue
```diff
@@ -27,9 +33,12 @@ interface LabelValueWithRoute extends LabelValue {
   route: RouteLocationNamedRaw | undefined
 }
 
-const props = defineProps<{
+const props = withDefaults(defineProps<{
   tags: LabelValue[] | Record<string, string> | null | undefined
-}>()
+  shouldTruncate?: boolean
+}>(), {
+  shouldTruncate: false,
```
johncowen, line 40: I think there was a note a while ago saying everyone should generally prefer using shorter names for things (so just `something` instead of `isSomething`, `hasSomething`, `shouldSomething`) 

It would be nice for us to try and stick to that where possible. So that being said, could we call this `truncate`?

src/app/common/TagList.vue
```diff
@@ -1,5 +1,11 @@
 <template>
-  <KTruncate>
+  <component
+    :is="props.shouldTruncate ? 'KTruncate' : 'div'"
+    :width="props.shouldTruncate ? 'auto' : undefined"
+    :class="{
+      'tag-list': !props.shouldTruncate,
```
johncowen, line 6: oooh look, would using the `AnonymousComponent` mean you don't need to remember to do any of this? Not totally sure, but if so that would be another +1 for `AnonymousComponent`

johncowen (commented):
Thanks for doing this, a small detail but an important one I think.

I left a couple of potential improvements, lmk and then we can get this 👍 I think.

---

Not for this PR I don't think, but I did wonder if we should 'ellipsis truncate' these when the truncation is opened. I'm in two minds though, as its just as easy for me to see these altogether in the summary panel now by just clicking the table row. So I suppose I'm just thinking out loud here.

Screenshot 2023-11-22 at 10 24 47

src/app/common/TagList.vue
kleinfreund: But we do need the `div` to attach `.tag-list` as without KTruncate, we don’t have tag list styles.

src/app/common/TagList.vue
kleinfreund: Not really, no. We need the `width` conditional only when using KTruncate and we need the `tag-list` class only when _not_ using KTruncate. The latter we can’t even do with AnonymousComponent because we do need an actual element.

See also https://github.com/kumahq/kuma-gui/pull/1798#discussion_r1401841381.

src/app/common/TagList.vue
johncowen: Ah gotcha, cool lets stick with the `div` then

src/app/common/TagList.vue
kleinfreund: I’m not a fan of that decision because it makes for generally more weird-to-read code and sometimes straight up confusing code (most commonly when the variables are "simple" verbs or nouns like "copy", "truncate", "error", etc.; less so for things like "loading").

That being said, at least in a typed code base, one can usually discover that a variable is a boolean and that a prop is expected to get a boolean value.

I’m not sure that’s a rule we should/must follow and we should probably decide on that separately. For the time being, seeing as pretty much all our booleans are named in this is/has/are/should style, I’m gonna stick to this. It’s relatively easy to find cases of this later on and change them, if we want to.

johncowen: > As part of the Kongponents redesign effort, we’re going to attempt to standardize our component prop names. We’d like to move away from prefixing props, especially booleans, in favor of naming them without the prefix.

Would be kinda nice to start making our naming similar to the majority of our components, especially seeing as we've literally just upgraded to v9 so this standardization felt like a good time to start but 🤷 

> That being said, at least in a typed code base, one can usually discover that a variable is a boolean and that a prop is expected to get a boolean value.

Yeah completely agree with that one

Gimme 2 tics let me approve this

johncowen (approved):

PR #1799, opened 2023-11-22
chore: upgrade to vite 5

### LINKED ISSUE #1791: Upgrade Vite 4 to 5
Migration guide: https://vitejs.dev/guide/migration.html

**To do**:

- Update package @vitejs/plugin-vue
- Remove package vite-plugin-rewrite-all
- Update package vite-svg-loader
- Set package.json `type` field to `module`
- Change file extension of the following files to `.cjs` as they don’t support (or aren’t imported) as ES modules:
  - dev-utilities/deprecatedUtilityClasses.js
  - .eslintrc.js
  - stylelint.config.js
- Change module format of the following files to ESM:
  - docgen.config.js
- Change import `import createEsbuildPlugin from '@badeball/cypress-cucumber-preprocessor/esbuild'` to `import { createEsbuildPlugin } from '@badeball/cypress-cucumber-preprocessor/esbuild'`

**Blockers**:

- [vite-plugin-html](https://github.com/vbenjs/vite-plugin-html) using deprecated features:
  - `plugin 'vite:html' uses deprecated 'enforce' option. Use 'order' option instead.`
  - `plugin 'vite:html' uses deprecated 'transform' option. Use 'handler' option instead.`
  - **Possible strategy**: Use [Vite’s native environment variable replacing mechanism](https://vitejs.dev/guide/env-and-mode.html#html-env-replacement).
- [vite-plugin-html-env](https://github.com/lxs24sxl/vite-plugin-html-env) using deprecated features:
  - `plugin 'rollup-plugin-html-env' uses deprecated 'transform' option. Use 'handler' option instead.`
  - **Possible strategy**: Use [Vite’s native environment variable replacing mechanism](https://vitejs.dev/guide/env-and-mode.html#html-env-replacement).

kleinfreund:
Updates Vue 3.3.x to 3.4.x.

Upgrades vite from v4 to v5 (see https://vitejs.dev/guide/migration.html#migration-from-v4).

Sets `pkg.type` to `module`.

Changes the extension of the files `dev-utilities/deprecatedUtilityClasses.js`, `.eslintrc.js`, and `stylelint.config.js` to `.cjs` as they’re CommonJS modules and can’t be converted to ESM.

Changes the module format of the file `docgen.config.js` to ESM.

Removes vite-plugin-rewrite-all as it’s no longer needed.

Removes vite-plugin-html in favor of using Vite’s native environment variable replacing logic.

Resolves #1791.


johncowen:
@kleinfreund could you let me know what this upgrade is giving us, apart from it being the new major

kleinfreund:
Seeing as Vite is our primary tool for all things development, we should really try and stick to its latest versions. Over time, we’ll be more and more locked out of fixes and security updates that we want or need.

johncowen:
Right, but seeing as its only on 5.0.2 and the public site doesn't yet have docs for v5, it could also give us unforeseen breakages and security bugs, plus hard to find docs and probably some other things.

I'm not saying lets not ever upgrade, but I'm just wondering jumping on the latest thing might cause more issues just for lack of being a little more patient.

I'm not totally sure about `vite`s support, but I would also guess in the immediate future if anything was to be amiss with the far more stable and "battle tested" v4 which we are currently on, it would be released as a bugfix for v4. Again not saying we should stay on on v4 forever.

---

So, could you let me know what this upgrade is giving us, apart from it being the new major?

kleinfreund:
> and the public site doesn't yet have docs for v5,

What makes you say that? The website seems to reflect v5 for me.

johncowen:
😆 there's no v5 in the Version menu

---

**So, could you let me know what this upgrade is giving us, apart from it being the new major?**

johncowen:
So here's how I would have answered my own question:

> So, could you let me know what this upgrade is giving us, apart from it being the new major?

Yeah sure:

- I mentioned in the description that the upgrade would mean we can remove `vite-plugin-rewrite-all `, Vite fixed the issue they had with `.` characters in URL paths which means we won't need the code here https://github.com/kumahq/kuma-gui/blob/6b453886e8eec507c171134429f5654b01b6865d/vite.config.production.ts#L30-L35
- ...
- ...
- ...

---

Just so we don't go back there. 

- I agree that updating software means the software you are using doesn't eventually become outdated. If you can explain to me what else this upgrade is giving us there might be more we can agree on.
- Yes, I saw that the site has the version 5 docs by default, I guess must be stupid and immediately looked in the Version menu to see what versions were in there, I found it funny 🤷 . Thanks for pointing that out.

---

So before this starts sounding like a backwards Monty Python sketch again:

> So, could you let me know what this upgrade is giving us, apart from it being the new major?

lahabana:
Let's reconsider merging this in January

vite.config.development.ts
```diff
@@ -1,12 +1,11 @@
...
+import { defineConfig, mergeConfig, type UserConfigFn, type UserConfig } from 'vite'
 
 import { config as prodConfig } from './vite.config.production'
 // https://vitejs.dev/config/
 export const config: UserConfigFn = (env) => {
   return mergeConfig(
     prodConfig(env),
-    ({
-    } as UserConfig),
+    {} satisfies UserConfig as UserConfig,
```
johncowen, line 8: Is the double `satisfies UserConfig as UserConfig` needed here?

.env.development
```diff
@@ -1 +1,4 @@
 VITE_ZONE_CREATION_FLOW=enabled
+# The following are templating placeholders for the index.html file. They’re going to be replaced using server-side templating when serving the GUI application in production.
+VITE_BASE_GUI_PATH=/gui
+VITE_KUMA_CONFIG='{"baseGuiPath":"/gui","apiUrl":"http://localhost:5681","version":"2.4.0","product":"Kuma","mode":"global","environment":"universal","storeType":"postgres","apiReadOnly":false}'
```
johncowen, line 4: Note: Not keen on this, but fine by me for now (not a blocker)

johncowen (commented):
LGTM, I had 1 question before 👍 

Also, while I remember, when we merge this try to remember to mention in the message that we did Vue here also.

vite.config.development.ts
kleinfreund: This PR has a couple of changes looking like this. They’re not essential to this upgrade and are a result of me trying to solve type issues I started getting on this change set (specifically the one where I’m adopting this in kong-mesh-gui). The configs here showed TypeScript errors related to not being able to use this config. Ultimately, this turned out to be caused by concurrent versions of Vite being installed. I needed to deduplicate them in the lock file to solve it.

I think it’s still good to leave those changes: If we actually change the config and add something here, the satisfies UserConfig bit will make sure the config is type-checked and alert us of things like misspelled options.

.env.development
kleinfreund: I’m not a huge fan of this either. It’s a good enough stop gap for what we need, but a bit more flexibility would be nice. Perhaps one of the plugins we used for this purpose becomes viable again. We can always switch back to them, I think.

vite.config.development.ts
johncowen: I suppose I meant the `as UserConfig` bit? I still get type errors if I remove it and typo a property? But I also get no type errors if I remove the `as UserConfig` and don't make any typos? Not a biggie, and I guess you just be getting errors if you remove the `as UserConfig`? But let me know on this one and we can approve/merg.

kleinfreund: You’re completely right. We don’t need the `as UserConfig` bit at all here. The `satisfies UserConfig` bit still seems to be necessary as I stop getting errors then.

I’ve removed the `as UserConfig` bit now.

johncowen (approved):

PR #1803, opened 2023-11-22
feat(dataplanes): update summary appearance

kleinfreund:
**feat: color kuma.io tags differently**

**feat(dataplanes): update summary appearance**

Updates the summary appearance to the new design.

Replaces uses of `--AppGap` with the appropriate design system variables.


src/app/common/TagList.vue
```diff
@@ -46,8 +49,9 @@
   return labels.map((tag) => {
     const { label, value } = tag
     const route = getRoute(tag)
+    const isKuma = label.includes('kuma.io/')
```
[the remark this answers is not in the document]
johncowen: I suppose this is vaguely valid 🤷 , `not-kuma.io` would get highlighted and we don't want that to happen, not a biggie but is there a different approach to avoid this?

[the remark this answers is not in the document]
kleinfreund: Changing this to something more robust like `label.includes('.kuma.io/') || label.startsWith('kuma.io/')` doesn’t really make things better. We can’t parse tag keys as URLs because they aren’t.

For the sake of argument we could fake the URL syntactically and use something like this:

```suggestion
    let isKuma = false
    try {
      const tagKeyAsUrl = new URL('https://' + label)
      isKuma = tagKeyAsUrl.hostname.endsWith('.kuma.io') || tagKeyAsUrl.hostname === 'kuma.io'
    } catch {}
```

That’s the most robust thing I can think of, but I also think it’s really doubtful that this is actually useful to get 100% accurate. We don’t process the result of this call other than by applying different colors.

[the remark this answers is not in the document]
johncowen: Well I was just thinking of fixing it instead of ignoring this alert instead, up to you. I would like to avoid this coming up in PRs either way. Would this thing appear in the Security tab in GH as well as PRs if we don't do either?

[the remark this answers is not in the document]
kleinfreund: > Well I was just thinking of fixing it instead of ignoring this alert instead, up to you. I would like to avoid this coming up in PRs either way.

What exactly needs fixing here? We’re not actually parsing URLs here so the CodeQL alerts isn’t valid. At most we’re talking about false positive matches for this `isKuma` check. `label.includes('.kuma.io/') || label.startsWith('kuma.io/')` would avoid matching `not-kuma.io`.

> Would this thing appear in the Security tab in GH as well as PRs if we don't do either?

They’ll be listed here for closed alerts: https://github.com/kumahq/kuma-gui/security/code-scanning?query=is%3Aclosed+branch%3Amaster

johncowen:
I took a look at the Preview site and had some questions:

1. I vaguely remember saying we should show non-truncated tags in the side bar. I can see a single tag in the listing and I click for the summary I probably want to see them all, especially considering we have room in the summary. (I know the design doesn't show this, but I think that could just be an oversight, maybe at least double check)
2. I also remember talking about colors in the sync, we've gone with purple for kuma here but the designs have blue for kuma and gray. I've seen the gray being used in the designs for the thing I'm working on also. So it would be good to confirm that either way, and if necessary open an issue or do that in a separate PR.

I also noticed this which looks a bit weird, but again could be for another PR:

Screenshot 2023-11-23 at 13 56 59

Otherwise I think this is looking a lot better, lemme take a scan of the code.

src/app/common/HorizontalDefinitionCard.vue
```diff
@@ -0,0 +1,40 @@
...
+  align-items: flex-start;
+  font-weight: $kui-font-weight-bold;
+}
+</style>
+
+<style lang="scss">
+.definition-card-container > * {
+  min-width: 0;
+}
+</style>
```
johncowen, line 40: Rather than copy/paste the code from `DefinitionCard` and tweak a line or two, is there anything else we can do here. I know sometimes the copy/paste thing is ok, just it feels like in this case there could be a better way to do this.

I think at the very least we should use different classnames, extra "at the very least" for the non-scoped style. But let me know if you have any ideas for a better way to do this.

kleinfreund:
> I vaguely remember saying we should show non-truncated tags in the side bar. I can see a single tag in the listing and I click for the summary I probably want to see them all, especially considering we have room in the summary.

Let’s check on this in the design sync then.

> I also remember talking about colors in the sync, we've gone with purple for kuma here but the designs have blue for kuma and gray.

That has changed in the design since I implemented this. Let’s check in on this in the design sync.

src/app/common/TagList.vue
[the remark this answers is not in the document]
johncowen: Nothing _needs_ fixing, just trying to suggest something that might work and avoid all this, as I said completely up to you.

---

What about using:

```html
data-key="kuma.io/service"
```

and then targetting that with css 

```css
[data-key*=".kuma.io/"], [data-key^="kuma.io/"]
```

I would guess that would remove the alert and is probably simpler?

You could even split by `/` and use `[0]` - `data-key="kuma.io"` `[class$=".kuma.io"], [class="kuma.io"]`, or something like that?

Would any of those work?

Just to be clear, no problem to leave as is if this alert doesn't continue to alert us anywhere.

src/app/common/TagList.vue
```diff
@@ -100,4 +104,13 @@ function getRoute(tag: LabelValue): RouteLocationNamedRaw | undefined {
...
 }
+
+.kuma-badge {
+  background-color: $kui-color-background-decorative-purple-weakest !important;
+}
+
+.kuma-badge,
+.kuma-badge a {
+  color: $kui-color-text-decorative-purple !important;
+}
```
johncowen, line 115: Is there any way to avoid the `!important`s here?

johncowen (commented):
I left a few suggestions, lmk

src/app/common/HorizontalDefinitionCard.vue
kleinfreund: I decided to integrate this into DefinitionCard via an optional `layout` prop. Initially I thought there’d be more differences, but that didn’t turn out that way.

src/app/common/TagList.vue
kleinfreund: I changed this a bit to avoid the `!important` even though in this case it was a perfectly fine use: we really don’t want anything to be able to override these colors again. That’s what `!important` is for.

src/app/common/TagList.vue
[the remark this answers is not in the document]
kleinfreund: I did change this to use `label.includes('.kuma.io/') || label.startsWith('kuma.io/')` in the off chance someone uses a non-conforming label including something liek `not-kuma.io/`, but I don’t think more caution is warranted here. The alert can be closed and won’t bother us again unless we change the code.

kleinfreund:
I’ve made a couple of changes here after the design sync:

- TagList
  - Uses the `neutral` appearance by default and the `default` appearance for kuma.io tags
  - Automatically truncates in any context if there are more than 10 tags
- Dataplane list view: Uses TagList for its Services column

I also added some mock data for non-kuma.io tags.

src/app/common/TagList.vue
```diff
@@ -31,13 +39,18 @@ import type { RouteLocationNamedRaw } from 'vue-router'
...
 }
 
 const props = withDefaults(defineProps<{
   tags: LabelValue[] | Record<string, string> | null | undefined
   shouldTruncate?: boolean
+  hideLabelKey?: boolean
+  alignment?: 'left' | 'right'
 }>(), {
   shouldTruncate: false,
+  hideLabelKey: false,
```
johncowen, line 52: How come we need this? How does someone know what the tag is if we only show the value? Is that what this is for?

kleinfreund: This is for the special case of the dataplane list where we show a specific subset of tags by value: service tags. The column reads "Services" so it’s unambiguous what they are. I wanted to avoid using a separate KTruncate/KBadge style listing for this.

johncowen: Oh I see! I quite like that, thanks!

src/test-support/FakeKuma.ts
```diff
@@ -242,15 +245,19 @@ export class KumaModule {
         'kuma.io/service': service ?? this.serviceName(),
         'kuma.io/protocol': this.protocol(),
         ...(zone && { 'kuma.io/zone': zone }),
+        ...additionalTags,
```
johncowen, line 248: I'm getting more and more concerned about the fact that our mocks are slowly becoming more complicated than the application itself.

It took me a while the other day to figure out why something wasn't showing that I thought should have been. We are also removing type safety/validation in areas where its sometimes important to and adding sometimes incorrect 'alarms' everywhere instead (the `true` vs `1` thing in the name of "making things simpler" from the other day)

I've mentioned before the usefulness of being able to see the API JSON structure in the single isolated mock file. I can just open it up and see what it is without having to go on a trip down the rabbit hole.

i.e. `fake.kuma.` helpers shouldn't really be used as an `import`.

Generally, they should be used for fake **values** not entire JSON structures. What is happening is very similar to these `<DoEverythingComponent />`s, they eventually get so complicated that bugs start to appear that get hard to find and fix and everything generally gets harder and harder to work with.

Its almost even more important here as we don't really want to start manually writing tests for our mock files.

Just so its said again, I get the duplication thing/issue and these ideally would be autogenerated, so we need to try and strike a balance until we get to autogeneration. If we never get to autogenerated mocks then we will soon end up in a situation where are mocks are as difficult to work with as our application used to be.

And yes I know, types! Just so its said, the autogeneration could be using the generated API data types to do that, but I want to avoid using the sometimes incorrect manually written data types i.e. the sooner we can autogenerate all this the better.

Finally, when amending mocks for testing purposes please think twice before creating multiple functions that are essentially `import` to import little bits of entire JSON structure. Ok sometimes maybe thats what we want, but sometimes its better to make 'value' type helpers and then compose so you can see the entire structure of the JSON in the single file.

---

In saying all of the above, there is a perfect opportunity here to make a `fake.kuma.tags()` helper function, which might help to explain when its a good opportunity to use a `fake.kuma.` helper, and when its not.

`tags` are arbitrary key/values, the property names don't really matter to us, its just a `Record<string, string>` that are mostly anything but also can have special kuma.io ones in. So yes its a JSON structure but its pretty much a 'value' seeing as we don't really care what that structure is.

It would be nice to have a `fake.kuma.tags()`

```js
tags: fake.kuma.tags(),
// or
tags: {
  ...fake.kuma.tags(),
  "kuma.io/zone": "zone"
}
// or something similar to that ^
```

wdyt think of the `fake.kuma.tags` idea?

---

Just to be clear this is non-blocking for this PR, just I keep seeing PRs with these mocks changes in, and I don't want to block any fixes/features etc but I do want to make sure our mocks don't get into the same state as the application used to be.

johncowen (approved):
LGTM! This looks much better overall, I like some of the extra details also 👏 

(I left one more inline comment but again not a blocker)

src/test-support/FakeKuma.ts
kleinfreund: Yeah, I get your concern. The dataplane mocks are more difficult to follow/assess at a glance nowadays. We use those utilities for 5 mocked endpoints currently (though we could consider dropping the /dataplanes/_overview mock as we’re not using this endpoint currently).

I’ll think some more about this. Using a tags utility is definitely worthwile.

johncowen: > (though we could consider dropping the /dataplanes/_overview mock as we’re not using this endpoint currently).

I'm pretty sure this would be a good idea also, I think theres only repetition because we are supporting endpoints we don't use.

👍

PR #1805, opened 2023-11-22
chore(dataplanes): truncate name column

kleinfreund:
## Changes

## Notes

- I’ve made this change specific to the DPP list for now. We could consider generalizing this to all list views, but I think this should be a good first step.
- I’ve updated the logic for DPP name generation in our mocks to be a bit more realistic. This isn’t quite complete, yet. Ideally, the DPP mocks would have the same zone prefix in the name as they have a zone tag in somewhere in `networking`. Another day.

### DIFF
--- a/src/app/data-planes/views/DataPlaneListView.vue
+++ b/src/app/data-planes/views/DataPlaneListView.vue
@@ -40,7 +40,6 @@
               <DataPlaneList
                 v-else
                 data-testid="data-plane-collection"
-                class="data-plane-collection"
                 :page-number="parseInt(route.params.page)"
                 :page-size="parseInt(route.params.size)"
                 :total="data?.total"
--- a/src/app/data-planes/components/DataPlaneList.vue
+++ b/src/app/data-planes/components/DataPlaneList.vue
@@ -1,5 +1,6 @@
 <template>
   <AppCollection
+    class="data-plane-collection"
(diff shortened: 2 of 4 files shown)


src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -143,4 +143,8 @@ import type { MeSource } from '@/app/me/sources'
   // Removes the bottom margin as we’re aligning the label with the select in a horizontal layout.
   margin-bottom: 0 !important;
 }
+
+.data-plane-collection :deep(.name-column) {
+  max-width: 400px;
+}
```
johncowen, line 149: How come we split this CSS in two? As it is now would mean that the Service > Data Plane Proxy listing wouldn't truncate, is that right?

johncowen: @kleinfreund wasn't sure if you'd seen this?

kleinfreund: I had, but forgot about it again. Thanks for the ping.

I’ll have a look at this now. I wasn’t entirely sure about the approach initially so I only rolled it out in one place. I’ll make sure it also applies to the other DPP listing.

kleinfreund: I changed this to apply to both Dataplane lists now.

johncowen (approved):

PR #1806, opened 2023-11-22
fix(dataplanes): service address and port not always being set

kleinfreund:


johncowen (approved):
👍 

We can improve on this later

PR #1807, opened 2023-11-22
feat(dataplanes): Add ServiceTraffic component

johncowen:
This PR adds 2 things:

1. A `ServiceTraffic` component and related sub-components. This is the component only and is not hooked up to any data.
2. In order to be able to build/PR/merge the above component in isolation without adding it to our actual application. I've added a very rough sandbox in our docs that can be seen via `yarn docs:dev` (we can add PR previews at a later date which will be much better)

Screenshot 2023-11-22 at 17 37 40

## ServiceTraffic component

I'm looking for feedback on the component itself, plus its subcomponents. Hopefully it's clear where the data will be plugged into this in a later PR and finally integrated into our application.

## Docs/Component Sandbox

I wanted to start getting this in as I would really like to use this more for building/manually testing/previewing our reusable components. It has the additional benefit of being able to work on, PR and merge components that aren't yet used in our application. Almost like feature flags but instead of feature flagging our in-production application, we just use a completely different application. In the future we should be able to gradually build entire features in isolation using this sandbox application, then just enable them in production when we want to.

As you can see the code here is very much in-progress. I went out of my way not to alter any of our application code just yet, but I think some injection related things will change ever so slightly to allow us to reuse things better.

Lastly I'd be happy to close this PR either way and wait for everything to be integrated, but I wanted to get feedback early on both of the above features and I also wanted to do a final CI check on the sandbox stuff and start thinking about CI integration. Let me know either way. I'd also be happy splitting these features into separate PRs, but really in order to test/experience one you need the other.

---

FIXMEs should be resolved and removed before merge, I'll probably add some inlines here also.


.vitepress/theme/index.ts
```diff
@@ -1,6 +1,7 @@
 // https://vitepress.dev/guide/custom-theme
 import { h } from 'vue'
 import Theme from 'vitepress/theme'
+import '../../src/assets/styles/main.scss'
```
kleinfreund, line 4: **Issue**: This breaks the dark theme. Seeing as you’re trying to isolate the components in an iframe, we probably don’t need this, right?

image

docs/frames/default.md
```diff
@@ -0,0 +1,134 @@
...
+})
+</script>
+
+<div id="sandboxed-component" ref="el"></div>
+
+<style scoped>
+#sandboxed-component {
+  width: 100%;
+  height: 100vh;
+  padding: 0.5rem;
```
kleinfreund, line 131: **Suggestion**: To ensure this doesn’t look weird in dark mode.

```suggestion
  padding: 0.5rem;
  background-color: #fff;
```

image

.vitepress/theme/style.css
```diff
@@ -1,3 +1,8 @@
+iframe {
+  width: 100%;
+  border: 1px solid #CCC;
+  padding: 50px;
```
kleinfreund, line 4: **Suggestion**: To ensure this doesn’t look weird in dark mode.

```suggestion
  padding: 50px;
  background-color: #fff;
```

kleinfreund (commented):
A few notes on the docs site interaction with its dark theme for now.

src/app/services/components/service-traffic/README.md
```diff
@@ -0,0 +1,110 @@
+# ServiceTraffic
```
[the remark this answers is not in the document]
kleinfreund: I had this comment pending before https://github.com/kumahq/kuma-gui/pull/1807/commits/5c8c30df67903b361e979643385e1c3706dbe2df was pushed. Just leaving it here for reference.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficGroup.vue
```diff
@@ -0,0 +1,44 @@
...
+  &:is(.type-outbound) {
+    border-color: $kui-color-border-primary-weak;
+    background-color: $kui-color-background-primary-weakest;
+  }
+  &:is(.type-inbound) {
+    padding: 0;
+    border-radius: 0;
+    border: 0;
+  }
+}
```
kleinfreund, line 43: **Suggestion**: I don’t think `:is()` is necessary here.

```suggestion
.kong-card {
  &.type-passthrough {
    /* FIXME(jc): These colors is in the design but don't seem to be in @kong/design-tokens */
    border-color: #00D6A4;
    background-color: #ECFFFB;
  }
  &.type-outbound {
    border-color: $kui-color-border-primary-weak;
    background-color: $kui-color-background-primary-weakest;
  }
  &.type-inbound {
    padding: 0;
    border-radius: 0;
    border: 0;
  }
}
```

src/app/data-planes/components/data-plane-traffic/ServiceTrafficGroup.vue
```diff
@@ -0,0 +1,44 @@
...
+const props = defineProps<{
+  type: string
+}>()
+</script>
+<style lang="scss" scoped>
+.service-traffic-group :deep(.k-card-body) {
+  display: flex;
+  flex-direction: column;
+  gap: $kui-space-40;
+}
```
kleinfreund, line 22: **Suggestion**: Instead of overriding `.k-card-body`, I suggest adding a wrapping element around `<slot name="default" />` which these styles are attached to. This way we don’t depend on class names out of our control.

As an aside, for `<slot name="default" />`, you can always omit `name="default"` as it’s the default name.

src/app/data-planes/components/data-plane-traffic/DataPlaneTraffic.vue
```diff
@@ -0,0 +1,53 @@
...
+.service-traffic {
+  display: flex;
+  flex-direction: column;
+  gap: $kui-space-40;
+}
+.kong-card {
+  &:is(.header) {
+    border: 0;
+  }
+}
```
kleinfreund, line 27: **Question**: There’s a lot of KCard overriding going on in the components here. We override `padding`, in this instance `border`, often `border-color` and `background-color`. Does it really make sense to use KCard here? It seems to me the cards are sufficiently different that trying to stick to using KCard turns out more of a hindrance than a benefit.

src/app/services/components/service-traffic/README.md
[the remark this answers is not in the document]
johncowen: Resolving this one lemme know if it needs reopening

src/app/data-planes/components/data-plane-traffic/ServiceTrafficGroup.vue
johncowen: I removed all the places where I was using `.kong` CSS selectors or similar.

As for the  `<slot name="default" />` I don't think it matters either way, most of me thinks being explicit is good for when/if more slots get added but 🤷 . I didn't change it.

src/app/data-planes/components/data-plane-traffic/DataPlaneTraffic.vue
johncowen: Gonna resolve this one

src/app/common/data-card/DataCard.vue
```diff
@@ -0,0 +1,58 @@
...
+  display: flex;
+  gap: $kui-space-40;
+  align-items: flex-start;
+}
+.card {
+  flex-direction: column;
+}
+.title {
+  /* FIXME(jc): Usually this is 1.5 but that looks way to large */
+  line-height: 1.2;
```
johncowen, line 36: @kleinfreund do you have any input here? I looked a little deeper into it which resulted in https://github.com/kumahq/kuma-gui/pull/1816 (which is in draft as it still doesn't feel quite right to me I'm gonna check in with @adamdehaven before I go any further with that draft PR)

I do think everything should be 1.3 though, I'll make some more comment on that draft PR.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficGroup.vue
```diff
@@ -0,0 +1,45 @@
...
+.type-passthrough,
+.type-outbound {
+  padding: $kui-space-40;
+  border-radius: $kui-border-radius-40;
+}
+.service-traffic-group {
+  &.type-passthrough {
+    /* FIXME(jc): These colors is in the design but don't seem to be in @kong/design-tokens */
+    border-color: #00D6A4;
+    background-color: #ECFFFB;
```
johncowen, line 33: @kleinfreund do you know if I'm missing something here, or do you have any suggestions? I can check in with design-tokens folks if not.

src/app/common/data-card/DataCard.vue
adamdehaven: Really you should/could be using the SCSS variables from `@kong/design-tokens`

If you do, the `kui-font-size` and `kui-line-height` variables are meant to be used 1:1, meaning if you have `$kui-font-size-40` you would typically use `$kui-line-height-40` alongside it.

johncowen:
@kleinfreund just to check is there anything else here apart from my FIXMEs and the dark mode in the sandbox thing you mentioned originally?

src/app/data-planes/components/data-plane-traffic/ServiceTrafficGroup.vue
adamdehaven: If you're using colors not in the design system, they typically fall in two categories:

1. You're doing something custom; should you be? Maybe, maybe not.
2. The Figma design you're referencing hasn't been updating to utilize the tokens. Check with your designer

src/app/common/data-card/DataCard.vue
johncowen: Ah you're around! I wanted to chat to you about this along with the PR mentioned above, I know its not ideal. Umm let me think about what you said here a bit in the context of #1816 it might help with that.

Umm if you are only "around" but not "around" (😆 ) we can catch up early next week 👍  (otherwise ping me on Slack)

Oh also see the other FIXME here on the colors

src/app/data-planes/components/data-plane-traffic/ServiceTrafficGroup.vue
johncowen: Gotcha, let me check again but unless I'm misremembering the there was a green that was in the token set, but not with a semantic alias, so it looked like I couldn't get to it. let me look again and I'll come back to you.

(From what I remember the design also uses the non-semantic color name so `green-000` , 000 being whatever it was on the scale, again let me triple check)

johncowen: Here's the hexcode in design-tokens:

https://github.com/Kong/design-tokens/blob/c64b570fd649762bbbfb6937425852a18b663833/tokens/alias/color/index.json#L147-L149

and the Figma

Screenshot 2023-11-24 at 15 10 24

The border hex is the most important one. I guess we could just use a different color that exists in the semantic colors but it feels like since design are using these colors we should have access to them? I'm fine with that but we should also tell design not to use these.

---

I remember now I was unsure about the background one (which does exist) and I've added a commit to fill it out here. I think the not-terribly-important reason why I was unsure was: the color doesn't mean `success` it means 'green' i.e. we want to use 'green' for passthrough, not 'success' for passthrough. So if we were to expose the colors (see above) I would rather say 'green' over 'success' <- this one definitely isn't as important as not having the color at all.

kleinfreund: I think this is a case where it’s expected to have custom colors. We're not indicating something in a particular state/meaning the design system could know about here. We're arbitrarily assigning these greens to a card for passthrough traffic. Same thing for outbound traffic which happens to use colors present in the tokens.

Should these just generally use custom colors then?

johncowen: Hmm yeah quite possibly. We could even make a `--kuma-green-40` so at least we match up with the value/name in the design.

I think I would rather have access to `--kui-green-40` though, but I get why there's a reason to restrict people from using those (they should prefer semantic ones where possible). I think my main thing is, if design have access to it, and we need to implement the design, then we need access also.

> Same thing for outbound traffic which happens to use colors present in the tokens.

Also agree with this ^, although `primary` isn't as bad as using `success`

Lets see what @adamdehaven comes back with then

kleinfreund:
> @kleinfreund just to check is there anything else here apart from my FIXMEs and the dark mode in the sandbox thing you mentioned originally?

I’m mostly fine with this, I think. I think this thread should await an outcome: https://github.com/kumahq/kuma-gui/pull/1807#discussion_r1404423938

I’m not a fan of structural selectors like `.class > dl > div > dd`, etc. They create an unnecessarily tight coupling of styles with exact DOM tree structure that can be avoided, but you do you.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficGroup.vue
johncowen: Ok so that sounds like "please use the colors you have access to in the tokens"? i.e. we choose a different color from design-tokens that we do have access to?

johncowen:
> I’m mostly fine with this, I think. I think this thread should await an outcome: https://github.com/kumahq/kuma-gui/pull/1807#discussion_r1404423938

I've got some feedback on this, but left it as hex with a TODO, which I think is fine for the minute.

> I’m not a fan of structural selectors like .class > dl > div > dd, etc. They create an unnecessarily tight coupling of styles with exact DOM tree structure that can be avoided, but you do you.

Its only `.class > dl > div > :is(dt, dd)` i.e. any immediate dt, dd combinations (`dt`'s and `dd`s should always go in `dl > div`s. This is kinda similar to `ul` styling any immediate `li`s. Plus, I think we do this quite a lot (sometimes even without the `>`)

Let me know on this one, oh I also disabled dark mode.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficGroup.vue
adamdehaven: Just use the hex if you need to; Jason will be back next week and can ask him how to proceed if we need to expose some green tokens

.vitepress/theme/index.ts
kleinfreund: Resolving this following https://github.com/kumahq/kuma-gui/pull/1807#issuecomment-1828302119.

docs/frames/default.md
kleinfreund: Resolving this following https://github.com/kumahq/kuma-gui/pull/1807#issuecomment-1828302119.

.vitepress/theme/style.css
kleinfreund: Resolving this following https://github.com/kumahq/kuma-gui/pull/1807#issuecomment-1828302119.

kleinfreund (approved):

Issue #1565, opened 2023-09-27
Tweak logic for showing Envoy data (xds config, stats, clusters)


kleinfreund:
### Description

Blocked on https://github.com/kumahq/kuma/issues/5216.

Once https://github.com/kumahq/kuma/issues/5216 is implemented:

- Remove the logic for only querying Envoy data when an associated resource is online (handled in https://github.com/kumahq/kuma-gui/pull/1809)
  - This logic doesn’t make sense: we would make an additional request (to load ZoneOverview) in order to potentially avoid the request for the Envoy data (meaning we actually make two requests in the positive case)
  - This logic is broken because for ZoneIngress and ZoneEgress we don’t look at the corresponding ZoneOverview’s status but the respective ZoneIngreeOverview/ZoneEgressOverview
- Make sure we correctly display the error response information based on however https://github.com/kumahq/kuma/issues/5216 will be solved

In other words, we distinguish (in the backend) between the following scenarios:

- Resource is online and we get Envoy data
- Resource is offline and we get a 400 (or something) and tell people why they can’t see Envoy data
- Resource doesn’t exist. Standard 404

johncowen:
Just wanted to note here that we shouldn't be querying any Envoy API admin calls if the Zone is offline. From reading the description here I couldn't figure out if this issue meant we would still query the Envoy admin API if the zone was offline:

> Remove the logic for only querying Envoy data when an associated resource is online

We shouldn't query Envoy admin APIs if the zone is offline.

We could consider either:

- Showing warnings in the tabs themselves that we can't show you anything if the Zone is offline (which I think we kinda do now)
- Just don't show the tabs, and maybe showing a warning at the top of the overview saying you can see Envoy data once the zone comes online.

kleinfreund:
> Just wanted to note here that we shouldn't be querying any Envoy API admin calls if the Zone is offline.

Why?

slonka:
> Why?

This results in an error log on our side

```
stream to execute Stats operations is not yet connected
```

kleinfreund:
And it’s better to avoid this error at the cost of adding a new API call in the UI (calling /zones+insights so we know whether a Zone is online)?

slonka:
> And it’s better to avoid this error at the cost of adding a new API call in the UI (calling /zones+insights so we know whether a Zone is online)?

Don't we already have this info? If not then I guess it might be better to change the level of that log message. WDYT @lukidzi ?

[...]

PR #1809, opened 2023-11-23
chore: remove online check for envoy data

kleinfreund:
## Changes

Removes the online check from the EnvoyData component. It wasn’t correctly implemented for ZoneIngress and ZoneEgress where we would need to check for the status of the Zone not the ZoneIngress/ZoneEgress.

## Notes

- This is one half of #1565. The other half will be ensuring that we correctly surface information from the error responses added in https://github.com/kumahq/kuma/issues/5216.

### DIFF
--- a/src/app/common/EnvoyData.vue
+++ b/src/app/common/EnvoyData.vue
@@ -1,16 +1,6 @@
 <template>
   <div>
-    <KAlert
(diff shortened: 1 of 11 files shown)


johncowen (approved):
LGTM!

PR #1815, opened 2023-11-24
chore(dataplanes): tweak mocks

kleinfreund:
## Changes

chore: add tags mock utility

chore(dataplanes): inline mock utilities in networking utility

## Notes

- Related conversation: https://github.com/kumahq/kuma-gui/pull/1803#discussion_r1404323972

  We currently use `dataplaneNetworking` in five mocks (of which one isn’t strictly necessary because we don’t use the associated endpoint). For the remaining four mocks, I don’t want to repeat the contents of `dataplaneNetworking`. Ideally we had a generator method for an entire Dataplane object and an entire DataplaneOverview object. That would cut it down to two mocks (i.e. tolerable repetition) and the effect to see the entire object structure in one place. For now, `dataplaneNetworking` is a relatively good compromise because of how far up it appears in the `Dataplane` (on `networking`) and `DataplaneOverview` (on `dataplane.networking`) it appears in those objects.

### DIFF
--- a/src/test-support/FakeKuma.ts
+++ b/src/test-support/FakeKuma.ts
@@ -4,9 +4,6 @@ import deepmerge from 'deepmerge'
 import type {
   DataplaneNetworking,
   DataPlaneProxyStatus,
-  DataplaneGateway,
-  DataplaneInbound,
-  DataplaneOutbound,
   ServiceStatus,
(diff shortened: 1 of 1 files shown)


johncowen (approved):
This looks a bit better I think, thanks for doing this!

PR #1816, opened 2023-11-24
chore: Reduce line-height overwrites, add base font-size

johncowen:
I keep noticing several bits of weirdness to do with line-heights and font-sizes. This PR deals mainly with line-heights and centralizes them using relative units in our base stylesheet. I also noticed setting `--kui-line-height` also solves some issues mainly with k-cards.

The font-size addition was a late amend when I noticed some weirdness there also. At least for the moment, this amend means we can control our base font size, which is important for us seeing as we use `rems` quite a bit. I guess there are other places where we might change our `rem` usage to use `--kui-font-size` more, but still I guess there will be places where we might use `rem`.

This PR originally grew out of https://github.com/kumahq/kuma-gui/pull/1807#pullrequestreview-1748060631

I'm still a little unsure about setting `--kui-line-height-20` ourselves but given everything I'd rather do that now and have all our line-heights make sense, we can always undo this at a later date. All in all there are more things that make me want to do this than not want to do this.

### DIFF
--- a/src/assets/styles/_base.scss
+++ b/src/assets/styles/_base.scss
@@ -4,6 +4,26 @@
   box-sizing: border-box;
 }
 
+:root.kuma-ready {
+  // TODO(jc) ideally --kui-line-height would be set to $kui-line-height-20;
+  // $kui-line-height is currently 16px, but 1.3 gives the same relative height
+  // across font-sizing, which resolves all our line-height issues
+  --kui-line-height-20: 1.3;
+  // we set the base font-size for our rems here incase it ever gets set to
+  // something other than 16px
+  font-size: $kui-font-size-40;
+}
+
+body,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  line-height: var(--kui-line-height-20);
+}
+
 body {
   margin: 0;
   tab-size: 2;
@@ -11,7 +31,6 @@ body {
   font-family: $kui-font-family-text;
   font-weight: $kui-font-weight-regular;
   font-size: $kui-font-size-40;
-  line-height: 1.5;
 }
 
 blockquote,
@@ -38,7 +57,6 @@ h3,
 h4,
 h5,
 h6 {
-  line-height: 1.3;
   font-weight: $kui-font-weight-semibold;
 }
 
@@ -174,7 +192,6 @@ select {
   border: $kui-border-width-10 solid $kui-color-border;
   border-radius: 3px;
   font-size: $kui-font-size-40;
-  line-height: $kui-line-height-30;
   padding: 10px 30px 10px 13px;
   background-color: $kui-color-background;
   background-image: url('@/assets/images/chevron-down.svg?url');
--- a/src/assets/styles/_base.scss
+++ b/src/assets/styles/_base.scss
@@ -4,7 +4,7 @@
   box-sizing: border-box;
 }
 
-:root.kuma-ready {
+:root {
   font-size: $kui-font-size-40;
 }
 
@@ -95,6 +95,7 @@ ol {
     margin-top: $kui-space-50;
   }
 }
+
 p:empty {
   display: none;
 }
--- a/src/assets/styles/_base.scss
+++ b/src/assets/styles/_base.scss
@@ -5,12 +5,6 @@
 }
 
 :root.kuma-ready {
-  // TODO(jc) ideally --kui-line-height would be set to $kui-line-height-20;
-  // $kui-line-height is currently 16px, but 1.3 gives the same relative height
-  // across font-sizing, which resolves all our line-height issues
-  --kui-line-height-20: 1.3;
-  // we set the base font-size for our rems here incase it ever gets set to
-  // something other than 16px
   font-size: $kui-font-size-40;
 }
 
@@ -21,7 +15,7 @@ h3,
 h4,
 h5,
 h6 {
-  line-height: var(--kui-line-height-20);
+  line-height: 1.3;
 }
 
 body {
--- a/src/assets/styles/_forms.scss
+++ b/src/assets/styles/_forms.scss
@@ -12,7 +12,6 @@
 
 .form-card .card-content:not(.increase-specificity) {
   font-size: $kui-font-size-40;
-  line-height: 1.5;
 }
 
 .form>*+* {
@@ -72,7 +71,6 @@
   display: inline-flex;
   margin-bottom: $kui-space-40;
   font-size: $kui-font-size-30;
-  line-height: $kui-line-height-30;
   font-weight: $kui-font-weight-semibold;
 }
 
--- a/src/app/application/components/app-view/AppView.vue
+++ b/src/app/application/components/app-view/AppView.vue
@@ -127,13 +127,11 @@ onBeforeUnmount(() => {
   margin-bottom: 20px; /* 2rem */
 
   h1, h2, h3, h4, h5, h6  {
-    line-height: 36px;
     overflow: hidden;
     text-overflow: ellipsis;
   }
 }
 .app-view-title-bar h1 {
-  line-height: 1.3;
   font-weight: $kui-font-weight-semibold;
   font-size: $kui-font-size-80;
 }
--- a/src/app/common/DefinitionCard.vue
+++ b/src/app/common/DefinitionCard.vue
@@ -55,7 +55,6 @@ const props = withDefaults(defineProps<{
 .definition-card--vertical .definition-card-container {
   flex-grow: 1;
   font-size: $kui-font-size-60;
-  line-height: 1.5;
 }
 </style>
 
--- a/src/app/common/data-card/DataCard.vue
+++ b/src/app/common/data-card/DataCard.vue
@@ -39,8 +39,6 @@
   flex-direction: column;
 }
 .title {
-  line-height: $kui-line-height-20;
-
   font-weight: $kui-font-weight-semibold;
   font-size: $kui-font-size-60;
 }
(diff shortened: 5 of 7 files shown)


src/assets/styles/_base.scss
```diff
@@ -3,6 +3,20 @@
 *::after {
   box-sizing: border-box;
 }
+:root {
+  --kui-line-height-20: 1.3;
+}
```
johncowen, line 8: Whilst this seems to sort out all our problems / need for overwrites. It definitely doesn't feel right setting this here like this. Wondering what to do? For the moment I reused this in our own styles, but alternatively I can drop this "setting a variable we don't own" and move the overwrite on line 15 of _forms.scss up here so its altogether maybe?

Dunno still thinking myself, also would like guidance/ to hear from @adamdehaven when hes around.

src/assets/styles/_forms.scss
```diff
@@ -12,7 +12,6 @@
 
 .form-card .k-card-body:not(.increase-specificity) {
   font-size: $kui-font-size-40;
-  line-height: 1.5;
```
johncowen, line 15: This is the overwrite I refer to in the above comment

src/assets/styles/_base.scss
adamdehaven: We don't currently have standardized "base page styles" for native elements (e.g. headers, paragraph tags, etc.).

This is something coming in the future and will be exported from one of our existing packages (e.g. you could import a generic base stylesheet) but still TODO

johncowen: Umm, are these `--kui-line-height-20` vars meant to be set by the end user?

The reason I ask is that the `.k-card-body` below uses --kui-line-height-20` in its styles but as its unset currently and then defaults to something that doesn't look right, so we have to override it.

As soon as I set `--kui-line-height-20` to what it looks like our GUI uses as a standard/default line height, everything falls into place.

kleinfreund: > The reason I ask is that the `.k-card-body` below uses `--kui-line-height-20` in its styles but as its unset currently and then defaults to something that doesn't look right, so we have to override it.

That is intentional. The CSS custom properties are intended to be the application-level customization layer and the Sass variables are the defaults. This allows Kongponents styles to be customized without resorting to overriding CSS rule sets.

The issue with setting only one `--kui-line-height-XX` property is that it only affects uses of exactly this property. Any property with a number different from `XX` won’t be affected by this setting.

Could you provide us with some before/after screenshots of places where we need to change `--kui-line-height-20`, please?

kleinfreund: Additionally, I think the base font size should be set on `html` or `body`. Why add `.kuma-ready`?

johncowen:
@kleinfreund I think the majority of your inline comments can be addressed with this:

Did you see that `.k-card-body` uses `--kui-line-height`, therefore setting `--kui-line-height` to `1.3` means `.k-card-body` uses that, which means we no longer have to set line-heights for everything as if anything does inherit from that it will still be `1.3` i.e. the default for the application (which is also a relative unit and therefore just works in the majority of cases). This covers your suggestions I think and is the entire point of this PR.

---

> That is intentional. The CSS custom properties are intended to be the application-level customization layer and the Sass variables are the defaults. This allows Kongponents styles to be customized without resorting to overriding CSS rule sets.

Ok great!

> Could you provide us with some before/after screenshots of places where we need to change --kui-line-height-20, please?

Again see above for the point of the PR, but I think it's more the fact that we can remove all of the places where we resort to overriding the CSS by using the CSS custom property (as you've mentioned above). This PR sets the line-height in one place and then it generally works for the entire application without having to add more and more CSS overwrites throughout the application.

There are a few places where this does improve things visually here's one of them:

## Before

Screenshot 2023-11-30 at 14 34 57

## After

Screenshot 2023-11-30 at 14 35 13

I guess there are others, but the thing is I should barely need to think about it now.

> Additionally, I think the base font size should be set on html or body. Why add .kuma-ready?

In a HTML document `:root` is `html`, I added `kuma-ready` as we own that and I don't want to apply this rule until our application is ready.

I think I covered everything but lemme know if not.

kleinfreund:
Don’t get me wrong, I’m very much in favor of using relative line-heights because in the vast majority of cases, there isn’t a need for a more fine-grained control of line-height.

However, since we do use Kongponents and its styles are exclusively written with non-relative font-size/line-height pairings, we also have more constraints to take into account.

Specifically, I think setting _one_ `--kui-line-height-XX` property (and yes I did see that, I commented on it [here](https://github.com/kumahq/kuma-gui/pull/1816#discussion_r1410336970) and [here](https://github.com/kumahq/kuma-gui/pull/1816#discussion_r1410353312)) to change the absolute line-height of the `line-height-XX` token to a relative one is problematic. It breaks the scale between that and other line height tokens and partially defeats the purpose of using design tokens in this way. This is how the individual line height tokens would apply with those changes:

| line-height | value |
| ----------- | ----- |
| 10          | 12px  |
| 20          | 1.3   |
| 30          | 20px  |
| 40          | 24px  |
| 50          | 28px  |
| 60          | 32px  |
| 70          | 36px  |
| 80          | 40px  |
| 90          | 48px  |
| 100         | 56px  |

If we really, _really_ want to override the line height tokens, we should override all line height tokens so we don’t break the scale and risk any updates to that scale breaking our visuals. And I kinda don’t want to override any of those line height tokens.

Ultimately, if that’s how you want to handle this, I’m fine with it. It just really doesn’t seem like intended design token use.

---

> In a HTML document `:root` is `html`, I added `kuma-ready` as we own that and I don't want to apply this rule until our application is ready.

The point was very much about why `.kuma-ready` is there. It shouldn’t be. It doesn’t provide anything useful. The variable setting code show go into the variables file also.

kleinfreund:
> Do you think we should set the document font size before the application is ready?

Yes, 100%. There is no reason to have a set of base styles that don’t always apply. Base styles should always be set and use tag selectors so they apply to _all_ elements for that tag and can be easily overridden by more specific things. On that note, `html` is a preferable selector as it has lower specificity than `:root`.

adamdehaven:
For what it's worth, for MinK, the app shell handles hiding the host app so this REALLY isn't needed there. 

I'll also note that this thread will likely be mute next week as a lot of the classes you're targeting `k-card-body` as an example are changing or being removed. 

Once the reskin effort is complete, there should be **zero** style overrides on things like card text, etc. that live inside a component.

lahabana:
https://github.com/kumahq/kuma-gui/pull/1853 needs to happen first

johncowen:
> I'm still a little unsure about setting --kui-line-height-20

Looks like since https://github.com/kumahq/kuma-gui/pull/1853 we no longer need to do this 🎉

johncowen:
@adamdehaven @kleinfreund I wanted to resurrect this one now that the KCard amends have gone in.

I noticed since those amends went in things are much better, and this PR removes all the `line-height` overwrites we had. Instead we have 1 single `line-height` in the base stylesheet that is applied to `body` and headers. It is still a relative value, but that means we don't have to worry about `line-height` for anything that it non-kongponent.

I think the font-size 16 is no longer required but even when that issue was fixed upstream I was half "maybe we should keep this here for if it happens again" @adamdehaven will completely leave that one up to you whether you want me to remove or you think we should keep it in here.

Following that last point if we are leaving it in, I'd be fine using `html` and removing the `.kuma-ready` if I know that I'm not ever going to clobber any font sizes in the small period of time where our application is loaded but not yet `ready`. Overall it might be easier to say "just remove this font-size entirely, if the issue happens again, it happens again, we know where to fix it if it does" - I'm also ok with that.

If you could let me know either way on those points and we can get this one merged or closed.

So TLDR;

1. font-size remove entirely or leave.
2. If the answer to 1. is leave, then let me know on the selector options.
3. line-heights are we ok with what is here now?

kleinfreund:
> 1. font-size remove entirely or leave.

I reckon it’d be the most safe to always set a `font-size` on `html`.

> 2\. If the answer to 1. is leave, then let me know on the selector options.

I don’t think there is a point in differentiating between "HTML file is discovered and its styles are applied" and "We set .kuma-ready" when it comes to styles. That just delays when styles are applied and I can’t think of any reason why we’d want that.

We should apply `font-size` to the selector `html`.

> 3\. line-heights are we ok with what is here now?

I think so. Perhaps we can also reconsider setting some of those font-sizes in a form context, but we don’t have to do that now.

johncowen:
>I don’t think there is a point in differentiating between "HTML file is discovered and its styles are applied" and "We set .kuma-ready" when it comes to styles. That just delays when styles are applied and I can’t think of any reason why we’d want that.

The original issue came from the fact that the MinK loading template applied a font-size different to ours on `html` from what I remember. I'm guessing this was because of setting the font-size of something for the loader, or previous to our application being ready. If that is still necessary (or someone adds something again in the future), then we wouldn't clobber that with this rule due to the `.kuma-ready`

kleinfreund:
> The original issue came from the fact that the MinK loading template applied a font-size different to ours on `html` from what I remember. I'm guessing this was because of setting the font-size of something for the loader, or previous to our application being ready. If that is still necessary (or someone adds something again in the future), then we wouldn't clobber that with this rule due to the `.kuma-ready`

I think we established that this wasn’t intentionally set

> Which makes me want to resolve the `kuma-ready` thing, as if we are removing it then I think that we might want the higher specificity of `:root`?
> 
> Anyway, let me know if you follow why I think the `kuma-ready` thing is probably a good idea.

I don’t think we _want_ the higher specificity, but it probably won’t cause issues, so I’m fine with using `:root`.

adamdehaven:
You don't need kuma-ready. 

Root should be fine; anything more specific on a global level might conflict with Kongponents or other components which isn't desired.

johncowen:
Cool thanks all, let me remove the `kuma-ready` but leave the `:root` I think that was the consensus here

PR #1818, opened 2023-11-27
chore(services): fix single resource mocks

kleinfreund:
Fixes an issue with not correctly determining the service type in the single resource type mock for services.


johncowen (approved):
Not totally sure what the fix was here but LGTM

kleinfreund:
> Not totally sure what the fix was here but LGTM

This logic was wrong:

```
  let serviceType
  if (name.includes('gateway_builtin')) {
    serviceType = 'gateway_builtin'
  } else if (name.includes('gateway_delegated')) {
    serviceType = 'gateway_delegated'
  } else {
    serviceType = fake.kuma.serviceType()
  }
```

We weren’t considering names that include internal or external.

PR #1832, opened 2023-11-28
chore(deps): ignore Kongponents in Dependabot

kleinfreund:
## Changes

Ignores the Kongponents updates in Dependabot for the time being. Dependabot currently tries to update Kongponents’ alpha prerelease to PR prerelease versions which we definitely need to prevent.

## Notes

- See example PR we don’t want: https://github.com/kumahq/kuma-gui/pull/1831

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,11 @@ updates:
     open-pull-requests-limit: 10
     labels:
       - "dependencies"
+    ignore:
+      # TODO: Remove this once Kongponents v9 is used from a mainline release again (i.e. neither alpha or beta).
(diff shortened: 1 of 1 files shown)


kleinfreund:
> Although, before doing that, I'm kinda interested to see if its possible to somehow pin to the alpha tag that kongponents are using to publish updates to the alpha. If we can, that would mean we would only get updates published to the alpha tag, and not the pr releases, wdyt?

Were this an option, I would update the dependabot configuration, but there isn’t anything for that degree of control. The only thing that is possible for npm packages is pinning to certain numbers _in the semantic version_ (e.g. `9.x`) or to ignore certain types of updates (major, minor, patch).

kleinfreund:
> Lets make an issue for re-enabling this once we come out of alpha?

https://github.com/kumahq/kuma-gui/issues/1840

kleinfreund:
So evidently this didn’t work: https://github.com/kumahq/kuma-gui/pull/1848. Dependabot even reads the setting as intended:

> updater | 2023/11/30 15:43:29 INFO <job_756546464>   9.x - from .github/dependabot.yml

:/

PR #1834, opened 2023-11-28
refactor(dataplanes): move getTags into data transformation file

kleinfreund:
Draft PR. Includes https://github.com/kumahq/kuma-gui/pull/1814 because it was branched off of that PR’s feature branch.

---

Moves the `getTags` dataplane utility into its data transformation file.


src/app/data-planes/data/index.ts
```diff
@@ -52,3 +53,28 @@ export function getStatusAndReason(dataplaneOverview: DataplaneOverview): { stat
...
+  const { gateway, inbound } = dataplaneOverview.dataplane.networking
+
+  if (inbound) {
+    tags = inbound
+      .flatMap((inbound) => Object.entries(inbound.tags))
+      .map(([key, value]) => key + separator + value)
+  }
+
+  if (gateway) {
+    tags = Object.entries(gateway.tags).map(([key, value]) => key + separator + value)
```
johncowen, line 69: Do you have a reason why you would prefer/change these to use addition here?

If not, could we switch these back to use backticks/template literals instead of addition, especially seeing as this is what it was using before.

johncowen: This looks like addition to me, these are strings, not numbers. We should use template literals always for joining strings, for one it prevents `valueOf` being used instead of `toString`.

We shouldn't get into the habit of thinking addition for string joining is ok and there is no reason to change this here.

kleinfreund: I’ve changed it because it really doesn’t matter, but please consider whether this kind of suggestion is really worth the time spent. The plus operator is extremely well-established to do string concatenation in JavaScript and especially since we’re dealing with a pretty well-typed code base, we actually know that none of the values here can even be numbers in theory.

Confusing an expression involving strings and the plus operator with number addition is not a problem in our code base because we’ve established better knowledge about the values at play.

johncowen (approved):

PR #1839, opened 2023-11-29
chore(dataplanes): Add data source for traffic retrieval

johncowen:
Adds a new `DataSource` `source` (in application code) for retrieving `stats` endpoint and parsing it into something we can use for the `DataPlaneTraffic` component.

The aim of this PR is to get something we can use in the application after this PR is merged, so I want to figure out what properties we should be using for the rendered stats here. For the moment I did what I'd understood was correct but then it didn't seem to show me anything, so I changed some things so that I had some numbers appearing. From what I understand this all depends on exactly what it is we want to show. We should be able to decide this and just tweak some words to show what we want.

Notes:

- I used an entirely different DataSource URI, this may or may not be merged into the one we currently use in the application, I kinda wanted to avoid touching application code for the moment.
- I commented out the stats in the title blocks (I'll note inline). I really want to get the names of the stats we want to be using here right before we finish these up.
- The mock stats I'm using are from an actual install with some requests thrown at it, so it should at least show me something other than zeros.

Lastly, please ignore the lint failure for the moment. I can't merge without sorting it out, but I don't want to talk about that before getting this PR reviewed.

Once we get this merged, it should just be a case of copying this over to the application we can then iterate on it from there.

Screenshot 2023-11-29 at 15 50 26

Also see https://github.com/kumahq/kuma-gui/pull/1807


src/app/data-planes/components/data-plane-traffic/DataPlaneTraffic.vue
```diff
@@ -6,7 +6,7 @@
       <template #title>
         <slot name="title" />
       </template>
-      <slot name="data" />
+    <!-- <slot name="data" /> -->
```
johncowen, line 9: Here's the bit I purposefully commented out for now

kleinfreund (approved):
Kind of just want to get this in and spend more time reviewing the result once this UI is hooked up to the application.

The linter error should be relatively easy to address. It looks like it’s about CSS rule set ordering.

johncowen:
I added a small amend for the CSS linter here, will wait a few minutes and then go ahead and merge if there are no further comments.

PR #1853, opened 2023-11-30
refactor: update KCard usage in Kongponents v9

kleinfreund:
Please refer to https://alpha--kongponents.netlify.app/guide/migrating-to-version-9.html#kcard for breaking changes in KCard.

---

Removes the `<template #body>` and corresponding `</template>` tags from all uses of KCard as its `body` slot was removed.


src/app/common/data-card/DataCard.vue
```diff
@@ -2,18 +2,16 @@
...
+    <dl>
+      <div class="card">
+        <dt class="title">
+          <slot name="title" />
+        </dt>
+        <dd class="body">
+          <slot name="default" />
+        </dd>
+      </div>
+    </dl>
```
johncowen, line 14: CSS needs adding to set these font-sizes back to the smaller `-20` (I think its `-20`)  now that kcard doesn't do that

kleinfreund: Shouldn’t those cards use the same font size as KCard?

johncowen: Not according to the designs I don't think, the size for the data here is smaller than normal body copy

kleinfreund: I placed an override in the component. Worth checking with design whether the changes to KCard should be reflected in the design for these components or not.

kleinfreund:
I’ve updated the PR now to install the correct version of Kongponents.

Issue #1264, opened 2023-07-28
Add data correction/reshaping in a data layer separate from the view layer


johncowen:
### Description

1. Move data reshaping/logic type functions over to a "data layer". These should be within a `module-name/data` folder (for a rough example see https://github.com/kumahq/kuma-gui/tree/467e78de210f481c2b47d4002d1d1a357838f6d9/src/app/zones/data)
2. Avoid undefined/null arrays and potentially objects. What to do with arrays is clearer, objects could be less clear, see accompanying MADR on the overall approach for guidance. For other primitive values again try to use best judgement using the MADR for guidance. (for a rough example see https://github.com/kumahq/kuma-gui/blob/467e78de210f481c2b47d4002d1d1a357838f6d9/src/app/zone-ingresses/data/index.ts)
3. Refactor and combine above data reshaping/logic type functions so we are reusing what we can. What exactly this involves will depend on lots of things and will be case by case.
4. Move as many calls to these "data layer" functions into the `sources.ts` files, so the data is reshaped as soon as we receive it.

Depending on the situation the above can be done in a different order depending on the module/whatever is easiest and don't have to all be in the same PR.

- [x] DataPlanes
  - #1871
  - #1909
  - #1908
- [x] ControlPlanes
  - #1914
- [x] Diagnostics
  - #1914
- [x] Onboarding (has no data sources)
- [ ] Subscriptions
  - https://github.com/kumahq/kuma-gui/pull/1915
- [x] Meshes
  - #1912
- [x] Policies
  - #1888
- [x] Services
  - #1886
- [x] Zone Egresses
  - https://github.com/kumahq/kuma-gui/pull/1869
- [x] Zone Ingresses
  - https://github.com/kumahq/kuma-gui/pull/1867
- [x] Zones 
  - https://github.com/kumahq/kuma-gui/pull/1854
- [ ] Final review incase there are any stray bits or missed bit once the above list of modules is complete.

lahabana:
Let's do that on a per module basis

johncowen:
I updated the description here and included a rough approach plus a list of modules we'll need to cover. We can use that to track progress.

PR #1854, opened 2023-11-30
chore(zones): Zone data layer pt1

johncowen:
Reduces the majority of our `Zone` related reshaping functions

See https://github.com/kumahq/kuma-gui/issues/1264


src/app/zones/data/index.ts
```diff
@@ -1,36 +1,87 @@
...
+export const ZoneInsight = {
+  fromObject: (item?: PartialZoneInsight): ZoneInsight | undefined => {
+    // if item isn't set don't even try augmenting things
+    return isSet<PartialZoneInsight>(item)
+      ? ((item) => {
+        // figure out the connectedSubscription by looking at the connectTime
+        // and disconnectTime of the last subscription
+        const connectedSubscription: KDSSubscription | undefined = !Array.isArray(item.subscriptions)
+          ? undefined
+          : [item.subscriptions[item.subscriptions.length - 1]].find((item) => item.connectTime?.length && !item.disconnectTime)
```
kleinfreund, line 42: **Issue**: We should use the last subscription. It’s the most recent one. We used to try and find the first one that has no `disconnectTime`, but that’s not necessary. 

And let’s avoid this "array with one entry plus find/every/some method" pattern just to get to a one-line expression. The following is much, much more clear:

```suggestion
        const subscriptions = item.subscriptions ?? []
        const connectedSubscription: KDSSubscription | undefined = subscriptions[subscriptions.length - 1]
```

src/app/zones/data/index.ts
```diff
@@ -1,36 +1,87 @@
...
+
+        // using the connectedSubscription find the config for the zone if it exists, is valid JSON and is not null and
+        // turn it into an object
+        const config: Record<string, unknown> = (() => {
+          const str = isSet<string>(connectedSubscription?.config) ? connectedSubscription.config : '{}'
+          try {
+            return JSON.parse(str)
+          } catch (e) { }
+          return {}
+        })()
```
kleinfreund, line 52: **Issue**: We shouldn’t expect the possibility that `JSON.parse(str)` can throw. Instead, we should rely on the assumption that, if the ZoneInsight was updated following an applied config, that that config was parseable in the first place. If not, I’d assume the config to not make it into a KDS subscription.

On top of that, hiding the error here is dangerous because we won’t know that the error occurred. Much better to let this cause an exception that we can actually fix (because it’s not an expected application scenario).

```suggestion
        const config: Record<string, unknown> = isSet<string>(connectedSubscription?.config) ? JSON.parse(connectedSubscription.config) : {}
```

kleinfreund (changes requested):

src/app/zones/data/index.ts
johncowen: I added a `console.error` here which I think addresses your concern about us not knowing if this throws, but I guess we shouldn't expect that. But if it does happen, `config` would just be a `{}` and things would continue on as normal.

src/app/zones/data/index.ts
johncowen: Are you absolutely sure the last subscription always has a `connectTime` and never has a `disconnectTime`? At the very least the types say that `connectTime` can be `undefined`? In which case I would guess it hasn't connected?

src/app/zones/routes.ts
```diff
@@ -34,13 +34,6 @@ export const routes = (
                 module: 'zone-cps',
               },
               component: () => import('@/app/zones/views/ZoneListView.vue'),
-              children: [
-                {
-                  path: ':zone',
-                  name: 'zone-cp-summary-view',
-                  component: () => import('@/app/zones/views/ZoneSummaryView.vue'),
-                },
-              ],
```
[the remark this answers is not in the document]
johncowen: I think this is good to resolve, lemme know if not.

src/app/zones/data/index.ts
```diff
@@ -1,36 +1,87 @@
...
+export const ZoneOverview = {
+  fromObject: (item: PartialZoneOverview): ZoneOverview => {
+    const insight = ZoneInsight.fromObject(item.zoneInsight)
+    const zone = Zone.fromObject(item.zone)
+    return {
+      ...item,
+      zoneInsight: insight,
+      zone,
+      // first check see if the zone is disabled, if not look for the connectSubscription
+      state: !zone.enabled ? 'disabled' : typeof insight?.connectedSubscription !== 'undefined' ? 'online' : 'offline',
```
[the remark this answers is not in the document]
johncowen: We use `status` elsewhere for an object name not a string FSM state. This is a string FSM state, not an object of properties/data/infinite state. `state` describes this characteristic very well and would not be confused with the `status` object elsewhere. How big of an issue is this?

src/app/zones/data/index.ts
```diff
@@ -1,36 +1,89 @@
...
+        // figure out the connectedSubscription by looking at the connectTime
+        // and disconnectTime of the last subscription
+        const connectedSubscription: KDSSubscription | undefined = !Array.isArray(item.subscriptions)
+          ? undefined
+          : [item.subscriptions[item.subscriptions.length - 1]].find((item) => item.connectTime?.length && !item.disconnectTime)
+
+        // using the connectedSubscription find the config for the zone if it exists, is valid JSON and is not null and
+        // turn it into an object
+        const config: Record<string, unknown> = (() => {
+          const str = isSet<string>(connectedSubscription?.config) ? connectedSubscription.config : '{}'
```
kleinfreund, line 47: **Comment**: I noticed an issue with this utility. Take the following reduced example:

```ts
const array: { prop?: string }[] = []
const element = array.at(0)
const test = isSet<string>(element?.prop) ? element.prop : null
//                                          ^^^^^^^ 'element' is possibly 'undefined'.
```

When the TypeScript compiler infers that `element` has a union type like `Interface | undefined`, the `isSet` utility doesn’t function as a complete type guard anymore.

Note that this just happens to not be surfaced when using `array[0]` because by default, TypeScript considers `array[i]` safe and returns `T`. With `noUncheckedIndexedAccess` enabled, it would be `T | undefined` and the same error would happen, I think.

johncowen: 🤔 

`str` is always a `string` here, I believe?
Also, if `connectedSubscription?.config` is a string, then we access it as `connectedSubscription.config` because we know its set?

Could you explain a little more what the issue is here, or is this just a comment?

src/app/zones/data/index.ts
johncowen: From a conversation offline:

1. Whilst connectTime in our frontend has always been marked as potentially being unset. `connectTime` is guaranteed to be set, so we should amend that throughout the application (and therefore after this PR)
2. It is not guaranteed that the last subscription doesn't have a `disconnectTime`, all subscriptions can have a disconnectTime when the CP disconnects all the subscriptions (from what I understand from the offline convo with backend folks)

Let me know if I understood this wrong ^

src/app/zones/data/index.ts
johncowen: I'm guessing I can resolve this, let me know if not

src/app/zones/data/index.ts
johncowen: Also to be clear, I'm keeping the code using the same style that it was using before. I personally think this is clearer than your suggestion and using less `?-y` things. I may change this in pt2 of this (a follow up PR though), but I would probably still use `.find` seeing as it returns `undefined` and therefore `connectedSubscription` is less likely to be set to `null`.

With this comment and my previous comment, I'd like to resolve this, let me know.

src/app/zones/data/index.ts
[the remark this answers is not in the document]
kleinfreund: Okay, if you insist, then let’s call it state. Can you please create an issue to rename this to state everywhere then? We call it status virtually everywhere and introducing a different name here will be confusing over time.

src/app/zones/data/index.ts
kleinfreund: > 2\. It is not guaranteed that the last subscription doesn't have a `disconnectTime`

Just to be explicit: this is normal and should be expected. The last subscription _may_ have a disconnect time; all prior subscriptions _must_ have a disconnect time.

---

Feel free to resolve this. I really think this style of expression is problematic and unnecessarily complex, and it should be changed. I hope you do this in that follow up. At the very least, please consider the following suggestion: should the subscriptions ever come in as an empty array, this code will cause an exception and using `Array.prototype.at` exposes the issue to the type system (see explanation below).

```suggestion
        const connectedSubscription: KDSSubscription | undefined = !Array.isArray(item.subscriptions)
          ? undefined
          : [item.subscriptions.at(item.subscriptions.length - 1)].find((item) => item.connectTime?.length && !item.disconnectTime)
```

---

For the sake of explaining why I think it’s problematic:

> Also to be clear, I'm keeping the code using the same style that it was using before. I personally think this is clearer than your suggestion and using less `?-y` things. I may change this in pt2 of this (a follow up PR though), but I would probably still use `.find` seeing as it returns `undefined` and therefore `connectedSubscription` is less likely to be set to `null`.

I do really feel very strongly about this and want it to be changed here.

The code in the PR currently uses this expression:

```ts
const connectedSubscription: KDSSubscription | undefined = !Array.isArray(item.subscriptions)
  ? undefined
  : [item.subscriptions[item.subscriptions.length - 1]].find((item) => item.connectTime?.length && !item.disconnectTime)
```

This expression is semantically rather complex and hides a possible bug:

First: `item.subscriptions[item.subscriptions.length - 1]` appears to be of type `KDSSubscription` but should more accurately appear to be of type `KDSSubscription | undefined`. That’s because we don’t have `noUncheckedIndexedAccess` enabled. Replacing it with `item.subscriptions.at(item.subscriptions.length - 1)` illustrates the problem with that: suddenly `item` in the `find` call is `KDSSubscription | undefined` and `item.connectTime` becomes an error.

This code would lead to an exception if `item.subscriptions` would be an empty array. Currently that wouldn’t happen, but that might change very easily.

Second: With `[item.subscriptions.at(item.subscriptions.length - 1)].find((item) => item.connectTime?.length && !item.disconnectTime)` we’re creating an array of finite length 1 to then call `Array.prototype.find` just so we can leverage the fact that `find` returns `undefined` when it doesn’t find a match. That’s unnecessary.

We could make all of that extremely transparent with the change I suggested:

```suggestion
        const lastSubscription = (item.subscriptions ?? []).at(-1)
        const connectedSubscription = lastSubscription && !lastSubscription.disconnectTime ? lastSubscription : undefined
```

It also maintains that the "empty value" is `undefined` and performs the check for the connected subscription in a direct manner without utilizing `find` on a one-element array.

> using less `?-y` things

It uses the exact same amount of ternary expressions (1). Also, ternary expressions are great because they make it more obvious what the returned value can be. The one-element array and find style makes that needlessly obscure.

kleinfreund:
As a general remark, could you please add unit tests for the data transformation steps in the zones module? As far as I can tell, hardly any of it is currently covered with unit tests.

src/app/zones/data/index.ts
kleinfreund: VS Code shows it. Not sure how to reproduce it another way.

src/app/zones/data/index.ts
johncowen: I just traced back a few other things mentioned here

> This expression is semantically rather complex and hides a possible bug:
> This code would lead to an exception if item.subscriptions would be an empty array. Currently that wouldn’t happen, but that might change very easily.

https://github.com/kumahq/kuma-gui/blob/e490ea4c0f84a30e040fa2008e6283caabc811c0/src/types/index.d.ts#L46

`subscriptions` here is always set according to our types?

The only way this could have been `undefined` previously to this PR is because `zoneIngress` could be `undefined`, but in this PR we check for that before even getting to this code, so that case can't happen here. i.e. As things stand right now, `subscriptions` cannot be `undefined`.

If anything we need to remove the `Array.isArray` check, but ideally I wanted to make sure the rest of the areas we need this sort of thing was looked at before a final nip and tuck.

As mentioned above I have a PR for zone-ingress coming up, which currently has subscriptions as a child of an `any` definition, this is using `Array.slice` and `find` to always use an array and return an `item | undefined`.

---

🤔  I'm going to move this approach over here now rather than later as I'm "pretty sure" that it will end up being like this throughout the application, plus bringing this over here now instead of later I think resolves your issue. Please note tho this amend will continue to have a `Array.isArray` even though `subscriptions` here is always supposed to be an array, and it 'may' change again later depending on what the rest ends up looking like.

Resolving this one, but let me know if I need to reopen.

johncowen:
Ok on the unit tests, ideally I wanted to submit this separately (partly why this PR is a pt1):

https://github.com/kumahq/kuma-gui/commit/13c016524e7cb7adcd8459d5b337043ef528a197

Happy to add it on here, but I'd still rather do it separately.

---

There are several things in the above which could be talking points but here's a quick/rough list of things I'm trying to achieve:

1. Avoid/prevent us writing to the API response objects directly. We may want to make all these readonly once we get to autogenerated, in the meantime at least I'm using `Object.freeze` but I don't want to use that in the application.
2. Use our mocks to generate testing objects for us. They are currently our single source of truth for our mock data.
3. Continue not returning our typed entities from the mocks
4. Decide what to test in the cases where we assume things/types are correct but sometimes aren't because of reasons
5. Avoid feeling like I'm writing tests for the sake if it (possibly failed a bit here). I really want to avoid the trap of overdoing tests. There's probably lot to say here so I think the best thing to say is its always a balance and I'd rather the effort on this was trying to strike that balance rather than going too far one way or the other.
6. Do something that is consistent for unit testing across our data layer and easy to write.

Some of the above depends on doing more of the data layer work to ensure that this sort of stuff is flexible enough, yet as easy as possible to use.

Seeing as previously:

1. There have never really been any unit tests for any of this
2. We have plenty of e2e tests now that cover the majority of this

...and given that  all of the tests in that commit currently pass but the code may extend the lifetime of this PR, I'd be inclined to stick to my plan of PRing this separately.

Let me know what you think, i.e. start adding this here, or do it as a follow up.

src/app/zones/data/index.ts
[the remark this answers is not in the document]
johncowen: Can you be a little more specific here, I'm guessing you don't mean the backend properties, seeing as we can't really change those?

I think as we move forwards adding our 'frontend' properties to our 'backend' types I'll add a `state` on another few entities, but seeing as this is the first instance of us having these 'frontend properties' theres nowhere to change, I'll just make these new `state` props as I go?

I wondered of you meant things like `ResourceStatus` but they can and do show more than just the `state`, so status is kinda fine for those. Anyway lemme know, i can certainly add an issue if I know exactly what we need to change.

I just want to point out again that `state` is meant to represent a single finite state, not a grouping of properties/data.

---

Tiny little note whilst looking at this I think `StatusInfo` isn't used anymore so I'll make a PR to delete that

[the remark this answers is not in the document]
kleinfreund: I mean the stuff we virtually everywhere call status. For example, the stuff we pass to StatusBadge, the component, itself, etc.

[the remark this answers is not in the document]
johncowen: Ok honestly I think `StatusBadge` comes into what I said about `ResourceStatus` but I kinda don't mind either way, feel free to make an issue if you want to go ahead and change those thing I'm not that fussed.

Just to underline again, the new `state` frontend property is specifically to differentiate between the finite state of certain entities and the places where sometimes `status` is a string `state` and sometimes its a bag of properties.

From now on, on all our entities that have a `state`, `state` is the finite state of the thing, and the `status` could be anything.

kleinfreund:
I’m fine with you adding tests in a follow up, but I really think we should have a rather fine-grained coverage of the data transformations happening during data sourcing. Our Cypress tests, while covering a lot of it incidentally, aren’t a great tool to ensure we actually implement this stuff correctly (and prevent regressions reliably). The removal or change of a Cypress test could easily remove the incidental coverage of the data transformations.

This is a situation where redundancy in test coverage is almost unavoidable and not something we should fight.

And as far as "overdoing it with tests" goes, we really, really don’t have this problem for data transformers right now as we’re just introducing the mechanism. There is no need to be careful when our _direct_ test coverage of that mechanism is so low.

johncowen:
> I’m fine with you adding tests in a follow up

Great we can discuss over in that then.

Is there anything outstanding here?

src/app/zones/data/index.ts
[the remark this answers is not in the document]
kleinfreund: I’m confused about this now. In this PR, you motivated the change to call what we currently call "status" virtually everywhere "state". Personally, I don’t like that because (1) status is already an appropriate name for the concept and (2) state isn’t a better name (it’s more generic).

You want to distinguish a scalar status representation from a representation as an object, but we don’t have any instances where a resource’s status is represented as anything but a simple string. We have Zones (online, offline, disabled), Dataplanes (online, offline, partially degraded), and Services (online, offline, partially degraded, not available). Only in the case of Dataplanes do we accompany the status with a `reason` array to further clarify a certain status. But even in that case, the status is a simple string and not a dictionary.

If we rename this concept in our application, it follows we shouldn’t just do it in some places. That’s why I asked you to create a ticket for this. I don’t want to call this "state" here and then "status" everywhere else. Then we could just keep the name "status" here and it would be less confusing.

---

I previously said I’m fine if you think this is the right thing to do, but I’m not fine with using two different names for the status concept when we currently only have the one which is completely appropriate.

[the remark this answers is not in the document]
johncowen: What about `status` here?

https://github.com/kumahq/kuma-gui/blob/ec0bb869c7dcd3fdf9241215935c7dfccecfaac2/src/types/index.d.ts#L41

and here

https://github.com/kumahq/kuma-gui/blob/ec0bb869c7dcd3fdf9241215935c7dfccecfaac2/src/types/index.d.ts#L178

?

If you're fine with this, lets stop here. If it ends up being a bad choice we can come back and change.

This convo is resolved so lets leave it at that.

kleinfreund (approved):

PR #1855, opened 2023-12-04
refactor(dataplanes): use new _rules API

### LINKED ISSUE #1838: Migrate to new _rules API
### Description

For inspect policies. We should use the newer API that scales better and work in global.

We can also discuss design a little if we review this as some outputs are a bit better.

https://github.com/kumahq/kuma/blob/bf05cb9456e0456e70ed8769dc88c18ad75e903c/api/openapi/specs/api.yaml#L20-L47

kleinfreund:
Changes the dataplane policies view to only show the "Policies" section in non-global mode.

Migrates to using the new `_rules` API and updates StandardDataplanePolicies accordingly.

Resolves #1838.


features/mesh/dataplanes/DataplanePolicies.feature
```diff
@@ -1,61 +1,339 @@
...
+      Then the "$policies-view" element contains "__rule-matches-hash__:waFsoIISmZDTfFWYqxvY265/GASHYEWvHQTwMh/bpuU= and kuma.io/service:backend_kuma-demo_svc_3001"
+      And the "$policies-view" element contains "!__rule-matches-hash__:waFsoIISmZDTfFWYqxvY265/GASHYEWvHQTwMh/bpuU= and kuma.io/service:backend_kuma-demo_svc_3001"
+
+    Scenario: Policies tab has expected content (MeshTimeout with from & to rules)
+      Given the environment
+        """
+        KUMA_DATAPLANEPROXYRULE_ENABLED: false
+        KUMA_DATAPLANERULE_COUNT: 1
+        KUMA_DATAPLANETORULE_COUNT: 3
+        KUMA_DATAPLANEFROMRULE_COUNT: 1
```
johncowen, line 132: Ouch these names read really awkwardly, umm 🤔 , I have some questions as I'm not that familiar with the data here:

1. Do we ever test something where we absolutely need the TO and FROM rules to be different?
2. Wondering if we can just use `KUMA_RULE`, I know we have different sort of rules, but I don't think we'll ever need to worry about trying to set the amount of RULES for dataplanes and accidentally setting it for roles also? Definitely still wondering about this one, I'm not sure what I think either.
3. If we want differently named _RULES (see above) are these POLICY_RULES? Maybe not.

I think all in all we've unfortunately not used `_` in other places, but I think the original thinking was to not use `_` for model/entity names but to use `_` for PARENT_CHILD things which would make this read better overall (so `KUMA_FAULTINJECTIONS` but `KUMA_DATAPLANE_RULES`)

Anyway as you can see from this comment I'm not entirely sure myself here, and its just a naming thing, but still, these are really hard to look at. Let me know what you think on any of the above before we decide to do anything about it, if anything

(If anything above isn't clear please ask)

features/mesh/dataplanes/DataplanePolicies.feature
```diff
@@ -1,61 +1,339 @@
...
+                  origin:
+                    - mesh: default
+                      name: default
+                      type: MeshTimeout
+                  conf:
+                    connectionTimeout: 2s
+                    idleTimeout: 20s
+                    http:
+                      requestTimeout: 10s
+        """
```
johncowen, line 213: Do we need to mock out all of this data?

src/app/data-planes/components/PolicyTypeEntryList.vue
```diff
@@ -100,15 +100,26 @@
...
-import { PolicyTypeEntry, TableHeader } from '@/types/index'
+import type { LabelValue, TableHeader } from '@/types/index.d'
+import type { RouteLocationNamedRaw } from 'vue-router'
+
+export type PolicyTypeEntryConnection = {
+  sourceTags: LabelValue[]
+  destinationTags: LabelValue[]
+  name: string | null
+  origins: Array<{ name: string, route: RouteLocationNamedRaw }>
+  config: string | null
```
johncowen, line 116: I know we haven't got to moving policy type things over to our data layer yet, but is there a straight forward alternative to `null`? If not don't worry this will no doubt be changed pretty soon.

src/app/data-planes/components/RuleEntryList.vue
```diff
@@ -19,87 +19,74 @@
...
-            :headers="tableHeaders"
+            :class="{
+              'policy-type-table--with-matchers': props.showMatchers,
+            }"
+            :fetcher="() => ({ data: ruleEntry.rules, total: ruleEntry.rules.length })"
+            :headers="[
+              ...(props.showMatchers ? [{ label: 'Matchers', key: 'matchers' }] : []),
+              { label: 'Origin policies', key: 'origins' },
+              { label: 'Conf', key: 'config' },
+            ]"
```
johncowen, line 30: Not looking too much into it myself I wondered why this couldn't use AppCollection? I guess there might be more reasons apart from the fact that the code being changed here came before we had AppCollection?

src/app/data-planes/components/RuleEntryList.vue
```diff
@@ -110,34 +97,33 @@
...
-    required: false,
-    default: 'entry-list',
-  },
-
-  ruleEntries: {
-    type: Object as PropType<RuleEntry[]>,
-    required: true,
-  },
+import type { InspectRuleMatcher } from '@/types/index.d'
+import { useI18n } from '@/utilities'
```
johncowen, line 105: Could you import this from `@app/application` please? `@/utilities` will cease to exist pretty soon.

src/test-support/mocks/fs.ts
```diff
@@ -93,11 +93,11 @@ export const fs: FS = {
   '/meshes/:mesh/dataplanes/_overview': _21,
   '/meshes/:mesh/dataplanes/:name': _19,
   '/meshes/:mesh/dataplanes/:name/_overview': _22,
+  '/meshes/:mesh/dataplanes/:name/_rules': _128,
```
johncowen, line 96: Self-note: I had a script to auto generate this file originally, I must remember to pull that out and bring it over here.

johncowen (commented):
I took a very quick look at this (thanks for the info in the description btw), and I had a few initial questions which will help out when I review this proper, lmk on those.

features/mesh/dataplanes/DataplanePolicies.feature
kleinfreund: > 1. Do we ever test something where we absolutely need the TO and FROM rules to be different?

Very much so. The meaning changes if for what in the test scenario is expected to be 2 to rules and no proxy and from rules to suddenly have from rules, for example. So we do need to control that on the level of "rule types", not just a "KUMA_RULE".

> 2\. Wondering if we can just use `KUMA_RULE`, I know we have different sort of rules, but I don't think we'll ever need to worry about trying to set the amount of RULES for dataplanes and accidentally setting it for roles also? Definitely still wondering about this one, I'm not sure what I think either.

We could consider removing the `DATAPLANE` prefix, but there is in fact the potential for clashes as rules also exist in an RBAC context, so I don’t think that’s worthwhile.

> 3\. If we want differently named _RULES (see above) are these POLICY_RULES? Maybe not.

Not sure I understand the question completely. The association of those env entries (and the rest of the code) with dataplanes is partially arbitrary: it could also belong to policies. But really, what’s more accurate is to say these entities exist on the relationship between dataplanes and policies. I picked dataplanes because that’s the scope in which the _rules API is defined.

> I think all in all we've unfortunately not used `_` in other places, but I think the original thinking was to not use `_` for model/entity names but to use `_` for PARENT_CHILD things which would make this read better overall (so `KUMA_FAULTINJECTIONS` but `KUMA_DATAPLANE_RULES`)

I just stuck to the pattern that the env entries started out with which was `KUMA_` followed by more or less resource/entity name and finally a modifier for its "nature" or "type": `_COUNT` or `_ENABLED`.

We could just add underscores to separate all words. Then at least this becomes more readable.

I’m not set on keeping the `DATAPLANE` bit in there, but I do kinda want to avoid clashes with other concepts like RBAC.

features/mesh/dataplanes/DataplanePolicies.feature
kleinfreund: Most of it, yes. The only bits that we don’t test against are the `conf`s so they could technically be random.

It’s already pretty limiting to need to resort to four different env entries to control the test data here. There are things that can’t be expressed in tests like this like an empty `matchers` array in one rule while having two `matchers` items in another. This always happens as soon as an API has arrays within arrays that need controlling in tests. You’re limited to one `KUMA_*_COUNT` variable and every nested array will have to use it.

features/mesh/dataplanes/DataplanePolicies.feature
kleinfreund: For the time being, I changed the names to always separate words by underscore.

johncowen (approved):
I think the functionality of this looks good to me, I also prefer it to what we had previously.

There's probably some comment to be said for some of the code, but we are in the middle of the data layer thing, and when all this goes over to use that approach it will address most of those things.

I'm guessing it also will mean that we'll be doing less of the "test that Vue prints this data to the DOM" testing which will address some of the e2e test comments I was wondering about here, umm to explain a little more, I think I'm a little concerned about the size of these tests for something that doesn't give us a huge return on that size/effort. Not a blocker tho.

Not a big thing, but I did notice that a lot (well at least some) of the test assertions look in what is almost the entire page (`$policy-view`) to assert for text. You might want to change those in the short term to check/assert a selector that is more specific, completely up to you tho.

src/test-support/mocks/src/meshes/_/dataplanes/_/_rules.ts
```diff
@@ -0,0 +1,282 @@
...
+              ],
+            },
+          ],
+          type: 'MeshTimeout',
+          warnings: [],
+        },
+      ],
+    } satisfies InspectRulesForDataplane,
+  }
+}
```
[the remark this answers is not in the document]
kleinfreund: Let me try if this works out-of-the-box. It’s not terribly important to keep this around, but this is a case where reviewing the result and the code pretty much needs a concrete example that is somewhat appropriate.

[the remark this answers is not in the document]
kleinfreund: I registered this under the `test-dataplane` name. I’ll consider adding more examples of this kind for testing purposes, but I’ll do so separately. I half-expect a bit of iteration to be happening on this view going forward.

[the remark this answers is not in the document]
johncowen: Ah great, glad it works then.

> I’ll consider adding more examples of this kind for testing purposes

Please only do this sort of thing where necessary, testing should ideally uses the tooling we have to do make mocks in the tests. In the past I've found it to only be appropriate in uber complicated setups (which is generally reproducing specific edge case bugs found "in the wild"), those setups were far more complicated that the sort of thing I've seen here up until now.

If we start adding loads of these static examples everywhere it becomes harder and harder to keep them up to date, and then people start testing again static responses again etc etc, and this is what this "dynamic mocks" idea is supposed to avoid.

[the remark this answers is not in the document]
johncowen: P.S. If you mean "manual" testing purposes thats kinda ok sometimes also.

PR #1866, opened 2023-12-05
feat: add feature flagged dataplane traffic view

johncowen:
Enables dataplane traffic view behind a feature flag.

Pretty much a copy/paste of https://github.com/kumahq/kuma-gui/pull/1839 behind a feature flag but there are a couple of additional type casts.

Note: there is still placeholder code in here that is not rendered via corresponding the slot as the comment is commented out. 

Also, I hit the same CSS lint issue as in https://github.com/kumahq/kuma-gui/pull/1839 here. I'll bring the issue up at a later date.

Make sure you have a cookie set to `KUMA_TRAFFIC_ENABLED=1` to view. The view only shows on non-gateways.

### DIFF
--- a/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -1,5 +1,12 @@
 <template>
-  <RouteView name="data-plane-detail-view">
+  <RouteView
+    v-slot="{ route }"
+    :params="{
+      mesh: '',
+      dataPlane: '',
+    }"
+    name="data-plane-detail-view"
+  >
     <AppView>
       <template
         v-if="warnings.length > 0"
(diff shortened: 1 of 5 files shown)


src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -311,6 +463,24 @@ const warnings = computed(() => {
 </script>
 
 <style lang="scss" scoped>
+.traffic {
+  padding: 0;
+  container-type: inline-size;
+  container-name: traffic;
+  .columns {
```
kleinfreund, line 470: **Suggestion**: Please use a different class name to avoid the conflict with the global `columns` styles.

kleinfreund (commented):
General question: Should the existing inbounds section be visible while the feature flag resolves to true?

johncowen:
> General question: Should the existing inbounds section be visible while the feature flag resolves to true?

I think so yeah, the feature flag is so folks can play with this, so I don't think it matters whether we are also showing inbounds or not.

When we remove the feature flag we should decide whether we are removing Inbounds entirely at that point.

src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: 🤔 This is `.traffic .columns` within a scoped block so I doubt it will conflict with anything?

The only way would be if we were to put more columns inside `.traffic` but specifically in this file. Considering the size of the columns here already I'm not sure if that would happen, and if in the extremely weird case it did we would be editing this file, so we could just update the selector then?

I dunno, did I miss something there?

On another note: just speaking out loud more than anything, I'm unsure how Vue would deal with `container-name`s within `scoped` I would assume that aren't scoped.

kleinfreund: > 🤔 This is .traffic .columns within a scoped block so I doubt it will conflict with anything?

Not in that direction, no, but if you add a use of `columns` inside of `.traffic` that doesn’t want these new styles, you have a conflict.

Just a good practice to avoid this problem from even being possible by not re-using global class names like that.

> just speaking out loud more than anything, I'm unsure how Vue would deal with `container-name`s within `scoped` I would assume that aren't scoped.

Scoping is done on the level of selector lists and property declarations (i.e. `container-name: traffic) and at rules (i.e. `@container`) should be affected by it at all.

johncowen: > Not in that direction, no, but if you add a use of columns inside of .traffic that doesn’t want these new styles, you have a conflict.

> > The only way would be if we were to put more columns inside .traffic but specifically in this file. Considering the size of the columns here already I'm not sure if that would happen, and if in the extremely weird case it did we would be editing this file, so we could just update the selector then?

kleinfreund: Still, avoiding the clash proactively is the right thing to do because you completely remove the need to even remember that there can be a clash.

johncowen: > Scoping is done on the level of selector lists and property declarations (i.e. container-name: traffic) and at rules (i.e. @container`) should be affected by it at all.

Was that a typo as it doesn't look like container-names get scoped? I guess you meant that it's not affected?

Screenshot 2023-12-06 at 13 06 44

johncowen: > and if in the extremely weird case it did we would be editing this file, so we could just update the selector then?

Was this just a suggestion or is this a blocker?

kleinfreund: > Was that a typo as it doesn't look like container-names get scoped? I guess you meant that it's not affected?

Yes, that was a typo, it should read "shouldn’t be".

> Was your original comment just a suggestion or is this a blocker?

Just a suggestion. One I thought was a straightforward improvement and strictly better.

kleinfreund (commented):
Can we get some basic tests covering this? I know this is being iterated on behind a feature flag, but I think this is a good time to start adding some test coverage as this PRs hooks-up these features to the application.

johncowen:
The feature isn't finished yet, there is still hardcoded data in here commented out etc.

We are really interested in getting this into a place where we can test it against some real installs to get some feedback on that. Once we know what we are building is the right thing, we can add some more tests on top then. If this blows up it really doesn't matter as its behind a feature flag. I even used a brand new datasource uri for the data instead of sharing an existing to avoid editing existing application code.

FYI: There are tests on the data parsing (https://github.com/kumahq/kuma-gui/blob/master/src/app/data-planes/data/stats.spec.ts)

I'm guessing you are requesting e2e tests before getting this approved?

kleinfreund (approved):

PR #1867, opened 2023-12-05
chore(zones): ZoneIngress data layer pt1 

johncowen:
Reduces the majority of our ZoneIngress related reshaping functions.

This is pretty much the same approach as https://github.com/kumahq/kuma-gui/pull/1854 but has the additional weirdness of two types of ZoneIngress depending one whether its the root object or the child object. For the moment I'm copy/pasta'ing around this, but I a better approach would be to apply `Entity` conditionally, but I don't want to do that (or something else) here yet.

See https://github.com/kumahq/kuma-gui/issues/1264

Related https://github.com/johncowen/kuma-gui/commit/b11e78a5484df618da0b89dac585b81eaab8ea74 which aren't PRed yet as I also want to do ZoneEgress and then at least reduce repetition some more.

Oh I also updated some mocks here to:

1. replicate the fact that `subscriptions` never seems to be an empty array, its only ever absent or at least length=1
2. replicate random available services in the listing API.

### DIFF
--- a/src/app/zone-ingresses/data/index.ts
+++ b/src/app/zone-ingresses/data/index.ts
@@ -1,24 +1,100 @@
-import type { ZoneIngressOverview as PartialZoneIngressOverview } from '@/types/index.d'
+import type { PaginatedApiListResponse as CollectionResponse } from '@/types/api.d'
+import type {
+  ZoneIngressOverview as PartialZoneIngressOverview,
+  ZoneIngress as PartialZoneIngress,
+  KDSSubscription,
+} from '@/types/index.d'
 export type { AvailableService } from '@/types/index.d'
-type PartialZoneIngress = PartialZoneIngressOverview['zoneIngress']
 
-type RequiredZoneIngress =
-  Required<Pick<PartialZoneIngress,
-  'availableServices'
-  >>
-  & PartialZoneIngressOverview['zoneIngress']
+type PartialZoneIngressInsight = any
+type PartialInternalZoneIngress = PartialZoneIngressOverview['zoneIngress']
 
+// TODO(jc) Theres probably a better way to not copy/pasta this i.e. make `Entity` composable
+type InternalZoneIngress = {
+  socketAddress: string
+  advertisedSocketAddress: string
+} & Required<Pick<PartialInternalZoneIngress, 'availableServices'>> & PartialInternalZoneIngress
+
+export type ZoneIngress = {
+  config: PartialZoneIngress
+  socketAddress: string
+  advertisedSocketAddress: string
+} & Required<Pick<PartialZoneIngress, 'availableServices'>> & PartialZoneIngress
+// end TODO
+
+// TODO(jc) currently in our manually written types ZoneIngressInsight is any therefore
+// we have no Partial*
+export type ZoneIngressInsight = {
+  connectedSubscription?: KDSSubscription
+  subscriptions: KDSSubscription[]
+}
 export type ZoneIngressOverview = PartialZoneIngressOverview & {
-  zoneIngress: RequiredZoneIngress
+  zoneIngress: InternalZoneIngress
+  zoneIngressInsight?: ZoneIngressInsight
+  state: 'online' | 'offline'
+}
+// TODO(jc) Theres probably a better way to not copy/pasta this i.e. make `Entity` composable
+const InternalZoneIngress = {
+  fromObject: (item: PartialInternalZoneIngress): InternalZoneIngress => {
+    return {
+      ...item,
+      availableServices: Array.isArray(item.availableServices) ? item.availableServices : [],
+      socketAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
+      advertisedSocketAddress: item.networking?.advertisedAddress && item.networking?.advertisedPort ? `${item.networking.advertisedAddress}:${item.networking.advertisedPort}` : '',
+    }
+  },
+}
+
+export const ZoneIngress = {
+  fromObject: (item: PartialZoneIngress): ZoneIngress => {
+    return {
+      ...item,
+      config: item,
+      availableServices: Array.isArray(item.availableServices) ? item.availableServices : [],
+      socketAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
+      advertisedSocketAddress: item.networking?.advertisedAddress && item.networking?.advertisedPort ? `${item.networking.advertisedAddress}:${item.networking.advertisedPort}` : '',
+    }
+  },
+}
+// end TODO
+export const ZoneIngressInsight = {
+  fromObject: (item?: PartialZoneIngressInsight): ZoneIngressInsight | undefined => {
+    // if item isn't set don't even try augmenting things
+    return isSet<PartialZoneIngressInsight>(item)
+      ? ((item) => {
+        const subscriptions: KDSSubscription[] = Array.isArray(item.subscriptions) ? item.subscriptions : []
+        // figure out the connectedSubscription by looking at the connectTime
+        // and disconnectTime of the last subscription
+        const connectedSubscription = subscriptions.slice(-1).find((item) => item.connectTime?.length && !item.disconnectTime)
+        return {
+          ...item,
+          subscriptions,
+          connectedSubscription,
+        }
+      })(item)
+      : undefined
+  },
 }
 export const ZoneIngressOverview = {
   fromObject: (item: PartialZoneIngressOverview): ZoneIngressOverview => {
+    const insight = ZoneIngressInsight.fromObject(item.zoneIngressInsight)
+    const zoneIngress = InternalZoneIngress.fromObject(item.zoneIngress)
     return {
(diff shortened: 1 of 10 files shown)


src/app/zone-ingresses/views/ZoneIngressListView.vue
```diff
@@ -42,29 +42,29 @@
...
+              :items="data?.items"
               :error="error"
               :empty-state-message="t('common.emptyState.message', { type: 'Zone Ingresses' })"
               :empty-state-cta-to="t('zone-ingresses.href.docs')"
               :empty-state-cta-text="t('common.documentation')"
               :is-selected-row="(row) => row.name === route.params.zoneIngress"
               @change="route.update"
             >
-              <template #name="{ row }">
+              <template #name="{ row: item }">
```
kleinfreund, line 61: **Suggestion**: Could we add the explicit type annotations for the rows here so we have a better sense of safety that what we’re accessing here is correct?

```suggestion
              <template #name="{ row: item }: { row: ZoneIngressOverview }">
```

src/app/zone-ingresses/views/ZoneIngressConfigView.vue
```diff
@@ -33,7 +33,7 @@
           <template v-else>
             <ResourceCodeBlock
               id="code-block-zone-ingress"
-              :resource="data"
+              :resource="data.config"
```
kleinfreund, line 36: **Issue**: This isn’t correct. The YAML shown here is the ZoneIngress resource, not the config.

```suggestion
              :resource="data"
```

src/app/zone-ingresses/data/index.ts
```diff
@@ -1,24 +1,100 @@
...
+// end TODO
+export const ZoneIngressInsight = {
+  fromObject: (item?: PartialZoneIngressInsight): ZoneIngressInsight | undefined => {
+    // if item isn't set don't even try augmenting things
+    return isSet<PartialZoneIngressInsight>(item)
+      ? ((item) => {
+        const subscriptions: KDSSubscription[] = Array.isArray(item.subscriptions) ? item.subscriptions : []
+        // figure out the connectedSubscription by looking at the connectTime
+        // and disconnectTime of the last subscription
+        const connectedSubscription = subscriptions.slice(-1).find((item) => item.connectTime?.length && !item.disconnectTime)
```
kleinfreund, line 68: **Suggestion**: As discussed earlier, find the _first_ not-disconnected subscription

```suggestion
        const connectedSubscription = subscriptions.find((item) => item.connectTime?.length && !item.disconnectTime)
```

kleinfreund (changes requested):

src/app/zone-ingresses/views/ZoneIngressListView.vue
johncowen: Oh that's weird I thought we'd addressed this centrally?

When we made AppCollection generic was there a reason why we didn't pass that through:

https://github.com/kumahq/kuma-gui/blob/f50cd686172fa54593dc5a03bb1da772be012d51/src/app/application/components/app-collection/AppCollection.vue#L88

i.e. make the above

```
:row="row as Row"
```

src/app/zone-ingresses/views/ZoneIngressConfigView.vue
johncowen: Not sure if I'm not following you or if you missed this:

https://github.com/kumahq/kuma-gui/pull/1867/files#diff-9fb9e8d988eecfed0988d235c43f9055bc6a50b7e5b79890f25915568a72183eR19

? lmk

kleinfreund: Hm, I see.

I think it’s a bit weird to call the ZoneIngress resource "config". What we’re doing here is effectively storing the original version of a resource as it comes from the API.

Wouldn’t the `$raw` concept in the MADR work better for this?

src/app/zone-ingresses/views/ZoneIngressListView.vue
kleinfreund: Yeah, we can do that. Then that would be working centrally.

johncowen: Lemme make a separate PR, then we can rebase that to all these data layer PRs

src/app/zone-ingresses/views/ZoneIngressConfigView.vue
kleinfreund: Can we call this `.resource` perhaps? Config is a little unfortunate at least in code as we have a few clashes with this (notable in Zones, we have a "Config" tab showing the Zone CP config (and not the Zone or ZoneOverview resource); everywhere else we call this tab "YAML" for this reason).

src/app/zone-ingresses/data/index.ts
johncowen: I made an issue for this one as I think in previous PRs it still looking at the last https://github.com/kumahq/kuma-gui/issues/1890.

Seeing as we also want to amend the `connectTime` thing it would be good to do all entities/resources at once so that they are all consistent. 

If you feel strongly about it happy to do this one here now and do the others later.

src/app/zone-ingresses/views/ZoneIngressListView.vue
johncowen: x ref https://github.com/kumahq/kuma-gui/pull/1889

src/app/zone-ingresses/views/ZoneIngressConfigView.vue
johncowen: Is Zones/ZoneCPs the only place where we use the name `.config` in code?

kleinfreund: We currently have two places where we show a config:

- Zone CP config tab: the Zone CP config (will also be used for _the_ Zone in the new `zone` mode)
- Diagnostics view: the CP config (I think; also not sure why this is called Diagnostics)

In contrast, where we show resource YAMLs (which we _do_ sometimes call configs), we call them YAML in the UI to disambiguate.

johncowen: Just to confirm a few things:

> Diagnostics view: the CP config (I think; also not sure why this is called Diagnostics)

This one is from the `/config` endpoint so I don't think it has a `.config` property. Its the entire response that we use right?

> Zone CP config tab: the Zone CP config (will also be used for the Zone in the new zone mode)

This one is the same format response that we get from `/config` but in the subscriptions `.config` property is that right?

> (will also be used for the Zone in the new zone mode)

Do I remember right that this will need a new API endpoint with the word `/config` in it that will work like the main `/config` endpoint (ie. no `.config` property we just use the entire response)?

kleinfreund: > This one is from the `/config` endpoint so I don't think it has a `.config` property. Its the entire response that we use right?

Correct.

> This one is the same format response that we get from `/config` but in the subscriptions `.config` property is that right?

I’m not completely sure about this one. It’s at least extremely similar, but I can’t say if there are small or big difference between those two.

> Do I remember right that this will need a new API endpoint with the word `/config` in it that will work like the main `/config` endpoint (ie. no `.config` property we just use the entire response)?

Not quite: for _the_ Zone in `zone` mode, what we show in the Zone detail view’s `config` tab is simply the response from `/config` (i.e. the same as Diagnostics). It’s the exact same endpoint, not a new one.

johncowen: Ok gotcha, ok so I definitely want to keep the `.config` word here then, if anything I would change the augmented `.config` property I added recently here:

https://github.com/kumahq/kuma-gui/blob/a4c2a9e8083e7a80c8cfc1eb32a99f321f0d729a/src/app/zones/data/index.ts#L15

I want to use `.config` for the user-land YAML config (not `.yaml` as I don't want to use the format in the name of the thing).

The 'config' from the `/config` endpoint (and related responses/properties) are not a representation of this user-land YAML config.

I just noticed that `Zones` can have a very small user-land YAML config and even though we don't display that in the GUI as yet, if we ever do that should be in `.config`, and the current `.config` (that I recently added) needs to be renamed.

kleinfreund: So the thing we call Config in the UI would be something like `controlPlaneConfig` on the object and the thing we call YAML in the UI would be `config` on the object? (Note I didn’t propose the name `yaml` but `resource`)?

I can live with that even though it’s weirdly cross-wired.

johncowen: > So the thing we call Config in the UI would be something like controlPlaneConfig on the object and the thing we call YAML in the UI would be config on the object?

Yes something like that, possibly `controlPlaneConfig ` or `controlPlane.config` or I dunno I can't think of something that says "config but not supplied by YAML"

> I can live with that even though it’s weirdly cross-wired.

Great, lets resolve this one then, and then we can figure out naming/shaping the outlier in `Zone.config` later.

PR #1869, opened 2023-12-05
chore(zones): ZoneEgress data layer pt1

johncowen:
Reduces the majority of our ZoneEgress related reshaping functions, copy/pasta'ed from https://github.com/kumahq/kuma-gui/pull/1867 with `availableServices` and `advertisedSocketAddress` removed.

This is pretty much the same approach as https://github.com/kumahq/kuma-gui/pull/1854 but has the additional weirdness of two types of ZoneEgress depending one whether its the root object or the child object. For the moment I'm copy/pasta'ing around this, but I a better approach would be to apply `Entity` conditionally, but I don't want to do that (or something else) here yet.

See https://github.com/kumahq/kuma-gui/issues/1264 and https://github.com/kumahq/kuma-gui/pull/1867

No tests here yet, or even a branch this time, I just copy pasted it from https://github.com/kumahq/kuma-gui/pull/1867 and didn't want to copy pasta the tests again at least not yet.

Also note this is currently in draft until https://github.com/kumahq/kuma-gui/pull/1867 is approved (this PR may need amends depending on what happens with https://github.com/kumahq/kuma-gui/pull/1867)

### DIFF
--- a/src/app/zone-egresses/data/index.ts
+++ b/src/app/zone-egresses/data/index.ts
@@ -1,0 +1,93 @@
+import type { PaginatedApiListResponse as CollectionResponse } from '@/types/api.d'
+import type {
+  ZoneEgressOverview as PartialZoneEgressOverview,
+  ZoneEgress as PartialZoneEgress,
+  KDSSubscription,
+} from '@/types/index.d'
+
+type PartialZoneEgressInsight = any
+type PartialInternalZoneEgress = PartialZoneEgressOverview['zoneEgress']
+
+// TODO(jc) Theres probably a better way to not copy/pasta this i.e. make `Entity` composable
+type InternalZoneEgress = {
+  socketAddress: string
+} & PartialInternalZoneEgress
+
+export type ZoneEgress = {
+  config: PartialZoneEgress
+  socketAddress: string
+} & PartialZoneEgress
+// end TODO
+
+// TODO(jc) currently in our manually written types ZoneEgressInsight is any therefore
+// we have no Partial*
+export type ZoneEgressInsight = {
+  connectedSubscription?: KDSSubscription
+  subscriptions: KDSSubscription[]
+}
+export type ZoneEgressOverview = PartialZoneEgressOverview & {
+  zoneEgress: InternalZoneEgress
+  zoneEgressInsight?: ZoneEgressInsight
+  state: 'online' | 'offline'
+}
+// TODO(jc) Theres probably a better way to not copy/pasta this i.e. make `Entity` composable
+const InternalZoneEgress = {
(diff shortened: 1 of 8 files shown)


src/app/zone-egresses/data/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+  },
+}
+// end TODO
+export const ZoneEgressInsight = {
+  fromObject: (item?: PartialZoneEgressInsight): ZoneEgressInsight | undefined => {
+    // if item isn't set don't even try augmenting things
+    return isSet<PartialZoneEgressInsight>(item)
+      ? ((item) => {
+        const subscriptions: KDSSubscription[] = Array.isArray(item.subscriptions) ? item.subscriptions : []
```
lahabana, line 58:
```suggestion
        const subscriptions: XDSSubscription[] = Array.isArray(item.subscriptions) ? item.subscriptions : []
```

src/app/zone-egresses/data/index.ts
```diff
@@ -0,0 +1,93 @@
...
+        // figure out the connectedSubscription by looking at the connectTime
+        // and disconnectTime of the last subscription
+        const connectedSubscription = subscriptions.slice(-1).find((item) => item.connectTime?.length && !item.disconnectTime)
```
lahabana, line 61:
```suggestion
        const connectedSubscription = subscriptions.first((item) => item.connectTime?.length && !item.disconnectTime)
```

Be permissive maybe

src/app/zone-egresses/data/index.ts
johncowen: I'm guessing this might be that there is an XDSSubscription type in the backend somewhere, or maybe we just have the wrong name for the type? FYI: This was a KDSSubscription before this PR. If we do need to add additional type to model an XDSSubscription or change the name I think we should do that after this PR. I'm gonna resolve this for now, but feel free to unresolve if you think we should address this here. It would also be good to understand why we don't already have an XDSSubscription type in the frontend.

src/app/zone-egresses/data/index.ts
johncowen: Following our sync on this we are defo going to do this. We want to do it holistically over the entire project though so we don't do it in one place and not another (I made an issue so we also don't forget to do it https://github.com/kumahq/kuma-gui/issues/1890)

Handily some other things have also come up related to this and how we figure out `lastUpdatedTime` and some other things to do with subscriptions so that will all be thought about addresses at once everywhere.

Resolving this here for now, but there will be a quick follow up pretty soon.

kleinfreund (approved):

PR #1871, opened 2023-12-05
refactor(dataplanes): hook-up data transformations to data sourcing

kleinfreund:
Hooks up data transformation utilities to all dataplane data sources (except anything releated to the "Policies" tab).

### DIFF
--- a/src/app/data-planes/components/DataPlaneList.vue
+++ b/src/app/data-planes/components/DataPlaneList.vue
@@ -16,7 +16,7 @@
     :page-number="props.pageNumber"
     :page-size="props.pageSize"
     :total="props.total"
-    :items="props.items ? transformToTableData(props.items) : undefined"
+    :items="props.items"
     :error="props.error"
     :is-selected-row="props.isSelectedRow"
     @change="emit('change', $event)"
@@ -28,7 +28,7 @@
       <slot name="toolbar" />
     </template>
 
-    <template #name="{ row: item }: { row: DataPlaneOverviewTableRow }">
+    <template #name="{ row: item }">
       <RouterLink
         class="name-link"
         :title="item.name"
@@ -48,25 +48,29 @@
       </RouterLink>
     </template>
 
-    <template #services="{ row }: { row: DataPlaneOverviewTableRow }">
+    <template #type="{ row }">
+      {{ t(`data-planes.type.${row.dataplaneType}`) }}
+    </template>
+
+    <template #services="{ row }">
       <KTruncate
         v-if="row.services.length > 0"
         width="auto"
       >
         <div
-          v-for="(tag, index) in row.services"
+          v-for="(service, index) in row.services"
           :key="index"
         >
-          <TextWithCopyButton :text="tag.value">
+          <TextWithCopyButton :text="service">
             <RouterLink
               :to="{
                 name: 'service-detail-view',
                 params: {
-                  service: tag.value,
+                  service,
                 },
               }"
             >
-              {{ tag.value }}
+              {{ service }}
             </RouterLink>
           </TextWithCopyButton>
         </div>
@@ -77,28 +81,34 @@
       </template>
     </template>
 
-    <template #status="{ row }: { row: DataPlaneOverviewTableRow }">
+    <template #certificate="{ row }">
+      <template v-if="row.dataplaneInsight.mTLS?.certificateExpirationTime">
+        {{ formatIsoDate(row.dataplaneInsight.mTLS.certificateExpirationTime) }}
+      </template>
+
+      <template v-else>
+        {{ t('data-planes.components.data-plane-list.certificate.none') }}
+      </template>
+    </template>
+
+    <template #status="{ row }">
       <StatusBadge :status="row.status" />
     </template>
 
-    <template #warnings="{ row: item }: { row: DataPlaneOverviewTableRow }">
-      <KTooltip
-        v-if="Object.values(item.warnings).some((item) => item)"
-      >
-        <template
-          #content
-        >
+    <template #warnings="{ row }">
+      <KTooltip v-if="row.isCertExpired || row.warnings.length > 0">
+        <template #content>
           <ul>
-            <template
-              v-for="(warning, i) in item.warnings"
-              :key="i"
-            >
-              <li v-if="warning">
-                {{ t(`data-planes.components.data-plane-list.${i}`) }}
-              </li>
+            <template v-if="row.warnings.length > 0">
+              <li>{{ t('data-planes.components.data-plane-list.version_mismatch') }}</li>
+            </template>
+
+            <template v-if="row.isCertExpired">
+              <li>{{ t('data-planes.components.data-plane-list.cert_expired') }}</li>
             </template>
           </ul>
         </template>
+
         <WarningIcon
           class="mr-1"
           :size="KUI_ICON_SIZE_30"
@@ -111,7 +121,7 @@
       </template>
     </template>
 
-    <template #details="{ row }: { row: DataPlaneOverviewTableRow }">
+    <template #details="{ row }">
       <RouterLink
         class="details-link"
         data-testid="details-link"
@@ -138,28 +148,14 @@
 import { KUI_ICON_SIZE_30 } from '@kong/design-tokens'
 import { ArrowRightIcon } from '@kong/icons'
 
-import { getDataplaneType, getIsCertExpired, getStatusAndReason, getTags, getWarnings } from '../data/index'
+import type { DataplaneOverview } from '../data/index'
 import AppCollection from '@/app/application/components/app-collection/AppCollection.vue'
 import StatusBadge from '@/app/common/StatusBadge.vue'
 import TextWithCopyButton from '@/app/common/TextWithCopyButton.vue'
 import WarningIcon from '@/app/common/WarningIcon.vue'
-import type { DataPlaneOverview, LabelValue, StatusKeyword } from '@/types/index.d'
 import { useI18n } from '@/utilities'
 
 const { t, formatIsoDate } = useI18n()
-
-type DataPlaneOverviewTableRow = {
-  type: string
-  name: string
-  mesh: string
-  services: LabelValue[]
-  status: StatusKeyword
-  warnings: {
-    version_mismatch: boolean
-    cert_expired: boolean
-  }
-  certificate: string
-}
 
 type ChangeValue = {
   page?: number
@@ -170,11 +166,10 @@ const props = withDefaults(defineProps<{
   total?: number
   pageNumber: number
   pageSize: number
-  items: DataPlaneOverview[] | undefined
+  items: DataplaneOverview[] | undefined
   error: Error | undefined
   isSelectedRow: ((row: any) => boolean) | null
   summaryRouteName: string
-  canUseZones: boolean
 }>(), {
   total: 0,
   isSelectedRow: null,
@@ -183,41 +178,6 @@ const props = withDefaults(defineProps<{
 const emit = defineEmits<{
   (event: 'change', value: ChangeValue): void
 }>()
-
-function transformToTableData(dataPlaneOverviews: DataPlaneOverview[]): DataPlaneOverviewTableRow[] {
-  return dataPlaneOverviews.map((dataPlaneOverview) => {
-    const { mesh, name } = dataPlaneOverview
-    const type = t(`data-planes.type.${getDataplaneType(dataPlaneOverview)}`)
-
-    const tags = getTags(dataPlaneOverview)
-    const services = tags.filter((tag) => tag.label === 'kuma.io/service')
-
-    const { status } = getStatusAndReason(dataPlaneOverview)
-
-    let certificate
-    if (dataPlaneOverview.dataplaneInsight?.mTLS?.certificateExpirationTime) {
-      certificate = formatIsoDate(dataPlaneOverview.dataplaneInsight.mTLS.certificateExpirationTime)
-    } else {
-      certificate = t('data-planes.components.data-plane-list.certificate.none')
-    }
-
-    const warnings = getWarnings(dataPlaneOverview, props.canUseZones)
-    const isCertExpired = getIsCertExpired(dataPlaneOverview)
-
-    return {
-      name,
-      type,
-      mesh,
-      services,
-      status,
-      warnings: {
-        version_mismatch: warnings.length > 0,
-        cert_expired: isCertExpired,
-      },
-      certificate,
-    }
-  })
-}
 </script>
 
 <style lang="scss" scoped>
--- a/src/app/data-planes/data/index.ts
+++ b/src/app/data-planes/data/index.ts
@@ -42,7 +42,7 @@ export type DataplaneOverview = PartialDataplaneOverview & {
   services: string[]
 }
 
-export const DataplaneNetworking = {
+const DataplaneNetworking = {
   fromObject(partialDataplaneNetworking: PartialDataplaneNetworking): DataplaneNetworking {
     // TODO: Normalize inbound and outbound omitempty arrays.
     // TODO: Consider to determine inbound address and service address here.
@@ -63,7 +63,7 @@ export const Dataplane = {
   },
 }
 
-export const DataplaneInsight = {
+const DataplaneInsight = {
   fromObject(partialDataplaneInsight: PartialDataplaneInsight | undefined): DataplaneInsight {
     const subscriptions = Array.isArray(partialDataplaneInsight?.subscriptions) ? partialDataplaneInsight.subscriptions : []
     const connectedSubscription = subscriptions.find((subscription) => !subscription.disconnectTime)
--- a/src/app/data-planes/data/index.ts
+++ b/src/app/data-planes/data/index.ts
@@ -1,8 +1,12 @@
 import { getIsConnected } from '@/app/subscriptions/data'
+import type { PaginatedApiListResponse } from '@/types/api.d'
 import type {
-  DataPlaneOverview as DataplaneOverview,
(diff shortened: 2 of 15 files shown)


johncowen:
I know this is in draft but just wanted to say this is looking great! 🙌

src/app/data-planes/components/DataPlaneList.vue
```diff
@@ -77,28 +81,34 @@
       </template>
     </template>
 
-    <template #status="{ row }: { row: DataPlaneOverviewTableRow }">
+    <template #certificate="{ row }: { row: DataplaneOverview }">
+      <template v-if="row.dataplaneInsight.mTLS?.certificateExpirationTime">
```
kleinfreund, line 85: **Comment**: I’m considering to always set `mTLS` o we can save the optional chaining for stuff like this.

johncowen: Q: What's your thinking on how we would have things without mTLS set?

johncowen: I'm guessing that mTLS can sometimes be empty/absent to mean that mTLS isn't enabled on this Dataplane?

kleinfreund: Yes, that is the case. That’s one of the reasons why I didn’t change this in this PR. I might not change it at all if we rely on this logic. An alternative would be:

- introduce a field `hasMtlsEnabled: boolean` to `DataplaneOverview` and use that for such checks
- always write a (possibly empty) `mTLS` object onto `dataplaneInsight` so that accesses like this one are simpler (i.e. don’t need to use optional chaining)

In any case, that’s definitely something that I want to consider separately from this PR.

johncowen: Cool, gotcha, ok yeah defo something worth considering, thanks for sharing your thinking

src/app/data-planes/data/index.ts
```diff
@@ -13,56 +17,138 @@ export type DataplaneWarning = {
...
+    networking: DataplaneNetworking
   }
+  dataplaneInsight: DataplaneInsight
+  dataplaneType: 'standard' | 'builtin' | 'delegated'
+  status: 'online' | 'offline' | 'partially_degraded'
+  unhealthyInbounds: Array<{ service: string, port: number }>
+  lastUpdateTime?: string | undefined
+  warnings: DataplaneWarning[]
+  isCertExpired: boolean
+  serviceTags: LabelValue[]
```
johncowen, line 42: Should this just be `services`?

src/app/data-planes/data/index.ts
```diff
@@ -13,56 +17,138 @@ export type DataplaneWarning = {
...
+  connectedSubscription?: DiscoverySubscription
+}
+
+export type DataplaneOverview = PartialDataplaneOverview & {
+  dataplane: {
+    networking: DataplaneNetworking
   }
+  dataplaneInsight: DataplaneInsight
+  dataplaneType: 'standard' | 'builtin' | 'delegated'
+  status: 'online' | 'offline' | 'partially_degraded'
```
johncowen, line 37: I take it you don't want to change this to `state`?

src/app/data-planes/data/index.ts
```diff
@@ -84,24 +170,13 @@ export function getTags(dataplaneOverview: DataplaneOverview): LabelValue[] {
...
-export function getWarnings(dataplaneOverview: DataplaneOverview, canUseZones: boolean): DataplaneWarning[] {
-  const subscriptions = dataplaneOverview.dataplaneInsight?.subscriptions ?? []
-  if (subscriptions.length === 0) {
-    return []
-  }
-
-  const lastSubscription = subscriptions[subscriptions.length - 1]
-  if (!('version' in lastSubscription) || !lastSubscription.version) {
+function getWarnings({ subscriptions }: DataplaneInsight, tags: LabelValue[], canUseZones: boolean): DataplaneWarning[] {
+  const lastSubscription = subscriptions.at(-1)
```
johncowen, line 178: Similar to above this should use `connectedSubscription`

src/app/data-planes/sources.ts
```diff
@@ -103,7 +107,7 @@ export const sources = (api: KumaApi) => {
     },
 
     '/meshes/:mesh/dataplane-overviews/:name': async (params) => {
-      return api.getDataplaneOverviewFromMesh(params)
+      return DataplaneOverview.fromObject(await api.getDataplaneOverviewFromMesh(params), can('use zones'))
```
johncowen, line 110: This is more of a self-note/let you know that I'm still looking/thinking about this. Ideally `zones` should be dealt with in zones. I need to refresh my mind on this to decide what (if anything) to do here though.

johncowen (commented):
I left some comments on some of the more important/straight forwards things, plus a question or 2. Let me know on those. FYI I'm still looking over the cdoe for this.

src/app/data-planes/data/index.ts
```diff
@@ -13,56 +17,138 @@ export type DataplaneWarning = {
...
+}
+
+export type DataplaneInsight = PartialDataplaneInsight & {
+  subscriptions: DiscoverySubscription[]
+  connectedSubscription?: DiscoverySubscription
+}
+
+export type DataplaneOverview = PartialDataplaneOverview & {
+  dataplane: {
+    networking: DataplaneNetworking
```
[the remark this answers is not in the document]
kleinfreund: No, it’s not a Dataplane object. It’s a `dataplane` property holding an object with the sole property `networking`. There is no point in trying to make this into a full `Dataplane` object either. The fields `mesh`, `type` (i.e. `'Dataplane'`), `name`, `creationTime`, and `modificationTime` are not useful and the case of the times not relevant because we can’t actually know them for the underlying Dataplane object.

src/app/data-planes/data/index.ts
```diff
@@ -13,56 +17,138 @@ export type DataplaneWarning = {
...
-  return lastSubscription.status.lastUpdateTime
+export const DataplaneOverview = {
+  fromObject(partialDataplaneOverview: PartialDataplaneOverview, canUseZones: boolean): DataplaneOverview {
+    const dataplaneInsight = DataplaneInsight.fromObject(partialDataplaneOverview.dataplaneInsight)
+    const networking = DataplaneNetworking.fromObject(partialDataplaneOverview.dataplane.networking)
+
+    const dataplaneType = getDataplaneType(networking)
+    const status = getStatus(partialDataplaneOverview)
+    const unhealthyInbounds = getUnhealthyInbounds(networking)
+    const lastUpdateTime = dataplaneInsight.subscriptions.at(-1)?.status.lastUpdateTime
```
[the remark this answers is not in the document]
kleinfreund: The `lastUpdateTime` should also yield a result for disconnected Dataplanes and that’s not possible with the concept of `connectedSubscription`.

src/app/data-planes/data/index.ts
kleinfreund: I don’t think so. Why should we use `connectedSubscription` here? This information is also useful if the Dataplane is disconnected.

src/app/data-planes/data/index.ts
kleinfreund: Hm, kinda nice to know you’re not looking at a list of service names, but it doesn’t matter much. I changed it.

src/app/data-planes/data/index.ts
```diff
@@ -13,56 +17,138 @@ export type DataplaneWarning = {
...
 }
 
-export function getLastUpdateTime(dataplaneOverview: DataplaneOverview): string | undefined {
-  const subscriptions = dataplaneOverview.dataplaneInsight?.subscriptions ?? []
-  if (subscriptions.length === 0) {
-    return undefined
+export type DataplaneNetworking = PartialDataplaneNetworking
+
+export type Dataplane = PartialDataplane & {
+  networking: DataplaneNetworking
```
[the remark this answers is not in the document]
johncowen: Ok so this is in place ready for more changes in a follow up PR? If so feel free to resolve.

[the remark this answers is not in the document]
kleinfreund: Yep.

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
johncowen: Oh I see this is like the 'child' one 🤔 I think these should still be explicitly named types (I think not totally sure), happy to leave this as is for now. Feel free to resolve.

src/app/data-planes/data/index.ts
johncowen: Oh sorry I meant that this should be an array of services, not tags. Where we use this its not that we want tags its that we want a list if service names right? (or did I misunderstand this?)

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
johncowen: In the same sense that we shouldn't be looking for the last subscription elsewhere, should we be finding the one/subscription with the latest `lastUpdateTime` here?

If so, I'm not super precious about getting this tweaked in this PR just if we are doing this sort of thing we should do it here or in a follow up.

Let me know on the above question.

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
kleinfreund: Just to clarify: what do you want to be a named typed? The `{ networking: DataplaneNetworking }` structure?

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
kleinfreund: > In the same sense that we shouldn't be looking for the last subscription elsewhere, should we be finding the one/subscription with the latest `lastUpdateTime` here?

To determine what’s the latest subscription by `lastUpdateTime`, we would have to iterate over all subscriptions, convert the `lastUpdateTime` to `Date` objects and keep record of the latest one.

If we do this here (for `lastUpdateTime`), we should also do it for determining whether we can set a `connectedSubscription`: taking the latest subscription and checking if that subscription has no `disconnectTime`. If so, that’s the `connectedSubscription`.

As far as I understood it, going back to the mode of discovering the connected subscription by finding the _first_ not disconnected one (in contrast to just checking for that condition on the last subscription) was meant as a safe guard in case we encounter bugs of the nature where subscriptions were out of order.

If now we introduce the new measure of only looking at the subscription with the latest `lastUpdateTime`, we potentially wouldn’t find the a not disconnected subscription if it doesn’t happen to be the latest.

To avoid _that_, we would have to perform the searches like this:

- For `lastUpdateTime`: find the subscription with the latest `lastUpdateTime` overall (and take its `lastUpdateTime`)
- For `connectedSubscription`: find the _not disconnected_ subscription with the latest `lastUpdateTime`

Is this what we want? It seems quite convoluted.

src/app/data-planes/data/index.ts
kleinfreund: As far as I understand it, all subscriptions but the latest one are irrelevant when it comes to presenting current information about a Dataplane. None of its data might still be in effect so showing warnings based on that would be incorrect.

src/app/data-planes/data/index.ts
kleinfreund: Oh, I see. Sure, that makes sense as well. I changed this to be a `string[]`.

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
johncowen: Well to give you an example of what I was thinking over here we have an `InternalZoneIngress` type that isn't exported and is only used for the 'child'. But we use a separate type definition there.

https://github.com/kumahq/kuma-gui/pull/1867/files#diff-9fb9e8d988eecfed0988d235c43f9055bc6a50b7e5b79890f25915568a72183eR13-R16

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
johncowen: I think there are some very good points here.

> we would have to iterate over all subscriptions, convert the lastUpdateTime to Date objects 

I would like to convert all our Dates to Date objects anyway.

> If we do this here (for lastUpdateTime), we should also do it for determining whether we can set a connectedSubscription: taking the latest subscription and checking if that subscription has no disconnectTime. If so, that’s the connectedSubscription.

🤔 so you are saying that we can potentially have two subscriptions without a disconnectTime? I think the difference here is the phrase `lastUpdate`. If the one connected subscription can be anywhere in the list I would have guessed the `lastUpdateTime` also could be anywhere in the list?

Lets resolve this one if you are ok doing that, I would like to look/ask/think about this further outside of this PR, but I would guess this will come up again.

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
kleinfreund: > what do you want to be a named typed? The `{ networking: DataplaneNetworking }` structure?

So is that a "Yes" to that second question?

(Slightly related: I removed the export and tests for the DataplaneNetworking and DataplaneInsight transformer objects. The tests are redundant and the objects aren’t meant to be used outside of this file.)

src/app/onboarding/views/OnboardingDataplanesView.vue
```diff
@@ -124,7 +124,7 @@ async function getAllDataplanes() {
         const { name, mesh } = dataPlane
 
         const dataPlaneOverview = await kumaApi.getDataplaneOverviewFromMesh({ mesh, name })
-        const { status } = getStatusAndReason(dataPlaneOverview)
+        const status = getStatus(dataPlaneOverview)
```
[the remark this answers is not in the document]
johncowen: Oh I see, thanks!

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
kleinfreund: > Lets resolve this one if you are ok doing that, I would like to look/ask/think about this further outside of this PR, but I would guess this will come up again.

Fine by me.

> 🤔 so you are saying that we can potentially have two subscriptions without a disconnectTime?

At least in the bugs we’ve seen, yes, that was part of the problem: not only were the subscriptions out of order, there were multiple not disconnected subscriptions.

Everything being how it’s supposed to be however would mean (1) there is only ever one not disconnected subscription and (2) the last subscription is the only interesting/relevant one to look at.

Consider also the detail that the API will actually not populate some of the data of older subscriptions (e.g. the config in the case of `ZoneOverview`). I think it would be good to have all of those rules written down somewhere that we can look at (preferably as close to the source (i.e. the backend) as possible).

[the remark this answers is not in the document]
johncowen: ok gotcha, definitely worth thinking about this one further, I think there will be more work to come around here, not only here but in zones also. Lets resolve this one for now.

src/app/data-planes/data/index.ts
kleinfreund: Yes, that’s how I understand it. Overall I think it’s more useful to show the same kinds of warnings for a disconnected Dataplane that you would for a connected one.

johncowen: Ok gotcha, thanks, similar to the over thread on this PR I think there'll be a bit of work off the back of this. Feel free to resolve this one.

src/app/data-planes/data/index.ts
johncowen: Lets resolve this one for now.

johncowen (approved):
Pretty sure this is good to go, I think the conversations that came off the back of this are resolved/resolveable for the moment.

Thanks again, nice PR!

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
kleinfreund: Just for reference, we could consider adding the following two functions to the subscriptions data layer for this purpose:

```ts
export function getLatestSubscription<S extends KDSSubscription | DiscoverySubscription>(subscriptions: S[]): S | undefined {
  let latestSubscription: S | undefined

  let currentLastUpdateTime = 0
  for (const subscription of subscriptions) {
    const lastUpdateTime = (new Date(subscription.status.lastUpdateTime)).getTime()

    if (lastUpdateTime > currentLastUpdateTime) {
      currentLastUpdateTime = lastUpdateTime
      latestSubscription = subscription
    }
  }

  return latestSubscription
}

export function getConnectedSubscription<S extends KDSSubscription | DiscoverySubscription>(subscriptions: S[]): S | undefined {
  const latestSubscription = getLatestSubscription(subscriptions)

  if (latestSubscription && !latestSubscription.disconnectTime) {
    return latestSubscription
  }

  return undefined
}
```

And store both `latestSubscription` and `connectedSubscription` on a `*Insight` object (both for ZoneOverview and DataplaneOverview). This would isolate both modules from changes when we decide to change the rules.

src/app/data-planes/data/index.ts
kleinfreund: Other related thread for reference: https://github.com/kumahq/kuma-gui/pull/1871#discussion_r1422451135

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
johncowen: I think its a yes, yes 😅 

We can decide/sort this after though

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
johncowen: > And store both latestSubscription and connectedSubscription on a *Insight object (both for ZoneOverview and DataplaneOverview). 

Yes exactly that ^, please don't go ahead and do it just yet. I'd like to get everything moved over to this data layer `fromObject` API first. Then we can do these kind of things separately as a whole across the entire project/all entities.

PR #1872, opened 2023-12-05
chore(deps): address breaking changes in marked

kleinfreund:

### DIFF
--- a/vite.config.production.ts
+++ b/vite.config.production.ts
@@ -46,12 +46,12 @@ export const config: UserConfigFn = ({ mode }) => {
           schema: DEFAULT_SCHEMA.extend(
             new Type('tag:yaml.org,2002:text/markdown', {
               kind: 'scalar',
-              construct: (data) => {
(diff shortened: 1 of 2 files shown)


vite.config.production.ts
```diff
@@ -51,7 +51,8 @@ export const config: UserConfigFn = ({ mode }) => {
                 // for which we use FormatJS under the hood. FormatJS requires you to escape any XML/HTML looking
                 // things, plus ICU '{' and '}', hence this replace.
                 // If we ever need !!text/markdown for anything else we should do something like !!text/icu+markdown
-                return marked(data).replace(/</g, "'<'")
+                const str = marked(data) as string
```
kleinfreund, line 54: **Comment**: Awaiting this didn’t work and made a bunch of tests fail. That’s not super surprising given that this a callback function passed to another thing to consume.

It looks like in our use, we’re not actually getting a promise, so type assertion should be fine here.

johncowen (approved):

Issue #1686, opened 2023-10-30
Support zone mode


lahabana:
### Description

With https://github.com/kumahq/kuma/issues/8190 we'll have the GUI in a zone. This means that the GUI should support this. There will likely be NO difference between standalone and zone GUI as zone is the future of Zone.

We may add a page to show KDS connection state of something like this.

kleinfreund:
In https://github.com/kumahq/kuma-gui/pull/1875, I’m adding the `zone` entry to our translation files so that the UI at least renders in zone mode.

lahabana:
There's an extra field called: `zone_instance_id` that can be present (it's new so it dpends). Probably worth showing

kleinfreund:
Note: The "name" of the Zone in zone mode is in /config

Also, /config includes whether KDS is enabled. If it’s not enabled, there isn’t going to be a Zone Insight. If it’s enabled, it behaves like in global mode; however, the subscriptions data is different.

kleinfreund:
xref: https://github.com/kumahq/kuma/issues/8767

kleinfreund:
We changed our mind on this again and will do the following on the UI side:

- global mode: nothing changes
- zone mode + KDS enabled: nothing changes, the GUI will be disabled on the backend side
- zone mode + KDS disabled: make everything behave like standalone mode behaves currently (essentially, replace mentions of `standalone` with `zone`, i.e. deprecate standalone mode)

PR #1875, opened 2023-12-06
fix: ui not rendering in zone mode

kleinfreund:
Fix the UI not rendering at all in zone mode on account of the mode not being listed in our translation files. This is just a small fix to make the UI work in Kuma again. More work is needed to actually support zone mode.

### DIFF
--- a/src/locales/en-us/common/index.yaml
+++ b/src/locales/en-us/common/index.yaml
@@ -14,6 +14,7 @@ common:
     mode:
       standalone: Standalone
       global: Multi-Zone
+      zone: Zone
   warnings:
     CERT_EXPIRED: !!text/markdown |
       The certificate for this dataplane has expired


johncowen (approved):
Once this is in could we do another PR with the following:

Wherever this is needed can we make it look for `common.product.mode` first and if not use a defaultMessage of `t(http.api.values.${mode})`? Let me know if thats problematic for any reason tho.

kleinfreund:
> Once this is in could we do another PR with the following:
> 
> Wherever this is needed can we make it look for `common.product.mode` first and if not use a defaultMessage of `t(http.api.values.${mode})`? Let me know if thats problematic for any reason tho.

Could you leave a comment about this in https://github.com/kumahq/kuma-gui/issues/1686? I don’t intend to pick up that task currently. This PR is really just what it set out to be: fix the GUI flat out not working in zone mode.

PR #1879, opened 2023-12-07
feat: allow copying service name from service/dataplane list views

### LINKED ISSUE #1747: Add possibility to copy service name from Dataplane proxy view
### Description

It would be nice to be able to copy the service name from data plane view instead of going back to service

kleinfreund:
**feat(services): add copy button to service name in list view**

**refactor(CopyButton): use type syntax for props**

**feat(dataplanes): show copy button for services in list view**

Add support for showing a copy button to the TagList component.

Resolves #1747.

### DIFF
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -225,11 +225,13 @@ const click = (e: MouseEvent) => {
 </script>
 
 <style lang="scss" scoped>
-.app-collection :deep(td:first-child),
-.app-collection :deep(td:first-child > *) {
+.app-collection :deep(td:first-child) {
   color: inherit;
   font-weight: $kui-font-weight-semibold;
-  text-decoration: none;
+}
+
+.app-collection :deep(td:first-child a) {
+  color: currentColor;
 }
 
 .app-collection-toolbar {
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -225,13 +225,10 @@ const click = (e: MouseEvent) => {
 </script>
 
 <style lang="scss" scoped>
-.app-collection :deep(td:first-child) {
+.app-collection :deep(td:first-child),
+.app-collection :deep(td:first-child *) {
   color: inherit;
   font-weight: $kui-font-weight-semibold;
-}
-
-.app-collection :deep(td:first-child a) {
-  color: currentColor;
   text-decoration: none;
 }
 
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -232,6 +232,7 @@ const click = (e: MouseEvent) => {
 
 .app-collection :deep(td:first-child a) {
   color: currentColor;
+  text-decoration: none;
 }
 
 .app-collection-toolbar {
--- a/src/app/common/CopyButton.vue
+++ b/src/app/common/CopyButton.vue
@@ -96,7 +96,7 @@ export default {
 
 .copy-button[data-tooltip-text]::after {
   background-color: var(--tooltip-background-color);
-  border-radius: $kui-border-radius-20;
+  border-radius: $kui-border-radius-10;
   color: $kui-color-text-inverse;
   content: attr(data-tooltip-text);
   font-weight: $kui-font-weight-regular;
--- a/src/app/common/CopyButton.vue
+++ b/src/app/common/CopyButton.vue
@@ -28,50 +28,23 @@
 <script lang="ts" setup>
 import { KUI_ICON_SIZE_30 } from '@kong/design-tokens'
 import { CopyIcon } from '@kong/icons'
-import { KButton, KClipboardProvider } from '@kong/kongponents'
 
-import type { PropType } from 'vue'
-
-const props = defineProps({
-  text: {
-    type: String,
-    required: false,
-    default: '',
-  },
-
-  getText: {
-    type: Function as PropType<() => string | Promise<string>>,
-    required: false,
-    default: null,
-  },
-
-  copyText: {
-    type: String,
-    required: false,
-    default: 'Copy',
-  },
-
-  tooltipSuccessText: {
-    type: String,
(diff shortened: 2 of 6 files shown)


johncowen (commented):
I just took a very quick look at this, and clicking on the copy button in the listing view navigates me to the detail view.

Apart from that I'm not totally sold on the copy button being in the tag thing, but as I said I just took a very quick look and noticed the above thing, so leave that one with me some more.

kleinfreund:
> clicking on the copy button in the listing view navigates me to the detail view.

Good catch. I’ve fixed this now. We had this before and the strategy was making sure the links are nested inside `TextWithCopyButton` and not the other way around.

> I'm not totally sold on the copy button being in the tag thing, but as I said I just took a very quick look and noticed the above thing, so leave that one with me some more.

Me neither. I was thinking to stop using the TagList component specifically for the Dataplane list view. Let me try that. That might just be the solution here.

kleinfreund:
> I was thinking to stop using the TagList component specifically for the Dataplane list view. Let me try that. That might just be the solution here.

I’ve implemented this now. I think I like this better.

kleinfreund:
> Oh, one thing I noticed which might not be fixable easily here (but could be, lmk) is the non-rounded corners on the tooltip/notification things:

I added the Kongponents default border radius to this now. I have work planned to replace CopyButton with KCopy completely, but that’s not priority right now.

kleinfreund:
> I added the Kongponents default border radius to this now. I have work planned to replace CopyButton with KCopy completely, but that’s not priority right now.

Huh, I didn’t realize this already had a border radius. Let me look at this some more.

Okay, so specifically it’s not applied for the services in the Dataplane list. Strange. I’ll look at this a bit. If it’s not easy to fix, let’s do that separately.

kleinfreund:
Okay, yes, let me look at that separately. The issue is the KTruncate use using `overflow: hidden` which cuts-off parts of the tooltip making it look like it has no border-radius. That’s not exactly trivial to fix.

johncowen:
> That’s not exactly trivial to fix.

Cool don't worry on that one then thanks for looking.

I noticed this also, looks like the service names have turned blue in this PR:

Screenshot 2023-12-07 at 15 13 53

kleinfreund:
> I noticed this also, looks like the service names have turned blue in this PR:

I changed the AppCollection styles for this to fix that and also removed an unnecessary `text-decoration: none;`.

johncowen:
> and also removed an unnecessary text-decoration: none;.

Oh wait, that means we get an underline on hover which I don't think we want

kleinfreund:
> Oh wait, that means we get an underline on hover which I don't think we want

Isn’t that _exactly_ what we want? We have the following base rule for links:

```css
a:hover,
a:focus {
  text-decoration: underline
}
```

So why shouldn’t it apply to the name cells in AppCollection?

johncowen:
It's not what we want (unless something has changed that I'm not aware of). We spoke about this a good while ago now.

The name of the thing in the row (i.e. the first cell) shouldn't look/act like a link, the entire row is the clickable thing. The text in the first cell shouldn't be blue/underlined/underline on hover or anything that makes it look like a link. It should be black and slightly bold, as close as possible to this:

Screenshot 2023-12-08 at 10 18 07

> We have the following base rule for links:

This is why this rule is there, to overwrite the link base rule.

kleinfreund:
> It's not what we want (unless something has changed that I'm not aware of). We spoke about this a good while ago now.
> 
> The name of the thing in the row (i.e. the first cell) shouldn't look/act like a link, the entire row is the clickable thing. The text in the first cell shouldn't be blue/underlined/underline on hover or anything that makes it look like a link. It should be black and slightly bold, as close as possible to this:

I’ve set this declaration again as it’s not related to this PR’s purpose, but I really, really disagree with this entire statement.

It’s not clear why something that’s a link shouldn’t look like any other link in the application by default. In this case, a special rule applies to the color and font-weight as a first cell item in tables, but why the link shouldn’t be underlined on hover is unclear.

The link also very much acts like a link: You can activate it to navigate, you can focus it, you can use the whole array of browser link interactions like MouseWheel+Click, Ctrl+Click, etc.

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -225,10 +225,13 @@ const click = (e: MouseEvent) => {
...
 <style lang="scss" scoped>
-.app-collection :deep(td:first-child),
-.app-collection :deep(td:first-child > *) {
+.app-collection :deep(td:first-child) {
   color: inherit;
   font-weight: $kui-font-weight-semibold;
+}
+
+.app-collection :deep(td:first-child a) {
+  color: currentColor;
```
johncowen, line 234: I don't think there are any changes required in this file. You can just leave it as it is. I think this has only ended up being changed because of the little visual changes made here which we've now undone, but we've undone them differently to how they were in the first place instead of just undoing.

Another note: `color: currentColor` is a little weird seeing as we are setting the `color` to the current color, which we are setting to the current color, which we are setting to the current color etc etc. I think `inherit` is what we mean here.

kleinfreund: The changes are required because otherwise the color of the name cell link for services is now blue.

I’m going to leave this as-is because it works.

johncowen: Oh I see, gotcha! Could we just do this then:

```css
.app-collection :deep(td:first-child),
.app-collection :deep(td:first-child *) {
  color: inherit;
  font-weight: $kui-font-weight-semibold;
  text-decoration: none;
}
```

i.e. remove the immediate child selector `>`? Its the removal of this in your above code that solves the issue I think not the change to use `a` or the weird `color: currentColor` (I'd rather not use that if poss, this is my biggest "what the what!?")

kleinfreund: `currentColor` isn’t weird. I encourage you to read up on it and let’s not dwell on PR suggestions that end up being nitpicks. This wasn’t helpful.

I’ve changed this _again_ to what you suggested.

johncowen (approved):
Thanks!

PR #1884, opened 2023-12-08
chore(dataplanes): tweak new rules API-based policies

### LINKED ISSUE #1883: Tweak appearance of Policies tab
### Description

Add an empty state to the Dataplane policies tab if there are no policies matching.

Also check column widths. Long matchers blow out column.

Also, make the `and` red as well (like the `!`). Wrap matcher expressions to new line.

kleinfreund:
**chore(dataplanes): add empty state to policies tab**

**chore(dataplanes): tweak new rules API-based policies**

Change table layout to always have fixed column widths.

Make sure that table cell content that becomes too long can wrap onto new lines.

Explicitly place each matcher’s individual expressions on a separate line (after the `and`).

Resolves #1883.

image

image

### DIFF
--- a/src/app/data-planes/components/StandardDataplanePolicies.vue
+++ b/src/app/data-planes/components/StandardDataplanePolicies.vue
@@ -3,56 +3,62 @@
     data-testid="standard-dataplane-policies"
     class="stack"
   >
-    <KCard v-if="props.showPoliciesSection">
-      <PolicyTypeEntryList
-        id="policies"
-        :policy-type-entries="policyTypeEntries"
-        data-testid="policy-list"
-      />
+    <KCard v-if="(props.showPoliciesSection ? props.sidecarDataplanes.length === 0 : true) && props.inspectRulesForDataplane.rules.length === 0">
+      <EmptyBlock />
     </KCard>
 
-    <KCard v-if="proxyRule">
-      <h3>{{ t('data-planes.routes.item.proxy_rule') }}</h3>
+    <template v-else>
+      <KCard v-if="props.showPoliciesSection">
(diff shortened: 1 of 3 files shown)


src/app/data-planes/components/StandardDataplanePolicies.vue
```diff
@@ -3,56 +3,62 @@
...
+        </h3>
+
+        <div
+          v-for="(fromRule, index) in fromRuleInbounds"
+          :key="index"
+        >
+          <h4>{{ t('data-planes.routes.item.port', { port: fromRule.port }) }}</h4>
+
+          <RuleEntryList
+            :id="`from-rules-${index}`"
```
johncowen, line 54: Is this just because `KCodeBlock` requires an `id` to be set which gets used internally? To rephrase that a little, is this id necessary for anything outside of `KCodeBlock`?

johncowen (approved):
I had one question, but otherwise LGTM!

src/app/data-planes/components/StandardDataplanePolicies.vue
kleinfreund: Yes. A code block needs a unique ID to avoid label/input association clashes.

kleinfreund: Not really something I’m concerned about at the moment. Setting these IDs is simple enough and not something we do over and over.

johncowen: Yeah don't worry, didn't need doing here, was just checking/questioning for future.

PR #1886, opened 2023-12-11
refactor(services): set-up data transformations

kleinfreund:
Set-up data transformations for the services module.

### DIFF
--- a/src/app/services/sources.ts
+++ b/src/app/services/sources.ts
@@ -1,7 +1,9 @@
+import { ExternalService, ServiceInsight } from './data'
 import { DataSourceResponse } from '@/app/application/services/data-source/DataSourcePool'
 import type KumaApi from '@/services/kuma-api/KumaApi'
 import type { PaginatedApiListResponse as CollectionResponse } from '@/types/api.d'
-import type { ExternalService, ServiceInsight } from '@/types/index.d'
+
+export type { ExternalService, ServiceInsight } from './data'
 
 type CollectionParams = {
   mesh: string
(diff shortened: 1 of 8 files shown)


src/app/services/sources.ts
```diff
@@ -56,7 +58,7 @@ export const sources = (api: KumaApi) => {
         tag: [`kuma.io/service:${service}`],
       })
 
-      return items.length > 0 ? items[0] : null
+      return items.length > 0 ? ExternalService.fromObject(items[0]) : null
```
johncowen, line 61: How much of a knock on effect would it be to remove this `null` here? (preferably replace it with a `[]`)

src/app/services/views/ServiceListView.vue
```diff
@@ -74,14 +74,14 @@
...
+                {{ row.serviceType }}
               </template>
 
-              <template #addressPort="{ rowValue }">
+              <template #addressPort="{ row }">
                 <TextWithCopyButton
-                  v-if="rowValue"
-                  :text="rowValue"
+                  v-if="row.addressPort"
+                  :text="row.addressPort"
```
johncowen, line 84: 👏 Glad we are using `row` now and not `rowValue`.

johncowen (approved):
LGTM!

src/app/services/sources.ts
kleinfreund: First of all, this is not returning multiple items. It can only ever be one or none, so `[]` doesn’t work.

This is an interesting case: in `src/app/services/components/ExternalServiceDetails.vue:15` we check against `null` to show a specific empty state. However, we can’t easily change this to `undefined` because we’ve chosen for `undefined` to mean a `DataSource` is loading.

So this will need some more thought.

PR #1887, opened 2023-12-11
chore(deps-dev): Manually bump @badeball/cypress-cucumber-preprocessor from 19.1.0 to 19.2.0

johncowen:
Manually bumped to replace https://github.com/kumahq/kuma-gui/pull/1822 which is failing, possibly due to this being added:

https://github.com/kumahq/kuma-gui/pull/1822/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR8018

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "@babel/core": "7.23.5",
     "@babel/preset-env": "7.23.5",
     "@babel/preset-typescript": "7.23.3",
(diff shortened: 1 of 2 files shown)


johncowen:
Yeah for sure, lets keep an eye out for it I suppose, if we start to see it often then we should look into it, if not then thats this particular problem solved for us I suppose 🤷

PR #1888, opened 2023-12-11
refactor(policies): set-up data transformations

kleinfreund:
Set-up data transformations for the policies module.

### DIFF
--- a/src/app/policies/data/index.ts
+++ b/src/app/policies/data/index.ts
@@ -4,6 +4,8 @@ import type {
   PolicyEntity as PartialPolicy,
 } from '@/types/index.d'
 
+export { PolicyType } from '@/types/index.d'
+
 export type PolicyDataplane = PartialPolicyDataplane
 
 export type Policy = PartialPolicy
--- a/src/app/policies/data/index.ts
+++ b/src/app/policies/data/index.ts
@@ -1,0 +1,39 @@
+import type { PaginatedApiListResponse } from '@/types/api.d'
+import type {
+  PolicyDataplane as PartialPolicyDataplane,
+  PolicyEntity as PartialPolicy,
+} from '@/types/index.d'
+
+export type PolicyDataplane = PartialPolicyDataplane
+
+export type Policy = PartialPolicy
+
+export const PolicyDataplane = {
+  fromObject(partialPolicyDataplane: PartialPolicyDataplane): PolicyDataplane {
+    return partialPolicyDataplane
+  },
+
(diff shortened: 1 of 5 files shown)


src/app/policies/data/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+  },
+}
+
+export const Policy = {
+  fromObject(partialPolicy: PartialPolicy): Policy {
+    return partialPolicy
+  },
+
+  fromCollection(partialPolicys: PaginatedApiListResponse<PartialPolicy>): PaginatedApiListResponse<Policy> {
```
johncowen, line 31: nit: `policies`

src/app/policies/views/PolicySummaryView.vue
```diff
@@ -100,18 +100,19 @@
 </template>
 
 <script lang="ts" setup>
+import type { Policy } from '../data'
 import DefinitionCard from '@/app/common/DefinitionCard.vue'
 import EmptyBlock from '@/app/common/EmptyBlock.vue'
 import ResourceCodeBlock from '@/app/common/ResourceCodeBlock.vue'
-import type { PolicyEntity, PolicyType } from '@/types/index.d'
+import type { PolicyType } from '@/types/index.d'
```
johncowen, line 107: How come this type isn't coming from `../data`, its an API type right?

johncowen (commented):
I had a couple of tiny things, but generally this looks fine to me.

I remember that there's a tonne of other Policy reshaping happening but that's in Dataplanes is it?

src/app/policies/data/index.ts
kleinfreund: Oops, fixed.

src/app/policies/views/PolicySummaryView.vue
kleinfreund: I haven’t added any transformations for this API because we need next to nothing to change on it. The only thing we use in one place (Dataplane policies) is creating a `policyTypesByName` object from `policyTypesData.policies.reduce((obj, policyType) => Object.assign(obj, { [policyType.name]: policyType }), {})`, but I’m not sure we really need to manage this on the data layer.

And as for the type import: this one is used in a bunch of places and I don’t really want to update all the imports right now seeing also that we don’t need them to be different now. There is probably a couple of things that are going to change with regards to type organization going forward, so I’m not too worried about this one being import like that.

johncowen: Could you re-export it from `../data`?

I want `@/types/index.d` to be gone from our application.

If not don't worry, if you could either make the change or not, then feel free to resolve.

johncowen (approved):
There's an optional thing left but otherwise LGTM!

PR #1907, opened 2023-12-12
refactor(onboarding): use DataSource for dataplanes view

kleinfreund:
Migrates the OnboardingDataplanesView component to use DataSource. This is necessary in order to remove the access of the data layer transformation utility in this file.

### DIFF
--- a/src/app/data-planes/sources.ts
+++ b/src/app/data-planes/sources.ts
@@ -7,8 +7,8 @@ import {
   SidecarDataplane,
 } from './data'
 import type { Can } from '../application/services/can'
-import { defineSources } from '@/app/application/services/data-source'
-import { DataSourceResponse } from '@/app/application/services/data-source/DataSourcePool'
+import type { DataSourceResponse } from '@/app/application'
+import { defineSources, type Source } from '@/app/application/services/data-source'
 import { normalizeFilterFields } from '@/app/common/filter-bar/normalizeFilterFields'
 import { parse, getTraffic } from '@/app/data-planes/data/stats'
 import type { TrafficEntry } from '@/app/data-planes/data/stats'
@@ -19,9 +19,10 @@ import type { PolicyTypeEntry } from '@/types/index.d'
 export type { Dataplane, DataplaneOverview } from './data'
 
 export type DataplaneSource = DataSourceResponse<Dataplane>
+
 export type DataplaneOverviewSource = DataSourceResponse<DataplaneOverview>
-export type DataPlaneCollection = CollectionResponse<DataplaneOverview>
-export type DataPlaneCollectionSource = DataSourceResponse<DataPlaneCollection>
+export type DataplaneOverviewCollection = CollectionResponse<DataplaneOverview>
+export type DataplaneOverviewCollectionSource = DataSourceResponse<DataplaneOverviewCollection>
 
 export type EnvoyDataSource = DataSourceResponse<object | string>
 
@@ -42,8 +43,18 @@ const includes = <T extends readonly string[]>(arr: T, item: string): item is T[
   return arr.includes(item as T[number])
 }
 
-export const sources = (api: KumaApi, can: Can) => {
+export const sources = (source: Source, api: KumaApi, can: Can) => {
   return defineSources({
+    '/dataplanes': (params) => {
+      const { size, page } = params
+      const offset = size * (page - 1)
+      const canUseZones = can('use zones')
+
+      return source(async () => {
+        return DataplaneOverview.fromCollection(await api.getAllDataplaneOverviews({ size, offset }), canUseZones)
+      }, { interval: 1000 })
+    },
+
     '/meshes/:mesh/dataplanes/:name': async (params) => {
       return Dataplane.fromObject(await api.getDataplaneFromMesh(params))
     },
--- a/src/app/data-planes/sources.ts
+++ b/src/app/data-planes/sources.ts
@@ -45,7 +45,7 @@ const includes = <T extends readonly string[]>(arr: T, item: string): item is T[
 
 export const sources = (source: Source, api: KumaApi, can: Can) => {
   return defineSources({
-    '/dataplanes': (params) => {
+    '/dataplanes/poll': (params) => {
       const { size, page } = params
       const offset = size * (page - 1)
       const canUseZones = can('use zones')
--- a/features/onboarding/dataplanes-overview/Index.feature
+++ b/features/onboarding/dataplanes-overview/Index.feature
@@ -3,37 +3,69 @@ Feature: onboarding / dataplanes-overview / index
     Given the CSS selectors
       | Alias         | Selector                               |
       | next-button   | [data-testid='onboarding-next-button'] |
-      | loading       | [data-testid='loading']                |
       | state-success | [data-testid='state-success']          |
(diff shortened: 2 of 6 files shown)


src/app/data-planes/sources.ts
```diff
@@ -155,5 +156,31 @@ export const sources = (api: KumaApi, can: Can) => {
         size,
       }), can('use zones'))
     },
+
+    '/dataplanes/with-status-polling': async (params) => {
```
johncowen, line 160: You don't want `async` here

kleinfreund: Interesting. Why does this change the behavior of DataSource?

---

Unrelated to that issue, I think there is another reason we can’t use the retry mechanism for this as it’s currently built.

In this view, we want to poll `/dataplanes` until we no longer find any offline Dataplanes of the 10 that are fetched. _While_ this takes place, however, we want to show the current Dataplanes. But with the retry mechanism being exception-based, we can’t actually return the data _and_ trigger a retry. I’ve changed the approach to polling in https://github.com/kumahq/kuma-gui/pull/1907/commits/7752e4a7d8ee8234adcbe421392ce53f75405180 to trigger DataSource externally.

Do you think there is a better way?

johncowen: > Why does this change the behavior of DataSource?

It doesn't necessarily, the `sources` originally where a map of URIs to functions returning promises. That promise resolves the emitted `data` for the DataSource. Here the `async` turns the CallableEventSource created via `source()` into `return Promise.resolve(CallableEventSource)` instead of just `return CallableEventSource`. So the DataSource uses the CallableEventSource resolved by the `async`/Promise returned by the function as the `data` emitted by the DataSource.

Largely this is down to us supporting two methods for making `sources`:

1. Just return a promise which resolves your data. This approach is super simple and is enough so the majority of what we need to do is mostly a straight forwards one-liner, i.e. with little code.
2. You can also use `source(() => Promise<>, config)` which returns a CallableEventSource instead of a promise. This option lets you configure the CallableEventSource with more options instead of the defaults that are used if you just return a promise, i.e. polling, retries, backoff and probably lots of other things.

Hope that helps, feel free to ask if the above isn't clear.

---

> Do you think there is a better way?

I had a very quick look at this and the straight forward answer is "yes probably".

I think I added just normal polling when I did the retry stuff, but I'd have to look again, maybe I just added the scaffold for it, or built the retry stuff with polling in mind, I can't remember. All the below depends on whether the polling needs finishing off or not (I just took a quick look and it looks like you can just set `source(() => Promise, {interval: 1000})`)

Off the top of my head, the DataSource should just be set to poll eternally. Then in the view we add the DataSource to the page as usual and let it poll, check the `data.items` until the length hits `10` then remove the DataSource from the page (which will stop the polling).

I'm guessing that once the DataSource is no longer on the page we need to keep hold of its data? Which I'm guessing we can do by keeping the data using `onchange` and doing something in JS land.

---

Apart from the above approach, I've had the much spoken about DataLoader code here now for a while but not wanted to push it as yet. DataLoader maintains the `data` on "disconnection" (i.e an error after a successful load). So with that we could poll, then in the `sources` file check for `data.items.length === 10` and throw an error. The DataLoader will consider that a "disconnection", but we completely control with DataLoader what we show on "disconnection" so we don't have to show an error, and crucially we would still have access to the `data` and the polling would stop without removing the DataLoader from the page.

I would guess you might be able to do something like that ^ with DataSource also, as I think the polling would stop on error, but the `data` and `DataSource` could be kept on the page, maybe try that out and see.

---

Completely separately, how come we poll for all dataplanes instead of just ones in the `default` mesh?

kleinfreund: > Completely separately, how come we poll for all dataplanes instead of just ones in the default mesh?

We don’t know about the user’s meshes in this context and we also don’t care.

johncowen: Just wanted to note, while I appreciate you getting this migrated over, to DataSource and I'm glad we are doing it, we should probably leave this in draft for the moment and continue on with the rest of the data layer migration before revisiting this.

I can see that in order to use the data layer stuff in onboarding it needs to use DataSource (which is a good thing), but we can hold off on worrying about Onboarding until the rest of the migration is done.

Thanks for putting the draft up though!

kleinfreund: Just FYI, I completely changed the approach here to be somewhat close to what you suggested in an earlier comment.

I’m gonna resolve this thread now. If there’s anything related to discuss, let’s use a new one.

johncowen: Gotcha, I just took a very quick look and this is looking much better, thanks!

src/app/data-planes/sources.ts
```diff
@@ -42,8 +43,18 @@ const includes = <T extends readonly string[]>(arr: T, item: string): item is T[
   return arr.includes(item as T[number])
 }
 
-export const sources = (api: KumaApi, can: Can) => {
+export const sources = (source: Source, api: KumaApi, can: Can) => {
   return defineSources({
+    '/dataplanes': (params) => {
```
johncowen, line 48: At least for the moment can we change this to `/dataplanes/poll` so we know at a glance in application-land that its a polling one? Basically something so we know its not a basic "get me all the dataplanes once" one but is also polling.

johncowen (commented):
Looks good, I had one little comment on adding something in the URI for the data source so we know from looking that its a polling one, should be a 👍 after that though.

src/app/data-planes/sources.ts
kleinfreund: Done.

kleinfreund:
I’ve just seen the test that relies on the polling fail. I suspect it can be flaky because the command timeout happens before the polling kicks in. Let me see if I can guard this a little better.

src/app/data-planes/sources.ts
johncowen: Thanks!

Just while we are here, as little further note/potential follow up at some point.

I would really like to change all our sources to use `/dataplanes` (plural) for listings and actions on listings (i.e. `/dataplanes/poll`, `/dataplanes/by/type/` etc etc) and `/dataplane/:name` (singular) for items so we don't get clashes (generally you need to perform actions on listings but not on items)

johncowen:
Gotcha, lemme know

kleinfreund:
> Gotcha, lemme know

The cause was that sometimes, the test saw a partially degraded dataplane first which doesn’t pass the check for showing the waiting state in the UI. That’s fixed now.

Technically, this view wants to do something our status currently doesn’t represent: poll as long as any of the Dataplanes isn’t _connected_. We do so by checking against the offline state, but offline can also happen for disconnect Dataplanes. Maybe we should store the connected status separately from the Dataplane health status.

Anyway, that’s a preexisting issue with this view, so I’m leaving this for the moment.

johncowen:
> Technically, this view wants to do something our status currently doesn’t represent: poll as long as any of the Dataplanes isn’t connected. We do so by checking against the offline state, but offline can also happen for disconnect Dataplanes. Maybe we should store the connected status separately from the Dataplane health status.

Kinda follow what you are saying here, but I'd have to look a little closer to completely get it, maybe make an issue for this one? and we can follow up at some point.

Let me give this one last look.

johncowen (approved):
👍

PR #1908, opened 2023-12-12
refactor(dataplanes): add data transformations for dataplane policies

kleinfreund:
Follow up to #1871.

Move all data transformations happening in the Dataplane Policies tab components into the data layer.

### DIFF
--- a/src/app/data-planes/components/BuiltinGatewayPolicies.vue
+++ b/src/app/data-planes/components/BuiltinGatewayPolicies.vue
@@ -8,14 +8,25 @@
         Gateway policies
       </h3>
 
-      <ul v-if="meshGatewayRoutePolicies.length > 0">
+      <ul v-if="gatewayDataplane.routePolicies.length > 0">
         <li
-          v-for="(policy, index) in meshGatewayRoutePolicies"
+          v-for="(policy, index) in gatewayDataplane.routePolicies"
           :key="index"
         >
           <span>{{ policy.type }}</span>:
 
-          <router-link :to="policy.route">{{ policy.name }}</router-link>
+          <RouterLink
+            :to="{
+              name: 'policy-detail-view',
+              params: {
+                mesh: policy.mesh,
+                policyPath: props.policyTypesByName[policy.type]!.path,
+                policy: policy.name,
+              },
+            }"
+          >
+            {{ policy.name }}
+          </RouterLink>
         </li>
       </ul>
 
@@ -25,7 +36,7 @@
 
       <div>
         <div
-          v-for="(listenerEntry, index) in meshGatewayListenerEntries"
+          v-for="(listenerEntry, index) in gatewayDataplane.listenerEntries"
           :key="index"
         >
           <div>
@@ -50,7 +61,18 @@
                     <div class="dataplane-policy-header">
                       <div>
                         <div>
-                          <b>Route</b>: <router-link :to="routeEntry.route">{{ routeEntry.routeName }}</router-link>
+                          <b>Route</b>: <RouterLink
+                            :to="{
+                              name: 'policy-detail-view',
+                              params: {
+                                mesh: routeEntry.route.mesh,
+                                policyPath: props.policyTypesByName[routeEntry.route.type]!.path,
+                                policy: routeEntry.route.name,
+                              },
+                            }"
+                          >
+                            {{ routeEntry.route.name }}
+                          </RouterLink>
                         </div>
 
                         <div>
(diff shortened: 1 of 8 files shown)


src/app/data-planes/components/BuiltinGatewayPolicies.vue
```diff
@@ -50,7 +61,18 @@
                     <div class="dataplane-policy-header">
                       <div>
                         <div>
-                          <b>Route</b>: <router-link :to="routeEntry.route">{{ routeEntry.routeName }}</router-link>
+                          <b>Route</b>: <RouterLink
+                            :to="{
+                              name: 'policy-detail-view',
+                              params: {
+                                mesh: routeEntry.route.mesh,
+                                policyPath: props.policyTypesByName[routeEntry.route.type]!.path,
```
johncowen, line 69: How come the `!` here?

src/app/data-planes/components/PolicyTypeEntryList.vue
```diff
@@ -20,56 +20,71 @@
...
+              { label: 'On', key: 'name' },
+              { label: 'Conf', key: 'config' },
+              { label: 'Origin policies', key: 'origins' },
+            ]"
             :cell-attrs="getCellAttributes"
             disable-pagination
             is-clickable
           >
-            <template #sourceTags="{ rowValue }">
+            <template #sourceTags="{ row }: { row: PolicyTypeEntryConnection }">
```
johncowen, line 34: I know I've asked before about using AppCollection here I think you said its because we don't need anything that it gives us here. But just incase, if we were to use it does it render anything that we absolutely don't want, or has an interface that we can'f fill out for reasons? i.e. if we were to use AppCollection would anything bad happen?

src/app/data-planes/data/index.ts
```diff
@@ -57,6 +66,50 @@ export type DataplaneOverview = PartialDataplaneOverview & {
...
+  port: number
+  hostName: string
+  routeEntries: MeshGatewayRouteEntry[]
+}
+
+export type MeshGatewayDataplane = PartialMeshGatewayDataplane & {
+  listenerEntries: MeshGatewayListenerEntry[]
+  routePolicies: Meta[]
+}
+
```
johncowen, line 112: I wondered why these types are in `data` not `@/types`. Umm, question, if we had auto-generated types would these ones be auto generated?

johncowen (commented):
I'm guessing this is mostly a lift and shift PR (i.e. just move the things), but I also see a little bit of change apart from that, is it easy to explain why that change was needed here (if I saw that correctly)?

Apart from that I had a couple of questions, but I'm guessing there will be a follow up here to do similar refactors to what we have been doing once its all moved over?

src/app/data-planes/components/BuiltinGatewayPolicies.vue
kleinfreund: This used to use the assertion `as PolicyType` only tell TypeScript that this can’t be `undefined`. Using `!` is producing the same result type-wise and is simpler

src/app/data-planes/components/PolicyTypeEntryList.vue
kleinfreund: It would mostly work, but two things that would be wrong:

- there would be different styles for the first column
- if there is any link in the table (which is the case here), that will be triggered by clicking the row

src/app/data-planes/data/index.ts
kleinfreund: Good question. I changed this a bit so that the data file only defines/exports types it governs (i.e. the transformed types). In other words: the things that would be auto-generated (hopefully anyway) are back in the types file.

kleinfreund:
> I'm guessing there will be a follow up here to do similar refactors to what we have been doing once its all moved over?

Possibly, but I’m not too unhappy with the state of this in this PR. There will probably be room for simplifications in some parts of the functions moved into the data file, but they’re not super important now.

src/app/data-planes/components/PolicyTypeEntryList.vue
johncowen: Gotcha, thanks!

src/app/data-planes/data/index.ts
johncowen: Awesome, sounds great!

johncowen (approved):
Ok yeah I think this PR is good as is as its moving the stuff over from the view to the data layer. I would say there is probably plenty still do be done here and I think all the policy stuff should go in a file on its own somewhere else.

I think/hope moving on from here things will be reduced somewhat and it will be easier to see exactly what should go where.

No new tests on this one though? But I'm ok adding those in a follow up

Lets get this in!

PR #1909, opened 2023-12-12
refactor(dataplanes): add data transformations for inbounds

kleinfreund:
Follow up to #1871.

Add data transformations for inbounds.

### DIFF
--- a/src/app/data-planes/data/index.ts
+++ b/src/app/data-planes/data/index.ts
@@ -7,6 +7,8 @@ import type {
   DataPlaneOverview as PartialDataplaneOverview,
   DiscoverySubscription,
   LabelValue,
+  DataplaneInbound as PartialDataplaneInbound,
+  DataplaneOutbound as PartialDataplaneOutbound,
 } from '@/types/index.d'
 
 export type DataplaneWarning = {
@@ -17,7 +19,20 @@ export type DataplaneWarning = {
   }
 }
 
-export type DataplaneNetworking = PartialDataplaneNetworking
+export type DataplaneInbound = PartialDataplaneInbound & {
+  health: {
+    ready: boolean
+  }
+  addressPort: string
+  serviceAddressPort: string
+}
+
+export type DataplaneOutbound = PartialDataplaneOutbound
+
+export type DataplaneNetworking = Omit<PartialDataplaneNetworking, 'inbound' | 'outbound'> & {
+  inbounds: DataplaneInbound[]
+  outbounds: DataplaneOutbound[]
+}
 
 export type Dataplane = PartialDataplane & {
   networking: DataplaneNetworking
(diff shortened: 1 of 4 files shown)


src/app/data-planes/data/index.ts
```diff
@@ -17,7 +19,20 @@ export type DataplaneWarning = {
   }
 }
 
-export type DataplaneNetworking = PartialDataplaneNetworking
+export type DataplaneInbound = PartialDataplaneInbound & {
+  health: {
+    ready: boolean
+  }
+  addressPort: string
+  serviceAddressPort: string
```
johncowen, line 27: We need to make these consistent I think? Can we use `socketAddress` to mean both address and port? (and `serviceSocketAddress`), this is what I've used over in `Zones`

src/app/data-planes/data/index.ts
```diff
@@ -44,12 +59,28 @@ export type DataplaneOverview = PartialDataplaneOverview & {
...
-    // Address: `${inbound.address ?? props.data.dataplane.networking.advertisedAddress ?? props.data.dataplane.networking.address}:${inbound.port}`
-    // Service address: `${inbound.serviceAddress ?? inbound.address ?? props.data.dataplane.networking.address}:${inbound.servicePort ?? inbound.port}`
+    const { inbound, outbound, ...rest } = partialDataplaneNetworking
+
+    const inbounds: DataplaneInbound[] = (inbound ?? []).map((inbound) => {
+      // An inbound without a health property is considered healthy.
+      const health = { ready: !inbound.health || inbound.health.ready }
+      const addressPort = `${inbound.address ?? partialDataplaneNetworking.advertisedAddress ?? partialDataplaneNetworking.address}:${inbound.port}`
+      const serviceAddressPort = `${inbound.serviceAddress ?? inbound.address ?? partialDataplaneNetworking.address}:${inbound.servicePort ?? inbound.port}`
+
```
johncowen, line 69: For the properties we use to create our new properties, do we want to make sure that the possibly strings are always strings (i.e. `''` if unset)?

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -301,7 +302,7 @@
                     </template>
 
                     <template #body>
-                      <TextWithCopyButton :text="`${inbound.serviceAddress ?? inbound.address ?? props.data.dataplane.networking.address}:${inbound.servicePort ?? inbound.port}`" />
+                      <TextWithCopyButton :text="inbound.serviceAddressPort" />
```
johncowen, line 305: Much nicer not having to worry about all of the previous stuff in the red.

johncowen (commented):
Couple of questions on this one

src/app/data-planes/data/index.ts
kleinfreund: Ehm, but is that even the same concept?

src/app/data-planes/data/index.ts
kleinfreund: You mean `inbound.address`, `partialDataplaneNetworking.advertisedAddress`, etc? What would be the  benefit of that?

src/app/data-planes/data/index.ts
johncowen: Umm I thought so but maybe not? Don't worry for now.

src/app/data-planes/data/index.ts
johncowen: Well at a minimum its always a `string` not a string some of the time. Which is half the point of this line of work.

I think with these types of properties whether to use `string` or `string | undefined` is case by case but I think it makes sense to use `string` for this one. Don't worry though, feel free to resolve if not.

kleinfreund: Arguably `string | undefined` is the safer type because it gives you one more surface to consider the question: Can I just use this?

Sure, with `string | undefined`, a lot of things will be the same as if you’d just use `string` with the "empty" value being the empty string. For example, nothing in Vue’s templating will complain to render out `{{ inbound.address }}` and end up with the string "undefined" on the page. However, in some cases, you do get extra safety through type checks: passing a value of type `string | undefined` to a function accepting only `string` will error. Switching to using the empty string for such values is therefore worse because now it’s only your memory making sure (or not) that you checks first if you can use the value.

I’m going to resolve this.

johncowen (approved):

PR #1912, opened 2023-12-13
refactor(meshes): set-up data transformations

kleinfreund:
Set-up data transformations for the meshes module.

### DIFF
--- a/src/app/meshes/data/index.ts
+++ b/src/app/meshes/data/index.ts
@@ -1,0 +1,80 @@
+import { getDataplaneStatusCounts } from '@/app/data-planes/data'
+import { getServiceTypeCount } from '@/app/services/data'
+import type { PaginatedApiListResponse } from '@/types/api.d'
+import type {
+  Backend,
+  MeshBackend,
+  Mesh as PartialMesh,
+  MeshInsight as PartialMeshInsight,
+} from '@/types/index.d'
+
+export type Mesh = PartialMesh & {
+  config: PartialMesh
+  mtlsBackend: Backend | undefined
+  metricsBackend: Backend | undefined
+}
+
+export type MeshInsight = PartialMeshInsight & {
+  dataplanes: Required<PartialMeshInsight['dataplanes']>
+  dataplanesByType: {
+    standard: Required<PartialMeshInsight['dataplanesByType']['standard']>
+    gateway: Required<PartialMeshInsight['dataplanesByType']['gateway']>
+    gatewayBuiltin: Required<PartialMeshInsight['dataplanesByType']['gatewayBuiltin']>
+    gatewayDelegated: Required<PartialMeshInsight['dataplanesByType']['gatewayDelegated']>
+  }
+  services: Required<PartialMeshInsight['services']>
+  totalPolicyCount: number
+}
+
+export const Mesh = {
+  fromObject(partialMesh: PartialMesh): Mesh {
+    const mtlsBackend = getBackend(partialMesh.mtls)
+    const metricsBackend = getBackend(partialMesh.metrics)
+
+    return {
+      ...partialMesh,
+      config: partialMesh,
+      mtlsBackend,
+      metricsBackend,
+    }
+  },
+}
+
+export const MeshInsight = {
+  fromObject(partialMeshInsight: PartialMeshInsight): MeshInsight {
+    const dataplanes = getDataplaneStatusCounts(partialMeshInsight.dataplanes)
+    const dataplanesByType = {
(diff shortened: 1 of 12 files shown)


src/types/index.d.ts
```diff
@@ -567,10 +567,18 @@ export interface ZoneEgressOverview extends MeshEntity {
...
+    expiration?: string
+  }
+  requestTimeout?: string
+}
+
 export interface Backend {
   name: string
   type: string
-  conf?: Record<string, string | number>
+  conf?: Conf
```
johncowen, line 580: How come we went from `Record<string, string | number>` to `object`?

johncowen (commented):
Few questions but generally looks good to me.

src/app/meshes/data/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  dataplanes: Required<DataplaneStatusCount>
+  dataplanesByType: {
+    standard: Required<DataplaneStatusCount>
+    gateway: Required<DataplaneStatusCount>
+    gatewayBuiltin: Required<DataplaneStatusCount>
+    gatewayDelegated: Required<DataplaneStatusCount>
+  }
+  services: Required<ServiceTypeCount>
+  totalPolicyCount: number
+}
```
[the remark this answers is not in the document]
kleinfreund: I went slightly differently about this now because I also need this for nested structures (i.e. `dataplanesByType`) while still relying on the underlying partial type:

```ts
export type MeshInsight = PartialMeshInsight & {
  dataplanes: Required<PartialMeshInsight['dataplanes']>
  dataplanesByType: {
    standard: Required<PartialMeshInsight['dataplanesByType']['standard']>
    gateway: Required<PartialMeshInsight['dataplanesByType']['gateway']>
    gatewayBuiltin: Required<PartialMeshInsight['dataplanesByType']['gatewayBuiltin']>
    gatewayDelegated: Required<PartialMeshInsight['dataplanesByType']['gatewayDelegated']>
  }
  services: Required<PartialMeshInsight['services']>
  totalPolicyCount: number
}
```

src/types/index.d.ts
kleinfreund: Because that type, it turns out, isn’t correct. The `conf` here is very similar to the conf field we find on, for example, the _rules API. It can contain a bunch of things including nested objects (which the `Record<string, string | number>` type doesn’t allow).

I noticed that while adding the tests: the type checking there told me my input had unexpected properties.

src/app/meshes/data/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  },
+}
+
+function getBackend(meshBackend?: MeshBackend): Backend | undefined {
+  if (meshBackend?.enabledBackend && Array.isArray(meshBackend.backends)) {
+    const enabledBackend = meshBackend.backends.find((backend) => backend.name === meshBackend.enabledBackend)
+
+    if (enabledBackend !== undefined) {
+      return enabledBackend
+    }
```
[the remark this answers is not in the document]
kleinfreund: > Unless I missed something, seeing as `find` returns the thing or `undefined` you should be able to just return the result of `find` here.

Correct, I changed this.

> I think we should make `meshBackend.backends` always an array before we get here also. Once those two are done, seeing as you are pretty much just calling `find` the util function is pretty redundant.

I’m gonna leave this as-is for now. Changing this so that mesh backends that are set  always have `backends` just in order to simplify the `getBackend` function would ultimately be more work/code right now and we don’t access this anywhere else.

src/types/index.d.ts
johncowen: Doesn't `object` include `() => {}`? Should this be `Record<string, unknown>`?

kleinfreund: You’re right. I changed the `Conf` type to be `Record<string, unknown>`.

src/app/meshes/data/index.ts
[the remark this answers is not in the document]
johncowen: Yep no prob, thanks for the simplification, resolving

johncowen (approved):
Good to go!

kleinfreund:
> Good to go!

FYI, I just pushed a small change that adds a `config` property to `Mesh` because I realized https://github.com/kumahq/kuma-gui/pull/1913 was happening and so I fixed it here as well.

johncowen:
Ah yes I just saw that and wondered where else we need to keep an eye out for that! Thanks!

johncowen:
(just looked again in case, still good to go!)

(we should make an e2e test for these things to check that they don't contain one of our augmented properties)

PR #1913, opened 2023-12-14
fix: showing data-transformed objects as YAML

kleinfreund:


johncowen (approved):
😬 sorry I should have spotted/remembered this also! Thanks for fixing up.

PR #1914, opened 2023-12-14
refactor(control-planes): set-up data transformations

kleinfreund:
Set-up data transformations for the control-planes module.

Moves the data source for `/config` into the control-planes module as it belongs to it.


johncowen (commented):
> Moves the data source for /config into the control-planes module as it belongs to it.

Perfect!

---

I know it sounds silly but I was wondering about `config` and `.config`. Imagine if we ever added a property to config, it would turn up in the GUI config panel.

I doubt it will ever happen but maybe lets at leats add a comment in the ControlPlaneConfig.fromObject method to remind someone to make a `.config` property if they ever add any extra props there. wdyt?

kleinfreund:
I think it’s easy enough to leave this as-is and move it around if we change out mind. The way we’ve currently show config makes the likelihood to add properties to it extremely unlikely. Added to that, I don’t think people look at Diagnostics ever. In our testing sessions, I’ve never seen anyone look at it because you don’t need the GUI to show you /config.

johncowen (approved):
Well up to you, its just a comment 🤷 but I don't mind either way, agree its unlikely

PR #1915, opened 2023-12-14
chore: add subscription transformers and use them in zones and dataplanes

johncowen:
This PR has gone back and forth a tiny bit on how we "find" certain things related to subscriptions:

- We no longer assume we should just use the last subscription for anything:
  1. For connection state we look for all subscriptions that don't have a disconnectTime, if we get multiple we take the one with the latest update time.
  2. For subscription config for KDSSubscriptions we simply use a `find` to find the first one with a `config` set and then ignore the rest.
  3. For version we look for all subscriptions with a version property, if we get multiple we take the one with the latest update time.
- We no longer use the `subscription.status.lastUpdateTime` for anything (unless we are just showing that in a subscription list), instead of use the modificationTime of the `*Overview`

Everything uses a new `Subscription` transformer, and we have a special `KDSSubscription` and `DiscoverySubscription` transformers that in turn use the `Subscription` one. I've called these `*Collection` but the more I look at it they could be `SubscriptionInsight.fromObject` which might be quite nice. So I might change that in a follow up at some point.

There are a few e2e test updates here just reflecting the new behaviour, and I amended existing unit tests.

There's probably more work to come here around warnings and probably some other things, but I wanted to stop before I got too carried away in a single PR.


johncowen:
I added the suggestions here, but I also added a little code and comments, lmk!

src/app/subscriptions/data/index.ts
```diff
@@ -1,3 +1,38 @@
...
+  connectedSubscription?: T
+  version?: T['version']
+}
+export const SubscriptionCollection = {
+  fromArray: <T extends Subscription>(items?: T[]): SubscriptionCollection<T> => {
+    const subscriptions = Array.isArray(items) ? items : []
+    // make a copy of the subscriptions and then sort with the latest one first
+    // so we don't change the order of the original array, plus any finding we
+    // do will always gives us the latest subscription if there are multiple
+    const subs = subscriptions.slice()
```
kleinfreund: **Comment**: While it’s true that this will leave the order of the initial array intact, it doesn’t enforce immutability as `Array.prototype.slice()` will only create a shallow copy meaning that `subs[0].name = 'test'` will still mutate `items`.

In short, we could skip this for now if we do this for the objective of enforcing immutability.

kleinfreund (approved):

src/app/subscriptions/data/index.ts
johncowen: Yeah its not deep, but if we then want to add properties to subscriptions that should probably be cloned elsewhere.

Just to note the deep freezing that we are doing in https://github.com/kumahq/kuma-gui/pull/1910 should always tell us if we try to mutate anything, even if there are no tests for it specifically. It caught some mutating in a previous iteration of this code and I think there is more of that elsewhere, those tests will tell us once we get those in.

Al in all, its fine to add when if/we need it as the test approach in https://github.com/kumahq/kuma-gui/pull/1910 should fail if we try to mutate anything original without cloning it first.

PR #1918, opened 2023-12-15
refactor(dataplanes): clean-up getStatus function

kleinfreund:
Moves the status determination logic back into its own function.

Moves the duplicated `isSet` function into its own utility file.

### DIFF
--- a/src/app/data-planes/data/index.ts
+++ b/src/app/data-planes/data/index.ts
@@ -1,4 +1,4 @@
-import { DiscoverySubscriptionCollection } from '@/app/subscriptions/data'
+import { DiscoverySubscriptionCollection, type DiscoverySubscription } from '@/app/subscriptions/data'
 import type { ApiKindListResponse, PaginatedApiListResponse } from '@/types/api.d'
 import type {
   DataPlane as PartialDataplane,
@@ -24,6 +24,7 @@ import type {
   RuleEntryRule,
   SidecarDataplane as PartialSidecarDataplane,
 } from '@/types/index.d'
+import { isSet } from '@/utilities/isSet'
 
 export type { TrafficEntry } from './stats'
 
@@ -140,11 +141,10 @@ export const DataplaneOverview = {
     const networking = DataplaneNetworking.fromObject(partialDataplaneOverview.dataplane.networking)
 
     const dataplaneType = getDataplaneType(networking)
+    const status = getStatus(networking, dataplaneInsight.connectedSubscription)
     const tags = getTags(networking)
     const warnings = getWarnings(dataplaneInsight, tags, canUseZones)
     const isCertExpired = getIsCertExpired(dataplaneInsight)
-
-    const state = isSet(dataplaneInsight.connectedSubscription) ? 'online' : 'offline'
     const services = tags.filter((tag) => tag.label === 'kuma.io/service').map(({ value }) => value)
     const zone = tags.find((tag) => tag.label === 'kuma.io/zone')?.value
 
@@ -155,24 +155,7 @@ export const DataplaneOverview = {
       },
       dataplaneInsight,
       dataplaneType,
-      status: networking.gateway
-        ? state
-        : (() => {
-          const unhealthyInbounds = networking.inbounds.filter((inbound) => !inbound.health.ready)
-          switch (true) {
-            case unhealthyInbounds.length === networking.inbounds.length:
-              // All inbounds being unhealthy means the Dataplane is offline.
-              return 'offline'
-            case unhealthyInbounds.length > 0:
-              // Some inbounds being unhealthy means the Dataplane is partially
-              // degraded.
-              return 'partially_degraded'
-            default:
-              // All inbounds being healthy means the Dataplane’s status is
-              // determined by whether it’s connected to a control plane.
-              return state
-          }
-        })(),
+      status,
       warnings,
       isCertExpired,
       services,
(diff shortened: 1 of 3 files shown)


src/app/data-planes/data/index.ts
```diff
@@ -230,7 +229,7 @@ function getStatus(networking: PartialDataplaneNetworking, connectedSubscription
       return 'partially_degraded'
     default:
       // All inbounds being healthy means the Dataplane’s status is determined by whether it’s connected to a control plane.
-      return state
+      return isConnected ? 'online' : 'offline'
```
johncowen, line 232: How come you changed things to check `isConnected` twice and return `online` or `offline` twice? I think setting `state` once and returning that is better no? It's a little detail, and it was net new code it wouldn't bother me so much, but I think we are changing it here makes it worse, we may as well just leave as is?

I think overall its the same functionality no? Or did I miss something here?

---

Separately (and not for this PR) I wanted to revisit this function seeing as it seems like even though there is no connected subscription, this dataplane can still report a `partially_degraded` state, which feels a little strange to me. If there is no connected subscription I would have thought the state of this dataplane is offline? I dunno maybe I'm wrong?

kleinfreund: I called this `isConnected` because that’s what it represents: the conntectness of the Dataplane. It doesn’t represent the status or state of the Dataplane because, as the function shows, there is more to that than just its connectedness.

johncowen: Its not about what its called, I kinda don't mind its just a local function, its that fact that you do the same conditional twice when there is no need as far as I can see?

kleinfreund: Eh, that doesn’t really matter. The condition isn’t evaluate twice and it reads nicer this way. I’m leaving this as-is.

johncowen:
Hey @kleinfreund I noticed this has been rebased and rejigged a little, just to check, does the current state of this PR following the rebase give us any extra functionality?

kleinfreund:
> does the current state of this PR following the rebase give us any extra functionality?

What makes you ask that? It’s not intended to be a feature and doesn’t add extra functionality. It’s cleaning-up what’s currently a messy inline status logic. It really should be a dedicated function. While I was there, I noticed the `isSet` function was duplicated four times.

johncowen:
> What makes you ask that? It’s not intended to be a feature and doesn’t add extra functionality.

Because I wanted to check, at a glance it looks like it didn't but I could have been wrong so wanted to check. Thanks for confirming.

---

If this adds no additional functionality I think we should just close this one.

kleinfreund:
> If this adds no additional functionality I think we should just close this one.

I disagree which is why instead of closing it I rebased and changed it which you already know.

johncowen:
Hey, sorry but I still haven't changed my opinion here. If anything its gotten even stronger seeing as we have less instances of `isSet` as a result of reusing `Subscriptions` (https://github.com/kumahq/kuma-gui/pull/2081)

Just to be clear: I distinctly moved away from the changes you have re-implemented here as I prefer to see this stuff in place rather than scattered throughout several different functions. I don't know why I would approve a PR that undoes what I purposefully did only a few days previous?

Honestly, I think you should close this PR instead of re-purposing an older one thats no longer relevant or changes any functionality I honestly don't see anything beneficial here, sorry.

kleinfreund:
> Hey, sorry but I still haven't changed my opinion here. If anything its gotten even stronger seeing as we have less instances of `isSet` as a result of reusing `Subscriptions` (#2081)

That doesn’t mean we should routinely duplicate utility functions and type definitions. Moving this function into its own file doesn’t have any downsides and it avoids having it be duplicated all the time. What’s not to like?

> Just to be clear: I distinctly moved away from the changes you have re-implemented here as I prefer to see this stuff in place rather than scattered throughout several different functions.

I don’t. That’s why I’m moving back to having the status determination logic in a separate function. It’s exactly _one_ separate function, not "several different" and it’s perfectly reasonable to separate the data transformation function from the application logic that is determining a Dataplane’s status.

> I don't know why I would approve a PR that undoes what I purposefully did only a few days previous?

Why you would approve it? Because it’s a straightforward improvement. You can now see the code related to the data transformation (in terms of structure) at one glance without needing to scroll past many lines of code that have to do with application logic.

johncowen:
> What’s not to like?

1. It's the tiniest of functions that is currently duplicated once.
2. Now that we are avoiding `null` and we don't need the guard, the code here can just be moved back to `const state = typeof dataplaneInsight.connectedSubscription === 'undefined' ? 'online' : 'offline'` the same as we do elsewhere:

https://github.com/kumahq/kuma-gui/blob/bc136b3f4329f90ac4abac8e307385b9c25143f6/src/app/zones/data/index.ts#L85

This would mean we use the function exactly once, no duplication. I guess even that usage would go at some point.

3. These types of PRs that "prematurely" trying to "optimize the code" by moving everything off into multiple little functions which end up in a large difficult to follow structure (similar to how the app was not so long ago). Additionally this is without waiting to see how things shake out once everything is complete. i.e. we now only use the function once because we've reused work from elsewhere. You see how splitting everything off into tiny little functions, especially early on, can become completely unnecessary?
4. The fact that this PR did something entirely different originally, but instead of just closing it and considering opening a new one, its just been completely amended in order to stay open.
5. The amount of time it takes to discuss these things. this offers no extra functionality whatsoever. We aren't even supposed to be working on this right now. I wouldn't mind so much if there was a functionality change here (see my original question).

---

> It’s exactly one separate function, not "several different"

Screenshot 2024-01-25 at 13 07 59

"Several different" ^. I count 4 previously, 5 with this PR, we should be going down not up.

---

Another thing since I've looked at this again, I also don't like that this function originally (and your revert) had a dependency on `gateway` which is unnecessary. If it's a gateway, just use the state, don't bother calling another function to just return the state.

There's nothing here that I like I'm afraid, it doesn't offer any additional functionality and it puts things back to how we had it previously which we are trying to move away from.

If you want to talk about this more, please can we take it to a different discussion offline where we can come to a decision quicker.

johncowen (approved):
Let go with this

PR #1920, opened 2023-12-18
chore: minor amends to dataplane viz

johncowen:
- Adds `meshtrace_zipkin` to excluded list
- Switches tx/rx to the reverse on all cards as we are using `downstreams` for everything. (I'm guessing I need to do it for everything and not just outbounds, please let me know if not)

### DIFF
--- a/src/app/data-planes/sources.ts
+++ b/src/app/data-planes/sources.ts
@@ -62,7 +62,10 @@ export const sources = (api: KumaApi, can: Can) => {
       // inbounds is anything starting with `localhost_`
       const inbounds = getTraffic(json, (key) => key.startsWith('localhost_'))
 
-      // outbounds are anything else unless it starts with something in the below list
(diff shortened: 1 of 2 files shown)


src/app/data-planes/sources.ts
```diff
@@ -74,6 +77,7 @@ export const sources = (api: KumaApi, can: Can) => {
           'outbound_passthrough',
           'access_log_sink',
           'ads_cluster',
+          'meshtrace_zipkin',
```
lahabana, line 80: We have https://github.com/kumahq/kuma/pull/8654 so looks like you can already add a blanket (starts with `_`)

lahabana (approved):

src/app/data-planes/sources.ts
johncowen: Awesome 🎉

PR #1921, opened 2023-12-18
chore: apply service names to all types of inbounds/outbounds

johncowen:
This PR is a bit of a prequel to some other work I will be doing on the DataPlane viz, in that we would like to match the inbound `localhost_8080` inbound name with the service name of the inbound. In order to do that this PR first adds a `service` property/field to our DataplaneInbounds derived from the `kuma.io/service` tag.

I noticed that this reshape was almost already there but only for unhealthyInbounds, and unhealthyInbounds only contain the `service` and `port` and have all their other properties removed.

I've changed the reshaping to:

1. Add a `service` field to all inbounds and outbounds (derived from the `kuma.io/service` tag)
2. I moved things around a little as I realised I could remove the `getStatus` and `getUnhealthyInbounds` functions which where doubling/tripling up various `.filter`ing for unhealthiness.
3. Use `Array.isArray` for anything that should always be arrays instead of using truthiness `??` operators. This point applies to `inbounds` and `outbounds`
4. I figured similarly for point 3, I could use `isSet` for the healthiness check for inbounds.

Once I'd done the above I had to make a few amends to:

1. The types, because I have told the type system that `kuma.io/service` is always set on inbounds and outbounds.
2. The mocks, reflecting the type changes in point 1, plus some places where it was no longer ok to pass through the entire inbound (namely with `t` usage)
3. The unit tests for reflecting the slightly changed inbounds

Please let me know if there are any questions.

### DIFF
--- a/src/app/data-planes/data/index.spec.ts
+++ b/src/app/data-planes/data/index.spec.ts
@@ -192,6 +192,7 @@ describe('dataplanes data transformations', () => {
                     'kuma.io/service': '',
                     'kuma.io/zone': '',
                   },
+                  service: '',
                   port: 1,
                   addressPort: 'http://example.org:1',
                   serviceAddressPort: 'http://example.org:1',
@@ -380,6 +381,7 @@ describe('dataplanes data transformations', () => {
                   tags: {
                     'kuma.io/service': '',
                   },
+                  service: '',
                   port: 1,
                   addressPort: 'http://example.org:1',
                   serviceAddressPort: 'http://example.org:1',
@@ -485,6 +487,7 @@ describe('dataplanes data transformations', () => {
                   tags: {
                     'kuma.io/service': '',
                   },
+                  service: '',
                   port: 1,
                   addressPort: 'http://example.org:1',
                   serviceAddressPort: 'http://example.org:1',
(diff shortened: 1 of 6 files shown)


src/app/data-planes/data/index.spec.ts
```diff
@@ -659,10 +674,21 @@ describe('dataplanes data transformations', () => {
           },
           dataplaneType: 'standard',
           status: 'partially_degraded',
-          unhealthyInbounds: [{
-            port: 1,
-            service: 'service-1',
-          }],
+          unhealthyInbounds: [
```
lahabana, line 677: This unhealthyInbouds stuff is a little odd no? Could it just be a funciton on the object?

src/app/data-planes/data/index.ts
```diff
@@ -78,29 +81,31 @@ export type MeshGatewayDataplane = PartialMeshGatewayDataplane & {
...
 
     return {
       ...rest,
-      inbounds,
-      outbounds,
+      inbounds: inbounds.map((item) => {
+        return {
+          ...item,
+          // If a health property is unset the inbound is considered healthy
+          health: { ready: !isSet(item.health) ? true : item.health.ready },
```
lahabana, line 97: In 2.6 we also have a `State`: https://github.com/kumahq/kuma/blob/eb4b3d241742ce75950008b56daa58a3a9c402ad/api/mesh/v1alpha1/dataplane.proto#L162

lahabana (approved):

src/app/data-planes/data/index.spec.ts
johncowen: Yeah its very odd.

> Could it just be a funciton on the object?

Exactly, well I plan on just using `inbounds.filter(inbound => !inbound.heath.ready)` whereever we use it instead of having a `unhealthlyInbounds` property.

When we come back to work on this stuff more in earnest I'll be looking to alter this.

Theres a bit of tech debt here to sort out still and I'm still picking my way through it. This PR at least makes unhealthyInbounds have the same properties as a normal inbound.

src/app/data-planes/data/index.ts
johncowen: Ah perfect! I noticed that the other day but didn't realize how new it was 🎉 , much better!

PR #1932, opened 2023-12-19
feat(dataplanes): Use service name for naming the inbounds instead of localhost

johncowen:
There's a few things going on here, but I mainly wanted to get something up for a PR preview to get some feedback on what we are rendering here.

Please make sure you have `KUMA_TRAFFIC_ENABLED=true` before previewing:

[Preview](https://deploy-preview-1932--kuma-gui.netlify.app/gui/meshes/default/data-planes/frontend-ac58fd4fbc-hjaqs.firewall.kuma-system-proxy-1/overview)

Screenshot 2023-12-22 at 11 35 55

Additionally:

- There is also a fix in here for showing the aggregate passthrough traffic correctly.
- Overall there is quite a lot of work here in beginning to flesh out the mock for the `stats` endpoint which includes making sure our stateless mocks use "fake state" by setting the seed used for randomization based on the stateful URL, therefore making sure the randomness produces the same results on two different API URLs.

Important note:

~This is unfinished/WIP as of right now I'm looking to make sure I'm rendering what we want correctly etc. Following that I'll probably make some changes before taking out of draft. As an extra important note, once I'm super sure what data we want and what our reshaping needs might be, I would guess we might move some of this shaping to the data layer. For the moment, in order to have something for feedback, I'm kinda overusing `find`s etc where in under other circumstances I might not choose to do that.~

I've amended the rendering above slightly to what we want here so I'm taking this PR out of draft, but I have not yet changed the way we pick out the data we want (so I'm still overusing `find`s). I'm pretty sure I want to reshape our inbounds everywhere to be an object keyed by port instead of an array, that way we can just refer to the port of the inbound to find the thing rather than having to search through an array. I rather do that in a follow up though especially considering this feature is currently feature flagged, so I'm taking this out of draft for review.

### DIFF
--- a/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -145,7 +145,21 @@
                         :tx="item[meta.protocol]?.[`${meta.direction}_cx_rx_bytes_total`] as (number | undefined)"
                         :requests="meta.protocol === 'http' ? ['http1_total', 'http2_total', 'http3_total'].reduce((prev, key) => prev + (item.http?.[`${meta.direction}_rq_${key}`] as (number | undefined) ?? 0), 0) : undefined"
                       >
-                        {{ item.name }}
+                        <template
+                          v-for="port in [
+                            item.name.split('_')[1],
+                          ]"
+                          :key="port"
+                        >
+                          <template
+                            v-for="inbound in [
+                              props.data.dataplane.networking.inbounds.find(item => `${item.port}` === `${port}`),
+                            ]"
+                            :key="inbound"
+                          >
+                            {{ inbound?.tags['kuma.io/service'] ?? `` }}:{{ port }}
+                          </template>
+                        </template>
                       </ServiceTrafficCard>
                     </template>
                   </template>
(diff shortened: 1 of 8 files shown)


johncowen:
Following a little sync, work to be done here before taking out of draft:

1. Only use `:port` on the inbounds
2. Add a clickable `kuma.io/service` underneath the protocol in inbounds

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -138,15 +138,33 @@
...
+                        v-for="port in [
+                          item.name.split('_')[1],
+                        ]"
+                        :key="port"
                       >
-                        {{ item.name }}
-                      </ServiceTrafficCard>
+                        <template
+                          v-for="inbound in [
+                            props.data.dataplane.networking.inbounds.find(item => `${item.port}` === `${port}`),
```
johncowen, line 149: This is the bit I will change in another PR. We are picking off the `_portnumber` from the stat, then searching for that across all inbounds, for every inbound. Whilst we mostly have 1 inbound I'd rather do this differently as mentioned in the description for the PR.

johncowen:
Theres more work to come on this to address how we access/find data we need for this, but I wanted to take this one out of draft and get approval/merge as it is with a quick follow up for the data improvements.

lukidzi (approved):
Looks really good. We can in the future think about grouping by protocol in different colors, some collapsing maybe to not show all when there is many of them

PR #1936, opened 2023-12-22
chore(dataplanes): remove unhealthyInbounds property

johncowen:
Related:

- https://github.com/kumahq/kuma-gui/pull/1932
- https://github.com/kumahq/kuma-gui/pull/1921#pullrequestreview-1791223339

I'm 99% sure I want to store inbounds somewhere as an object keyed by port rather than an array. I'm just finishing review the code involved with inbounds, but I'm pretty sure I'm going to add a pre-shaped `.inbounds` property somewhere with this different shape.

---

We also noticed that we currently have a reshaped `unhealthyInbounds` property with no corresponding `healthyInbounds` property. Its also much easier using JS `filter` to find the inbounds that are unhealthy since we did the data layer work.

This PR removes all usage of the `unhealthyInbounds` property and replaces it with `inbounds.filter`.

I also noticed a semi-bug rendering issue with the tooltips:

### Before:
Screenshot 2023-12-22 at 12 32 24

### After:

Screenshot 2023-12-22 at 12 33 02

So I fixed this up while I was here.

Lastly I noticed that we had one last places in `dataplanes` where we weren't using the `version` from `SubscriptionCollection.fromArray` and instead using `.at(-1)` which is just due to order of PRs so I switched that up here also.


src/app/data-planes/data/index.ts
```diff
@@ -86,6 +85,8 @@ const DataplaneNetworking = {
     const { inbound, outbound, ...rest } = networking
 
     const inbounds = Array.isArray(inbound) ? inbound : []
+    // TODO(jc): Confirm whether we still ever get outbounds, from what I understand
+    // we no longer get this property
```
jakubdyszkiewicz, line 89: we can if we have Universal DPP without transparent proxying

jakubdyszkiewicz (approved):

src/app/data-planes/data/index.ts
johncowen: Ah perfect! Thanks Jakub! Let me replace this with an explanatory comment

johncowen:
I just made a little change to the TODO comment [here](https://github.com/kumahq/kuma-gui/pull/1936#pullrequestreview-1801866285) to make it an explanatory one, gonna go ahead and merge, ty!

PR #1956, opened 2024-01-04
chore: rollout ResourceCodeBlock using DataSources

johncowen:
Rollout of https://github.com/kumahq/kuma-gui/pull/1873 across the entire application.


src/app/common/ResourceCodeBlock.vue
```diff
@@ -117,9 +107,7 @@ const fetcher = async (_params?: SingleResourceParameters): Promise<Entity> => {
 }
 
 async function getYamlAsKubernetes() {
-  const resourceKubernetes = await (typeof props.resourceFetcher !== 'undefined' ? props.resourceFetcher({ format: 'kubernetes' }) : fetcher())
-
-  return toYamlRepresentation(resourceKubernetes)
```
johncowen, line 122: All of the `resourceFetcher` stuff here is the old stuff which was kept in so we could support both implementations at once (both old and new). Now we have rolled everything out, we don't need this older  approach.

src/app/services/routes.ts
```diff
@@ -48,13 +48,6 @@ export const routes = () => {
             module: 'services',
           },
           component: () => import('@/app/services/views/ServiceListView.vue'),
-          children: [
-            {
-              path: ':service',
-              name: 'service-summary-view',
-              component: () => import('@/app/services/views/ServiceSummaryView.vue'),
-            },
-          ],
```
johncowen, line 57: I noticed we don't use `ServiceSummaryView` as clicking on a Service row takes you straight to a detail page instead of a summary drawer. So instead of updating the ResourceCodeBlock in a file we don't use, I just deleted the file (plus a related test, plus the route configuration you can see here)

johncowen (commented):
Inlines to help review:

features/mesh/services/ServiceSummary.feature
```diff
@@ -1,45 +0,0 @@
-Feature: Service summary
```
slonka, line 1: edit: we don't use this view

slonka (approved):

PR #1957, opened 2024-01-04
chore: refactor UpgradeCheck to use DataSource

johncowen:
Moves UpgradeCheck to use DataSource.

The DataSource now gets the latest version, and then we check in the template to see if that version is the same as the version we are running using `KUMA_VERSION`. I used all the same logic sort of code in the DataSource `sources.ts` file.

---

I will probably remove the `UpgradeCheck.vue` component entirely in a later PR seeing as it doesn't really contain anything anymore and we only use it once in `ApplicationShell` here https://github.com/kumahq/kuma-gui/blob/b4ff4d670eab3d349ff1524014f9ff423fab3ec3/src/app/kuma/components/ApplicationShell.vue#L25. `ApplicationShell` itself is also only used once, so it makes sense to just put the code from `UpgradeCheck.vue` into it so the code is all front and center.

### DIFF
--- a/src/app/common/UpgradeCheck.vue
+++ b/src/app/common/UpgradeCheck.vue
@@ -1,9 +1,10 @@
 <template>
-  <div
-    class="upgrade-check"
+  <DataSource
+    v-slot="{ data }"
+    :src="`/control-plane/version/latest`"
   >
     <KAlert
-      v-if="showNotice"
+      v-if="env('KUMA_VERSION') !== data?.version"
       data-testid="upgrade-check"
       class="upgrade-check-alert"
       appearance="info"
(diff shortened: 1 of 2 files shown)


src/app/common/UpgradeCheck.vue
```diff
@@ -26,67 +27,17 @@
...
-    const comparison = compare(latestVersion.value, currentVersion)
-    showNotice.value = comparison === 1
-  } else {
-    const timespan = 3 // months
-    const today = new Date()
-    const refDate = new Date('2020-06-03 12:00:00')
-    const later = new Date(refDate.getFullYear(), refDate.getMonth() + timespan, refDate.getDate())
-
-    // compare dates and handle the notice accordingly
-    showNotice.value = today.getTime() >= later.getTime()
```
johncowen, line 83: I didn't bother replicating this stuff. I chased this back a long time ago after changing some code here a while back and it seems weird and I think it was original meant to store something in localstorage.

Essentially it shows the upgrade message if we can't retrieve the remote version number, which I do in this PR now but without all the date checking stuff.

slonka (approved):

PR #1958, opened 2024-01-04
feat(dataplanes): add showInactive toggle for dataplane traffic view

johncowen:
[Preview](https://deploy-preview-1958--kuma-gui.netlify.app/gui/meshes/default/data-planes/frontend-ac58fd4fbc-hjaqs.firewall.kuma-system-proxy-1/overview#KUMA_TRAFFIC_ENABLED=true)

Adds a "Show inactive" toggle so you can toggle between either showing the viz with services that have no traffic/not showing services that have no traffic.

Note: We decided to put a straightforwards toggle at the top of the list instead of a card at the bottom with the thinking that when you have lots of services you won't want to scroll to the bottom of the list every time you want to use the toggle.

---

I added some unrelated work here to make it possible to set feature flags via a deeplink (dev and PR previews only), I split that off into a separate commit.

---

This PR doesn't persist the setting into the URL/queryParam just yet. I've done this separately as there's quite a bit of unrelated change and will come in an upcoming PR.

### DIFF
--- a/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -175,6 +175,14 @@
               </DataPlaneTraffic>
               <DataPlaneTraffic>
                 <template #actions>
+                  <KInputSwitch
+                    v-model="showInactive"
+                  >
+                    <template #label>
+                      Show inactive
+                    </template>
+                  </KInputSwitch>
+
                   <KButton
                     appearance="primary"
                     @click="refresh"
@@ -230,7 +238,7 @@
                       :key="protocol"
                     >
                       <ServiceTrafficCard
-                        v-if="(protocol !== 'http' ? item.tcp?.downstream_cx_rx_bytes_total : item.http?.downstream_rq_total) as (number | undefined ) ?? 0 > 0"
+                        v-if="showInactive || ((protocol !== 'http' ? item.tcp?.downstream_cx_rx_bytes_total : item.http?.downstream_rq_total) as (number | undefined ) ?? 0 > 0)"
                         :protocol="protocol"
                         :traffic="item"
                       >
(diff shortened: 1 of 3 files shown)


lahabana:
> Note: We decided to put a straightforwards toggle at the top of the list instead of a card at the bottom with the thinking that when you have lots of services you won't want to scroll to the bottom of the list every time you want to use the toggle.

Was there input from design on this?

johncowen:
Unfortunately not no. Agree its not best so would be good to get a bit of design thought sooner rather than later. I think the idea was that we should just get something functional in whilst waiting to iterate on the look. We can talk more in today's sync.

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -175,6 +175,14 @@
               </DataPlaneTraffic>
               <DataPlaneTraffic>
                 <template #actions>
+                  <KInputSwitch
+                    v-model="showInactive"
```
johncowen, line 179: If https://github.com/kumahq/kuma-gui/pull/1966 goes in first, this v-model will be switched ot just use `route.params.inactive`, which following #1966 has a two way binding. If #1966 goes in last, there'll be another PR to change this up.

src/services/env/CookiedEnv.ts
```diff
@@ -1,16 +1,36 @@
...
+    const hashEnv = strToEnv(document.location.hash.substring(1))
+    doc.location.hash = ''
+    if (hashEnv.length > 0) {
+      doc.cookie = hashEnv.map(([key, value]) => `${key}=${value}`).join(';')
+    }
+  }
+  const cookies = strToEnv(doc.cookie).reduce((prev, [key, value]) => {
+    prev[key] = value
+    return prev
+  }, {} as Record<string, string>)
```
johncowen, line 33: The changes here are just to pick up any `#KUMA_THING=something` in any links and set the cookie so we can deep link and enable feature flags at the same time

kleinfreund (approved):

PR #1968, opened 2024-01-08
fix(policies): policy type list being too wide

### LINKED ISSUE #1962: Policies sidebar is too wide on bigger screens
### What happened?

image

kleinfreund:
Resolves #1962.

Not sure where/how this was introduced. Could be related to the KCard changes?


johncowen (approved):
> Could be related to the KCard changes?

Yeah, I think you're right there, thanks for the fix!

PR #1982, opened 2024-01-09
feat(dataplanes): add zone column

kleinfreund:
Adds a Zone column to Dataplane lists.

### DIFF
--- a/features/mesh/dataplanes/Index.feature
+++ b/features/mesh/dataplanes/Index.feature
@@ -32,6 +32,7 @@ Feature: mesh / dataplanes / index
                     ready: true
                   tags:
                     kuma.io/service: service-1
+                    kuma.io/zone: zone-1
           dataplaneInsight:
             mTLS:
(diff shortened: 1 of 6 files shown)


johncowen (commented):
I'm just starting to look at this now but I saw that in https://github.com/kumahq/kuma-gui/pull/1728 where this was removed there was an entry in the description, specifically saying "remove zone columns"

> Removes the protocol and zone columns.

I'd originally thought this disappearing was an accident but seems like it was on purpose, can you remember why we removed it originally? Was it just to make the table have less columns or for some other reason?

lmk, gonna look over this in the meantime.

johncowen (approved):

johncowen:
> I don’t totally recall how we motivated this, but I can imagine a reason was to free up space in the table. By itself, the Zone column isn’t particularly useful; however, the issue this PR is resolving provides one: the zone no longer appearing as part of the Dataplane name.

Cool, thanks!

PR #1993, opened 2024-01-10
chore: fix flaky tests

kleinfreund:
**chore(policies): fix flaky test**

Fixes a flaky policy test caused by sourcing a route parameter from a prop (see https://github.com/kumahq/kuma-gui/pull/1993#discussion_r1447134988 for details).

Avoids double navigations (specifically in the "Clicking the link goes to the detail page and back again" and "Clicking policy types in the sidebar switches listing" tests).

**chore(dataplanes): fix flaky test**

Fixes a flaky test by avoiding quick double navigations (specifically in the "Clusters tab has expected content" and "Policies tab has expected content" tests).

### DIFF
--- a/src/app/policies/views/PolicySummaryView.vue
+++ b/src/app/policies/views/PolicySummaryView.vue
@@ -24,12 +24,14 @@
               :to="{
                 name: 'policy-detail-view',
                 params: {
-                  policy: props.name,
+                  mesh: route.params.mesh,
+                  policyPath: route.params.policyPath,
+                  policy: route.params.policy,
                 },
               }"
             >
               <RouteTitle
-                :title="t('policies.routes.item.title', { name: props.name })"
+                :title="t('policies.routes.item.title', { name: route.params.policy })"
               />
             </RouterLink>
           </h2>
@@ -112,7 +114,6 @@ import EmptyBlock from '@/app/common/EmptyBlock.vue'
 import ResourceCodeBlock from '@/app/common/ResourceCodeBlock.vue'
 
 const props = withDefaults(defineProps<{
-  name: string
   policy?: Policy
   policyType: PolicyType
 }>(), {
--- a/features/mesh/dataplanes/DataplaneDetailsBuiltinGateway.feature
+++ b/features/mesh/dataplanes/DataplaneDetailsBuiltinGateway.feature
@@ -4,14 +4,14 @@ Feature: Dataplane details for built-in gateway
       | Alias             | Selector                                                           |
       | detail-view       | [data-testid='data-plane-detail-tabs-view']                        |
       | policies-view     | [data-testid='data-plane-policies-view']                           |
-      | overview-tab      | #data-plane-detail-view-tab a                                      |
-      | policies-tab      | #data-plane-policies-view-tab a                                    |
       | warnings          | [data-testid='dataplane-warnings']                                 |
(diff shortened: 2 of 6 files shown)


src/app/policies/views/PolicySummaryView.vue
```diff
@@ -24,12 +24,14 @@
               :to="{
                 name: 'policy-detail-view',
                 params: {
-                  policy: props.name,
+                  mesh: route.params.mesh,
+                  policyPath: route.params.policyPath,
+                  policy: route.params.policy,
```
kleinfreund, line 29: **Comment**: Turns out this was the fix for the flaky policy test I was actually after. Specifically, the `policy` param was sometimes missing and causing failures in the test when sourcing it via `props.name`. It’s a little strange because `props.name` in turn was sourced from the exact same `route.params.policy` (just in a different component). I suppose this has to do with components mounting/unmounting and `props.name` becoming unavailable.

I’ll keep the changes in the policies’ Index.feature because the following didn’t really make sense:

```
    When I click the "$item:nth-child(1) td:first-of-type a" element
    Then I click the "$item:nth-child(1) [data-testid='details-link']" element
```

There’s no reason to click the name link (which will open the summary) only to immediately click the details link.

Also, because the initial navigation (`When I visit the "/meshes/default/policies/circuit-breakers" URL`) happened once in the test setup and once in a scenario which caused double navigation. Not a problem by itself, but it does make the test slower for no reason.

johncowen: Cool yeah I like that we prefer to take primitive values from the URL for `*View.vue` components rather than keep passing them down 👍

johncowen (approved):
This LGTM the only question I had was:

I noticed that we changed from clicking the tab to go to the policies tab, to just visiting with a direct URL. Is that right? and are there any other places where we might have slightly changed the tests here? (it was a little hard to figure from the diff on its own)

---

I vaguely remember that we have tests somewhere that just click around navigation without to make sure clicking around as opposed to visiting URLs works.

I'm just looking to see if we have ones that cycle through tabs, have you seen any anywhere and if so does it cycle through the policy tab (...and if not do you think its worth adding something like that separately?)

kleinfreund:
> I noticed that we changed from clicking the tab to go to the policies tab, to just visiting with a direct URL. Is that right? and are there any other places where we might have slightly changed the tests here? (it was a little hard to figure from the diff on its own)
> 
> I vaguely remember that we have tests somewhere that just click around navigation without to make sure clicking around as opposed to visiting URLs works.
> 
> I'm just looking to see if we have ones that cycle through tabs, have you seen any anywhere and if so does it cycle through the policy tab (...and if not do you think its worth adding something like that separately?)

Basically, I wanted to avoid unnecessarily navigating to a view that we’re not testing in all scenarios.

We could consider adding a generic test in our application tests that just try all the navigation tabs once so we can _always_ ignore doing this in the module-specific tests.

As of this PR, we won’t have this covered for the Policies tab. If I have some time tomorrow, I’ll add a general one and remove unnecessary tab navigations in other tests.

johncowen:
Sounds great!

> We could consider adding a generic test in our application tests that just try all the navigation tabs once so we can always ignore doing this in the module-specific tests.

yeah that ^ 👍 

thanks!

Issue #1960, opened 2024-01-08
Implement tray for DPP outbounds


lahabana:
### Description

Let's discuss here further details

johncowen:
See:

- https://github.com/kumahq/kuma-gui/pull/1997
- https://github.com/kumahq/kuma-gui/pull/2015
- https://github.com/kumahq/kuma-gui/pull/2041

johncowen:
The above 3 PRs are merged, there will be follow up work, but that will be cleanup and any refactoring of things I noticed while I was in "feature mode".

Closing this

Issue #1959, opened 2024-01-08
inbound tray for DPP details


lahabana:
### Description

Should show entries with labels and stats.

johncowen:
The above 3 PRs are merged, there will be follow up work, but that will be cleanup and any refactoring of things I noticed while I was in "feature mode".

Closing this

PR #1997, opened 2024-01-10
feat(dataplanes): add basic dataplane viz summary drawers

johncowen:
Adds initial inbound/outbound summary drawers for the dataplane viz:

Mostly a copy paste from previous impl. of drawers apart from:

- Two way binding boolean route params no only execute once in our RouteView tree (only executing for the current route)
- ~I could not get the KSlideout not to close on clicking another link, even though I copied the impl. exactly as we have it elsewhere. Therefore I added a shim following the impl. in Kongponents but adding an additional check for links or "linked divs". I installed `vueuse/core` for this which is also a dependency of kongponents and used its onClickOutside plus a throttle to ignore the first-most emitted event when a div is clicked which is immediately followed by the `a.click()` event. If theres a better way to reliably solve this any suggestions would be appreciated.~ We changed this to a different approach
- `inbounds` and `outbounds` now have a `protocol` property derived from the tag but defaulting to `tcp`.

The contents of the drawers is purposefully very minimal until we know exactly what we want to show in those, I stopped short of filling the straightforwards things out as I wanted to talk about the usefulness of showing the exact same information in the drawer as we can already see in the viz itself (i.e. number of 2xx, 5xx requests etc). So this PR is partly technical but also to start off that conversation.

~Oh, lastly, I added a `@ts-ignore` for a typescript error I was getting in CI but not locally. I'm more concerned about the "works on my machine" aspect of this, so if you could remove the `@ts-ignore` and lemme know if it also "works on your machine" that'd be great!~ <== this solved itself as we stopped using vueuse's `onClickOutside`, still weird that it failed in CI but not locally.

Related:

- #1959 
- #1960


kleinfreund:
> * I could not get the KSlideout not to close on clicking another link, even though I copied the impl. exactly as we have it elsewhere. Therefore I added a shim following the impl. in Kongponents but adding an additional check for links or "linked divs". I installed `vueuse/core` for this which is also a dependency of kongponents and used its onClickOutside plus a throttle to ignore the first-most emitted event when a div is clicked which is immediately followed by the `a.click()` event. If theres a better way to reliably solve this any suggestions would be appreciated.

I’ve been looking into this. As far as I can tell, there are two things influencing this issue:

- the `click` event listener in `ServiceTrafficCard` fires twice when clicking on a card. That’s because `$a.click()` inside said listener will itself trigger the `click` listener a second time. A `if (!e.isTrusted) { return }` at the start of the listener will prevent this.
- When clicking on a (different) card while a tray is already open, what I see is the following:
  - the SummaryView’s close mechanism is triggered (i.e. a `route.replace` to the detail view route). **Note that the associated `router.push` in RouteView did _not_ happen, yet**.
  - the card’s click mechanism is triggered (i.e. to open the summary tray)
  - The `router.push` of the card’s close mechanism happens (it’s delayed because of the use of requestAnimationFrame)

I confirmed this by (1) removing the amends to KSlideout and (2) changing `click` as follows (i.e. adding the aforementioned `isTrusted` check and also using a requestAnimationFrame callback):

```ts
  if (!e.isTrusted) {
    return
  }

  const $el = (e.target as HTMLElement).closest('.service-traffic-card')
  if ($el) {
    const $a = $el.querySelector('a')
    if ($a !== null) {
      window.requestAnimationFrame(() => {
        console.log('triggered manual click for summary')
        $a.click()
      })
    }
  }
}
```

This does seem to fix the issue you’re describing.

Now, this isn’t exactly nice either. I strongly dislike the need to introduce a event loop timing mechanism just because we also use it in our router glue code. Manually triggering link clicks is also not great.

But I think it’s more important we don’t try to patch KSlideout like that.

Perhaps we can refine this approach more. Perhaps if we managed to avoid triggering the SummaryView close mechanism **if** we’re clicking on other cards we can avoid the RouteView timing issue.

johncowen:
Cool thanks I went with that ^ instead

src/app/common/SummaryView.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <KSlideout
+    ref="slideOutRef"
```
kleinfreund, line 3: **Suggestion**: This is no longer needed.

```suggestion
```

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
```diff
@@ -110,5 +125,11 @@ const props = defineProps<{
   display: flex;
   flex: 1 1 auto;
 }
-
+.title :deep(> a) {
+  text-decoration: none;
+  color: inherit;
+}
+.data-card:has(a) {
```
kleinfreund, line 132: **Question**: Should we be using `:has` already? It’s a really new feature that is only supported in Firefox as of version 121 which is the current stable release.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
```diff
@@ -110,5 +125,11 @@ const props = defineProps<{
   display: flex;
   flex: 1 1 auto;
 }
-
+.title :deep(> a) {
+  text-decoration: none;
```
kleinfreund, line 129: **Question**: How come we’re removing the hover underline style here as well?

src/app/data-planes/views/DataPlaneInboundSummaryView.vue
```diff
@@ -0,0 +1,87 @@
...
+    <AppView>
+      <template #title>
+        <div class="summary-title-wrapper">
+          <img
+            aria-hidden="true"
+            src="@/assets/images/icon-wifi-tethering.svg?url"
+          >
+
+          <h2 class="summary-title">
+            Inbound :{{ route.params.service }}
```
kleinfreund, line 18: **Question**: I noticed this in a few PRs recently: is it intentional that you’re not putting UI texts into the message files?

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
johncowen: Well I'm glad you asked! This only adds a pointer cursor instead of the default. If it was something more visible/important I wouldn't have used it.

I think more generally, I suppose it depends on what our browser support is? It would be good to figure out some sort of rule to follow.

Safari's support is from May last year (which also introduced `Array.at`), Chrome's from September last year. Firefox is the most recent (but also I would guess less used) at December only a month back.

Back to the less general issue of using `:has` to change the cursor to a pointer, I'm not too concerned, but would be fine to change it to use a different approach instead, lmk.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
johncowen: Because this is what we do with all our clickable blocks. Actual in-text links should be blue with underline on hover. Clickable blocks shouldn't have titles that look like in-text links i.e. they should look like more like buttons not text-links.

src/app/data-planes/views/DataPlaneInboundSummaryView.vue
johncowen: Oh sorry, no its just that this is mainly still in prototyping mode. So I'm not thinking about i18n etc. until we are sure we know what we are showing, otherwise we'll accidentally end up with loads of unused i18n texts (we have accumulated a few of those already I think). I think we are just about coming out of prototype mode, so I'll probably have PR that fills all these in pretty soon.

Oh you've just reminded my of another question I had, Its unrelated to this though so I'll make a new comment.

src/app/common/SummaryView.vue
johncowen: Oh thanks!

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
kleinfreund: > Because this is what we do with all our clickable blocks.

Aren’t these the _only_ type of that kind of element?

> Clickable blocks shouldn't have titles that look like in-text links i.e. they should look like more like buttons not text-links.

But why? These titles don’t look like buttons (and rightfully so, they should absolutely not look like buttons) nor do they act like buttons. They act like links.

I guess what I’m saying is it’s not really clear why there is a special rule for links in certain UI fragments to have no underline on hover.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
kleinfreund: It would be better to control this via a prop on DataCard. `.data-card:has(a)` isn’t great: it relies on a link in a data card implying that it is linked up. But what if there are other links in it? Then this selector would match for cards that aren’t linked up like that.

On the matter of browser support, _a month old_ is definitely not old enough to consider the feature well-supported. We shouldn’t use a case-by-case heuristic á la "we can use more modern features for things that aren’t so critical". That just makes the decision making very messy.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
johncowen: > Aren’t these the only type of that kind of element?

- Table rows are clickable blocks
- Tabs are clickable blocks
- Dropdown MenuItems are clickable blocks
- I would guess we might have the odd button that is a a clickable block (KButton has an attribute for doing that at least https://alpha--kongponents.netlify.app/components/button.html#to). This is for making links look like buttons.

And now that you bring it up there are probably a few other blocks that we have that aren't currently linked but I think we should probably link them somewhere, plus there are quite a few blocks in Konnect that have clickable blocks and their titles don't underline on hover so I'm just following that styling as always.

kleinfreund:
> The above is the `data-planes/view` folder, which is only going to grow in size (`*Inbound*TabView.vue` + `*Outbound*TabView.vue`).
> 
> I'm considering breaking this entire component and related summary/tab routes into it's own micro-service/frontend-module, then all code related directly to this viz would be self contained in its own place (not `data-planes`). How does that sound to you?

That’s fine by me. What would you call it? traffic? You’d probably want to keep both inbound and outbound related things in one place so these terms wouldn’t work.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
johncowen: > I guess what I’m saying is it’s not really clear why there is a special rule for links in certain UI fragments to have no underline on hover.

Off the top of my head I think I would say:

"If the link is a textual link (i.e. within text) then it should be blue and underline on hover, otherwise its probably a UI 'button' therefore don't use the blue/underline on hover."

I guess there might be exceptions to that ^ but those will be clear exceptions not the general rule.

johncowen:
> That’s fine by me.

Cool

>  What would you call it? traffic? You’d probably want to keep both inbound and outbound related things in one place so these terms wouldn’t work.

Yeah agree, not sure if traffic is a great name though, can you think of a good word that covers both inbound and outbound? Really they are services, but we are specifically saying these services are connected to things and its the connections that we are trying to describe.. umm `connections`? `streams`, or maybe just stick with `traffic`... 🤔 

I think I'm kinda liking `connections/` wdyt?

kleinfreund:
I like traffic a little better than connections. Connections isn’t too bad, but it’s somewhat ambiguous (with Zones/Dataplanes and their subscriptions). And I think traffic is a general label/bucket is quite accurate for what this is about.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
kleinfreund: I still think these rules are incredibly arbitrary and specifically it’s not clear why links that happen to be part of clickable boxes shouldn’t be underlined on hover. But of course I’m also of the opinion that links should generally be underlined (i.e. not just on hover) so this just follows from that.

Well, if you think that’s correct, then let’s go with that.

johncowen: Resolving this one.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
johncowen: Ok no prob, I moved this higher up (its only the passthrough card that doesn't link/need a pointer so I added the pointer based on that)

---

> But what if there are other links in it?

There never are, it was only to exclude the passthrough in `ServiceTrafficCard` which is only used here.

> On the matter of browser support, a month old is definitely not old enough to consider the feature well-supported.

Its a month old for Firefox (which is far from the only browser you take into account when discussing 'browser support'), which at a rough count/guess is probably less than 10% of users I'd say 90% of those Firefox users are on the the last latest version (120) so are likely to be on the latest (121) in the next few weeks.

For that very small amount of users, they will just not see a pointer cursor in this currently feature flagged new feature unless they rollover the title, just for the next few weeks.

Granted there will always be the odd person that will have an older browser for the next few years, but thats what graceful degradation is all about, the functionality isn't completely broken, its just not as nice as if you had an up-to-date browser.

Again, more than happy to discuss and decide upon a browser support matrix for all browsers, in fact I think we should do that, but it should include graceful degradation exceptions.

Anyway, I'm resolving this seeing as I moved the rule, please reopen if you want anything more here.

johncowen:
(just to be clear the new module creation/move will happen in a different PR)

johncowen:
@kleinfreund I think all the conversation here have been resolved but I'm still looking for an approval, is there anything left to be done here?

kleinfreund (approved):

PR #1998, opened 2024-01-10
ci: add restrictive permissions to some workflows

kleinfreund:
Adds very restrictive workflow-level permissions to create-gui-pr.yml and dispatch-merged-pr-notification.yml. There’s a good chance other permissions are needed (like `id-token: write`), but it’s hard to determine this exactly. I plan to achieve the most restrictive set of permissions possible so which is why I’m starting with pretty much no permissions at all. If more permissions turn out to be necessary, I’ll add them in a follow-up PR.

Docs: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

### DIFF
--- a/.github/workflows/create-gui-pr.yml
+++ b/.github/workflows/create-gui-pr.yml
@@ -21,6 +21,9 @@ on:
       # See “Filter pattern cheat sheet” https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
     branches: [master, 'release-[0-9]+.[0-9]+']
 
+permissions:
+  contents: read # needed for actions/checkout
+
 # Ensures that we only run one workflow per branch at a time.
 # Already running workflows will be cancelled.
 concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/dispatch-merged-pr-notification.yml
```diff
@@ -5,14 +5,15 @@ on:
     types: [closed]
     branches: [master, 'release-[0-9]+.[0-9]+']
 
+permissions: {}
```
kleinfreund, line 8: **Comment**: This is the syntax for setting all permissions to `none`.

.github/workflows/create-gui-pr.yml
```diff
@@ -38,57 +41,48 @@ jobs:
...
 
-      - name: Build dist
-        run: yarn run build
+      - run: yarn run build
 
-      - name: Generate GitHub app token
-        # https://github.com/tibdex/github-app-token
-        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a
+      # https://github.com/tibdex/github-app-token
+      - uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a
```
kleinfreund, line 65: **Comment**: This _might_ need `id-token: write`, but I’m not sure about it. Seeing as the action’s documentation doesn’t mention it, I left it out for now. If the action works without it, even better. If it doesn’t, we can add it.

johncowen (approved):
👍 glad we are doing this

kleinfreund:
> 👍 glad we are doing this

Just FYI, I’m gonna merge this now. If the workflows break because of it, I’ll file follow up PRs as soon as I can. But there might be a window of limbo where these workflows are broken.

kleinfreund:
Results:

- ✅ dispatch-merged-pr-notification.yml (https://github.com/kumahq/kuma-gui/actions/runs/7489350679)
- ✅ create-gui-pr.yml after running tests via `workflow_run` (https://github.com/kumahq/kuma-gui/actions/runs/7489414396)

So, it turns out `it-token: write` is **not** necessary.

PR #1999, opened 2024-01-11
feat: add GUI disabled mode to index template

### LINKED ISSUE #467: Move disable page to gui
### Description

Currently there's this static page inline in kuma: https://github.com/kumahq/kuma/blob/master/pkg/api-server/gui_handler.go#L10-L29

Would be nice to have it in here and have it styled as well

kleinfreund:
Resolves #467.

**Note**: To test this, open `public/no-gui.html` in a browser.

image


kleinfreund:
Extremely confusingly, I’m getting

```
Error: src/services/env/Env.ts(125,5): error TS2353: Object literal may only specify known properties, and 'disabled' does not exist in type 'PathConfig'.
Error: vite.config.preview.ts(68,21): error TS2353: Object literal may only specify known properties, and 'disabled' does not exist in type 'PathConfig'.
```

in CI but not locally. The error also doesn’t seem appropriate because the type does have that property.

johncowen:
> in CI but not locally.

Sounds like the same thing I had in that other PR 🤔 

---

On the PR as a whole, was there an option do just serve up a completely different `index.html` file from the binary if the GUI is disabled? That way we could keep it all completely separate from the GUI (If you think about it the GUI disabled page is a completely different application)

kleinfreund:
> On the PR as a whole, was there an option do just serve up a completely different `index.html` file from the binary if the GUI is disabled? That way we could keep it all completely separate from the GUI (If you think about it the GUI disabled page is a completely different application)

In principle, that’s definitely an option. The backend, of course, knows when the GUI is disabled and could serve, for example, a `no-gui.html` file instead. We could easily add this file. The benefit is that everything added in this PR in terms of logic goes away.

As long as the backend always serves that file for any URL opened by a user (consider an already opened tab on a sub page that’s being refreshed), there shouldn’t be any downsides.

@lahabana do you think this is feasible?

kleinfreund:
> Sounds like the same thing I had in that other PR 🤔

Just to note: this happened after I merged another PR that changed the Env.ts file and so there was a conflict in that file. Seeing as it didn’t happen locally, I think it’s too much of a coincidence that an error related to that file happens while that file has a conflict. But that of course raises the question _how_ that can lead to errors in `yarn lint:ts`. For it to find this error, it would have had to at least see the changes in the vite.preview.ts file. So that means in the workflow run, this PR’s files were checked out (we know this of course because tests run correctly on PRs). At this point I have no idea how this could fail.

kleinfreund:
@johncowen I put this file in public/ directly for now. Do you think it should be in the root and go through a Vite step? It doesn’t seem to be necessary, I think.

johncowen:
Yeah I think keep this as simple as possible and away from our main application as possible (which as little dependencies on our application as possible), so I'd agree, definitely copy/add its own default styles etc, and if possible just inline an SVG for the logo so its literally just an html file. The other benefit there is there not a lot to go wrong if its just some entirely self-contained, dependency-less HTML/CSS/SVG.

johncowen (commented):
I think this looks good to me, only thing I had was we should undo the JS changes here. I would guess the disabled var might be removed from the backend PR, and even if it isn't it will always be true right?

kleinfreund:
> I think this looks good to me, only thing I had was we should undo the JS changes here. I would guess the disabled var might be removed from the backend PR, and even if it isn't it will always be true right?

Not sure that’s right. It’s currently there and has the expected false/true values (see https://github.com/kumahq/kuma/pull/8813/files#diff-61deaedfd1df568bafeeb1487c07721e607084382d112af0a02db439676f39f5) so for now I’ll be keeping this here, too.

kleinfreund:
The field is gonna be removed again, so I removed the changes here now.

johncowen:
Ok thanks, how come there are still image changes and JS changes here? Can we make this one just the GUI disabled mode index template?

This PR should be one single dependency free file, which will therefore be completely isolated and kinda 'frozen in time' and therefore very unlikely to break.

kleinfreund:
I’m keeping the changes I made to the TS files in this PR. I don’t understand why type definitions like that are repeated over files like that and took the opportunity working on these files to correct that oversight.

johncowen:
As always my point is this should just go in on its own as a file on its own, if I just see one single HTML file in this PR I know that we are adding one single dependency free file that we will never change and therefore never break. So I can just immediately hit approve.

As it stands now, I'm still searching in the project to try and figure out what the image changes are for and trying to check to make sure we haven't deleted something we are using here. The same goes for the type changes but thats not as important.

johncowen:
> You should use your best judgment when facing such decisions. A good approach for this is to put yourself in the shoes of the person who will review your patch: will they understand your changes and reasoning just by reading your commit history? Will they find unrelated changes in a particular commit? They shouldn't!

https://github.com/kumahq/kuma/blob/master/CONTRIBUTING.md#commit-atomicity

I get that sometimes its easier to just do things as you go, and thats ok. But if someone asks and you don't have a reason to keep it in the PR, then we should default to just splitting it off into another one. If the reviewer doesn't ask, then fine.

I think we should consider adding something similar to the above link to our repository also.

kleinfreund:
Could you at least reference what image file you’re concerned about?

- I added the image file I’m using in the no-gui.html so we actually have a source for it and can use it elsewhere
- I removed another file that wasn’t used and didn’t belong in this project in the first place (there are actually many more unused image files in the assets directory)

johncowen:
> As always my point is this should just go in on its own as a file on its own, if I just see one single HTML file in this PR I know that we are adding one single dependency free file that we will never change and therefore never break. So I can just immediately hit approve.

Could you do whatever else it is you are doing in another PR please.

> But if someone asks and you don't have a reason to keep it in the PR, then we should default to just splitting it off into another one. If the reviewer doesn't ask, then fine.

---

> I added the image file I’m using in the no-gui.html so we actually have a source for it and can use it elsewhere

Fine do it in a followup, I actually want to use an SVG elsewhere also but it would have been nice to just see the HTML file and go 👍 ✅ 

> I removed another file that wasn’t used and didn’t belong in this project in the first place (there are actually many more unused image files in the assets directory)

Again, fine, just do it in another PR, I see you've put up another PR removing all of them. Which kinda shows you what I mean, instead of just removing this random one, if you've noticed loads aren't used, do them all together in a different PR about removing unused files.

johncowen (approved):

PR #2011, opened 2024-01-15
chore: remove logger

### LINKED ISSUE #1750: remove Logger
### Description

We don't really use it at all.

There are a handful of places where we make a call to Logger, but these can all be removed and the Logger class (and related classes) can also be removed.

kleinfreund:
Resolves #1750.


johncowen (approved):
Oh weird I thought we still had noop calls to logger in the app in places, I guess they've just been removed over time

---

One last thing left in `@/utilities` 🎉

PR #2012, opened 2024-01-15
feat(policies): hide legacy policy types if there are no legacy policies

kleinfreund:
Hides the legacy policy types in the policy type list view if there no legacy policies at all. Ignores MeshGateway policies which we always show.

### DIFF
--- a/features/mesh/policies/Index.feature
+++ b/features/mesh/policies/Index.feature
@@ -108,7 +108,10 @@ Feature: mesh / policies / index
 
     When I visit the "/meshes/default/policies/meshfaultinjections" URL
 
-    Then the "[data-testid^='policy-type-link-']" element exists 14 times
+    Then the "[data-testid='policy-type-link-FaultInjection']" element doesn't exist
+    And the "[data-testid='policy-type-link-MeshFaultInjection']" element exists
+    # Always shows MeshGateway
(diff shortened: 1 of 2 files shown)


features/mesh/policies/Index.feature
```diff
@@ -92,3 +70,81 @@ Feature: mesh / policies / index
...
+            total: 0
+          VirtualOutbound:
+            total: 0
+          MeshFaultInjection:
+            total: 2
+      """
+
+    When I visit the "/meshes/default/policies/meshfaultinjections" URL
+
+    Then the "[data-testid^='policy-type-link-']" element exists 27 times
```
johncowen, line 150: I kinda thought a lot about avoiding the counting and just setting one policy to 0 and looking for that, assuming that if one zero count legacy policy was showing then all the others are also. Then we wouldn't need to know that there are 27 that was have in our mocks. Its a bit of a quibble though so I don't mind really.

johncowen (approved):
LGTM, I left a little comment but kinda nit/opinion based so good to merge for me

features/mesh/policies/Index.feature
kleinfreund: I simplified this test as follows:

- In the second test, only setting one legacy policy type in the mocks
- In both tests, test against the presence (or lack of it) of a legacy policy type in the list

PR #2016, opened 2024-01-15
fix(zones): If there are no zone details pages, don't link a tag to one

johncowen:
Ideally I'd like TagList to work differently, but for now this prevents any `kuma.io/zone` tags from linking to a potentially non-existent page, a bit surprised we've not seen this earlier, which makes me think I might be missing something 🤔 

I guess in reality, zone tags wouldn't exist on old style standalone, but guessing that in single-zone mode things might be tagged with the name of the single zone.

---

**Edit: 5th Feb**

It looks like my guess here turned out to be correct:

>> but guessing that in single-zone mode things might be tagged with the name of the single zone.

It looks like when in "non-federated zone mode" we do get `kuma.io/zone` tags (set to `default`), which means if we try to link to the Zone detail page (using `default` as the zone name) then we will get an error (zone detail routes don't exist in "non-federated zone mode").

**Note**: Base branch is `2.6.x`, so will also require a port to `master`

### DIFF
--- a/src/app/common/TagList.vue
+++ b/src/app/common/TagList.vue
@@ -27,6 +27,7 @@
 <script lang="ts" setup>
 import { computed } from 'vue'
 
+import { useCan } from '@/app/application'
 import { LabelValue } from '@/types/index.d'
 import { getLabels } from '@/utilities/getLabels'
 import type { RouteLocationNamedRaw } from 'vue-router'
@@ -44,6 +45,7 @@ const props = withDefaults(defineProps<{
   shouldTruncate: false,
   alignment: 'left',
 })
+const can = useCan()
 
 const tagList = computed<LabelValueWithRoute[]>(() => {
   const labels = Array.isArray(props.tags) ? props.tags : getLabels(props.tags)
@@ -67,6 +69,9 @@ function getRoute(tag: LabelValue): RouteLocationNamedRaw | undefined {
   try {
     switch (tag.label) {
       case 'kuma.io/zone': {
+        if (!can('use zones')) {
+          return undefined
+        }
         return {
           name: 'zone-cp-detail-view',
           params: {


kleinfreund:
> I guess in reality, zone tags wouldn't exist on old style standalone, but guessing that in single-zone mode things might be tagged with the name of the single zone.

I’m not sure this is the case. Considering that in zone mode, the GUI will be disabled, I don’t think we should special-case this logic here.

Effectively, the GUI currently only needs to handle global mode.

johncowen:
> I’m not sure this is the case. Considering that in zone mode, the GUI will be disabled, I don’t think we should special-case this logic here.

As discussed in https://github.com/kumahq/kuma-gui/issues/1686#issuecomment-1881484611:

> - global mode: nothing changes
> - zone mode + KDS enabled: nothing changes, the GUI will be disabled on the backend side
> - zone mode + KDS disabled: make everything behave like standalone mode behaves currently (essentially, replace mentions of `standalone` with `zone`, i.e. deprecate standalone mode)

> **zone mode** + KDS disabled: make **everything behave like standalone** mode behaves currently

---

On your other point:

> Effectively, the GUI currently only needs to handle global mode.

If this is the case why do we have several areas in the GUI where we check to see if you can `use zones`? Including in zones routing config, where we only add certain routes if `KUMA_MODE="global"` (not `KUMA_MODE="zone"`)

---

I think to get past this and get this resolved, you should probably ask yourself if the change here does any harm.

> but guessing that in single-zone mode things might be tagged with the name of the single zone.

If this ^ is the case, this PR does no harm, but fixes an issue as thing stand currently. If its not the case, this PR does no harm.

In the meantime, as you've seen offline, I'm trying to confirm that "zone mode" or `KUMA_MODE=zone` is infact our old standalone mode which is now "single zone mode" as its the same as before just instead of there not being any zones, everything belongs to the one and only single zone.

johncowen:
> but guessing that in single-zone mode things might be tagged with the name of the single zone.

I asked in team sync, from what I understood in "non-federated zone" mode (aka/up-until-now "single-zone" mode) nothing has a `kuma.io/zone` on it. With that being confirmed, I'm just gonna take another look as I would guess the mocks are wrong, as from what I remember I was getting a error being thrown in `console`, following this confirmation we should never end up with a `kuma.io/zone` on anything in "non-federated zone" mode.

johncowen:
Putting this in draft for the moment, I'm sure now that this is a mocks issue (I tried this again briefly when I made the last comment and I could still see the error happening in when using the GUI with mocks in non-federated zone mode)

Will come back here later to look into it further.

johncowen:
I figured whatever the fix was here it will be nothing like the code in this PR so figured I could close this PR and make an issue instead to keep track of it rather than a draft PR.

See https://github.com/kumahq/kuma-gui/issues/2052

kleinfreund (approved):

PR #2030, opened 2024-01-16
chore(deps-dev): remove names from lockfile

kleinfreund:


johncowen (approved):
really could do with getting to the bottom of this, I guess dependabot isn't using the same yarn version as us? Or its getting called with extra flags or something?

kleinfreund:
I’m gonna play around with https://github.com/dependabot/cli to see if I can determine what Dependabot uses to produce the changed lockfile. But for now, I don’t really have a theory what’s happening. Neither `yarn install` nor `yarn upgrade $package` produce the "wrong" lockfile as we’re seeing it.

PR #2042, opened 2024-01-17
chore: add primary-actions to code-block and use it across the app

johncowen:
In the last feature we added, I tried to just go through and use what we had to make the feature.

Now we are coming back to cleanup/refactor and refine any bits which I held off on whilst building out the feature.

This PR adds a `primary-actions` slot to `CodeBlock` (I noticed it already has a `secondary-actions` slot, but no `primary-`)., and then uses that to add a `[Refresh]` button wherever we have that.

Lastly, the `CodeBlock.vue` file needs moving into the new `code-block/` folder I added here, but I figured it might be easier to review the changes first, then move once we get an a-ok here (I'll probably also put `EnvoyData` and `ResourceCodeBlock` in the same folder)


src/app/common/CodeBlock.vue
```diff
@@ -74,22 +87,21 @@ async function handleCodeBlockRenderEvent({ preElement, codeElement, language, c
 }
 </script>
 
-<style lang="scss">
-.code-block code {
-  // DO NOT REMOVE OR OVERRIDE THIS.
-  // Believe it or not, this avoids a critical performance issue with highlighting code and specifically highlighting lines using Prism’s line-highlight plugin.
-  // https://github.com/PrismJS/prism/issues/2062
-  display: block !important;
```
kleinfreund, line 82: **Question**: Have you tested and verified that this is not an issue anymore?

johncowen: I played with it in Chrome and whilst I've never seen the issue before, I didn't notice any issues when removing this.

I also noticed from reading linked GH issue in the comment (which is from 2019, but runs to 2022 / Chrome 77 to 88, so roughly between 5 and 2 years old) that it seems to have been fixed, possibly regressed and then fixed, its kinda hard to tell. In any case its clear it was a browser bug, and as I'm not noticing anything terrible rn, it seems fine to me.

Could you have a look yourself and see if you notice anything? I'd be fine adding this back in if so. lmk

btw, have you ever seen this bug/performance issue yourself? If so, could you explain how to reproduce?

johncowen:
> Generally wondering: what’s the purpose of this PR? I feel like adding the Refresh button like we’re currently doing is good enough and more flexible (you can put the refresh button anywhere in the UI). Why do we need this to be part of the CodeBlock component?

Not totally sure if I understand you question, but:

> Why do we need this to be part of the CodeBlock component?

The refresh button is not part of the CodeBlock component. You still add that from The Outside via a 'primary-actions' slot which then appears above the CodeBlock like a toolbar. We already had a configurable `secondary-actions` slot, but the current layout requires a separate area above the CodeBlock. I also didn't want to try adding a refresh button into `secondary-actions` as it covers up the text that I want to see.

> Generally wondering: what’s the purpose of this PR?

Mainly, you now no longer need to repeat the styles for the 'toolbar' layout for whatever tooling buttons you need to add to the code block. Ideally I would have like to add this button to the actual KCodeBlock 'toolbar', but it doesn't look like you can configure those from The Outside.

---

Reading your comment again I do think I've understood you correctly, and I think you might have misunderstood what is happening here:

> I feel like adding the Refresh button like we’re currently doing is good enough and more flexible (you can put the refresh button anywhere in the UI)

There is no Refresh button in the CodeBlock component, only a slot to put arbitrary buttons. As these buttons are added from the The Outside its easier to add actions/events to them without a tonne of extra plumbing (as would be the case if I'd hardcoded buttons inside the CodeBlock itself)

Does that help to answer your questions?

src/app/common/CodeBlock.vue
kleinfreund: > btw, have you ever seen this bug/performance issue yourself? If so, could you explain how to reproduce?

Yes. When I built the code block component, I noticed this effect. The performance implication then was massive (echoing the descriptions in the issue).

I’m gonna do my own test (also making sure nothing else sets `display: block` on the `code` element).

To reproduce, all you would do then was run any query (or a series of queries) in the code block. The syntax highlighting became so slow that the search itself appeared to be extremely slow (which wasn’t the case). You’d definitely notice it if it still happens.

johncowen: @kleinfreund did you get a chance to look at the CodeBlock in Chrome? I've tried again since and not noticed anything.

Apart from that was there anything else blocking on this PR?

FYI I'm waiting on this PR before I do the module move and then i18n everything.

johncowen: P.S. FYI, I also just realized I kinda need this for the other work I'm doing now also.

I'm looking to reuse the `.../traffic` DataSource on all our stats pages, which means I can't use `EnvoyData` (the actual stats are no longer just `data` but `data.raw`). This was actually one of the reasons I did this PR - so that I could use CodeBlock everywhere and feed it things shaped in all different ways.

kleinfreund: I couldn’t reproduce this issue anymore, so I’m gonna resolve this.

kleinfreund:
My question is rather why this is related to the CodeBlock via the introduction of primary actions at all: We use refresh buttons in other places, too, and they don’t always refresh just the contents of a code block like they do for the envoy data triplet. It looks like what’s being generalized here is the concept of an actions toolbar and more specifically the repetition of its styles is avoided. That could be done via a CSS rule set in `src/assets/styles/_components.scss` which we could use in other places, too.

Where the refresh button is located currently on the envoy data pages -- i.e. above the code block -- is rather coincidental.

johncowen:
> My question is rather why this is related to the CodeBlock via the introduction of primary actions at all

1. CodeBlock has a secondary actions slot, but no primary actions slot. At the very least thats a little weird. This gives it a primary slot above the editor, where we commonly add more actions (a refresh action that refreshes the content of the codeblock)
2. I would add it to secondary actions but it puts the button in the wrong place. Ideally I would be able to add buttons to the KCodeBlock toolbar but I can't.

Is there a difference is here to when you added the secondary-actions block? I'm doing exactly the same thing here, but I need the positioning to be different so I don't think I can use the secondary actions block without some strange CSS.

A larger point is: **We currently have 4 different components just to show code**, this feels very weird.

Ideally we should have one and only one that is sufficiently configurable. This is the start of a step towards reducing those 4 components (which is essentially the same component) down to 1 or 2 components.

johncowen:
@kleinfreund As explained in https://github.com/kumahq/kuma-gui/pull/2042#discussion_r1458838858 I'm waiting on this to move on with this work, I would appreciate it if we can find a way to resolve this a little quicker. I'd be happy to catchup offline in order to get this resolved in one sitting.

kleinfreund (approved):

johncowen:
Quick note: I just added a commit on the end here to do:

> Lastly, the CodeBlock.vue file needs moving into the new code-block/ folder I added here, but I figured it might be easier to review the changes first, then move once we get an a-ok here (I'll probably also put EnvoyData and ResourceCodeBlock in the same folder)

PR #2047, opened 2024-01-18
fix(NavTabs): tab click target being too small

kleinfreund:

### DIFF
--- a/src/app/common/NavTabs.vue
+++ b/src/app/common/NavTabs.vue
@@ -12,7 +12,6 @@
       #[`${tab.routeName}-anchor`]
     >
       <RouterLink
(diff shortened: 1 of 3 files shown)


src/app/common/NavTabs.vue
```diff
@@ -2,7 +2,7 @@
   <KTabs
     :tabs="kTabs"
     :model-value="currentTabHash"
-    :show-panels="false"
+    hide-panels
```
kleinfreund, line 5: **Comment**: Noticed we were using the wrong prop (which wasn’t even called `showPanels`). Not sure if that changed earlier already.

johncowen (approved):

PR #2071, opened 2024-01-22
feature: add DataCollection component and use it for Dataplane viz

johncowen:
Feature-wise this PR:

- Adds EmptyStates/EmptyBlocks to the DPP viz (which we also use for the "in-viz ErrorState")
- Adds a Loading spinner into the Outbounds column
- Reduces the amount of re-renders of the DPP viz when clicking ono the Cards

To do this I made a new component called `DataCollection` which:

- Provides a component for filtering/finding, sorting (and pagination outside of a AppCollection table should we need that) and also a default EmptyState for empty collections (which is slottable)
- Replaces some of our usage of `<template v-for="">`

I've moved the finding of the exact 'item' we need to show in the drawer to the drawer itself.

I've done this because we need to use `route.params.service` in order to find the 'item'. Moving the find (and the dependency on `route.params.service`) out of the main view and into the drawer means that the main view doesn't re-render whenever `route.params.service` changes, which happens whenever you click on a card.

### Loading State

Screenshot 2024-01-23 at 10 19 58

### `/stats` Error State

Screenshot 2024-01-23 at 10 20 05

### DIFF
--- a/src/app/application/components/data-collection/DataCollection.vue
+++ b/src/app/application/components/data-collection/DataCollection.vue
@@ -1,0 +1,84 @@
+<template>
+  <slot
+    v-if="items.length === 0"
+    name="empty"
+    :items="items"
+  >
+    <EmptyBlock />
+  </slot>
+  <slot
+    v-else
+    name="default"
+    :items="paginated"
+  />
+
+  <slot
+    v-if="props.page !== 0 && items.length > 0"
+    name="pagination"
+    :items="paginated"
+  >
+    <KPagination
+      :total-count="items.length"
+      :current-page="props.page"
+      :initial-page-size="props.pageSize"
+      :page-sizes="[15, 30, 50, 75, 100]"
+      @page-changed="({ page }: PaginationChangeEvent) => {
+        emit('change', {
+          page,
+          pageSize: props.pageSize,
+        })
+      }"
+    />
+  </slot>
+</template>
+<script lang="ts" generic="T extends {} = {}" setup>
+import { computed } from 'vue'
+
+import EmptyBlock from '@/app/common/EmptyBlock.vue'
+type PaginationChangeEvent = {
+  page: number
+}
+const props = withDefaults(defineProps<{
+  // type: string
+  paginationType?: 'server' | 'client'
+  page?: number
+  pageSize?: number
+  items: T[]
+  predicate?: (item: T) => boolean
+  comparator?: ((item: T) => number) | undefined
+  find?: boolean
+}>(), {
+  // type: '',
+  paginationType: 'server',
+  page: 0,
+  pageSize: 50,
+  predicate: () => true,
+  comparator: undefined,
+  find: false,
+})
+const emit = defineEmits<{
+  (e: 'change', value: {page: number, pageSize: number}): void
+  (e: 'error', error: Error): void
+}>()
+const items = computed(() => {
+  if (props.find) {
+    const found = props.items.find(props.predicate)
+    return typeof found === 'undefined' ? [] : [found]
+  } else {
+    const found = props.items.filter(props.predicate)
+    if (typeof props.comparator !== 'undefined') {
+      return found.sort(props.comparator)
+    }
+    return found
+  }
+})
+const paginated = computed(() => {
+  if (props.paginationType === 'client') {
+    const cursor = props.pageSize * (props.page - 1)
+    return items.value.slice(cursor, cursor + props.pageSize)
+  } else {
+    return items.value
+  }
+})
+
+</script>
--- a/src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
+++ b/src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
@@ -21,90 +21,97 @@
     <template
       v-if="props.traffic"
     >
-      <!-- rx and tx are purposefully reversed in all the below to rx=tx and tx=rx here due to the direction of the traffic (downstream) -->
-      <dl>
-        <template
-          v-if="props.protocol === 'passthrough'"
-        >
+      <template v-if="props.traffic.name.length > 0">
+        <!-- rx and tx are purposefully reversed in all the below to rx=tx and tx=rx here due to the direction of the traffic (downstream) -->
+        <dl>
           <template
-            v-for="(item, key) in [
-              (['http', 'tcp'] as const).reduce((prev, protocol) => {
-                const direction = 'downstream'
-                // sum both the properties we need from both protocols
-                return Object.entries(props.traffic?.[protocol] || {}).reduce((prev, [key, value]) => {
-                  return [`${direction}_cx_tx_bytes_total`, `${direction}_cx_rx_bytes_total`].includes(key) ?
-                    { ...prev, [key]: (value as number) + (prev[key] ?? 0) } :
-                    prev
-                }, prev)
-              }, {} as Record<string, number>),
-            ]"
-            :key="key"
+            v-if="props.protocol === 'passthrough'"
+          >
+            <template
+              v-for="(item, key) in [
+                (['http', 'tcp'] as const).reduce((prev, protocol) => {
+                  const direction = 'downstream'
+                  // sum both the properties we need from both protocols
+                  return Object.entries(props.traffic?.[protocol] || {}).reduce((prev, [key, value]) => {
+                    return [`${direction}_cx_tx_bytes_total`, `${direction}_cx_rx_bytes_total`].includes(key) ?
+                      { ...prev, [key]: (value as number) + (prev[key] ?? 0) } :
+                      prev
+                  }, prev)
+                }, {} as Record<string, number>),
+              ]"
+              :key="key"
+            >
+              <div>
+                <dt>{{ t('data-planes.components.service_traffic_card.tx') }}</dt>
+                <dd>{{ t('common.formats.bytes', { value: item.downstream_cx_rx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
+              </div>
+              <div>
+                <dt>{{ t('data-planes.components.service_traffic_card.rx') }}</dt>
+                <dd>{{ t('common.formats.bytes', { value: item.downstream_cx_tx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
+              </div>
+            </template>
+          </template>
+          <template
+            v-else-if="props.protocol === 'grpc'"
+          >
+            <div>
+              <dt>{{ t('data-planes.components.service_traffic_card.grpc_success') }}</dt>
+              <dd>{{ t('common.formats.integer', { value: props.traffic.grpc?.success as (number | undefined) ?? 0 }) }}</dd>
+            </div>
+            <div>
+              <dt>{{ t('data-planes.components.service_traffic_card.grpc_failure') }}</dt>
+              <dd>{{ t('common.formats.integer', { value: props.traffic.grpc?.failure as (number | undefined) ?? 0 }) }}</dd>
+            </div>
+          </template>
+          <template
+            v-else-if="props.protocol === 'http'"
+          >
+            <div
+              v-for="value in [props.traffic.http?.downstream_rq_1xx as (number | undefined) ?? 0].filter(item => item !== 0)"
+              :key="value"
+            >
+              <dt>{{ t('data-planes.components.service_traffic_card.1xx') }}</dt>
+              <dd>{{ t('common.formats.integer', { value: value }) }}</dd>
+            </div>
+            <div>
+              <dt>{{ t('data-planes.components.service_traffic_card.2xx') }}</dt>
+              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_2xx as (number | undefined) ?? 0 }) }}</dd>
+            </div>
+            <div
+              v-for="value in [props.traffic.http?.downstream_rq_3xx as (number | undefined) ?? 0].filter(item => item !== 0)"
+              :key="value"
+            >
+              <dt>{{ t('data-planes.components.service_traffic_card.3xx') }}</dt>
+              <dd>{{ t('common.formats.integer', { value: value }) }}</dd>
+            </div>
+            <div>
+              <dt>{{ t('data-planes.components.service_traffic_card.4xx') }}</dt>
+              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_4xx as (number | undefined) ?? 0 }) }}</dd>
+            </div>
+            <div>
+              <dt>{{ t('data-planes.components.service_traffic_card.5xx') }}</dt>
+              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_5xx as (number | undefined) ?? 0 }) }}</dd>
+            </div>
+          </template>
+          <template
+            v-else
           >
             <div>
               <dt>{{ t('data-planes.components.service_traffic_card.tx') }}</dt>
-              <dd>{{ t('common.formats.bytes', { value: item.downstream_cx_rx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
+              <dd>{{ t('common.formats.bytes', { value: props.traffic.tcp?.downstream_cx_rx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
             </div>
             <div>
               <dt>{{ t('data-planes.components.service_traffic_card.rx') }}</dt>
-              <dd>{{ t('common.formats.bytes', { value: item.downstream_cx_tx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
+              <dd>{{ t('common.formats.bytes', { value: props.traffic.tcp?.downstream_cx_tx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
             </div>
           </template>
-        </template>
-        <template
-          v-else-if="props.protocol === 'grpc'"
-        >
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.grpc_success') }}</dt>
-            <dd>{{ t('common.formats.integer', { value: props.traffic.grpc?.success as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.grpc_failure') }}</dt>
-            <dd>{{ t('common.formats.integer', { value: props.traffic.grpc?.failure as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-        </template>
-        <template
-          v-else-if="props.protocol === 'http'"
-        >
-          <div
-            v-for="value in [props.traffic.http?.downstream_rq_1xx as (number | undefined) ?? 0].filter(item => item !== 0)"
-            :key="value"
-          >
-            <dt>{{ t('data-planes.components.service_traffic_card.1xx') }}</dt>
-            <dd>{{ t('common.formats.integer', { value: value }) }}</dd>
-          </div>
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.2xx') }}</dt>
-            <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_2xx as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-          <div
-            v-for="value in [props.traffic.http?.downstream_rq_3xx as (number | undefined) ?? 0].filter(item => item !== 0)"
-            :key="value"
-          >
-            <dt>{{ t('data-planes.components.service_traffic_card.3xx') }}</dt>
-            <dd>{{ t('common.formats.integer', { value: value }) }}</dd>
-          </div>
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.4xx') }}</dt>
-            <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_4xx as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.5xx') }}</dt>
-            <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_5xx as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-        </template>
-        <template
-          v-else
-        >
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.tx') }}</dt>
-            <dd>{{ t('common.formats.bytes', { value: props.traffic.tcp?.downstream_cx_rx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.rx') }}</dt>
-            <dd>{{ t('common.formats.bytes', { value: props.traffic.tcp?.downstream_cx_tx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-        </template>
-      </dl>
+        </dl>
+      </template>
+    </template>
+    <template v-else>
+      <KSkeletonBox
+        :width="10"
+      />
     </template>
   </DataCard>
 </template>
--- a/src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
+++ b/src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
@@ -110,7 +110,7 @@
     </template>
     <template v-else>
       <KSkeletonBox
-        :width="10"
+        width="10"
       />
     </template>
   </DataCard>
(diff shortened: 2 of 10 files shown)


src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -202,67 +213,71 @@
...
           </KCard>
           <RouterView
-            v-if="route.params.service && [
-              traffic?.outbounds,
-              props.data.dataplane.networking.inbounds,
-            ].every(item => typeof item !== 'undefined')"
             v-slot="child"
           >
             <SummaryView
+              v-if="child.route.name !== route.name"
```
kleinfreund, line 280: **Question**: This doesn’t look right. Why are we no longer guarding this by the specific route parameter (i.e. `route.params.service`) we know we need to guard against?

johncowen: If child route is ever not the name of this route i.e. `"data-plane-detail-view"` then it means we must be in a "sub view" i.e. in this case a summary view seeing as "data-plane-detail-view"` has no other types of sub view, therefore show the SummaryView. That makes sense I think?

src/app/application/components/data-collection/DataCollection.vue
```diff
@@ -0,0 +1,84 @@
...
+      :initial-page-size="props.pageSize"
+      :page-sizes="[15, 30, 50, 75, 100]"
+      @page-changed="({ page }: PaginationChangeEvent) => {
+        emit('change', {
+          page,
+          pageSize: props.pageSize,
+        })
+      }"
+    />
+  </slot>
```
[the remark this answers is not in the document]
johncowen: Nowhere

>  (and pagination outside of a AppCollection table should we need that)

I want this component to be built in a way that makes sure it thinks about pagination for if we need to paginate this view (outbound lists). Therefore I built it with that in mind. Now if we need to paginate the outbound listing if it has 100's of inbounds, its pretty much done.

src/app/data-planes/views/DataPlaneDetailView.vue
kleinfreund: That is coincidentally true as long as don’t have other child routes in this view that aren’t summary trays. The check for `route.params.service` is more robust. A check for the intended sub routes would be even more robust.

src/app/application/components/data-collection/DataCollection.vue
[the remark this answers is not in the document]
kleinfreund: Shouldn’t we only add this when we actually need it? As far as I know, the designs don’t include pagination for this either. This seems like premature optimization.

src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: There're no other child routes in this route that aren't a SummaryViews are there? Whilst I wouldn't completely rule it out I'm 90% sure we'd probably never have a subroute here that isn't a SummaryView and if that ever did happen we would have to change quite a bit more I think.

src/app/application/components/data-collection/DataCollection.vue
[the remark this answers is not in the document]
johncowen: Premature optimization would be building something to be more optimized when it thats not necessary.

This isn't about optimization its about making sure that DataCollection is built in a way that we can use it for paging if we ever need to. I don't want to have to rebuild it all completely differently if someone says "can we add pagination to this non-table thing please?"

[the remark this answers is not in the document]
johncowen: Or even: I don't want to have to rebuild it all completely differently if someone says "can we add frontend pagination to this table thing please?"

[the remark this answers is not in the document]
kleinfreund: Aside from the pedantry about whether I’m using the right term here, we don’t currently have concrete plans to use this bit of code. It’s doubtful having to "rebuild it all completely differently" would become necessary in the event that we end up adding pagination here.

Looking past pagination, this component offers conditionally rendering an EmptyBlock and no functionality beyond that.

So I’m kind of confused: Why do we need this right now? It doesn’t seem the right time to introduce an abstraction over the search/filtering capabilities that are yet to be added in this view.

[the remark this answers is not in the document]
johncowen: > Aside from the pedantry about whether I’m using the right term here

There's a big difference between building something so its flexible for the long term (not knowing what features might be required in the future) and micro and/or premature optimization, as long as you know making sure this worked alongside KPagination isn't about optimizing performance.

Happy to remove/comment out the KPagination bit now the component has been around it. lmk

> Why do we need this right now?

Well one more thing is it means I need to do less of `<template v-for>` which I know you don't particularly like.

src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: Feels like I can resolve this one?

src/app/application/components/data-collection/DataCollection.vue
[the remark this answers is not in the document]
johncowen: > It doesn’t seem the right time to introduce an abstraction over the search/filtering capabilities that are yet to be added in this view.

We filter by `inactive` in this view

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
```diff
@@ -21,90 +21,97 @@
...
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.tx') }}</dt>
-            <dd>{{ t('common.formats.bytes', { value: props.traffic.tcp?.downstream_cx_rx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-          <div>
-            <dt>{{ t('data-planes.components.service_traffic_card.rx') }}</dt>
-            <dd>{{ t('common.formats.bytes', { value: props.traffic.tcp?.downstream_cx_tx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
-          </div>
-        </template>
-      </dl>
```
johncowen, line 107: Just to note the majority of this red/green change was down to extra nesting, so there's far less change than it looks, the KSkeletonBox was also added here for a tiny loading state for the stats in the cards.

src/app/application/components/data-collection/DataCollection.vue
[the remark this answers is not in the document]
kleinfreund: This is definitely much less about the introduction of a new component than about the addition of features that we’re not using. Since this PR contains work that was deemed relevant for the upcoming release of Kuma, it would’ve made more sense to keep it slim and isolated to those feature additions.

Nonetheless, I’m going to resolve this thread now so we can get this in.

src/app/data-planes/views/DataPlaneInboundSummaryView.vue
```diff
@@ -20,27 +20,22 @@
...
+        <DataCollection
+          v-slot="{ items }"
+          :items="props.data"
+          :predicate="(item) => `${item.port}` === route.params.service"
+          :find="true"
+        >
           <component
             :is="child.Component"
-            :data="props.data"
+            :data="items[0]"
```
kleinfreund, line 35: **Question**: What’s this about? Why are we getting a list of items here of which we’re unconditionally picking the first item? As far as I can tell, this is a context in which we never have a list of items to begin with: we’re clicking on an inbound for which we show a summary tray.

So I guess the question really is, how come this is even using DataCollection?

johncowen: A few things here, but I'm glad you've asked as there's some important detail in your questions which you may have missed from the PR desc:

> I've done this because we need to use `route.params.service` in order to find the 'item'. Moving the find (and the dependency on route.params.service) out of the main view and into the drawer means that the main view doesn't re-render whenever route.params.service changes, which happens whenever you click on a card.


First up:

> What’s this about? Why are we getting a list of items here of which we’re unconditionally picking the first item?

`items` in this context is never zero length, it its zero length we show the empty slot instead of the default slot. I guess this is another reason why we need (well, its nice to have) the DataCollection component now.

https://github.com/kumahq/kuma-gui/pull/2071/files#diff-dfcb4cdef54236d2334cfa3be3f50687776d5e558709f1f3034de861ce86d343R3

> As far as I can tell, this is a context in which we never have a list of items to begin with: we’re clicking on an inbound for which we show a summary tray.

There's a change just below `data: DataplaneInbound[]`, ie. we now pass the entire list into the Index/Tab view (the one with the actual tabs in it). So in this context we do have a list of items to begin with. We then find the 'selected' item in this view instead of the `Detail.vue`. This means we don't depend on `route.params.service` in the `Detail.vue` and it therefore re-renders far less. I get that it's a little strange but its just to get around the fact that I don't want to re-render the entire viz every time `route.params.service` changes.

For example, if you checkout `master`, you'll see that the `Detail.vue` re-renders every time you click on _any_ card. Whereas with this approach it only re-renders if you are looking at one type of  'bound and you switch to looking at an the other type of 'bound. If you are just looking at/clicking on various outbounds, then the viz doesn't re-render.

> So I guess the question really is, how come this is even using DataCollection?

In order to find the 'bound that `route.params.service` relates to, avoid the `undefined` result/trap and show an EmptyState/Block if we can't find a 'bound. DataCollection does all this for you (apart from you having to provide the `predicate`)

Does that help?

kleinfreund: I understand in principle what you’re explaining.

> For example, if you checkout `master`, you'll see that the `Detail.vue` re-renders every time you click on _any_ card.

What exactly do you see re-render in DataPlaneDetailView.vue on `master`? As far as I can tell, only the value passed to DataPlaneOutboundSummaryView.vue `props.data` will change in response to `route.params.service` being different, but neither component seems to _fully_ re-render or re-mount.

johncowen: I usually drop `{{ Math.random() }}` in places and see what changes. If it doesn't change its not getting included in the re-render/vdom pass because it has no changing dependencies. As soon as you refer to `route.params.service` in the view (which changes every time you click a card) the Math.random()'s will change.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
```diff
@@ -21,90 +21,97 @@
...
+              <dt>{{ t('data-planes.components.service_traffic_card.rx') }}</dt>
+              <dd>{{ t('common.formats.bytes', { value: props.traffic.tcp?.downstream_cx_tx_bytes_total as (number | undefined) ?? 0 }) }}</dd>
+            </div>
+          </template>
+        </dl>
+      </template>
+    </template>
+    <template v-else>
+      <KSkeletonBox
+        :width="10"
```
kleinfreund, line 113: **Suggestion**: This should be a string

```suggestion
        width="10"
```

src/app/data-planes/views/DataPlaneInboundSummaryView.vue
johncowen: Maybe this helps:

### `master`

dom-changes-master

### This PR

dom-changes-this-pr

kleinfreund: This re-render is triggered by the `.filter(item => (item.protocol === 'tcp' ? item.tcp?.downstream_cx_rx_bytes_total : item.http?.downstream_rq_total) as (number | undefined) ?? 0 > 0)` call which is no longer present in this PR. That is expected for the same reasons `Math.random()` will trigger a re-render provided Vue checks for it (which it must when `route.params.service` changes).

So while `route.params.service` provides cause for Vue to even go looking for changes, it’s the function calls in the template that force the re-render.

johncowen: > So while route.params.service provides cause for Vue to even go looking for changes

This is what I'm avoiding, vdom 'should' avoid any DOM changes else, but I've found it's more "it depends" a lot with Vue, I think the soon to come Vue upgrade will improve things a little.

> This re-render is triggered by the .filter(item => (item.protocol === 'tcp' ? item.tcp?.downstream_cx_rx_bytes_total : item.http?.downstream_rq_total) as (number | undefined) ?? 0 > 0) call which is no longer present in this PR

Not sure I follow, how is it no longer present?

https://github.com/kumahq/kuma-gui/pull/2071/files#diff-2bdf25833f5d0d9db633e05f8f1a0b057bed9b116d43591323b47acfd8a75db8R235

kleinfreund: The structure is different and most importantly, in doesn’t set `:key="outbounds"` on the `template` with the `v-for`. In fact changing the key there to `:key="index"` and adding `index` to the loop also avoids this re-render.

johncowen: Hmmm, I've kinda read you shouldn't use the `index` for the key 🤷 , plus I went back to `master` and tried it and the `{{Math.random}}` check continues to change/update no matter where I place it.

On the other hand, I did put the filtering back in the `*Detail.vue` whilst looking into this:

---

The `RouterView` part will update no matter what you do, so you _could_ use `route.params.service` inside the RouterView seeing as that is checking for a re-render anyway. So kinda "if we are checking for a re-render there we may as well add a dependency" not necessarily "avoid the dependency and therefore the check".

...but a few other issues have come up I think:

---

I did realize that, in doing the filtering in the `*Detail.vue` I then either have to accept "Thing | undefined" instead of just "Thing" in the `SummaryView.vue` and then guard for that `undefined` thing in the `SummaryView.vue`, which was something I wanted to avoid or not show anything at all inside the summary drawer if the find is `undefined`. 

We also need to revert back to checking the type of the 'bound in the `Detail.vue` to decide how to filter, it was kinda nice having the actual `In/OutboundView.vue` to do that in instead.

I think another thing I'd want to avoid is somebody at a later date, thinking "oh look we have `route.params.service` as a dependency I may as well use that up here at the top of the View" without realizing that in using the dependency elsewhere they've now caused Vue to consider the whole thing for a re-render.

---

Is there anything else on this PR or is this the last thing?

kleinfreund: > Hmmm, I've kinda read you shouldn't use the `index` for the key 🤷 

That’s news to me. Do you recall why? In this case, we’re setting an object reference as the key which naturally will trigger re-renders more frequently.

> plus I went back to `master` and tried it and the `{{Math.random}}` check continues to change/update no matter where I place it.

Which is expected. Vue needs to check for changed trees when part of the template's dependencies change and so every function in the template _will_ be called again and its sub tree re-rendered. That’s correct and expected behavior. It’s futile to try and avoid the checks just so that functions in templates aren’t called again. If that becomes an issue, the thing to change are the function calls. Not the source of the Vue checks for re-rendering.

---

> I did realize that, in doing the filtering in the `*Detail.vue` I then either have to accept "Thing | undefined" instead of just "Thing" in the `SummaryView.vue` and then guard for that `undefined` thing in the `SummaryView.vue`, which was something I wanted to avoid or not show anything at all inside the summary drawer if the find is `undefined`.

One reliable way to avoid this is to have a dedicated data source in the summary view. Of course one way or another a "not found" handling will need to exist in the summary view. The approach here doesn’t really prevent that from being necessary, it just dresses it up differently.

> I think another thing I'd want to avoid is somebody at a later date, thinking "oh look we have `route.params.service` as a dependency I may as well use that up here at the top of the View" without realizing that in using the dependency elsewhere they've now caused Vue to consider the whole thing for a re-render.

We should write templates that allow Vue to determine "aha, I don’t actually need to update this" instead of trying to prevent it from making those sort of checks in the first place. Vue checking for this is not the issue here. It’s the templates using function calls which includes all the in-place filtering and finds we do in a couple of places and, evidently, using object references as `key` values for `v-for` template trees.

---

> Is there anything else on this PR or is this the last thing?

Not really, but the topic discussed here is pretty much the main content of this PR and I’m not keen on the "move a list of items into a summary view component so it can find the one item it needs" approach. The summary views should continue to deal with one item, not a list of items (unless of course their purpose is to show lists of things).

There is definitely some unfortunate cross dependencies happening between the detail and summary views that we should refine and work out, but there have got to be better ways than making the summary view search through a list of items that it shouldn’t be concerned with. Double so if that is motivated with trying to avoid Vue internal behavior that by itself isn’t problematic.

johncowen: > That’s news to me. Do you recall why?

https://learnvue.co/articles/vue-for-loop-tips

> **But don't use index as the key!**
> The purpose of specifying a key is for Vue to link each item in our list to its corresponding vnode. Then, when the orders of our keys change - the orders of the elements will as well.
> If we use index as our key, the order of the keys will never change - since index always increases sequentially. This acts the same as not declaring a key since the index will be the same at each element's place.

That one ^ was just the first one that came to mind.

---

> Which is expected. Vue needs to check for changed trees when part of the template's dependencies change and so every function in the template will be called again and its sub tree re-rendered.

It's expected if you add a dependency that you need, i.e if that dependency changes then you would expect things to re-render. As I keep explaining here, **I'm removing the dependency so that happens less.**

---

Even more importantly than the minor dependency removal to avoid Vue re-renders (which I've been doing throughout the application for quite a while now), is the thinking behind it:

`route.params.service` is defined by the SummaryView, not the DetailView.

https://github.com/kumahq/kuma-gui/blob/fcb5c74292a43037d476a756186821e8896fad9b/src/app/data-planes/routes.ts#L16-L18

On this point alone, we should only be accessing `route.params.service` from the SummaryView, not its parent.

---

Another important point.

In every single 'item' view. We use DataSource to filter the entire list of items by the "name of the thing". The only difference is, we use the backend via an API to do that. We ask the DataSource to give us the item from the list that corresponds to "name of the thing" in the item view.

This is exactly the same as what we are doing here. Only we are doing it on the frontend via a DataCollection rather that a DataSource. I could just imagine it being:

```
<DataSource src="/mesh/name-of-data-plane/inbounds/name-of-the-thing" />
```

If DataSource knows that `inbounds` are held on the frontend then it would work exactly the same. (of course that fact that we wouldn't make an additional HTTP API request here would be entirely hidden from the end user/engineer)

---

All in all, whilst the Vue re-rendering thing is something I'd rather avoid, even without that reason, I wouldn't change my approach here.

> Not really, but the topic discussed here is pretty much the main content of this PR and I’m not keen on the "move a list of items into a summary view component so it can find the one item it needs" approach

The main content of the PR is adding improved error/empty and loading states. Things that are important for our GUI users. I made a very simple component to help me to do that and avoid the `item[0] === undefined` 'trap' plus provide an alternative to `<template v-for="">` which you have said on numerous occasions you don't like.

Kinda of rhetorical questions, not looking for answers here:

- How important is your disagreement of where I've put the filter?
- Which do you prefer `<template v-for="">` or `<DataCollection>`?

Let's continue this at https://github.com/kumahq/kuma-gui/pull/2071#issuecomment-1909719863, as I think we should do your option 2. I'm going to resolve this thread, feel free to re-open if you think it shouldn't be resolved.

Lastly:

> Is there anything else on this PR or is this the last thing?

> Not really,

If we are resolving this one, could you let me know if there is anything else, otherwise/if not lets get this approved and merged.

johncowen:
> If we are resolving this one, could you let me know if there is anything else, otherwise/if not lets get this approved and merged.

Please see https://github.com/kumahq/kuma-gui/pull/2071#discussion_r1466244659

lmk

src/app/data-planes/views/DataPlaneInboundSummaryView.vue
kleinfreund: > It's expected if you add a dependency that you need, i.e if that dependency changes then you would expect things to re-render. As I keep explaining here, **I'm removing the dependency so that happens less.**

Haven’t I demonstrated that the reason for why what you showed me above re-renders **isn’t** the presence of `route.params.service`? It’s complete coincidence that the removal of its use here prevents the re-renders of parts of the view here. If you changed any other remaining dependencies of the template, you’d see the very same re-rendering effects.

This will happen with all code that roughly has the following structure:

```vue
<template
  v-for="outbounds in [
    route.params.inactive ? traffic.outbounds : traffic.outbounds.filter(item => (item.protocol === 'tcp' ? item.tcp?.downstream_cx_rx_bytes_total : item.http?.downstream_rq_total) as (number | undefined) ?? 0 > 0),
  ]"
  :key="outbounds"
>
```

Upon seeing _any_ dependency change, Vue is forced to re-evaluate the `Array.prototype.filter` call because it can’t cache results of function calls. That then changes `outbounds` which forces a re-render via the `key` attribute.

You can add a dependency anywhere in the template and change it and observe how completely unrelated parts of the template are forced to re-render because of this.

Changing `key` to use a value that doesn’t unintentionally update like that fixes this without any other work needed. It also doesn’t matter whether that is a `v-for` index in this case as this is just one of these single element array hacks we use.

---

> On this point alone, we should only be accessing `route.params.service` from the SummaryView, not its parent.

While that seems logical (and I agree), you buy that purity by trading in the access of the list of inbounds in the summary view/data collection proxy component. For the same reason why the detail view shouldn’t deal with a route parameter it doesn’t own, the summary view shouldn’t deal with more data than it needs. Since the detail view _already has that data_ and we are, at least for the time being, avoiding to make an additional request in the summary view, we should not move the list of inbounds into it.

---

> All in all, whilst the Vue re-rendering thing is something I'd rather avoid, even without that reason, I wouldn't change my approach here.

I thought the re-rendering was your entire motivation for it?

> This means we don't depend on route.params.service in the Detail.vue and it therefore re-renders far less. I get that it's a little strange but its just to get around the fact that I don't want to re-render the entire viz every time route.params.service changes.

Seeing as you don’t need it to avoid the re-renders, I don’t understand why we still need DataCollection in a context where we don’t want to render a collection of any kind in the first place.

```vue
        <DataCollection
          v-slot="{ items }"
          :items="props.data"
          :predicate="(item) => `${item.port}` === route.params.service"
          :find="true"
        >
          <component
            :is="child.Component"
            :data="items[0]"
          />
        </DataCollection>
```

It’s specifically this bit of code in the summary view that I don’t agree with. For one, the summary shouldn’t do the filtering. It shouldn’t have access to the list of inbounds. For another, it shouldn’t happen through DataCollection which I thought was an abstraction _for lists of things_ for tasks like filtering in the UI. This is basically an `Array.prototype.filter` call in the form of a component. It doesn’t make sense to me.

johncowen:
@kleinfreund a quick ping as I thought we were going to approve and merge this, maybe I misunderstood?

The last inline conversation was resolved, and whilst I see a load of response there, I don't think its be re-opened. I'd understood from you last non-inline comment [here](https://github.com/kumahq/kuma-gui/pull/2071#issuecomment-1909719863) that you were ok to merge?

Could you let me know what we are waiting on? Do you have anything to add?

kleinfreund:
> @kleinfreund a quick ping as I thought we were going to approve and merge this, maybe I misunderstood?
> 
> The last inline conversation was resolved, and whilst I see a load of response there, I don't think its be re-opened. I'd understood from you last non-inline comment [here](https://github.com/kumahq/kuma-gui/pull/2071#issuecomment-1909719863) that you were ok to merge?
> 
> Could you let me know what we are waiting on? Do you have anything to add?

It sounded like either option that was okay for me to move forward with wasn’t okay with you. In the case of merging this PR, I’m expecting the approach of using DataCollection to be reworked in a follow up:

> merge this PR, but redo the approach of using DataCollection for summary trays in a follow up

However, you made it clear in https://github.com/kumahq/kuma-gui/pull/2071#discussion_r1466244659 that you wouldn’t change your approach

> All in all, whilst the Vue re-rendering thing is something I'd rather avoid, even without that reason, I wouldn't change my approach here.

So, with the PR being as it is, I can’t approve this.

To make it concrete:

- The move of the filtering of the inbound list by selected inbound into the summary view is a step in the wrong direction. I understand that it would be nice for the detail view to not know about the selected inbound. At the same time having the summary view know about all inbounds is worse.
- The use of DataCollection for a find operation that can be accomplished without it easily doesn’t make sense. This is pretty well illustrated by the fact that we’re unconditionally accessing the first item of the resulting `items` property.

I’m happy to consider alternatives to how we pass the right data from a view to its summary view because the current approach has its flaws, but the solution provided here makes things more complicated without avoiding those flaws.

johncowen:
> Let's continue this at https://github.com/kumahq/kuma-gui/pull/2071#issuecomment-1909719863, as I think we should do your option 2

?

kleinfreund:
Are you telling me I’ve misunderstood the following?

> All in all, whilst the Vue re-rendering thing is something I'd rather avoid, even without that reason, I wouldn't change my approach here.

johncowen:
I don't know if you've misunderstood or not, but considering that after that I said 

> Let's continue this at https://github.com/kumahq/kuma-gui/pull/2071#issuecomment-1909719863, as I think we should do your option 2

I dunno, sounds like you were ok me not changing it here in this PR?

---
You are saying lets approve and merge this, I am saying yes lets approve and merge this, yet it continues to be a discussion.

Can we just get this approved and merged please and we can cicle back afterwards.

kleinfreund (approved):

PR #2078, opened 2024-01-23
fix(policies): legacy policy types showing

kleinfreund:
Resolves #2077.

Signed-off-by: Philipp Rudloff <philipp.rudloff@konghq.com>

PR #2079, opened 2024-01-23
chore(policies): fix policy list test

kleinfreund:
Fixes a misspelled policy type name in mock data causing a test to behave incorrectly. Also switches the assertion order to first check for element visibility.

### DIFF
--- a/features/mesh/policies/Index.feature
+++ b/features/mesh/policies/Index.feature
@@ -97,7 +97,7 @@ Feature: mesh / policies / index
             total: 0
           FaultInjection:
             total: 0
-          HealthChecks:
+          HealthCheck:
             total: 0
           MeshGatewayRoute:
             total: 0
@@ -125,10 +125,10 @@ Feature: mesh / policies / index
 
     When I visit the "/meshes/default/policies/meshfaultinjections" URL
 
-    Then the "[data-testid='policy-type-link-FaultInjection']" element doesn't exist
-    And the "[data-testid='policy-type-link-MeshFaultInjection']" element exists
+    Then the "[data-testid='policy-type-link-MeshFaultInjection']" element exists
     # Always shows MeshGateway
     And the "[data-testid='policy-type-link-MeshGateway']" element exists
+    And the "[data-testid='policy-type-link-FaultInjection']" element doesn't exist
 
   Scenario: Shows legacy policy types if there are any legacy policies applied
     Given the URL "/mesh-insights/default" responds with
--- a/features/mesh/policies/Index.feature
+++ b/features/mesh/policies/Index.feature
@@ -124,10 +124,6 @@ Feature: mesh / policies / index
       """
 
     When I visit the "/meshes/default/policies/meshfaultinjections" URL
-    And the URL "/mesh-insights/default" was requested with
-      """
-      method: GET
-      """
 
     Then the "[data-testid='policy-type-link-MeshFaultInjection']" element exists
     # Always shows MeshGateway
@@ -144,10 +140,6 @@ Feature: mesh / policies / index
       """
 
     When I visit the "/meshes/default/policies/meshfaultinjections" URL
-    And the URL "/mesh-insights/default" was requested with
-      """
-      method: GET
-      """
 
     Then the "[data-testid='policy-type-link-FaultInjection']" element exists
     And the "[data-testid='policy-type-link-MeshFaultInjection']" element exists
--- a/features/mesh/policies/Index.feature
+++ b/features/mesh/policies/Index.feature
@@ -124,6 +124,10 @@ Feature: mesh / policies / index
       """
 
     When I visit the "/meshes/default/policies/meshfaultinjections" URL
+    And the URL "/mesh-insights/default" was requested with
+      """
+      method: GET
+      """
 
     Then the "[data-testid='policy-type-link-MeshFaultInjection']" element exists
     # Always shows MeshGateway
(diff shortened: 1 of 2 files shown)


johncowen (commented):
I might be reading this wrong:

> Fixes a misspelled policy type name in mock data causing a test to behave incorrectly. 

I changed the code back to how it was previous to https://github.com/kumahq/kuma-gui/pull/2078 (and previous to the changes to the tests in this PR)

I then _only_ fixed the typo and expected the tests to fail, but they didn't.

I then also used this entire PR but with the old `||` from https://github.com/kumahq/kuma-gui/pull/2078 which seems was the cause of the bug, but all the tests continued to pass?

Maybe I'm missing something? Lmk.

kleinfreund:
Prior to the fix from https://github.com/kumahq/kuma-gui/pull/2078, the test `Hides legacy policy types if there are no legacy policies applied` passes when it should fail.

One cause of this is that the negative assertion happens immediately after the visit step and is almost certain to pass. However, changing this around (like I did here), still doesn’t fail that test.

It turns out that in the scenario, briefly, the UI is showing the filtered list. That happens when mesh insights aren’t available, yet. Once they load, the list is filtered _incorrectly_ which would fail that test finally (which is expected without the fix), but the assertions at this point were already made. We can confirm this by placing a `When I wait for 1000 milliseconds/ms` after the visit step.

So it’s not the misspelled "HealthChecks" (which could interfere with the test through letting the mocks populate the HealthCheck total with a non-zero value) that doesn’t fail the test as expected.

To make the test behave correctly, I’m adding a step to wait for the mesh insight request. This should avoid false positives now.

johncowen:
Awesome sounds good, let me take another look at this.

johncowen:
If I checkout this PR and I change the `&&` back to `||` here:

https://github.com/kumahq/kuma-gui/blob/f15aca9ce25ef4ba3de041782d2eaaddd4fbbe47/src/app/policies/components/PolicyList.vue#L230

And I run the tests they still pass? The idea is they shouldn't if I make that ^ change right?

johncowen:
Ok this is failing for me now, which is great

> Now I’m tried hard to not solve this on the application layer because I don’t want to make waiting for /mesh-insights/:mesh a hard dependency for rendering this view. In those edge cases where /mesh-insights/:mesh isn’t available, yet, we should be able to render the view still. So when that API and with that props.meshInsight?.policies isn’t available, we should show the full policy type list.

I think I get what you are saying, and I almost agree with what you mean if I do follow correctly. I think generally we don't want to be solving test-y things in the application layer, but if I've understood you correctly what you are saying about not wanting a hard dependency on `/mesh-insights/:mesh` sounds like a good application/UX concern, so I'm glad you pointed that out.

The bit I'm not sure on is:

> So when that API and with that props.meshInsight?.policies isn’t available, we should show the full policy type list.

Until we know the counts of policies shouldn't we _not_ show the legacy ones?

kleinfreund:
> Until we know the counts of policies shouldn't we _not_ show the legacy ones?

Definitely no. In that scenario we don’t know if they have any legacy policies applied so we should err on the side of caution: if they have any applied, they’ll see the expected entries. If they don’t, no problem either. But if we hide the legacy policy types and they have some applied, that would be confusing.

johncowen:
You don't think things appearing for a little while and then disappearing after no user interaction might look a little weird for everybody whether they have legacy policies or not? I think things being added with no user interaction is fine (thats basically loading) but in this case disappearing might seem a little weird.

What kinda time period are we expecting here? Its likely to be sub-second right? So with them appearing then disappearing it will look like a little flash of content? Whereas the tiny extra sub-second wait that only somebody that might be using legacy policies would experience would be less of an issue?

This is a tiny little detail so I kinda don't mind in the grand scheme of things, but if it was me I'd definitely go with avoiding the flash of content for everybody, rather than the sub-second wait for somebody who might be using legacy policies.

1 sec I'm gonna approve this anyway, but if you change you mind lmk and we can come back here.

johncowen (approved):
This one is another reason to have a combined "element exists but the "$something" element doesn't exist" or something similar, so that everytime you want to check for something not being there you _have_ to check for some other content first.

---

Anyway, this LGTM now, test fails if I put it back previous to the fix 👍

kleinfreund:
Keep in mind we’re effectively talking about two separate scenarios here:

- mesh insights are still being loaded and arrive delayed. This should, if any, be a barely notable delay
- mesh insights just simply not being available yet. In this case, they _won’t_ arrive at all. That scenario is what I’m covering for with my "erring on the safe side" argument

johncowen:
> mesh insights just simply not being available yet. In this case, they won’t arrive at all. That scenario is what I’m covering for with my "erring on the safe side" argument

What's the response when mesh insights isn't available at all?

kleinfreund:
> What's the response when mesh insights isn't available at all?

404. Not how the DataSource for it doesn’t have error handling. It’s the same we do on the mesh detail view.

johncowen:
> - mesh insights are still being loaded and arrive delayed. This should, if any, be a barely notable delay

The "barely notable delay" in them disappearing is is the "flash of content"

> 404.

Ok cool, thanks. Right so:

> - mesh insights just simply not being available yet. In this case, they won’t arrive at all. That scenario is what I’m covering for with my "erring on the safe side" argument

You'd show them on 404, or any error really.

Me personally I'd want to avoid the flash 🤷 , and from what I understand mesh-insights not being available is pretty rare 🤔 and as I said, I don't really mind in the grand scheme of things.

PR #2082, opened 2024-01-23
feat(dataplanes): add visualization for built-in gateways

### LINKED ISSUE #2049: dataplane view for builtin gateway
### Description

HEre's an example DPP info that should be useful for mocks.

[Archive.zip](https://github.com/kumahq/kuma-gui/files/13978271/Archive.zip)

kleinfreund:
Handle built-in gateways in the new Dataplane detail view display.

Changes the Dataplane stats data source to expect a `service` parameter to account for that part being dynamic in the case of gateways and provides `'localhost_'` for standard proxies as the value for this parameter. This allows us to maintain one data source definition for stats.

Resolves #2049.


src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -11,7 +11,7 @@
   >
     <DataSource
       v-slot="{ data: traffic, error, refresh }: StatsSource"
-      :src="props.data.dataplaneType === 'standard' ? `/meshes/${route.params.mesh}/dataplanes/${route.params.dataPlane}/stats` : ''"
+      :src="props.data.dataplaneType === 'delegated' ? '' : `/meshes/${route.params.mesh}/dataplanes/${route.params.dataPlane}/stats/${props.data.dataplaneType === 'builtin' && props.data.dataplane.networking.gateway ? props.data.dataplane.networking.gateway.tags['kuma.io/service'] : 'localhost_'}`"
```
kleinfreund, line 14: **Comment**: Not in love with this, but I didn’t want to introduce a second source: it would be effectively duplicate the current stats source with the only difference being the hard-coded `localhost_` service prefix.

johncowen: At least short-term, this approach is what we should do.

There are some details though which you might like to tweak but I don't mind either way.

For sidecars the "name of the thing" is "localhost", for gateways its "service-name". The underscore is a little unfortunate, but thats just because its what the envoy line format seems to use to separate the port (but inconsistently between sidecars and gateways it seems). I wonder myself where the port comes into it with gateways.

Following that, other options could be:

- Look for `=== 'localhost'` in the data layer and add the `_`. Whilst that has the potential to clash, I was wondering the other night what happens if I make a `kuma.io/service: localhost_8888`, so I think that could have a different knock on elsewhere. 
- Give `DataPlane` or `DataplaneOverview` an `inboundName/Prefix` property (or similar) in the data layer and derive it there, then use `/meshes/${route.params.mesh}/dataplanes/${route.params.dataPlane}/stats/${route.params.dataPlane.inboundName}` instead.

Whatever you do it will have to happen where we want the DataSource to sync between multiple sources also.

johncowen: Looking at the below code, if you do the second option I mention in this comment, a lot of the conditionals can then depend on the existance of the `inboundName/Prefix` property. If thats empty, don't show the viz.

src/app/data-planes/routes.ts
```diff
@@ -57,6 +57,32 @@ export const routes = () => {
...
+                component: () => import('@/app/data-planes/views/DataPlaneBuiltinGatewaySummaryView.vue'),
+                children: [
+                  {
+                    path: 'overview',
+                    name: 'data-plane-builtin-gateway-outbound-summary-overview-view',
+                    // TODO: Create dedicated outbound component
+                    component: () => import('@/app/data-planes/views/DataPlaneBuiltinGatewaySummaryOverviewView.vue'),
+                  },
+                ],
+              },
```
johncowen, line 85: How come we need different Views for gateways, can't we just use the DataPlane ones that already exist?

kleinfreund: I’m not totally sure about this, yet. Simply put: I don’t know _if_ there are more differences or less. One difference so far is that for standard proxies, we show "Inbound :8080", but that doesn’t work for gateways. That’s why I split them up for now.

That being said, this might totally not how things will shake out. I’ve not made any changes and mostly just copied the summary view content into them.

src/app/data-planes/views/DataPlaneDetailView.vue
kleinfreund: > * Give `DataPlane` or `DataplaneOverview` an `inboundName/Prefix` property (or similar) in the data layer and derive it there, then use `/meshes/${route.params.mesh}/dataplanes/${route.params.dataPlane}/stats/${route.params.dataPlane.inboundName}` instead.

I might do this eventually, but I’m gonna wait and let the rest come together before tweaking this further. What I currently have should be correct and workable for the time being. Would be good to improve though.

src/app/data-planes/routes.ts
johncowen: Ok cool, well my opinion on the general direction here would be:

As things stand they should probably just reuse the same views, similar to how the detail views and tabs are reused between sidecars and dataplanes currently.

Plus there are still things I want to tighten up here and I wouldn't want to duplicate too much _if_ I could avoid it

But will leave it up to you for the min

src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: Sounds good 👍

src/app/data-planes/views/DataPlaneInboundSummaryView.vue
```diff
@@ -15,7 +15,13 @@
           >
 
           <h2 class="summary-title">
-            Inbound :{{ route.params.service }}
+            <template v-if="props.gateway">
+              {{ route.params.service }}
+            </template>
+
+            <template v-else>
+              Inbound :{{ route.params.service.replace('localhost_', '') }}
```
kleinfreund, line 23: **Comment**: I needed to change the route parameter from being just the port to `localhost_${port}` so that I could generalize handling the display of the "compressed" stats code blocks in the various summaries: They can now filter and replace using `${route.params.service}.` and no longer have a hard-coded reference to `'localhost_'`.

johncowen (approved):

PR #2115, opened 2024-01-30
ci: uses commit hashes for third party action versions

kleinfreund:
Replaces all version identifiers for third party actions with commit hashes and the appropriate version comment (Dependabot updates the comment as well). This ensures that we get Dependabot updates for new non-major versions of actions instead of silently opting into using them.


johncowen (approved):
🎉

PR #2119, opened 2024-01-31
chore(policies): fix flaky list view test

kleinfreund:
Add a content-based assertion in the middle of the "Clicking the link goes to the detail page and back again" so that the "navigate" + "assert URL" + "navigate" chain doesn’t trigger the missing route params error.

Remove the `key` attribute from `PolicyList`. It no longer seems needed to trigger a re-render when switching between policy types.


johncowen (approved):
Thanks!

PR #2120, opened 2024-01-31
chore: fix flaky tests relying on incomplete subscription mock data

kleinfreund:
Add explicit `KUMA_SUBSCRIPTION_COUNT` values to tests mocking `subscriptions` with _partial_ objects. If a mocked API has fewer subscriptions (determined by a faker-js dice roll) than what a test mocks out, we can run into errors like "Cannot read lastUpdateTime from undefined". That happens when the extra objects from the test have partial data (like not mocking out `status` in the case of the above error). I added more variables than necessary (as we usually create at least one subscription object and so tests only adding one don’t flake) so we’re on the safe side.

### DIFF
--- a/features/main-overview/DetailViewContent.feature
+++ b/features/main-overview/DetailViewContent.feature
@@ -105,6 +105,7 @@ Feature: Overview: Detail view content
       KUMA_ZONE_COUNT: 2
       KUMA_MESH_COUNT: 3
       KUMA_MODE: global
+      KUMA_SUBSCRIPTION_COUNT: 2
       """
     And the URL "/global-insight" responds with
       """
--- a/features/onboarding/dataplanes-overview/Index.feature
+++ b/features/onboarding/dataplanes-overview/Index.feature
@@ -21,6 +21,7 @@ Feature: onboarding / dataplanes-overview / index
     Given the environment
       """
       KUMA_DATAPLANE_COUNT: 1
+      KUMA_SUBSCRIPTION_COUNT: 1
       """
     And the URL "/dataplanes/_overview" responds with
       """
--- a/features/zones/zone-cps/egresses/Item.feature
+++ b/features/zones/zone-cps/egresses/Item.feature
@@ -9,6 +9,7 @@ Feature: zones / egresses / item
     And the environment
       """
       KUMA_MODE: global
+      KUMA_SUBSCRIPTION_COUNT: 2
       """
 
   Scenario: Detail view has expected content
--- a/features/zones/zone-cps/ingresses/item/Index.feature
+++ b/features/zones/zone-cps/ingresses/item/Index.feature
@@ -11,6 +11,7 @@ Feature: zones / ingresses / item
     And the environment
       """
       KUMA_MODE: global
+      KUMA_SUBSCRIPTION_COUNT: 2
       """
 
   Scenario: Clicking through the secondary navigation


johncowen (commented):
Thanks for looking at this!

One thing that could be in relation this:

> I added more variables than necessary (as we usually create at least one subscription object and so tests only adding one don’t flake) so we’re on the safe side.

If we are adding the `SUBSCRIPTION_COUNT` "globally" in the background then I think we should remove the places where we set it per test. wdyt?

kleinfreund:
> One thing that could be in relation this:
> 
> > I added more variables than necessary (as we usually create at least one subscription object and so tests only adding one don’t flake) so we’re on the safe side.
> 
> If we are adding the `SUBSCRIPTION_COUNT` "globally" in the background then I think we should remove the places where we set it per test. wdyt?

We could preset variables like this globally, yes. Depending on which value we set it to, those instances can be removed from the tests.

Without that, the changes here should be a good bit more robust.

johncowen:
Not sure if I explained myself properly.

If I'm not mistaken the tests here set `KUMA_SUBSCRIPTION_COUNT` in two places, both in the `Background` and in the `Scenario`, we should only do it in one place.

For example:

https://github.com/kumahq/kuma-gui/pull/2120/files#diff-e83ea8d5259783656d29e3a191f4460554a0457e6462aef94d0ff765ae3782c3R11
https://github.com/kumahq/kuma-gui/pull/2120/files#diff-e83ea8d5259783656d29e3a191f4460554a0457e6462aef94d0ff765ae3782c3L45


If the approach here is to use `Background`, then we should probably remove it from the `Scenario` (or alternatively add it to the Scenarios where it's missing)

johncowen:
👍 

I just noticed does this bit need a `disconnectTime: !!js/undefined` to avoid a potential flake?

https://github.com/kumahq/kuma-gui/pull/2120/files#diff-61c9d2287811884ae04444895cae863a245228372fdf8ccfea08a2b9def96e62R146-R147

?

kleinfreund:
> I just noticed does this bit need a `disconnectTime: !!js/undefined` to avoid a potential flake?
> 
> https://github.com/kumahq/kuma-gui/pull/2120/files#diff-61c9d2287811884ae04444895cae863a245228372fdf8ccfea08a2b9def96e62R146-R147
> 
> ?

Hm. I’m not sure. I don’t recall this test being flaky and it would be pretty obvious where it fails if it were. I’d rather leave this is as-is and see if it actually happens.

johncowen:
Well looking at:

```gherkin
And the "[data-testid='zone-control-planes-status']" element contains "1/2"
```

https://github.com/kumahq/kuma-gui/pull/2120/files#diff-61c9d2287811884ae04444895cae863a245228372fdf8ccfea08a2b9def96e62R166

If https://github.com/kumahq/kuma-gui/pull/2120/files#diff-61c9d2287811884ae04444895cae863a245228372fdf8ccfea08a2b9def96e62R146-R147 randomly had a `disconnectTime` that wasn't `undefined`, it would fail the test wouldn't it? `"[data-testid='zone-control-planes-status']"` would contain `0/2`?

kleinfreund:
> If https://github.com/kumahq/kuma-gui/pull/2120/files#diff-61c9d2287811884ae04444895cae863a245228372fdf8ccfea08a2b9def96e62R146-R147 randomly had a `disconnectTime` that wasn't `undefined`, it would fail the test wouldn't it? `"[data-testid='zone-control-planes-status']"` would contain `0/2`?

I understand that, but it doesn’t seem to be doing that at all. Have you seen this test flake?

johncowen:
> Have you seen this test flake?

I wouldn't remember if I had seen it flake or not, but if this PR is to "fix flakey tests relying on incomplete subscription mock data" this one feels like it theoretically could flake due to relying on incomplete subscription mock data. I'd rather preempt that rather than hope it will never flake. i.e. prevention rather than cure?

Is there a reason not to add this? it looks like the test case depends on there being a connected subscription and as our mocks introduce randomness its likely that at some point it will flake (if it hasn't already)?

kleinfreund:
> I wouldn't remember if I had seen it flake or not, but if this PR is to "fix flakey tests relying on incomplete subscription mock data" this one feels like it theoretically could flake due to relying on incomplete subscription mock data. I'd rather preempt that rather than hope it will never flake. i.e. prevention rather than cure?
> 
> Is there a reason not to add this? it looks like the test case depends on there being a connected subscription and as our mocks introduce randomness its likely that at some point it will flake (if it hasn't already)?

There’s a reason not to add this: the test doesn’t currently behave flaky.

I’m not gonna add it. I see no reason why.

johncowen (approved):
Lets get this in as is then

PR #2126, opened 2024-02-01
chore: add DataLoader component

johncowen:
Adds a new `DataLoader` component, which is can be used as a bit of a "DataSource-with-loading-and-error-states" or a "DataSource-with-dependant-data-and-errors" or a mix of the two.

This is a little bit of a WIP in that I 'think' this is the API I want, I've purposefully used it in three areas to try it out and get a feel for it:

1. A simple 'item' view in Diagnostics, where I pretty much just want a DataSource with automatic loading and error states.
2. A listing that has a visual skeleton in our Mesh listing, where I pretty much just want a DataSource with just an error state (the loading state is contained in the visual we feed the data to)
3. A more complex UI (DataPlanePolicy view) where we load in several sets of data which are sometimes dependant on each other but sometimes not, here I want to load data without any blocking loading states and also load data with a blocking loading state, that also depends on the previous non-blocking data load (plus errors with the same dependency idea). I also use DataCollection here to help with Empty states seeing as we don't have `AppCollection` with its embedded Empty state.

---

You can use DataLoader in one of three ways:

1. Exactly like a DataSource. Give it a `src` and it will load the data, but unlike DataSource itself, it will block showing its child nodes until the data is loaded. If there is an error it will automatically show a default error state using our ErrorBlock.
2. Instead of giving it a `src` you can give it an array "dependant data/errors". DataLoader will block showing its child nodes until _all_ of the dependant data is loaded, or if _any_ of the errors is defined, then show the error state.
3. A mix of the above two, i.e. a `src` _and_ dependent data/errors. DataLoader will block showing its child nodes until _all_ of the dependant data _and_ the `src` is loaded, or if _any_ of the errors is defined (including the potential error from the `src`, then show the error state.

All of the above can also use a `loading="true/false"` which will control whether to show the child nodes when in a loading state or not (sometimes we have components that have their own loading state).

---

Separately I noticed that the more complex DataPlanePolicyView seemed a little simpler to follow using DataLoader as the majority of the loading/error/empty functionality happens elsewhere.

Related: https://github.com/kumahq/kuma-gui/issues/1474

### DIFF
--- a/src/app/application/components/data-source/DataLoader.vue
+++ b/src/app/application/components/data-source/DataLoader.vue
@@ -1,0 +1,100 @@
+<template>
+  <DataSource
+    v-slot="{ refresh }"
+    :src="props.src"
+    @change="(data) => srcData = data"
+    @error="(e) => srcError = e"
+  >
+    <template
+      v-if="allData.length > 0 && allData.every(item => typeof item !== 'undefined')"
+    >
+      <!-- if the data is loaded and then we get an error -->
+      <!-- keep the old data visible but also show a disconnected slot -->
+      <template
+        v-if="allErrors.length > 0"
+      >
+        <slot
+          name="disconnected"
+          :data="props.src !== '' ? allData[0] : undefined"
+          :errors="props.src !== '' ? allErrors[0] : undefined"
+          :refresh="props.src !== '' ? refresh : () => {}"
+        >
+          <!-- KAlert -->
+        </slot>
+      </template>
+      <slot
+        name="default"
+        :data="props.src !== '' ? allData[0] : undefined"
+        :errors="props.src !== '' ? allErrors[0] : undefined"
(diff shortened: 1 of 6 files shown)


src/app/application/components/data-source/DataLoader.vue
```diff
@@ -0,0 +1,100 @@
...
+</template>
+<script lang="ts" setup>
+import { computed, ref } from 'vue'
+
+import ErrorBlock from '@/app/common/ErrorBlock.vue'
+import LoadingBlock from '@/app/common/LoadingBlock.vue'
+
+const props = withDefaults(defineProps<{
+  data: any[]
+  errors: (Error | undefined)[]
```
kleinfreund, line 75: **Question**: What’s the thought behind accepting data and error as arrays? We seem to use only the first one of them.

kleinfreund (approved):

PR #2127, opened 2024-02-01
chore: Remove non-unique HTML id's

johncowen:
Removes all instances of non-unique/hardcoded id usage.

Theres a thing I want to explain/ask which I'll note inline.

### DIFF
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -12,6 +12,7 @@
       aria-atomic="true"
     />
     <slot
+      :id="UniqueId"
       name="default"
       :t="t"
       :env="env"
@@ -31,7 +32,7 @@ import { computed, provide, inject, ref, watch, onBeforeUnmount, reactive } from
 import { useRoute, useRouter } from 'vue-router'
 
 import { ROUTE_VIEW_PARENT } from '.'
-import { useCan, useI18n } from '../../index'
+import { useCan, useI18n, uniqueId } from '../../index'
 import {
   urlParam,
   normalizeUrlParam,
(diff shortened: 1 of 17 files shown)


src/app/application/components/route-view/RouteView.vue
```diff
@@ -67,7 +68,16 @@ const props = withDefaults(defineProps<{
...
+const _id = uniqueId(props.name)
+class UniqueId {
+  static toString() {
+    return _id
+  }
 
+  toString() {
+    return uniqueId(props.name)
+  }
+}
```
johncowen, line 80: Yeah I know right? 😅 

So this is a bit of a magic string that you can either:

```
{{ id }} => 'name-of-the-route-12345' << this one never changes no matter how many times you use it so it's unique to the route but not the usage
{{ id }} => 'name-of-the-route-12345' << this one never changes no matter how many times you use it so it's unique to the route but not the usage
{{ id }} => 'name-of-the-route-12345' << this one never changes no matter how many times you use it so it's unique to the route but not the usage

{{ new id }} => 'name-of-the-route-12345' << this one changes everytime you call it
{{ new id }} => 'name-of-the-route-12346' << this one changes everytime you call it
{{ new id }} => 'name-of-the-route-12347' << this one changes everytime you call it
```

The above approach felt better than overwriting the `toString` of a function (`(() => {}).toString = () => {}`) as there is no overwriting and its perfectly valid javascript and typescript.

Alternatively, I can just:

1. Keep it as a function and use `<template v-for="id in [uniqueId()]">` at the very top of the template
2. v-slot both a `{staticId, uniqueIDCreatingFunction}` so you explicitly choose which one to choose.
3. Keep the static `v-slot="{id}"` without the funny business (i.e. id is just a string) and then use `<input id="${ id }-1" />`, `<input id="${ id }-2" />` i.e. manually make the id's unique within the same file.

johncowen (commented):
Leaving the inline comment on the 'thing':

kleinfreund (approved):

PR #2130, opened 2024-02-02
chore: ensure defineSources returns the correct return types

johncowen:
A little while back in https://github.com/kumahq/kuma-gui/pull/1525 we added automatic types to our sources

> `"/services/:name": (params) => { params.name }`

The thing I didn't realise at the time was we had lost the return type of the source.

This PR corrects that so now the sources have automatic types for their URI based parameters, plus the resulting return types are correct for each individual source. Now this is correct it means we can start looking at automatically passing these through to our DataSources so that the URIs auto-complete and the emitted data from the DataSources correctly corresponds to the URIs (see https://github.com/kumahq/kuma-gui/issues/1474)

---

I've not gone ahead and rolled this out as yet, but I'd like to after this PR has gone in in a separate PR. In the meantime you can look at `data-planes/sources.ts` which already uses `defineSources`.

### DIFF
--- a/src/app/application/services/data-source/index.ts
+++ b/src/app/application/services/data-source/index.ts
@@ -7,8 +7,14 @@ export type DataSourceResponse<T> = {
   error: Error | undefined
   refresh: () => void
 }
+type PaginationParams = {
+  size: number
+  page: number
+  search: string
+  cacheControl: string
+}
 
-type ExtractRouteParams<T extends string> =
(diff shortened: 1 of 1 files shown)


kleinfreund (approved):

johncowen:
Small note: I remembered a couple of extra things I wanted to do in the changes here, but that are mainly naming/typescript-y things so I can do that in a later PR, gonna merge this.

PR #2136, opened 2024-02-05
fix: prevents a blinking upgrade alert

johncowen:
The fix and the comment in the code probably explains this PR the best, please ask if its not clear tho.

I'll inline comment the fix as there is also the version bump in here.

Tangent: Ideally this version needs to be in one place (in the package.json file), so I should be moving to do that at some point soonish.

### DIFF
--- a/src/app/kuma/components/ApplicationShell.vue
+++ b/src/app/kuma/components/ApplicationShell.vue
@@ -26,8 +26,9 @@
               v-slot="{ data }"
               :src="`/control-plane/version/latest`"
             >
+              <!-- make sure we have data but don't show errors or loaders -->
               <KAlert
-                v-if="env('KUMA_VERSION') !== data?.version"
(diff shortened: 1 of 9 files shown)


src/app/kuma/components/ApplicationShell.vue
```diff
@@ -26,8 +26,9 @@
               v-slot="{ data }"
               :src="`/control-plane/version/latest`"
             >
+              <!-- make sure we have data but don't show errors or loaders -->
               <KAlert
-                v-if="env('KUMA_VERSION') !== data?.version"
+                v-if="data && env('KUMA_VERSION') !== data.version"
```
johncowen, line 31: This is the fix, when data is `undefined` it doesn't equal `KUMA_VERSION`, which means the upgrade alert can show for a sort period (and every time you refresh the browser)

PR #2150, opened 2024-02-06
feat(services): move external services into separate tab

### LINKED ISSUE #2133: Move external service out of the services list
### Description

We've decided to move these to their own tab in the mesh listing as it's clearer

kleinfreund:
Split the services module into a services (for `ServiceInsight`s) and an external services (for `ExternalService`s) modules.

Resolves #2133.


kleinfreund:
Strangely, the very simple test I added in features/mesh/external-services/Item.feature fails in CI but not locally. I have no idea why. The root cause is that the external service routes aren’t available which means that the route setup didn’t work the same as it does locally.

kleinfreund:
@johncowen Does https://github.com/kumahq/kuma-gui/pull/2150#issuecomment-1930042185 ring a bell? No idea why those tests would only fail in CI.

johncowen:
1 sec just pulling this down to look properly

johncowen:
Ok so I think this is because in kuma-land our feature flags are only available in a development build (we only add the cookied env there). There are other envs vars that work in production but that is because either:

1. They are proper prod env vars that we pass through via the index.html
2. They work on the mocks and are passed through to those via Cypress.

This new one isn't one of the above types.

For now, off the top of my head, if you want to be writing tests for this here, verify the tests locally and then `@skip` the tests for the PR. When we remove the feature flag we can remove the `@skip` and they will all pass. If they don't pass we won't be able to get the PR that removes the feature flag through, and therefore it won't be deployed to users, which means we are still achieving the thing we want to achieve - failing tests blocking deployment to users.

Another option would be elsewhere to hook the feature flag up to a proper launchdarkly feature flag and test over there, although I prefer that less, the tests should stay with the code.

Lastly, in the past I've added a different type of feature flags for prod/end users to use to try things out. These wouldn't be based on cookies, but I don't think we should do this for this feature.

TLDR; I'd go with the test locally and @skip route for the moment.

kleinfreund:
> For now, off the top of my head, if you want to be writing tests for this here, verify the tests locally and then `@skip` the tests for the PR.

Yeah this is what I was gonna do to continue. Thanks!

I don’t think we need a user-facing proper feature flag for this. It’s rather a "development safety and isolation" flag. So this should be adequate.

johncowen:
Pretty sure this is going to be re-worked, let me know if not @kleinfreund otherwise ping me when its ready again ty!

kleinfreund:
> Pretty sure this is going to be re-worked, let me know if not @kleinfreund otherwise ping me when its ready again ty!

I’ve reworked this now. I went for a more minimal approach to integrate the external services list into the Services tab mainly with the objective to keep the current route structures of the services module intact. I didn’t want to change routes conditionally based on the feature flag and I wanted to avoid the route overlapping of `services/:internalService` and something like `services/external/:externalService`.

I also didn’t bother tweaking the visuals of the "secondary nav tabs" too much and plan to tweak them later.

src/app/external-services/data/index.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+              'kuma.io/service': 'service',
+            },
+          },
+        },
+      },
+    ])('.fromObject: $message', ({ parameters, expected }) => {
+      expect(ExternalService.fromObject(...parameters)).toStrictEqual(expected)
+    })
+  })
+})
```
johncowen, line 60: Could this use the data layer test plugin, simila rot Zones and Dataplanes? I think it would also make it clearer what this is testing (I think its just the `.config`?

src/app/external-services/views/ExternalServiceListView.vue
```diff
@@ -0,0 +1,159 @@
...
+                :to="{
+                  name: 'external-service-list-view',
+                  params: {
+                    mesh: route.params.mesh,
+                  },
+                }"
+              >
+                <b>External</b>
+              </RouterLink>
+            </LinkBox>
```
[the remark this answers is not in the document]
kleinfreund: I haven’t looked at the design yet so I’m not entirely sure what we’ll need. In any case, this is unfinished and I’ll pass over this again once we have them. During that step, it might well be that I’m swapping this out again.

In short, there will be a PR adjusting the changes here (and the future ones for gateways) to the design.

src/app/external-services/sources.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+
+    '/meshes/:mesh/external-services/:name/as/kubernetes': async (params: CollectionParams & { name: string }) => {
+      const { mesh, name } = params
+      return api.getExternalService({ mesh, name }, {
+        format: 'kubernetes',
+      })
+    },
+  }
+}
```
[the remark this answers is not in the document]
kleinfreund: Sure, I’ll change this up.

kleinfreund:
> I not totally sure about moving a "type" of service into its own module, but it might be because we are feature flagging this, I need to look at it a bit more.

We’re developing this feature behind a feature flag and I don’t want to touch the existing services URL paths. That’s one reason why I moved external services into its own module: to have separate external-services/ URL paths. From the business logic side of things, I also think this makes sense because we do want to treat Internal Services and External Services as largely distinct things. They’re different entities (ones a resource, one isn’t), too.

This might change again if it ends up being more convenient to handle all this within one services module but for the time being, I think this is the right way to go.

johncowen:
> I not totally sure about moving a "type" of service into its own module, but it might be because we are feature flagging this, I need to look at it a bit more.

We chatted a bit offline about this, and its mainly because whilst from a user perspective these are all "services" and there are two different "types", when it comes to the data of the things they are entirely different, so kinda make sense to have a separate module, at least short term (this may change in future with `MeshService` etc)

URLs-wise, sounds like this is more for feature flagging reasons which also make sense to me. Although once we are in a position to remove the feature flag, do we think it would make more sense to have `/mesh/default/services/internal` & `/mesh/default/services/external` (i.e. have the segmented control in the `/mesh/default/services` route as the parent, and then nest `/internal` / `/external` under that?)

src/app/external-services/data/index.spec.ts
kleinfreund: This file’s content is extracted from the existing services module tests. They will probably need touching ones we work on actual changes to the display of external services so we can handle changing the test setup for this then.

johncowen:
Ah we both wrote at the same time 😆 I think both comments more or less match up, let me know on the `defineSources` and test plugin thing, and in the meantime I'm gonna take another look at this.

src/app/external-services/views/ExternalServiceListView.vue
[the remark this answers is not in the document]
kleinfreund: For reference: https://kongponents.konghq.com/components/segmented-control.html

It looks like KSegmentedControl always wraps its elements into a button element with click handler. This would be problematic for us since we’d be using links inside the elements.

src/app/external-services/sources.ts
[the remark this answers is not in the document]
kleinfreund: Done.

src/app/external-services/data/index.spec.ts
johncowen: Oh gotcha, sounds good to me.

src/app/external-services/views/ExternalServiceListView.vue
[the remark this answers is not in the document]
johncowen: Doh, yeah not ideal at all. We should consider bring this up with the kongponents folks I think, feels very much like the KTabs component not being able to work with links, we can chat about that later. Will resolve this for the moment.

src/shims-vue-router.d.ts
```diff
@@ -6,5 +6,10 @@ declare module 'vue-router' {
      * The module this route belongs to
      */
     module?: string
+
+    /**
+     * Whether to ignore this route in the forming of navigational tabs
+     */
+    shouldIgnoreInNavTabs?: boolean
```
johncowen, line 13: How come this is needed?

johncowen: Did I understand right that we would have a nested routing structure here once we remove the feature flag? i.e.

- `/mesh/default/services/`
  - `/mesh/default/services/internal`
  - `/mesh/default/services/external`

If so that would mean that this ceases to be something we need to work around because we are using nested routing. That being said can we just filter out the route with the name `external-service-list-view`. FYI I don't want this sort of `route.meta` to become a thing. If for some reason this sort of workaround is needed, its more likely that the code is going against the natural nested routing approach, and its more important to stick with that nested routing approach rather than workaround things.

Lastly I get why you are trying to not alter the URL structure too much for the feature flag reasons, but this is also why I think its fine to just use the route name in the filter instead of adding new `route.meta` i.e. its a temporary thing.

johncowen (commented):
I think apart from the `shouldIgnoreInTabNavs` thing, this looks good to me. So if we can sort that out I can give this a 👍 I think

src/shims-vue-router.d.ts
kleinfreund: > Did I understand right that we would have a nested routing structure here once we remove the feature flag? i.e.
> 
> * `/mesh/default/services/`
>   
>   * `/mesh/default/services/internal`
>   * `/mesh/default/services/external`

I haven’t made up my mind on this, yet. _Internal_ services aren’t really a thing. We have ServiceInsights which represent all manner of service in Kuma which is confusing because those services aren’t very similar. We might get a dedicated resource in MeshService that will replace ServiceInsight, but that’s not a concrete plan, yet. The term "Internal" kind of just appears because we have ExternalServices. Long story, short: I want to avoid the notion that we have two types of services that are just slight variations.

To that end, I prefer the routes using `services/` (which in the future could be replaced by `mesh-services/`) and `external-services/`.

> If so that would mean that this ceases to be something we need to work around because we are using nested routing. That being said, is it possible to just filter out the route with the name `external-service-list-view`? FYI I don't want this sort of `route.meta` to become a thing. If for some reason this sort of workaround is needed, its more likely that the code is going against the natural nested routing approach, and its more important to stick with that nested routing approach rather than workaround things.
> 
> Lastly I get why you are trying to not alter the URL structure too much for the feature flag reasons, but this is also why I think its fine to just use the route name in the filter instead of adding new `route.meta` i.e. its a temporary thing.

Not sure what’s bad using `route.meta` as a place to store such filter decisions. It’s functionally the same as filtering the route by name but simpler.

It’s true that nesting both types of services under `services/` would allow us to avoid having to filter the external service route from participating in nav tabs generation, but before I know what URL paths we’ll settle on, I’m not going to re-write the route setup for this PR (again) just so I can avoid the small additions to `route.meta`.

kleinfreund: > Is there any other reason to feature flag this "services to internal/external services" feature?

The other reason was that this set of features will take multiple PRs to land. We want to minimize the amount of UI reorganization changes.

The feature flag will remain in place during the development here even if the necessary API capabilities are available.

johncowen: Ok so if the only potential reason not to do nested routing here is possibly due to the feature flagging, will this be moved to use nested routing once we remove the feature flag?

kleinfreund: Possibly. I’m not making a decision on this now.

johncowen:
Should this be happening now, or is it just our mocks?

Screenshot 2024-02-14 at 11 17 30

kleinfreund:
> Should this be happening now, or is it just our mocks?

This was fixed in https://github.com/kumahq/kuma-gui/pull/2150/commits/a310d1525980ca81c6db0aa9219c80e6a8204efc. I had updated the name of the parameter in response to the backend PR going in (which uses `type` rather than `serviceType`).

johncowen (approved):
Ok cool, thanks for sorting that.

Lets go with this then

PR #2151, opened 2024-02-06
chore: various amends for gateway data-viz

johncowen:
Various amends for the Dataplane Gateway Viz

- Adds an `inboundName` and `inbound[].name` properties at the data layer level (see https://github.com/kumahq/kuma-gui/pull/2082#discussion_r1466188690)
- Fills in a single inbound for built-in gateways based on inferring the data from the tags of the gateway, again in the data layer. This is only temporary, gateways can have multiple inbounds.
- Using the above, removes the duplication/fork in `DataPlaneDetailView` for showing the Traffic components. This also has the benefit that all the empty/loading/error states work the same as the view when we are viewing a sidecar/standard dataplane.
- Replaces the hardcoded `HTTP` protocol for the current single gateway inbound with the correct protocol from the stats endpoint.
- Replaces instances of a `Cluster` 'protocol' with `tcp` for gateway outbounds. Whilst this is hardcoded to `tcp` for gateway outbounds only, I figure its better than showing `Cluster` for a protocol. Figuring out the protocol for gateway outbounds is going to be a little more involved using the `listeners` property of the `/stats` endpoint.
- I hid our filled out gateway inbounds from anywhere else we show inbounds for the moment at least, using DataCollection to do that in places, although I'm guessing that if a gateway has inbounds then we should also show them in other places (we currently don't)

All in all while we still don't show multiple inbounds if there are any, and the protocol for the outbounds could be wrong. I figured there was enough improvement here (no duplication, better states, no `Cluster` protocol etc) to PR something and continue with these issues.

~Lastly, filling out the tests here was a little unwieldy and makes it harder to add our own properties in the data layer. I did enough here to make the tests pass by adding all the new properties to all the dataplane tests, but I'm thinking my next PR could be replacing these tests with the approach we use in Zones which is easier to manage and test.~ This is no longer the case now https://github.com/kumahq/kuma-gui/pull/2154 has gone in an rebased

### DIFF
--- a/src/app/data-planes/data/index.ts
+++ b/src/app/data-planes/data/index.ts
@@ -29,19 +29,19 @@ import { isSet } from '@/utilities/isSet'
 
 export type { TrafficEntry } from './stats'
 
-export type DataplaneInbound = PartialDataplaneInbound & {
+export type Connection = {
+  name: string
+  service: string
+  protocol: string
+}
+export type DataplaneInbound = PartialDataplaneInbound & Connection & {
   health: {
     ready: boolean
   }
-  service: string
   addressPort: string
   serviceAddressPort: string
-  protocol: string
 }
-
-export type DataplaneOutbound = PartialDataplaneOutbound & {
-  service: string
-  protocol: string
+export type DataplaneOutbound = PartialDataplaneOutbound & Connection & {
 }
 
 export type DataplaneGateway = PartialDataplaneGateway & {}
@@ -49,6 +49,8 @@ export type DataplaneGateway = PartialDataplaneGateway & {}
 export type DataplaneNetworking = Omit<PartialDataplaneNetworking, 'inbound' | 'outbound'> & {
   inbounds: DataplaneInbound[]
   outbounds: DataplaneOutbound[]
+  inboundName: string
+  type: 'delegated' | 'builtin' | 'standard'
 }
 
 export type Dataplane = PartialDataplane & {
@@ -85,7 +87,7 @@ export type MeshGatewayDataplane = PartialMeshGatewayDataplane & {
 }
 
 const DataplaneNetworking = {
-  fromObject(networking: PartialDataplaneNetworking): DataplaneNetworking {
+  fromObject(networking: PartialDataplaneNetworking, defaultHealth: boolean): DataplaneNetworking {
     // remove singular inbound/outbound to be replaced with plural versions
     const { inbound, outbound, ...rest } = networking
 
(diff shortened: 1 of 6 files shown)


src/app/data-planes/data/index.ts
```diff
@@ -29,26 +29,28 @@ import { isSet } from '@/utilities/isSet'
...
+export type DataplaneOutbound = PartialDataplaneOutbound & Connection & {
 }
 
 export type DataplaneGateway = PartialDataplaneGateway & {}
 
 export type DataplaneNetworking = Omit<PartialDataplaneNetworking, 'inbound' | 'outbound'> & {
   inbounds: DataplaneInbound[]
   outbounds: DataplaneOutbound[]
+  inboundName: string
+  type: 'delegated' | 'builtin' | 'standard'
```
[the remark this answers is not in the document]
johncowen: We use the `type` further below in `networking` to figure out whether we just reshape sidecar inbounds that we are given in the response, or we "make" one for the builtin gateway based on the info we have in networking. So we need to know the `type` here.

Then, instead of figuring it out again for `DataPlaneOverview` we just do:

```
DataplaneOverview.type = DataplaneOverview.DataplaneNetworking.type
```

I would guess at some point in the future we will always access the "networking type" (i.e. sidecar or gateway) from `dataplaneOverview.dataplane.networking.type` and then lose the DataPlaneOverview one, but at the minute that would require more change that isn't required in this PR.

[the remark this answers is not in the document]
kleinfreund: I see why that makes it beneficial for the property to be accessible on `networking`.

It be nicer though to have `dataplaneType` on the top-level of a resource.

As a benefit of it being on `networking`, it also becomes available on `Dataplane`s.

src/app/data-planes/data/index.ts
```diff
@@ -141,9 +165,9 @@ const DataplaneInsight = {
 export const DataplaneOverview = {
   fromObject(partialDataplaneOverview: PartialDataplaneOverview, canUseZones: boolean): DataplaneOverview {
     const dataplaneInsight = DataplaneInsight.fromObject(partialDataplaneOverview.dataplaneInsight)
-    const networking = DataplaneNetworking.fromObject(partialDataplaneOverview.dataplane.networking)
 
-    const dataplaneType = getDataplaneType(networking)
+    const networking = DataplaneNetworking.fromObject(partialDataplaneOverview.dataplane.networking, typeof dataplaneInsight.connectedSubscription !== 'undefined')
```
kleinfreund, line 169: **Question**: Could you explain this addition a bit, please?

How come the health property becomes dependent on data from `dataplaneInsight` here?

src/app/data-planes/data/index.ts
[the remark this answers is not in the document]
johncowen: Not sure if we are on the same page here or not, it sounds like maybe we are, did you notice that currently there is still a `dataplaneType` on the top level resource? So the same value is in two places:

https://github.com/kumahq/kuma-gui/pull/2151/files#diff-6cd4b378952d5211d00c3e499e5d33d9d1c1e40cfcc5e470b4940a5d97eaa8a1R184

I think moving forwards there is a possibility we'll change this so its only accessible from one place `dataplaneOverview.dataplane.networking.type`. I don't really see a reason if a top-level one is just for avoiding typing the word `.networking.` i.e. `dataplaneOverview.dataplane.networking.type` vs `dataplaneOverview.dataplaneType`.

I think I can resolve this one but let me know if not.

src/app/data-planes/data/index.ts
kleinfreund: My question is about the business details of this change and not about code organization. So when I say "the health property becomes dependent on data from `dataplaneInsight`" that is definitely the case: for the single inbound here, ready will change depending on whether there is a connected subscription as you explain here:

> In "filling it out" we don't want to always hardcode the health `ready: true` and therefore show it as "online" because in the case of gateways thats not true if there is no `connectedSubscription`. Therefore we pass in a `true`/`false` based on whether there is a `connectedSubscription`.

My question is basically: does that make sense?

This doesn’t happen anywhere else. We don’t change the health object of inbounds if a regular Dataplane has no connected subscription so I find it weird to have this influence here. If we show an individual inbound’s readyness in places, it’s confusing to me that sometimes, the presence of a connected subscription would change the readyness status and sometimes it wouldn’t.

johncowen: Just to be clear because you mention a "regular Dataplane" are you asking:

> does that make sense for sidecars and gateways?

or

> does that make sense for gateways?

?

kleinfreund: The latter:

> does that make sense for gateways?

johncowen: Ok thanks. What single thing do we use currently to understand the status of a gateway dataplane?

kleinfreund: The presence of a connected subscription. But what we don’t do, is make that information influence an unrelated property (the health object on inbounds).

johncowen: Ok:

1. We don't use this `builtin.inbound.health.ready` information anywhere as yet, its currently just a base default.
2. In order to show "builtin gateway status" everywhere in the GUI we use the presence of a connected subscription.
3. We have a choice between two options, `builtin.inbound.health.ready:` `true` or `false`. We don't want a three-way boolean. We either pick one based on nothing, or we pick one based on something _as a default_. To be honest in this case if we didn't want to use whether the dataplane is connected or not _as a default_ for inbound health, I'd probably rather it was `false`.

A slightly related question, if the dataplane isn't connected to the dataplane, how do you know the health of the inbound?

kleinfreund: > A slightly related question, if the dataplane isn't connected, how do you know the health of the inbound?

I don’t know.

I guess let’s resolve this, but I definitely think it’s weird that we write to an object here that we usually use exclusively as the source of data. The fact that we don’t have this data from the get go kind of implies there’s no meaningful information here that could be called "inbound health". It’s extra weird that this behaves differently depending on whether we "make up" this single inbound or use inbounds from the API resource.

kleinfreund (approved):

PR #2158, opened 2024-02-08
feat(gateways): add new gateway views

### LINKED ISSUE #1994: Gateway view
### Description

Currently gateways are not very liked.

MeshGateway are shown in policies when they are not policies. Hard to make a difference between a provided gateway and a mesh gateway...

We should probably add a tab "Gateways" to:

image

In the case of a builtin gateway we'd be able to show the routes etc attached.

kleinfreund:
Add a new Gateways tab to the meshes modules. It links to a built-in gateway list view which is a listing of MeshGateway objects (which represent built-in gateways). The list’s entries link to a built-in gateway detail view. A secondary navigation provides a link to a new delegated gateway list view which is a listing of ServiceInsight objects of type `gateway_delegated`).

Remove both built-in and delegated ServiceInsight objects from the service list view.

Resolves #1994.

---

Related pull requests:

- https://github.com/kumahq/kuma-gui/pull/2149
- https://github.com/kumahq/kuma-gui/pull/2150.


src/app/gateways/routes.ts
```diff
@@ -0,0 +1,82 @@
...
+  const item = (): RouteRecordRaw[] => {
+    return [
+      {
+        path: 'gateways',
+        name: 'gateway-detail-view',
+        children: [
+          {
+            path: 'builtin/:gateway',
+            name: 'builtin-gateway-detail-tabs-view',
+            component: () => import('@/app/gateways/views/BuiltinGatewayDetailTabsView.vue'),
```
kleinfreund, line 13: **Comment**: We technically don’t need tabs here, yet, but I think there’s a good chance we will soon. In any case, we can always remove it if we don’t have any use for tabs here.

src/app/policies/components/PolicyList.vue
```diff
@@ -94,7 +94,7 @@
             :empty-state-cta-text="t('common.documentation')"
             :headers="[
               { label: 'Name', key: 'name' },
-              ...(props.currentPolicyType.isTargetRefBased ? [{ label: 'Zone', key: 'zone' }] : []),
+              ...(props.currentPolicyType.isTargetRefBased || props.currentPolicyType.name === 'MeshGateway' ? [{ label: 'Zone', key: 'zone' }] : []),
```
kleinfreund, line 97: **Comment**: This is technically a small bug fix. We should show the Zone column for MeshGateway. However, this will soon go away again because MeshGateways won’t be shown under policies in the future.

johncowen (approved):
Looks great! I had a few things to chat about but nothing blocking, and most of this we can probably cover in this weeks design sync and follow up with anything that comes out of that:

1. Breadcrumbs and titling. I almost think that we should just to `Mesh > name > Gateways > name` for these instead of including the subtype. It then makes a nice rule for consistency "Kuma Nouns are used in breadcrumbs and titling, whereas 'subtypes' of those Kuma Nouns aren't". We can chat in design sync about this though.
2. Do we need to filter `MeshGateway` out of the policy listing view so it doesn't appear?
3. I noticed that the "naming of things" is wrong. This is in the `gateway` module so according to what we decided all these `*View.vue` files needs to begin with `Gateway`. Honestly I really don't care too much (as we decided not to do folders) so please don't go off and change just because of this, but I wanted to point out how unnatural I still think the thing we decided to stick to consistently is.

I'm good to follow up on 1 and 2 if we need to 👍

kleinfreund:
> 1. Breadcrumbs and titling. I almost think that we should just to `Mesh > name > Gateways > name` for these instead of including the subtype. It then makes a nice rule for consistency "Kuma Nouns are used in breadcrumbs and titling, whereas 'subtypes' of those Kuma Nouns aren't". We can chat in design sync about this though.

Let’s talk about this in the design sync, yes. I think we shouldn’t label them as the same. It could create the expectation that they’re interchangable and that isn’t the case for the two service types or the two gateway types.

> 2\. Do we need to filter `MeshGateway` out of the policy listing view so it doesn't appear?

Yes, we do. I had this in here initially, but removed it again. However, as long as we hide this change set behind a feature flag, we need to keep showing MeshGateways in policies. We could feature flag this up there, too, but I’m not sure that’s worth the effort. Do note that this will need the `/policies` API to be changed anyway (I’ll file a ticket for this).

Long story, short: when unflagging this feature, I’ll also do some simplifications in services (the details tabs view will no longer need to control data sourcing) and adjustments in policies (removing any handling of MeshGateways).

johncowen:
Gotcha, sounds good 👍

kleinfreund:
> Gotcha, sounds good 👍

I just talked with Charly about breadcrumbs. I think there are two options and we thought the first of them is better:

- Meshes > `:mesh` > Built-in Gateways > `:name`
- Meshes > `:mesh` > Gateways > Built-in > `:name`

The option of just having "Gateways" or "Services" in the breadcrumbs isn’t great in my opinion: you’d lose the option to go from a delegated gateway detail view to its list view (same for external services).

I think what we have currently gives you all you really need. The second option above would only allow you to go from delegated gateway detail view to built-in gateway list view (and from external service detail view to service detail view). I’m not sure that’s super useful.

johncowen:
> The option of just having "Gateways" or "Services" in the breadcrumbs isn’t great in my opinion: you’d lose the option to go from a delegated gateway detail view to its list view (same for external services).

This is a good point I think, but only if the word `Gateways` would take you back to the page where you were but _without the previous filter selected_ i.e. in my mind we would always show the word `Gateways` but if you are looking at a `built-in` clicking that `Gateway` word in the breadcrumbs would take you back to the `Gateways` tab, with the `built-in` filter selected.

So more briefly, my other option would be exactly the same as your first option here:

> - Meshes > :mesh > Built-in Gateways > :name

But just lose the word `Built-in`/`Delegated`. The link/href would still be for `built-in`/`delegated` depending, just as it is now.

kleinfreund:
> So more briefly, my other option would be exactly the same as your first option here:
> 
> > * Meshes > :mesh > Built-in Gateways > :name
> 
> But just lose the word `Built-in`/`Delegated`. The link/href would still be for `built-in`/`delegated` depending, just as it is now.

@lahabana What do you think about that? I personally don’t see a reason to drop the "Built-in"/"Delegated" prefix there. It makes clear where you’re going.

PR #2159, opened 2024-02-08
fix(policies): columns not always showing correctly

kleinfreund:
fix(policies): add regression test for columns not always showing correctly

fix(policies): columns not always showing correctly

Trigger a re-render of AppCollection when the current policy type changes to make sure the table re-renders with the correct columns.

refactor(policies): move route access into list view

### DIFF
--- a/features/mesh/policies/Index.feature
+++ b/features/mesh/policies/Index.feature
@@ -147,3 +147,12 @@ Feature: mesh / policies / index
     And the "[data-testid='policy-type-link-MeshFaultInjection']" element exists
     # Always shows MeshGateway
     And the "[data-testid='policy-type-link-MeshGateway']" element exists
+
+  Scenario: Regression test: Zone column is visible when navigating from legacy policy type
+    When I visit the "/meshes/default/policies/circuit-breakers" URL
+
+    Then the "$items-header" element exists 2 times
+
+    When I click the "[data-testid='policy-type-link-MeshFaultInjection']" element
+
+    Then the "$items-header" element exists 4 times
--- a/src/app/policies/components/PolicyList.vue
+++ b/src/app/policies/components/PolicyList.vue
@@ -17,7 +17,7 @@
           :to="{
             name: 'policy-list-view',
             params: {
-              mesh: route.params.mesh,
+              mesh: props.mesh,
               policyPath: policyType.path,
             },
           }"
@@ -189,7 +189,6 @@
 import { KUI_ICON_SIZE_30 } from '@kong/design-tokens'
 import { ArrowRightIcon } from '@kong/icons'
 import { computed } from 'vue'
-import { useRoute } from 'vue-router'
 
 import type { PolicyType } from '../data'
 import { PolicyCollection } from '../sources'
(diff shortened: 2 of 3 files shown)


features/mesh/policies/Index.feature
```diff
@@ -147,3 +147,12 @@ Feature: mesh / policies / index
     And the "[data-testid='policy-type-link-MeshFaultInjection']" element exists
     # Always shows MeshGateway
     And the "[data-testid='policy-type-link-MeshGateway']" element exists
+
+  Scenario: Regression test: Zone column is visible when navigating from legacy policy type
```
johncowen, line 151: Aren't all tests regression tests? i.e. bit of a nit, but do we need this prefix?

johncowen (commented):
I had 1 little nitty question, but I was wondering in which PR this regressed?

features/mesh/policies/Index.feature
kleinfreund: I think it’s useful to know when a test is dedicated to a regression.

kleinfreund:
> I had 1 little nitty question, but I was wondering in which PR this regressed?

https://github.com/kumahq/kuma-gui/pull/2119

johncowen:
Oh was it the removal of the key here then?

https://github.com/kumahq/kuma-gui/pull/2119/files#diff-15b4ba78b79c8245ad6c3be64f3cf17c428c9de4cb80d83cb54f1382786ebf14L55

kleinfreund:
> Oh was it the removal of the key here then?
> 
> https://github.com/kumahq/kuma-gui/pull/2119/files#diff-15b4ba78b79c8245ad6c3be64f3cf17c428c9de4cb80d83cb54f1382786ebf14L55

Since that is the only code change there, yes. I could’ve also re-introduced the `key` on PolicyList, but figured this was working as well.

johncowen:
How possible would it be to split this `PolicyList` component into two nested routes/`*View.vue`s?

- `/gui/meshes/default/policies` <-- this would be the "policy type list" that is currently on the left
  - `/gui/meshes/default/policies/:type` <-- this would be the "policy list" that is currently on the right

If I follow correctly what the issue is, this would mean that every time you click a "policy type" on the left, the "policy" table on the right would completely re-render just through building this with nested routes rather than using things like `:key`.

kleinfreund:
> How possible would it be to split this `PolicyList` component into two nested routes/`*View.vue`s?
> 
> * `/gui/meshes/default/policies` <-- this would be the "policy type list" that is currently on the left
>   
>   * `/gui/meshes/default/policies/:type` <-- this would be the "policy list" that is currently on the right
> 
> If I follow correctly what the issue is, this would mean that every time you click a "policy type" on the left, the "policy" table on the right would completely re-render just through building this with nested routes rather than using things like `:key`.

Sure, that’s possible. It’s also much, much, much more effort than just putting the `key` attribute here.

johncowen (approved):
👍 I made an issue to change this to use nested routes

https://github.com/kumahq/kuma-gui/issues/2162

PR #2181, opened 2024-02-13
chore: various amends for gateway data-viz pt2

johncowen:
Continuation of https://github.com/kumahq/kuma-gui/pull/2151

> All in all while we still don't show multiple inbounds if there are any, and the protocol for the outbounds could be wrong. I figured there was enough improvement here (no duplication, better states, no Cluster protocol etc) to PR something and continue with these issues.

So this is continuing mainly with multiple inbounds on gateways, but also showing non-zero values for the gateways outbounds.

---

Various fixes for the built-in gateway dataplane viz. I've tried to fix by continuing to normalize the way that we figure out stats/inbound/outbounds for both sidecars and gateways. I might add some inline comments 🤔 

- Supports multi inbounds to gateways in the dataplane viz
- Fixes showing any stats in the outbounds cards for gateway outbounds. It looks like the stats for gateway outbounds can only be found from `upstream` named stats, which I believe make sense as I 'think' its the direction of traffic from the gateway to each of the `outbounds`. But would be good to confirm with people. Either way, according to the stats traffic on anything `downstream` here is always zero.

~There's potentially a tiny bit more work I want to do here, so in draft while I decide. 🤔  I might just do it in a follow up seeing as its mostly cleanup/refactor seeing as this is user facing fixes I'm inclined to PR it as is and then follow up.~ I decided this see https://github.com/kumahq/kuma-gui/pull/2181#issuecomment-1943572869

### DIFF
--- a/src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
+++ b/src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
@@ -30,7 +30,8 @@
             <template
               v-for="(item, key) in [
                 (['http', 'tcp'] as const).reduce((prev, protocol) => {
-                  const direction = 'downstream'
+                  // FIXME: confirm this can change to upstream
+                  const direction = props.direction
                   // sum both the properties we need from both protocols
                   return Object.entries(props.traffic?.[protocol] || {}).reduce((prev, [key, value]) => {
                     return [`${direction}_cx_tx_bytes_total`, `${direction}_cx_rx_bytes_total`].includes(key) ?
@@ -67,7 +68,7 @@
             v-else-if="props.protocol === 'http'"
           >
             <div
-              v-for="value in [props.traffic.http?.downstream_rq_1xx as (number | undefined) ?? 0].filter(item => item !== 0)"
+              v-for="value in [props.traffic.http?.[`${props.direction}_rq_1xx`] as (number | undefined) ?? 0].filter(item => item !== 0)"
               :key="value"
             >
               <dt>{{ t('data-planes.components.service_traffic_card.1xx') }}</dt>
@@ -75,10 +76,10 @@
             </div>
             <div>
               <dt>{{ t('data-planes.components.service_traffic_card.2xx') }}</dt>
-              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_2xx as (number | undefined) ?? 0 }) }}</dd>
+              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.[`${props.direction}_rq_2xx`] as (number | undefined) ?? 0 }) }}</dd>
             </div>
             <div
-              v-for="value in [props.traffic.http?.downstream_rq_3xx as (number | undefined) ?? 0].filter(item => item !== 0)"
+              v-for="value in [props.traffic.http?.[`${props.direction}_rq_3xx`] as (number | undefined) ?? 0].filter(item => item !== 0)"
               :key="value"
             >
               <dt>{{ t('data-planes.components.service_traffic_card.3xx') }}</dt>
@@ -86,11 +87,11 @@
             </div>
             <div>
               <dt>{{ t('data-planes.components.service_traffic_card.4xx') }}</dt>
-              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_4xx as (number | undefined) ?? 0 }) }}</dd>
+              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.[`${props.direction}_rq_4xx`] as (number | undefined) ?? 0 }) }}</dd>
             </div>
             <div>
               <dt>{{ t('data-planes.components.service_traffic_card.5xx') }}</dt>
-              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.downstream_rq_5xx as (number | undefined) ?? 0 }) }}</dd>
+              <dd>{{ t('common.formats.integer', { value: props.traffic.http?.[`${props.direction}_rq_5xx`] as (number | undefined) ?? 0 }) }}</dd>
             </div>
           </template>
           <template
@@ -98,11 +99,11 @@
           >
             <div>
               <dt>{{ t('data-planes.components.service_traffic_card.tx') }}</dt>
-              <dd>{{ formatBytes(props.traffic.tcp?.downstream_cx_rx_bytes_total as (number | undefined) ?? 0) }}</dd>
+              <dd>{{ formatBytes(props.traffic.tcp?.[`${props.direction}_cx_rx_bytes_total`] as (number | undefined) ?? 0) }}</dd>
             </div>
             <div>
               <dt>{{ t('data-planes.components.service_traffic_card.rx') }}</dt>
-              <dd>{{ formatBytes(props.traffic.tcp?.downstream_cx_tx_bytes_total as (number | undefined) ?? 0) }}</dd>
+              <dd>{{ formatBytes(props.traffic.tcp?.[`${props.direction}_cx_tx_bytes_total`] as (number | undefined) ?? 0) }}</dd>
             </div>
           </template>
         </dl>
@@ -122,10 +123,14 @@ import type { TrafficEntry } from '../../data'
 import { useI18n } from '@/app/application'
 import DataCard from '@/app/common/data-card/DataCard.vue'
 const { t } = useI18n()
-const props = defineProps<{
+const props = withDefaults(defineProps<{
   protocol: string
   traffic?: TrafficEntry
-}>()
+  direction?: 'upstream' | 'downstream'
+}>(), {
+  traffic: undefined,
+  direction: 'downstream',
+})
 const click = (e: MouseEvent) => {
   const $target = e.target as HTMLElement
   if (e.isTrusted && $target.nodeName.toLowerCase() !== 'a') {
--- a/src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
+++ b/src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
@@ -30,7 +30,7 @@
             <template
               v-for="(item, key) in [
                 (['http', 'tcp'] as const).reduce((prev, protocol) => {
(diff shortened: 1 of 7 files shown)


src/app/data-planes/data/stats.ts
```diff
@@ -54,39 +99,57 @@ export const parse = (lines: string): EnvoyStats => {
...
-      traffic[key][item] = value
+      if (item === 'unknown') {
+        const obj = value as Record<string, any>
+        appProtocols.forEach((protocol) => {
+          if (typeof obj[protocol] !== 'undefined') {
+            traffic[key].protocol = protocol
+            traffic[key][protocol] = obj[protocol]
+          }
+        })
+      }
```
johncowen, line 152: This is the bit where we detect `grpc` or `http2` on `unknown` and hoist it up. Its the same approach as the red that was removed.

johncowen (commented):
Added some inline, partly for me, partly for review:

johncowen:
I've decided to take this out of draft as is:

1. It fixes the majority of user facing issues left with the gateway viz.
2. There is still an issue with not knowing the protocol of an outbound of a gateway proxy, but that was an issue anyway.
3. I'm dying to move all of this over to a `/connections` module and test out better and then clean up some things, and I can never find a good time to do this.

Taking it out of draft and approved means:

1. Users get the fixes
2. I can concentrate on figuring out exatcly how delegated gateways fit into this (it should be extremely similar to sidercars but I would guess there might be "details"
3. Once we have all three variations done, I'll understand all we have to do and therefore it will be a good time to test out and refactor/tighten up.

Just incase thats not clear ^ there are code-y things here I'd like to sort out, but they are code-y things, not user facing things. Taking this out of draft.

johncowen:
Just to note here:

> I can concentrate on figuring out exactly how delegated gateways fit into this (it should be extremely similar to sidecars but I would guess there might be "details"

I just tried this out with delegated and it looks like I can just remove the conditionals that were guarding us from showing anything for delegated gateways and everything works.

I'll PR that separately tho as I'm making a custom empty state for the fact that inbounds are empty for delegated gateways "because delegated"

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
```diff
@@ -30,7 +30,8 @@
             <template
               v-for="(item, key) in [
                 (['http', 'tcp'] as const).reduce((prev, protocol) => {
-                  const direction = 'downstream'
+                  // FIXME: confirm this can change to upstream
+                  const direction = props.direction
```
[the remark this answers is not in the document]
kleinfreund: I guess this would be best answered by @lahabana

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -221,51 +233,58 @@
...
-                        Non mesh traffic
-                      </ServiceTrafficCard>
-                    </ServiceTrafficGroup>
-                    <DataCollection
-                      v-slot="{ items }"
-                      :predicate="route.params.inactive ? undefined : (item) => ((item.protocol === 'tcp' ? item.tcp?.downstream_cx_rx_bytes_total : item.http?.downstream_rq_total) as (number | undefined) ?? 0) > 0"
-                      :items="traffic.outbounds"
+                    <!-- Outbounds for gateways report actual traffic on the upstream so we switch to upstream here for non-standard-->
+                    <template
+                      v-for="direction in [props.data.dataplane.networking.type !== 'standard' ? 'upstream' : 'downstream'] as const"
```
kleinfreund, line 238: **Suggestion**: I feel like this would be better solved by storing a new property on `networking` in the data layer.

src/app/data-planes/views/DataPlaneInboundSummaryStatsView.vue
```diff
@@ -30,15 +30,41 @@
...
+                  .join('\n')
+
+              } else {
+                const name = `localhost_${route.params.service.split(':')[1]}`
+                return data.raw.split('\n')
+                  .filter((item) => item.includes(`.${name}.`))
+                  .map((item) => item.replace(`${name}.`, ''))
+                  .join('\n')
+              }
+            })()"
```
kleinfreund, line 67: **Suggestion**: Can we introduce a function for this somewhere? There is a ton of stuff going on here that shouldn’t be in the template here.

johncowen: Yes definitely, was gonna do something in a follow up, let me make this a little better here first.

src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: Just so I'm clear, do you mean the `direction` we should use to figure out the traffic?

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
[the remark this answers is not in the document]
lahabana: Reading: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/intro/terminology#terminology

> Downstream: A downstream host connects to Envoy, sends requests, and receives responses.
> Upstream: An upstream host receives connections and requests from Envoy and returns responses.

Then:
- when traffic is HTTP we have HTTPConnection manager https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/stats
- when traffic is TCP we have https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/tcp_proxy_filter#statistics
- when there's a listener (inbounds and outbounds when not using builtin gateway) https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/stats

It's not mega clear to me where you are getting the stats from here but there's some info in these docs that should help you get a clearer idea of what we're doing here. In any case we can chat some more!

src/app/data-planes/views/DataPlaneDetailView.vue
kleinfreund: Yes.

src/app/data-planes/components/data-plane-traffic/ServiceTrafficCard.vue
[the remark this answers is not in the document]
johncowen: note: We are gonna change the direction for outbounds for both sidecars and gateways

src/app/data-planes/data/stats.ts
```diff
@@ -33,8 +42,44 @@ export const parse = (lines: string): EnvoyStats => {
...
+
+      // 10.244.0.11_8081.http.edge-gateway.downstream_rq_1xx
+      parts = _key.substring(pos + 1).split('.')
+
+      // 8081.http.edge-gateway.downstream_rq_1xx
+      const [port, protocol, service, name] = parts
+      parts = ['listener', protocol, service.search(trailingPortRe) !== -1 ? service : `${service}_${port}`, name]
+      // listener.http.edge-gateway_8081.downstream_rq_1xx
+      _key = parts.join('.')
+    }
```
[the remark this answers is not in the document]
johncowen: We are eventually gonna go back to not using `listener.` at all, which I'm glad about 😅 , but we need some changes made so its possible to get the non-rolled up info from the `/stats` endpoint. I won't be spending too much time tweaking this code whatever as it will go pretty soon.

src/app/data-planes/data/stats.ts
```diff
@@ -54,39 +99,57 @@ export const parse = (lines: string): EnvoyStats => {
...
+          // so we can pretend this doesn't happen
           port: key.split('_').at(-1) ?? '',
-          // TODO(jc): Not totally sure how we are going to get the protocol
-          // from the outbounds for gateways (these only have `cluster.`) so
-          // lets set it to `tcp` for now so as to now show `Cluster` as the
-          // protocol
-          protocol: item === 'cluster' ? 'tcp' : item,
+          // FIXME we need to detect the protocol based on the existence of
+          // a http property in the stats, just not sure which one yet
+          protocol: item === 'unknown' ? 'http' : item,
```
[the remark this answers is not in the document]
johncowen: For outbounds on built-in gateways, we are going to look for non-zero values in the stats to figure out the protocol (i.e. if something `http` is non-zero then its `http`). For the times when everything is `0` we are going to show the card in some sort of unknown protocol state.

johncowen:
Just a small note, I've added a few additional inlines following a sync on this. I need to pick this back up again and decide to merge this PR and follow up with those additional amends in a separate PR or do here. I'm likely to followup considering this PR on its own is an improvement (multiple http inbounds on gateway and stats for http outbounds on gateway)

I also want to make a few slight refactors mentioned in the inline conversation above before looking for approval/merge of this PR anyway. So will be back with an update here soonish.

src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: This will depend on whether its an inbound or outbound, which are different things so I don't think there will be a need for its own property. But I can re look at this in a follow up.

src/app/data-planes/views/DataPlaneInboundSummaryStatsView.vue
johncowen: I've moved this into a function for the minute, it will probably go into a `/connections/data/index.ts` file later on.

johncowen:
Ok I've:

1. Removed the `kafka` reference.
2. Converted FIXMEs to TODOs (for the follow up)
3. 'Sunk' the inline template function down into the `script` tag

Apart from that I want to try and get this approved and in. As explained in this and the last PR, there are good user improvements in this PR (multiple HTTP inbounds for gateways, stats for HTTP outbounds for gateways).

The things that this still doesn't do (TCP inbounds for gateways, protocol sniffing for outbounds for gateways) didn't work previous to this PR so those facts should be non-blocking here. And the good news is, at the very least we have a plan for resolving these outstanding things. But I'd rather do that in a follow up.

So to TLDR; if theres something that is user broken here, then it should block if its not mentioned above. If its a code-y thing, feel free to ask/suggest/hide-behind-the-sofa, but those things are likely to be resolved in the follow ups. That being said if you feel super strongly about something like that, please feel free to mention.

PR #2156, opened 2024-02-07
refactor: update Kongponents to KTooltip milestone

kleinfreund:
**chore(KTooltip): address breaking changes**

Replace `label` with `text` prop.

Removes `props.positionFixed` as its default value is now `true`.

Signed-off-by: Philipp Rudloff <philipp.rudloff@konghq.com>

---

Previous Kongponents v9 adoption PRs:

- https://github.com/kumahq/kuma-gui/pull/1532
- https://github.com/kumahq/kuma-gui/pull/1853
- https://github.com/kumahq/kuma-gui/pull/1878
- https://github.com/kumahq/kuma-gui/pull/1961
- https://github.com/kumahq/kuma-gui/pull/2118

PR #2118, opened 2024-01-31
refactor: update Kongponents to KModal/KPrompt milestone

kleinfreund:
**chore(KModal): address breaking changes**

Rename `props.isVisible` to `props.visible`.

Rename `canceled` event to `cancel`.

Remove `header-content` slot use as it’s not needed anymore. The `props.title` string will be used automatically.

Change `body-content` slot to default slot.

Change `action-buttons` slot to `footer-actions` slot.

**chore(KPrompt): address breaking changes**

Rename `props.isVisible` to `props.visible`.

Rename `canceled` event to `cancel`.

Change `body-content` slot to default slot.

**chore(zones): remove divider from single item dropdown item**

Signed-off-by: Philipp Rudloff <philipp.rudloff@konghq.com>

---

Previous Kongponents v9 adoption PRs:

- https://github.com/kumahq/kuma-gui/pull/1532
- https://github.com/kumahq/kuma-gui/pull/1853
- https://github.com/kumahq/kuma-gui/pull/1878
- https://github.com/kumahq/kuma-gui/pull/1961

PR #2184, opened 2024-02-14
refactor: update Kongponents to KSlideout/KTruncate milestone

kleinfreund:
This milestone adds a `title` slot to `KSlideout` which we should be making use of. This caused a bit more refactoring work of how we currently handle summary trays.

---

**chore(KTruncate): address breaking changes**

Rename `.k-truncate-container` to `.truncate-container`.

**chore(KSlideout): address breaking changes**

Replace `isVisible` with `visible` prop.

Replace `preventCloseOnBlur` with `:closeOnBlur="false"`.

Update custom styles to address changes in markup and layout.

**refactor(SummaryView): use KSlideout title slot**

Delete unused summary code in the mesh and zone list views.

Move the use of SummaryView from the parent views into the concrete *SummaryView components. The primary reason for this change is that this allows us to use the `title` slot of KSlideout. Not using it isn’t great because it would make for inconsistent use of KSlideout across applications, but it’s also proving more challenging to provide custom styles that re-implement the same mechanics and looks of the slideout title.

---

### DIFF
--- a/features/application/MainNavigation.feature
+++ b/features/application/MainNavigation.feature
@@ -95,4 +95,4 @@ Feature: application / MainNavigation
 
   Scenario: Tertiary navigation
     When I visit the "/meshes/default/data-planes/dp-name/overview/inbound/driver:8080/stats" URL
-    And the "#data-plane-inbound-summary-stats-view-tab.active" element exists
+    And the "#connection-inbound-summary-stats-view-tab.active" element exists
--- a/features/mesh/dataplanes/DataplaneSummary.feature
+++ b/features/mesh/dataplanes/DataplaneSummary.feature
@@ -4,7 +4,7 @@ Feature: Dataplane summary
       | Alias                | Selector                                       |
       | item                 | [data-testid='data-plane-collection'] tbody tr |
       | summary              | [data-testid='summary']                        |
-      | close-summary-button | $summary [data-testid^='close-button-']        |
+      | close-summary-button | $summary [data-testid='slideout-close-icon']   |
     And the environment
       """
       KUMA_SUBSCRIPTION_COUNT: 2
--- a/features/mesh/policies/PolicySummary.feature
+++ b/features/mesh/policies/PolicySummary.feature
@@ -1,10 +1,10 @@
 Feature: Policy summary
   Background:
     Given the CSS selectors
-      | Alias                | Selector                                   |
-      | item                 | [data-testid='policy-collection'] tbody tr |
-      | summary              | [data-testid='summary']                    |
-      | close-summary-button | $summary [data-testid^='close-button-']    |
+      | Alias                | Selector                                     |
+      | item                 | [data-testid='policy-collection'] tbody tr   |
+      | summary              | [data-testid='summary']                      |
+      | close-summary-button | $summary [data-testid='slideout-close-icon'] |
     And the URL "/meshes/default/meshfaultinjections" responds with
       """
       body:
--- a/features/zones/zone-cps/egresses/ZoneEgressSummary.feature
+++ b/features/zones/zone-cps/egresses/ZoneEgressSummary.feature
@@ -4,7 +4,7 @@ Feature: Zone Egress summary
       | Alias                | Selector                                        |
       | item                 | [data-testid='zone-egress-collection'] tbody tr |
       | summary              | [data-testid='summary']                         |
-      | close-summary-button | $summary [data-testid^='close-button-']         |
+      | close-summary-button | $summary [data-testid='slideout-close-icon']    |
     And the URL "/zoneegresses/_overview" responds with
       """
       body:
--- a/features/zones/zone-cps/ingresses/ZoneIngressSummary.feature
+++ b/features/zones/zone-cps/ingresses/ZoneIngressSummary.feature
@@ -4,7 +4,7 @@ Feature: Zone Ingress summary
       | Alias                | Selector                                         |
       | item                 | [data-testid='zone-ingress-collection'] tbody tr |
       | summary              | [data-testid='summary']                          |
-      | close-summary-button | $summary [data-testid^='close-button-']          |
+      | close-summary-button | $summary [data-testid='slideout-close-icon']     |
     And the URL "/zone-ingresses/_overview" responds with
       """
       body:
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "dependencies": {
     "@kong-ui-public/i18n": "2.1.3",
     "@kong/icons": "1.8.14",
-    "@kong/kongponents": "9.0.0-alpha.105",
+    "@kong/kongponents": "9.0.0-alpha.107",
     "@vueuse/core": "10.8.0",
     "brandi": "5.0.0",
     "deepmerge": "4.3.1",
--- a/src/app/common/SummaryView.vue
+++ b/src/app/common/SummaryView.vue
@@ -2,10 +2,9 @@
   <KSlideout
     ref="slideOutRef"
     class="summary-slideout"
-    prevent-close-on-blur
-    close-button-alignment="end"
+    :close-on-blur="false"
     :has-overlay="false"
-    is-visible
+    visible
     :max-width="props.width"
     offset-top="var(--app-slideout-offset-top, 0)"
     data-testid="summary"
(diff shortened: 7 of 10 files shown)


johncowen:
How come we can't just do this:

https://github.com/kumahq/kuma-gui/compare/master...johncowen:kuma-gui:scratch/k-slideout

Seems like the above resolves the breaking changes in the kongponents alpha, and is far less change in order to be able to upgrade?

kleinfreund:
> Seems like the above resolves the breaking changes in the kongponents alpha, and is far less change in order to be able to upgrade?

Because it completely sidesteps the slideout styles which comes more notable with this change in KSlideout. Consider the following screenshots, both having their content portion scrolled down a bit:

You branch:

image

This branch:

image

KSlideout now fixes its title to the top of the viewport like this.

johncowen:
Hmm, thats not great. I can no longer scroll my listing with the KSlideout open. This is probably important for the primary usecase of our drawers - being able to quickly click between several rows and see the drawer without having to click open, close, open, close etc etc all the time. 🤔

kleinfreund:
> Hmm, thats not great. I can no longer scroll my listing with the KSlideout open. This is probably important for the primary usecase of our drawers - being able to quickly click between several rows and see the drawer without having to click open, close, open, close etc etc all the time. 🤔

This should be brought up on https://github.com/Kong/kongponents/pull/2020.

johncowen:
Yep defo, already commented

johncowen:
> Because it completely sidesteps the slideout styles which comes more notable with this change in KSlideout. Consider the following screenshots, both having their content portion scrolled down a bit:

So I think this was my point during our design sync. We should be addressing the breaking changes only here if we can, not completely re-implementing how our application is structured.

I added a `background-color: $kui-color-background;` to my scratch branch and even added a small border radius incase:

https://github.com/kumahq/kuma-gui/compare/master...johncowen:kuma-gui:scratch/k-slideout

Screenshot 2024-02-20 at 17 21 00

This looks fine to me as a "lets make sure we address the breaking changes as simply as possible", but I'm also glad we got to find the scroll bar issues 🙌 

I would like to change slightly how we do our SummaryView titling, but:

1. I don't think it requires so much change (i.e. I think we can do it in a few extra lines whilst still making use of the summary views new title slot)
2. I don't think its super urgent.
3. I don't want to do it in one of these "Kongponents upgrades" PRs, especially if it requires so much change to our application. If it requires this much change and we can address the breakages with a few lines. Then lets address the breakages first and circle back.

Let me know what you think.

kleinfreund:
Just why wouldn’t we use the KSlideout’s title slot?

The proposed minimal additions you suggest to make this change smaller seem still far off what the titles are supposed to look like.

> 3\. I don't want to do it in one of these "Kongponents upgrades" PRs, especially if it requires so much change to our application. If it requires this much change and we can address the breakages with a few lines, then lets address the breakages first and circle back.

What’s the effective difference between doing it here and doing it separately? It seems very much like the ideal opportunity to do this.

kleinfreund:
without the SummaryView refactoring work, I can get roughly to an okay state with custom CSS that looks like this (scrolled tray):

image

This has issues: when the title wraps, the fixed content is taller and we can’t account for it dynamically with these styles.

While that’s technically a less invasive change, it also requires quite some heavy-handed CSS customization that I’d rather avoid.

I’ve pushed this approach up here for reference, but I don’t think I want to ship that kind of Kongponents mutilation.

johncowen:
Please can we just address the breaking changes in the kongponents upgrade in this PR?

We can move to use the title slot in a follow up. FYI: The change I would like to make:

1. Uses the KSlideout title slot
2. Doesn't have any CSS overwrites.
3. Has far less change than in the previous iteration of this PR or this one.

johncowen:
> This has issues: when the title wraps, the fixed content is taller and we can’t account for it dynamically with these styles.

> While that’s technically a less invasive change, it also requires quite some heavy-handed CSS customization that I’d rather avoid.

> I’ve pushed this approach up here for reference, but I don’t think I want to ship that kind of Kongponents mutilation.

Just so I follow what you are saying, would you be happy merging this iteration? It sounded like you weren't and it was just for reference, but it changes far less that the previous iteration and the changes are more focussed and easier to remove when we move to use KSlideout's title slot, so if you absolute can't go with the branch I proposed, then I'd be happy merging this in the interim seeing as it addresses the breaking change without to much widespread change.

Lmk

kleinfreund:
> Please can we just address only the breaking changes in the kongponents upgrade in this PR?
> 
> We can move to use the title slot in a follow up. FYI: The change I would like to make:
> 
> 1. Uses the KSlideout title slot
> 2. Doesn't have any CSS overwrites.
> 3. Has far less change than in the previous iteration of this PR and less than this iteration.

I think I laid out quite clearly how that isn’t possible. The way we put titles in our summary tries is completely and fully at odds with how the title slot is styled. Something I addressed in the previous iteration of this PR which you rejected based on reasons that aren’t comprehensible. We can’t have "uses title slot + doesn’t have CSS overwrites + doesn’t change existing summary views".



> Just so I follow what you are saying, would you be happy merging this iteration? It sounded like you weren't and it was just for reference? It changes far less that the previous iteration and the changes are more focussed and easier to remove when we move to use KSlideout's title slot. So, if you absolutely can't go with the branch I proposed with the tiny CSS overwrite, then I'd be happy merging this in the interim seeing as it addresses the breaking change without too much widespread change.
> 
> Lmk

First of all, what you proposed in your branch isn’t enough because it doesn’t fix the title content to the top of a summary try, it only does so for the close button. I don’t want to have that much deviation from the the KSlideout title styles.

What I have currently in this PR addresses this, but it also has a flaw: offsetting the summary tray content from the top needs to account for varying heights of the title content but we don’t know the height of it so we can only use fixed values (one of the flaws of not using KSlideout’s title mechanism). I accounted for wrapping content which leaves non-wrapping content with a slightly too big gap.

All that being said, if it’s that important to you that this PR doesn’t contain the summary view refactor I had in the previous iteration of it and we do that in a follow up, then I’m okay with merging this current version. In that case, I want to do this as an _immediate_ follow up.

kleinfreund:
> I'd be happy to put a PR in straight after that moves to use KSlideout's title slot with no CSS overwrites or major application re-writes, lmk once this has gone in.

I’m confused. We have to rewrite some of the application to move the titles. We’ll still have to do that.

johncowen:
As I said, lets address the breaking changes first. That's the most important thing here.

After that, we can look at using KSlideout's title slot in a separate PR.

Whether we use it or not makes no difference to our GUI or users of our GUI. The panel has no sticky title as it stands now, and its not as if people are blocked because the title in the side panel isn't sticky.

Longer term it would be nice for kongponents folks if we could use it, but even then they've said they're expecting some consumers may want to CSS overwrite things in the title slot (the PR I have in mind has no CSS overwrites)

kleinfreund:
@johncowen Please review this again. The adoption of this milestone is being picked up in the next days.

src/app/application/components/app-view/AppView.vue
```diff
@@ -21,6 +21,7 @@
         <slot name="title" />
 
         <div
+          v-if="$slots.actions"
```
johncowen, line 24: How come we needed this?

kleinfreund: I noticed that this renders an empty element even when there are no actions slotted-in. This messes with the parent's flex layout. Generally, optional slots should be guarded like this.

johncowen: From that I understood that it wasn't needed for the upgrade?

I know this sounds like a nit, but it would be really good to get used to doing these types of PRs (upgrades, important bugfixes) as only whats needed to do the upgrade or important bugfix. That way its more likely we avoid regressions or further errors from doing an upgrade or important bugfix.

Sorry, I know its this tiny bit of code, but could this one go in another PR if poss? Also, is this currently causing an issue anywhere (no reason not to fix if not, just curious).

I get the feeling this PR isn't going in that soon as the kongponents side is still in draft, but if not I'm going to approve anyway incase you need to merge quick without removing this.

johncowen (approved):
Seems fine for the minute

src/app/application/components/app-view/AppView.vue
johncowen: I think this _needs_ to be removed, 99% sure. So even for the 1% that I'm wrong it doesn't hurt to remove and really it shouldn't be in this PR. Let me dismiss the approval, sorry about that

johncowen: This isn't addressing the breaking changes, and I think adding this change here creates a bug elsewhere. Could you you remove it from this PR please? We can circle back after if necessary.

kleinfreund: If this creates a bug somewhere then surely that needs fixing, right? Can you tell me what this could break and file an issue for it?

A hand-wavy suggestion to remove the appropriate guard on an optional slot doesn’t seem to break the idea of having these adoption PRs be fairly minimal.

kleinfreund: FYI I did remove this now, but I expect a better explanation for why this insistence was necessary.

johncowen: This insistence was necessary because I've been saying throughout this PR that we should just be addressing the breaking changes in the kongponents upgrade here, and the before and after screengrabs you gave above showing the line wrapping if the title is exactly a certain length with no breaking characters is not something I would cout as just addressing the breaking changes.

---

Apart from this, FYI:

https://github.com/kumahq/kuma-gui/pull/1729

Screenshot 2024-02-23 at 10 26 24

johncowen (approved):

PR #2186, opened 2024-02-15
refactor: simplify NavTabs component

kleinfreund:
Integrate the logic for determining the routes that are to be displayed as navigational tabs into the NavTabs component.

### DIFF
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -23,10 +23,8 @@
         replace: routeReplace,
         params: routeParams,
         back: routerBack,
-        children: children,
-        active: (active: RouteLocationNormalizedLoaded) => {
-          return children.find(item => item.name === active.name || (item.meta && item.meta.module === active.meta.module))
-        },
+        children,
+        active,
       }"
     />
   </div>
@@ -46,7 +44,7 @@ import {
   beforePaint,
 } from '../../utilities'
 import { useEnv } from '@/utilities'
-import type { RouteLocationNormalizedLoaded, RouteRecordRaw } from 'vue-router'
+import type { RouteRecordRaw } from 'vue-router'
 export type RouteView = {
   addTitle: (item: string, sym: Symbol) => void
   removeTitle: (sym: Symbol) => void
@@ -54,7 +52,7 @@ export type RouteView = {
   removeAttrs: (sym: Symbol) => void
 }
 
-export type StringNamedRouteRecordRaw = RouteRecordRaw & {
+type StringNamedRouteRecordRaw = RouteRecordRaw & {
   name: string
 }
 
@@ -102,6 +100,7 @@ const children: StringNamedRouteRecordRaw[] = (router.getRoutes().find((route) =
   item.name = String(item.name)
   return item as StringNamedRouteRecordRaw
 }) ?? [])
+const active = computed(() => children.find((item) => item.name === route.name || item.meta?.module === route.meta.module))
 
 const routeView = {
   addTitle: (item: string, sym: Symbol) => {
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -54,6 +54,10 @@ export type RouteView = {
   removeAttrs: (sym: Symbol) => void
 }
 
+export type StringNamedRouteRecordRaw = RouteRecordRaw & {
+  name: string
+}
+
 const win = window
 const env = useEnv()
 const can = useCan()
@@ -64,9 +68,6 @@ const sym = Symbol('route-view')
 
 type Params = { [P in keyof T]: T[P] }
 type RouteReplaceParams = Parameters<typeof router['push']>
-type StringNamedRouteRecordRaw = RouteRecordRaw & {
-  name: string
-}
 
 const props = withDefaults(defineProps<{
   name: string
--- a/src/app/common/NavTabs.vue
+++ b/src/app/common/NavTabs.vue
@@ -66,14 +66,3 @@ const currentTabHash = computed(() => {
   return activeTab.hash
 })
 </script>
-
-<style lang="scss" scoped>
-.nav-tabs {
-  overflow-x: auto;
-  width: 100%;
-}
-
-.nav-tabs :deep(.tab-item) {
-  white-space: nowrap;
-}
-</style>
--- a/src/app/common/NavTabs.vue
+++ b/src/app/common/NavTabs.vue
@@ -7,7 +7,7 @@
     data-testid="nav-tabs"
   >
     <template
-      v-for="tab in props.tabs"
+      v-for="tab in tabs"
       :key="`${tab.routeName}-anchor`"
       #[`${tab.routeName}-anchor`]
     >
@@ -22,36 +22,59 @@
 </template>
 
 <script lang="ts" setup>
-import { KTabs, Tab } from '@kong/kongponents'
-import { PropType, computed } from 'vue'
-import { useRoute } from 'vue-router'
+import { computed } from 'vue'
+import { useRoute, useRouter, type RouteRecordRaw } from 'vue-router'
 
-export interface NavTab {
+import { useI18n } from '@/utilities'
+import type { Tab } from '@kong/kongponents'
+
+interface NavTab {
   title: string
   routeName: string
   module: string
 }
 
+const { t } = useI18n()
 const route = useRoute()
+const router = useRouter()
 
-const props = defineProps({
-  tabs: {
-    type: Array as PropType<NavTab[]>,
-    required: true,
-  },
+const props = withDefaults(defineProps<{
+  anchorRouteName: string
+  i18nPrefix: string
+  filterPredicate?: (route: RouteRecordRaw) => boolean
+}>(), {
+  filterPredicate: () => true,
 })
 
-const kTabs = computed<Tab[]>(() => props.tabs.map((tab) => ({
-  title: tab.title,
-  hash: '#' + tab.routeName,
-})))
+const tabs = computed<NavTab[]>(() => {
+  const routes = router.getRoutes().find((route) => route.name === props.anchorRouteName)?.children ?? []
+
+  return routes
+    .filter(props.filterPredicate)
+    .map((route) => {
+      const referenceRoute = typeof route.name === 'undefined' ? route.children?.[0] as RouteRecordRaw : route
+      const routeName = referenceRoute.name as string
+      const module = referenceRoute.meta?.module ?? ''
+      const title = t(`${props.i18nPrefix}.${routeName}`)
+
+      return { title, routeName, module }
+    })
+})
+
+const kTabs = computed<Tab[]>(() => {
+  return tabs.value.map((tab) => ({
+    title: tab.title,
+    hash: '#' + tab.routeName,
+  }))
+})
+
 const currentTabHash = computed(() => {
   const modules = route.matched
     .map((route) => route.meta.module ?? '')
     .filter((module) => module !== '')
   modules.reverse()
 
-  const activeTab = props.tabs.find((tab) => {
+  const activeTab = tabs.value.find((tab) => {
     if (tab.routeName === route.name) {
       return true
     }
@@ -62,7 +85,7 @@ const currentTabHash = computed(() => {
 
     return false
   })
-  const routeName = activeTab?.routeName ?? props.tabs[0].routeName
+  const routeName = activeTab?.routeName ?? tabs.value[0].routeName
 
   return '#' + routeName
 })
--- a/src/app/common/NavTabs.vue
+++ b/src/app/common/NavTabs.vue
@@ -1,68 +1,32 @@
 <template>
   <KTabs
     :tabs="tabs"
-    :model-value="currentTabHash"
+    :model-value="props.activeRouteName ? '#' + props.activeRouteName : tabs[0].hash"
     hide-panels
-    class="nav-tabs"
-    data-testid="nav-tabs"
   >
     <template
       v-for="tab in tabs"
-      :key="`${tab.routeName}-anchor`"
-      #[`${tab.routeName}-anchor`]
+      :key="tab.title"
+      #[`${tab.title}-anchor`]
     >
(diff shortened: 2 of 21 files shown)


src/app/services/views/ServiceDetailTabsView.vue
```diff
@@ -52,8 +52,19 @@
...
+              if (data.serviceType !== 'external' && route.name === 'service-config-view') {
+                return false
+              }
+
+              if (data.serviceType === 'external' && route.name === 'service-data-plane-proxies-view') {
+                return false
+              }
+
+              return true
+            }"
```
kleinfreund, line 67: **Comment**: This won’t be necessary once we remove the KUMA_GATEWAYS_UI feature flag and can be removed again.

johncowen (commented):
I really like and appreciate that you wanted to simplify the JS/TS used/repeated in out tab views here, and I like the overall solution.

But! Putting this functionality inside NavTabs means it only works for NavTabs and would need to be repeated for any other differently designs UI interfaces we have that need the same functionality.

So, I've tried to build ontop of what you have done here with https://github.com/kumahq/kuma-gui/pull/2192, could you have a look at that and see if we can bring both these 2 PRs together?

src/app/application/components/route-view/RouteView.vue
```diff
@@ -54,6 +54,10 @@ export type RouteView = {
   removeAttrs: (sym: Symbol) => void
 }
 
+export type StringNamedRouteRecordRaw = RouteRecordRaw & {
+  name: string
+}
```
kleinfreund, line 59: **Comment**: Somehow, I couldn’t import this without moving it up here.

src/app/meshes/views/MeshDetailTabsView.vue
```diff
@@ -18,8 +18,11 @@
       </template>
 
       <NavTabs
-        class="route-mesh-view-tabs"
-        :tabs="tabs"
+        :children="route.children"
+        :active="route.active"
+        i18n-prefix="meshes.routes.item.navigation"
+        :filter-predicate="(route) => route.name !== 'external-service-list-view'"
```
johncowen, line 24: If you reuse `DataCollection` here, you don't need to reimplement the same functionality inside `NavTabs` (the presentational component).

src/app/common/NavTabs.vue
```diff
@@ -22,49 +22,48 @@
...
+      const hash = '#' + routeName
+      const title = t(`${props.i18nPrefix}.${routeName}`)
+      const isActive = props.active(currentRoute)?.name === routeName
 
-    if (modules.includes(tab.module)) {
-      return true
-    }
+      return { title, hash, routeName, isActive }
+    })
+})
```
johncowen, line 61: If we pretend that we are using `DataCollection` to do filtering outside this component already (i.e. meaning the `filter` isn't required), is this computed function just reshaping data to be fed to `KTabs`, it kinda looks that way to me?

Looks like we could do?

```vue
<KTabs
  :tabs="items.map(item => ({
      title: t(`path.${item.name}`),
      hash: `#${item.name}`,
    }))"
  :model-value="`#${route.active(active)?.name ?? ''}`"
  hide-panels
  data-testid="nav-tabs"
>
  <template
    v-for="item in items"
    :key="`${item.name}`"
    #[`${item.name}-anchor`]
  >
    <RouterLink
      :data-testid="`${item.name}-tab`"
      :to="{ name: item.name }"
    >
      {{ t(`path.${item.name}`) }}
    </RouterLink>
  </template>
</KTabs>
```

johncowen (commented):
I added some inline suggestions, it looks to me like we don't need a the `NavTabs` wrapper at all? I think even the additional styles we've added into `NavTabs` are no longer required?

kleinfreund:
> I added some inline suggestions, it looks to me like we don't need a the `NavTabs` wrapper at all? I think even the additional styles we've added into `NavTabs` are no longer required?

What makes you say that? To avoid NavTabs, we would have to duplicate all the code inside NavTabs for all its usages. That’s awful.

kleinfreund:
> is there anything extra that having `NavTabs` brings us?

Just the usage example you provided illustrates how much worse dropping this abstraction would be. I’m not going to do that.

> and if I'm not mistaken these CSS rules have now been added to `KTabs`?

That much is correct, but besides the other simplifications I already did here, it’s the only thing we can drop.

johncowen:
What is the "abstraction" abstracting and why is it necessary? i.e. what does `NavTab`s bring us?

kleinfreund:
> What is the "abstraction" abstracting and why is it necessary? i.e. what does `NavTab`s bring us?

Not having to write all the stuff you’d need to write if it didn’t exist. Using it like this:

```vue
<NavTabs
  :children="route.children"
  :active="route.active"
  i18n-prefix="data-planes.routes.item.navigation"
/>
```

is much nicer than

```vue
<KTabs
  :tabs="items.map(item => ({
      title: t(`path.${item.name}`),
      hash: `#${item.name}`,
    }))"
  :model-value="`#${route.active(active)?.name ?? ''}`"
  hide-panels
>
  <template
    v-for="item in items"
    :key="`${item.name}`"
    #[`${item.name}-anchor`]
  >
    <RouterLink
      :to="{ name: item.name }"
    >
      {{ t(`path.${item.name}`) }}
    </RouterLink>
  </template>
</KTabs>
```

So, it saves us a lot of repetition.

src/app/common/NavTabs.vue
kleinfreund: > What do you mean by "virtual code"?

I meant "inlining all virtually code"

> Why would we want to use `computed` here?

It cashes its results and unchanging dependencies won’t result in a recompuation and therefor prevent unnecessary rerenderings.

johncowen: Do you think we should cache the result of a `map` across a list of 5 or 6 static items?

johncowen:
But its not like its something we can't just copy/pasta right, and as I pointed out before this gives us access to the RouterLinks from the `*View.vue` files (where we have access to the router). NavTabs locks those away so we can't get to them when we need to.

How come we don't need similar "abstractions" in places like this:

https://github.com/kumahq/kuma-gui/blob/3b2ffb1f2e6f86757b067aeefc4138f5d7b7052c/src/app/gateways/views/GatewayListTabsView.vue#L31-L50

How come we only need it for KTabs?

src/app/meshes/views/MeshDetailTabsView.vue
kleinfreund: > What if we don't have tabs depending on what application we are in?

I’m not aware of a case where this occurs currently. Currently, the degree of flexibility we need is deciding what has tabs or not.

> haven't locked this functionality up inside of NavTabs.

Since this isn’t currently something that’s conditional, I don’t see how adding DataCollection is any better than simply removing NavTabs where we don’t need them.

johncowen:
Also, looking at that, what happens if we need to add a `param` to one of the RouterLinks in the tabs as we needed to do with `LinkBox`

kleinfreund:
> But its not like its something we can't just copy/pasta right, and as I pointed out before this gives us access to the RouterLinks from the `*View.vue` files (where we have access to the router). NavTabs locks those away so we can't get to them when we need to.

But what’s the point? We don’t need any access to those links from within a view component. NavTabs has proven very useful at what it does and this PR cuts down on unnecessary repetition by a lot.

> How come we don't need similar "abstractions" in places like this:

Because it’s two links without anything special other than adding an `active` class and because we don’t use a Kongponent here that we need to interface with. Were we using KSegmentedControl (provided it could handle links), it’s quite likely that an abstraction would be beneficial for that as well.

kleinfreund:
> Also, looking at that, what happens if we need to add a `param` to one of the RouterLinks in the tabs as we needed to do with `LinkBox`

Then we’ll consider that as it happens.

johncowen:
I just checked the PR that brought NavTabs in https://github.com/kumahq/kuma-gui/pull/962#issuecomment-1571824253:

> What part of this PR will survive once KTabs is updated to support usages? Feels like we'll be left with a component that brings nothing and we'd likely just undo what is done here no?

KTabs has been updated, and there is no functional value in `NavTabs` that I can see, plus it sounds like we planned to just undo what was done there as we are left with a component that brings nothing?

If we've chosen to 'simplify' NavTabs at this point, then we should just simplify it by just undoing it now, as mentioned in the above PR, so we don't get a slow stream of PRs going up over time that are further tweaking and changing a component that we don't actually need 🤷

johncowen:
Something I've bumped into today, I've been working on a PR to move all of the DataPlane Viz stuff into it own module and hit a thing where the new NavTab abstraction would be problematic:

Both inbounds and outbounds summary views use the same 3 tab names `Overview`, `Stats` and `Clusters` I would guess they will always be the same whether you are in inbounds or outbounds, so I wanted them to use the same locales text.

Doing this is impossible with this new NavTabs abstraction because its too coupled to the route, I _have_ to have these texts repeated, I _have_ to have one for each route and it must be the exact name of the route.

See https://github.com/kumahq/kuma-gui/compare/master...johncowen:kuma-gui:chore/connections-module#diff-9de42debc26bc42cb9c169537812c1585f92d61d4b4d437d47863a7cf240fbf1R23-R39 for the alternative with the much more decoupled `KTabs`.

This is very similar to the point above somewhere where I said this new abstraction makes it difficult to add params to a route used for a tab, but also the other thing I realized with this new NavTabs implementation is that we are using `t` inside reusable components, something we said we would avoid doing:

> When writing components we should "default to" avoiding using t within generic/re-usable components. i.e. we should pass translated text into the components from the outside using slots as much as possible (and occasionally attributes). Mostly this means t would most commonly be used in route/view components.

kleinfreund:
What you’re proposing here is to maximize the repetition that NavTabs abstracts away. That is _exactly_ what I’m trying to minimize with the initial addition of NavTabs and also with this change set. Instead of repeating the usage code of KTabs for the dozen uses of NavTabs, we get to simply use what you see in this PR. That’s strictly better.

Now the things you bring up as "impossible" to do with NavTabs are either not relevant right now, or, in fact, easily possible. For example, it’s possible and desirable to adjust an abstraction component like this to meet future requirements. The need to do so isn’t a disadvantage.

> Something I've bumped into today, I've been working on a PR to move all of the DataPlane Viz stuff into it own module and hit a thing where the new NavTab abstraction would be problematic:
> 
> Both inbounds and outbounds summary views use the same 3 tab names `Overview`, `Stats` and `Clusters` I would guess they will always be the same whether you are in inbounds or outbounds, so I wanted them to use the same locales text.
> 
> Doing this is impossible with this new NavTabs abstraction because its too coupled to the route, I _have_ to have these texts repeated, I _have_ to have one for each route and it must be the exact name of the route.

You can pass an arbitrary i18n prefix to NavTabs making it possible to not duplicate these message entries. However, even if that weren’t possible, duplicating those entries is fine, too. It costs us next to nothing to do so.

---

I’m decidedly against removing NavTabs and I disagree with the statement that it has no functional value. The contents of this PR illustrate very clearly that that isn’t true.

johncowen:
I had a thought last night that may or may not work which might be address both our concerns here, let me try it out today, be back later.

johncowen:
Here's what I came up with:

```vue
<Tabs
  :selected="route.active(active)?.name"
>
  <template
    v-for="item in route.children"
    :key="`${item.name}`"
    #[`${item.name}-action`]
  >
    <RouterLink
      :to="{ name: item.name }"
    >
      Text
    </RouterLink>
  </template>
</Tabs>
```

Briefly, addressing your concerns it:

1. Reduces the repetition needed for both setting `KTabs::tabs` _and_ having a `<template>` loop. Instead we just have a single, normal and very recognisable Vue `<template>` loop.
2. Removes the need to set `hide-panels`, I can't think of an occasion where we would want to not hide the panels.
3. Replaces the confusingly named `:model-value` with ':selected'. Whilst I would prefer this to be better still, I'm not sure we can go much further with Vue and KTabs.
4. Hides the fact that KTabs wants everything prefixed with a `#`, and thus avoids more repetition.
5. Keeps a wrapper abstraction around `KTabs`

and addressing my concerns it:

1. Allows us full control over what goes in a tab, its `href` (including any route params) and its `text`.
2. Doesn't use any private types (`StringNamedRouteRecordRaw`), which if I'd have known we needed to export that I would have thought twice about adding it.
3. Is incredibly clear what is happening. It's just a Vue loop.
4. Doesn't use `t` internally.
5. If we don't want to stick to not using KTabs directly (something I'd half rather do), this looks very much like KTabs still.

Lmk

src/app/application/components/route-view/RouteView.vue
```diff
@@ -101,6 +100,7 @@ const children: StringNamedRouteRecordRaw[] = (router.getRoutes().find((route) =
   item.name = String(item.name)
   return item as StringNamedRouteRecordRaw
 }) ?? [])
+const active = computed(() => children.find((item) => item.name === route.name || item.meta?.module === route.meta.module))
```
kleinfreund, line 103: **Comment**: Instead of providing `active` as a function determining the active route, I turned it into a computed property. For one, this avoids requiring the consumer of RouteView to also access `useRoute()` (RouteView already accesses it and so it can make that decision completely internally). For another, it avoids recomputations triggered by functions in templates.

src/app/services/views/ServiceDetailTabsView.vue
```diff
@@ -51,10 +51,30 @@
...
+                if (data.serviceType !== 'external' && name === 'service-config-view') {
+                  return false
+                }
+
+                if (data.serviceType === 'external' && name === 'service-data-plane-proxies-view') {
+                  return false
+                }
+
+                return true
+              })"
```
kleinfreund, line 66: **Comment**: This needing to take place in the template is a downside with this approach because we can’t cache the result using `computed`. Seeing as the need for filtering will go away with https://github.com/kumahq/kuma-gui/pull/2190, it’s not a big problem though.

kleinfreund:
I’ve adapted/integrated the suggestion for the NavTabs interface you provided in the last update to this PR. This should be a fairly good compromise although I’m still not a fan of the amount of repetition it incurs.

johncowen (approved):
Thanks!

PR #2188, opened 2024-02-15
fix: tag list linking to wrong service

kleinfreund:
Change how TagList auto-links to detail views for `kuma.io/service` annotations. Since we no longer have _one_ dedicated detail view for all service types, we can’t unconditionally link to them anymore. Instead, we need to make TagList is contextually aware of the corresponding dataplane type or service type.

### DIFF
--- a/src/app/common/filter-bar/FilterBar.vue
+++ b/src/app/common/filter-bar/FilterBar.vue
@@ -222,12 +222,15 @@ watch(() => currentFields.value, function (newFields, oldFields) {
   }
 })
 
+watch(() => props.query, () => {
+  currentQuery.value = props.query
+  recomputeFields(currentQuery.value)
+}, { immediate: true })
+
 watch(() => currentQuery.value, function () {
   if (currentQuery.value === '') {
     tokenizerError.value = null
   }
-
-  isShowingSuggestionBox.value = true
 })
 
 type CommandKeywords = 'submitQuery' | 'jumpToNextSuggestion' | 'jumpToPreviousSuggestion' | 'closeSuggestionBox'
--- a/src/app/common/TagList.vue
+++ b/src/app/common/TagList.vue
@@ -37,15 +37,37 @@ interface LabelValueWithRoute extends LabelValue {
   isKuma: boolean
 }
 
+const dataplaneTypeToServiceTypeMap = {
+  standard: 'internal',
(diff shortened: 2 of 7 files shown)


kleinfreund:
I’ve put this back in draft because I’m not sure this will work.

What we want is to be able to link to the correct detail view whenever a `kuma.io/service` annotation appears in a `TagList`. But I’m not sure that’s possible: the `kuma.io/service` determines the `name` of a corresponding `ServiceInsight` object, but this is not _generally_ the same as the name of an `ExternalService` or `MeshGateway` resource (the name _can_ be the same, but it doesn’t have to).

For example, an `ExternalService` named `external-service-1` with `kuma.io/service` tag `service-1` will have a corresponding `ServiceInsight` object named `service-1` (crucially, the two names are not the same).

This means that it’s only really possible to link to the correct detail view if we can determine that a `kuma.io/service` annotation belongs to either a `ServiceInsight` object of either serviceType `internal` or `gateway_delegated` (because those detail views query `/meshes/:mesh/service-insights/:name`).

@lahabana, does that sound correct to you?

src/app/common/filter-bar/FilterBar.vue
```diff
@@ -222,12 +222,15 @@ watch(() => currentFields.value, function (newFields, oldFields) {
...
+  currentQuery.value = props.query
+  recomputeFields(currentQuery.value)
+}, { immediate: true })
+
 watch(() => currentQuery.value, function () {
   if (currentQuery.value === '') {
     tokenizerError.value = null
   }
-
-  isShowingSuggestionBox.value = true
```
johncowen, line 230: How come these couple of changes were needed? (both the red and green in this file)

kleinfreund: With the changes here, what we now have are situations where the query parameter of a route can be programmatically changed through something that isn’t an already open filter box: clicking on a service tag in a dataplane summary view doesn’t navigate because the list view is already rendered, but it does populate the query in the URL. What happened then was that clicking on such a link would _open_ the suggestion box and that’s addressed with this change.

There’s also a change here that makes sure that changing `props.query` is reflected in `currentQuery`.

johncowen: Gotcha, thanks

johncowen (approved):

PR #2192, opened 2024-02-16
chore: exposes RouteView::route.children and use for GatewayTabs

johncowen:
Exposes a `RouteView::route.children` so that we can easily make different types of tab-like UI interfaces automatically based on our route config, which in turn means we can completely define this sort of sub navigation from The Outside.

This PR then uses that for the `LinkBox` in `GatewayListTabsView`, but the functionality could also be re-used for all sorts of similar sub navigation widgets.

Note: Whilst DataCollection `predicate` functionality isn't being used here, as DataCollection shows nothing if `items` is zero length, this means that if we configured our router to have no children here, the tab interface would automatically not show/render at all.

### DIFF
--- a/src/app/gateways/views/GatewayListTabsView.vue
+++ b/src/app/gateways/views/GatewayListTabsView.vue
@@ -11,56 +11,56 @@
         mesh: '',
       }"
     >
-      <AppView>
-        <template #title>
-          <h2>
-            <RouteTitle :title="t(`${currentRoute.name === 'builtin-gateway-list-view' ? 'builtin' : 'delegated'}-gateways.routes.items.title`)" />
-          </h2>
-        </template>
-
-        <template #actions>
-          <LinkBox>
-            <RouterLink
-              :class="{
-                'active': currentRoute.name === 'builtin-gateway-list-view',
-              }"
-              :to="{
-                name: 'builtin-gateway-list-view',
-                params: {
-                  mesh: route.params.mesh,
-                },
-              }"
+      <RouterView
+        v-slot="{ Component, route: active }"
+      >
+        <AppView>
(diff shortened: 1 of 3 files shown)


src/app/gateways/views/GatewayListTabsView.vue
```diff
@@ -11,56 +11,56 @@
...
-              }"
-              :to="{
-                name: 'delegated-gateway-list-view',
-                params: {
-                  mesh: route.params.mesh,
-                },
-              }"
+      <RouterView
+        v-slot="{ Component, route: active }"
+      >
```
kleinfreund, line 16: **Question**: Why use this instead of keeping to checking `currentRoute.name`? This seems unnecessarily verbose.

johncowen: I understood you are referring to the bit I removed here?

```js
import { useRoute } from 'vue-router'
...
...
const currentRoute = useRoute()
```

Yeah, I suppose that ^ seems more verbose to me than `{ route: active }`, and especially unnecessary doing all the import/using if its already exported from `RouterView`?

kleinfreund: Except that this made it necessary to move the RouterView all the way up the tree and hook up its `<component ...>` bit, too. The diff for this PR could’ve been much smaller without this change. Also, it’s quite strange to say that needing to import something is verbose. How can an import be verbose? Never mind that the import was already here to begin with.

Imports are fine and good. It’s not needed to go out of your way and remove them, especially if that requires changing a bunch of other stuff in otherwise working code.

kleinfreund (approved):

PR #2211, opened 2024-02-22
Feat/move dataplane list and config into overviews

kleinfreund:
feat(delegated-gateways): move dataplane list into overview

feat(external-services): move config into overview

feat(services): move dataplane list and config into overview

Signed-off-by: Philipp Rudloff <philipp.rudloff@konghq.com>

Resolves #2209.

PR #2212, opened 2024-02-22
feat(builtin-gateways): turn affected DPPs into dataplane list

### LINKED ISSUE #2210: Turn Affected DPPs in Built-in Gateway detail view into a regular Dataplane list
### Description

Follow up to https://github.com/kumahq/kuma-gui/issues/2209

kleinfreund:
Follow up to #2211.

Replace the Affected DPPs section in the built-in gateway detail view with a regular Dataplane list.

Resolves #2210.

### DIFF
--- a/src/app/gateways/views/BuiltinGatewayDetailView.vue
+++ b/src/app/gateways/views/BuiltinGatewayDetailView.vue
@@ -10,9 +10,6 @@
       :params="{
         mesh: '',
         gateway: '',
-        codeSearch: '',
-        codeFilter: false,
-        codeRegExp: false,
         page: 1,
         size: me.pageSize,
         query: '',
--- a/src/app/gateways/views/BuiltinGatewayDetailView.vue
+++ b/src/app/gateways/views/BuiltinGatewayDetailView.vue
@@ -1,116 +1,259 @@
 <template>
-  <RouteView
-    v-slot="{ route, t }"
-    name="builtin-gateway-detail-view"
-    :params="{
-      mesh: '',
-      gateway: '',
-      codeSearch: '',
-      codeFilter: false,
-      codeRegExp: false,
-      dataplane: '',
-    }"
+  <DataSource
+    v-slot="{ data: me }: MeSource"
+    src="/me"
   >
-    <AppView>
-      <DataSource
-        v-slot="{ data, error }: MeshGatewaySource"
-        :src="`/meshes/${route.params.mesh}/mesh-gateways/${route.params.gateway}`"
-      >
-        <ErrorBlock
-          v-if="error"
-          :error="error"
-        />
-
-        <LoadingBlock v-else-if="data === undefined" />
-
-        <div
-          v-else
-          class="stack"
+    <RouteView
+      v-if="me"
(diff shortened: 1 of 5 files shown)


src/app/gateways/views/BuiltinGatewayDetailView.vue
```diff
@@ -1,116 +1,259 @@
...
+        mesh: '',
+        gateway: '',
+        codeSearch: '',
+        codeFilter: false,
+        codeRegExp: false,
+        page: 1,
+        size: me.pageSize,
+        query: '',
+        s: '',
+        dataPlane: '',
```
johncowen, line 20: nit: pretty sure we don't need all of these params here, we only should need the ones that are dependencies of this `*View.vue`

johncowen (commented):
I left one little nit, but I think otherwise this is good to go.

One little suggestion here which is completely up to you at this moment. But if you use `DataLoader` instead of `DataSource` you don't have to add loaders and errors everywhere so its just a one liner

Completely up to you, at some point there will be a project wide refactor adding DataLoader everywhere I guess.

Actually on second glance you have multiple DataSources being used here plus 2 loaders (one of our full loading blocks plus the skeleton loader from AppCollection). I 'think' you could build this differently to only use the one skeleton loader.

Anyway this DataLoader thing is an improvement/suggestion rather than anything, lmk on the unused params inline comment.

kleinfreund:
> One little suggestion here which is completely up to you at this moment. But if you use `DataLoader` instead of `DataSource` you don't have to add loaders and errors everywhere so its just a one liner

Yeah, it’d probably be a good idea to do this, but I’ll look into it in a separate piece of work. DataLoader should be able to handle the dependency of data source 2 (Dataplanes) on data source 1 (MeshGateway).

src/app/gateways/views/BuiltinGatewayDetailView.vue
```diff
@@ -1,116 +1,256 @@
...
+                <template #toolbar>
+                  <FilterBar
+                    class="data-plane-proxy-filter"
+                    :placeholder="`tag: 'kuma.io/protocol: http'`"
+                    :query="route.params.query"
+                    :fields="{
+                      name: { description: 'filter by name or parts of a name' },
+                      protocol: { description: 'filter by “kuma.io/protocol” value' },
+                      tag: { description: 'filter by tags (e.g. “tag: version:2”)' },
+                      zone: { description: 'filter by “kuma.io/zone” value' },
```
johncowen, line 75: Shall we use `can('use zones')` here seeing as we are hiding the column?

johncowen: Actually if we want to do that, lets do that as a follow up everywhere instead of just here.

johncowen (approved):
LGTM!

I had a suggestion, but shouldn't be done here, lmk what you think on that separately.

src/app/gateways/views/BuiltinGatewayDetailView.vue
kleinfreund: 👍🏻

PR #2213, opened 2024-02-22
fix: make attrs utilities more type-safe

kleinfreund:
Changes the type of `attrs` values being passed around to `Partial<Record<string, string>>` to make type errors when accessing data based on empty records (i.e. `{}` as opposed to `{ class: '...' }` visible.

Fix the unsafe passing of `flat.class` to the `difference` function leading to a `Cannot read properties of undefined (reading 'includes')` error.


johncowen:
Q: Where do we call this function without there being a `class` property?

johncowen (approved):
There's more to be done here, but this is fine for now. Thanks for looking into it

PR #2215, opened 2024-02-23
fix: unflagged reference to feature-flagged route

kleinfreund:

### DIFF
--- a/src/app/data-planes/components/DataPlaneList.vue
+++ b/src/app/data-planes/components/DataPlaneList.vue
@@ -76,7 +76,7 @@
             </RouterLink>
 
             <RouterLink
-              v-else-if="row.dataplaneType === 'delegated'"
+              v-else-if="row.dataplaneType === 'delegated' && props.canUseGatewaysUi"
               :to="{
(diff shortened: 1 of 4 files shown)


johncowen (commented):
Aren't we removing the DataPlaneList component in https://github.com/kumahq/kuma-gui/pull/2211?

kleinfreund:
> Aren't we removing the DataPlaneList component in #2211?

Yep, but I still needed access to the feature flag here now.

johncowen:
So you want this in before the other one? Or if we get the other one in, is this no longer required?

kleinfreund:
> So you want this in before the other one? Or if we get the other one in, is this no longer required?

The fix of this PR will still be required regardless of whether https://github.com/kumahq/kuma-gui/pull/2211 goes in first or not. It will require a small change of course to access the feature flag directly.

I just didn’t want to mix that PR and this fix.

johncowen (approved):

johncowen:
Note here's what the issue looks like:

Screenshot 2024-02-27 at 10 37 24

PR #2217, opened 2024-02-23
fix: new service sub navigation not working

kleinfreund:
Fix the new service sub navigation not working on account of the refactoring of NavTabs now relying on automatic `route.children` which in this cases was an empty array. To make this work, I’ve done part of the refactoring I had planned for the time of unflagging the KUMA_GATEWAYS_UI features.


johncowen:
Is there a way to manually test this with the flag turned off? I'm guessing there are tests to ensure that if we broke the "flag off" functionality they would warn us if we'd broken something by accident?

kleinfreund:
> Is there a way to manually test this with the flag turned off? I'm guessing there are tests to ensure that if we broke the "flag off" functionality they would warn us if we'd broken something by accident?

I’m not sure how automatic we can get this under the current setup. I’ve added a rudimentary test in https://github.com/kumahq/kuma-gui/pull/2217/commits/4ff269a1ab9588d2908790f1a724520060d971af for the old service routes, but I expect it to fail in CI because of the feature flag environment thing not picking up this kind of variable at present.

johncowen:
Ah ok, I just managed to turned it off in the PR preview, I must have added my cookie wrong or something.

🤔 yeah, I guess we just keep in mind that if we've enabled this by default, but in prod elsewhere this is disabled, we might be looking at this going "This is fine" but its broken in prod.

Off the top of my head I think it might be time to use gherkin tags so we can switch between different users with different flags enabled or something, that way we can essentially run tests we are otherwise skipping.

---

I did spot one tiny thing by accident but I don't think its important, going here:

https://deploy-preview-2217--kuma-gui.netlify.app/gui/meshes/default/services/internal?page=1&size=50

With the flag off shows me:

Screenshot 2024-02-23 at 15 43 58

But I'd only managed to get to that URL because I was switching the flags back and forth, so I think its fine.

---

I just saw you added the skip so I think we are good to go with this, I guess just keep in mind to be careful that we aren't breaking anything in the feature disabled version. I guess we'll be removing the flag pretty soon?

Let me take another quick look but I think this should be good to go.

kleinfreund:
> 🤔 yeah, I guess we just keep in mind that if we've enabled this by default, but in prod elsewhere this is disabled, we might be looking at this going "This is fine" but its broken in prod.
> 
> Off the top of my head I think it might be time to use gherkin tags so we can switch between different users with different flags enabled or something, that way we can essentially run tests we are otherwise skipping.

Yeah, I think it would be great to have a way to do both:

- Ship something with a feature flag turned off and test it the flagged stuff still (good for providing access to the features in kuma/kong mesh for testing before release)
- Ship something with a feature flag turned on and test the unflagged stuff still

kleinfreund:
> I just saw you added the skip so I think we are good to go with this, I guess just keep in mind to be careful that we aren't breaking anything in the feature disabled version. I guess we'll be removing the flag pretty soon?

I think, yes. This feature set is intended to ship with Kuma 2.7 and the primary objective of the feature flag is to minimize too much fluctuation in the views for users that see it before that release.

---

I’ll briefly consider if something needs to be done about accessing /services/internal with the feature flag off, but as you said, this is probably too minor to be a real issue.

kleinfreund:
> But I'd only managed to get to that URL because I was switching the flags back and forth, so I think its fine.

This shouldn’t ever be a problem for a real user indeed: the services view doesn’t have a summary tray and so the URL path `/services/internal` doesn’t shadow an otherwise working route. The detail view is on `/services/:service/overview`.

johncowen (approved):

PR #2219, opened 2024-02-26
chore: move inbounds/outbounds related things to connections module

johncowen:
Moves all things related to the DataPlane Inbound/Outbound viz into a `connections/` module, which seemed like a good name to mean either inbound or outbound. Honestly I'm kind in two minds about this one but I mainly want to do this because the `data-planes` module was getting more difficult to manage and look at with ~20 similarly named views all in the same folder. 

Apart from the inline comment there shouldn't be any changes here, only moving things around.


src/app/common/SummaryView.vue
```diff
@@ -42,15 +42,15 @@ const emit = defineEmits<{
...
+  h1, h2 {
     --icon-before: url('@/assets/images/icon-wifi-tethering.svg');
   }
 }
 .summary-slideout :deep(.app-view-title-bar) {
   display: flex;
   // Accounts for the absolutely-positioned close button
   margin-right: calc($kui-space-30 + 24px);
-  h1, h2, h3, h4, h5, h6 {
+  h1, h2 {
```
johncowen, line 53: We only use icons currently on h1 and h2s, but this is also a temporary fix for a little CSS bug where we are also putting icons on the h3's inside the drawer. We'll before moving to use KSlideout's title slot soon, which will make this easier. Happy to leave this out and wait for the KSlideout upgrade also, I'm easy.

PR #2239, opened 2024-02-27
chore: sniff the protocol of unknown outbounds

johncowen:
Looks at the stats values to try to figure out whether a service is http or not. See inline comments for a little more info.

### DIFF
--- a/src/app/connections/data/index.ts
+++ b/src/app/connections/data/index.ts
@@ -137,9 +137,20 @@ export const getTraffic = (json: ConnectionStats, filter: (key: string) => boole
           // we only get these on outbounds at the moment (and we don't need the port for the outbound)
           // so we can pretend this doesn't happen
           port: key.split('_').at(-1) ?? '',
-          // TODO we need to detect the protocol based on the existence of
-          // a http property in the stats, just not sure which one yet
-          protocol: item === 'unknown' ? 'http' : item,
+          protocol: item === 'unknown' ? 'tcp' : item,
+        }
+        // on initially setting the TrafficEntry we may have guessed tcp
+        // make an extra check on the stats themselves to try and figure
+        // if this protocol is actually http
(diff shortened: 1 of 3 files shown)


src/app/connections/data/index.ts
```diff
@@ -137,9 +137,20 @@ export const getTraffic = (json: ConnectionStats, filter: (key: string) => boole
...
+        // if this protocol is actually http
+        // unfortunately there doesn't seem to be a way to do the same for grpc
+        if (item === 'unknown') {
+          const obj = value as Record<string, any>
+          switch (true) {
+            // use a stat that can only be non-zero for http (or gRPC)
+            case obj.upstream_rq_total !== 0:
+              traffic[key].protocol = 'http'
+            // if we figure a way to do the same for gRPC add we can add a case here
+          }
```
johncowen, line 153: If we don't know the protocol, we default to `tcp`, then check to see if we've had any HTTP requests, if so we 'upgrade' to HTTP. I checked to see if we could use the same approach to 'upgrade' to gRPC but it does't look it.

If there has been zero traffic, then the protocol will always read `tcp` if its an outbound for a gateway.

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -244,7 +244,7 @@
                   >
                     <!-- Outbounds for gateways report actual traffic on the upstream so we switch to upstream here for non-standard-->
                     <template
-                      v-for="direction in [props.data.dataplane.networking.type !== 'standard' ? 'upstream' : 'downstream'] as const"
+                      v-for="direction in ['upstream'] as const"
```
johncowen, line 247: We also decided to always use `upstream` stats for outbounds

kleinfreund (approved):

PR #2240, opened 2024-02-28
chore: add custom error handler for matcher not found router errors

kleinfreund:
Add a custom error handler to the Vue application that, by default, behaves like the standard error handling (i.e. any Vue realm errors are re-thrown). Specifically for the vue-router errors corresponding to "Matcher not found" errors, it adds a custom error message to the `Error` object because, for some reason, vue-router strips this out in production builds and makes troubleshooting unnecessarily hard.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -38,6 +38,25 @@ async function mountVueApplication() {
   )
 
   const app = await get($.app)((await import('./app/App.vue')).default)
+
+  app.config.errorHandler = function (error) {
+    // Patches the vue-router MATCHER_NOT_FOUND error message back into the error object because vue-router explicitly creates an `Error` object with no message in production environments.
+    if (
+      error instanceof Error &&
+      error.message === '' &&
+      'type' in error &&
+      error.type === 1 &&
+      'location' in error
+    ) {
+      // Changing `error.message` causes Vue to throw the error twice even when `errorHandler` throws the same `Error` object.
+      error.message = `No match for ${JSON.stringify(error.location)}`
+
+      throw error
+    }
+
+    throw error
+  }
+
   app.mount('#app')
 }
 


src/main.ts
```diff
@@ -38,6 +38,25 @@ async function mountVueApplication() {
...
+
+  app.config.errorHandler = function (error) {
+    // Patches the vue-router MATCHER_NOT_FOUND error message back into the error object because vue-router explicitly creates an `Error` object with no message in production environments.
+    if (
+      error instanceof Error &&
+      error.message === '' &&
+      'type' in error &&
+      error.type === 1 &&
+      'location' in error
+    ) {
```
kleinfreund, line 50: **Comment**: I’m unsure if we should add the following here:

```suggestion
    if (
      import.meta.env.PROD &&
      error instanceof Error &&
      error.message === '' &&
      'type' in error &&
      error.type === 1 &&
      'location' in error
    ) {
```

Technically, that’s the most consistent because vue-router only hides the error message in production builds. So with this check, we’d keep the original vue-router error in dev environments.

kleinfreund: It’s the same in practice. Though from reading vue-router source code, I know their error message has an extra conditional adding `JSON.stringify(router.currentRoute.value)` at the end. I left that out for the moment. I also didn’t ever see it being used in practice for some reason, but I reckon we could manually add it.

More importantly, should the nature of the error message change in vue-router, this extra condition would ensure we’d see it in dev.

johncowen: What about if we append to the message rather than overwrite? `error.message = '${error.message} (new stuff)'` (or similar)? Does that cover those bases?

kleinfreund: I don’t want to change the message in dev at all, hence the check for whether `error.message` is empty. I supposed that indirectly includes the prod check as the message is only empty in prod.

johncowen: > So with this check, we’d keep the original vue-router error in dev environments.

But we'd only overwrite it if the message is empty? i.e. we only throw the original vue-router error away if its empty?

> More importantly, should the nature of the error message change in vue-router, this extra condition would ensure we’d see it in dev.

So following the above, I understand that if it becomes non-empty, maybe due to some sort of Vue upgrade or something, we don't overwrite it?

---

Sounds to me like we don't need to add anything DEV/PROD forking. If you're happy with that I can probably 👍 this.

kleinfreund: > But we'd only overwrite it if the message is empty? i.e. we only throw the original vue-router error away if its empty?

We don’t throw it away exactly. We keep all of it intact and re-throw it, we just add a message if it’s empty. The crucial information are still present even on their `Error` object. It’s just that they decided that an error message isn’t necessary/wanted in prod and create the error using `new Error()` but with intact properties, stack trace, etc.

> So following the above, I understand that if it becomes non-empty, maybe due to some sort of Vue upgrade or something, we don't overwrite it?

Yes, correct. If they change their mind (unlikely) we’d start getting their errors because `error.message` wouldn’t be `''`.

---

Yeah, I think we can do without the explicit prod check.

johncowen (approved):

PR #2241, opened 2024-02-28
chore: make use of KSlideout's new title slot

johncowen:
Make use of KSlideout's new title slot and gradually remove old CSS.

There's a little CSS left here in SummaryView, but I'm extremely likely to move that into AppView at some point soon, which means SummaryView will have no CSS whatsoever.

### DIFF
--- a/src/app/common/SummaryView.vue
+++ b/src/app/common/SummaryView.vue
@@ -10,13 +10,21 @@
     data-testid="summary"
     @close="emit('close')"
   >
+    <template #title>
+      <XTeleportSlot :name="id" />
+    </template>
     <slot />
   </KSlideout>
 </template>
 
 <script lang="ts" setup>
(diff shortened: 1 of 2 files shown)


src/app/common/SummaryView.vue
```diff
@@ -10,13 +10,21 @@
     data-testid="summary"
     @close="emit('close')"
   >
+    <template #title>
+      <XTeleportSlot :name="id" />
```
kleinfreund, line 14: **Question**: Why do we need to keep the title of summary views slotted into AppView in the first place? Shouldn’t we just move the to be slotted into KSlideout directly?

johncowen: Well one reason is all our `*View.vue`s look the same, there are no "special" ones just because it goes in a side panel. I suppose this is the biggest reason but there are also a couple of others.

kleinfreund (approved):

PR #2243, opened 2024-02-28
fix: DataLoader exported errors > error

johncowen:
Typo in https://github.com/kumahq/kuma-gui/pull/2126 `errors` > `error`.

DataLoader should export the same shape as DataSource

### DIFF
--- a/src/app/application/components/data-collection/DataCollection.vue
+++ b/src/app/application/components/data-collection/DataCollection.vue
@@ -31,7 +31,7 @@
     />
   </slot>
 </template>
-<script lang="ts" generic="T extends {} = {}" setup>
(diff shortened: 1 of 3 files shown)


src/app/application/components/data-collection/DataCollection.vue
```diff
@@ -31,7 +31,7 @@
     />
   </slot>
 </template>
-<script lang="ts" generic="T extends {} = {}" setup>
+<script lang="ts" generic="T" setup>
```
johncowen, line 34: Vaguely related, T can be `undefined` for DataCollection::items

kleinfreund (approved):

PR #2248, opened 2024-03-01
refactor(gateways): use DataLoader for detail views

kleinfreund:
**chore(DataLouder): make data and errors optional**

Make the `data` and `errors` prop so users of DataLoader can actually not provide them.

Remove the unused default value from the `src` prop. Providing an empty source should require an explicit action on the usage side.

**refactor(gateways): use DataLoader for detail views**

### DIFF
--- a/src/app/application/components/data-source/DataLoader.vue
+++ b/src/app/application/components/data-source/DataLoader.vue
@@ -71,14 +71,13 @@ import ErrorBlock from '@/app/common/ErrorBlock.vue'
 import LoadingBlock from '@/app/common/LoadingBlock.vue'
 
 const props = withDefaults(defineProps<{
-  data: any[]
-  errors: (Error | undefined)[]
+  data?: any[]
+  errors?: (Error | undefined)[]
   src: string
   loader?: boolean
 }>(), {
   errors: () => [],
   data: () => [],
-  src: '',
   loader: true,
 })
 
--- a/src/app/application/components/data-source/DataLoader.vue
+++ b/src/app/application/components/data-source/DataLoader.vue
@@ -73,11 +73,12 @@ import LoadingBlock from '@/app/common/LoadingBlock.vue'
 const props = withDefaults(defineProps<{
   data?: any[]
   errors?: (Error | undefined)[]
-  src: string
+  src?: string
   loader?: boolean
 }>(), {
   errors: () => [],
   data: () => [],
+  src: '',
   loader: true,
 })
 
--- a/src/app/gateways/views/BuiltinGatewayDetailView.vue
+++ b/src/app/gateways/views/BuiltinGatewayDetailView.vue
@@ -23,7 +23,7 @@
           :src="`/meshes/${route.params.mesh}/mesh-gateways/${route.params.gateway}`"
         >
           <DataLoader
-            v-slot="{ data: dataplanesData, error: dataplanesError }: DataplaneOverviewCollectionSource"
+            v-slot="{ data: dataplanesData }: DataplaneOverviewCollectionSource"
             :src="data === undefined ? '' : `/meshes/${route.params.mesh}/dataplanes/for/${data.selectors[0].match['kuma.io/service']}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
             :data="[data]"
             :errors="[error]"
@@ -45,7 +45,6 @@
                 ]"
                 :items="dataplanesData?.items"
                 :total="dataplanesData?.total"
-                :error="dataplanesError"
                 :is-selected-row="(row) => row.name === route.params.dataPlane"
                 summary-route-name="builtin-gateway-data-plane-summary-view"
                 :empty-state-message="t('common.emptyState.message', { type: 'Data Plane Proxies' })"
--- a/src/app/gateways/views/BuiltinGatewayDetailView.vue
+++ b/src/app/gateways/views/BuiltinGatewayDetailView.vue
@@ -18,29 +18,19 @@
       }"
     >
       <AppView>
-        <DataSource
+        <DataLoader
           v-slot="{ data, error }: MeshGatewaySource"
           :src="`/meshes/${route.params.mesh}/mesh-gateways/${route.params.gateway}`"
         >
-          <ErrorBlock
-            v-if="error"
-            :error="error"
-          />
-
-          <LoadingBlock v-else-if="data === undefined" />
-
-          <KCard v-else>
-            <DataSource
-              v-slot="{ data: dataplanesData, error: dataplanesError }: DataplaneOverviewCollectionSource"
-              :src="`/meshes/${route.params.mesh}/dataplanes/for/${data.selectors[0].match['kuma.io/service']}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
-            >
-              <ErrorBlock
-                v-if="dataplanesError !== undefined"
-                :error="dataplanesError"
-              />
-
+          <DataLoader
+            v-if="data"
+            v-slot="{ data: dataplanesData, error: dataplanesError }: DataplaneOverviewCollectionSource"
+            :src="`/meshes/${route.params.mesh}/dataplanes/for/${data.selectors[0].match['kuma.io/service']}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
+            :data="[data]"
+            :errors="[error]"
+          >
+            <KCard v-if="dataplanesData">
               <AppCollection
-                v-else
                 class="data-plane-collection"
                 data-testid="data-plane-collection"
                 :page-number="route.params.page"
@@ -206,9 +196,9 @@
                   />
                 </SummaryView>
               </RouterView>
-            </DataSource>
-          </KCard>
-        </DataSource>
+            </KCard>
+          </DataLoader>
+        </DataLoader>
       </AppView>
     </RouteView>
   </DataSource>
@@ -220,9 +210,7 @@ import { ArrowRightIcon } from '@kong/icons'
 
 import type { MeshGatewaySource } from '../sources'
 import AppCollection from '@/app/application/components/app-collection/AppCollection.vue'
-import ErrorBlock from '@/app/common/ErrorBlock.vue'
 import FilterBar from '@/app/common/filter-bar/FilterBar.vue'
-import LoadingBlock from '@/app/common/LoadingBlock.vue'
 import StatusBadge from '@/app/common/StatusBadge.vue'
 import SummaryView from '@/app/common/SummaryView.vue'
 import WarningIcon from '@/app/common/WarningIcon.vue'
--- a/src/app/gateways/views/BuiltinGatewayDetailView.vue
+++ b/src/app/gateways/views/BuiltinGatewayDetailView.vue
@@ -18,18 +18,18 @@
       }"
     >
       <AppView>
-        <DataLoader
+        <DataSource
           v-slot="{ data, error }: MeshGatewaySource"
           :src="`/meshes/${route.params.mesh}/mesh-gateways/${route.params.gateway}`"
         >
           <DataLoader
-            v-if="data"
             v-slot="{ data: dataplanesData, error: dataplanesError }: DataplaneOverviewCollectionSource"
-            :src="`/meshes/${route.params.mesh}/dataplanes/for/${data.selectors[0].match['kuma.io/service']}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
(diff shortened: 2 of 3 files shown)


src/app/gateways/views/BuiltinGatewayDetailView.vue
```diff
@@ -18,29 +18,19 @@
       }"
     >
       <AppView>
-        <DataSource
+        <DataLoader
```
johncowen, line 21: If I follow what you are doing here then this should be a DataSource which is non-blocking, the idea is you can then pass its data and errors into the single DataLoader (the DataLoader _is_ blocking).

Also, and perhaps separately to this PR, I've noticed that this DataSource should probably be in the Tabs view and its data passed down seeing as we use it in both tabs. Let me know if this point isn't clear

src/app/gateways/views/BuiltinGatewayDetailView.vue
```diff
@@ -18,29 +18,19 @@
...
-              v-slot="{ data: dataplanesData, error: dataplanesError }: DataplaneOverviewCollectionSource"
-              :src="`/meshes/${route.params.mesh}/dataplanes/for/${data.selectors[0].match['kuma.io/service']}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
-            >
-              <ErrorBlock
-                v-if="dataplanesError !== undefined"
-                :error="dataplanesError"
-              />
-
+          <DataLoader
+            v-if="data"
```
johncowen, line 26: FYI: I've been using `data!` inside `DataLoader`s for the moment. It should always be set inside the default slot, and once I finish up the DataSource URI/src auto typing it will end up being typed correctly. Kinda of a nit this so I guess up to you.

src/app/gateways/views/BuiltinGatewayDetailView.vue
```diff
@@ -18,29 +18,19 @@
...
+            v-slot="{ data: dataplanesData, error: dataplanesError }: DataplaneOverviewCollectionSource"
+            :src="`/meshes/${route.params.mesh}/dataplanes/for/${data.selectors[0].match['kuma.io/service']}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
```
johncowen, line 28: I just noticed this, but is it ok to just use `selectors[0]`, can we have selected dataplanes at `selectors[1]` etc do you know?

src/app/gateways/views/DelegatedGatewayDetailView.vue
```diff
@@ -18,19 +18,12 @@
       }"
     >
       <AppView>
-        <DataSource
+        <DataLoader
```
johncowen, line 21: Same as above, mostly (but there will be exceptions) you only want one DataLoader on a page and if there are multiple HTTP requests to be made, use non-blocking DataSources to feed the data into the single blocking DataLoader

src/app/application/components/data-source/DataLoader.vue
```diff
@@ -71,14 +71,13 @@ import ErrorBlock from '@/app/common/ErrorBlock.vue'
...
-  data: any[]
-  errors: (Error | undefined)[]
+  data?: any[]
+  errors?: (Error | undefined)[]
   src: string
   loader?: boolean
 }>(), {
   errors: () => [],
   data: () => [],
-  src: '',
```
johncowen, line 81: How come you removed this?

johncowen (commented):
I like it! I added a few comments, I think mostly just cos I haven't explained the DataSource/DataLoader relationships properly, let me know if you have any questions there.

src/app/application/components/data-source/DataLoader.vue
kleinfreund: `props.src` is not optional and this will never be useful. Not providing `props.src` is a user error. Default values for props are for optional props.

src/app/gateways/views/BuiltinGatewayDetailView.vue
kleinfreund: This needs to be blocking because the result of it becomes part of the next DataLouder’s `props.src`: the request URL depends on data from this response.

src/app/gateways/views/BuiltinGatewayDetailView.vue
kleinfreund: I think the `v-if` directive is much more expressive in saying "use this content if you have data and you may access it in the following descendant nodes". Since I access `data` further down, this also saves me from constantly using `data!`.

src/app/gateways/views/BuiltinGatewayDetailView.vue
kleinfreund: As far as I know, this should be safe, but I’m also not certain (cc @lahabana).

If we need to consider multiple selectors, that eithers means we’d expect multiple Dataplane lists or alternatively we would need to query the Dataplane API with multiple `kuma.io/service` tag filter parameters (does the API handle that? From a quick test this doesn’t seem to be the case: `curl http://localhost:5681/meshes/default/dataplanes\?tag\=kuma.io/service:redis_kuma-demo_svc_6379\&tag\=kuma.io/service:demo-app_kuma-demo_svc_5000` only takes the last tag filter into account).

src/app/application/components/data-source/DataLoader.vue
johncowen: `props.src` is optional, you purposefully don't have to add a `:src` to a DataLoader, you can just pass it data from other DataSources if you like.

If it would have to have a sync on the thinking behind the interactions/relationships with DataSource/DataLoader (or at least the intended functionality 😅 ) let me know.

src/app/gateways/views/BuiltinGatewayDetailView.vue
johncowen: Oh I see, 1 sec lemme look again

src/app/application/components/data-source/DataLoader.vue
kleinfreund: So then why is `props.src` not optional in the definition?

johncowen: That would be the better fix here, looks like I somehow missed that when I initially added this. I guess the examples I added happened to all have `:src`s

src/app/gateways/views/BuiltinGatewayDetailView.vue
johncowen: cool I don't really mind either way at the min. Just to note the `v-if` is runtime, I don't believe the `!` is? FYI: one of the reasons for DataLoader working as it does is to minimize the need for conditionals everywhere (once the URIs are typed all neither of these approaches will be needed I don't think)

Gonna resolve this one.

src/app/application/components/data-source/DataLoader.vue
kleinfreund: I changed this now.

src/app/gateways/views/DelegatedGatewayDetailView.vue
kleinfreund: I changed this back, too, now.

src/app/gateways/views/BuiltinGatewayDetailView.vue
johncowen: I think if you use:

```
 :src="typeof data === 'undefined' ? '' : `/meshes/${route.params.mesh}/dataplanes/for/${data.selectors[0].match['kuma.io/service']}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"

```

For the `:src` for the DataLoader and remove its `v-if="data"` (and remember to turn the wrapping one into a DataSource) then we will get what we want here. lmk

kleinfreund: But why wouldn’t I just use DataLoader on the top level if I want it to be blocking anyway?

src/app/application/components/data-source/DataLoader.vue
johncowen: Awesome, thanks!

src/app/gateways/views/BuiltinGatewayDetailView.vue
johncowen: 🤔 actually yeah I guess its fine as it is also, I suppose I'm just automatically thinking of the "most of the time we should only have one DataLoader thing" - also I think if we move the MeshGateway call to the route above this one to share it, you'll have to do the sort of thing in the suggestion eventually anyway. But yeah, this is fine I think. Will resolve this one.

src/app/gateways/views/DelegatedGatewayDetailView.vue
```diff
@@ -77,17 +67,14 @@
...
+                  :errors="[error]"
                 >
-                  <ErrorBlock
-                    v-if="dataplanesError !== undefined"
-                    :error="dataplanesError"
-                  />
-
                   <AppCollection
-                    v-else
+                    v-if="dataplanesData"
```
johncowen, line 77: Ah this is the thing we have to be a little careful of. AppCollection has its own loader/skeleton, which is very coupled to its look/feel. So we need to use AppCollections loader but use DataLoaders errors, so here you should be able to use `DataLoader::loader="false"` to render DataLoaders default slot even when loading ..... aaaaaah ok, right this is a reason why we should use DataSource for the other conversation I just resolved 😅 

Yeah I just tried, so in both places where we load a AppCollection in this PR, use `DataLoader::loader="false"` and remove the `v-if="dataplanesData/data"`. In the above conversation you'll also need to implement the suggestion I had to use DataSource at the top-level.

Once you've done this you'll see we have a skeleton loader from AppCollection instead of a spinner.

Sorry for the confusion, lmk if this isn't clear.

kleinfreund: Hm, this is a pretty inconvenient state when interacting with DataLoader. DataSource doesn’t include an error block so with that it’s okay to reason about whether you need to put an error block in the view component or let AppCollection handle it via KTable. But with DataLoader, one should be able to make the assumption that one _definitely_ doesn’t have to deal with that question. But I guess that’s owed to the fact that we can’t just stop using AppCollection’s (read: KTable’s) loading state stuff (though I’d rather use ours).

I tried implementing these suggestions in the last suggestion.

johncowen: Yeah this:

> But I guess that’s owed to the fact that we can’t just stop using AppCollection’s (read: KTable’s) loading state stuff

--

Its a little awkward, but the loading state is tied to AppCollection because the skeleton needs to look like a table. I thin it might look more awkward here because the loading logic is a tiny bit more involved. If you take a look at `MeshListView` its way less awkward/more straightforwards. Its just the addition of a `loader="false"`.

Let me take another look at this with it switched up.

johncowen: K I think this looks good now, let me take one last scan and I think I can 👍 . 2tics

johncowen: Ok, it looks like BuiltIn if fine but delegated I think needs a tiny bit more work. If you set `KUMA_LATENCY=3000` you'll see this:

Screenshot 2024-03-01 at 13 11 58

At a first glance, you _may_ need 2 DataLoaders here, a spinner for the top content and a skeleton via AppCollection for the list of dataplanes underneath.

I might be mistaken though, let me know either way

kleinfreund: What I did here now was to move the outer data source for the ServiceInsight further in so that it only wraps around the top card. I changed it to a DataLoader because it needs to have its independent loading and error handling.

I also changed the AppCollection still having a `v-if` which prevented the KTable loading mechanism to handle the loading state. It’s also a DataLoader but one with `:loader="false"`.

johncowen: gotcha lemme take another look in a sec

johncowen:
This looks great to me now, one thing I spotted that isn't in the diff here is that we need to remove `:error` from both instances of `AppCollection` here. `DataLoader` deals with the errors now so ApplCollection isn't even rendered anymore if there is an error.

Here https://github.com/kumahq/kuma-gui/blob/90e00f27ebc4b80c7b39800af2319f76a883b224/src/app/gateways/views/DelegatedGatewayDetailView.vue#L91

And here https://github.com/kumahq/kuma-gui/blob/90e00f27ebc4b80c7b39800af2319f76a883b224/src/app/gateways/views/BuiltinGatewayDetailView.vue#L48

Let me know on those then I think this is good to go

johncowen (approved):
Nice PR, thanks!

src/app/gateways/views/BuiltinGatewayDetailView.vue
kleinfreund: FYI I opened https://github.com/kumahq/kuma/issues/9481 and we’ll settle this there.

PR #2249, opened 2024-03-01
feat(builtin-gateways): add rules-based policies

### LINKED ISSUE #2208: Built-in Gateways: visualize how traffic is routed to destination services using data from the /_rules API
In the built-in gateway detail view, show how traffic coming in through a MeshGateway’s listeners is routed to its corresponding destination services.

**References**:

- https://kuma.io/docs/2.6.x/guides/gateway-builtin
- https://kuma.io/docs/2.6.x/policies/meshhttproute
- https://github.com/kumahq/kuma/issues/9589

## How it works

Initially, what we have in this view is a MeshGateway resource and its listeners. Additionally, we query the `/_rules` API which gives us the routing rules defined for this MeshGateway.

These rules are the result of creating MeshHTTPRoutes policies which target the MeshGateway (e.g. with a top-level targetRef with kind `MeshGateway` and the MeshGateway’s name). They specify rules that decide how the traffic is routed from the listeners to their corresponding destination services (this information is _defined_ in MeshHTTPRoutes but exposed to us here via the `/_rules` API). For example, a rule could specify that only requests with a certain request header are routed to service A while all others are routed to service B.

For determining how a request’s traffic is routed, a listener is always selected first. There are a few rules for selecting a listener: exact hostname match, prefix hostname match, order of listeners. Once a listener is selected, only routes matching this listeners are relevant. Routes only applying to other listeners are not relevant at all.

Furthermore, a MeshGateway can specify `tags` in its listeners and MeshHTTPRoutes can specify tags in their top-level targetRef.

**MeshGateway**:
`/meshes/default/meshgateways/demo-app`

<details>
<summary>YAML</summary>

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshGateway
mesh: default
metadata:
  labels:
    kuma.io/origin: zone
  name: demo-app
spec:
  selectors:
  - match:
      kuma.io/service: demo-app_gateway
  conf:
    listeners:
    - hostname: foo.example.com
      port: 80
      protocol: HTTP
    - hostname: bar.example.com
      port: 80
      protocol: HTTP
```

</details>

**MeshHTTPRoute**:
`/meshes/default/meshhttproutes/demo-app.kuma-system`

<details>
<summary>YAML</summary>

```yaml
apiVersion: kuma.io/v1alpha1
kind: MeshHTTPRoute
metadata:
  labels:
    kuma.io/mesh: default
    kuma.io/origin: zone
  name: demo-app
  namespace: kuma-system
spec:
  targetRef:
    kind: MeshGateway
    name: demo-app
  to:
  - targetRef:
      kind: Mesh
    hostnames:
    - foo.example.com
    rules:
    - default:
        backendRefs:
        - kind: MeshService
          name: demo-app_kuma-demo_svc_5000
          weight: 1
      matches:
      - path:
          type: PathPrefix
          value: /
  - targetRef:
      kind: Mesh
    rules:
    - default:
        backendRefs:
        - kind: MeshService
          name: redis_kuma-demo_svc_6379
          weight: 1
      matches:
      - path:
          type: PathPrefix
          value: /
```

</details>

**`/_rules` API response**:
`/meshes/default/meshgateways/demo-app/_rules`

<details>
<summary>YAML</summary>

```yaml
resource:
  mesh: default
  name: demo-app
  type: MeshGateway
rules:
- fromRules: []
  toRules:
  - conf:
      rules:
      - matches:
        - path:
            value: "/"
            type: PathPrefix
        default:
          backendRefs:
          - kind: MeshService
            name: demo-app_kuma-demo_svc_5000
            weight: 1
      hostnames:
      - foo.example.com
    matchers: []
    origin:
    - mesh: default
      name: demo-app.kuma-system
      type: MeshHTTPRoute
  - conf:
      rules:
      - matches:
        - path:
            value: "/"
            type: PathPrefix
        default:
          backendRefs:
          - kind: MeshService
            name: redis_kuma-demo_svc_6379
            weight: 1
    matchers: []
    origin:
    - mesh: default
      name: demo-app.kuma-system
      type: MeshHTTPRoute
  type: MeshHTTPRoute
  warnings: []
```

</details>

## Presentation

We’re thinking of a UI that allows a user to select a listener to "drill down" and then show (1) the rules at play and (2) which destination services you’d resolve to based on these individual rules.

I’m adding a screenshot of how this would look from the point of view of the data hierarchy. Per listener on the MeshGateway, we would show one or multiple entries representing the rules relevant for this listener. The thing labeled "Rules" in the screenshot are the criteria that decide whether a given request is routed to the service linked to under the label "Services". That service is the destination service of the traffic, provided the request matches the rules.

image

## Questions

- **Q**: (1) How does a MeshHTTPRoute apply to only a subset of a MeshGateway’s listeners? (2) Does the `/_rules` API call resolve this or will we have to do some form of filtering on the response in the UI?

  **A**: (1) By matching listeners being specified in `spec.to[].hostnames`. (2) The selected hostnames are available on `rules[].toRules[].conf.hostnames`.

- **Q**: For the purposes of this visualization, the only relevant entries from the `/_rules` API are `toRules` with `type: 'MeshHTTPRoute'`, correct?

  **A**: For the time being, yes. We should also consider MeshTCPRoute and, later, other policy types influencing this view.

- **Q**: In the context of MeshGateway and MeshHTTPRoutes targeting them, can backendRefs have a targetRef.kind other than `MeshService` or `MeshServiceSubset`? I’m asking because I want to know if we need to safe-guard this in some way when creating links (e.g. if we have a `kind: MeshGateway` here we would need to create different links).

  **A**: No, only MeshService and MeshServiceSubset are relevant here.

- **Q**: MeshHTTPRoutes _may_ filter traffic based on hostname (but they don’t have to). We want a UI that groups traffic to destination services by listeners first. (1) I assume that means that routes _without_ hostnames filter apply to all listeners of a MeshGateway, correct? (2) Similarly, I assume that routes with hostnames filter apply to all listeners that don’t specify a hostname (or which specify a matching hostname), correct?

  **A**: Two important parts need to be considered here: (1) a MeshGateway’s listener without `hostname` implicitly has a wildcard hostname meaning requests to all origins can be handled by this listener. (2) a route filtering by hostnames applies to a listener if the intersection with a listener’s hostname is non-zero. In other words, taking point 1 into account, a route filtering by hostnames will always apply to any listener without a hostname (the intersection of the implicit wildcard hostname with any hostnames list is always non-zero). In other _other_ words, a route only _doesn’t_ apply to a listener if it filters by hostnames and the listener doesn’t have a matching (explicit) hostname.

- **Q**: Can `MeshGateway.conf.listeners[].hostname` be set to `'*'` explicitly, and does it mean the same as omitting the field (i.e. no hostname or `hostname: '*'` means any hostname) ?

  **A**: Yes, and yes.

- **Q**: Can `MeshHTTPRoute.spec.to[].hostnames[]` be set to `['*']` explicitly, and does it mean the same as omitting the field (i.e. no hostnames or `hostnames: ['*']` means any hostnames) ?

  **A**: Yes, and yes.

- **Q**: Can `MeshGateway.conf.listeners[].hostname` or `MeshHTTPRoute.spec.to[].hostnames[]` have values like `*.foo.example.org`?

  **A**: Yes, they can.

- **Q**: For each listener, should we only show rules from routes with a top-level targetRef whose tags are a subset of those on the listener?

  **A**: Yes, we should filter rules by tags.

- **Q**: For each listener, should we only show rules from routes whose to targetRef have a hostnames list matching that of the listener?

  **A**: No, we shouldn’t filter rules by hostname.

- **Q**: How do MeshHTTPRoute and MeshTCPRoute policies match a MeshGateway’s listeners?

  **A**: MeshHTTPRoutes match listeners with protocol HTTP or HTTPS; MeshTCPRoutes match listeners with protocol TCP or TLS.

- **Q**: Should we show backendRef weights? If yes, should we also show `weight: 1` (the default)?

  **A**: --

kleinfreund:
You can use the custom [`test-meshgateway` entity](https://deploy-preview-2249--kuma-gui.netlify.app/gui/meshes/default/gateways/builtin/test-meshgateway/overview) to look at a "real enough" and not randomized example.

## Changes

**chore: mock MeshHTTPRoute and MeshGateway _rules**

**feat(builtin-gateways): add rules-based policies**

Update the built-in gateway detail view to show which rules from MeshHTTPRoutes and MeshTCPRoutes apply to which listeners and to which destination services they resolve.

Move the Dataplanes table into its own tab because selecting a listener in the new UI makes the content beneath it jump around.

Resolves #2208.

## Screenshots

image


johncowen:
Heads up incase: I just noticed we are probably going to tread on toes here between this one and https://github.com/kumahq/kuma-gui/pull/2247, did you want to try and get 2247 in first?

kleinfreund:
I’m going to open this up for review now. There are some items left to discuss with design mostly around the topic of information hierarchy in the Routes and Services part of the view. There’s quite a lot of relevant information that needs to be shown. The most tricky aspect is nicely representing multiple rules (i.e. the pair of matches and default) within a route card.

johncowen:
Did I see right that we aren't using nested routes for this? And if not how come we aren't? I guess there's a reason not to?

kleinfreund:
> Did I see right that we aren't using nested routes for this? And if not how come we aren't? I guess there's a reason not to?

That’s a bit of a vague question, but I’m guessing you mean the boundary between the view where a selected listener is in play and the one where there isn’t. I had it like that before and found no benefits to it (rather only complications) and changed it again.

johncowen:
> (rather only complications)

> And if not how come we aren't? I guess there's a reason not to?

i.e. Could you detail these complications for me?

kleinfreund:
> i.e. Could you detail these complications for me?

It’s just more cumbersome to use a RouterView as the interface for a statically resolved component. You get no prop checking and need a separate route. There’s no need for a separate route here.

kleinfreund:
> Please could you use nested routing for this.
> 
> There have been several PRs over the past few weeks moving areas that weren't using nested routing to use nested routing. Doing this fixed several bugs and made several UX improvements (not to mention the DX improvements)

If you can point out what can be improved by using nested routes here, I will consider it. For the time being, I’ll consider this suggestion one of technicality. I’m not interested in changing things around "just because".

johncowen:
The entire project uses nested routing to do this sort of thing, I don't see any reason to do anything different here.

At a general level, it prevents very many negative things as has been seen in moving the places where it wasn't being used to use it.

src/app/data-planes/data/index.ts
```diff
@@ -277,9 +288,8 @@ export const Rule = {
         // with a single entry so it looks like to and from rules
         const proxy = typeof item.proxyRule !== 'undefined'
           ? [{
-            ...Rule.fromObject(item.proxyRule as InspectBaseRule),
+            ...Rule.fromObject(item.proxyRule as InspectBaseRule, 'proxy'),
```
johncowen, line 291: I would rather these methods always took a single argument, this is why previously the ruleType was applied here instead of inside `Rule`. Moreover if we are still adding the `type` here also we may aswell continue to add the `ruleType` here.

If it gets to a point where we absolutely must pass extra things in then fair enough, but I don't think we at that point.

src/app/data-planes/data/index.ts
```diff
@@ -222,24 +223,36 @@ export const SidecarDataplane = {
...
-      ruleType: '',
+      ruleType,
       ...rest,
-      config: Object.keys(conf || {}).length > 0 ? conf : {},
+      config: {
+        ...conf,
+        // An omitted or empty hostnames list implies the wildcard hostname (meaning any hostnames apply).
+        hostnames: Array.isArray(conf.hostnames) && conf.hostnames.length > 0 ? conf.hostnames : ['*'],
+        rules: Array.isArray(conf.rules) ? conf.rules : [],
+      },
```
johncowen, line 255: Isn't config the stuff we show in the UI? Shouldn't we avoid amending that? Or at least provide some sort of `raw` value that we don't touch to then display?

src/app/gateways/components/RuleMatch.vue
```diff
@@ -0,0 +1,99 @@
...
+          <dt>
+            <KBadge appearance="neutral">
+              Query parameter
+            </KBadge>
+          </dt>
+
+          <dd>
+            <span class="key">Type:</span> <span>{{ param.type }}</span>
+            <span class="key">Name:</span> <span>{{ param.name }}</span>
+            <span class="key">Value:</span> <span>{{ param.value }}</span>
```
johncowen, line 47: Tiny thing: What do you think of a nested `dl` here?

src/app/gateways/components/RuleMatch.vue
```diff
@@ -0,0 +1,99 @@
...
+          <span class="key">Value:</span> <code>{{ props.match.path.value }}</code>
+        </dd>
+      </div>
+    </template>
+
+    <template v-if="props.match.queryParams && props.match.queryParams.length > 0">
+      <div
+        v-for="(param, index) in props.match.queryParams"
+        :key="index"
+      >
```
johncowen, line 36: I know a `div` is fine within a `dl` in order to wrap the dt/dd combo. Not totally sure about nested divs any idea?

src/app/gateways/views/BuiltinGatewayDetailView.vue
```diff
@@ -1,242 +1,45 @@
...
-  <DataSource
-    v-slot="{ data: me }: MeSource"
-    src="/me"
+  <RouteView
+    v-slot="{ route }"
+    name="builtin-gateway-detail-view"
+    :params="{
+      mesh: '',
+      gateway: '',
+      listener: '0',
```
johncowen, line 8: I don't know if you saw my comment offline, but I think we can use a `hostname:port` combo for an identifier for these? wdyt?

johncowen (commented):
Additional to the nested routes thing, I did a quick scan here and added a few more comments/questions. Lemme know on those.

src/app/gateways/components/RuleMatch.vue
kleinfreund: Ah, good point. I don’t know for sure, but in this case, we don’t need two levels anyway, so I removed one.

src/app/data-planes/data/index.ts
kleinfreund: Good point. Seeing as we’re now accessing specific data from a `conf` object, using `raw` becomes the preferred name for this, I think. I’ve changed this now and replaced the few instances of accessing `config` elsewhere to `raw` (for `Rule`, not for other objects that might still use `config`; we should change those, too).

src/app/gateways/views/BuiltinGatewayDetailView.vue
kleinfreund: I’m not entirely certain that is reliable. There is certainly validation in place preventing a clash of `${hostname}:${port}`, but I don’t know if there are edge cases that make this more tricky than it seems. While it’s not exactly pretty, I think the index is a reliable way to identify a listener for now (until someone adds an non-last listener, but the same problem exists when changing hostname or port).

Also, note that as far as I can tell, only the port is required. In the case of an absent hostname, the hostname is implied to be `*`.

Slightly related: I would like for the `/_rules` API to identify which listeners their rules apply to. This, too, would need to use _some_ form of identifier (even if it’s just the listener index).

src/app/gateways/components/RuleMatch.vue
kleinfreund: I briefly thought about it, and didn’t really have a clear picture about it. It feels a bit "spammy" to mark up everything as a definition list. I’m not sure how meaningful the notion of "this term is defined by another definition list" is.

src/app/data-planes/data/index.ts
kleinfreund: There was a specific reason why I needed to move the addition of `ruleType` inside the transformer: I wanted a stricter type for `ruleType` so that autocompletion of the field knows of the three possible values. That meant that in order for `fromObject` to return a `Rule`, it needs to set one of the three allowed values (`''` not being one of them). And that can only happen inside the `fromObject` function.

Now, theoretically `type` could also be pulled into the function, but for `type` there is neither an advantage nor a forcing factor: it doesn’t come from `rule` and its resulting type will only be `string` anyway. While it would be nice to have a `PolicyTypeName` union type of all policy type names to make this more precise, it would require quite a bit more work to establish one.

src/app/gateways/views/BuiltinGatewayDetailView.vue
johncowen: Yeah I'm not 100% sure either, but considering there is validation stopping you from adding a clash of a `hostname:port` combo, I'd say its extremely unlikely we can't rely on it. Again considering there is validation preventing you from doing this, I'd almost say it would be a bug if someone did manage to do that.

Absent hostname would produce a `/*:80/` in the URL.

---

Separately, if we are using the URLs as state (ie. bookmarkable, sharable state), what happens if I share a link to listener `/3/` and then I delete listener number 2? Listener 3 would become listener 2 right? Then the link I just shared would be wrong.

lahabana:
- Compact Name and Value for `matches` 
- Add tags for meshServiceSubset (represent the targetRef). Can we design a component for targetRef rendering/linking? --> Potentialy do as a follow up issue/PR
- For weight e.g: :wrench::20 can we have a weight icon and a tooltip to say `weight`. We can ask team to get an icon for this: https://kong.github.io/icons/

src/app/data-planes/data/index.ts
johncowen: `type` should always be string seeing as policies are pluggable.

So your reasoning here not to have `'' | 'to' | 'from' | 'proxy'` is incase you accidently search for a `ruleType=''`? (which would return none)

kleinfreund:
> 1. Compact Name and Value for `matches`
> 2. Add tags for meshServiceSubset (represent the targetRef). Can we design a component for targetRef rendering/linking? --> Potentialy do as a follow up issue/PR
> 3. For weight e.g: 🔧:20 can we have a weight icon and a tooltip to say `weight`. We can ask team to get an icon for this: https://kong.github.io/icons/
> 4. Show a `*` matcher when `matchers` is empty (similar to what we do for policies).
> 5. Remove `Overview and type` to just have `MeshHTTPRoute` or `MeshTCPRoute` in top left (like we do for listener type).
> 6. Needs support for filters. (how do this relate to point 8).
> 7. Discuss with design to not show matchers and origin by default (have a toggle for users to see them).
> 8. Discuss with design the possibility of having Matches and Services next to each other instead of one after the other.

I’ve done the following now:

> 1. Compact Name and Value for `matches`
> 2. Add tags for meshServiceSubset (represent the targetRef). Can we design a component for targetRef rendering/linking? --> Potentialy do as a follow up issue/PR
> 3. For weight e.g: 🔧:20 can we have a weight icon and a tooltip to say `weight`.
> 4. Show a `*` matcher when `matchers` is empty (similar to what we do for policies).
> 5. Remove `Overview and type` to just have `MeshHTTPRoute` or `MeshTCPRoute` in top left (like we do for listener type).

Left to do:

> 6. Needs support for filters. (how do this relate to point 8).
> 7. Discuss with design to not show matchers and origin by default (have a toggle for users to see them).
> 8. Discuss with design the possibility of having Matches and Services next to each other instead of one after the other.

... and the weight icon thing, but that can be done in a follow up, too.

src/app/gateways/views/BuiltinGatewayDetailView.vue
kleinfreund: > what happens if I share a link to listener `/3/` and then I delete listener number 2? Listener 3 would become listener 2 right?

Yes. But the link to whatever `${hostname}:${port}` would’ve been for listener 2 would also no longer work. 🤷🏻 

I’ll consider updating this, but I might do it in a follow up.

src/app/data-planes/data/index.ts
kleinfreund: > So your reasoning here not to have `'' | 'to' | 'from' | 'proxy'` is incase you accidently search for a `ruleType=''`? (which would return none)

Why would I allow `''` in the first place? It’s not useful and I can easily avoid it by pulling the rule type stuff inside.

lahabana:
Seems like we've lost the path type now:

image

src/app/data-planes/data/index.ts
johncowen: Well yeah but if ideally we want to stick to the single argument `fromObject` interface, it kinda depends what we prefer to keep. I'm not sure in what circumstance anyone would try and search ruleType by `''` even if Typescript allowed them to do that.

We could also have a private type in the middle to allow us to create a rule without a `ruleType` and then add the type where we did previously. Let me look/think about this a little more.

Btw this isn't a blocker.

src/app/gateways/views/BuiltinGatewayDetailView.vue
johncowen: > But the link to whatever ${hostname}:${port} would’ve been for listener 2 would also no longer work.

And that would be the correct behaviour becuase I've deleted it. "Listener 3" would still be available at the same link of `hostname:port`.

kleinfreund: For the sake of trying how this would work, I implemented this locally and there’s a problem with it: we don’t know `${hostname}:${port}` ahead of getting the MeshGateway. That means we can’t load the view with a pre-selected listener. We would need to update the URL after loading the MeshGateway. I’m going to leave this as an index for now.

kleinfreund:
> Seems like we've lost the path type now:

I’ve fixed this.

I also implemented this now:

> 6\. Needs support for filters. (how do this relate to point 8).

This is a bit of a best effort implementation as there’s quite a bit to handle with filters. We can iterate on this in a follow up.

kleinfreund:
I created an issue for the two outstanding items that depend on design input we discussed previously: https://github.com/kumahq/kuma-gui/issues/2368

The rest of the feedback was addressed and this PR can be re-reviewed.

src/app/data-planes/data/index.ts
johncowen: What if we used `to` as a default instead of `` inside `Rule.FromObject`, then (as was the case previous to the changes here) we immediately overwrite it with the value from the API always. That way we get the types you are after and we keep the single argument `fromObject` API?

There are other ways to do this (we could have multiple `To/From/ProxyRule.fromObject` etc) but I think this is the simplest and I think it answers both concerns, wdyt?

johncowen:
Is it just me or am I not seeing any listeners now on the preview site now? No matter which one I click on I get this:

Screenshot 2024-04-02 at 12 31 45

kleinfreund:
> Is it just me or am I not seeing any listeners now on the preview site now? No matter which one I click on I get this:

I can definitely get gateways with listeners entries, but it does seem a little "too rare" for preview purposes. I can see if I can tweak this somehow.

src/app/data-planes/data/index.ts
kleinfreund: I think that’d be acceptable. I’ve changed this now.

johncowen:
Ok cool, if it's just a mock thing don't worry too much about it

src/test-support/mocks/src/meshes/_/meshgateways.ts
```diff
@@ -36,14 +38,27 @@ export default ({ env, fake, pager }: EndpointDependencies): MockResponder => (r
...
-            listeners: [
-              {
-                hostname: 'foo.example.com',
-                port: 8080,
-                protocol: 'HTTP',
+            listeners: Array.from({ length: listenerCount }).map(() => {
+              return {
+                hostname: `${fake.internet.domainWord()}.${fake.internet.domainName()}`,
+                port: fake.internet.port(),
+                protocol: fake.helpers.arrayElement<'HTTP' | 'HTTPS' | 'TCP' | 'TLS'>(['HTTP', 'HTTPS', 'TCP', 'TLS']),
```
johncowen: Do we need to specify the types for these things? (I think I saw another one further up also)

src/test-support/FakeKuma.ts
```diff
@@ -294,6 +295,43 @@ export class KumaModule {
...
+            value: 'test-1',
+          },
+          {
+            name: 'X-Test-Header',
+            value: 'test-2',
+          },
+        ],
+      }),
+    }
+  }
```
johncowen, line 334: I'd like to try to avoid having to many 'heavy' things on the fake kuma object. Smaller things aren't so bade (like `tags` for example). Here something like `kuma.httpHeaders` etc would be better than the full 'heavier' ruleMatch object.

Similarly to our actual application frontend templates, our mock templates should be made up of these smaller 'components' to give more flexibility/visibility - sometimes a bit of repetition is an ok trade for flexibility/visibility.

Tinier thing which might be a part-consequence of the above is that I'm not entirely sure whether we'd ever get two headers with the same name in an API response - not sure what would happen in real kuma if you tried to do that.

Anyway, bit of a comment more than anything, not a hard request unless you wanted to change something here yourself.

src/app/gateways/components/RuleFilter.vue
```diff
@@ -0,0 +1,145 @@
...
+  display: inline-flex;
+  flex-wrap: wrap;
+  gap: $kui-space-40;
+  align-items: baseline;
+}
+
+.text-neutral {
+  color: $kui-color-text-neutral;
+}
+</style>
```
johncowen, line 145: Would it make sense for this and RuleMatch go into the `rules` module?

johncowen (commented):
I had a couple more comments, but generally I think we should aim to get this in and circle back/iterate at a later date with some things like:

1. Nested routing
2. More stable `hostname:port` route segments
3. Possibly DataCollection for filtering/finding relationships

Lmk on the other comments and I think we are good to merge this (assuming @lahabana s comments are covered)

src/app/gateways/components/RuleMatch.vue
```diff
@@ -0,0 +1,105 @@
...
+        <dt>
+          <KBadge appearance="neutral">
+            Header
+          </KBadge>
+        </dt>
+
+        <dd>
+          <div class="list">
+            <span><span class="text-neutral">{{ t('http.api.property.type') }}:</span> {{ header.type ?? 'Exact' }}</span>
```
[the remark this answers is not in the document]
kleinfreund: I wanted to do this, but held off for the time being because the source of data is nested _quite_ deep in the _rules API. Eventually, we should probably do this though.

src/test-support/mocks/src/meshes/_/meshgateways.ts
kleinfreund: Yes, the method is otherwise typed as returning an array of strings which fails the expected type.

src/app/gateways/components/RuleMatch.vue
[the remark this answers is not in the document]
johncowen: Cool lets consider this in later PRs then 👍

src/app/gateways/components/RuleFilter.vue
kleinfreund: Yes, I’m moving them provided the directory already exists.

src/test-support/mocks/src/meshes/_/meshgateways.ts
johncowen: What if we used `as const` on these instead?

kleinfreund: I don’t see a reason to do that. Although a bit verbose, this gives us what we need here.

src/app/gateways/components/RuleFilter.vue
kleinfreund: Done.

kleinfreund:
> * Fix i18n problem (don't i18n keys)
> * Don't use a pill for type/action use `key:value`
> 
> image
> * For request mirror don't put the number in the service pill. Put it aside as `50%`

I’ve addressed these items now.

kleinfreund:
@johncowen Please finish reviewing this.

johncowen (commented):
I noticed that I could not see a TCP listener with Routes/Services and when I tried it out it looks like I don't see the service on a TCP listener/route?

Screenshot 2024-04-04 at 11 09 25

Could we take a look at that?

Apart from that I added a couple of more minor things inline. lmk

src/app/gateways/views/BuiltinGatewayDataplanesView.vue
```diff
@@ -0,0 +1,250 @@
...
+          :src="`/meshes/${route.params.mesh}/mesh-gateways/${route.params.gateway}`"
+        >
+          <div class="stack">
+            <DataLoader
+              v-slot="{ data: dataplanesData }: DataplaneOverviewCollectionSource"
+              :src="meshGateway === undefined ? '' : `/meshes/${route.params.mesh}/dataplanes/for/${meshGateway.selectors[0].match['kuma.io/service']}?page=${route.params.page}&size=${route.params.size}&search=${route.params.s}`"
+              :data="[meshGateway]"
+              :errors="[error]"
+              :loader="false"
+            >
```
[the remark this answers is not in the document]
kleinfreund: I’ve changed this.

src/test-support/mocks/src/meshes/_/meshgateways/_/_rules.ts
```diff
@@ -0,0 +1,130 @@
+import type { EndpointDependencies, MockResponder } from '@/test-support'
+import type { InspectBaseRule, InspectRulesForDataplane } from '@/types/index.d'
+
+export default ({ env, fake }: EndpointDependencies): MockResponder => (req) => {
+  const mesh = req.params.mesh as string
+  const name = req.params.name as string
+
+  const haxProxyRuleOverride = env('KUMA_DATAPLANE_PROXY_RULE_ENABLED', '')
+  const haxProxyRule = haxProxyRuleOverride !== '' ? haxProxyRuleOverride === 'true' : fake.datatype.boolean()
```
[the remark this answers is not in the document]
kleinfreund: Fixed.

src/test-support/mocks/src/meshes/_/meshgateways/test-meshgateway/_.ts
```diff
@@ -0,0 +1,51 @@
...
+            },
+            tls: {
+              mode: 'PASSTHROUGH',
+            },
+          },
+        ],
+      },
+    } satisfies MeshGateway,
+  }
+}
```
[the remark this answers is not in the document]
kleinfreund: For PR previews for one and generally for testing with more real/less random data.

kleinfreund:
> Could we take a look at that?

That looks like an invalid TCP route. It has no `matches` property in its `rules` objects and so the `_rules` API doesn’t have any rules for this.

lahabana:
LGTM from a design POV

johncowen (approved):
Awesome, here's what I'm seeing with that applied 👍 

Screenshot 2024-04-04 at 12 37 13

Lets get this in!

Issue #2162, opened 2024-02-08
split PolicyList component into two nested routes and remove policy list component


johncowen:
### Description

See https://github.com/kumahq/kuma-gui/pull/2159#issuecomment-1934313523

> - `/gui/meshes/default/policies` <-- this would be the "policy type list" that is currently on the left
>  - `/gui/meshes/default/policies/:type` <-- this would be the "policy list" that is currently on the right
>
> If I follow correctly what the issue is, this would mean that every time you click a "policy type" on the left, the "policy" table on the right would completely re-render just through building this with nested routes rather than using things like `:key`.

PR #2250, opened 2024-03-01
chore(policies): refactor policies and fix zone links

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/2162

This PR was mainly a refactor of the PolicyList component to split it into nested routes. But whilst doing this I spotted several tiny bugs that I've fixed up here also, at least one of those bugs will require a fix on 2.6.x using the older version of the code.

A good deal of the green here is taking that code that was in PolicyList and splitting it into two separate routes.


src/app/policies/views/PolicyListView.vue
```diff
@@ -15,95 +15,222 @@
...
-                    })"
+                  <AppCollection
+                    class="policy-collection"
+                    data-testid="policy-collection"
+                    :empty-state-message="t('common.emptyState.message', { type: `${policyType.name} policies` })"
+                    :empty-state-cta-to="t('policies.href.docs', { name: policyType.name })"
+                    :empty-state-cta-text="t('common.documentation')"
+                    :headers="[
+                      { label: 'Name', key: 'name' },
+                      ...(can('use zones') && policyType.isTargetRefBased || policyType.name === 'MeshGateway' ? [{ label: 'Zone', key: 'zone' }] : []),
```
johncowen, line 91: Without this `can` the links that appear in this column might not resolve, this is the fix that will need to go to 2.6 as far as I understand it. Please shout if there is something I've missed here.

src/app/policies/views/PolicyTypeListView.vue
```diff
@@ -0,0 +1,161 @@
...
+                    v-slot="{ items }"
+                    :items="data!.policies"
+                  >
+                    <template
+                      v-for="legacy in [typeof meshInsight?.policies === 'undefined' ? items : items.filter(item => {
+                        // legacy policies are those that aren't targetRef, aren't MeshGateway and are also in use
+                        return !item.isTargetRefBased && item.name !== 'MeshGateway' && (meshInsight.policies?.[item.name]?.total ?? 0) > 0
+                      })]"
+                      :key="legacy"
+                    >
```
kleinfreund, line 56: **Suggestion**: This is the point in the template where you should introduce a component so that you can use computed properties of data source data instead of forcing you to use these hellish `<template v-for>` constructs to create local variables out of increasingly complex inline function calls.

It’s a problem that function calls are routinely being (re-)introduced in the template, because, as I pointed out quite a few times, it forces Vue to re-render the trees as Vue can’t know whether a function needs reevaluation — it has to call them again. That’s what computed properties are for. You should use them liberally.

The current PolicyList component does a good job of that. For one, it’s great that there’s a place where loading of data happens and one where all the presentational duties of the page are carried out. For another, it’s much easier to understand what happens in a view if one doesn’t have to digest virtually _all_ its logic in one gigantic template. Finally, and most importantly, it minimizes re-renders through the use of computed properties.

kleinfreund: It’s a suggestion.

src/app/policies/views/PolicyListView.vue
```diff
@@ -15,95 +15,222 @@
...
-          :src="`/*/policy-types`"
+      <DataCollection
+        v-slot="{ items: types }"
+        :predicate="(policyType) => typeof policyType !== 'undefined' && policyType.path === route.params.policyPath"
+        :find="true"
+        :items="props.policyTypes ?? [undefined]"
+      >
+        <template
+          v-for="policyType in [types[0]!]"
+          :key="policyType"
```
kleinfreund, line 26: **Suggestion**: This is a lot of unnecessary abstraction here. First of all, policy types can’t be empty. They can error, but they can’t be empty. Secondly, why are we using a Vue component for the task of finding an item in a list? A simple computed property does this job so much better. Besides which, using a component intended to show a collection of things with a empty block baked in for something that will 100% of times show one thing and not a collection is really weird as well.

src/app/policies/views/PolicyTypeListView.vue
```diff
@@ -0,0 +1,161 @@
...
+              class="policy-list-content"
+            >
+              <KCard
+                class="policy-type-list"
+                data-testid="policy-type-list"
+              >
+                <!-- block on policy types but not meshInsight -->
+                <DataLoader
+                  :data="[data]"
+                >
```
kleinfreund, line 44: **Question**: Is there a better way to do this? Using a DataLoader in a strictly not data loading capacity is extremely odd/unintuitive.

johncowen: Not sure if I follow. This is using DataLoader in a data loading capacity. It is blocking on policyTypes loading and showing a loading spinner inside the KCard until they are loaded.

---

Separately, one thing I noticed that I think we are missing here is the error state / `:errors-[error]`. I'll add that in shortly, ty!

src/app/policies/views/PolicyListView.vue
johncowen: > First of all, policy types can’t be empty.

They can be empty before they are loaded:

https://github.com/kumahq/kuma-gui/pull/2250/files#diff-d2633793eb7b7dee98cca8d616577a208c7eba06b7b8f983e4b7c0fce8986a52R104

> Secondly, why are we using a Vue component for the task of finding an item in a list?

Partly because I wanted an alternative to `<template v-for>` which you don't like, and partly because we get an automatic empty state with it.

> Besides which, using a component intended to show a collection of things with a empty block baked in for something that will 100% of times show one thing and not a collection is really weird as well.

See above, it can be empty before they are loaded.

> A simple computed property does this job so much better.

Computed properties don't have automatic empty states

src/app/policies/views/PolicyTypeListView.vue
johncowen: I added a commit to pass through the error to the DataLoader here. Let me know if we can resolve this one

johncowen:
Side note:

I wonder if this flake will get fixed by this refactor 🤔 , I've hit it a few times lately.

Screenshot 2024-03-06 at 11 43 49

src/app/policies/views/PolicyListView.vue
kleinfreund: > They can be empty before they are loaded:

They’re not empty then; they’re not loaded yet, and more specifically _are being loaded_. From a user perspective, you never expect to see an empty state for policy types.

> Partly because I wanted an alternative to `<template v-for>` which you don't like, and partly because we get an automatic empty state with it.

It looks like this neither needs an empty state nor is it able to avoid using the template v-for hack as one is following it immediately.

There are two pieces here why I think DataCollection is inadequate for what you’re trying to get to here:

- Passing in `[undefined]` when `props.policyTypes` is not set. Of course DataCollection expects an array of things. Providing it with `[undefined]` says both "I can’t provide it with a list of the right things right now" and "I'm still gonna provide it with a list of wrong things". This can only happen while still loading policy types, but there is a DataLoader that already covers the loading/empty scenario further up. That already enables you to expect a non-empty and set `props.policyTypes`. That’s handled in the current component in `master` so I don’t see how we can’t keep this convenience. You also have to guard against `undefined` in the `predicate` prop which is another signal that something is wrong here. The response from `/policies` can’t have `undefined` policy type definitions.
- There’s now a new instance of the `v-for` variable setting template hack with a type assertion on accessing `types[0]!`.

> Computed properties don't have automatic empty states

Of course not. The point? You can add an empty state if you need one (which you don’t here). That has nothing to do with computed properties.

---

So, from looking at this, neither does this part of the code need an empty state as that is already handled further up when loading the data nor does it need any of the patch work of selecting the current policy type. This component should be able to expect (1) a loaded set of `PolicyType`s and (2) the current `PolicyType` object as found from the current policy type path from the path segment. That’s an ideal use case for a computed property.

kleinfreund: Why do we want to show the empty state for an unknown policy type? The current behavior (showing an error) state is what I’d expect here.

kleinfreund: It’s expected that it does that. You manually typed in a URL to a policy type that we don’t know about (and that you can’t get to from anywhere in the UI) and you get a generic 404 error. That’s about to be expected.

Should we need to handle arbitrary policy types at some point due to some form of plug-and-play policy concept being introduced, many things will have to be rethought and this doesn’t help.

Seeing that the outfall of this is a bunch of convoluted logic in this component that we otherwise wouldn’t need (and actually probably even with it, we could handle this much better), I’m not okay with doing it that way.

To sum up what I expect to happen here:

- for unknown policy types, the view should show an error page as it does in `master`
- the PolicyListView component shouldn’t try and handle empty states and the like for policy type sourcing as that’s already done in PolicyTypeListView.

The rest of the PR looks more or less good to me even though I still kind of question why we need two different views for this. The separation of data-loading view and data-presenting component makes more sense to me.

johncowen: > It’s expected that it does that. 

🤔  I'd say its a bug. It gets even weirder when the CircuitBreaker item isn't in the left hand menu at all, and then it shows me an "About CircuitBreaker" intro text. Why do we show a piece of intro text for something that doesn't even exist? It's not that there aren't any, it's that that don't exist. As a user I'd be like "what is a circuit breaker and why are you showing me this when I asked for something entirely different?

> Should we need to handle arbitrary policy types at some point due to some form of plug-and-play policy concept being introduced, many things will have to be rethought and this doesn’t help.

As far as I am aware we have been handling these cases for the very very long time in the code. How come we shouldn't handle them here? We may aswell just pretend the list of policies is hardcoded in the code? (obvs I don't thin kwe should do that and hence this fix)

johncowen: btw is this a suggestion or a blocker?

src/app/policies/views/PolicyTypeListView.vue
johncowen: I'm gonna resolve this one, as I think its fine now, let me know if not and we can reopen.

src/app/policies/views/PolicyTypeListView.vue
johncowen: I took another look at this and found I could reuse the DataCollection for the filtering here, therefore removing one instance of `<template v-for>` and reducing the gigantic(?) template from 115 lines to 107 lines. I think I can resolve this one now but let me know if not.

src/app/policies/views/PolicyListView.vue
johncowen: So just to note what was said in the sync.

We are going to go with what is implemented in this PR, but we are going to change the "No Data" text to say something like "This policy doesn't exist"..

johncowen: I went with this:

Screenshot 2024-03-12 at 11 25 53

The text is also in i18n, so super easy to change if we want to. Will push this on the end now and resolve this conversation.

johncowen:
Thats all the conversations here resolved @kleinfreund let me know on this one. The only convo I've left open is my own inline which was to remind me to make a smaller PR to fix one of the issues covered here on `2.6.x`. I'll do that once this is merged.

src/app/policies/views/PolicyListView.vue
johncowen: I added one more commit here to move the custom text to the body rather than the title, that way it looks much more like everywhere else where we do this:

### Non existing policy type

Screenshot 2024-03-13 at 10 47 54

### Existing empty state for non existing policies

Screenshot 2024-03-13 at 10 47 45

kleinfreund (approved):

PR #2269, opened 2024-03-05
fix(policies): ensure legacy policies section has an empty state

johncowen:
Whilst playing around with kuma I spotted a bug when using non-federated/single-zone mode where we show a legacy policies section but we don't show an empty state, and instead its just a blank space which looks broken:

### Before:
Screenshot 2024-03-05 at 11 37 17

### After:

Screenshot 2024-03-05 at 11 39 56

I noticed that this was for "sidecar" dataplanes only, "gateway" dataplanes where showing an empty state, but I decided to add a KCard around the empty state in both places so there is a little change there also to make them consistent.

Theres a question about whether we should show the "Legacy Policies" block at all if its empty, but seeing as we do for gateway dataplanes I decided at least for the moment to continue to show it.

I stopped short of doing a bit more work here as I know there is work in progress that is altering this code for different reasons, but there is likely to be a follow up in this area later.

### DIFF
--- a/src/app/data-planes/components/PolicyTypeEntryList.vue
+++ b/src/app/data-planes/components/PolicyTypeEntryList.vue
@@ -4,7 +4,7 @@
     multiple-open
   >
     <AccordionItem
-      v-for="(policyTypeEntry, index) in legacyPolicyTypeEntries"
+      v-for="(policyTypeEntry, index) in items"
       :key="index"
     >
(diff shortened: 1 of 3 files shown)


johncowen:
This is another place where we could use `Object.groupBy` (https://github.com/kumahq/kuma-gui/pull/2287) but we can circle round later.

PR #2270, opened 2024-03-05
refactor: update Kongponents to KAlert/KEmptyState/KPagination milestone

kleinfreund:
**chore(KAlert): address breaking changes**

Use default slot instead of `alertMessage` slot.

Rename the `is-dismissibile` prop to `dismissible` (also remove two instances that didn’t have the corresponding close handling).

Rename `closed` event to `dismiss`.

Rename the `alertIcon` slot to `icon`.

Remove the `size` prop.

**chore(KPagination): address breaking changes**

**chore(KEmptyState): address breaking changes**

Replace the `cta-is-hidden` prop with `:action-button-visible="false"`.

Remove the use of the `icon` and `iconSize` prop. There is no longer a way to use the Kongponent "state" icons like `stateNoData` with KEmptyState so I removed it from our use of the component.

Replace the use of the `title` slot with providing the `title` prop. This required changing out ErrorBlock and EmptyBlock components to do the same. The only place where we really made use of slotting in not-just-text content was the ErrorBlock component where title and the badge list now appear on separate rows.

Replace the use of the `message` slot with using the default slot instead.

---

- https://github.com/kumahq/kuma-gui/pull/2184

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "dependencies": {
     "@kong-ui-public/i18n": "2.1.3",
     "@kong/icons": "1.8.14",
-    "@kong/kongponents": "9.0.0-alpha.107",
+    "@kong/kongponents": "9.0.0-alpha.113",
     "@vueuse/core": "10.9.0",
     "brandi": "5.0.0",
     "deepmerge": "4.3.1",
--- a/src/app/zones/views/ZoneCreateView.vue
+++ b/src/app/zones/views/ZoneCreateView.vue
@@ -295,49 +295,44 @@
                       data-testid="error"
                     />
 
-                    <KEmptyState
-                      v-else
-                      cta-is-hidden
-                    >
-                      <template #title>
-                        <template v-if="(typeof data === 'undefined')">
-                          <ProgressIcon
-                            data-testid="waiting"
-                            display="inline-block"
-                            :color="KUI_COLOR_TEXT_NEUTRAL_WEAK"
-                          />
-                          {{ t('zones.form.scan.waitTitle') }}
-                        </template>
-                        <template
+                    <KEmptyState v-else>
+                      <template #icon>
+                        <ProgressIcon
+                          v-if="data === undefined"
+                          data-testid="waiting"
+                          display="inline-block"
+                          :color="KUI_COLOR_TEXT_NEUTRAL_WEAK"
+                        />
+
+                        <CheckCircleIcon
                           v-else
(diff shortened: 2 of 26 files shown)


johncowen (commented):
I know this is still in draft for the moment, but just to say that this currently looks good to me.

package.json
```diff
@@ -124,5 +124,6 @@
       "./cypress/main.ts",
       "./cypress/support/step_definitions/*.ts"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.21+sha256.dbed5b7e10c552ba0e1a545c948d5473bc6c5a28ce22a8fd27e493e3e5eb6370"
```
johncowen, line 128: Did this sneak in with the yarn/pnpm investigation you've been doing?

kleinfreund: Argh, I need to address this in my local environment. I updated corepack and since this newest version, every single `yarn` command I run adds this (not just `yarn install`, `yarn dev`, too, etc.). I’ll remove it here.

Maybe we can add the `packageManager` entry though? It seems like there might be a bit of standardization happening here. In any case, it doesn’t belong here, so I’ll remove it.

kleinfreund: Done.

johncowen: No worries. Yeah definitely could add this later. I read `corepack` is potentially being moved to stable soon

src/app/zones/views/ZoneCreateView.vue
```diff
@@ -298,49 +295,44 @@
...
+                            :to="{
+                              name: 'zone-cp-detail-view',
+                              params: {
+                                zone: name,
+                              },
+                            }"
+                          >
+                            {{ t('zones.form.scan.completeButtonLabel', { name }) }}
+                          </KButton>
+                        </p>
```
johncowen, line 335: Apart from removing the `cta-is-hidden`attribute and the switch from `#message` to `#default`, does this make any other changes?

kleinfreund: Yes: This is now using the `icon` slot instead of adding the icon to the title.

johncowen (approved):
One thing I noticed is that certain nodes are being rendered even if they are empty, specifically https://github.com/Kong/kongponents/blob/c9dd37b7a7a9360aaac3e8695fc8f6eeab4c601c/src/components/KEmptyState/KEmptyState.vue#L27

I think thats an upstream issue tho.

PR #2272, opened 2024-03-05
chore: removes read usage of kumaApi

johncowen:
Generally our onboarding area is pretty frozen, but it still has older code that we have to continue supporting unless we get rid of it (in this case `kumaApi` usage vs DataSource/DataLoader).

I had a bit of time whilst waiting on other PRs so, this PR switches the remaining areas of onboarding to use DataSource/DataLoader, and in doing so:

1. Removes all the older imperative code from onboarding and replaces it with DataSource
2. Changes some places where we were looking for a count of at least one zone/dataplane/zone-ingress and instead looks for at least a count of one _and_ one is online.
3. Uses the newer `_overview` endpoints for gathering data
4. Removes the last occurrences of us using `useKumaApi()` / `kumaApi` for reading data, which means we are closer to being able to use an OpenAPI spec auto-generated typescript http client.
5. I noticed that the URL for the documentation link was wrong in multizone onboarding, so I corrected that with a little more i18n markdown-ing while I was here. This then led to being unable to use `v-html` and `class` on the same element due to a clash with eslint rules (`class` has to be above `v-html` which means you can't disable the `v-html` on the next line because an eslint fix that reorders `class` and `v-html` means its no longer on the next line). Due to this I disabled the no-v-html rule entirely. Thinking being, we need to/will need to use this a lot. That being said, `v-html` should only be used for i18n purposes.

I might add some inlines as this PR snowballed into something bigger than I originally imagined.

---

An important note here. Ideally our DataSource URIs would be split into using plural-name/singular-name i.e. `/zone-cps` and `/zone-cp/:name`. This means we can safely add 'actions' on collections of things without worrying about clashing with the `:name` part of a singular. 

For example, in the case of this PR, I very easily clashed with `/zone-cps/online` with `/zone-cps/:name` without realising. If we used `/zone-cp/:name` instead (i.e. the singular for the singular request) this wouldn't be an issue. Also importantly, if we ever need to perform an 'action' request on a singular item the URI would be `/zone-cp/:name/:action` meaning we again avoid the clash.

To temporarily work around this issue in this PR I've used the 'DNS trick' `/zone-cps/~online` seeing as we can't use the `~` character in the name of a zone-cp and ordered the source to come before `/zone-cps/:name` therefore avoiding a potential clash. But we should change all our source URIs to have plural/singular forms.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -96,6 +96,7 @@ const INLINE_NON_VOID_ELEMENTS = [
           ignoreWhenEmpty: true,
           ignores: ['router-link', 'pre', ...INLINE_NON_VOID_ELEMENTS],
         }],
+        'vue/no-v-html': 'off',
         // Reason: https://github.com/vuejs/eslint-plugin-vue/issues/2259
         'vue/no-setup-props-destructure': 'off',
         // Disallow Kongponents utility classes
--- a/features/onboarding/add-new-services-code/Index.feature
+++ b/features/onboarding/add-new-services-code/Index.feature
@@ -1,4 +1,4 @@
-Feature: onboarding / dataplanes-overview / index
+Feature: onboarding / add-new-services-code / index
   Background:
     Given the CSS selectors
       | Alias           | Selector                               |
@@ -20,11 +20,15 @@ Feature: onboarding / dataplanes-overview / index
       """
       KUMA_DATAPLANE_COUNT: 1
       """
-    And the URL "/dataplanes" responds with
+    And the URL "/dataplanes/_overview" responds with
       """
       body:
         items:
           - name: dataplane-test
+            dataplaneInsight:
+              subscriptions:
+                - connectTime: 2021-02-17T07:33:36.412683Z
+                  disconnectTime: !!js/undefined
       """
     Then the "$loading" element doesn't exist
     And the "$is-connected" element exists
--- a/features/onboarding/multi-zone/Index.feature
+++ b/features/onboarding/multi-zone/Index.feature
@@ -1,32 +1,48 @@
 Feature: onboarding / multi-zone / index
   Background:
     Given the CSS selectors
-      | Alias             | Selector                               |
-      | next-button       | [data-testid='onboarding-next-button'] |
-      | loading           | [data-testid='loading']                |
-      | zone-connected    | [data-testid='zone-connected']         |
-      | ingress-connected | [data-testid='zone-ingress-connected'] |
+      | Alias       | Selector                                   |
+      | next-button | [data-testid='onboarding-next-button']     |
+      | loading     | [data-testid='loading']                    |
+      | connected   | [data-testid='zone-online-ingress-online'] |
 
   Scenario: Loading the zone list and showing the state
     Given the environment
       """
       KUMA_ZONE_COUNT: 0
+      KUMA_ZONEINGRESS_COUNT: 0
       """
     When I visit the "/onboarding/multi-zone" URL
     Then the "$loading" element exists
-    And the "$zone-connected" element doesn't exist
+    And the "$connected" element doesn't exist
     And the "$next-button[disabled]" element exists
     And the environment
       """
       KUMA_ZONE_COUNT: 1
+      KUMA_ZONEINGRESS_COUNT: 1
       """
-    And the URL "/zones" responds with
+    And the URL "/zones/_overview" responds with
       """
       body:
         items:
           - name: zone-test
+            zone:
+              enabled: true
+            zoneInsight:
+              subscriptions:
+                - connectTime: 2021-02-17T07:33:36.412683Z
+                  disconnectTime: !!js/undefined
+      """
+    And the URL "/zone-ingresses/_overview" responds with
+      """
+      body:
+        items:
+          - name: zone-ingress-test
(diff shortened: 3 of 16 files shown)


.eslintrc.cjs
```diff
@@ -96,6 +96,7 @@ const INLINE_NON_VOID_ELEMENTS = [
           ignoreWhenEmpty: true,
           ignores: ['router-link', 'pre', ...INLINE_NON_VOID_ELEMENTS],
         }],
+        'vue/no-v-html': 'off',
```
johncowen, line 99: If there is a suggestion for an alternative to doing this without it being annoying it would be much appreciated.

features/onboarding/add-new-services-code/Index.feature
```diff
@@ -20,11 +20,15 @@ Feature: onboarding / dataplanes-overview / index
       """
       KUMA_DATAPLANE_COUNT: 1
       """
-    And the URL "/dataplanes" responds with
+    And the URL "/dataplanes/_overview" responds with
```
johncowen, line 23: We now check `_overview` for counts of things, I think this is fine, but let me know if not.

features/onboarding/multi-zone/Index.feature
```diff
@@ -1,32 +1,48 @@
...
-    And the "$zone-connected" element doesn't exist
+    And the "$connected" element doesn't exist
     And the "$next-button[disabled]" element exists
     And the environment
       """
       KUMA_ZONE_COUNT: 1
+      KUMA_ZONEINGRESS_COUNT: 1
       """
-    And the URL "/zones" responds with
+    And the URL "/zones/_overview" responds with
```
johncowen, line 24: Again we use `/_overview` here now

src/app/onboarding/components/OnboardingAlert.vue
```diff
@@ -8,8 +8,9 @@
   >
     <template #alertMessage>
       <div class="onboarding-alert-content">
-        <!-- eslint-disable-next-line vue/no-v-html  -->
-        <div v-html="t('main-overview.detail.onboarding.message', { name: t('common.product.name') })" />
+        <div
+          v-html="t('main-overview.detail.onboarding.message', { name: t('common.product.name') })"
+        />
```
johncowen, line 13: I just removed the comment from here seeing as we disable it globally.

src/app/onboarding/locales/en-us/index.yaml
```diff
@@ -11,12 +8,30 @@ onboarding:
...
+        Zone status: {zone, select,
+          online { *Connected* }
+          offline { **Disconnected** }
+          other { - }}
+
+        Zone ingress status: {ingress, select,
+          online { *Connected* }
+          offline { **Disconnected** }
+          other { - }}
+
```
johncowen, line 25: Corrected the docs URL, and i18n'ed the rest while I was here.

src/app/onboarding/views/OnboardingAddNewServicesCodeView.vue
```diff
@@ -8,135 +8,106 @@
...
+
+          <template #navigation>
+            <OnboardingNavigation
+              next-step="onboarding-dataplanes-view"
+              previous-step="onboarding-add-new-services-view"
+              :should-allow-next="typeof data !== 'undefined'"
+            />
+          </template>
+        </OnboardingPage>
+      </DataSource>
```
johncowen, line 99: This is mostly a re-nest due to the DataSource

src/app/onboarding/views/OnboardingDataplanesView.vue
```diff
@@ -9,114 +9,103 @@
...
+                    class="mb-4"
+                    data-testid="dataplanes-table"
+                    :fetcher-cache-key="JSON.stringify(data)"
+                    :fetcher="((data) => {
+                      return () => {
+                        return {
+                          data: data.items,
+                          total: data.items.length,
+                        }
+                      }})(data)"
```
johncowen, line 70: I noticed that `KTable::fetcher` isn't reactive, i.e. if i change the value of `:fetcher` `KTable` will still use the initial one, which I guess is probably a bug. Not totally sure i might file an issue. Wdyt? Should this be considered a bug?

---

I left in the function factory incase it ever gets fixed

johncowen (commented):
Inlines:

slonka:
@johncowen do you need any help to move this forward? it was stuck for 3 weeks

johncowen:
Yeah it would be good to get a review on it. I think most of the change is just re-using patterns from what we have elsewhere so there shouldn't be anything to controversial.

I think maybe the only places I was specifically hoping for a second pair of eyes was these things from the PR desc:

> 2. Changes some places where we were looking for a count of at least one zone/dataplane/zone-ingress and instead looks for at least a count of one and one is online.
> 3. Uses the newer _overview endpoints for gathering data

Just an additional note incase, there is nothing urgent here and nothing here is needed for 2.7, but in the same breath it would be good to get this merged, otherwise we may as well close it (which would be a shame as it is valuable seeing as we want to get to a OpenAPI generated HTTP client)

If you want to take a look yourself lemme know if I can help further with anything

kleinfreund (approved):

johncowen:
Rebased this one also, will hold off merge for a little while incase, ty!

PR #2273, opened 2024-03-06
chore: adds XAction component

johncowen:
Another little PR I had in mind whilst I'm awaiting reviews of others:

Adds an eXtra eXtension to `RouterLink` in the form of an `XAction` component which can be used instead of `RouterLink` (we can eventually take that out of the global scope)

Mainly to paper over the awkward `inactive: route.params.inactive ? null : undefined` things (I included one example of this here) which really should just be `inactive: true | false`. But also adds dynamic node choosing based on existence `href`/`to` properties and I'm sure there'll be more things we might want to add here in future.

### DIFF
--- a/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -203,7 +203,7 @@
                                 }
                               "
                             >
-                              <RouterLink
+                              <XAction
                                 data-action
                                 :to="{
                                   name: ((name) => name.includes('bound') ? name.replace('-outbound-', '-inbound-') : 'connection-inbound-summary-overview-view')(String(_route.name)),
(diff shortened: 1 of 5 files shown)


slonka:
@johncowen same as the other PR - anything we can help with?

johncowen:
Yep defo, feel free to review if you're ok with that. This one is pretty straight forwards codewise I think. The majority of it is converting `true`/`false` to `null` / `undefined` but instead of doing that in the application code we hide it away in a component.

Separately, there is also code to decide whether to use a `<a>` or a `<button>` depending on whether you have passed a `href/to` or not. If we have a URL/href the element should always be a `<a>` if we dont' it should always be a `<button>`. Again we hide this away behind a `<XAction />` component so as an application dev I don't need to think about it, or potentially get that wrong.

We do this conversion so we get `/url?inactive` and `/url` instead of `/url?inactive=true` and `/url?inactive=false` in the browser location bar.

Let me know if there is anything else I can help with

(oh also if its better to rebase these before review lemme know, I usually leave them until during or after review to avoid rebasing everyday as the days go by with these PRs sitting here)

johncowen:
ty! This one needed a rebase, so I'll leave hanging for a little while incase, but then go ahead and merge

PR #2274, opened 2024-03-07
fix link to GUI docs

jakubdyszkiewicz:
Fix the link


johncowen (approved):
😮 thanks!

I found an old link that is now 404ing yesterday also, I think we might want to add some sort of link checking in our repo 🤔

PR #2276, opened 2024-03-08
refactor(dataplanes): moves dataplane viz to pick inbounds based on IP

johncowen:
TBD


kleinfreund (approved):

johncowen:
@kleinfreund I add 3 more tiny commits on the end here, I'll merge shortly if I hear nothing back on those.

Worth looking at as I'm using the "tagged anchor" thing here, plus an annoying re-addition of `inline-block` on some icons, which I would like to avoid having to deal with.

PR #2277, opened 2024-03-08
feature(dataplanes): add inbound rules to the summary panel

johncowen:
Note: This PR may or may not change depending when https://github.com/kumahq/kuma-gui/pull/2276 goes in, as I think https://github.com/kumahq/kuma-gui/pull/2276 gives lets us expand on this easier.

This PR adds the policy rules for standard dataplane inbounds to the summary panel:

Screenshot 2024-03-08 at 13 51 06

Please let me know if there are any other suggestions for presenting this visually.

Similar visuals for policy rules for gateway inbounds are still to come (see above note on https://github.com/kumahq/kuma-gui/pull/2276), as are rules for when you click on outbounds (which I think will need separate information from the service insight endpoint which we don't request as yet)

### DIFF
--- a/src/app/connections/views/ConnectionInboundSummaryOverviewView.vue
+++ b/src/app/connections/views/ConnectionInboundSummaryOverviewView.vue
@@ -1,101 +1,143 @@
 <template>
   <RouteView
-    v-slot="{ t }"
+    v-slot="{ route, t }"
     name="connection-inbound-summary-overview-view"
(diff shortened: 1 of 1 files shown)


johncowen:
I've been working on adding more data to this one such as `matchers` and `origins`. I'm gonna merge this and follow up pretty quick rather than add it on the end here.

PR #2287, opened 2024-03-12
chore: Object.groupBy polyfill

johncowen:
Now Typescript comes with built-in declarations for `Object.groupBy` its a little less annoying to polyfill use Javascript's `Object.groupBy`, also see the description in https://github.com/kumahq/kuma-gui/pull/2128.

This PR adds a polyfill and uses it an a couple of places where conversations have arisen lately. most of the change here is as a result of a re-nest, but the code that came up in past conversations is now just a one-liner.

There are a few places I'll be switching this in at some point (including the code in above linked PR)

I tested the polyfill manually in Firefox 118.

### DIFF
--- a/src/app/data-planes/components/RuleEntryList.vue
+++ b/src/app/data-planes/components/RuleEntryList.vue
@@ -4,116 +4,114 @@
     multiple-open
   >
     <template
-      v-for="policies in [props.rules.reduce<Record<string, Rule[]>>((prev, item) => {
-        if(typeof prev[item.type] === 'undefined') {
-          prev[item.type] = []
-        }
-        prev[item.type].push(item)
-        return prev
-      }, {})]"
+      v-for="policies in [Object.groupBy(props.rules, (item, i) => item.type)]"
       :key="policies"
     >
-      <AccordionItem
+      <template
         v-for="(items, type) in policies"
         :key="type"
       >
-        <template #accordion-header>
-          <h3 class="policy-type-heading">
-            <PolicyTypeTag :policy-type="type">
-              {{ type }}
-            </PolicyTypeTag>
-          </h3>
-        </template>
-
-        <template #accordion-content>
-          <template
-            v-for="hasMatchers in [items.some((item) => item.matchers.length > 0)]"
-            :key="hasMatchers"
-          >
-            <div class="policy-list">
-              <AppCollection
-                class="policy-type-table"
-                :class="{
-                  'has-matchers': hasMatchers,
-                }"
-                :total="items.length"
-                :items="items"
-                :headers="[
-                  ...(hasMatchers ? [{ label: 'Matchers', key: 'matchers' }] : []),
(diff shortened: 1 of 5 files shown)


src/app/data-planes/components/StandardDataplanePolicies.vue
```diff
@@ -73,7 +65,7 @@
 
             <RuleEntryList
               class="mt-2"
-              :rules="rs"
+              :rules="rs!"
```
kleinfreund, line 68: **Question**: How come this is needed now? Shouldn’t this change be a simple replacement of "groupBy" code that has no other impact?

src/app/data-planes/components/RuleEntryList.vue
```diff
@@ -4,116 +4,114 @@
...
-                :headers="[
-                  ...(hasMatchers ? [{ label: 'Matchers', key: 'matchers' }] : []),
-                  { label: 'Origin policies', key: 'origins' },
-                  { label: 'Conf', key: 'config' },
-                ]"
-              >
-                <template
-                  #matchers="{ row }"
+        <AccordionItem
+          v-if="items"
```
kleinfreund, line 15: **Question**: How come this is change was needed/made? Shouldn’t this change be a simple replacement of "groupBy" code that has no other impact?

src/app/data-planes/components/StandardDataplanePolicies.vue
johncowen: Looks like the Typescript definition can return `[] | undefined` instead of `[]` for a 'group' which as far as I understand it can never happen. The key just isn't there if there is no group and therefore doesn't get looped over. Actually without looking back at it properly I'm not sure if this is the Typescript `groupBy` definition or Vue's for loop.

src/app/data-planes/components/RuleEntryList.vue
johncowen: I wrapped the AccordianThing in a further `template` because I unfortunately also now need to do a v-if="" otherwise TS complains, I think this is the same issue as our conversation above just approached differently.

src/app/data-planes/components/StandardDataplanePolicies.vue
kleinfreund: FYI I dug around to get some clarity on this and it seems appropriate (although somewhat annoying) that the result of `Object.groupBy` is a `Partial<...>`, see https://github.com/microsoft/TypeScript/issues/47171#issuecomment-1848451195

src/app/data-planes/components/RuleEntryList.vue
kleinfreund: See also https://github.com/kumahq/kuma-gui/pull/2287#discussion_r1524855069

kleinfreund (approved):

johncowen:
FYI I'm considering add `:group-by` to DataCollection

PR #2295, opened 2024-03-14
chore(zones): adds modal layer and uses it in zones

johncowen:
This PR started out as something smaller, but again grew as I found things:

1. Adds a `modal-layer` to attach all our modals to at an Application-wide level (in ApplicationShell)
2. Changes the modals we have here to use the new `modal-layer`
3. Uses `KToggle` to control modal open/close state.
4. `KToggle` unfortunately has a little problem with it in that it exposes `isToggled`, but its not a `boolean` as you would expect. Its a Vue `ref` and hence you need to access `isToggled.value` from within a Vue template, which is strange. I made `XDisclosure` to paper over this (I called it XDisclosure because I wanted something that folks didn't think was a 'toggle button'). Seeing as its a eXtra eXtension to KToggle I namespaced it as such.
5. ~Fixed what I think is a tiny bug in deciding whether to show the Zone creation button or not (I'll note inline)~
6. Made the Zone creation button use our new `AppView::actions` `XTeleportSlot` which simplifies things slightly.

Possibly another couple of things, but I stopped here as I want to return to return to other work.


slonka:
@johncowen - anything we can help with here? should I do a review? This one does not have conflicts.

johncowen:
Yeah same again, feel free to take alook. Umm I think the main thing with this one was making sure our modals are always on the equivalent of the highest `z-index` no matter where we use them. (we do this using a Teleport: deeper dive if it helps https://vuejs.org/guide/built-ins/teleport.html)

kleinfreund (approved):

Issue #2303, opened 2024-03-15
Clusters tabs: show empty state if there’s no clusters data


kleinfreund:
### What happened?

See title

PR #2304, opened 2024-03-15
chore: moves connection stats/clusters to use DataLoader/DataCollection

johncowen:
Moves stats/clusters  to use `DataLoader` and `DataCollection`. DataCollection has an "auto empty state", so if no matching lines are found, then instead of getting an empty code editor we can an empty state instead.

I also changed the URI for the DataLoaders to use the `.../stats/ip-address` here so things synced up correctly. This also meant that I had to re-nest some `template`s in `DataPlaneDetailView`. Therefore it looks like there is loads of change in that file when actually its pretty much just a re-nest, there is one tweak in that file but I'll inline comment to point that out.

Closes https://github.com/kumahq/kuma-gui/issues/2303

### DIFF
--- a/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -12,148 +12,148 @@
       v-slot="{ data: traffic, error, refresh }: StatsSource"
       :src="`/meshes/${route.params.mesh}/dataplanes/${route.params.dataPlane}/stats/${props.data.dataplane.networking.address}`"
     >
-      <!-- if we are a builtin gateway proxy i.e. a 'gateway' proxy -->
-      <!-- use its first and only inbounds as a template  -->
-      <!-- and fill the inbounds out from the stats once they are loaded -->
-      <template
-        v-for="inbounds in [props.data.dataplane.networking.type === 'gateway' ? Object.entries<any>(traffic?.inbounds ?? {}).reduce<DataplaneInbound[]>((prev, [key, value]) => {
-          const [ip, port] = key.split('_')
-          // As we are just 'finding' inbounds from stats without knowing them
-          // from the dataplane overview API request, we will wrongly 'find'
-          // the envoy admin inbound that is used for kumas /stats API. Ignore
-          // the envoy admin inbound when we find it
-          if(port === (props.data.dataplane.networking.admin?.port ?? '9901')) {
-            return prev
-          }
-          return prev.concat([
-            {
-              ...props.data.dataplane.networking.inbounds[0],
-              name: key,
-              port: Number(port),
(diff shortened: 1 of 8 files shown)


src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -12,148 +12,148 @@
       v-slot="{ data: traffic, error, refresh }: StatsSource"
       :src="`/meshes/${route.params.mesh}/dataplanes/${route.params.dataPlane}/stats/${props.data.dataplane.networking.address}`"
     >
-      <!-- if we are a builtin gateway proxy i.e. a 'gateway' proxy -->
-      <!-- use its first and only inbounds as a template  -->
-      <!-- and fill the inbounds out from the stats once they are loaded -->
-      <template
-        v-for="inbounds in [props.data.dataplane.networking.type === 'gateway' ? Object.entries<any>(traffic?.inbounds ?? {}).reduce<DataplaneInbound[]>((prev, [key, value]) => {
-          const [ip, port] = key.split('_')
```
johncowen, line 20: This is the `<template />` I moved deeper in

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -12,148 +12,148 @@
...
-                    type="inbound"
+                <ConnectionGroup
+                  type="inbound"
+                >
+                  <!-- if we are a builtin gateway proxy i.e. a 'gateway' proxy -->
+                  <!-- use its first and only inbounds as a template  -->
+                  <!-- and fill the inbounds out from the stats once they are loaded -->
+                  <template
+                    v-for="inbounds in [props.data.dataplane.networking.type === 'gateway' ? Object.entries<any>(traffic?.inbounds ?? {}).reduce<DataplaneInbound[]>((prev, [key, value]) => {
+                      const [ip, port] = key.split('_')
```
johncowen, line 138: The `<template />` mentioned is now here

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -213,239 +213,240 @@
...
-                <component
-                  :is="child.Component"
-                  :data="(child.route.name as string).includes('-inbound-') ? inbounds : traffic?.outbounds || {}"
-                />
-              </SummaryView>
-            </RouterView>
+              <component
+                :is="child.Component"
+                :data="(child.route.name as string).includes('-inbound-') ? props.data.dataplane.networking.inbounds : traffic?.outbounds || {}"
+                :dataplane-overview="props.data"
```
johncowen, line 343: As a consequence of moving the template I also tweaked this line, and added the `:dataplane-overview` which I then use in the sub views.

johncowen (commented):
Inlines:

lukidzi (approved):

Issue #2251, opened 2024-03-01
Add possibility to move to the outbound service in dp view


lahabana:
### Description

I'm tracking 5xx requests and I've got:

<img width="1450" alt="image" src="https://github.com/kumahq/kuma-gui/assets/702355/f7362810-d2a9-49fb-b985-fbb065b571d4">

Ideally I'd like to be able to get to the dataplanes of the api-play-001 service

Issue #2124, opened 2024-01-31
Add matching policies in tray of dataplane view


lahabana:
### Description

This is super useful to actually understand what config you have on an inbound or an outbound

johncowen:
PRs addressing this:

- https://github.com/kumahq/kuma-gui/pull/2277
- https://github.com/kumahq/kuma-gui/pull/2289

Those are just for inbounds on sidecars, so we might want to split this issue up into further issues, instead of of a longer running umbrella issue.

johncowen:
Closed by https://github.com/kumahq/kuma-gui/pull/2315

There are more PRs to come here that do more minor things such as group the rules by policy type (https://github.com/kumahq/kuma-gui/pull/2322) plus some more design related PRs, but the top-level task here is complete.

PR #2315, opened 2024-03-19
chore(dataplanes): various dataplane viz additions

johncowen:
### Outbound service links

Outbound card now have a link to their respective service so you can click through to a list of dataplanes of the outbound and then click into the viz for the outbound (which is shows you it as an inbound). This means you can keep clicking through all hops, from start to finish. We infer the service name by removing the `-hash` from the end of the outbound (if that exists)

Screenshot 2024-03-20 at 11 34 02

### Outbound (currently ungrouped) rules display 

Adds outbound rules to the drawer. We filter out MeshHTTP/TCPRoutes from here, and then do simple simple filtering based on the `matchers` to figure out whether we should show certain rules or not.

Screenshot 2024-03-20 at 11 36 18

There are a few small fixes to the data re-shaping here and I also changed the random mocks so they link together better (across `_overview`, `stats` and `_rules`).

Lastly I added a bunch of static fixtures so its easier to test against something realistic for this. It's far too easy to mock out random data that might not actually happen. On top of these I added some straight forward e2e tests to give a bit of confidence that things are continuing to display properly.

---

- Part of https://github.com/kumahq/kuma-gui/issues/2124
- Closes https://github.com/kumahq/kuma-gui/issues/2251
- Has https://github.com/kumahq/kuma-gui/pull/2318 applied to avoid the flakey test.


src/app/data-planes/data/index.ts
```diff
@@ -76,7 +77,19 @@ export type MeshGatewayDataplane = PartialMeshGatewayDataplane & {
   listenerEntries: MeshGatewayListenerEntry[]
   routePolicies: Meta[]
 }
-
+const DataplaneOutbound = {
+  fromObject(item: PartialDataplaneOutbound): DataplaneOutbound {
+    return {
+      ...item,
+      name: item.tags['kuma.io/service'],
```
kleinfreund, line 84: **Question**: Why do we need `name` _and_ `service`?

johncowen: We don't use this right now but in the same way that we fill out what we know at this point and append it later for inbounds this would do the same. See for gateway inbounds, we use the name of the service for the name and then once we know they port (later on in the `*View.vue`) we append that onto the end of the name.

kleinfreund: But why do we need two properties with the same value?

johncowen: Its not the same value, in some cases we append the port to it later.

kleinfreund: Where does this happen? And do we actually mutate the DataplaneOverview object for that or is it just happening in place somewhere?

kleinfreund (approved):

PR #2316, opened 2024-03-19
chore: refactors service unit tests to use our test utils

johncowen:
Moves our `service` unit tests to use our less verbose test utils

(This has a tmp commit applied to sort a flake-y test that I PRed separately here https://github.com/kumahq/kuma-gui/pull/2318)


features/mesh/dataplanes/DataplanePolicies.feature
```diff
@@ -410,24 +417,6 @@ Feature: Dataplane policies
...
-                          kuma.io/service: demo-app_kuma-demo_svc_5000
-                        policies:
-                          CircuitBreaker:
-                            name: circuit-breaker-1
-          policies:
-            TrafficLog:
-              name: traffic-log-1
-            TrafficTrace:
-              name: traffic-trace-1
-        """
```
kleinfreund, line 430: **Question**: Are you sure this can be removed? The endpoint seems to be called in this test and I’ve recently seen this test flake (though I haven’t looked into this).

(I’m aware that this only shows here because the PR isn’t rebased and that this mock was actually already removed)

johncowen: I merged master up into here so this has now gone from this PR.

The test was flaking before I removed this, and the PR I added to change this including an amend to fix the flake.

Nothing from this mock is being asserted on, apart from the mock having an amount of data).

Resolving this seeing as its not in this PR.

kleinfreund (approved):

PR #2318, opened 2024-03-19
fix(tests): unflake built-in gateway policy test

johncowen:
I've spotted a couple of new flakes lately, this test one of them.

The mock here mocked out a HTTP endpoint that didn't seem to be being called at all. So I've removed that and added some `_COUNT` env vars that should always give us the data we are looking for the test.

This seems to stabilise this test, but there are still a couple of others that have popped up. Hopefully I get figure those out after this one which seems to be the most flakey.

I also applied this here https://github.com/kumahq/kuma-gui/pull/2316 after this test flaking multiple times there, and its seems to have fixed the flake.

### DIFF
--- a/features/mesh/dataplanes/DataplanePolicies.feature
+++ b/features/mesh/dataplanes/DataplanePolicies.feature
@@ -399,6 +399,13 @@ Feature: Dataplane policies
         """
         KUMA_MODE: zone
         """
+      Given the environment
+        """
+        KUMA_DATAPLANE_PROXY_RULE_ENABLED: true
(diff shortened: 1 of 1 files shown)


lukidzi (approved):

kleinfreund:
> The mock here mocked out a HTTP endpoint that didn't seem to be being called at all.

Just to note, I don’t think that’s correct. I see this API being called locally in this view in zone mode.

johncowen:
Ah thanks, lemme double check this.

PR #2319, opened 2024-03-21
chore: amends to app wide headers css

johncowen:
Reduces the size of our headers slightly. 

The following screenshot shows the majority of the affected headers:

Screenshot 2024-03-21 at 10 17 35

~There are few semi-hacks here such as minus margins, but its to keep the certain buttons inline whilst keeping margins the same whether the button appears or not, and is probably the best way to do this (hence only a semi-hack)~

### DIFF
--- a/src/app/gateways/views/GatewayListTabsView.vue
+++ b/src/app/gateways/views/GatewayListTabsView.vue
@@ -48,9 +48,3 @@
 <script lang="ts" setup>
 import LinkBox from '@/app/common/LinkBox.vue'
 </script>
-<style lang="scss" scoped>
-:deep(.link-box) {
-  margin-top: -10px;
-  margin-bottom: -8px;
-}
-</style>
--- a/src/app/gateways/views/GatewayListTabsView.vue
+++ b/src/app/gateways/views/GatewayListTabsView.vue
@@ -48,3 +48,9 @@
 <script lang="ts" setup>
 import LinkBox from '@/app/common/LinkBox.vue'
 </script>
+<style lang="scss" scoped>
+:deep(.link-box) {
+  margin-top: -10px;
+  margin-bottom: -8px;
+}
+</style>
--- a/features/mesh/dataplanes/DataplanePolicies.feature
+++ b/features/mesh/dataplanes/DataplanePolicies.feature
@@ -399,6 +399,13 @@ Feature: Dataplane policies
         """
         KUMA_MODE: zone
         """
+      Given the environment
+        """
+        KUMA_DATAPLANE_PROXY_RULE_ENABLED: true
+        KUMA_DATAPLANE_RULE_COUNT: 1
+        KUMA_DATAPLANE_TO_RULE_COUNT: 1
+        KUMA_DATAPLANE_FROM_RULE_COUNT: 1
+        """
       And the URL "/meshes/default/dataplanes/dataplane-gateway_builtin-1/_overview" responds with
         """
         body:
@@ -409,24 +416,6 @@ Feature: Dataplane policies
                 type: BUILTIN
                 tags:
                   kuma.io/service: service-1
-        """
-      And the URL "/meshes/default/dataplanes/dataplane-gateway_builtin-1/policies" responds with
-        """
-        body:
-          listeners:
-            - hosts:
-                - routes:
-                  - destinations:
-                      - tags:
-                          kuma.io/service: demo-app_kuma-demo_svc_5000
-                        policies:
-                          CircuitBreaker:
-                            name: circuit-breaker-1
-          policies:
-            TrafficLog:
-              name: traffic-log-1
-            TrafficTrace:
-              name: traffic-trace-1
         """
 
       When I visit the "/meshes/default/data-planes/dataplane-gateway_builtin-1/policies" URL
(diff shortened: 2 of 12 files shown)


src/app/gateways/views/GatewayListTabsView.vue
```diff
@@ -48,3 +48,9 @@
 <script lang="ts" setup>
 import LinkBox from '@/app/common/LinkBox.vue'
 </script>
+<style lang="scss" scoped>
+:deep(.link-box) {
+  margin-top: -10px;
+  margin-bottom: -8px;
+}
```
kleinfreund, line 55: **Issue**: We should really, really not use styles like this for general alignment matters. There is no reason the vertical margins of the targeted element here should be pulled in like that. This should be removed everywhere where it was added in this PR.

kleinfreund (changes requested):

src/app/gateways/views/GatewayListTabsView.vue
johncowen: I mention this in the desc:

> There are few semi-hacks here such as minus margins, but its to keep the certain buttons inline whilst keeping margins the same whether the button appears or not, and is probably the best way to do this (hence only a semi-hack)

Happy to use another approach to keep the margins the same whether there is a button or not. Previous to this PR the code that was there meant that the margins were out here:

Screenshot 2024-03-25 at 09 14 03

Let me know if you have a different approach that works.

johncowen: Oh this specific one was for when we show the linkbox and when we don't. Because our headers are now much smaller, having a linkbox conditionally show means that the margins can look wrong when we have a linkbox.

With this change in either case now the margins for the header always the same.

kleinfreund: Using negative margins to artificially reduce the height of the button in the actions element just so that the title bar is the same height as if the button wasn’t there is a pretty bad approach to make the title bars the same height:

- It’s not accurate. In this PR, the title bar (i.e. the `.card-header` element) without button has a height of 21 pixels. The title bar with a button has a height of 21.9 pixels. So they’re not actually aligned.
- It looks visually bad. The button now clips into the KCard padding which it really shouldn’t.
- It’s brittle because it depends on the height of the buttons. If the button height changes, the offset values will need to be adjusted.

There are two other strategies to handle this:

- Fixed height. Making the title bar the same height everywhere so that horizontally laid out elements like on our overview page align with their content. Easy to implement. The only downside is that it, too, might need to be manually adjusted if the height of the action items or headings changes. But at least it only needs to be done in one place. Strictly better than the negative margin approach.
- Not to bother. Instead, let content be laid naturally. Even easier to "implement" (it needs no implementation, just code removal). The content in the overview page wouldn’t be aligned, but that’s alright in my opinion.

johncowen: > It’s not accurate.

Are you saying its 0.1 pixel out?

> It looks visually bad.

Could you screengrab what you are seeing?

> It’s brittle because it depends on the height of the buttons. If the button height changes, the offset values will need to be adjusted.

Agreed, but thats not a huge issue I don't think.

> There are two other strategies to handle this:

Both of these have their problems, and most notably the second one doesn't solve the issue. We want these things to be aligned and to be consistent with all our other headers.

If you could pop up a screengrab of what you are looking at that looks visually bad it would help (also to confirm exactly which one of these instances we are talking about) that'd be great! ty

kleinfreund: > Are you saying its 0.1 pixel out?

0.9 pixels.

> Could you screengrab what you are seeing?

image

The top padding is too short.

johncowen: Looks spot on to me?

Screenshot 2024-03-25 at 12 22 47

Which is the bit that looks visually bad?

johncowen: I think everything should line up.

I think this stuff you are talking about "clipping the padding" is quite a common thing in user interfaces, and generally an approach to doing that is to use negative margins.

I think this looks more than fine, and better than it did previously 🤷

johncowen: Heres another example of "clipping the padding":

Screenshot 2024-03-25 at 12 34 01

johncowen: What about this:

Screenshot 2024-03-25 at 12 53 28

kleinfreund: What about it? That UI doesn’t seem related to what we’re discussing here.

johncowen: Well its our UI and the baselines of the text and the button text line up?

johncowen: Here's how it was previously, note how the padding, especially for the Mesh title, looks far too big because of the fixed height.

Screenshot 2024-03-25 at 13 03 22

johncowen: > That text happens to line up because (1) the content is vertically centered

Yes, but it "clips the padding" of the box. So its exactly the same as what we are talking about here and its in our UI (which was one of your points)

johncowen: I decided to remove the negative margins here. 

The headers have different margins now, and despite that fact that this PR was supposed to be resolving this I kinda want to get this in 🤷 

Screenshot 2024-03-26 at 11 53 38

kleinfreund (approved):

Issue #2252, opened 2024-03-01
clicking on kuma.io/service links seems to add multiple browser history entries


johncowen:
### What happened?

We recently changed our kuma.io/service links to link to a dataplane list with a pre-filled search query.

If I'm not mistaken this seems to add several extra history entries to the browser, meaning when I click to go back I have to click several times to get back to where I was originally.

PR #2321, opened 2024-03-22
fix: fixes TagList clicking browser history

johncowen:
Simplifies FilterBar and as a consequence fixes https://github.com/kumahq/kuma-gui/issues/2252

P.S. I took the code from https://github.com/kumahq/kuma-gui/pull/1721 and everything just worked.

Closes https://github.com/kumahq/kuma-gui/issues/2252


kleinfreund:
I’m pretty confused about this PR: the contents of this PR seem to be a rather large refactoring of everything related to the filter bar UI rather than a bug fix. What’s the actual bug fix here?

It seems like linking to something like http://localhost:8080/gui/meshes/default/data-planes?query=tag:+%22kuma.io/service:panel%22 causes two history entries because of this code on usages of FilterBar causing a second navigation:

```vue
@fields-change="route.update({
  query: $event.query,
  s: $event.query.length > 0 ? JSON.stringify($event.fields) : '',
})"
```

Surely there is a way to fix the bug that doesn’t involve throwing out virtually all filter bar code?

johncowen:
Oh sorry I forgot to mention, I just resurrected the code I had here: https://github.com/kumahq/kuma-gui/pull/1721 and everything just worked. Seemed to be easier than looking into what the cause was.

kleinfreund (approved):

johncowen:
I'd thought a weird Cypress disabled input thing had sorted itself out via Cypress updates or something, meaning I could remove some tweaks to prevent a flake here - turns out I was wrong:

Screenshot 2024-04-04 at 12 44 34

So I undid that change

PR #2324, opened 2024-03-25
fix(services): fixes internal service navigation test

johncowen:
Fixes (🤞 ) flakey service insight by mocking the serviceType to be `internal`

---

I found that the above change still produced the flake, then I also remembered that we also have a very similar flake in `policies / index` which made me look for commonalities that might cause both tests to occasionally flake, which points to KTable.

After looking at KTable's source I've tried adding an additional guard in our own `AppCollection` which should avoid any potentially incorrect and/or racey behaviour in KTable. There might not be any, but we won't know if this fixes both flakes or not for a little while I suppose. Thought it was worth a try.

This is green right now, so I would say we merge and let it sit for a while and report if we see either of these flakes types of flakes:

Screenshot 2024-03-15 at 14 49 33

Screenshot 2024-03-19 at 09 55 20

### DIFF
--- a/features/mesh/services/Index.feature
+++ b/features/mesh/services/Index.feature
@@ -16,6 +16,7 @@ Feature: mesh / services / index
       body:
         items:
           - name: service-1
+            serviceType: internal
       """
     When I visit the "/meshes/default/services/internal" URL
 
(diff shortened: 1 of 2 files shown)


kleinfreund (approved):

johncowen:
And almost straight away I just saw this in a separate PR:

Screenshot 2024-03-27 at 11 41 52

So the second guess here didn't seem to help 🤷 , but I think the first `internal` addition one did.

PR #2345, opened 2024-03-26
chore: split off an msw module and use service container in vite

johncowen:
TBD

Part 2 of https://github.com/kumahq/kuma-gui/pull/2314


johncowen:
Hmm, so of course anything we pull into vite can't use vite (i.e. `@/` aliasing and `.vue` file loading) which our development.ts file (and files it imports) has.

I'm gonna close this for the moment and come back at a better time.

Issue #2302, opened 2024-03-15
Hide Legacy policies when there are none instead of showing an empty state


kleinfreund:
### Description

See title

johncowen:
Also see https://github.com/kumahq/kuma-gui/pull/2269

> Theres a question about whether we should show the "Legacy Policies" block at all if its empty, but seeing as we do for gateway dataplanes I decided at least for the moment to continue to show it.

But we decided that because they are 'legacy' we shouldn't show this block afterall.

PR #2347, opened 2024-03-26
chore(rules): moves rules related things into the rules module

johncowen:
This is a mix of things that is partly finishing up recent past work and getting ready to continue on with ongoing work.

---

We've recently split off some things into a `@/app/rules` module, so I moved the Rules related `/data` and `sources.ts` into this module. I'm also looking to refactor and combine as much as possible to do with listing of our rules (we currently use an AccordionList to do that), so I've moved and reduced some things around a little to make it easier to see where I can do that.

Lastly, once i'd done that reduction/move around, I also spotted I could also close https://github.com/kumahq/kuma-gui/issues/2302 pretty easily, so I've done that here also. Due to this I also reorganized the tests around this to make them easier to read i.e. reduced the mocking required, hoisted the CSS selectors to apply to the entire file, reordered the tests to put tests of the same type together etc. The coverage is/should be the same.

Closes https://github.com/kumahq/kuma-gui/issues/2302


kleinfreund:
@johncowen Can you please rebase this one?

johncowen:
Oh hey, sorry yeah, just finishing up one more 2.7 thing (https://github.com/kumahq/kuma-gui/issues/2371) then I'll come back and rebase this, gimme 30mins or so

johncowen:
@kleinfreund this one has been rebased

src/app/data-planes/components/BuiltinGatewayPolicies.vue
```diff
@@ -17,7 +17,7 @@
               name: 'policy-detail-view',
               params: {
                 mesh: policy.mesh,
-                policyPath: props.policyTypesByName[policy.type]!.path,
+                policyPath: props.types[policy.type]?.path ?? '',
```
kleinfreund, line 20: **Question**: Why is this being changed? This doesn’t look right. We should assume that a policy type appearing anywhere has a corresponding object found on the `/policies` API response. Allowing for `''` as a policy path is never correct so this code shouldn’t look like that’s possible. (Same with similar occurrences in this change set)

johncowen: I was looking to "deal with it" rather than "ignore it". 

> Allowing for '' as a policy path is never correct so this code shouldn’t look like that’s possible.

Allowing for an `undefined` access error is also never correct. Thing is with `''` we can probably deal with it inside our own `XAction` and fail a little more gracefully than just crashing.

I kinda don't mind either way right now seeing as we aren't using `XAction` everywhere yet. To be fair at the moment the error we get is with the `undefined` access is more descriptive should we ever get an error (highly unlikely I think?), but that reason will go away once we fail more gracefully.

Let me know if you feel strongly about this and I'd be happy to change it back.

Was there anything else in this PR or was it just this?

johncowen: @kleinfreund just noticed this is still outstanding, any thoughts on this?

kleinfreund: I guess I don’t agree with the philosophy behind why this is better. If you ever get to the `''` code path, something’s already gone wrong before this point and the user should see an error of some kind.

What we’re dealing with here is basically that we aren’t able to model that `props.policyTypesByName[policy.type]` is a safe access on the type level. Solving this with a type assertion seems appropriate because it keeps the decision where the problem arises: on the type level. Using a conditional property access with a fallback value is more intransparent and signals "hey, if there isn’t a fitting object, fall back to this value". But that’s a nonsensical application scenario.

In any case, I don’t care too strongly about it as both variants work. Resolving this.

kleinfreund (approved):

Issue #2051, opened 2024-01-19
Check for kuma.io/display-name when displaying names


johncowen:
### Description

We should do this in our data layer.

Check for a `kuma.io/display-name` and use that instead of the name if its present

At least 2 options:

1. We should also move/duplicate the actual `name:` to something else (like `id:`) for if we need it for other things .If `kuma.io/display-name` doesn't exist, `name:` stays as it is, and we just copy the same value to `id:`.
2. Keep `name:` as it is but create a new field called `label` or `displayName`, then change where we use `name` in the View to use `theNewFieldName:` instead. If `kuma.io/display-name` is not set we just copy the `name:` to the new field.
3. ....

jakubdyszkiewicz:
> Triage: let's also display kube namespace if we do this. On Kubernetes, the display name is a policy name without a namespace

PR #2348, opened 2024-03-27
chore(dataplanes): use kuma.io/display-name if it exists

johncowen:
Uses `kuma.io/display-name` as the dataplane `name` if it exists.

We also always set a new `.id` property, which we use for hrefs instead of `name` now.

I added a couple of places to show the `namespace` of a dataplane based on the namespace label (if it exists), but I didn't add this into our list/table views.

Oh I also removed the HTML `title`s we have in a couple of places, I figured those are less necessary now, but let me know if those should stay.

This PR only applies this to everywhere we have dataplanes (or should do). I noticed we probably could do this for policies also, which I'll do separately.

Closes https://github.com/kumahq/kuma-gui/issues/2051

If you view the preview make sure to set it to use a kubernetes-like environment via your cookies or via this link: https://deploy-preview-2348--kuma-gui.netlify.app/gui/#KUMA_ENVIRONMENT=kubernetes

### Screenshots of where the namespace was conditionally added:

Screenshot 2024-03-27 at 18 52 23

Screenshot 2024-03-27 at 18 52 31

### DIFF
--- a/features/mesh/dataplanes/DataplaneDetailsTraffic.feature
+++ b/features/mesh/dataplanes/DataplaneDetailsTraffic.feature
@@ -1,4 +1,4 @@
-Feature: Dataplane traffic
+Feature: mesh / dataplanes / DataplaneDetailsTraffic
   Background:
     Given the CSS selectors
       | Alias           | Selector                                            |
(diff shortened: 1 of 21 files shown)


features/mesh/dataplanes/DataplaneDetailsTraffic.feature
```diff
@@ -77,33 +77,3 @@ Feature: Dataplane traffic
...
-        dataplane:
-          networking:
-            gateway:
-              type: DELEGATED
-      """
-
-    When I visit the "/meshes/default/data-planes/dpp-1-name-of-gateway_delegated/overview" URL
-
-    And the "$detail-view" element contains "dpp-1-name-of-gateway_delegated"
-    And the "$traffic" element doesn't exist
```
johncowen, line 109: This are no longer required. I made an issue about that fact that these were passing also

https://github.com/kumahq/kuma-gui/issues/2350

johncowen:
I had a little chat with @jakubdyszkiewicz and really we need the namespace in the table. I'm just toying with the idea of putting it underneath the name as like a sub-title rather than another column. Only thing is we don't have this pattern anywhere already.

I might just do a little try and see what it looks like.

Whichever way really we need the namespace in the table before we merge this.

johncowen:
I've added a conditional namespace column for dataplane tables so we have something in this PR, but we can iterate a little later.

Two things I noticed:

1. KTable headers don't seem to be reactive so I've had to use the key trick to re-render the table when/if they change.
2. I'm basing the conditional on the first item in the list having a namespace, and I'm guessing there are cases where its possible that the first item doesn't have a namespace and others do. I considered using a filter/find to check all items in your page of results, but then the column could appear and disappear when you click between pages. I'm not sure how edge case-y this is so thought I'd mention here before overthinking it.

Just to be clear, the namespace is now in the table, so this PR is fine to be merged - but highly likely to be iterated on.

kleinfreund (approved):

PR #2367, opened 2024-04-02
Chore/update vitest manually

kleinfreund:
chore(deps-dev): regenerate msw file

chore(deps-dev): update vitest manually

I noticed that somehow Dependabot refuses to update Vitest and it’s not clear why, see: https://github.com/kumahq/kuma-gui/pull/2313

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "@typescript-eslint/eslint-plugin": "7.5.0",
     "@typescript-eslint/parser": "7.5.0",
     "@vitejs/plugin-vue": "5.0.4",
-    "@vitest/coverage-istanbul": "1.3.1",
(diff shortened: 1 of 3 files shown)


johncowen (approved):
Oh good spot, completely forgot I'd seen that, must've got sidetracked.

I took a quick guess/look at why it might have been dropped by dependabot, looks like `coverage-istanbul` had a peerDependency for vitest of of exactly `1.3.1` so I'm guessing one requires the other and vice-versa, then I found this:

https://github.com/kumahq/kuma-gui/pull/2309

Guessing some sort of order combined with the above meant this never went in

kleinfreund:
Could be that the lack of grouped updates in Dependabot makes updates like this impossible then.

PR #2369, opened 2024-04-02
fix: unsafe access to namespace

kleinfreund:


johncowen (approved):
Thanks!

I think we should wrap all our AppCollections in DataCollections and use empty states in DataCollection, which would help prevent this kinda thing form happening. I think I started that work already somewhere

Issue #2370, opened 2024-04-02
In dataplane view rework columns


lahabana:
### Description

- Add icon for type and put it at the start of the row
- Move the zone column after namespace
- Remove type column (it's now an icon).

Issue #2293, opened 2024-03-14
Apply styles and custom logic only to specifically marked "name" columns in AppCollection


kleinfreund:
### Description

We have some code that relies on the first column of an AppCollection to be the "name" column. This isn’t always true and causes a bunch of weird logic and styles that aren’t robust at all:

Instead, we should specifically mark the name columns with a selector on only apply custom styles and custom logic if this selector is found in a table row.

Problematic examples:

https://github.com/kumahq/kuma-gui/blob/878e471739bc71fce5fd60cafdbe8170c71f2b39/src/app/application/components/app-collection/AppCollection.vue#L215-L223

https://github.com/kumahq/kuma-gui/blob/878e471739bc71fce5fd60cafdbe8170c71f2b39/src/app/application/components/app-collection/AppCollection.vue#L227-L239

https://github.com/kumahq/kuma-gui/pull/2292/files

johncowen:
Note: This is kind of a dupe of https://github.com/kumahq/kuma-gui/issues/2275

kleinfreund:
I’m closing this as a duplicate of https://github.com/kumahq/kuma-gui/issues/2275 (which still needs addressing as we have several `td:first-child` selectors in AppCollection)

Issue #2275, opened 2024-03-08
change table row clicking to use a 'tagged' link instead of the first


johncowen:
### Description

See https://github.com/kumahq/kuma-gui/pull/2247#discussion_r1514399022

We'd kinda always figured if we needed this to be a bit more flexible we would move to tag the link to use for the row click rather than just the first.

The change is pretty straightforwards, but then we'll have to go over every table/App collection and tag the link which is a bit of chore work.

johncowen:
This is done, you can now use a `<a data-action />` attribute to mark this link as the one to use for the row click

PR #2373, opened 2024-04-02
chore(dataplanes): revisit dpp column ordering and style

johncowen:
Continuation of https://github.com/kumahq/kuma-gui/pull/2348

> I'm basing the conditional on the first item in the list having a namespace, and I'm guessing there are cases where its possible that the first item doesn't have a namespace and others do. I considered using a filter/find to check all items in your page of results, but then the column could appear and disappear when you click between pages. I'm not sure how edge case-y this is so thought I'd mention here before overthinking it.

We decided to always show the namespace column and use iconography/tooltip for the proxy type:

Screenshot 2024-04-04 at 16 56 09

~Current icons are temporary, but even then if you have better suggestions for the temporary icons used here I'd be happy to change.~

The proxy icon is built in differently to the rest which is hopefully very temporary, and should be switched up to use our icon set soon.

Closes https://github.com/kumahq/kuma-gui/issues/2370

---

There's a larger thing beginning here also with `XIcon` which always has an embedded Tooltip (conditionally enabled depending on the default slot). Thinking being, the majority of the icons we use should really have some sort of tootlip with them. This makes it really easy to add that tooltip, meaning we are more likely to add them.

Following on from this I did enough work to make sure we can replace/support our current `WarningIcon`, and I checked it where we use it in Zones only, but this will need rolling out separately (I can also do it here if its better, it shouldn't be too much work)


johncowen:
We are just waiting on a final icon, gonna open this up for code review in the meantime.

johncowen:
Quick note here: We are gonna change the icons to be east/west arrows for sidecars, and north/south arrow for gateways. Then use a green color for delegated gateways and a kuma purple for builtin. Sidecars will stay kuma purple.

johncowen:
In the end we've gone with east/west icon for sidecars, and a "point-y-thing-into-globe-y-thing" for gateways. Both types of gateways use the same icon - i.e. it's a gateway at this view in the GUI, even if I have a mix of types, I figure its not super important to know that its delegated or builtin, just that its a gateway. Once you've clicked in, you can see the specific type. If there are strong opinions on an alternative to this thinking, let me know as I'd be happy to change up for something else.

@kleinfreund there has been quite a bit of back and forth on this one, but I think we are good for a review on here now 👍

kleinfreund:
One thing I noticed just now that must’ve happened some time before is that the names of Dataplanes in the list view no longer include the type string (e.g. `-gateway_builtin`) in the mock data. That breaks the type being persisted when going from list to detail view and should be fixed.

src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -219,7 +219,12 @@ function getRowAttributes(row: Row): Record<string, string> {
 const click = (e: MouseEvent) => {
   const $tr = (e.target as HTMLElement).closest('tr')
   if ($tr) {
-    const $a: HTMLAnchorElement | null = $tr.querySelector('td:first-child a')
+    const $a: HTMLAnchorElement | null = ['td:first-child a', '[data-action]'].reduce<HTMLAnchorElement | null>((prev, item) => {
+      if (prev === null) {
+        return $tr.querySelector(item)
+      }
+      return prev
+    }, null)
```
kleinfreund, line 227: **Question**: How come this code is becoming more complex by the week? It should become simpler by selecting a specifically marked element and then being done with it.

See https://github.com/kumahq/kuma-gui/issues/2275 and https://github.com/kumahq/kuma-gui/issues/2293.

src/app/x/components/x-icon/XIcon.vue
```diff
@@ -0,0 +1,96 @@
+<template>
```
kleinfreund, line 1: **Question**: What’s you general plan with this component? Is this a long-term thing you intend to keep?

It’s generally a bad idea to have a generic icon component that statically imports all its available icons because that will always include them in the resulting code bundles regardless of whether they’re used.

src/app/application/components/app-collection/AppCollection.vue
johncowen: AppCollection is going to be completely redone soon, and I would think that the AppCollection component will be replaced by an `XTable` that does less.

Just that won't happen before 2.7 goes out.

For the moment we just support two ways i.e. if it can't find an anchor in the first cell, then look for a `[data-action]` instead. That sort of 'complexity' I'm more than ok with, especially seeing as this code will soon be removed from here entirely seeing as we need this functionality in other places apart from AppCollection.

johncowen:
> One thing I noticed just now that must’ve happened some time before is that the names of Dataplanes in the list view no longer include the type string (e.g. -gateway_builtin) in the mock data. 

I added something back in for this

kleinfreund (approved):

PR #2344, opened 2024-03-26
refactor: update Kongponents to KBreadcrumbs milestone

kleinfreund:
**chore(KBreadcrumbs): address breaking changes**

Replace `.k-breadcrumbs-item` with `.breadcrumbs-item-container`.

Signed-off-by: Philipp Rudloff <philipp.rudloff@konghq.com>

---

Previous Kongponents v9 adoption PRs:

- https://github.com/kumahq/kuma-gui/pull/1532
- https://github.com/kumahq/kuma-gui/pull/1853
- https://github.com/kumahq/kuma-gui/pull/1878
- https://github.com/kumahq/kuma-gui/pull/1961
- https://github.com/kumahq/kuma-gui/pull/2118
- https://github.com/kumahq/kuma-gui/pull/2156
- https://github.com/kumahq/kuma-gui/pull/2184
- https://github.com/kumahq/kuma-gui/pull/2270

PR #2374, opened 2024-04-03
refactor: update Kongponents to KCodeBlock milestone

kleinfreund:
**chore(KCodeBlock): address breaking changes**

Replace `isSearchable` prop with `searchable`.

Replace `isProcessing` prop with `processing`.

Replace renamed class and data-testid attributes.

Reset some PrismJS styles interfering with the code block’s general theme.

**refactor: use KCopy in ResourceCodeBlock**

---

- https://github.com/kumahq/kuma-gui/pull/2184
- https://github.com/kumahq/kuma-gui/pull/2270
- https://github.com/kumahq/kuma-gui/pull/2344


src/app/common/code-block/CodeBlock.vue
```diff
@@ -95,13 +95,26 @@ async function handleCodeBlockRenderEvent({ preElement, codeElement, language, c
...
-:deep(.k-code-block-actions) {
+:deep(.code-block-actions) {
   position: sticky;
   z-index: 4;
   top: var(--AppHeaderHeight);
 }
+
 :deep(.k-highlighted-code-block) {
   border: none;
 }
```
johncowen, line 108: is `k-highlighted-code-block` still in the new KCodeBlock?

kleinfreund: No. I’ll remove this rule set as it’s covered by the new addition below.

johncowen:
Notes: Not required for a 2.7 release, also waiting on https://github.com/Kong/kongponents/pull/2100 which is currently draft. Would be good to get some advanced review here though seeing as it changes up how we do copying for the `Copy as K8s` functionality

src/app/common/code-block/ResourceCodeBlock.vue
```diff
@@ -90,25 +87,27 @@ const emit = defineEmits<{
...
+const copy = ref<(cb: CopyCallback) => void>(() => {})
+let promise = new Promise((resolve: Resolve, reject) => {
+  copy.value = (cb) => cb(resolve, reject)
+})
 
 const fetcher = async (_params?: SingleResourceParameters): Promise<Entity> => {
-  let res: Entity
   try {
-    res = await p
+    return await promise
```
johncowen, line 105: nit: Do we need to await the promise here if we are just returning it from an async function?

johncowen (commented):
I think the PRs this needs are still in draft but I had a couple of q's:

1. Does our devtime console logging for copying no work with this now, I don't think it does? Is it possible to keep that, maybe not?
2. Wanted to ask about losing the XDisclosure (née KToggle). I kinda like the fact that it managed your boolean state for you so you could call/fire a function/event and read data rather than setting data directly which seems to fit with the platform better. Not a biggie but thought I'd ask.

src/app/common/code-block/ResourceCodeBlock.vue
kleinfreund: Technically no. I removed it while rebasing the PR.

kleinfreund:
> 1\. Does our devtime console logging for copying no work with this now, I don't think it does? Is it possible to keep that, maybe not?

What do you mean by this? What dev-time logging?

> 2\. Wanted to ask about losing the XDisclosure (née KToggle). I kinda like the fact that it managed your boolean state for you so you could call/fire a function/event and read data rather than setting data directly which seems to fit with the platform better. Not a biggie but thought I'd ask.

There is no need for introducing a component for the purely programmatic interaction of flipping a boolean variable. It does nothing better and is pretty awkward to use. This removes at least one of the awkward aspects of this component.

johncowen:
Here's a link to the original PR for the devtime logging:

https://github.com/kumahq/kuma-gui/pull/1981

> There is no need for introducing a component for the purely programmatic interaction of flipping a boolean variable. It does nothing better and is pretty awkward to use

The biggest point of it was the fact that you only read properties and call functions, no setting properties directly, but 🤷 the q was mainly out of interest.

Anyway as I said neither are biggies. Shout me when the kongponent PR comes out of draft

johncowen:
Just to note this generally looks fine to me, lemme know once the kongponents PR is in and we have the new package version here.

johncowen (approved):

Issue #2372, opened 2024-04-02
Use display-name for policy list


lahabana:
### Description

Same as for dataplane for new policies

PR #2375, opened 2024-04-03
chore(policies): use display-name label for policy listing

johncowen:
Use `display-name` and `namespace` labels for policy listings

---

I noticed that the mock for the affected dataplanes endpoint was static, so I changed that up but also that this endpoint is paged, but we don't offer any paging and we use frontend searching on the single page. I stopped short of doing anything towards changing this tho. Not sure if its on purpose or just an accidental omission? We might want to follow up with changing that soon.

Closes https://github.com/kumahq/kuma-gui/issues/2372

### DIFF
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -393,6 +393,7 @@ export interface DataPlaneOverview extends MeshEntity {
   labels?: {
     'kuma.io/display-name'?: string
     'k8s.kuma.io/namespace'?: string
+    [key: string]: string | undefined
   }
   dataplane: {
     networking: DataplaneNetworking
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -390,7 +390,10 @@ export interface DataPlane extends MeshEntity {
  */
 export interface DataPlaneOverview extends MeshEntity {
   type: 'DataplaneOverview'
-  labels?: Record<string, string>
+  labels?: {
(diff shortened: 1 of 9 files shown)


src/types/index.d.ts
```diff
@@ -390,7 +390,10 @@ export interface DataPlane extends MeshEntity {
  */
 export interface DataPlaneOverview extends MeshEntity {
   type: 'DataplaneOverview'
-  labels?: Record<string, string>
+  labels?: {
+    'kuma.io/display-name'?: string
+    'k8s.kuma.io/namespace'?: string
```
kleinfreund, line 395: **Suggestion**: This change would prevent accessing other properties. It should be:

```suggestion
    'k8s.kuma.io/namespace'?: string
    [key: string]: string | undefined
```

johncowen: There's something to be said for not letting ourselves statically accessing properties that aren't specified here but 🤷

kleinfreund (approved):

johncowen:
> but we don't offer any paging and we use frontend searching on the single page. I stopped short of doing anything towards changing this tho. Not sure if its on purpose or just an accidental omission?

@kleinfreund any thoughts on this ^

kleinfreund:
> @kleinfreund any thoughts on this ^

Not really. This can definitely be improved, but users are relatively unlikely to run into limitations here

PR #2384, opened 2024-04-05
fix(gateways): ensure gateway linking uses ids rather than names

johncowen:
Similar to https://github.com/kumahq/kuma-gui/pull/2380


kleinfreund:
Can you please add some tests for this so we can’t mess this up anymore?

johncowen:
Yeah sure I can do that in a follow up?

kleinfreund:
Sure

kleinfreund (approved):

johncowen:
Tests added here https://github.com/kumahq/kuma-gui/pull/2386

PR #2386, opened 2024-04-05
chore(tests): add display-name/ns related navigation tests

johncowen:
Requires https://github.com/kumahq/kuma-gui/pull/2385

There wasn't really a good way to make a test file that covered all of these, so I made it fairly easy to copy/pasta across all 'modules' and did that.

You'll hopefully see these are all the same tests, just repeated in each module.


johncowen:
https://github.com/kumahq/kuma-gui/pull/2385 is on here now and all green

PR #2402, opened 2024-04-08
refactor: switch package manager to npm

### LINKED ISSUE #1837: Migrate away from Yarn classic (v1)
### Description

Yarn classic (v1) hasn’t been a recent version of Yarn for a while and is no longer maintained ([the last change was in September ’22](https://github.com/yarnpkg/yarn))

We should probably migrate away from it.

Options are Yarn, pnpm, and npm. See below comments for my evaluation:

- [Yarn modern (v4)](https://github.com/kumahq/kuma-gui/issues/1837#issuecomment-1898313283)
  - https://yarnpkg.com/
  - https://github.com/yarnpkg/berry/issues
- [pnpm (v8)](https://github.com/kumahq/kuma-gui/issues/1837#issuecomment-2034433592)
  - https://pnpm.io/
  - https://github.com/pnpm/pnpm/issues
- [npm (v10)](https://github.com/kumahq/kuma-gui/issues/1837#issuecomment-2034433777)
  - https://npmjs.com/
  - https://github.com/npm/cli/issues

**TL;DR verdict**: npm > pnpm > yarn

kleinfreund:
**IMPORTANT NOTE**: This PR commits the new `.npmrc` file. Double-check that your local clone does **not** have any npm tokens stored in it to avoid leaking it.

## Changes

Replace yarn v1 with npm v10.

Add .npmrc file holding the npm configuration.

Remove the deps script. From testing the CI integration of npm with deleted caches, it doesn’t seem to be necessary.

Remove the unused package `@types/marked`.

Resolves #1837.

## Notes

- **GitHub Actions and npm**: By default, GitHub Actions workflows have an npm version pre-installed (currently, that seems to be `10.5.0` which is very recent). Using actions/setup-node will also install npm (presumably as the byproduct of installing the specified node version). That’s why currently, in CI, the installed npm version (10.1.0) does **not** match the `pkg.engines.npm` value (arguably, actions/setup-node should use that).
- To avoid the create-gui-pr.yml workflow breaking with this change, we need https://github.com/kumahq/kuma-gui/pull/2487 and the Makefile setup backported to the current release branches.


lahabana:
Use a netlify.toml: https://docs.netlify.com/configure-builds/file-based-configuration/#sample-netlify-toml-file

.github/workflows/create-gui-pr.yml
```diff
@@ -51,7 +51,8 @@ jobs:
         with:
           node-version-file: '.nvmrc'
 
-      - run: yarn install --frozen-lockfile
+      - run: |
+          make clean-install/sync
```
johncowen, line 55: This should just be `install/sync` and in the Makefile `install/sync` can then call whatever it likes. The fact that the sync workflow installs 'cleanly' (whatever that means) is controlled in the Makefile itself, which we can have different versions of per branch.

Could we also remove the `clean-install` target, its really confusing and we don't need it. We can just make `install/sync` in turn call `install --frozen-lockfile`.

If this isn't clear or if you have different ideas/plans, could we hook up offline to prevent any further confusion.

kleinfreund: Oh, well. I’ve changed this now.

For the record, I think it’s a mistake to not distinguish between "install" and "clean-install" explicitly in the command names. It will likely lead to more confusion than just learning the difference between "install" and "clean-install". Now, you will always need to look up what the command actually does and can’t infer it from the command name. In fact, you might assume you can and make the wrong conclusion (e.g. that `make install/sync` runs `npm install` which it doesn’t).

johncowen: Ok cool, I think we should hook up later to make sure we are both on the same page here:

If I'm not mistaken, the additional change here means that `--frozen-lockfile` isn't used when it should be?

https://github.com/kumahq/kuma-gui/blob/8c2b54f8ecfc4c4e078bbff04fcec85ea1be504c/mk/install.mk#L6-L7

https://github.com/kumahq/kuma-gui/blob/8c2b54f8ecfc4c4e078bbff04fcec85ea1be504c/mk/install.mk#L2-L3

---

In my mind `create-gui-sync` should always call `install/sync` and `build/sync` on `master` (which is essentially on all branches). `install/sync` should in turn call `yarn/npm install --frozen-lockfile` (or even `make clean-install`, it doesn't matter really) This way if for some reason a future branch doesn't need a frozen-install or requires something else, we can do that per branch in the Makefile.

If you call the thing that we don't want to change `clean-install` then you are leaking that fact that it does a "clean install" into the name i.e. the thing we are avoiding changing. If we need to change it to not do a clean install, then really we'd need to change the name, and then backport it another 5 times.

All in all its a naming thing, but its similar to when something gets called `blue-button` instead of `primary-button`, and then designers change `blue-button` so its red and we end up with weird confusing naming.

I know its the tiniest of things but, I'd rather get it right now and then forget about it.

Shout me when you are back around.

kleinfreund: > If I'm not mistaken, the additional change here means that `--frozen-lockfile` isn't used when it should be?

It looks like you’re looking at an outdated version of this branch. This is what the install scripts currently look like:

```sh
.PHONY: install
install:
	npm install

.PHONY: install/sync
install/sync:
	npm clean-install
```

johncowen: Right but the previously backported Makefiles are wrong now aren't they? Because `install/sync` on 2.7 (etc) is like this:

https://github.com/kumahq/kuma-gui/blob/be68aadf480247c57827007cdaa6c8dc0cba6528/mk/install.mk#L1-L7

i.e. without the frozen-lockfile.

The release branches need `frozen-lockfile` don't they?

kleinfreund: This has all been addressed now so I’m resolving this.

package.json
```diff
@@ -34,9 +34,9 @@
     "docs:preview": "vitepress preview"
   },
   "engines": {
-    "node": ">=20.9"
+    "node": ">=20.9",
+    "npm": "^10.0.0"
```
johncowen, line 38: Maybe I misunderstood this from your PR description:

> That’s why currently, in CI, the installed npm version (10.1.0) does not match the pkg.engines.npm value (arguably, actions/setup-node should use that).

If in CI the installed npm version is 10.1.10 then this matches the `pkg.engines.npm` value doesn't it?

johncowen (commented):
I had a few questions to leave before testing this out properly here, lmk on those. In the meantime gonna take this for a spin.

.npmrc
```diff
@@ -0,0 +1 @@
+save-exact=true
```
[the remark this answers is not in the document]
kleinfreund: .npmrc is **the** npm configuration file. Avoiding pinning of versions to avoid using the configuration file makes no sense. What are we going to do if we want to configure something else? We shouldn’t avoid using the configuration file because a developer _can_ put credentials in them.

> At this point it feels kinda like exact pinning is causing more problems than it solves.

I principally disagree with this statement.

package.json
kleinfreund: > If in CI the installed npm version is 10.1.10 then this matches the `pkg.engines.npm` value doesn't it?

No, it does not. It _happens_ to fulfill it because of the Node version we’re installing (which also installs npm), but there is no guarantee for that and there exists no mechanism in `actions/setup-node` today that would allow us to configure the npm version.

If we want to make sure the npm version we define in `pkg.engines.npm` is used in CI, we would need to manually install it after every instance of `actions/setup-node` in CI. Considering this, it turns out the need to manually install pnpm in CI is actually an advantage.

.npmrc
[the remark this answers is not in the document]
johncowen: I think the answer to that is that if you use the recommended defaults that the tool comes with you then don't have to configure things, and things just tend to work with no issues.

If we need to configure things differently to the vast majority of the JS/TS ecosystem we should be asking ourselves why.

Anyway no action needed here

package.json
johncowen: Oh I think I get you.

From what I remember `npm` versions are bundled with `node` aren't they? So depending on which version of `node` you have pin to you get a certain version of `npm` also?

kleinfreund: > From what I remember `npm` versions are bundled with `node` aren't they?

Yes. That’s the reason why it’s _that_ particular version of npm that is installed in CI (as opposed to the one already available in GitHub Actions by default).

> So depending on which version of `node` you have pin to you get a certain version of `npm` also?

Correct. Any particular version of node does come with a particular version of npm (presumably because that’s the latest stable version npm that was available at the time that node version was released). There’s no reason however to not select/install a completely different npm version so long as it works with that node version (where there should almost never be any version conflicts).

johncowen: Ok so I think what we are saying here is that as we pin `node` that also pins the version of `npm`, in which case we could just remove this entry from `pkg.engines.npm`? 

I'm just thinking of less stuff to think about when we come to do node upgrades.

Not saying we need to do it here, just asking more than anything.

johncowen: Oh I thought you said we weren't controlling it?

> but there is no guarantee for that and there exists no mechanism in actions/setup-node today that would allow us to configure the npm version.

Or perhaps better said I'd understood we are controlling it via `node` pinning.

Anyway doesn't matter I was half thinking out loud. I think this PR is good 1 sec lemme approve.

johncowen (approved):

package.json
kleinfreund: > Oh I thought you said we weren't controlling it?

This is specific to CI where one way or another, the tooling we use (actions/setup-node) doesn’t honor `pkg.engines.npm` and always installs its own npm version.

Locally, that’s not the case and `pkg.engines.npm` gives us the useful validation for whether a developer uses a compatible version.

kleinfreund:
I’ve rebased this one more time.

PR #2404, opened 2024-04-08
fix: center aligns text of DefinitionCard titles

johncowen:
Note: I almost did this just in the `ControlPlaneStatus` component which is where the problem arises. But it seemed like if was fine to do at an application level. If anyone is unsure as to whether this is fine or not, I'd be happy to move it back to fixing it in the more focussed option within `ControlPlaneStatus`.

But, it would be super helpful to get a second opinion/confirmation as to whether doing this at an application level has no negative effect.

## Before

Screenshot 2024-04-08 at 15 44 23

## After

Screenshot 2024-04-08 at 15 44 09

---

As a sidenote I noticed that all of the text rendered by `DefinitionCard` is rendered within `div`s. We should probably change that at some point to be more semantic (either using a `dl`, or `h*`/`p` elements) as the `<div>` element should be used only when no other semantic element is appropriate.

### DIFF
--- a/src/app/common/DefinitionCard.vue
+++ b/src/app/common/DefinitionCard.vue
@@ -42,7 +42,7 @@ const props = withDefaults(defineProps<{
 
 .definition-card-title {
   display: flex;
-  align-items: flex-end;
+  align-items: center;
   gap: $kui-space-40;
 }
 


johncowen:
Sorry to ask, I just wanted to get a bit more of a stronger confirmation that this has no negative effects on the application as a whole, I know you added DefinitionCard and the CSS I changed here, you're absolutely sure this is good right, I'm not sure why it was originally `flex-end`?

kleinfreund:
> Sorry to ask, I just wanted to get a bit more of a stronger confirmation that this has no negative effects on the application as a whole, I know you added DefinitionCard and the CSS I changed here, you're absolutely sure this is good right, I'm not sure why it was originally `flex-end`?

I’m assuming you checked relevant uses of this component to make sure this isn’t having any negative effects. Did you?

I don’t remember whether there were specific cases that needed this to be particularly `flex-end`.

johncowen:
Ok sounds like we are good here, I'm gonna merge

PR #2411, opened 2024-04-09
chore: adds additional error message to XAction and use it in service/policy listings

johncowen:
I'm using this to try and get a flake/error on a couple of tests that commonly flake, whilst it could go in, I want to keep running it until I get a flake. Also, I'm considering adding something on this PR that reruns certain tests multiple times to make any potential flakes more likely to come up in the PR.

~DO NOT MERGE!11!! No honestly, don't 😅~

---

edit: I've run this a few times, and I can't get a flake out of it. I'd rather merge this seeing as it doesn't really change any functionality and uses `XAction` (which we'll probably end up rolling out everywhere) and see if flake continue to appear over time.

### DIFF
--- a/src/app/policies/views/PolicyListView.vue
+++ b/src/app/policies/views/PolicyListView.vue
@@ -104,7 +104,7 @@
                       @change="route.update"
                     >
                       <template #name="{ row }">
-                        <RouterLink
+                        <XAction
                           :to="{
                             name: 'policy-summary-view',
                             params: {
@@ -119,7 +119,7 @@
                           }"
                         >
                           {{ row.name }}
-                        </RouterLink>
+                        </XAction>
                       </template>
                       <template #namespace="{ row: item }">
                         {{ item.namespace.length > 0 ? item.namespace : t('common.detail.none') }}
@@ -157,7 +157,7 @@
                       </template>
 
                       <template #details="{ row }">
-                        <RouterLink
+                        <XAction
                           class="details-link"
                           data-testid="details-link"
                           :to="{
@@ -175,7 +175,7 @@
                             decorative
                             :size="KUI_ICON_SIZE_30"
                           />
-                        </RouterLink>
+                        </XAction>
                       </template>
                     </AppCollection>
                     <RouterView
--- a/src/app/services/views/ServiceListView.vue
+++ b/src/app/services/views/ServiceListView.vue
@@ -47,7 +47,7 @@
             >
               <template #name="{ row: item }">
                 <TextWithCopyButton :text="item.name">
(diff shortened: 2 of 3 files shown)


johncowen:
Going in for the 7th run, typically can't get it to flake 😆 

I had a quick look into the idea of tagging a test with `@flake` or something, and based on that running that one test multiple times. Initial look into that seemed like it wouldn't be trivial to do unfortunately. Considering merging this as is as it does no harm - who knows maybe just the extra code here has somehow made it so the flake doesn't happen 🤷‍♂️ .

Will run it a couple more times and see.

kleinfreund (approved):

johncowen:
Just wanted to note here that I've not seen the flake that this is investigating for quite a while, which is weird especially considering we've PRed quite a bit lately. Wondering if something somewhere has indirectly affected this, maybe https://github.com/kumahq/kuma-gui/pull/2393 🤷 

Anyway, this won't silence it if it happens but instead make it easier to figure out whats happening. Lets see how it goes.

johncowen:
Boom 💥 

https://github.com/kumahq/kuma-gui/actions/runs/8664390344/job/23760713700?pr=2433

Screenshot 2024-04-12 at 10 48 41

This flaked, and if I'm not mistaken this it doesn't include the extra message I tried to add. So either I added the message wrong or the links I thought were causing the flake, aren't actually causing the flake.

kleinfreund:
Last resort Cypress test debugging:

```js
document.body.insertAdjacentHTML('beforeend', `
  <div style="position: fixed; z-index: 100; top: 20px; left: 20px; padding: 10px; background: tomato; pointer-events: none;">
    <pre>${JSON.stringify(error)}</pre>
  </div>
`)
```

That way, if the error does happen, you’d see it in the video recording.

johncowen:
Its looks to me tho that we aren't catching the right error / we aren't wrapping the extra message around the error? The above code would just produce the error in the DOM rather than in the Cypress CLI output, but if we are still wrapping around the wrong error, we still won't see anything? Or did I misunderstand something there.

Ummm, I suppose TLDR on this comment is: I 'think' (but I'm not sure) that the problem is that we are wrapping the wrong error, not that we can't display a customized error message.

---

In some way I think we are one step forwards as this flake occurring without the extra error message discounts links that I thought were causing the error. Do you come to the same thinking?

kleinfreund:
Yeah, if the flake doesn’t produce the extra error, it can’t come from non-XAction instances.

What I suggested above was meant as a workaround for not seeing errors that might occur after another error that fails the test already.

Issue #2415, opened 2024-04-11
Links to the policy list view shouldn't start at CircuitBreakers


johncowen:
### Description

For example clicking here:

<img width="1117" alt="Screenshot 2024-04-11 at 14 19 50" src="https://github.com/kumahq/kuma-gui/assets/554604/61ec072a-d03c-4e55-ab71-8b1268bec2c8">

Takes me here:

<img width="1424" alt="Screenshot 2024-04-11 at 14 20 39" src="https://github.com/kumahq/kuma-gui/assets/554604/ee29ef09-6d82-453d-8113-01bbeeb43bae">

When it should probably take me to either MeshCircuitBreaker (because: the first that has a non-zero amount), or MeshAccessLog (because: the first in the list)

kleinfreund:
CircuitBreaker _used_ to be the first in the list which is why it was chosen. Now it should be MeshAccessLog. For the Policies tab (or links to the policy list view in general), we can’t link to the first one with non-zero policies because that would mean we’d always have to first evaluate `/mesh-insights/:mesh`.

johncowen:
Closed by https://github.com/kumahq/kuma-gui/pull/2417 (must've used the wrong keyword for autoclose)

PR #2417, opened 2024-04-11
fix(policies): redirect to first available policy type instead of CBs

johncowen:
Instead of always linking to CircuitBreakers when clicking any `[Policies]` link, we choose the first one in the list and redirect you to that one.

Closes https://github.com/kumahq/kuma-gui/issues/2415

### DIFF
--- a/src/app/policies/routes.ts
+++ b/src/app/policies/routes.ts
@@ -21,16 +21,6 @@ export const routes = () => {
             module: 'policies',
           },
           component: () => import('@/app/policies/views/PolicyTypeListView.vue'),
-          redirect: (to) => {
-            return {
-              ...to,
-              params: {
-                ...to.params,
-                policyPath: 'circuit-breakers',
-              },
-              name: 'policy-list-view',
-            }
-          },
           children: [
             {
               path: ':policyPath',
--- a/src/app/policies/views/PolicyTypeListView.vue
+++ b/src/app/policies/views/PolicyTypeListView.vue
@@ -8,8 +8,6 @@
       v-slot="{ route, t }"
       name="policy-list-view"
       :params="{
-        page: 1,
-        size: me.pageSize,
         mesh: '',
         policyPath: '',
         policy: '',
@@ -60,14 +58,14 @@
                         :key="current"
                       >
                         <div
-                          v-for="policyType in items"
+                          v-for="(policyType, i) in items"
                           :key="policyType.path"
                           class="policy-type-link-wrapper"
                           :class="{
                             'policy-type-link-wrapper--is-active': current && current.path === policyType.path,
                           }"
                         >
-                          <RouterLink
+                          <XAction
                             class="policy-type-link"
                             :to="{
                               name: 'policy-list-view',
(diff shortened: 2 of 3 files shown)


src/app/policies/views/PolicyTypeListView.vue
```diff
@@ -76,10 +74,11 @@
                                 policyPath: policyType.path,
                               },
                             }"
+                            :mount="route.params.policyPath.length === 0 && i === 0 ? route.replace : undefined"
```
johncowen, line 77: > CircuitBreaker used to be the first in the list which is why it was chosen. Now it should be MeshAccessLog. For the Policies tab (or links to the policy list view in general), we can’t link to the first one with non-zero policies because that would mean we’d always have to first evaluate /mesh-insights/:mesh.

https://github.com/kumahq/kuma-gui/issues/2415#issuecomment-2049735176

---

If we wanted to redirect to the first non-zero one we could just look at `meshInsight?.policies?.[policyType.name]?.total` like we do below

I decided to just go to the first one though (whatever that is from now into the future). I'm easy either way, but I kinda think we should just use the first existing even if its zero.

johncowen: Ah also, not totally sure on the name `:mount` here. I started with `:auto` and then tried to use `emit` instead of a prop and `@mount` felt like a better vue-like name, but then you can't detect when the `@mount` function changes I don't think? So I moved it back to `:mount`. If you had any naming suggestions, lmk.

kleinfreund: Might just call it `replace`. Is there really ever gonna be a different use case for this sort of navigation indirection?

kleinfreund (approved):

src/app/policies/routes.ts
```diff
@@ -21,16 +21,6 @@ export const routes = () => {
...
-          redirect: (to) => {
-            return {
-              ...to,
-              params: {
-                ...to.params,
-                policyPath: 'circuit-breakers',
-              },
-              name: 'policy-list-view',
-            }
-          },
```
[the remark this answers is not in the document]
johncowen: Because the first one could still be `CircuitBreaker` as far as I understand it? Plus why do one fix here and then a different one on `master` when the one on master is fine for both?

src/app/policies/views/PolicyTypeListView.vue
johncowen: Yeah I see what you're thinking,  I was thinking something like `redirect` also.  

Thing is you can put anything in this function not just something that redirects. I can also console.log in there if I wanted to so `:redirect/replace="() => console.log"` doesn't make sense. I think the function means that if there is another use case we haven't thought of, injecting a function/behaviour is a looser coupling/more flexible, and the fact that its a function means "do _something_ on mount".

The other option is that I make it `redirect=true` and then import `useRouter` into `XAction` and redirect from there.

The other thing I was thinking I should avoid is coupling it to RouteView/AppView, although now come to think of it maybe coupling it and doing something similar to RouteTitle would be good. 

I dunno, I'm still a bit in two minds 🤔 I think if you're not super strong on anything here we merge this, and possibly tweak after.

johncowen: Gonna resolve this one I think, I didn't notice you'd already approved thanks for that!

src/app/policies/routes.ts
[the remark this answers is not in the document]
johncowen: Gonna resolve this one and merge (just noticed you'd approved), let me know if I got it wrong on the "potentially there are CircuitBreakers" thing.

PR #2424, opened 2024-04-12
chore: update version to 2.7

kleinfreund:


johncowen (approved):
nice, thanks!. Would be good to have some sort of auto thing that would do this and branch for us I think 🤔

PR #2448, opened 2024-04-16
chore(deps): update kongponents

kleinfreund:

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "dependencies": {
     "@kong-ui-public/i18n": "2.1.5",
     "@kong/icons": "1.8.16",
-    "@kong/kongponents": "9.0.0-alpha.137",
+    "@kong/kongponents": "9.0.0-alpha.141",
     "@vueuse/core": "10.9.0",
     "brandi": "5.0.0",
     "deepmerge": "4.3.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1895,10 +1895,10 @@
   resolved "https://registry.yarnpkg.com/@kong/icons/-/icons-1.8.16.tgz#4284e8c3b587f0c903f5ef2f99bee5fabe3762b5"
   integrity sha512-aQvZUrX9Px7SqQkGlZnjlAf7aubW3mFobF9ojJ88lMPZ1XN4gDI0TKBcWtvGCtCqgbvtqe0LJD9gqAjTRqArEg==
 
-"@kong/kongponents@9.0.0-alpha.137":
-  version "9.0.0-alpha.137"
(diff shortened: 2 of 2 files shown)


johncowen:
I'm guessing the `@npm:` stuff flip-flopped back because this was a manual/local package update rather than one done by dependabot. Guessing they would get flip-flopped back again once they are involved in a dependabot upgrade? Do you know if that's the case? Just trying to figure out why the flip-flop/where the difference is.

kleinfreund:
> I'm guessing the `@npm:` stuff flip-flopped back because this was a manual/local package update rather than one done by dependabot. Guessing they would get flip-flopped back again once they are involved in a dependabot upgrade? Do you know if that's the case? Just trying to figure out why the flip-flop/where the difference is.

I assume that is the case, yes. I have no idea what Dependabot does differently to produce a different lock file but locally, I’ve just changed the package.json version and ran `yarn install`. I also tried a slightly different variant (`yarn add @kong/kongponents@...`) but the effect was the same.

johncowen:
Hmm yeah judging by this PR https://github.com/kumahq/kuma-gui/pull/1857 dependabot doesn't flip those weird dependencies back out. What do you think of just not committing those hunks? I'm not totally sure on that but I'm inclined to copy exactly what dependabot does if we can.

Lastly, it looks as though this specific `yarn.lock` dedupe issue (the `@npm:` one) is caused by the fact that kongponents' current versioning scheme causes issues with dependabot, and in us working around that issue we are manually upgrading kongponents, which in turn causes this specific yarn.lock flip-flopping issue. Does that line of thinking make sense to you?

kleinfreund:
> What do you think of just not committing those hunks? I'm not totally sure on that but I'm inclined to copy exactly what dependabot does if we can.

I’m fine with that. We don’t have too much we need to update locally and eventually nothing at all at which point this should resolve itself.



> Lastly, it looks as though this specific `yarn.lock` dedupe issue (the `@npm:` one) is caused by the fact that kongponents' current versioning scheme causes issues with dependabot, and in us working around that issue we are manually upgrading kongponents, which in turn causes this specific yarn.lock flip-flopping issue. Does that line of thinking make sense to you?

This issue has nothing to do with Kongponents. You can verify this by trying to update another dependency this way: the same result will occur. It just so happens that Kongponents is the only thing we currently update manually.

johncowen:
> You can verify this by trying to update another dependency this way: the same result will occur.

Right, but is this because we are trying to update the dependency, i.e. its not dependabot doing it?

1 sec lemme approve this also

johncowen:
Oh wait I just took a look on another PR and its looks like its flipping it there also 🤔 .

kleinfreund:
> Right, but is this because we are trying to update the dependency, i.e. its not dependabot doing it?

What do you mean by this?

johncowen (approved):

Issue #2432, opened 2024-04-12
Create `RouteView::src()/RouteView::uri()`


johncowen:
### Description

First iteration of this would hopefully look something like this:

```vue
<RouteView
  v-slot="{ route, t, uri }"
>
  <DataSource :src="uri<sources>(
    '/mesh/:mesh/dataplanes',
    {
      mesh: "default"
    }
  )"> 
    <!-- data here begins as undefined -->
    {{data?.items}} 
  </DataSource>


<DataLoader :src="uri<sources>(
    '/mesh/:mesh/dataplanes',
    {
      mesh: "default"
    }
  )">
    <!-- data is never undefined because DataLoad will block until data is defined before rendering the default block  -->
    {{data.items}} 
  </DataLoader>
</RouteView>
...
import type { sources } from '@/app/dataplanes/sources'
```

`uri` would return a `toString`able Token that would also carry the type of the source with it, this Token could then be inspected with Typescript inside of DataSource to figure out the type/shape of the data that the DataSource emits. The DataSource would still be able to call `toString` on the Token to access the bare string uri.

### Why?

1. Typechecking/autocompletion of DataSource/Loader uris
2. We currently have a little sharp edge where we have `...Source` types which are typed so that data can be `undefined`, which is the case for DataSource. But we also use these for DataLoader (which guarantees data is never `undefined`) and then have to perform an extra check just for this type to make sure the data isn't undefined. 

### Wait wat?!?! You said "first iteration", how come?

Ideally `uri` would know all the sources for the entire application and therefore you wouldn't have to pass through the sources in `uri<sources>`. Making `uri` know about all of the potentially injected sources in the entire application would depend on some typescript changes in out DI layer, and whilst this would also solve a similar issue for `can` and `i18n` typechecking/autocompletion, it would be a larger chunk of work for something which is essentially a DX feature.

This iteration should be far less effort/churn to achieve, and also it would be partly used if we ever come to do the second iteration. Second iteration would be an entirely different scope not specific to DataSource.

PR #2450, opened 2024-04-16
refactor(control-planes): refactors cp-details to use DataLoader

johncowen:
First stage refactor of the control-plane-detail page (usually our landing page) to use DataLoader/XAction and some of the newer patterns we have in the application:

1. Uses DataLoader instead of DataSource (I think there is a bit of extra detail i need to look into here that might produce a separate PR, I'll detail in that PR if so)
2. Uses XAction over RouterLink
3. Adds DataLoader#loadable slot. This state is `loading` and `loaded`. I'm thinking we can use this instead of the `loading=false` property on `DataLoader`. I'm pretty sure it will allow us to eventually avoid having to use `data!` sometimes, in that the default slot can export `T` and the loadable slot can export `T | undefined`, meaning in a large amount of cases when not using the loadable slot we can use `data` instead of `data!` (also related https://github.com/kumahq/kuma-gui/issues/2432)

There's more work to come here I think, but I'm going to try and use this view to do several under-the-hood upgrades. Just an extra reminder that there are some additional complications involving `ControlPlaneStatus` (ideally I would like have its KCard in the `*View.vue` but I don't think that is possible yet)

One thing I found is that I almost changed the lists back to use the loading spinner instead of the KTable skeletons by accident due to the component usage hiding what was going on. It's highly likely I'll delete the components used here seeing as they are just AppCollection wrappers. I would much rather it was clear what was happening all in one place and I'm pretty sure these components are not multi-use.

### DIFF
--- a/features/main-overview/DetailViewContent.feature
+++ b/features/main-overview/DetailViewContent.feature
@@ -2,7 +2,6 @@ Feature: Overview: Detail view content
   Background:
     Given the CSS selectors
       | Alias                       | Selector                                    |
-      | details                     | [data-testid='detail-view-details']         |
       | zone-control-planes-details | [data-testid='zone-control-planes-details'] |
       | meshes-details              | [data-testid='meshes-details']              |
     And the URL "/mesh-insights" responds with
@@ -89,7 +88,6 @@ Feature: Overview: Detail view content
 
     When I visit the "/" URL
     Then the page title contains "Overview"
-    And the "$details" element exists
 
     And the "[data-testid='zone-control-planes-status']" element doesn't exist
     And the "[data-testid='meshes-status']" element contains "3"
@@ -161,7 +159,6 @@ Feature: Overview: Detail view content
     When I visit the "/" URL
 
     Then the page title contains "Overview"
-    And the "$details" element exists
 
     And the "[data-testid='zone-control-planes-status']" element contains "1/2"
     And the "[data-testid='meshes-status']" element contains "3"
--- a/src/app/application/components/data-source/DataLoader.vue
+++ b/src/app/application/components/data-source/DataLoader.vue
@@ -23,6 +23,12 @@
         </slot>
       </template>
       <slot
+        name="loadable"
+        :data="props.src !== '' ? allData[0] : undefined"
+        :error="props.src !== '' ? allErrors[0] : undefined"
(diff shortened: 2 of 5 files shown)


johncowen:
> Adds DataLoader#loadable slot. This state is loading and loaded. I'm thinking we can use this instead of the loading=false property on DataLoader. I'm pretty sure it will allow us to eventually avoid having to use data! sometimes, in that the default slot can export T and the loadable slot can export T | undefined, meaning in a large amount of cases when not using the loadable slot we can use data instead of data! (also related https://github.com/kumahq/kuma-gui/issues/2432)

I have a PR to come after this that adds type support for DataSource uris. It's not _exactly_ what I wanted. There's an extra initial argument for `uri` due to Typescript's lack of support for having a mix of required and inferred generics, but it pretty much implements https://github.com/kumahq/kuma-gui/issues/2432 and the second iteration mentioned in that issue would remove the need for the initial argument.

pre-PR: https://github.com/johncowen/kuma-gui/compare/refactor/control-plane-detail...johncowen:kuma-gui:chore/uri

The 'old way' still works also so we can incrementally upgrade and we also have an escape hatch if the type support is too strict.

kleinfreund (approved):

kleinfreund:
> One thing I found is that I almost changed the lists back to use the loading spinner instead of the KTable skeletons by accident due to the component usage hiding what was going on. It's highly likely I'll delete the components used here seeing as they are just AppCollection wrappers. I would much rather it was clear what was happening all in one place and I'm pretty sure these components are not multi-use.

You can open these components and look at them. Putting everything in one file just so you can see all relevant code in that one file leads to these massive files which are increasingly hard to grok at a glance and the exact opposite of what you want it to do: having an overview of what’s relevant.

johncowen:
> You can open these components and look at them. Putting everything in one file just so you can see all relevant code in that one file leads to these massive files which are increasingly hard to grok at a glance and the exact opposite of what you want it to do: having an overview of what’s relevant.

Yeah for sure I can see your point, I think it very much depends on the individual case. In this case, I very much almost added the spinners back to these components instead of keeping the table skeletons, that to me made me think it could be clearer somehow. 🤔 maybe some comments around here might help? I might add those instead.

Thanks for the approval 👍

Issue #2387, opened 2024-04-08
Update all remaining instances of DataSource/ErrorBlock/LoadingBlock with DataLoader


johncowen:
### Description

A little while back we added a DataLoader component as a tool to manage the state of DataSource components and display relevant GUI.

We've used this now in quite a few areas, both simple and more complex interactions whilst adding new features to the GUI.  We should roll this _and_ `uri` out to the remaining places that haven't been updated as a consequence/whilst adding new features.

- [x] `control-planes`
- [ ] `data-planes` (also needs EnvoyData undoing)
- [x] `diagnostics`
- [x] `external-services`
- [ ] `gateways`
- [x] `kuma`
- [ ] `meshes`
- [ ] `onboarding`
- [ ] `policies`
- [x] `rules`
- [ ] `services`
- [ ] `subscriptions`
- [ ] `zone-egresses`
- [ ] `zone-ingresses`
- [ ] `zones`

johncowen:
I added a TODO list to the description to help record progress here, likely to be various PRs for one each TODO line.

johncowen:
Wanted to note that there has been good progress on this via work on other things (i.e. we've done some things in this task "in passing"). I had another little peek at what is outstanding and ticked off maybe on more section in the description above. Whilst a lot of the obvious areas have been upgraded there is still the odd bit that should be done.

Also, upgrading to use `uri` was a later addition and seems to be the biggest thing left.

Last note here, there is no rush to do this, but it will be nice once its all on the new stuff.

johncowen:
Searching the app there are no more ErrorBlocks in the app apart from the one in DataLoader. Which pretty much means we use DataLoader in the places we should. As noted in the last comment here 

> upgrading to use uri was a later addition and seems to be the biggest thing left.

> Last note here, there is no rush to do this, but it will be nice once its all on the new stuff.

i.e. we can just convert as we work on other stuff, if we want, its no biggie either way.

Closing this. Please reopen if you think we should have a specific issue to track this - IMO we don't

PR #2453, opened 2024-04-18
chore: Adds `uri` for generating typed string URIs for optional use with DataSource/Loader

johncowen:
~Putting up in draft for the moment in case folks want to play with it whilst I'm still working on it.~

Small chance of some more work to happen on here but any review comments/questions in the meantime are welcome 

---

Adds `RouteView::uri` for generating `toString`-able typed URIs (i.e. a string but not a string).

Closes https://github.com/kumahq/kuma-gui/issues/2432

See above 'Closes' issue for more information. 

- `DataSource#default` should emit `data: Data | undefined`
- `DataLoader#default` should emit `data: Data`
- `DataLoader#loadable` should emit `data: Data | undefined`

I've tried hard to try to:

- make sure that the entire data structure of the data is presented in the type hint rather than just the name of the alias to the structure
- ensure that you can still use a bare string with DataSources, but you don't get the types (you can continue to use our current approach), partly for backwards compat during rollout but partly incase there is something I've not thought of i.e. we have an escape hatch.

One thing I wanted to avoid but couldn't was providing `sources` as the first runtime argument to `uri`. It looks like you can't mix required generic parameters and inferred generic parameters in the same call with Typescript. I toyed with putting it as the last argument (so we can later easily drop it when we no longer need it), but figured you would probably naturally want this as the first argument, so you can then auto complete the second argument.

I think the other thing its worthwhile thinking about whilst reviewing this PR is that its super important that the user provides the same cache key/identifier/URI for different DataSources in different areas of the UI and this is why we go a little bit around the houses here to stick to the "you must provide a string URI to `:src` just like an ``" concept.

This PR should be finished up before continuing on with https://github.com/kumahq/kuma-gui/issues/2387

### DIFF
--- a/src/app/application/components/data-source/DataLoader.vue
+++ b/src/app/application/components/data-source/DataLoader.vue
@@ -1,9 +1,9 @@
 <template>
   <DataSource
     v-slot="{ refresh }"
-    :src="props.src"
+    :src="props.src as T"
     @change="(data) => srcData = data"
-    @error="(e) => srcError = e"
+    @error="(e: Error) => srcError = e"
   >
     <template
       v-if="allData.length > 0 && allData.every(item => typeof item !== 'undefined')"
@@ -15,8 +15,8 @@
       >
         <slot
           name="disconnected"
-          :data="props.src !== '' ? allData[0] : undefined"
-          :error="props.src !== '' ? allErrors[0] : undefined"
+          :data="srcData"
+          :error="allErrors[0]"
           :refresh="props.src !== '' ? refresh : () => {}"
         >
           <!-- KAlert -->
@@ -24,14 +24,14 @@
       </template>
       <slot
         name="loadable"
-        :data="props.src !== '' ? allData[0] : undefined"
-        :error="props.src !== '' ? allErrors[0] : undefined"
+        :data="srcData"
+        :error="srcError"
         :refresh="props.src !== '' ? refresh : () => {}"
       />
       <slot
         name="default"
-        :data="props.src !== '' ? allData[0] : undefined"
-        :error="props.src !== '' ? allErrors[0] : undefined"
+        :data="srcData as NonNullable<typeof srcData>"
+        :error="srcError"
         :refresh="props.src !== '' ? refresh : () => {}"
       />
     </template>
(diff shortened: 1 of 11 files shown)


src/app/application/components/data-source/DataLoader.vue
```diff
@@ -53,55 +53,60 @@
...
       >
         <LoadingBlock />
       </slot>
       <slot
         v-else
         name="default"
-        :data="props.src !== '' ? allData[0] : undefined"
-        :error="props.src !== '' ? allErrors[0] : undefined"
+        :data="srcData as NonNullable<typeof srcData>"
+        :error="srcError"
```
johncowen, line 73: I figured that you never use the exported data from `DataLoader` unless you have provided a `:src` (if you provide `data="[...]"` you already have access to the data in order to provide it in the first place). The DataLoader `error` state needs to use `allErrors[0]` though so you get a reference to the first error in a potential list of `:errors=[...]` that failed.

src/app/application/services/data-source/index.ts
```diff
@@ -31,6 +33,37 @@ export const defineSources = <T extends Record<PropertyKey, unknown>>(sources: E
...
+const cache = new Map()
+export const useUri = () => {
+  return <T extends SourceFactory, K extends keyof ReturnType<T>>(_typeRef: T, src: K, params: ExtractRouteParams<K>, query: Partial<PaginationParams> = {}) => {
+    const str = String(src)
+    if (!cache.has(str)) {
+      cache.set(str, compile(str))
+    }
+    const interpolate = cache.get(str)
+    const q = Object.entries(query).reduce<string[]>((prev, [key, value]) => prev.concat(`${key}=${value}`), [])
+    const uri = `${interpolate(params)}${q.length > 0 ? `?${q.join('&')}` : ''}`
```
johncowen, line 63: The above few lines are the only real meaningful runtime additions in this PR

1. Compile the path/uri/route (`/this/:thing`) into something you can interpolate with the variables using the `path-to-regexp` util
2. Add this to a cache so we only do this once per URI.
3. interpolate the route with the parameters
4. Potentially make a query string to add on the end.

The rest is just for Typescript

johncowen (commented):
Add a few inlines. Gonna take out of draft, although I'm still reviewing thinking on this one myself so there still might be the odd change.

kleinfreund (approved):

johncowen:
Will rebase this shortly. Just wanted to note that I'm still on the fence about putting the `sources` argument last instead of first. I'll merge this but I might change my mind before I roll this out.

PR #2455, opened 2024-04-18
ci: extract application error when annotating

kleinfreund:
**ci: add cypress-fail-on-console-error**

**ci: extract application error when annotating**

Make sure we’re handling application errors by stripping of the Cypress text (i.e. "Error: The following error originated from your application code, not from Cypress.") here first: https://github.com/kumahq/kuma-gui/actions/runs/8736436811/job/23971833245

```
     Error: The following error originated from your application code, not from Cypress.

  > Missing required param "service"

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.
```

## Screenshots

image

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -74,8 +74,13 @@ export default defineConfig({
         if (results && results.stats.failures > 0) {
           for (const test of results.tests.filter((test) => test.state === 'failed')) {
             if (test.displayError) {
-              const newLineIndex = test.displayError.indexOf('\n')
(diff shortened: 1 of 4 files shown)


kleinfreund:
Huh, strange. I deliberately introduced an error (adding a broken `<RouterLink :to="{ name: 'service-detail-view' }">Link</RouterLink>`) in the Zone create view but https://github.com/kumahq/kuma-gui/actions/runs/8736786887/job/23972394683 passed. That should really not happen. The test fails locally, also.

johncowen (approved):

PR #2456, opened 2024-04-18
refactor: removes DocumentationLink

johncowen:
Removes DocumentationLink and replaces with a 'typed' `XAction`.

There's a separate line of work in refactoring our Error/EmptyStates which needs to take into account their interactions with AppCollection so this kinda also bleeds into that work.

Pretty soon I plan on adding the implementation of `create` and `copy` `XAction` types, but for the moment they are noops.


src/app/x/components/x-icon/XIcon.vue
```diff
@@ -33,6 +33,7 @@ import {
   WarningIcon,
   PortalIcon,
   MeshIcon,
+  BookIcon,
```
kleinfreund, line 36: **Issue**: You should consider avoiding to create a icon wrapper component for reasons I lined out before. You’ll end up importing several icon components in one file which is bad for tree shaking. For example, let’s say you use a documentation link on one page: instead of just using the BookIcon code, you’ll also have to load all the code of the other icons currently in XIcon because they’ll be in the same JavaScript bundle. That’s the reason why modern icon libraries provide separate imports and bundles per icon. And that’s why I prefer the DocumentationLink component which doesn’t have that problem.

johncowen: We only import the icons we use in the application. This isn't a component meant for usage in other applications only our own. If we stop using an icon we just remove it from here and it will no longer be bundled.

Does that help? Maybe I don't understand the concern correctly?

kleinfreund: I’m not sure you’re following me.

On a page that uses _only_ BookIcon, the XIcon approach will cause a bundle that includes _all_ the code and icons found in XIcon to be loaded because they’re statically imported there. That causes more data to be loaded than necessary. The more icons you add to XIcon, the worse this will get.

johncowen: In what scenario will a user only ever visit one page in the application?

If there is a scenario, in that case would the user even care that a tiny SVG icon that might be on a page that they never visit? Even if I add another 10 icons here that are never used (which I'm not going to do), I doubt it would have a huge impact on load time.

Slightly related, I would actually like that our entire application code is loaded upfront in one go for all pages. I really don't like the pauses you get with no loading/error states in a incrementally loaded Vue application. Its extremely jarring as an initial (and most important) experience.

kleinfreund: > In what scenario will a user only ever visit one page in the application?

That doesn’t really matter. You shouldn’t develop a web application under the assumption any given user will always see a majority of its pages. First of all, that assumption generally won’t hold. Secondly, it favors a developer’s experience over a user’s by ignoring factors like JavaScript bundle sizes.

> Even if I add another 10 icons here that are never used (which I'm not going to do), I doubt it would have a huge impact on load time.

Maybe not, maybe yes. You might also end up with 20 or 30 icons in here. You might also develop more components of that style. Point is: it’s a bad practice to build wrapper components like this when there are better approaches. Here, there isn’t anything gained by putting the BookIcon inside XIcon.

> Slightly related, I would actually like that our entire application code is loaded upfront in one go for all pages.

😬

That is a very concerning statement in my opinion. That would have a negative impact on user experience. Application code tends to grow (certainly the case for this application) and so the total size of `dist/` gets bigger, too. Loading it all up front would lead to megabytes of data being loaded on every visit.

> I really don't like the pauses you get with no loading/error states in a incrementally loaded Vue application. Its extremely jarring as an initial (and most important) experience.

Indeed a real problem of single-page applications that needs a lot of careful attention to do away with. Server-side rendered applications have much less of that by nature. Importantly, they do so without requiring the entire application to be loaded up front.

---

Let’s get back to the PR: why does BookIcon need to be in XIcon?

johncowen: > That is a very concerning statement in my opinion. That would have a negative impact on user experience. Application code tends to grow (certainly the case for this application) and so the total size of dist/ gets bigger, too. Loading it all up front would lead to megabytes of data being loaded on every visit.

The thing is, that once you've loaded up all the library code and code that we reuse lots and barely changes (our "library code" like DataSource/DataLoader/XIcon/XAction etc), the application code is relatively small, its just markup/HTML (almost HTML, just it has extra minimal Vue bits attached).

If you are that worried about initial upfront loading time you should take a look at what's in the production built index.js files.

> Indeed a real problem of single-page applications that needs a lot of careful attention to do away with. Server-side rendered applications have much less of that by nature. Importantly, they do so without requiring the entire application to be loaded up front.

Yes but:

1. Our application isn't a Server-side rendered application
2. You can very easily add loading and error states to Server-side rendered applications, and this is the important thing, I can tell the user whats happening.

> Let’s get back to the PR: why does BookIcon need to be in XIcon?

Because I don't want an entirely different component just because the link happens to be a link to a documentation page. Its an "action" i.e. I click it to see some documentation, which therefore should just use `XAction`. In the same respect that I wouldn't have many many components called `<BlueText>Text</BlueText>` and `<BlackText>Text</BlackText>` I don't want many many components called `<LinkWithBookIcon />` etc etc. Its an `XAction` of a `doc` type, (or an action of a `create` type, or a `copy` type or a `reset`/`submit` type).

johncowen: I built a production deployment and served it over vite.preview like we were doing yesterday and saw this:

Screenshot 2024-04-19 at 11 58 11

It seems to me like all our icons (even the ones we don't use) plus a lot of other stuff we don't use is bundled into the initially loaded index.js file (~310kb gzipped):

> On a page that uses only BookIcon, the XIcon approach will cause a bundle that includes all the code and icons found in XIcon to be loaded because they’re statically imported there. That causes more data to be loaded than necessary. The more icons you add to XIcon, the worse this will get.

Therefore unless I'm mistaken, I don't think I'm making matters any worse in adding icons to XIcon

kleinfreund: Very well. I’m just pointing out what’s considered a bad pattern. The Vue bundling/"chunking" (?) including all of those icons probably has at least two causes: (1) some icons will simply have valid code paths from a lot of pages through things like empty and error states, Kongponents, etc. and (2) the build step making some decisions about what ends up in which bundle to optimize between the two extremes of "just one bundle for the entire app" and "one bundle for virtually every module".

It is also true that the initial bundle size loaded on app boot is already very big.



> Because I don't want an entirely different component just because the link happens to be a link to a documentation page. Its an "action" i.e. I click it to see some documentation, which therefore should just use `XAction`. In the same respect that I wouldn't have many many components called `<BlueText>Text</BlueText>` and `<BlackText>Text</BlackText>` I don't want many many components called `<LinkWithBookIcon />` etc etc. Its an `XAction` of a `doc` type, (or an action of a `create` type, or a `copy` type or a `reset`/`submit` type).

I mean that’s really just preference. If you want the docs link to use XAction, you could just include that in the component. Nothing prevents that. DocumentationLink isn’t even remotely similar to these BlueText and BlackText examples. It has an actual semantic purpose and also didn’t need much attention for a very long time. It works just fine and doesn’t need to be part of a bigger abstraction.

kleinfreund (approved):

src/app/x/components/x-icon/XIcon.vue
johncowen: Notice that its also preference of the authors of HTML for example `<button type="button/submit/reset">` not `ResetButton`, `SubmitButton`, `ButtonButton`, so seeing as they went that way I'm inclined to do the same.

And just to set your mind at rest, I would agree that its a bad pattern if the SVG size addition was huge and not cacheable, or we weren't already adding all of our SVGs to our index file. but in this case I think its a non-issue.

Thanks for the approval, gonna go ahead and merge.

I think theres a similar one over in https://github.com/kumahq/kuma-gui/pull/2452 if you wanted to take a look at that also.

PR #2465, opened 2024-04-19
chore: update mockServiceWorker file

kleinfreund:
This is an example for why it’s kind of annoying that Dependabot doesn’t "just" use `yarn install` (evidently) because its updates miss changed files like that.

### DIFF
--- a/public/mockServiceWorker.js
+++ b/public/mockServiceWorker.js
@@ -8,7 +8,7 @@
  * - Please do NOT serve this file on production.
  */
 
-const PACKAGE_VERSION = '2.2.13'
+const PACKAGE_VERSION = '2.2.14'
 const INTEGRITY_CHECKSUM = '26357c79639bfa20d64c0efca2a87423'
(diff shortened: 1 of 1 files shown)


johncowen:
I didn't totally read it myself but this might be of interrest

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions

> Dependabot creates pull requests to keep your dependencies up to date, and you can use GitHub Actions to perform automated tasks when these pull requests are created. For example, fetch additional artifacts, add labels, run tests, _or otherwise modifying the pull request_.

In any case, would be best to finish the switch of package manager before we look into this.

johncowen:
If I'm not mistaken following a brief look at the above link, we can listen for dependabot upgrade and then add a commit on the end to run `yarn install` (which would stop the flip-flopping).

Obvs if the npm switch avoid that also then great, but if not theres a possibility there.

kleinfreund:
Yep, we can consider adding some logic here after the npm migration 👍🏻

johncowen (approved):

PR #2487, opened 2024-04-24
chore: add build and install make commands

kleinfreund:
Add the commands `make install`, `make clean-install`, `make clean`, and `make build`. The main purpose of these commands is to provide a wrapper around the package manager-specific commands. This is needed for the create-gui-pr.yml workflow which currently runs `yarn run build`. Since release branches use the file from the `master` branch for this workflow (as it uses `workflow_dispatch`), they use `yarn run build`, too. If we switch to another package manager, create-gui-pr workflow runs triggered from release branches will start failing as they will then call, for example, `npm run build` while actually still using yarn. With the make commands, this can be solved.

Note: If we decide to merge this, I’ll back-port the Makefile setup to the 5 current release branches. Only then can create-gui-pr.yml be updated to use the make commands.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,6 @@ indent_size = 2
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[{Makefile,**.mk}]
+indent_style = tab
--- a/Makefile
+++ b/Makefile
@@ -2,3 +2,5 @@ SHELL := /usr/bin/env bash
 
 ## make help: if you're aren't sure use `make help`
 include mk/help.mk
+include mk/install.mk
+include mk/build.mk
--- a/mk/build.mk
+++ b/mk/build.mk
@@ -1,7 +1,7 @@
-.PHONY: build/sync
-build/sync:
-	make build
-
 .PHONY: build
 build:
 	yarn run build
+
+.PHONY: build/sync
+build/sync:
+	make build
--- a/mk/build.mk
+++ b/mk/build.mk
@@ -1,3 +1,7 @@
+.PHONY: build/sync
+build/sync:
+	make build
+
 .PHONY: build
 build:
 	yarn run build
--- a/mk/build.mk
+++ b/mk/build.mk
@@ -1,0 +1,3 @@
+.PHONY: build
+build:
+	yarn run build
--- a/mk/install.mk
+++ b/mk/install.mk
@@ -1,0 +1,11 @@
+.PHONY: install
+install:
+	yarn install
+
+.PHONY: clean-install
+clean-install:
+	yarn install --frozen-lockfile
+
+.PHONY: clean
+clean: ## Delete all node_modules directories
+	find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +
--- a/mk/install.mk
+++ b/mk/install.mk
@@ -2,9 +2,17 @@
 install:
 	yarn install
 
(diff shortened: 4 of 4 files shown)


johncowen (commented):
Super glad you're doing this! 🎉 

I had a couple of things to ask especially considering we should get everything done here before we backport it all those times.

---

First thing, not super important but could we add the following or similar to the `.editorconfig` file

```
[{Makefile,**.mk}]
indent_style = tab
```

---

I think we need to think about how this is going to grow, especially considering the multiple environments we have (at least `local` and `ci`) and, for this usecase, multiple branches.

Umm, before I talk through anything else, just to confirm: is `create-gui-pr` the only workfllow we need to cater for multi-branches using code from `master`?

--I'm kinda thinking we should make `install/sync` and `build/sync` targets (or similar) which in turn just call further targets (considering we'll never want to change the names of these ever again)

Might be worth hooking up offline to chat a little

kleinfreund:
I added a build/sync command now, but I don’t think install/sync makes much sense. Even in a workspace setup, you’d call `make install` whereas I can see how you might want to switch to something like `make build/other-project`.

johncowen:
> In any case, we can of course wrap make install by a make install/sync (and same for build) if you think it’s necessary that the make install command might change. Then again, I don’t see when it would be necessary to change that command where changing the underlying script (e.g. yarn install) wouldn’t work already.

Well we already have an additional `--frozen-lockfile` for the sync right?

Longer term, I'm torn between have the make targets know where they are via an env var (guessing GH gives some sort of `CI=true` env var?), but given that for this workflow we have the 5 times backport thing to think about I think having targets with super explicit "these ones are the ones for the sync which are painful if we need to change so don't use them for anything else" is maybe a good thing.

I think if just for these ones we use the some sort of `sync` naming which we only ever use for the sync (they probably just call other targets), then we can punt any other decisions to later.

---

ah gimme a sec I just saw your last comment, lemme look.

johncowen:
> There’s something to be said about this being the general standard to install.

Right yeah I almost said the same, but thats not really the point.

The main point is _if/when_ they need to be different they can be without have to do this after the fact (and then backport it all again)

johncowen:
Ah actually, the change would only happen on `master` seeing as the make file usage "disconnects" the "master workflow on a old branch" thing 🤔

kleinfreund:
> Ah actually, the change would only happen on `master` seeing as the make file usage "disconnects" the "master workflow on an old branch" thing 🤔 .

That’s right. _Only_ when you need to change the thing called in create-gui-pr.yml would you need to do a backport chain of PRs.



> I still think we use `install/sync` and `build/sync` they are just more explicit and "safer" into the future considering this weird "it always uses the master workflow on an older branch" thing.

Fine, I’ll add it. I still don’t think it’s useful/needed. 🤷🏻

johncowen:
I just took another look at the code and I think its just a case of changing the name of `clean-install` to `install/sync`? We don't need `clean-install` for anything else right?

kleinfreund:
> I just took another look at the code and I think its just a case of changing the name of `clean-install` to `install/sync`? We don't need `clean-install` for anything else right?

That doesn’t make sense to me. It’s circumstantial that there currently happens to be a 1 to 1 correspondence of `yarn install` is used in a local context and `yarn install --frozen-lockfile` is used in a CI context. Either command may have uses in local and CI contexts both.

PR #2503, opened 2024-04-25
chore(mk: update install makefiles (release-2.7)

kleinfreund:

### DIFF
--- a/mk/install.mk
+++ b/mk/install.mk
@@ -4,15 +4,7 @@ install:
 
 .PHONY: install/sync
 install/sync:
-	make install
-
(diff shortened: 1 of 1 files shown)


johncowen (approved):
ok so just to confirm: with all these approved, once they are in its only `master` that is incorrect?

https://github.com/kumahq/kuma-gui/blob/be68aadf480247c57827007cdaa6c8dc0cba6528/mk/install.mk#L1-L7

kleinfreund:
> ok so just to confirm: with all these approved, once they are in its only `master` that is incorrect?

Yep.

PR #2504, opened 2024-04-25
ci: use make commands in create-gui-pr workflow

kleinfreund:
ci: update create-gui-pr workflow to use make commands

chore(mk): update install makefiles


johncowen (approved):
Thanks! Much prefer to do this separate from the npm switch

Issue #2513, opened 2024-04-29
"Copy as K8s" no longer works


johncowen:
### What happened?

If you load a fresh page with a "Copy as K8s" button on it and then click the "Copy as K8s" button, nothing is copied into your clipboard (whatever was there beforehand remains).

This seems to be only on `master` not on `release-2.7` so I would imagine its as a result of https://github.com/kumahq/kuma-gui/pull/2374 and/or related Kongponents upgrades

kleinfreund:
Hm, I just tried it locally and it worked for me. This could be a browser-specific thing?

johncowen:
Maybe, I just tried it again, and unless I'm missing something/doing something wrong it's not working here

kleinfreund:
Hm, strange. I can somehow _sometimes_ reproduce it but not always?

johncowen:
Definitely something up here, we've both reproduced it and looks like there is something weird going on when we look deeper.

johncowen:
This still seems really broken, reopening.

Its also really confusing because since https://github.com/kumahq/kuma-gui/pull/2374 the `as K8s` button is actually a smaller button inside a button, so depending on where you click on the thing you can click one button or another.

PR #2517, opened 2024-04-29
fix: Copy as K8s

johncowen:
See https://github.com/kumahq/kuma-gui/pull/2374#discussion_r1560807919

Looks like this is one of those cases where we need to make sure we await for a promise and return its result rather than just return the promise to be awaited on elsewhere.

Probably why the original code was like:

https://github.com/kumahq/kuma-gui/blob/6c158bacad13f9562d9c31a300bd16008aaaa0a4/src/app/common/code-block/ResourceCodeBlock.vue#L101-L107

But we've lost that detail through lack some good comments and subsequent upgrades/refactors there. I'll probably either refactor this code slightly to make this clearer or add some clearer comments. Putting the PR up whilst I think about that a little.

As little sidenote: I really wanted to progress the dev time clipboard coping debug I added here https://github.com/kumahq/kuma-gui/pull/1981 into something we could also use for testing clipboard interactions. 

Unfortunately we removed the clipboard debugging here https://github.com/kumahq/kuma-gui/pull/2374, but this PR shows why it would be good to get that back in and then also progressed into something we can use for e2e tests.

Closes https://github.com/kumahq/kuma-gui/issues/2513

### DIFF
--- a/src/app/common/code-block/ResourceCodeBlock.vue
+++ b/src/app/common/code-block/ResourceCodeBlock.vue
@@ -102,6 +102,7 @@ let promise = new Promise((resolve: Resolve, reject) => {
 
 const fetcher = async (_params?: SingleResourceParameters): Promise<Entity> => {
   try {
+    // yes, this is one of those places where `return await` is important
     return await promise
   } finally {
     promise = new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


kleinfreund (approved):

johncowen:
I added the most minimal of comments here so we had something, I want to merge all our 3 CodeBlock components into one at some point anyway, so I guess I'll be back here at some point soonish, I can make things clearer then.

johncowen:
This wasn't the fix in the end see https://github.com/kumahq/kuma-gui/issues/2513#issuecomment-2093269940

Actual fix https://github.com/kumahq/kuma-gui/pull/2531

PR #2528, opened 2024-04-30
chore: add DataCollection#item, EmptyBlock:type

johncowen:
Adds an `item` slot to DataCollection, to be used instead of `:find="true"`

Previously when using `DataCollection:find="true"` you then also needed to do `items[0]` to always refer to the first and only item in the list. Now you can use the `#item="{ item }"` slot instead of `:find="true"`, eventually we can remove `:find` entirely.

---

Adds `EmptyBlock:type` and then uses that and makes use of `t` inside of a reusable component.

DataCollection also has `DataCollection:type` which is just passed through to `EmptyBlock`

There is a fair amount of decisions here based on supporting what we currently/used to have, whilst moving forwards with a different approach.

### DIFF
--- a/src/app/application/components/data-collection/DataCollection.vue
+++ b/src/app/application/components/data-collection/DataCollection.vue
@@ -1,55 +1,86 @@
 <template>
-  <slot
-    v-if="items.length === 0"
-    name="empty"
-    :items="items"
+  <template
+    v-if="slots.item"
   >
-    <EmptyBlock v-if="props.empty" />
-  </slot>
-  <slot
+    <template
+      v-for="item in [props.items.find(props.predicate)]"
+      :key="item"
+    >
+      <slot
+        v-if="item"
+        name="item"
+        :item="item as T"
+      />
+      <slot
+        v-else
+        name="empty"
+        :items="items"
+      >
+        <EmptyBlock
+          v-if="props.empty"
+          :type="props.type"
+        />
+      </slot>
+    </template>
+  </template>
+  <template
     v-else
-    name="default"
-    :items="paginated"
-  />
-
-  <slot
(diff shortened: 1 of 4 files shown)


src/app/application/components/data-collection/DataCollection.vue
```diff
@@ -1,55 +1,86 @@
...
+        :page-sizes="[15, 30, 50, 75, 100]"
+        @page-change="({ page }: PaginationChangeEvent) => {
+          emit('change', {
+            page,
+            pageSize: props.pageSize,
+          })
+        }"
+      />
+    </slot>
+  </template>
```
johncowen, line 63: A lot of the change here is just because I added an extra conditional/fork for the new `slots.item` which you'll see at the very top of this changeset.

src/app/application/services/i18n/I18n.ts
```diff
@@ -61,7 +61,7 @@ export default <T extends I18nRecord>(strs: T, env: Env['var']) => {
             // temporarily change any http.api terms to camelCase at runtime
             return camelCaseToWords((e as I18nError).key.split('.').pop()!)
           default:
-            if (rest[2]?.defaultMessage) {
+            if (typeof rest[2]?.defaultMessage === 'string') {
```
johncowen, line 64: Previously this being Javascript `''` would evaluate to `false`, meaning we couldn't use `''` as a default message when a i18n string wasn't found. We now specifically check for a string instead of truthyness.

You'll see where I use a `defaultMessage: ''` in EmptyBlock further down in this PR.

johncowen (commented):
Notes:

lukidzi (approved):

Issue #1448, opened 2023-09-07
improvement: Make dev scripts shareable


johncowen:
### Description

We have quite a few dev scripts inside `package.json` files, meaning they are harder to share/keep in sync.

`kuma` itself uses Makefiles to share/sync these types of dev time scripts. We should consider doing the same.

I would guess using a Makefile and a similar approach to kuma would generally make things easier to edit/add/organize including having a `make help` to list/describe/document targets.

Additionally we should try keep CI scripts inside the Makefile instead of the GH action yaml file (calling the Makefile from the yaml file)

Happy to hear any alternative approaches

johncowen:
More progress here https://github.com/kumahq/kuma-gui/pull/2691

johncowen:
This is complete, there is some follow up work to do. I'll link the issue here below thats tracking that.

PR #2529, opened 2024-04-30
chore: expands make usage with the more commonly used targets

johncowen:
Back in https://github.com/kumahq/kuma-gui/pull/2457 we added a very minimal Makefile based on kuma's and we plan on using this going forwards to give us a thin layer/abstraction over our various utility commands scripts.

This PR adds the minimal/commonly used targets so we can get used to using `make` as early as possible:

- installing (`make install`)
- running a dev server (`make run`)
- testing (`make test/unit` and `make test/e2e`)
- linting (`make lint`)

I've combined installing with running the dev server as installing it now super quick compared to yarn, but I'm not totally sure whether we would inadvertently update the lock file every so often with that. We'll see as we move forwards.

I've added some extra help to get started by way of basic node version checking and "escape hatch" type `run/%` which calls the corresponding package script so we don't have to have a make equivalent for every single package script, only the common ones. `make help` should be enough for folks to figure out how to get started.

I've continued to use include files given that we will have to inject different functionality depending on the environment. Idea being you will just need to use `make install` wherever you are and things should "Just Work(tm)"

Lastly, I updated our `DEVELOPER.md` to match and cleaned up our package scripts slightly and removed the scripts we don't no longer need.

Part of https://github.com/kumahq/kuma-gui/issues/1448


lukidzi (approved):

mk/install.mk
```diff
@@ -1,11 +1,13 @@
...
 .PHONY: install/sync
 install/sync:
 	npm clean-install
 
 .PHONY: clean
-clean: ## Delete all node_modules directories
+clean: ## Dev: Delete all node_modules directories
 	find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +
+
+
```
lukidzi, line 13: empty line to remove

mk/install.mk
```diff
@@ -1,11 +1,13 @@
 .PHONY: install
-install:
-	npm install
+install: check/node ## Dev: install all dependencies
+	@npm install
```
lukidzi, line 3: Should we install a specific version?

johncowen: A specific version of `node`?

At the moment at least we just warn and error out if you don't have the right version of node installed: 

https://github.com/kumahq/kuma-gui/pull/2529/files#diff-c6eeb169250244938a6aee582aa3134148d4c40d1a8f18ab793e67a735f166b8R6

It takes the pinned version from here

https://github.com/kumahq/kuma-gui/blob/master/.nvmrc

At the moment its up to the engineer to then make sure the correct version is installed (usually its calling `nvm`) but it could also be `asdf` or something else

PR #2545, opened 2024-05-12
chore: add eslint prod config

### LINKED ISSUE #1179: improvement: Add environment aware linting to prevent `console.logs` being left in code
### Description

We want to prevent console.logs from being left in production builds, but we do want to leave console.logs in the code whilst we are working.

We should make two different eslint config files so we can overwrite rules in a production/CI config via a development/local config.

The development config would disable or augment any console log rules (or other rules) in the production config, thus allowing adding console logs only during local development. This development config would then not be used in CI.

Additionally we should configure some git hooks to only run the production rules on commit, meaning they are unlikely to even reach CI. But if for whatever reason the git hook fails or does not get installed, then CI would catch any issues.

Also see https://github.com/kumahq/kuma-gui/pull/1140

schogges:
This is my first contribution to this project and I hope it's fine that I picked this issue 🙂 
My contribution includes the following changes:
1. fixed an eslint `import-order` issue in `src/app/x/index.ts`
2. added an eslint production config `.eslintrc.production.cjs`, which can be used in CI and a pre-commit hook. It extends the base config `.eslintrc.cjs` and sets the `no-console`-rule. I kept the original `.eslintrc.cjs` and explicitly disabled the `no-console` rule, although it didn't seem to be active yet. Since the current `.eslintrc.cjs` is taken as the config by default, it can be treated as the development version and ensures that there are no errors reported by the IDE during development. If this is not desired, I will change this according to your wishes. Another option could be to use an environment variable, e.g. `NODE_ENV`, and extend the configuration accordingly.
3. added [husky](https://typicode.github.io/husky/) for an easy pre-commit setup and further use. I hope this is fine, otherwise I could set this up via [pre-commit](https://pre-commit.com/). Note that if `ignore-scripts: true` is set, the `prepare` script will not be executed and must be run manually to complete the local setup for husky.

Closes #1179


schogges:
@johncowen I am not sure if this is intended, but would it actually make sense to change the `lint`-script and remove the `--fix`-flag such that it does not try to fix linting issues in ci? We could bring it back with a new script, e.g. `"lint:eslint:fix": "npm run lint -- --fix"` which can be used locally 🙂

johncowen:
Hey @schogges ! Thanks for dropping by!

I had a quick read of your PR description and comments and this sounds great!

re: your comment on the CI `--fix` if I understand correctly what you said makes perfect sense. If this adds pre-commit `fix`ing then thats much better than running that in CI 🙈 

I've not looked at the code properly here yet, so let me do that first incase I misunderstood and I'll come back to you. There might be a couple of over things you can help with here also (from what I remember we possibly don't need the `LINTER_MODE` stuff anymore)

Overall super glad we are getting pre-commit hooks here 🙌

schogges:
@johncowen thank you for your review and response 🙂 

> re: your comment on the CI `--fix` if I understand correctly what you said makes perfect sense. If this adds pre-commit `fix`ing then thats much better than running that in CI 🙈

currently it would fix linting issues in both the pre-commit and ci, but depending on what you prefer I can make sure to only fix them in pre-commit 👍 

> I've not looked at the code properly here yet, so let me do that first incase I misunderstood and I'll come back to you. There might be a couple of over things you can help with here also (from what I remember we possibly don't need the `LINTER_MODE` stuff anymore)

I'd be very happy to further improve the current changes based on your feedback and suggestions 🙂

schogges:
@johncowen thanks for the feedback and the suggestion for `lint-staged` 👍 I have chosen option 1 🙂  
Therefore I removed the `fix` from the `lint`-script, so during ci there will be no more `fix` happening. And as you suggested I added `lint-staged` to be used in the pre-commit hook. It will fix all fixable issues detected by `eslint` and `stylelint`.

schogges:
I have rebased this branch and re-installed `husky` and `lint-staged` with non-fixed versions.

johncowen:
Looks/sounds awesome, let me give it a whirl! brb

johncowen (approved):
Thanks, nice little dev improvement 👍

johncowen:
Hey @schogges , I went ahead and merged this one, hope you don't mind.

Do you fancy doing something else? I had something in mind you could do in the GUI application itself if you wanted to try that? Lmk and I can explain if so.

PR #2546, opened 2024-05-13
feat(meshservice): add mesh service listing and summary panel

johncowen:
Adds minimal "List plus Summary" views for MeshService. Minimal meaning the list only has a `name`, clicking on which shows a summary containing only the YAML config (which has the same functionality as all our resource YAML views). The plan is to then iterate on this in further PRs.

Note: there is no feature flagging here, all users will see the new `MeshServices` button. Please shout if this isn't desired, there is the possibility of flagging based on the output of `/config` if we wanted to do that.

Screenshot 2024-05-14 at 10 51 06

Screenshot 2024-05-14 at 10 51 22

### DIFF
--- a/src/app/services/views/MeshServiceListView.vue
+++ b/src/app/services/views/MeshServiceListView.vue
@@ -6,7 +6,7 @@
     <RouteView
       v-if="me"
       v-slot="{ route, t, uri }"
-      name="service-list-view"
+      name="mesh-service-list-view"
       :params="{
         page: 1,
         size: me.pageSize,
@@ -21,16 +21,15 @@
           >
             <h2>
               <RouteTitle
-                :title="t(`services.routes.items.title`)"
+                :title="t(`services.routes.mesh-service-list-view.title`)"
               />
             </h2>
           </XTeleportTemplate>
         </template>
         <KCard>
           <DataLoader
-            :src="uri(sources, '/meshes/:mesh/service-insights/of/:serviceType', {
+            :src="uri(sources, '/meshes/:mesh/mesh-services', {
               mesh: route.params.mesh,
-              serviceType: 'internal',
             },{
               page: route.params.page,
               size: route.params.size,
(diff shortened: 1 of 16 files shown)


src/app/services/views/ServiceListView.vue
```diff
@@ -14,133 +14,149 @@
...
+                      :name="route.params.service"
+                      :service="data?.items.find((item) => item.name === route.params.service)"
+                    />
+                  </SummaryView>
+                </RouterView>
+              </DataCollection>
+            </template>
+          </DataLoader>
+        </KCard>
+      </AppView>
```
johncowen, line 159: This file didn't really need changing so mcu, but as I used this as a copy/paste starting point for the new listing page, I upgraded it to use newer patterns first, then did the copy/paste.

src/app/services/views/MeshServiceListView.vue
```diff
@@ -0,0 +1,116 @@
...
+    v-slot="{ data: me }: MeSource"
+    src="/me"
+  >
+    <RouteView
+      v-if="me"
+      v-slot="{ route, t, uri }"
+      name="mesh-service-list-view"
+      :params="{
+        page: 1,
+        size: me.pageSize,
```
jijiechen, line 12:
```suggestion
        size: me.size,
```
Should it be `size`? I don't see `pageSize` defined anywhere.

src/app/services/views/ServiceListView.vue
```diff
@@ -14,133 +14,149 @@
...
+                  data-testid="service-collection"
+                  :headers="[
+                    { label: 'Name', key: 'name' },
+                    { label: 'Address', key: 'addressPort' },
+                    { label: 'DP proxies (online / total)', key: 'online' },
+                    { label: 'Status', key: 'status' },
+                    { label: 'Details', key: 'details', hideLabel: true },
+                  ]"
+                  :page-number="route.params.page"
+                  :page-size="route.params.size"
```
jijiechen, line 57: Oh, is `pageSize` defined here?

jijiechen (approved):

PR #2559, opened 2024-05-14
fix(onboarding): replaces KPops with SVG titles in onboarding

johncowen:
We noticed that the `KPop`s within onboarding weren't working due to using `KPop` inside of SVG.

We decided to replace them with an SVG title.

The diff probably looks more than it actually is but its just "take the text from the content of the KPop and put it in a `<title>` instead" - over and over again.


bartsmykla (approved):
LGTM

PR #2564, opened 2024-05-17
feat(meshservice): adds MeshService detail page

johncowen:
Direct Links:

- [Listing](https://deploy-preview-2564--kuma-gui.netlify.app/gui/meshes/default/services/mesh-services?page=1&size=50)
- [Detail](https://deploy-preview-2564--kuma-gui.netlify.app/gui/meshes/default/services/mesh-services/program-0/overview?page=1&size=50)

New things:

- Ports and Tags in listing rows
- Ports and Tags in the summary drawer
- New Detail > Overview page with Ports, Tags and Dataplane listing with search and summary drawer.
- New Detail > YAML page just repeating the same YAML from the summary drawer

Overall pretty happy we can build all this with almost zero Javascript 🎉 

### Listing

Screenshot 2024-05-17 at 14 09 36

### Listing with summary

Screenshot 2024-05-17 at 14 09 45

### Detail

Screenshot 2024-05-17 at 14 09 53

### Detail with summary

Screenshot 2024-05-17 at 14 09 59

### DIFF
--- a/src/app/services/data/index.ts
+++ b/src/app/services/data/index.ts
@@ -9,16 +9,22 @@ import type {
 export type ExternalService = PartialExternalService & {
   config: PartialExternalService
 }
-export type MeshService = PartialMeshService & {
+export type MeshService = Omit<PartialMeshService, 'spec' | 'status'> & {
   id: string
-  config: PartialMeshService
   namespace: string
+  labels: NonNullable<PartialMeshService['labels']>
   spec: {
     ports: NonNullable<PartialMeshService['spec']['ports']>
     selector: {
-      dataplaneTags: Record<string, string>
+      dataplaneTags: NonNullable<NonNullable<PartialMeshService['spec']['selector']>['dataplaneTags']>
     }
   }
+  status: {
+    addresses: NonNullable<PartialMeshService['status']['addresses']>
+    vips: NonNullable<PartialMeshService['status']['vips']>
+    tls: NonNullable<PartialMeshService['status']['tls']>
+  }
+  config: PartialMeshService
 }
 
 export type ServiceInsight = PartialServiceInsight & {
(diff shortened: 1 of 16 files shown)


johncowen:
Put back in draft, just realised I need to find dataplanes by the MeshService's `dataplaneTags` not just `kuma.io/service`

johncowen:
Since I originally PRed this, I've added:

- Addresses: Both in listing, summary and detail. I'm assuming there is usually one of these but could be multiple (they truncate if there are lots)
- VIPs: Only in the detail page, my thinking being these are more of an internal thing (let me know if this decision is bad). Again, assuming there is usually one of these but could be multiple (they truncate if there are lots)
- Zone: In the listing based on `labels['kuma.io/zone']` (similar to other areas of the GUI)

src/app/services/data/index.ts
```diff
@@ -55,13 +67,30 @@ export const MeshService = {
...
+          })(item.selector),
+        }
+      })(item.spec),
+      status: ((item = {}) => {
+        return {
+          tls: typeof item.tls !== 'undefined' ? item.tls : { status: '' },
+          vips: Array.isArray(item.vips) ? item.vips : [],
+          addresses: Array.isArray(item.addresses) ? item.addresses : [],
+        }
+      })(item.status),
```
johncowen, line 93: Lots of handwritten code here just to make sure things are set 🤔 , if we can't sort this some other way, it would be good longer term to see if we can automatically generate this sort of code, or at least stick to using a pattern that is common and easy to read and copy/paste throughout the application.

lahabana:
In details I think hostnames are more useful than ips

johncowen:
We _also_ show hostnames in the detail page, so both vips and addresses, (sorry I didn't update the screengrabs above after adding this new stuff)

Screenshot 2024-05-22 at 16 30 53

I'm not even totally sure if showing vips is any use, as far as I understand it they're for tproxy right?

lahabana (approved):

PR #2565, opened 2024-05-19
feat(data-planes): add and use nested DataPlanePoliciesSummaryView

### LINKED ISSUE #1916: Use policy summary tray for origin policies in Dataplane policies tab
### Description

For each origin in an entry shown in the Dataplane policies tab, we want to show a summary tray instead of going to the policy detail view directly.

schogges:
This change introduces the nested view for the `DataPlanePolicySummaryView` tray in the `DataPlanePoliciesView` instead of navigating to the `PolicyDetailView` directly.

- To keep the same style and layout for `PolicySummaryView` and `DataPlanePolicySummaryView` I have added ~~two~~ one new component~~s~~ by extracting ~~them~~ it from the current `PolicySummaryView`
  - ~~`PolicySummaryTitle` including the link to -> `PolicyDetailView`~~
  - `PolicySummary` for the content with a default slot for the `ResourceCodeBlock`
- Added browser tests

Screenshot 2024-05-19 at 23 27 34

Closes #1916


schogges:
@johncowen I hope I didn't miss it, but I couldn't find any browser test that covered the navigation from `DataPlanePoliciesView` to `PolicyDetailView` via "Origin Policy Link" and the browser tests are still passing after the change. In the issue you've mentioned that we are missing mock data, so I am wondering if it still makes sense to add something. During development I was able to test relatively reliably with the `MeshHTTPRoute` policy type.

johncowen:
Hey @schogges 👋 

Thanks for putting this up 🚀 

> @johncowen I hope I haven't overlooked it, but I couldn't find any browser test that covered the navigation from DataPlanePoliciesView to PolicyDetailView via "Origin Policy Link" and the browser tests are still passing after the change.

I would guess there were no browser tests for that as yet, it was just a link to another page previously so guessing thats why.

> In the issue you've mentioned that we are missing mock data, so I am wondering if it still makes sense to add something. 

Generally policies are pluggable and fully dynamic so we generally don't specifically add things for individual policy types, hence why we don't have them all mocked out. I quite often click around on the policy pages, see a 404 and think "oh! bug!" and then realize its not a bug at all, its because we don't have mocks for all policies 🤦 😆 . So I just wanted to point out that if you see 404 clicking around there, its just because of lack of mocks.

---

Not had a chance to look at this today unfortunately, will hopefully look first thing tomorrow 👍

schogges:
@johncowen thank you for the clarification 😄 there were a lot of 404s, but as mentioned I was able to test with the policy type `MeshHTTPRoute` and I also realised that in the browser tests I could mock the response of the requests individually 😃 therefore I have also added some tests for the new stuff.
I set this now to ready for review and looking forward to your review and feedback. Worth to mention is that I have never worked with Vue (I have a React background) and gherkin before and I hope that everything makes sense. I'd be happy to further improve based on your feedback 🙂

schogges:
@johncowen I have realised that some times there is an extra empty space that the `.slideout-content` could grow to and I could not find the reason yet 🤔 this doesn't seem to happen in the `PolicySummaryView` and the only difference in what is being used is the `DataSource` and `DataLoader` vs `DataCollection` and `EmptyBlock`. I'd further investigate that because it adds an additional and unnecessary scrollbar.
Screenshot 2024-05-21 at 10 44 57

features/mesh/dataplanes/DataplanePolicySummary.feature
```diff
@@ -0,0 +1,52 @@
...
+    Given the CSS selectors
+      | Alias                      | Selector                                                                   |
+      | summary-slideout-container | [data-testid='summary'] [data-testid='slideout-container']                 |
+      | summary-title              | $summary-slideout-container [data-testid='slideout-title']                 |
+      | summary-content            | $summary-slideout-container [data-testid='data-plane-policy-summary-view'] |
+  
+  Scenario: Policy Summary View has expected content
+    Given the environment
+      """
+      """
```
johncowen, line 12: If this is blank you shouldn't need it

src/app/data-planes/views/DataPlanePoliciesView.vue
```diff
@@ -163,17 +165,35 @@
                 </template>
               </div>
             </template>
+            <RouterView
+              v-if="route.params.policyPath && route.params.policy"
```
johncowen, line 169: When these update the entire view re-renders, which means that if you have any accordions open, they close again. If you rebase here you should be able to use `route.child()` instead i.e. "does this route have a child to show" if it does then show the summary view.

Once you've done that you should be able to remove lines 8 and 9 above. Which means the route has less dependencies and therefore re-renders less, and crucially, doesn't re-render when the `policy` or `policyPath` updates.

src/app/data-planes/views/DataPlanePolicySummaryView.vue
```diff
@@ -0,0 +1,74 @@
...
+            :title="t('policies.routes.item.title', { name: route.params.policy })"
+            :link-to="{
+              name: 'policy-detail-view',
+              params: {
+                mesh: route.params.mesh,
+                policyPath: route.params.policyPath,
+                policy: route.params.policy,
+              },
+            }"
+          />
```
johncowen, line 30: We usually lean towards avoiding creating components for this sort of thing. Reason being, its far easier to see what's happening if I just see the same that is in every other route I know its fine, whereas now I'm wondering "what is it extra that PolicySummaryTitle is doing 🤔 ", even though its doing nothing extra. Moreover its not really saving us from writing less code, which I think could have been the reasoning here.

i.e. sometimes copy/paste is fine - in fact sometimes its best.

I'd much rather see the h2 and application standard `RouteTitle` here, even tho that means a little bit of repetition.

src/app/data-planes/views/DataPlanePolicySummaryView.vue
```diff
@@ -0,0 +1,74 @@
...
+        </template>
+        <DataLoader
+          :data="[data]"
+          :errors="[error]"
+        >
+          <PolicySummary
+            v-if="data"
+            :policy="data"
+          >
```
johncowen, line 39: I can see there is a little more reason to add this as a reusable component and use it twice, and I like that you kept the `route` access outside the component by using a slot 👍 . Feel free to resolve this comment just wanted to point out I liked this.

johncowen (commented):
Sorry for the delay here, I left some comment.

I think most importantly the thing I noticed was that the accordion panes close if you click to open the summary panel. This was something we could (and should) fix elsewhere (https://github.com/kumahq/kuma-gui/pull/2581).

You should be able to rebase now and along with the comment I made on the `policyPath`/`policy` route params, it should prevent the accordions closing by avoiding a re-render of the route.

Let me know if you have any questions, I think this generally this is really nice considering you have very little context on the codebase 👍 

---

> @johncowen I have realised that some times there is an extra empty space that the .slideout-content could grow to and I could not find the reason yet 🤔 this doesn't seem to happen in the PolicySummaryView and the only difference in what is being used is the DataSource and DataLoader vs DataCollection and EmptyBlock. I'd further investigate that because it adds an additional and unnecessary scrollbar.

As long as its functionally ok I wouldn't worry too much about that in this PR, we can take a look at that later. There are some other things that aren't ideal about the code editor so I could look at those separately. Lets see where we get to with the accordion pane/close re-render thing first.

src/app/data-planes/views/DataPlanePoliciesView.vue
schogges: Ok, this is great 🙌 I removed the route params and use `route.child()` instead. However this wasn't enough. At least locally it still rerendered the whole route and closed the accordions. I was able to achieve the expected behaviour by relocating and duplicating the `RouterView` for the `SummaryView` into the different rule types. This worked out quite well for the `proxy` and `to` rules, but not for the `from` rules. Not sure what I am missing for them 😅 At least the route is not rerendered but the accordion closes when opening a policy summary from the `from` rules 🤔

src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: I fully agree 👍 though the main intention was consistency in terms for style and layout, I see that the benefit for this component was not enough to reason the abstraction. Removed it and brought back the `h2` and `RouterLink` to the respective policy views.

schogges:
@johncowen many thanks for your review and feedback ❤️ I rebased to the latest master and covered your change requests. Unfortunately the `lint-test` is failing because of some `console` methods being used in `src/app/x/components/x-copy-button/XCopyButtonDebug.vue`. Would you like me to fix them or will you take care of them?

src/app/data-planes/views/DataPlanePoliciesView.vue
johncowen: Ah sorry! Just noticed, the `v-if="route.child()"` needs to go on the `SummaryView`, so we always "render" the `RouterView` which is render-less, and conditionally render the `SummaryView` inside it if we have a route child.

I went back to 0ef4258ba25b35919005bf5171cb1bfd3aed67f9 and tried it there and it all seems fine.

I still don't completely get Vue's rendering algorithms sometimes 🤷‍♂️

johncowen:
> Edit: nvm, I guess I'll wait for https://github.com/kumahq/kuma-gui/pull/2582 being merged 😁

😅 yeah sorry about that! I felt a but bad taking out the split lint thing, but for now it makes things simpler I think, and personally I don't mind using the stricter one.

You should be able to rebase this and the fail should go away.

---

I also figured the extra detail for this, take a look above:

https://github.com/kumahq/kuma-gui/pull/2565#discussion_r1609868385

I 'think' after that we should be able to merge this

src/app/data-planes/views/DataPlanePoliciesView.vue
schogges: Perfect, that did the trick 👍 works like a charm now 😁

schogges:
> > Edit: nvm, I guess I'll wait for #2582 being merged 😁
> 
> 😅 yeah sorry about that! I felt a but bad taking out the split lint thing, but for now it makes things simpler I think, and personally I don't mind using the stricter one.
> 
> You should be able to rebase this and the fail should go away.

That's perfectly fine 😄 I agree that reducing it back to one config is a lot easier and more understandable 👍 

I have rebased and covered your change request from above. Seems to work as expected 🙂 

---

Once this is merged and you agree, I'd further investigate the issue I described above: https://github.com/kumahq/kuma-gui/pull/2565#issuecomment-2122111132 🙂

johncowen (approved):
Looks good! Thanks!

---

I think you you have perms to merge this yourself, if you do please make sure you Squash and Merge so we get one commit per PR. If you don't have the perms, I can merge it when I'm back in tomorrow.

---

On the scrollbar thing, I had a brief look and I noticed that you can get it to "double scroll" on other pages also if you make your browser window shorter, so not totally if what you are seeing is only on this page or not.

I think the thing to be careful of is (and now I think about it this is now ringing more bells with a previous conversation a few months back), we probably need to be able to scroll the main page and the summary panel as two separate scroll panes. The double scroll isn't ideal but I'm not sure what the alternative is.

Ah here we go found it https://github.com/Kong/kongponents/pull/2020#discussion_r1495941955

Up to you if you wanted to look into it, or an making issue is also fine if not. It could be part-related to our component library, I've noticed some other strange things with the code editor that will need looking at at some point. Anyway, I'll leave it up to you.

schogges:
> Looks good! Thanks!
> 
> I think you you have perms to merge this yourself, if you do please make sure you Squash and Merge so we get one commit per PR. If you don't have the perms, I can merge it when I'm back in tomorrow.

Thank you 🎉 need to wait for you to merge it, I have no write permissions 😄 

> On the scrollbar thing, I had a brief look and I noticed that you can get it to "double scroll" on other pages also if you make your browser window shorter, so not totally if what you are seeing is only on this page or not.
> 
> I think the thing to be careful of is (and now I think about it this is now ringing more bells with a previous conversation a few months back), we probably need to be able to scroll the main page and the summary panel as two separate scroll panes. The double scroll isn't ideal but I'm not sure what the alternative is.
> 
> Ah here we go found it [Kong/kongponents#2020 (comment)](https://github.com/Kong/kongponents/pull/2020#discussion_r1495941955)
> 
> Up to you if you wanted to look into it, or an making issue is also fine if not. It could be part-related to our component library, I've noticed some other strange things with the code editor that will need looking at at some point. Anyway, I'll leave it up to you.

Alright, I see 🤔 I will see what I can find 🙂 
In addition I have found that sometimes there are even three scrollbars. Below is a better depiction of the problem. The `body`, the `.slideout-container` (which includes the `.slideout-header`) and the `.slideout-content` are all scrollable. I believe that the `.slideout-container` should not be scrollable in these cases.

Screenshot 2024-05-22 at 23 52 49

johncowen:
Yikes 3 😆 , yeah we should defo look at that then, lemme know how it goes.

Merging this one 🎉

PR #2586, opened 2024-05-22
feature(zones): adds Zone Instance Id and versioning information

johncowen:
- Adds a Zone Instance Id to the title of the accordion bar, if this doesn't exist it will revert to use the GlobalControlPlanID as before.
- Adds ZoneCP version to the title of the accordion bar.


johncowen:
Kinda unsure whether I'm using the right versions here 🤔

johncowen:
Ok so I am using the correct version, but we need to do this slightly differently depending on application.

I'm gonna pull out the version notification thing so I can do it differently with that in mind, but leave the rest in so I can get this merged and move on/forward.

lahabana (approved):

PR #2589, opened 2024-05-26
fix(data-planes/DataPlanePolicySummaryView): extra spacing and scrollbar

### LINKED ISSUE #2588: Extra spacing and scrollbar in DataPlanePolicySummary
### What happened?

This is a follow-up issue to #2565 

**Description**

It sometimes happens that in the new `DataPlanePolicySummaryView` there is an extra spacing being added to the `.slideout-container`, which adds an additional and unnecessary scrollbar, ending up with up to 3 scrollbars in extreme cases (see depiction below). It highly depends on the viewport or how much space the data is taking up on the screen.

**Repro**

1. Go to `/gui/meshes/:mesh/data-planes/:dataPlane/policies/:policyPath/:policy`
2. Check the `.slideout-container` for having a scrollbar
3. Retry with other data or policy to reproduce this issue (e.g. refresh when running locally)

**Expected behaviour**

- ✅ The `body` shall remain scrollable when the height of the main content exceeds the height of the `body`
- ✅ The `.slideout-content` shall remain scrollable when the content exceeds the height of this element, being the only element in the `.slideout-container` that should be scrollable
- ❌ The `.slideout-container` shall not be scrollable

**Context**

Screenshot 2024-05-22 at 23 52 49

schogges:
This change makes sure that there is no extra spacing and scrollbar being added to the `.slideout-container` in the `DataPlanePolicySummaryView`.

The reason for the issue that adds the extra spacing which led to the extra scrollbar, is [an empty span](https://github.com/kumahq/kuma-gui/blob/9691ec3ab7ed877d5b10a802dff29076ba66304d/src/app/application/components/data-source/DataSource.vue#L7). It is part of the `DataSource` component and due to the [absolute positioning](https://github.com/kumahq/kuma-gui/blob/9691ec3ab7ed877d5b10a802dff29076ba66304d/src/app/application/components/data-source/DataSource.vue#L118) it was pulled up style-wise to the next element that has a `position` set, which is the `.slideout-container`'s `position: fixed` in this case. Therefore I've moved and surrounded the `DataSource` and `DataLoader` in the `DataPlanPolicySummaryView` with a container element that has `position: relative` set. This could also be solved by removing the empty `span` element from the `DataSource`, removing the `position: absolute` or by adding a `position` to an element in the `DataLoader` or `DataSource`, but this would have impacts on other areas too.

Closes #2588

### DIFF
--- a/src/app/application/components/data-source/DataSource.vue
+++ b/src/app/application/components/data-source/DataSource.vue
@@ -116,8 +116,8 @@ onBeforeUnmount(() => {
 // 'visually-hidden type' rule applied here so we are dependency free
 span {
   position: absolute !important;
-  width: 1px !important;
-  height: 1px !important;
+  width: 0px !important;
+  height: 0px !important;
   padding: 0 !important;
   margin: -1px !important;
   overflow: hidden !important;
--- a/src/app/data-planes/views/DataPlanePolicySummaryView.vue
+++ b/src/app/data-planes/views/DataPlanePolicySummaryView.vue
@@ -11,65 +11,63 @@
       codeRegExp: false,
     }"
   >
-    <AppView>
-      <template #title>
-        <h2>
-          <RouterLink
-            :to="{
-              name: 'policy-detail-view',
-              params: {
-                mesh: route.params.mesh,
-                policyPath: route.params.policyPath,
-                policy: route.params.policy,
-              },
-            }"
-          >
-            <RouteTitle
-              :title="t('policies.routes.item.title', { name: route.params.policy })"
-            />
-          </RouterLink>
-        </h2>
-      </template>
-      <div class="relative">
-        <DataSource
-          v-slot="{ data, error }: PolicySource"
-          :src="`/meshes/${route.params.mesh}/policy-path/${route.params.policyPath}/policy/${route.params.policy}`"
-        >
-          <DataLoader
-            :data="[data]"
-            :errors="[error]"
-          >
-            <PolicySummary
-              v-if="data"
-              :policy="data"
+    <DataSource
+      v-slot="{ data, error }: PolicySource"
+      :src="`/meshes/${route.params.mesh}/policy-path/${route.params.policyPath}/policy/${route.params.policy}`"
+    >
+      <AppView>
+        <template #title>
+          <h2>
+            <RouterLink
+              :to="{
+                name: 'policy-detail-view',
+                params: {
+                  mesh: route.params.mesh,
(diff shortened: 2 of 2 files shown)


schogges:
@johncowen Could you please explain to me the purpose of [the empty span](https://github.com/kumahq/kuma-gui/blob/9691ec3ab7ed877d5b10a802dff29076ba66304d/src/app/application/components/data-source/DataSource.vue#L7)? 🙂

schogges:
> Hey @schogges 👋
> 
> Sorry for the delay (been out for a couple of weeks)
> 
> > @johncowen Could you please explain what is the purpose of [the empty span](https://github.com/kumahq/kuma-gui/blob/9691ec3ab7ed877d5b10a802dff29076ba66304d/src/app/application/components/data-source/DataSource.vue#L7)? Looks like some Vue specific thing? 🙂
> > Depending on the purpose, another solution might make more sense 🤔
> 
> It's so when (if ever) we need to know if a DataSource is in the visible DOM or not using MutationObserver we can detect that (usecases being fetching data as you scroll/resize etc).
> 
> I think we should be good to just remove the span for the moment if its causing an issue. I don't think it will effect anything else if we do that, and if we need the mutation observer stuff later we can add something back in with better/less intrusive positioning.

Hi @johncowen 
I've also been out for the past week and sorry for the delay 🙂 

Thank you for the clarification. This actually sounds like a useful feature and I have tested other strategies instead of removing the `span`. A less intrusive way of using the `span` and keeping the functionality alive could be setting the `width` and/or `height` to `0px` or setting `display: none`. Please let me know what you think about this 🙂

johncowen:
Hey 👋 

If the width/height: 0px thing works for this then I think that would be best (I "think" using display: none would mean we wouldn't be able to detect when the DataSource comes into the viewport via MutationObserver).

So order or preference from me from most preferred to least preferred:

1. width/height: 0px
2. display: none
3. remove the span

Hopefully 1 works fine 👍

schogges:
> Hey 👋
> 
> If the width/height: 0px thing works for this then I think that would be best (I "think" using display: none would mean we wouldn't be able to detect when the DataSource comes into the viewport via MutationObserver).
> 
> So order or preference from me from most preferred to least preferred:
> 
> 1. width/height: 0px
> 2. display: none
> 3. remove the span
> 
> Hopefully 1 works fine 👍

I have rebased to the latest master and chose option 1. I could not reproduce the extra scrollbar anymore 🎉 😄

johncowen (approved):
Awesome, lets go with this!

PR #2607, opened 2024-06-10
chore(deps): kongponents/kpop/floating-ui upgrade

johncowen:
Kongponents/kpop/floating-ui upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "dependencies": {
     "@kong-ui-public/i18n": "^2.1.6",
     "@kong/icons": "^1.13.0",
-    "@kong/kongponents": "9.0.0-alpha.159",
+    "@kong/kongponents": "9.0.0-alpha.166",
     "@vueuse/core": "^10.10.0",
     "brandi": "^5.0.0",
     "deepmerge": "^4.3.1",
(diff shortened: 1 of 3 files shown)


src/app/data-planes/views/DataPlaneSummaryView.vue
```diff
@@ -76,7 +76,6 @@
                       >
                         <KTooltip
                           class="reason-tooltip"
-                          placement="bottomEnd"
```
johncowen, line 79: This was removed purposefully, in the upgrade the string was changed from `bottomEnd` to `bottom-end`, but seeing this is the only place where we use `placement`, I tried removing it to see if it made any difference, and it didn't.

FloatingUI probably has better positioning, and I'd rather not have to define/hardcode positioning on things anyway.

package.json
```diff
@@ -40,7 +40,7 @@
   "dependencies": {
     "@kong-ui-public/i18n": "^2.1.6",
     "@kong/icons": "^1.13.0",
-    "@kong/kongponents": "9.0.0-alpha.159",
+    "@kong/kongponents": "9.0.0-alpha.166",
```
johncowen, line 43: Note: We purposefully use an exact pin rather than a range pin here

lukidzi (approved):

jijiechen (approved):

Issue #2471, opened 2024-04-22
Integrate with MeshExternalService (externalService v2)


lahabana:
### Description

We'll need GUI improvements for this. @slonka and @lukidzi to add more info there once the spec is done.

johncowen:
Closed by https://github.com/kumahq/kuma-gui/pull/2619

Related potential follow on issue: https://github.com/kumahq/kuma-gui/issues/2470

PR #2619, opened 2024-06-11
feat(services): Adds support for MeshExternalService

johncowen:
Initial support for MeshExternalService

[Direct Link](https://deploy-preview-2619--kuma-gui.netlify.app/gui/meshes/default/services/mesh-external-services?page=1&size=50)

See https://github.com/kumahq/kuma-gui/issues/2471

### DIFF
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -620,6 +620,24 @@ export interface MeshService extends MeshEntity {
     addresses?: { hostname: string }[]
   }
 }
+export interface MeshExternalService extends MeshEntity {
+  type: 'MeshExternalService'
+  labels?: {
+    'kuma.io/display-name'?: string
+    'kuma.io/zone'?: string
+    'k8s.kuma.io/namespace'?: string
+    [key: string]: string | undefined
+  }
+  spec: {
+    match?: { type: string, port: number, protocol: string }
+    endpoints?: { address: string, port?: number }[]
+    tls?: { enabled: boolean }
+  }
+  status: {
+    vip?: { ip: string }
+    addresses?: { hostname: string }[]
+  }
+}
 
 export interface Zone {
   name: string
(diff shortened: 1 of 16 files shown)


src/types/index.d.ts
```diff
@@ -620,6 +620,24 @@ export interface MeshService extends MeshEntity {
...
+  labels?: {
+    'kuma.io/display-name'?: string
+    'kuma.io/zone'?: string
+    'k8s.kuma.io/namespace'?: string
+    [key: string]: string | undefined
+  }
+  spec: {
+    match?: { type: string, port: number, protocol: string }
+    endpoints?: { address: string, port?: number }[]
+    tls?: { enabled: boolean }
```
slonka, line 634: there are more fields in TLS, the whole spec is here: https://github.com/kumahq/kuma/blob/master/pkg/core/resources/apis/meshexternalservice/api/v1alpha1/meshexternalservice.go#L75-L87

slonka (commented):
One thing that I'm not sure of

src/types/index.d.ts
slonka: you can also check other parts of the schema

johncowen: Oh I think I'd decided to only show `tls.enabled` or not in the GUI, and forgot to do that! 😆  I can add that tho, thanks.

---

I'm only handwriting the TS for the props I need at the moment, I'm hoping very much that I can autogenerate these types very soon and therefore they'll be complete (and I won't have to handwrite them all)

johncowen: I added another commit on the end here adding TLS enabled/disabled everywhere within MeshExternalService. I also noticed that Port was missing on the detail page so I added that there also.

---

I'll be working separately on the openapi autogeneration, I doubt I'll be finished today so this PR shouldn't wait on it, I'm likely to PR it separately (also see https://github.com/kumahq/kuma/pull/10455)

slonka (approved):

PR #2620, opened 2024-06-12
feat(policies): support namespaced affected dataplanes

johncowen:
The policy detail page now looks like this:

Screenshot 2024-06-14 at 12 36 50

There are two tabs [Affected policies] and [YAML].  Affected policies are now in a table instead of a list.

I also removed the search, I understand that this list is likely to be small enough to not warrant a search (previously it didn't even have paging implemented, but it does now). If we do need `?name=` search I can implement that in a separate PR.

Support for https://github.com/kumahq/kuma/pull/10428


johncowen:
I switched to a table layout here 👍

lahabana (approved):
LGTM

Issue #1840, opened 2023-11-30
Re-enable Kongponents updates in Dependabot


kleinfreund:
### Description

Re-enable Kongponents update in Dependabot. They were disabled in https://github.com/kumahq/kuma-gui/pull/1832 and https://github.com/kumahq/kuma-gui/pull/1868.

Re-enable condition: Kongponents is installed using a non-prerelease tag (e.g. `v9.x.x`, not `v9.x.x-alpha.x`)

PR #2627, opened 2024-06-14
fix(deps): use correct semver constraint for kongponents alpha

johncowen:
Kongponents is currently going through an alpha series of release which we are following along with, these releases follow a `9.0.0-alpha.000` versioning scheme.

Kongponents also releases PR builds to the npm registry for every single PR, using a versioning scheme of `pr-<gh-pr-number>`.

If you use a range versioning for kongponents' alpha without additional constraints, it will mean that it will get updated to the _PR version_ rather than stay on the alpha version. (`pr-000` is deemed a newer version than `9.0.0-alpha`)

Usually we use caret/range versioning `^9.0.0-alpha.000`, but to prevent us being moving to a PR version for a while now we had used exact versioning for kongponents (`9.0.0-alpha.123` without the caret).

---

Up until now we've fixed this in the wrong place in various places:

- Using exact versioning for the kongponents version
- Disabling dependabot updates for kongponents

This has then caused several other issues:

- Not having automatic dependabot updates for kongponents meaning we need to 'remember' to update it.
- Not being able to specify a range semver not only means we don't get automatic tracking, but also means we are more likely to get duplicated versions of kongponents and other more general installation issues.
- I just hit another issue, which made me look to fix this correctly, when trying to install `@kong-ui-public/app-layout`:

Screenshot 2024-06-14 at 13 27 31

This above error happens because we current exact pin to `9.0.0-alpha.166` whereas `@kong-ui-public/app-layout` range pins its peerDependency to `^9.0.0-alpha.169` i.e. _at least_ alpha.169, therefore there is a clash between our exact version of 166 and the version that `app-layout` is requiring of _at least_ 169, even though I would guess that `@kong-ui-public/app-layout` doesn't require _at least_ 169, but probably a much lesser version of that.

https://github.com/Kong/public-ui-components/blob/main/packages/core/app-layout/package.json#L42

---

I tested this constraint on https://semver.npmjs.com/ which unfortunately doesn't use deeplinks so here are the screengrabs, notice there are no `pr` versions in the second screengrab:

Screenshot 2024-06-14 at 13 51 07

Screenshot 2024-06-14 at 13 51 15

---

Closes https://github.com/kumahq/kuma-gui/issues/1840

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,10 +7,6 @@ updates:
     open-pull-requests-limit: 10
     labels:
       - "dependencies"
-    ignore:
-      # TODO: Remove this once Kongponents v9 is used from a mainline release again (i.e. neither alpha or beta).
-      # Ignores Kongponents updates. This is necessary because Kongponents publishes prerelease versions for each pull request. Dependabot then creates PRs updating to `-pr.X.Y` which we definitely need to prevent. There doesn’t seem to be a way to update only the alpha version.
-      - dependency-name: "@kong/kongponents"
 
   - package-ecosystem: "github-actions"
     directory: "/"
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "dependencies": {
         "@kong-ui-public/i18n": "^2.1.6",
         "@kong/icons": "^1.13.0",
-        "@kong/kongponents": "9.0.0-alpha.166",
+        "@kong/kongponents": "^9.0.0-alpha < 9.0.0-beta",
         "@vueuse/core": "^10.10.0",
         "brandi": "^5.0.0",
         "deepmerge": "^4.3.1",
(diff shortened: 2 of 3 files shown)


michaelbeaumont (approved):

johncowen:
See https://github.com/kumahq/kuma-gui/pull/2636, specifically https://github.com/kumahq/kuma-gui/pull/2636/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L43

As we currently let dependabot edit our package file, it also bumps the constraint we specified here to from `< 9.0.0-beta` to `< 10.0.0` and then updates back to the `pr` version 😆 

I've PR'ed https://github.com/kumahq/kuma-gui/pull/2640 which hopefully means that we only give control of the lock file to dependabot instead of both the package and the lock file i.e. we re-take control of our dependencies.

Seeing as we always use caret ranges for versioning this means we get all non-breaking updates automatically, and we choose when we want to move to a breaking version instead of the bot. We can make use of `npm outdated` to check for updates to this.

I think this will be much better than our previous approach, but I'm going to make an issue to come back and re-evaluate this in a few months time once we have experienced what dependabot does over time.

PR #2657, opened 2024-06-18
chore(deps): remove babel

johncowen:
Seems to me like we no longer use babel.

1. I checked with some other projects and they don't directly depend on babel.
2. I checked a brand new vue project using vite and it doesn't directly depend on babel.
3. I added some severe errors to our babel.config.js file and tried running `make run` and `make build`, both targets ran fine. I would have expected some severe errors if we were using babel.
4. We don't use any crazy new javascript, everything we use should be supported natively in the browser, and we should continue to do that to avoid unnecessary config and dependencies.

Honestly this is super surprising to me, and I can't quite believe we have all this config/deps for something that isn't being used. I'm half guessing I've missed something here, but putting the PR up for visibility anyway.


bartsmykla (approved):
Let's get rid of it, everything seems to work and there is no `babel` mentioned nowhere explicitly. LGTM

PR #2658, opened 2024-06-19
fix: makes codeblock sticky toolbar use a variable height

johncowen:
We removed the stickiness on the toolbar here https://github.com/kumahq/kuma-gui/pull/2622 as it was based on a static value.

This lets the place that it sticks at be more configurable. I also added a background color to it using the dark theme (if we ever have a theme-able GUI this will either need to be moved to kongponents or made theme-able here)

### DIFF
--- a/src/app/common/SummaryView.vue
+++ b/src/app/common/SummaryView.vue
@@ -49,6 +49,9 @@ const emit = defineEmits<{
 </script>
 
 <style lang="scss" scoped>
+.summary-slideout {
+  --app-view-content-top: 0;
+}
 :deep(.slideout-title) {
   h1, h2, h4, h5, h6 {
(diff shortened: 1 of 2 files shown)


bartsmykla:
Example link to see the change: https://deploy-preview-2658--kuma-gui.netlify.app/gui/meshes/default/data-planes/circuit-proxy-0/xds-config

bartsmykla (approved):
LGTM

johncowen:
The other place for where it needed to be a variable sticking height is here:

https://deploy-preview-2658--kuma-gui.netlify.app/gui/meshes/default/data-planes/matrix-gateway_builtin-0/overview/outbound/monitor_svc_16207/stats

PR #2663, opened 2024-06-19
chore(deps): Upgrade kongponents 

johncowen:
See https://github.com/Kong/kongponents/pull/2157

### DIFF
--- a/features/zones/Delete.feature
+++ b/features/zones/Delete.feature
@@ -1,15 +1,15 @@
 Feature: zones / delete
+
   Background:
     Given the CSS selectors
-      | Alias          | Selector                                                       |
-      | items          | [data-testid="zone-cp-collection"]                             |
-      | item           | $items tbody tr                                                |
-      | actions-button | $item:nth-child(1) [data-testid='dropdown-trigger'] button     |
-      | delete-button  | $item:nth-child(1) [data-testid='dropdown-delete-item'] button |
-      | delete-prompt  | [data-testid="delete-zone-modal"]                              |
-      | confirm-button | $delete-prompt [data-testid='modal-action-button']             |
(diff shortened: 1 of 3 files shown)


features/zones/Delete.feature
```diff
@@ -1,15 +1,15 @@
...
-      | delete-button  | $item:nth-child(1) [data-testid='dropdown-delete-item'] button |
-      | delete-prompt  | [data-testid="delete-zone-modal"]                              |
-      | confirm-button | $delete-prompt [data-testid='modal-action-button']             |
-      | confirm-input  | $delete-prompt [data-testid='confirmation-input']              |
-
+      | Alias          | Selector                                                   |
+      | items          | [data-testid="zone-cp-collection"]                         |
+      | item           | $items tbody tr                                            |
+      | actions-button | $item:nth-child(1) [data-testid='dropdown-trigger'] button |
+      | delete-button  | $item:nth-child(1) [data-testid='dropdown-delete-item']    |
```
johncowen, line 9: The change here is the removal of `button`, which is no longer required because the kongponent update adds any additional props to the button itself now

The rest of the change here is just formatting

package.json
```diff
@@ -40,7 +40,7 @@
   "dependencies": {
     "@kong-ui-public/i18n": "^2.2.2",
     "@kong/icons": "^1.13.0",
-    "@kong/kongponents": "^9.0.0-alpha < 9.0.0-beta",
+    "@kong/kongponents": "^9.0.0-pr.2157.f80d6769.0",
```
johncowen, line 43: This needs updating to the actual version of kongponents once PR-2157 has been merged and released.

lukidzi (approved):

package.json
portikM: There will be no beta in v9 so maybe that constraint is not needed - no a blocker though

Issue #2623, opened 2024-06-13
subscriptions: Use table/panel GUI instead of accordion panes


johncowen:
### Description

Taken from https://github.com/kumahq/kuma-gui/issues/2164

> Finally each subscription should have a copy raw button to just get the json/yaml of the subscription.

We should switch from using accordion panes for subscriptions to using a table/summary panel layout like the rest of the GUI.

In the summary panel we can have 2 tabs, [Overview] and [YAML]. Overview shows the information that is currently showing in the accordion panes, and the YAML tab can contain the YAML for the subscription (along with copy buttons etc)

johncowen:
See https://github.com/kumahq/kuma-gui/pull/2664 for part of this

lahabana:
xref for 2 other implems:

https://github.com/kumahq/kuma-gui/issues/2920
https://github.com/kumahq/kuma-gui/issues/2921

Issue #2164, opened 2024-02-09
Improve zone inspect view


lahabana:
### Description

1. Since 2.6.0 there's a `zoneInstanceId` field which is more useful to users that the `globalInstanceId`. 

See:

```json
 "zoneInsight": {
  "subscriptions": [
   {
    "id": "792e2051-4f0c-4df6-ab03-1701b2b8af99",
    "globalInstanceId": "mink-control-plane-5c98999c7c-wsc2t-875e",
    "connectTime": "2024-02-06T12:38:04.230480952Z",
    "disconnectTime": "2024-02-06T16:14:13.958395810Z",
    "status": {
     "lastUpdateTime": "2024-02-06T15:14:54.374063306Z",
     "total": {
      "responsesSent": "45",
      "responsesAcknowledged": "45"
     }
    },
    "version": {
     "kumaCp": {
      "version": "2.6.0",
      "gitTag": "2.6.0",
      "gitCommit": "53744108710a5151b4b8121f7b9392ffc0f3651c",
      "buildDate": "2024-02-01T13:41:52Z",
      "kumaCpGlobalCompatible": true
     }
    },
    "generation": 86,
    "authTokenProvided": true,
    "zoneInstanceId": "kong-mesh-control-plane-84ccc98fc4-6zzzv-feca"
   },...
```
We should show this instead of the Global InstanceId as this is more useful
<img width="1464" alt="image" src="https://github.com/kumahq/kuma-gui/assets/702355/97aff79c-49f7-4594-ad9a-821b24eaff3d">

2. We should also show: the version at the top. Finally each subscription should have a `copy raw` button to just get the json/yaml of the subscription. We should also show if the version is not the latest to suggest upgrade to customers.
3. We should make it obvious when the zone is returning NACKs (I think we may need a backend change to be able to do this).

johncowen:
- ZoneInstanceID and Version in the accordion title: https://github.com/kumahq/kuma-gui/pull/2586

Still to do:

- Finally each subscription should have a copy raw button to just get the json/yaml of the subscription.
- We should also show if the version is not the latest to suggest upgrade to customers.
- We should make it obvious when the zone is returning NACKs (I think we may need a backend change to be able to do this).

lahabana:
[...]

PR #2664, opened 2024-06-20
feat(zones): move KDSSubscriptions to use a table/summary

johncowen:
Changes Zone Subscriptions only to use a table/summary view/interaction.

We decided that there would be far more space to show things if we used our common table/summary panel pattern instead of the accordion where there is very little space to show things:

Screenshot 2024-06-20 at 12 37 34

Screenshot 2024-06-20 at 12 37 44

See:

- https://github.com/kumahq/kuma-gui/issues/2623
- https://github.com/kumahq/kuma-gui/issues/2164

---

~Questions:~

~1. Do we want to show the Subscription YAML, the Subscription JSON config, or both?~
~2. If we want to show both, should we add a separate `[YAML]` tab or use the `Structured/YAML` dropdown?~

Note:

I'm guessing we'll want the same layout for DPP subscriptions, I plan to do this as a follow up and there will some changes made on top of here to make the panel work in both places.

Closes https://github.com/kumahq/kuma-gui/issues/2704

### DIFF
--- a/src/app/subscriptions/data/index.ts
+++ b/src/app/subscriptions/data/index.ts
@@ -26,7 +26,7 @@ export const DiscoverySubscriptionCollection = {
 }
 export const SubscriptionCollection = {
   fromArray: <T extends Subscription>(items?: T[]): SubscriptionCollection<T> => {
-    const subscriptions = Array.isArray(items) ? items : []
+    const subscriptions = Array.isArray(items) ? items.map(item => item) : []
 
     // make a copy of the original array so we don't change its order
     const subs = subscriptions.slice()
--- a/src/app/subscriptions/routes.ts
+++ b/src/app/subscriptions/routes.ts
@@ -1,0 +1,23 @@
+import type { RouteRecordRaw } from 'vue-router'
+export const routes = (): RouteRecordRaw[] => {
+  return [
+    {
+      path: 'subscription/:subscription',
+      name: 'subscription-summary-view',
+      redirect: { name: 'subscription-summary-overview-view' },
+      component: () => import('@/app/subscriptions/views/SubscriptionSummaryView.vue'),
+      children: [
+        {
+          path: 'overview',
+          name: 'subscription-summary-overview-view',
+          component: () => import('@/app/subscriptions/views/SubscriptionSummaryOverviewView.vue'),
+        },
+        {
+          path: 'config',
+          name: 'subscription-summary-config-view',
+          component: () => import('@/app/subscriptions/views/SubscriptionSummaryConfigView.vue'),
+        },
+      ],
+    },
+  ]
+}
--- a/src/app/subscriptions/views/SubscriptionSummaryOverviewView.vue
+++ b/src/app/subscriptions/views/SubscriptionSummaryOverviewView.vue
@@ -1,0 +1,145 @@
+<template>
+  <RouteView
+    name="subscription-summary-overview-view"
+    v-slot="{ t }"
+  >
+    <AppView>
+      <div
+        class="stack-with-borders"
+      >
+        <DefinitionCard
+          layout="horizontal"
+        >
+          <template #title>
+            Version
+          </template>
+
+          <template #body>
+            <template
+              v-for="version in [props.data.version?.kumaCp?.version]"
+            >
+              {{ version ?? '-' }}
+            </template>
+          </template>
+        </DefinitionCard>
+        <DefinitionCard
+          layout="horizontal"
+        >
+          <template #title>
+            Connected
+          </template>
+
+          <template #body>
+            {{ t('common.formats.datetime', { value: Date.parse(props.data.connectTime ?? '') }) }}
+          </template>
+        </DefinitionCard>
+        <DefinitionCard
+          v-if="props.data.disconnectTime"
+          layout="horizontal"
+        >
+          <template #title>
+            Disconnected
+          </template>
+
+          <template #body>
+            {{ t('common.formats.datetime', { value: Date.parse(props.data.disconnectTime) }) }}
+          </template>
+        </DefinitionCard>
+        <DefinitionCard
+          layout="horizontal"
+        >
+          <template #title>
+            Total Responses (sent/ack'ed)
+          </template>
+
+          <template #body>
+            <template
+              v-for="responses in [props.data.status?.total ?? {}]"
+            >
+              {{ responses.responsesSent }}/{{ responses.responsesAcknowledged }}
+            </template>
+          </template>
+        </DefinitionCard>
+        <DefinitionCard
+          v-if="props.data.zoneInstanceId"
+          layout="horizontal"
+        >
+          <template #title>
+            Zone Leader Instance ID
+          </template>
+
+          <template #body>
+            {{ props.data.zoneInstanceId }}
+          </template>
+        </DefinitionCard>
+        <DefinitionCard
+          v-else
+          layout="horizontal"
+        >
+          <template #title>
+            Global Instance ID
+          </template>
+
+          <template #body>
+            {{ props.data.globalInstanceId }}
+          </template>
+        </DefinitionCard>
+        <DefinitionCard
+          layout="horizontal"
+        >
+          <template #title>
+            ID
+          </template>
+
+          <template #body>
+            {{ props.data.id }}
+          </template>
+        </DefinitionCard>
+      </div>
+      <div
+        class="mt-8 stack-with-borders"
+      >
+        <div>
+          <slot name="default" />
+        </div>
+        <DefinitionCard
+          class="mt-4"
+          layout="horizontal"
+        >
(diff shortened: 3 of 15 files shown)


johncowen:
> 1. Do we want to show the Subscription YAML, the Subscription JSON config, or both?
> 2. If we want to show both, should we add a separate [YAML] tab or use the Structured/YAML dropdown?

I figured we probably don't want to show the YAML, its not like you'd want to use it for anything. If that incorrect we can always add as a follow up.

src/app/subscriptions/routes.ts
```diff
@@ -0,0 +1,23 @@
...
+      redirect: { name: 'subscription-summary-overview-view' },
+      component: () => import('@/app/subscriptions/views/SubscriptionSummaryView.vue'),
+      children: [
+        {
+          path: 'overview',
+          name: 'subscription-summary-overview-view',
+          component: () => import('@/app/subscriptions/views/SubscriptionSummaryOverviewView.vue'),
+        },
+        {
+          path: 'config',
```
lahabana, line 16: This needs to be double checked (@jakubdyszkiewicz will know) but AFAIR only the first subscription has a config as it's the one already shown in the `Config` tab in the overview. Therefore the tab should only show the overview

src/app/zones/views/ZoneDetailView.vue
```diff
@@ -102,14 +106,79 @@
...
-
-            <KCard class="mt-4">
-              <SubscriptionList
-                :subscriptions="props.data.zoneInsight.subscriptions"
+            <AppCollection
+              :headers="[
+                { ...me.get('headers.name'), label: 'Name', key: 'name' },
+                { ...me.get('headers.connected'), label: 'Connected', key: 'connected' },
+                { ...me.get('headers.disconnected'), label: 'Disconnected', key: 'disconnected' },
+                { ...me.get('headers.responses'), label: `Responses (sent/ack'ed)`, key: 'responses' },
```
lahabana, line 114: I think I'd add the version as well. It's useful when you want to follow when you are upgrading

lahabana (commented):
Some nits but I like the new view it looks more powerful

src/app/zones/views/ZoneDetailView.vue
```diff
@@ -102,14 +106,79 @@
             v-if="props.data.zoneInsight.subscriptions.length > 0"
           >
             <h2>{{ t('zone-cps.detail.subscriptions') }}</h2>
-
-            <KCard class="mt-4">
-              <SubscriptionList
-                :subscriptions="props.data.zoneInsight.subscriptions"
+            <AppCollection
+              :headers="[
+                { ...me.get('headers.name'), label: 'Name', key: 'name' },
```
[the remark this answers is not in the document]
johncowen: Gotcha, will sort 👍

src/app/subscriptions/data/index.ts
```diff
@@ -26,7 +26,7 @@ export const DiscoverySubscriptionCollection = {
 }
 export const SubscriptionCollection = {
```
lahabana, line 27: I think it's worth remembering that there are multiple subscriptions with different info. This is KDSSubscription but there are also DiscoverySubscription that are shown in there: https://deploy-preview-2664--kuma-gui.netlify.app/gui/meshes/default/data-planes/protocol-gateway_delegated-0/overview

johncowen: Yes definitely, tbh I'm guessing this code here will change a little to make it cope with both types, I made a half thought out reference to it in the PR desc also

> I'm guessing we'll want the same layout for DPP subscriptions, I plan to do this as a follow up and there will some changes made on top of here to make the panel work in both places.

Its kinda in the back of my head, and doing this will be coming up next 👍

src/app/zones/views/ZoneDetailView.vue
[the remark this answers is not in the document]
johncowen: I just noticed while doing this: A few days back I think we decided to change this piece of text to `Zone Leader instance ID`. So I'm 99% sure this column should be the same.

I'll update the screengrab below to use `Zone Leader instance ID` in the column also (instead of just `Zone Instance ID`)

Shout if that ^ isn't correct, otherwise I'll resolve this once I'm done.

Screenshot 2024-07-30 at 16 29 32

src/app/zones/views/ZoneDetailView.vue
johncowen: Adding the below extra column

Screenshot 2024-07-30 at 16 36 17

src/app/subscriptions/routes.ts
johncowen: Ahhh right gotcha, was super unsure about this (see questions in PR I crossed out 🫤 )

Looking at the issue this PR will hopefully close (https://github.com/kumahq/kuma-gui/issues/2164), it says this:

> We should also show: the version at the top. Finally each subscription should have a copy raw button to just get the json/yaml of the subscription. We should also show if the version is not the latest to suggest upgrade to customers.

So this means we want to show the subscription YAML itself not its JSON config I think, will check in with @jakubdyszkiewicz also, but I'm pretty sure this is what we need to do.

src/app/zones/views/ZoneDetailView.vue
johncowen: This is done like the above screengrab, resolving

src/app/subscriptions/routes.ts
johncowen: I checked in with @jakubdyszkiewicz offline, generally it is the case that only one subscription has a `.config` field.

So I'm pretty sure I just misunderstood the original ask.

I've already changed this so we show the subscriptions YAML in the tab instead, so I'm gonna resolve this convo.

but shout if there's anything else to do here.

johncowen:
K this one has had a few alterations after a first round of reviews, good to go again now 👍

src/app/zones/views/ZoneDetailView.vue
```diff
@@ -102,14 +106,85 @@
             v-if="props.data.zoneInsight.subscriptions.length > 0"
           >
             <h2>{{ t('zone-cps.detail.subscriptions') }}</h2>
-
-            <KCard class="mt-4">
-              <SubscriptionList
-                :subscriptions="props.data.zoneInsight.subscriptions"
+            <AppCollection
+              :headers="[
+                { ...me.get('headers.zoneInstanceId'), label: 'Zone Leader Instance ID', key: 'zoneInstanceId' },
```
lahabana, line 111: Btw should these labels be i18n?

johncowen: Glad you asked, I keep meaning to find the best place to mention this.

For the moment I've kinda given up doing i18n texts unless its useful for machine > human.

Our locale files are a bit of a mess as there is no consistency, lots of repetition and no real thought gone into them. I'm kinda avoiding adding to that until we have a change to put a bit more thought into it.

It will be easier once to add these to the i18n files once we are following a pattern to avoid repetition, whereas as the locale files get bigger and bigger it will be harder and harder to start following some sort of pattern.

I think I should probably put something in an issue actually, wdyt?

src/app/subscriptions/views/SubscriptionSummaryOverviewView.vue
```diff
@@ -0,0 +1,145 @@
...
+        <DefinitionCard
+          v-else
+          layout="horizontal"
+        >
+          <template #title>
+            Global Instance ID
+          </template>
+
+          <template #body>
+            {{ props.data.globalInstanceId }}
```
johncowen, line 84: We want to make it so that the mock results in this being shown sometimes

src/app/subscriptions/views/SubscriptionSummaryOverviewView.vue
```diff
@@ -0,0 +1,145 @@
...
+          <template #title>
+            Zone Leader Instance ID
+          </template>
+
+          <template #body>
+            {{ props.data.zoneInstanceId }}
+          </template>
+        </DefinitionCard>
+        <DefinitionCard
+          v-else
```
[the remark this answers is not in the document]
johncowen: Yeah this should show both zoneInstanceId and globalInstanceId

src/app/zones/views/ZoneDetailView.vue
johncowen: Lets keep i18ning things, structure isn't ideal at the minute but we can sort that out separately

johncowen: I made a couple of issues off the back of this:

- https://github.com/kumahq/kuma-gui/issues/2814

lahabana:
LGTM do we have issues to cover the same work for dataplane and zone proxies? If not can we open them and milestone them for 2.10 please?

lahabana (approved):

Issue #1639, opened 2023-10-22
Filter on policies list


johnharris85:
### What happened?

Should be able to filter on various fields on each policy list display

jakubdyszkiewicz:
> Triage: let's do filtering by name for now. The backend already supports this.

kleinfreund:
Note for later reference, I just tried a couple of APIs locally and the `name` prefix filter works for:

- /zones
- /zones/_overview
- /meshes
- /meshes/:mesh/dataplanes
- /meshes/:mesh/:policyTypePath

and doesn’t work for:

- /meshes/:mesh/service-insights

lahabana:
Needs to be done after: https://github.com/kumahq/kuma/issues/8908

johncowen:
Policy filtering was added in https://github.com/kumahq/kuma-gui/pull/2666

We should probably close this and make separate issues for any other areas where we want name based filtering.

---

Separately if we want to add something more than name based searching, we should again make a separate issue as I'm guessing that would need backend support?

lahabana:
Superseeded by: https://github.com/kumahq/kuma-gui/issues/3134

PR #2666, opened 2024-06-20
feat(policies): add name based search to policy listing

johncowen:
See https://github.com/kumahq/kuma-gui/issues/1639 (not clear whether we should close that seeing as it mentions doing search elsewhere other than policies)

Adds name based searching to the policy listings. Follows exactly the same approach we use for simple searching as elsewhere.

### Searching with result: 

Screenshot 2024-06-20 at 15 13 01

### Searching with no result: 

Screenshot 2024-06-20 at 15 13 08

### DIFF
--- a/src/app/policies/views/PolicyListView.vue
+++ b/src/app/policies/views/PolicyListView.vue
@@ -5,7 +5,7 @@
   >
     <RouteView
       v-if="me"
-      v-slot="{ route, t, can }"
+      v-slot="{ route, t, can, uri }"
       name="policy-list-view"
       :params="{
         page: 1,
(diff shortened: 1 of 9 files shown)


src/app/policies/views/PolicyListView.vue
```diff
@@ -27,158 +27,205 @@
...
+                          type="search"
+                          appearance="search"
+                          :value="route.params.s"
+                          :debounce="1000"
+                          @change="(e) => route.update({
+                            s: e,
+                          })"
+                        />
+                      </form>
+                    </search>
```
johncowen, line 108: This change above here in this file is the real change, the rest is mainly a result of re-nesting.

bartsmykla (approved):

PR #2676, opened 2024-06-21
chore(zones): add k8s/universal icons to zone rows instead of type

johncowen:
Uses icons/tooltips to show zoneCP type instead of an additional column.

I added them both on the zone listing on the global control plane overview page and the zone listing page.

---

Screenshot 2024-06-21 at 13 45 26

---

Screenshot 2024-06-21 at 13 45 40

### DIFF
--- a/features/zones/zone-cps/Index.feature
+++ b/features/zones/zone-cps/Index.feature
@@ -1,4 +1,5 @@
 Feature: zones / index
+
   Background:
     Given the CSS selectors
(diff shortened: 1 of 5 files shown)


bartsmykla:
It seems there is a small problem with hover state: 
image

johncowen:
@bartsmykla hover state should be better now, lmk when you get a sec 👍

bartsmykla (approved):
looks good now :-)

PR #2698, opened 2024-06-26
chore: creation of 'service-mesh' module

johncowen:
This PR is essentially one of those "move things around" PRs. In this case we are moving some service container configuration around so that:

- We have a `service-mesh` module that installs `zones`, `meshes`, `control-planes` at once.
- We can work our way to getting to the point that we can remove our top-level `services` folder (which is a confusingly named top-level folder). There are still a few more things to remove/move here but I felt like a good place to stop and PR now.

If you look at `main.ts` now the idea is you install your framework (`vue`), then the `application`, then the `service-mesh` module. Followed by an potentially optional modules (in this case `onboarding` and `diagnostics`)

As we are just moving things around here, there is no change to functionality/logic at all


lukidzi (approved):

johncowen:
Please do not merge yet, I'm waiting for other things to go in first

lukidzi (approved):

Issue #2468, opened 2024-04-19
Make table columns resizable


johncowen:
### Description

Kongponents recently added the ability to make the columns of KTable resizable/hideable.

We should see where we can make use of this.

johncowen:
Oh hey @miko-bautista 👋 I'm sure that will be fine! Thanks!

miko-bautista:
@johncowen Thank you for waiting! This is now **unblocked**! 🙌

PR #2727, opened 2024-07-04
chore: rollout table column resizing everywhere

johncowen:
Roll out of the code in https://github.com/kumahq/kuma-gui/pull/2625 to all our tables.

All tables should now have resizable columns that remember the sizes through browser refreshes (via `localStorage`)

Whilst the changeset looks big due to the amount of tables we have, the work is basically:

- Remove the `me` `DataSource` in favour of `RouteView::me`
- Add a `@resize` event listener to`AppCollection`
- Splat down the values from `me` i.e. `headers.headerName` onto the header JSON for the table/AppCollection.

There is a lot of green in the changeset due to the removal of the top-level `DataSource` component the the re-nesting that then occurs. I made some inline comments in the place that weren't specifically just rollout of the same thing again and again.

Part of https://github.com/kumahq/kuma-gui/issues/2468

### DIFF
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -1,7 +1,9 @@
 <template>
   <DataSource
-    v-slot="{ data: me }: MeSource"
-    :src="`/me/${props.name}`"
+    v-slot="{ data: me }"
+    :src="uri(sources, '/me/:route', {
+      route: props.name,
+    })"
   >
     <div
       class="route-view"
@@ -19,7 +21,7 @@
       <DataSink
         v-if="me"
         v-slot="{ submit }"
-        :src="`/me/mesh-list-view`"
+        :src="`/me/${props.name}`"
       >
         <slot
           :id="UniqueId"
@@ -58,7 +60,7 @@ import {
   beforePaint,
 } from '../../utilities'
 import { useUri } from '@/app/application/services/data-source'
-import type { MeSource } from '@/app/me/sources'
+import { sources } from '@/app/me/sources'
 import { useEnv } from '@/utilities'
 import { get } from '@/utilities/get'
 import type { RouteRecordRaw } from 'vue-router'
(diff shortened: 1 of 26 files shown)


src/app/application/components/route-view/RouteView.vue
```diff
@@ -19,7 +21,7 @@
       <DataSink
         v-if="me"
         v-slot="{ submit }"
-        :src="`/me/mesh-list-view`"
+        :src="`/me/${props.name}`"
```
johncowen, line 24: This probably should have been done in https://github.com/kumahq/kuma-gui/pull/2625, but seeing as https://github.com/kumahq/kuma-gui/pull/2625 was just doing the `mesh-list` anyway it doesn't really matter.

This change means we take the `<RouteView name="this-bit">` and use that as a unique key to save the data against in localStorage. This means every route has its own namespace to save things in.

src/app/application/index.ts
```diff
@@ -56,6 +56,7 @@ declare module '@vue/runtime-core' {
 }
 
 const $ = {
+  ...ME,
```
johncowen, line 59: We use/expose `ME` dependency injection tokens so we can overwrite them elsewhere.

src/app/kuma/index.ts
```diff
@@ -13,6 +13,9 @@ export const TOKENS = {
 }
 export const services = (app: Record<string, Token>): ServiceDefinition[] => {
   return [
+    [app.storagePrefix, {
+      service: () => 'kumahq.kuma-gui',
+    }],
```
johncowen, line 18: Sets the localStorage key to use a `kuma.kuma-gui` prefix. So each route saves its data in `kuma.kuma-gui:name-of-route`. We'll need to be able to alter this prefix depending on the application, hence we've exposed it as a dependency injection token (also see above comment)

src/app/me/sources.ts
```diff
@@ -1,15 +1,8 @@
...
-import type { DataSourceResponse } from '@/app/application'
 
-export type MeSource = DataSourceResponse<{
-  pageSize: number
-  headers: Record<string, { width: number }>
-}>
-
-export const sources = (_api: unknown, storage: Storage = window.localStorage) => {
-  const prefix = 'kumahq.kuma-gui'
+export const sources = (prefix: string = 'me', storage: Storage = window.localStorage) => {
```
johncowen, line 5: Everything here and above is plumbing for the prefix thing.

johncowen (commented):
Inlines:

lukidzi (approved):

PR #2766, opened 2024-07-17
fix: Alters main menu selection to not cause full re-render

johncowen:
Back in https://github.com/kumahq/kuma-gui/pull/2662 we moved the main navigation in App.vue to use `route.child()` for selection.

I noticed that this is causing a full page re-render whenever the `route` is changed. You can notice this because our Github button in the header flickers on every single navigation (caused by the re-render). When navigating between tabbed sub-pages you shouldn't see a flicker (but you will do when navigation between Home, Zones, Meshes).

Strangely `route.child()` doesn't cause full page re-renders in other areas, in fact it was used/made specifically to avoid full page re-renders.

For the moment I've added a fix straight into `App.vue` that avoids the issue. But I would guess at some point I will come back to try and ensure that `route.child()` or some iteration of it works consistently across all usages.

### DIFF
--- a/src/app/App.vue
+++ b/src/app/App.vue
@@ -12,7 +12,7 @@
         class: 'kuma-ready',
       }"
       data-testid-root="mesh-app"
-      v-slot="{ t, can, route }"
+      v-slot="{ t, can }"
     >
       <ApplicationShell
         class="kuma-application"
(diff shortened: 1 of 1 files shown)


lukidzi (approved):

johncowen:
Wanted to note that this PR is good and its a good fix for the regression explained in the PR desc, but!

> You can notice this because our Github button in the header flickers on every single navigation (caused by the re-render).

I noticed that only the Github button has its DOM elements re-rendered. Whilst this was a good indication that we were re-calculating the rendering of things that we shouldn't. The DOM for the GH button shouldn't get re-rendered at all following this recalculation. I took a look at the `vue-github-buttons` package that we use for this and I think its the way that the vie implementation is handled. The vue implementation further uses another `github-buttons` package, so theoretically we could replace the Vue implementation with our own that doesn't have a needless re-render, but still depend on `github-buttons`, which would be a small but worthwhile technical improvement.

I'll probably PR something to cover ^ at some point soon.

PR #2794, opened 2024-07-26
fix: silences/fixes some dev-time console warnings

johncowen:
I noticed a bunch of dev-time `console` warnings. This PR cleans them up a little.

### DIFF
--- a/src/app/zones/data/index.ts
+++ b/src/app/zones/data/index.ts
@@ -87,7 +87,7 @@ export const ZoneOverview = {
         payload: {},
       })
     }
-    if (!get(insight, 'version.kumaCp.kumaCpGlobalCompatible', 'true')) {
+    if (!get(insight, 'version.kumaCp.kumaCpGlobalCompatible', true)) {
       warnings.push({
         kind: 'INCOMPATIBLE_ZONE_AND_GLOBAL_CPS_VERSIONS',
         payload: {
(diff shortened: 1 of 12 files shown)


src/app/zones/data/index.ts
```diff
@@ -87,7 +87,7 @@ export const ZoneOverview = {
         payload: {},
       })
     }
-    if (!get(insight, 'version.kumaCp.kumaCpGlobalCompatible', 'true')) {
+    if (!get(insight, 'version.kumaCp.kumaCpGlobalCompatible', true)) {
```
johncowen, line 90: This one wasn't necessarily a console warning error, and doesn't fix anything, but its an issue in the same vein in that its incorrect but doesn't necessarily have any negative effect in our current app.

---

This should default to a boolean not a string if `kumaCpGlobalCompatible` isn't set. Our `get` utilities isn't very type safe as it always deals with `any` instead of using generics.

I plan to remove our usage of `get` if possible (we only use it in a few places), but if I can't completely get rid of it, then I'll try and convert it to use generics instead of `any` which would probably flag any similar errors like this.

vite.config.production.ts
```diff
@@ -26,6 +26,9 @@ export const config: UserConfigFn = () => {
         template: {
           compilerOptions: {
             whitespace: 'preserve',
+            isCustomElement: (item) => [
+              'search',
+            ].includes(item),
```
johncowen, line 31: `<search />` elements are a thing now, but Vue doesn't know about them

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/search

johncowen (commented):
Inlines:

lahabana (approved):

PR #2811, opened 2024-07-31
chore: prevent a re-render if a change of headers has same value

johncowen:
Prevents a table re-render when the `headers` reference changes but the value doesn't.

This PR also fixes the issue described in https://github.com/kumahq/kuma-gui/pull/2810 but in a different way.

Whilst both separately fix the issue and only one is theoretically needed, both a valuable improvements.

### DIFF
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -215,8 +215,10 @@ watch(() => props.pageNumber, function () {
     kTableMountKey.value++
   }
 })
-watch(() => props.headers, function () {
-  kTableMountKey.value++
+watch(() => props.headers, function (val, prev) {
+  if (JSON.stringify(val) !== JSON.stringify(prev)) {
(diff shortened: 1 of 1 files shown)


src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -215,8 +215,10 @@ watch(() => props.pageNumber, function () {
     kTableMountKey.value++
   }
 })
-watch(() => props.headers, function () {
-  kTableMountKey.value++
+watch(() => props.headers, function (val, prev) {
+  if (JSON.stringify(val) !== JSON.stringify(prev)) {
```
lahabana, line 219: There's no better way to do object eval than stringify?

johncowen: This is probably the most dependency free/fastest/simplest way (as long as you don't care about key order, which we don't here)

(btw: assuming thats a eval > equal typo?)

PR #2874, opened 2024-09-05
chore: removes EmptyBlock in favour of XEmptyState

johncowen:
TBD


johncowen:
Just a quick note, there is other work that this will clash with if I merge this now. As its a refactor I'd rather wait until the other work has gone in, shouldn't be too long now.

PR #2909, opened 2024-09-10
chore: Alter interface of MSW passthroughs for easier addition

johncowen:
Reverses logic of `mswPassthroughs` (the files to ignore from being mocked by MSW) to make it easier to work with.

Also uses MSWs `print` to print the warning to console, instead of just `console.warn`

### DIFF
--- a/src/app/msw/index.ts
+++ b/src/app/msw/index.ts
@@ -5,7 +5,7 @@ import type { Token, ServiceDefinition } from '@/services/utils'
 import type { SetupWorker } from 'msw/browser'
 
 type Handler = Parameters<typeof setupWorker>[0]
-type Passthrough = (req: Request) => string | string[] | void
+type Passthrough = (req: Request) => boolean | void
 
 type Msw = ReturnType<SetupWorker['start']>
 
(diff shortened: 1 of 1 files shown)


src/app/msw/index.ts
```diff
@@ -22,26 +22,25 @@ export const services = (_: Record<string, Token>): ServiceDefinition[] => {
...
     [$.msw, {
       service: (handlers: Handler[] = [], passthroughs: Passthrough[]) => {
         return setupWorker(...handlers).start({
           quiet: true,
-          onUnhandledRequest(req: Request) {
-            passthroughs.reduce<string[]>((prev, item) => prev.concat(item(req) ?? []), []).forEach(item => {
-              console.warn(item)
-            })
+          onUnhandledRequest(req: Request, print) {
+            if (!passthroughs.some((passthrough) => passthrough(req))) {
```
slonka, line 41: In another PR there was a fix for `some` because it behaves different than expected in some cases. Is this the case here as well or are we safe here?

johncowen: Ah, I wasn't 100% sure so I just triple checked this. In this case it works as intended I think:


```
// when an MSW endpoint isn't found, some passthroughs defined
[shouldIPassThrough()].some((bool) => bool) // true, passthrough i.e. don't print a warning

// when an MSW endpoint isn't found, but there are no "passthroughs defined"
[].some((bool) => bool) // false don't passthrough i.e. make sure we print a warning
```

Issue #2957, opened 2024-09-17
Add a warning for MeshExternalServices when `mesh` requirements are not met


slonka:
### Description

MeshExternalService requires:
- egress deployed (which requires mTLS and MeshTrafficPermissions)
- `routing.zoneEgress: true` (see bottom of https://kuma.io/docs/2.8.x/production/cp-deployment/zoneegress/#zone-egress)

would be nice to display a warning somewhere in `MeshExternalServices` tab if these requirements are not met

![image](https://github.com/user-attachments/assets/6d26dd07-851e-4525-a424-94eec10129b8)

so something like this maybe could work:

![image](https://github.com/user-attachments/assets/320ef02a-ea45-4ddb-8da5-cc7c776d64fc)

On the user side when they are not met the MeshExternalService won't work and will error out with:

```
/app # wget -O- http://mes-https.extsvc.mesh.local/
Connecting to mes-https.extsvc.mesh.local ([::ffff:242.0.0.1]:80)
wget: error getting response: Connection reset by peer
```

Which is not very helpful, not sure we can do something there 🤔

jakubdyszkiewicz:
It does not required `routing.zoneEgress: true` now. However, it requires mTLS

PR #2964, opened 2024-09-18
feat(mes): check for mTLS and connected zoneEgress + warning

johncowen:
- Checks for an enabled mTLS backend and shows a warning if one isn't available.
- Checks for a 'connected' ZoneEgress and shows a warning if one isn't available.

---

- [Example](https://deploy-preview-2964--kuma-gui.netlify.app/gui/meshes/default/services/mesh-external-services?page=1&size=50#KUMA_MTLS_ENABLED=0;KUMA_ZONEEGRESS_COUNT=0)

Closes https://github.com/kumahq/kuma-gui/issues/2957

Screenshot 2024-09-18 at 12 36 50


src/app/services/views/MeshExternalServiceListView.vue
```diff
@@ -13,179 +13,210 @@
...
-          :src="uri(sources, '/meshes/:mesh/mesh-external-services', {
-            mesh: route.params.mesh,
-          },{
-            page: route.params.page,
-            size: route.params.size,
-          })"
+      <template
+        v-for="warnings in [[
+          [t('services.mesh-external-service.notifications.mtls-warning'), typeof props.mesh.mtlsBackend === 'undefined'],
+          [t('services.mesh-external-service.notifications.no-zone-egress'), egresses && !egresses.items.find((egress) => typeof egress.zoneEgressInsight.connectedSubscription !== 'undefined')],
```
johncowen, line 28: This is the biggest addition. Just uses `[i18nText, conditionalWhichWhenTrueShowsTheI18nText]`. The rest is just renesting.

src/app/meshes/data/index.spec.ts
```diff
@@ -99,6 +99,9 @@ describe('meshes data transformations', () => {
               },
             ],
           },
+          routing: {
+            zoneEgress: false,
+          },
```
johncowen, line 104: I added a check in the above `fromObject` code to set `routing.zoneEgress` to `false` if routing is completely `undefined`. In the end we didn't need this here, but we use this pattern a lot and we might need this at some point later so I left it in.

johncowen (commented):
Notes:

lukidzi (approved):

PR #2971, opened 2024-09-18
fix: amend cluster/rule lookups to use resourceMeta.labels

johncowen:
Changes how we lookup/match cluster to rule to use the resourceMeta labels instead of the resourceMeta.name.

### DIFF
--- a/src/app/connections/data/index.ts
+++ b/src/app/connections/data/index.ts
@@ -79,7 +79,7 @@ export const ConnectionCollection = {
                 switch (true) {
                   case (str.indexOf('_msvc_') !== -1):
                     return 'MeshService'
-                  case (str.indexOf('_mesvc_') !== -1):
+                  case (str.indexOf('_extsvc_') !== -1):
                     return 'MeshExternalService'
                   case (str.indexOf('_mzsvc_') !== -1):
                     return 'MeshMultiZoneService'
--- a/src/app/connections/data/index.ts
+++ b/src/app/connections/data/index.ts
@@ -56,7 +56,9 @@ export const ConnectionCollection = {
               type: '',
               mesh: '',
               name: '',
+              namespace: '',
               zone: '',
+              port: '',
             },
             tcp,
             ...(typeof http !== 'undefined' ? { http } : {}),
@@ -66,11 +68,13 @@ export const ConnectionCollection = {
           // sniff the name to see if we are a new type of service
           const found = cluster.match(meshServiceRe)
           if (found) {
-            const [mesh, name, namespace, zone, _abbr, _port] = cluster.split('_')
+            const [mesh, name, namespace, zone, _abbr, port] = cluster.split('_')
             stats.$resourceMeta = {
               mesh,
-              name: `${name}.${namespace}`,
+              name,
+              namespace,
               zone,
+              port,
               type: ((str: string) => {
                 switch (true) {
                   case (str.indexOf('_msvc_') !== -1):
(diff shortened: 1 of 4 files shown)


src/app/connections/views/ConnectionOutboundSummaryOverviewView.vue
```diff
@@ -321,12 +321,20 @@ import PolicyTypeTag from '@/app/common/PolicyTypeTag.vue'
...
   data: Record<string, any>
   dataplaneOverview: DataplaneOverview
 }>()
+
+const ruleForCluster = (cluster: any, rule: ResourceRule) => {
+  return cluster.$resourceMeta.name === rule.resourceMeta.labels['kuma.io/display-name'] &&
+  cluster.$resourceMeta.namespace === rule.resourceMeta.labels['k8s.kuma.io/namespace'] &&
+  cluster.$resourceMeta.zone === rule.resourceMeta.labels['kuma.io/zone'] &&
+  (rule.resourceSectionName === '' || cluster.$resourceMeta.port === rule.resourceSectionName)
+}
```
johncowen, line 337: The main change is here:

We use the labels in the resourceMeta, instead of the other things in the resourceMeta to lookup the connection between a clusterName and a rule.

Additionally, if resourceMeta.resourceSectionName is non-empty, we also check that it equals the port form the cluster name.

src/app/connections/views/ConnectionOutboundSummaryOverviewView.vue
```diff
@@ -321,12 +321,20 @@ import PolicyTypeTag from '@/app/common/PolicyTypeTag.vue'
...
 
 const props = defineProps<{
   data: Record<string, any>
   dataplaneOverview: DataplaneOverview
 }>()
+
+const ruleForCluster = (cluster: any, rule: ResourceRule) => {
+  return cluster.$resourceMeta.name === rule.resourceMeta.labels['kuma.io/display-name'] &&
+  cluster.$resourceMeta.namespace === rule.resourceMeta.labels['k8s.kuma.io/namespace'] &&
+  cluster.$resourceMeta.zone === rule.resourceMeta.labels['kuma.io/zone'] &&
```
lahabana, line 335: I think if namespace or zone is empty we shouldn't try to compare

lahabana: @lobkovilya can you confirm?

src/app/connections/data/index.ts
```diff
@@ -66,11 +68,13 @@ export const ConnectionCollection = {
           // sniff the name to see if we are a new type of service
           const found = cluster.match(meshServiceRe)
           if (found) {
-            const [mesh, name, namespace, zone, _abbr, _port] = cluster.split('_')
+            const [mesh, name, namespace, zone, _abbr, port] = cluster.split('_')
             stats.$resourceMeta = {
               mesh,
-              name: `${name}.${namespace}`,
+              name,
```
lahabana, line 74: Should you name this displayName to not confuse between the different types of names?

src/app/connections/data/index.ts
```diff
@@ -66,11 +68,13 @@ export const ConnectionCollection = {
...
+              namespace,
               zone,
+              port,
               type: ((str: string) => {
                 switch (true) {
```
lahabana, line 79: MeshExternalService is `mes` not `mesvc`.

lahabana (approved):

Issue #2969, opened 2024-09-18
Builtin gateway is not using display-name etc


lahabana:
### Description

https://cloud.konghq.com/us/mesh-manager/81450f88-36f5-4ff6-932f-9c249398f6f7/meshes/default/gateways/builtin?page=1&size=50

See:

![Screenshot 2024-09-18 at 16 13 48](https://github.com/user-attachments/assets/d7b31b1b-93e6-474e-aa49-c9ae94af260f)

PR #2972, opened 2024-09-18
fix(meshgateways): add display-name support to MeshGateways

johncowen:
See title

Closes https://github.com/kumahq/kuma-gui/issues/2969


features/mesh/builtin-gateways/Item.feature
```diff
@@ -18,7 +18,7 @@ Feature: mesh / builtin-gateways / item
       KUMA_RULE_MATCHER_COUNT: 1
       KUMA_RULE_MATCH_COUNT: 2
       """
-    Given the URL "/meshes/default/meshgateways/gateway-1" responds with
+    Given the URL "/meshes/default/meshgateways/gateway-1.namespace" responds with
```
jijiechen, line 21: Can we make `namespace` something else? I think it's a variable, right?
Here it looks like a constant in the template.

jijiechen:
Some nits, but the feature looks OK.

jijiechen (approved):

features/mesh/builtin-gateways/Item.feature
johncowen: oh yeah 2 tics, will switch up

johncowen: I changed it to `kuma-system`, thanks!

PR #2980, opened 2024-09-19
chore(meshservice): removes Services from DDP listing in exclusive

johncowen:
Removes the `Services` column from the Dataplane listing, and the DataPlane Inbound card links, only when the Mesh is in meshServices Exclusive mode.

### DIFF
--- a/src/app/services/features.ts
+++ b/src/app/services/features.ts
@@ -1,5 +1,6 @@
 import type { Features } from '@/app/application'
 import type Env from '@/app/application/services/env/Env'
+import { Mesh } from '@/app/meshes/data'
 export const features = (env: Env['var']): Features => {
   return {
     'use meshservice': () => {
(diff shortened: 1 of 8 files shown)


src/app/services/features.ts
```diff
@@ -9,5 +10,8 @@ export const features = (env: Env['var']): Features => {
         return true
       }
     },
+    'use service-insights': (_can, mesh: Mesh) => {
```
jijiechen, line 13: very interesting naming. Is this a good practice to use spaces in object members?

jijiechen (approved):

src/app/services/features.ts
johncowen: In javascript it doesn't matter really, especially not in this case.

The reason for this is so the usage of these things reads nicely, we have a function called `can` (which you can see in this PR) that reads like "if you can use service-insights":

```js
if(can('use service-insights')) {

}
```

There are others like `create zones` i.e. `can('create zones') ? showCreateButton : dontShowCreateButton`

PR #2997, opened 2024-09-24
chore(e2e): fill out `_rules` API with toResourceRules

johncowen:
Various changes to make it easier to work with the new `toResourceRules` in the `_rules` HTTP endpoint:

1. Use the generated types for the mocks
2. Uses `KUMA_CLUSTER_NAME` to allow you to set a clusterName that is always filled with `rules` so you can see these during development.
3. Adds MeshTLS to the mocks for `/policies`

More upstream changes are required to generate the resulting TS in this PR.

Note: The structure for toResourceRules is going to change soon, by using the auto generated types this should make the work easier moving forwards when such changes are made. ~There is more work required later to convert our data re-shaping to use the auto generated types rather than the handwritten ones.~ 👈 I added a further commit on the end here to do this here.


src/app/application/components/data-source/DataLoader.vue
```diff
@@ -91,7 +91,7 @@ import ErrorBlock from '@/app/common/ErrorBlock.vue'
 import LoadingBlock from '@/app/common/LoadingBlock.vue'
 
 const props = withDefaults(defineProps<{
-  data?: TypeOf<T>[]
+  data?: unknown[]
```
schogges, line 94: Why do we move away from generic to unknown?

schogges (approved):
👍, leaving a Q 🙂

src/app/application/components/data-source/DataLoader.vue
johncowen: umm, I can hardly remember 😆 

eeem, I think it was because `:data="[]"` can be a mix/an array of differing types. `:src=""` is the thing we need to infer the type from, whereas we never need to infer the types of `:data="[]"` because we'll already have sourced those from a different DataSource component.

lukidzi (approved):

src/app/application/components/data-source/DataLoader.vue
slonka: should this conversation be persisted as a code comment?

PR #3017, opened 2024-09-30
chore(mk): move accessible targets to the "make main file" i.e. Makefile

johncowen:
Changes our approach to Makefile targets to make them more customisable upstream

---

The biggest problem here is that Make make its troublesome to simply override make targets.

Previously we would 'override' targets by not including the file that had the target we had to override, and then including a different file with our customisations in it for the relevant target. The thing is you would also need to override (or just copy) all targets in a file, rather than just the one you wanted to override.

This approach names all consumable make targets prefixed with a `.` i.e. `.install` vs `install`. We than add references to all these in our base `Makefile` (essentially `make`s `main` file). The `Makefile` is specific to each project so we can add what we want in there and it doesn't affect anything downstream. This also includes adding different documentation if need be.

For example, say in one make file we have:

```make
install: .install ## Docs for this here
```

And we need to add some sort of auth to that we can use 
```make
.install/auth:
  $(AUTH) $(MAKE) .install
install: .install/auth ## Different docs for this here
```

Both of the above examples will run our installation scripts without them being duplicated, but one is decorated with some sort of authentication before doing so.

Moving forwards we should add all "common" user scripts in this file using the same approach. We should avoid mixing the documented aliases in with the rest of the `./mk` files, and avoid mixing long form scripts into the `Makefile` itself. I've added a small comment at the top of the `Makefile` to this effect. 

Lastly, I've only moved previously documented targets into this form as these are the commonly used ones. We can move others later on if necessary/desirable.

### DIFF
--- a/mk/test.mk
+++ b/mk/test.mk
@@ -1,24 +1,24 @@
 .PHONY: test
 
-.PHONY: test/unit
-test/unit: install ## Dev: run unit tests and exit
+.PHONY: .test/unit
+.test/unit: install
 	@TZ=UTC \
 		FORCE_COLOR=1 \
 		npx vitest \
 			-c vite.config.production.ts \
 			run
 
-.PHONY: test/unit/watch
-test/unit/watch: install ## Dev: run unit tests but watch for changes
+.PHONY: .test/unit/watch
+.test/unit/watch: install
 	@TZ=UTC \
 		FORCE_COLOR=1 \
 		npx vitest \
 			-c vite.config.production.ts \
 
 
-.PHONY: test/e2e
-test/e2e: CYPRESS_SPEC?=**/*.feature
-test/e2e: ## Run browser-based e2e tests against a running GUI, you may want to set KUMA_BASE_URL=http://localhost:8080/gui and KUMA_TEST_BROWSER=chrome
+.PHONY: .test/e2e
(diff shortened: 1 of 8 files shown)


mk/test.mk
```diff
@@ -1,24 +1,24 @@
...
+.test/unit/watch: install
 	@TZ=UTC \
 		FORCE_COLOR=1 \
 		npx vitest \
 			-c vite.config.production.ts \
 
 
-.PHONY: test/e2e
-test/e2e: CYPRESS_SPEC?=**/*.feature
-test/e2e: ## Run browser-based e2e tests against a running GUI, you may want to set KUMA_BASE_URL=http://localhost:8080/gui and KUMA_TEST_BROWSER=chrome
```
slonka, line 21: why are we removing this comment? seems still related

slonka (commented):
1 Q

mk/test.mk
johncowen: We moved it over here

https://github.com/kumahq/kuma-gui/pull/3017/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R56

schogges (approved):

slonka (approved):

PR #3030, opened 2024-10-01
chore(e2e): adds localStorage setup and teardown step

johncowen:
Adds a gherkin step to add values to localStorage:

```gherkin
Given the localStorage
  """
    key:
        this:
           will: end
           up: json
           encoded: true     
  """
```
The root key will be the localStorage key, and the rest will get JSON stringified and used for the value. So the above will be:

```
| Key | Value                                                   |
| key | {"this": { will: "end", "up": "json", "encoded": true}} | 
```

We never use anything but this form of JSON encoding, and we shouldn't ever use anything else either.

---

I also combined `visit.ts` here. We used to have this separate for reasons, but we no longer need it to be separate.  (ln:103 down is just the contents of the deleted file)


cypress/support/step_definitions/index.ts
```diff
@@ -24,13 +24,20 @@ const YAML = {
...
   client.reset()
   env = {}
   selectors = {}
+  localStorage = new Set()
   useServer()
 })
+After(() => {
+  Array.from(localStorage).forEach((key) => {
+    window.localStorage.removeItem(key)
+  })
```
slonka, line 39: can't we just do `clear()`?

```suggestion
  window.localStorage.clear()
```

cypress/support/step_definitions/index.ts
```diff
@@ -84,6 +100,46 @@ Given('the URL {string} responds with', (url: string, yaml: string) => {
 
 // act
 
+When('I visit the {string} URL', function (path: string) {
```
slonka, line 103: this seems a bit unrelated to the PR title (not related to localStorage)

slonka (commented):
2 Qs

cypress/support/step_definitions/index.ts
johncowen: We only want to remove the keys that are used in the tests, not the entire `localStorage`. I might have other values in localStorage that I want to keep.

Ideally we would save previous values that we might overwrite, and then put those values back in at the end of the tests, but not super important as yet

cypress/support/step_definitions/index.ts
johncowen: Oh yeah totally fair 👍 

I mentioned in the desc:

> I also combined visit.ts here. We used to have this separate for reasons, but we no longer need it to be separate. (ln:103 down is just the contents of the deleted file)

But happy to put this in another PR also, lmk.

slonka: no need

slonka (approved):

Issue #3031, opened 2024-10-02
add possibility to show endpoints in config_dump


lahabana:
### Description

Similar to this PR for kumactl: https://github.com/kumahq/kuma/pull/11583

It would be nice in the GUI to provide an option to also show the endpoint in the config dump.

Probably a checkbox next to this refresh button: 
<img width="1640" alt="image" src="https://github.com/user-attachments/assets/6f6ff1de-b9dd-4bce-8538-033a31984551"> ?

You then just need to do the request with `include_eds=true`.

PR #3032, opened 2024-10-02
feat: add include_eds checkbox for XDS config views

johncowen:
Adds a checkbox to all XDS configuraiton pages that controls whether to show eds/endpoints in the XDS config or not.

**I wanted to double check the API URLs for this. I'm pretty sure the API endpoints here are correct, inside `/zoneegresses` and `/zoneingresses`. Just we have API endpoints for Zone Ingresses under `/zone-ingresses` also (note the hypen)**

You can also check this with a local cluster by using `KUMA_MOCK_API_ENABLED=false` so that the we use a local running instance of Kuma.

Closes https://github.com/kumahq/kuma-gui/issues/3031

### DIFF
--- a/src/app/data-planes/sources.ts
+++ b/src/app/data-planes/sources.ts
@@ -87,6 +87,17 @@ export const sources = (source: Source, api: KumaApi, can: Can) => {
         dataPath,
       })
     },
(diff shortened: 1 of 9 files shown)


Automaat (approved):

lahabana:
Tested locally LGTM!

PR #3047, opened 2024-10-09
chore(ci): rename GUI PR GHA to "release"

johncowen:
This PR changed our "Create GUI PR" workflow:

1. Naming. I changed the name to `release.yaml` / `Build and release to kumahq/kuma` because we are releasing the GUI to the binary/kuma. "Create GUI PR" sounds like I am creating a PR on `kumahq/kuma-gui`
~2. Added a `repository-data` job that fetches all "topics" from the repository~
~3. Uses the `repository-data` output to look for a `auto-release` "topic" which when present (as it is currently) performs a "release" as before.~

~**I plan on merging this with the `auto-release` "topic" applied to prove it works (validate the PR is made). I will then remove the `auto-release` topic temporarily.**~

~Notes:~

- ~I "think" I can't access the secrets in a run because I'm PRing from a forked PR, hence will currently fail until this is on `master`. I would level a better way to test this out (and GH Actions as a whole really) edit: I confirmed offline that my guess here is right.~
- ~In this PR we have two calls to get the GH token. One during the actual "release" and one to have permissions to access the repo topics. I plan on reusing the one in the `repository-data` job across the entire workflow once this PR is merged and we know it works as expected 🤞~

---

edit: [See comment below](https://github.com/kumahq/kuma-gui/pull/3047#issuecomment-2404660523), but TLDR; this PR now pretty much just renames the workflow action:

1. Naming. I changed the name to `release.yaml` / `Build and release to kumahq/kuma` because we are releasing the GUI to the binary/kuma. "Create GUI PR" sounds like I am creating a PR on `kumahq/kuma-gui`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,22 @@
-name: 'Create GUI update PR'
-run-name: 'Create GUI update PR (branch: ${{ github.event.workflow_run.head_branch || inputs.branch }})'
+name: 'Build and release to kumahq/kuma'
+run-name: |
+  Build and release to kumahq/kuma (branch: ${{ github.event.workflow_run.head_branch || inputs.branch }})
 
-# **Note 1**: You can merge a pull request into a release branch of the form “release-$MAJOR.$MINOR” (e.g. “release-2.1”) in order to create the GUI update PR in the matching release branch of the host repository. For this to work, the name of this repository’s release branch must match that of the host repository exactly.
-# **Note 2**: Since this workflow can be triggered using the `workflow_run` event type, one needs to pay special attention to the context in which runs will be executed based on this workflow. Most importantly, runs will be using the workflow file found **in the project’s default branch**. This means that context variables like `github.ref` and `github.ref_name` will refer to the default branch **and not the branch that caused the workflow_run event to trigger**. Also, it likely means that to change the behavior of the workflow in a release branch, one will actually have to update the workflow file in the default branch, too.
+# **Note 1**: You can merge a pull request into a release branch of the form
+# “release-$MAJOR.$MINOR” (e.g. “release-2.1”) in order to create the GUI
+# update PR in the matching release branch of the host repository. For this to
+# work, the name of this repository’s release branch must match that of the
+# host repository exactly.
+
+# **Note 2**: Since this workflow can be triggered using the `workflow_run`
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -19,8 +33,17 @@ on:
...
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
+    branches:
+      - master
+      - release-[0-9]+.[0-9]+
+  # Can be used for debugging workflow whilst in a PR
+  # pull_request:
+  #   types:
+  #     - opened
+  #     - reopened
+  #     - synchronize
```
johncowen, line 46: I left this in but commented for future us if we ever come to debug this again.

.github/workflows/release.yml
```diff
@@ -32,12 +55,39 @@ concurrency:
...
+  repository-data:
+    runs-on: ubuntu-latest
+    outputs:
+      token: ${{ steps.topics.outputs.token }}
+      topics: ${{ steps.topics.outputs.topics }}
+    steps:
+      - id: token
+        uses: actions/create-github-app-token@5d869da34e18e7287c1daad50e0b8ea0f506ce69 # v1.11.0
+        with:
+          app-id: ${{ secrets.APP_ID }}
```
johncowen, line 67: If I run this from the PR it tells me `app-id` is unset, so I'm guessing `secrets.APP_ID` is unset as I'm PRing from a forked PR and that when its merged all will work - I maybe be completely wrong on this tho!

johncowen (commented):
Notes:

.github/workflows/release.yml
johncowen: Confirmed offline that I can't access secrets here as I'm PR from a fork. Not sure how we can 100% test this end to end without merging. But in the scheme of things, if it breaks once its on master we can just fix it. I tested it quite a bit using `github.token` instead.

johncowen:
@lahabana showed me a button in the GH UI that does exactly what this PR does 😆 , so a lot of this PR is no longer needed.

---

There are some bits here that I'd like to keep, mostly the renaming of "Create GUI PR" to "Release". We'd also like to improve the "release" PRs to include more changelog information.

For the moment I'm going to put this back into draft. While I re-jiggle this about a bit.

slonka (approved):

PR #3049, opened 2024-10-09
chore: remove unnecessary env vars

johncowen:
Back in the mists of time we used a plugin in order to interpolate variables into our `index.html` file. We need this as in 'production' kuma itself uses our `index.html` as a template and interpolates variables into it.

Since then vite was upgraded to v5 which included functionality to interpolate variables, but _only_ shell environment variables. This led us to drop the plugin for variable interpolation and use environment variables instead. The only thing is this meant we also started adding single line stringified JSON into environment variables.

This has always sat a little strange to me, these variables don't need to be shell environment variables and the stringified JSON is awkward.

Additionally lately we've realised we want to use a variable from the `package.json` file (`"version": ...`) to prevent duplication of this value in several places. (see https://github.com/kumahq/kuma-gui/pull/2967)

Because of this last point, I decided it was a good time to revisit this.

---

This PR aims to replicate exactly what the kuma binary does. We use JS to search and replace the `golang` style `{{.BaseGuiPath}}` variables with the actual variables. ~This is achieved using a relatively simple almost single line "inline vite plugin".~ As we replicate what kuma does, it means we don't have to do anything special anymore for 'production', we just ship the template as is, then only in 'development' we interpolate variables.

This also removes the need for the shell environment variables and the awkward stringified single-line JSON, and means we can reference the `package.json#version` via JS where we need to.

Lastly, I removed some fallback JSON that we had for a case that I don't think I've ever seen happen (an un-parseable HTML file), and really, if the HTML is un-parsable we probably don't want anyone using the resulting GUI as its likely to be very broken and/or unadvisable to use.

Overall this PR simplifies things quite a bit.

---

## Post initial approval additions (https://github.com/kumahq/kuma-gui/pull/3049/commits/873ab088cac57795241586590443dc6d0c3a0806)

There has unfortunately been quite a bit of back and forth here, partly because I needed to check that this was all going to work elsewhere. I also noticed another area where we do similar template interpolating (our preview server that runs a production build during e2e tests).

Due to those realisations I:

- Created a `vite.plugins.ts` file to house the things I wanted to reuse across vite configs (this new file will soon be moved to a centralized config area)
- Added 2 vite plugins here.
  - A plugin that works with `configureServer` _and_ `configurePreviewServer` for local dev and for running a production build for e2e testing
  - A plugin that works for our docs site, that unfortunately needs to use a slightly different approach as we can't provide an index.html to vitepress to use, therefore we sneakily inject HTML into the index.html that vitepress uses.

All in all I'm happier that everything is now using the same approach, _and_ we do _nothing_ for production i.e. there is nothing to break.


vite.config.development.ts
```diff
@@ -1,11 +1,36 @@
+import { readFileSync as read } from 'node:fs'
 import { defineConfig, mergeConfig, type UserConfigFn, type UserConfig } from 'vite'
 
 import { config as prodConfig } from './vite.config.production'
+const version = JSON.parse(read('./package.json').toString()).version
 // https://vitejs.dev/config/
 export const config: UserConfigFn = (env) => {
   return mergeConfig(
     prodConfig(env),
```
slonka, line 9: unrelated, but why is there a `prodConfig` in a file called `development.ts`?

vite.config.development.ts
```diff
@@ -1,11 +1,36 @@
...
+              .replace('{{.BaseGuiPath}}', '/gui')
+              .replace('{{.}}', JSON.stringify({
+                baseGuiPath: '/gui',
+                apiUrl: 'http://localhost:5681',
+                version,
+                product: 'Kuma',
+                mode: 'global',
+                environment: 'universal',
+                storeType: 'postgres',
+                apiReadOnly: false,
```
slonka, line 27: This seems the same as https://github.com/kumahq/kuma-gui/pull/3049/files#diff-47b7ce02023c49f9552984f4026c55914f09f8d403fd8cc08f875f78cc0d1baeR8-R16 - should we apply DRY on it?

slonka (commented):
2 Qs

vite.config.development.ts
schogges: I think that `vite.config.production` is the base and in `vite.config.development` some pieces are overridden: [vite/mergeConfig](https://vite.dev/guide/api-javascript.html#mergeconfig) 🤔 not sure if it's worth to create another config called `base` 🙂

vite.config.development.ts
johncowen: 🤔 I think I'm gonna move this under and `kuma` property in our `package.json` and source it from there. We need to source package.json anyway for the version so we may as well put this JSON in there also if we want it centralized. Will sort today

vite.config.development.ts
slonka: I'd say if you put something prod specific in `production` and don't overwrite it in `dev` then we're in trouble because we might be sending some data cross env. I'd be heavily in favour of creating `base`.

johncowen: This is a bit of a weird one, and there are things to keep in mind mostly point 1 here:

1. We can't have a file called `vite.config.ts` in the root. If we do, vitepress will use it for its config automatically (ouch) and we don't want that happening. This is the biggest reason it got renamed to `.production`, at the time we just needed a different name, any different name, seeing as we already had a `.development` it got moved to `vite.config.production`. `.base` would also be fine for a name.
2. FYI `vite.development` was added incase we needed it (and I'm pretty sure we don't now) and `vite.config.preview` doesn't use the `production` file anyway from what I remember.
3. We have a longer running task to move all our config into a shareable JS package at `@kumahq/config`, and I was hoping to reorganize the vite config into that also, at which point all this would change. Next steps for this long running task is moving our repository to use npm workspaces though, so we are just waiting for a good time to do that before moving all these config files around.

I think the name will end up being `vite.config.app.ts`, which will source its config from `@kumahq/config`, then we'll have a `vite.config.e2e.ts` or something that will also source its config from `@kumahq/config` (and vitepress's config will also come from `@kumahq/config`.

In the meantime if we want to change it to something else I'm fine with that and would be happy to 👍 a PR, but take into account it will probably change again once we have `@kumahq/config` available everywhere.

johncowen:
Note to self: Remember to bump to 2.10 in here also now we've released 2.9

vite.config.development.ts
johncowen: I think we should leave this as is _for now_ i.e. in this PR.

I agree that the `.production` suffix isn't a great name and can be confusing, but seeing as all our config is going to be moved around pretty soon (hopefully) and the current state has been around for quite some time, I'd say lets leave as is in this PR and when we move the config files around lets make sure we don't keep this naming scheme.

I'm gonna resolve this convo, but let me know if you want to reopen

johncowen:
I've made a couple of additions here, one is straight-forwards one more complicated:

---

I bumped our package version. I'd planned to do this in a separate PR after this one had gone in, but seeing as we are already targeting 2.10 I bumped it here.

---

TDLR; All in all the JSON is now centralized in a vite plugin

In looking at DRYing out our JSON configuration that comes from kuma, I noticed that the path I'd specified in CliEnv was completely wrong. I'd specified `./package.json` from `./src/app/services`, and no such file exists. This made me realize that we don't even need CliEnv anymore. We had initially used this as a stop gap to be able to support our old CLI based e2e tests before we moved/migrated all of those over to use a browser runner (the CLI doesn't have access to index.html).

I figured I could just remove CliEnv, but then realized that we also used CliEnv for our engineering docs site. This originally will have just been for ease. So I re-worked things to add the index.html variables to the engineering docs index.html also. Meaning both our dev environment and the docs environment use _almost_ the same approach.

I say almost because I couldn't figure out how to provide a custom index.html file to vitepress (which is what our engineering docs uses) so I applied another little plugin there to add in the gotemplate `{{.}}` syntax, to then apply the same replacement plugin to it.

vite.config.development.ts
johncowen: See https://github.com/kumahq/kuma-gui/pull/3049#issuecomment-2441413261

slonka (approved):

johncowen:
Unfortunately I had to add more here post-review.

Please see the extra info in the PR description:

> Post initial approval additions (https://github.com/kumahq/kuma-gui/commit/873ab088cac57795241586590443dc6d0c3a0806)

schogges (approved):

PR #3079, opened 2024-10-21
fix(.vitepress): typescript support

schogges:
Adding dot-folders in tsconfig enables full ts support for files like `.vitepress/config.ts`


johncowen (approved):
LGTM!

PR #3088, opened 2024-10-22
chore: moves several components to our x namespace

johncowen:
We have various reasons to wrap, fix or otherwise rename some KComponents. This means that in order to know whether to use a `X` or `K` component, you need to know whether some sort of extra functionality/fixing/wrapping is required.

It would be much easier if all our generic GUI components were prefixed with `X` (we don't want to use `K` because we don't own that prefix)

This PR does several renames:

1. KBadge > XBadge (we have a spearate XBadge implem.)
2. KAlert > XAlert (just a rename for consistency)
3. KPop > XPop (just a rename for consistency)
4. KInputSwitch > XInputSwitch (just a rename for consistency)
5. CodeBlock > XCodeBlock (we have a separate implem. to KCodeBlock)

One of the consequences of moving CodeBlock to `x` is that the `code-block` folder also contained an additional component `ResourceCodeBlock`. I have a long running task of reducing the amount of CodeBlock components we have (I think we had 4 or 5 at one point), and at some point soon I plan on merging this `ResourceCodeBlock` into `XCodeBlock` meaning we will only have 1 CodeBlock component in our application.

Lastly, there are still some outstanding `K` components that need a rename of a wrap, but I didn't want to do too much in a single PR.


schogges (approved):
🚀

src/app/x/index.ts
```diff
@@ -61,16 +68,22 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
     [$.xVueComponents, {
       service: () => {
         return [
+          ['XCard', KCard],
+          ['XPop', KPop],
+          ['XInputSwitch', KInputSwitch],
+          ['XCheckbox', KCheckbox],
```
slonka, line 74: This is how we rename this right?

slonka (approved):
LGTM, only found places where I think we rename K* to X*

src/app/x/index.ts
johncowen: Yup, for a rename with no extra functionality this says "when I write <XCard>, use KCard".

Then if we ever need to wrap KCard with extra functionality we just make a XCard that wraps KCard and we change this to `['XCard', XCard]`.

You'll see we've already needed to do this wrapping with `['XBreadcrumbs', XBreadcrumbs]` for example (inside `XBreadcrumbs` we use `KBreadcrumbs` but add extra functionality).

Currently its confusing having two prefixes (K and X) depending on whether we need to add functionality or not, so we are in the process of just renaming them all to X even if they don't currently ned wrapping.

PR #3092, opened 2024-10-23
chore: adds XLayout component

johncowen:
Adds a `XLayout` component for laying out multiple components.

`XLayout` currently comes with two `types` of layout: `stack` and `separated`

- `stack` is the same as what `div class="stack"` does, renders the contents vertically with a gap in-between each item.
- `separated` is the same as the layout we use when rendering multiple badges and renders the contents inline/horizontally with a gap in-between each item.

---

We currently use `<div class="stack">` (and others div/class things) to lay out certain components. This uses `<XLayout type="stack">` to do exactly the same thing. But formalises it in the fact that is no longer just a div and a class, its a very simple component that you can use inline in your template with auto-complete-able layouts.

Another advantage is that we are making all our layouts consistent. For example, and specifically added in this PR, the margin between our tables and pagination is now shared with other "stacks". The margin is no longer a property (under the ownership) of a `KTable`, it's now owned by an XLayout. This way we can guarantee that the margin used for separating other components is exactly the same as the margin that separates a KTable/KPagination, previously we would have to guess/investigate what the margin was that was being used inside of KTable.

For `separated`, `TagList` currently uses this rule as CSS, but I'd like to be able to use it to render lists of `TargetRef` components, `Label` components, or maybe just Badge components. The idea is we can now re-use this simple layout via a composable layout component.

Lastly, I haven't refactored everything to use this as yet, I can do that in a follow up PR.

### DIFF
--- a/src/app/application/components/app-collection/AppCollection.vue
+++ b/src/app/application/components/app-collection/AppCollection.vue
@@ -154,7 +154,7 @@ const click = (e: MouseEvent) => {
   justify-content: flex-end;
   align-items: stretch;
   flex-wrap: wrap;
-  gap: $kui-space-60;
+  gap: $kui-space-70;
   font-size: $kui-font-size-40;
   width: 100%;
 }
--- a/src/app/application/components/data-collection/DataCollection.vue
+++ b/src/app/application/components/data-collection/DataCollection.vue
@@ -5,17 +5,35 @@
     <template
       v-if="slots.item"
     >
-      <template
-        v-for="item in [props.items.find(props.predicate)]"
-        :key="item"
-      >
+      <div>
+        <template
+          v-for="item in [props.items.find(props.predicate)]"
+          :key="item"
+        >
+          <slot
+            v-if="item"
+            name="item"
+            :item="item as T"
+          />
+          <slot
+            v-else
+            name="empty"
+            :items="items"
+          >
+            <XEmptyState
+              v-if="props.empty"
+              :type="props.type"
+            />
+          </slot>
+        </template>
+      </div>
+    </template>
+    <template
+      v-else
+    >
+      <div>
         <slot
-          v-if="item"
-          name="item"
-          :item="item as T"
-        />
-        <slot
-          v-else
+          v-if="items.length === 0"
           name="empty"
           :items="items"
         >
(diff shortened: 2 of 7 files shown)


src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -154,7 +154,7 @@ const click = (e: MouseEvent) => {
   justify-content: flex-end;
   align-items: stretch;
   flex-wrap: wrap;
-  gap: $kui-space-60;
```
johncowen, line 157: Looking elsewhere, this value was inconsistent, it should be `-70`

src/app/application/components/data-collection/DataCollection.vue
```diff
@@ -1,18 +1,36 @@
 <template>
-  <template
-    v-if="slots.item"
+  <XLayout
+    type="stack"
```
johncowen, line 3: All of this change is just wrapping in a new component

src/app/application/components/data-collection/DataCollection.vue
```diff
@@ -140,9 +144,6 @@ const paginated = computed(() => {
 
 </script>
 <style lang="scss" scoped>
-.pagination {
-  margin-top: $kui-space-70;
-}
```
johncowen, line 145: This is the value I was talking about in the PR description. We know let XLayout dictate the margin between the table and pagination (and it will be the same as all the other places we use XLayout)

src/assets/styles/_components.scss
```diff
@@ -7,7 +7,7 @@ Adapted from https://every-layout.dev/layouts/stack/.
 */
 
 .stack>*+* {
-  margin-block-start: $kui-space-80;
+  margin-block-start: $kui-space-70;
```
johncowen, line 10: I changed this because I knew it was slightly wrong, but eventually we will be able to delete this global style

johncowen (commented):
Notes:

schogges (approved):
LGTM, thanks for the explanations 🙂

johncowen:
Sorry! Just spotted one tiny extra thing here so popped another commit on the end.

I realized in adding XLayout to DataCollection, it would space the child content of the DataCollection (in our case the PolicyTypeListing).

I added a wrapping div around the contents of the default slot of the DataCollection to avoid this.

---

Separately I noticed our PolicyTypeList is a bunch of `div`s when it should actually be a `ul li` or similar.

lukidzi (approved):

PR #3095, opened 2024-10-25
fix(msw): inconsistent package version in generated file

schogges:
When running `make install` from outdated or cleaned `node_modules` the `PACKAGE_VERSION` in `public/mockedServiceWorker.js` gets aligned with the version specified in `package.json`, which causes a diff.

### DIFF
--- a/public/mockServiceWorker.js
+++ b/public/mockServiceWorker.js
@@ -8,7 +8,7 @@
  * - Please do NOT serve this file on production.
  */
 
-const PACKAGE_VERSION = '2.3.1'
+const PACKAGE_VERSION = '2.4.12'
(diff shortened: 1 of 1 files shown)


johncowen (approved):
👍 

Yeah this is super unfortunate and happens every single time dependabot does an upgrade unfortunately.

I have a vague plan to tell dependabot to also commit this change when it performs the upgrade, more details here:

https://github.com/kumahq/kuma-gui/pull/2918

I should make an issue for continuing that work ^ actually

johncowen:
Added an issue for continuing the dependabot work https://github.com/kumahq/kuma-gui/issues/3096

schogges:
> Added an issue for continuing the dependabot work #3096

Perfect 👍 🙂

PR #3119, opened 2024-10-29
feat(ServiceListTabsView): add explanatory sentences for service type views

### LINKED ISSUE #2928: Add separate explanatory sentences for Service types views
### Description

The sentences should go underneath the right hand side subnav, and we should have a separate sentence for each of Internal, External, MeshService, MeshMultiZoneService, MeshExternalService

schogges:
Moved and rephrased the title of the `ServiceListTabView` to separate explanatory sentences for each service type. For most of the services I included a link to the docs.

Screenshot 2024-10-29 at 11 01 47

Closes #2928

### DIFF
--- a/src/app/services/locales/en-us/index.yaml
+++ b/src/app/services/locales/en-us/index.yaml
@@ -27,15 +27,15 @@ services:
         mesh-service-list-view:
           label: MeshService
           description: !!text/markdown |
-            A [MeshService](https://kuma.io/docs/latest/networking/meshservice/#meshservice) represents a destination for traffic from elsewhere in the mesh and can define several networking details. The behaviour of this resource depends on the zone type. 
+            A <a href="https://kuma.io/docs/latest/networking/meshservice/#meshservice" target="_blank">MeshService</a> represents a destination for traffic from elsewhere in the mesh and can define several networking details. The behaviour of this resource depends on the zone type. 
         mesh-multi-zone-service-list-view:
           label: MeshMultiZoneService
           description: !!text/markdown |
-            A [MeshMultiZoneService](https://kuma.io/docs/latest/networking/meshmultizoneservice/#meshmultizoneservice) represents a group of `MeshService` resources in a loadbalanced multizone deployment. `MeshService` resources that are deployed across several clusters can be grouped by a `MeshMultiZoneService`.
+            A <a href="https://kuma.io/docs/latest/networking/meshmultizoneservice/#meshmultizoneservice" target="_blank">MeshMultiZoneService</a> represents a group of `MeshService` resources in a loadbalanced multizone deployment. `MeshService` resources that are deployed across several clusters can be grouped by a `MeshMultiZoneService`.
         mesh-external-service-list-view:
           label: MeshExternalService
           description: !!text/markdown |
-            **Experimental**: A [MeshExternalService](https://kuma.io/docs/latest/networking/meshexternalservice/#meshexternalservice) is a policy that allows an application or microservice to interact with explicit destinations that are not part of the mesh.
+            **Experimental**: A <a href="https://kuma.io/docs/latest/networking/meshexternalservice/#meshexternalservice" target="_blank">MeshExternalService</a> is a policy that allows an application or microservice to interact with explicit destinations that are not part of the mesh.
         service-list-view:
           label: Internal
           description: !!text/markdown |
@@ -43,7 +43,7 @@ services:
         external-service-list-view:
           label: External
           description: !!text/markdown |
-            An [ExternalService](https://kuma.io/docs/latest/policies/external-services/#external-service) is a policy that allows an application or microservice to interact with other services that are not part of the mesh.
+            An <a href="https://kuma.io/docs/latest/policies/external-services/#external-service" target="_blank">ExternalService</a> is a policy that allows an application or microservice to interact with other services that are not part of the mesh.
   detail:
     config: YAML
     data_plane_proxies: Data Plane Proxies
--- a/src/app/services/locales/en-us/index.yaml
+++ b/src/app/services/locales/en-us/index.yaml
@@ -23,14 +23,27 @@ services:
       config: 'YAML'
     items:
       title: Services
-      intro: !!text/markdown |
-        A Service represents an application or microservice that is registered with the mesh and can be managed, monitored and secured through the mesh's policies.
       navigation:
-        mesh-service-list-view: MeshService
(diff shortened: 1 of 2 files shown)


johncowen:
Looks good! I just had a few of q's:

1. I'd consider making the inline links open in a `target="_blank"` so we don't leave the app.
2. IMO they sentences should be full-width so they look the same as the rest of the app, thats my opinion tho, wdyt? Do you think its worth making this consistent with the rest or don't we mind here?
3. If we are going to link the name of the resource in the sentence we should probably do that in all the other sentences across the app? wdyt? (if you agree we can do this in a separate PR)
4. I just noticed we need to do the same as we do here for gateways ([builtin]/[delegated]). 🤔 probably good to do that in another follow up PR unless you were keen to do it here also, up to you.

Lmk, what you think then I reckon I can approve this, thanks!

schogges:
> Looks good! I just had a few of q's:
> 
> 1. I'd consider making the inline links open in a `target="_blank"` so we don't leave the app.
> 2. IMO they sentences should be full-width so they look the same as the rest of the app, thats my opinion tho, wdyt? Do you think its worth making this consistent with the rest or don't we mind here?
> 3. If we are going to link the name of the resource in the sentence we should probably do that in all the other sentences across the app? wdyt? (if you agree we can do this in a separate PR)
> 4. I just noticed we need to do the same as we do here for gateways ([builtin]/[delegated]). 🤔 probably good to do that in another follow up PR unless you were keen to do it here also, up to you.
> 
> Lmk, what you think then I reckon I can approve this, thanks!

1. Yes, that'd me much better 👍 
2. Ah yes, I also think we should aim for consistency 🙂  
3. Generally I think that linking to the docs improves the UX because it's more accessible. But probably also requires extra efforts here to keep it up to date and in sync with the docs 🤔 but I would agree to add links for the other resources to the docs (if available). Could also be multiple PRs then?
4. Makes sense 🙂  I'd prefer to keep PRs smaller and therefore go for a separate PR 🙂

johncowen:
Sounds good, lets do 1 and 2 here then and do 3 and 4 separately 👍 

I had 1 or 2 q's on the actual text/copy also, 1 tic.

src/app/services/locales/en-us/index.yaml
```diff
@@ -23,14 +23,27 @@ services:
...
-        service-list-view: Internal
-        external-service-list-view: External
+        mesh-service-list-view:
+          label: MeshService
+          description: !!text/markdown |
+            A <a href="https://kuma.io/docs/latest/networking/meshservice/#meshservice" target="_blank">MeshService</a> represents a destination for traffic from elsewhere in the mesh and can define several networking details. The behaviour of this resource depends on the zone type. 
+        mesh-multi-zone-service-list-view:
+          label: MeshMultiZoneService
+          description: !!text/markdown |
+            A <a href="https://kuma.io/docs/latest/networking/meshmultizoneservice/#meshmultizoneservice" target="_blank">MeshMultiZoneService</a> represents a group of `MeshService` resources in a loadbalanced multizone deployment. `MeshService` resources that are deployed across several clusters can be grouped by a `MeshMultiZoneService`.
```
johncowen, line 34: nit: I'm no subject matter expert here, but I wondered why we used the term `clusters` here instead of `zones`, should we use `zones` or maybe something like `zones/clusters`? Maybe the answer to this question, is just a "nope, clusters is exactly the term we want to use"

src/app/services/locales/en-us/index.yaml
```diff
@@ -23,14 +23,27 @@ services:
...
+        mesh-external-service-list-view:
+          label: MeshExternalService
+          description: !!text/markdown |
+            **Experimental**: A <a href="https://kuma.io/docs/latest/networking/meshexternalservice/#meshexternalservice" target="_blank">MeshExternalService</a> is a policy that allows an application or microservice to interact with explicit destinations that are not part of the mesh.
```
johncowen, line 38: How come this one is marked as experimental and not the others?

src/app/services/views/ServiceListTabsView.vue
```diff
@@ -9,33 +9,38 @@
...
+                    },
+                  }"
+                  :data-testid="`${name}-sub-tab`"
+                >
+                  {{ t(`services.routes.items.navigation.${name}.label`) }}
+                </XAction>
+              </template>
+            </XActionGroup>
+
+            <p
```
johncowen, line 40: Ah just noticed, this tag will have to be a `div v-html`.

markdown will always wrap a sentence with a `p` so you are gonna get double `p`s here if you use `p v-html`.

Good rule of thumb here is always use `div v-html` when dealing with i18n markdown.

johncowen (commented):
Few little comments on the text itself.

src/app/services/locales/en-us/index.yaml
schogges: Hmmm 🤔 I'm also not an expert, but I took this wording from the [docs](https://kuma.io/docs/latest/networking/meshmultizoneservice/#meshmultizoneservice)
> The most common use case for using MeshMultiZoneService is to group MeshServices that spawn across multiple clusters with failover to another cluster in case of unavailability in the local cluster.

src/app/services/locales/en-us/index.yaml
schogges: It's marked as experimental in the docs, but probably we should not do this here because we'd need to put an extra eye to keep it in sync? 🙂

src/app/services/views/ServiceListTabsView.vue
schogges: Ah, good point 👍 thank you 🙂

src/app/services/locales/en-us/index.yaml
johncowen: Cool, lets resolve this one then

src/app/services/locales/en-us/index.yaml
johncowen: Yeah I think you're right, lets take it out of here 👍

schogges: I'll remove experimental to reduce our efforts and keep it consistent. We can add it back if really necessary 🙂

schogges:
jfyi: after the changes I updated the screenshot in the description

johncowen (approved):
Looks good, thanks!

michaelbeaumont (commented):
:+1:

PR #3127, opened 2024-10-29
fix(services/locales): hardcoded docs base url in service type descriptions

schogges:
Replacing the hardcoded docs base URL with env `KUMA_DOCS_URL` and appending `KUMA_UTM_QUERY_PARAMS`.

### DIFF
--- a/src/app/services/locales/en-us/index.yaml
+++ b/src/app/services/locales/en-us/index.yaml
@@ -27,15 +27,15 @@ services:
         mesh-service-list-view:
           label: MeshService
           description: !!text/markdown |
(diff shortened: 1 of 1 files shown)


schogges:
> 👍
> 
> Nothing to do with your PR but, FYI I noticed that when we bump our package.json version to `2.10` all the docs URLs when testing locally or in a preview will read `/2.10.x/` and therefore 404. Not sure how much of a pain that might be 🤔
> 
> Not totally sure what to do there as yet, just talking out loud

We could probably add `/latest` after the base URL which then redirects to the latest stable version in the docs 🤔

Issue #2981, opened 2024-09-23
Gateway dataplane proxy view: filter input is missing after entering filter that results in 0 entries


Automaat:
### What happened?

When you try to filter dataplanes that match built in gateway and enter filter that returns 0 result, then filter input is missing and you cannot change your query without going back:

Normal view:
![image](https://github.com/user-attachments/assets/8ec1d918-a6a5-4175-8e68-e5ba29658d79)

With working filter:
![image](https://github.com/user-attachments/assets/34819e58-a458-4e3e-9863-128dca74d402)

With broken filter:
![image](https://github.com/user-attachments/assets/bf39f9c9-b20a-4adc-8191-4f80682830db)

PR #3089, opened 2024-10-23
fix(search): Always show search bars

johncowen:
This PR hoists all search/toolbars out of AppCollection and the DataLoader/Collections to its own "standalone component". Meaning it is always visible, throughout loading/erroneous and empty states.

---

There is a little CSS repetition here, but I hope to add a XLayout component pretty soon which I hope will cover this at some point.

Note I also took the `<search>` out of `FilterBar` and used it at the "call site" instead, in some places the `search` should contains the `select` also.

Closes https://github.com/kumahq/kuma-gui/issues/2981

Issue #2814, opened 2024-08-02
Reduce repetition and improve structure of i18n locales files


johncowen:
### Description

It should be really easy to figure out where to go to change the text via the i18n files. Being easy probably means being able to look at the URL of the page you are on and from that it being obvious which file and nested property corresponds to the text you are looking at.

As URLs/Routes are nested this possibly means a more nested structure that follows the URL.

But ideally we also want to reduce repetition, so just because we have a button that says `Create` in several nested URLs/Routes we should have several nested places in the locales files where we define the text `Create`.

There's probably more to say here but we should probably save those details/ideas until we come to do discuss this in more depth.

Vaguely related:

- https://github.com/kumahq/kuma-gui/issues/2430
- https://github.com/kumahq/kuma-gui/issues/2429

schogges:
We should consider that notification identifiers have to be globally unique.

lahabana:
First step will be to write a MADR to define the rules.

schogges:
In order to be more confident about a major normalization of i18n, we should make sure that our tests cover all parts of the UI, such that a test would fail in case we reference a missing key. The `Titles.feature` was supposed to catch these errors, so we should have a look and verify that it covers all parts of the UI. It just makes sure that all routes are mounted and rendered once.

Issue #2813, opened 2024-08-02
☔ normalise i18n


johncowen:
### Description

We should make more of an effort to get to a 100% i18n'ed application.

Things that might help:

- Introduce a lint rule to help us find these (and eventually prevent folks form adding strings to the application directly).
- We could do this module by module if it helps with definition of done/progress tracking. If that helps feel free to link to individual PRs for each of the below.

- [ ] `application`
- [ ] `common`
- [ ] `connections`
- [ ] `control-planes`
- [ ] `data-planes`
- [ ] `diagnostics`
- [ ] `external-services`
- [ ] `gateways`
- [ ] `kuma`
- [ ] `meshes`
- [ ] `onboarding`
- [ ] `poicies`
- [ ] `rules`
- [ ] `services`
- [ ] `subscriptions`
- [ ] `x`
- [ ] `zone-ingresses`
- [ ] `zone-egresses`
- [ ] `zones`
- [ ] `zones-crud`

lahabana:
Favouring #2814

johncowen:
> We should make more of an effort to get to a 100% i18n'ed application.

I wrote the above, but its never really been my opinion so:

Kind of gingerly adding/x-refing this conversation here, I think its the most appropriate place to add it, but seeing as we are favouring #2814 Ill add it there also.

https://github.com/kumahq/kuma-gui/pull/4299#issuecomment-3376335218

PR #3133, opened 2024-10-30
fix(search): Always show search bars (#3089)

johncowen:
Port of #3089 cherry-picked to `master`

---

This PR hoists all search/toolbars out of AppCollection and the DataLoader/Collections to its own "standalone component". Meaning it is always visible, throughout loading/erroneous and empty states.

---

There is a little CSS repetition here, but I hope to add a XLayout component pretty soon which I hope will cover this at some point.

Note I also took the `<search>` out of `FilterBar` and used it at the "call site" instead, in some places the `search` should contains the `select` also.

Closes https://github.com/kumahq/kuma-gui/issues/2981

### DIFF
--- a/src/app/common/filter-bar/FilterBar.vue
+++ b/src/app/common/filter-bar/FilterBar.vue
@@ -4,111 +4,109 @@
     class="filter-bar"
     data-testid="filter-bar"
   >
-    <search>
-      <form
-        ref="$form"
-        @submit.prevent="change"
+    <form
+      ref="$form"
+      @submit.prevent="change"
+    >
+      <button
+        class="focus-filter-input-button"
+        title="Focus filter"
+        type="button"
+        data-testid="filter-bar-focus-filter-input-button"
+        @click="focusFilterInput"
       >
-        <button
-          class="focus-filter-input-button"
-          title="Focus filter"
-          type="button"
-          data-testid="filter-bar-focus-filter-input-button"
-          @click="focusFilterInput"
-        >
-          <span class="visually-hidden">Focus filter</span>
+        <span class="visually-hidden">Focus filter</span>
 
-          <span class="filter-bar-icon">
-            <FilterIcon
-              decorative
-              data-testid="filter-bar-filter-icon"
-              hide-title
-              :size="KUI_ICON_SIZE_30"
-            />
-          </span>
-        </button>
+        <span class="filter-bar-icon">
+          <FilterIcon
+            decorative
+            data-testid="filter-bar-filter-icon"
+            hide-title
+            :size="KUI_ICON_SIZE_30"
+          />
+        </span>
+      </button>
 
-        <label
-          :for="`${props.id}-filter-bar-input`"
-          class="visually-hidden"
-        >
-          <slot>
-            {{ placeholderAndLabelFallback }}
-          </slot>
-        </label>
-        <input
-          :id="`${props.id}-filter-bar-input`"
-          ref="filterInput"
-          v-model="currentQuery"
-          class="filter-bar-input"
-          type="search"
-          :placeholder="currentPlaceholder"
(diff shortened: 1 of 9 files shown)


src/app/common/filter-bar/FilterBar.vue
```diff
@@ -4,111 +4,109 @@
...
-          data-testid="filter-bar-suggestion-box"
-        >
-          <div class="suggestion-list">
-            <p
-              v-if="tokenizerError !== null"
-              class="filter-bar-error"
-            >
-              {{ tokenizerError.message }}
-            </p>
+        <span class="visually-hidden">Focus filter</span>
```
schogges, line 18: Don't we use i18n?

src/app/common/filter-bar/FilterBar.vue
```diff
@@ -4,111 +4,109 @@
...
+          </p>
+
+          <button
+            v-else
+            type="submit"
+            class="submit-query-button"
+            :class="{ 'submit-query-button-is-selected': selectedSuggestionItemIndex === 0 }"
+            data-testid="filter-bar-submit-query-button"
+          >
+            Submit {{ currentQuery }}
```
schogges, line 79: s.a.

src/app/common/filter-bar/FilterBar.vue
```diff
@@ -4,111 +4,109 @@
...
+            <b>{{ fieldEntry.fieldName }}</b><span v-if="fieldEntry.description !== ''">: {{ fieldEntry.description }}</span>
 
             <button
-              v-else
-              type="submit"
-              class="submit-query-button"
-              :class="{ 'submit-query-button-is-selected': selectedSuggestionItemIndex === 0 }"
-              data-testid="filter-bar-submit-query-button"
+              class="apply-suggestion-button"
+              :title="`Add ${fieldEntry.fieldName}:`"
```
schogges, line 92: s.a.

src/app/common/filter-bar/FilterBar.vue
```diff
@@ -4,111 +4,109 @@
...
+              type="button"
+              :data-filter-field="fieldEntry.fieldName"
+              data-testid="filter-bar-apply-suggestion-button"
+              @click="applySuggestion"
             >
-              Submit {{ currentQuery }}
-            </button>
+              <span class="visually-hidden">Add {{ fieldEntry.fieldName }}:</span>
```
schogges, line 98: s.a.

src/app/policies/views/PolicyListView.vue
```diff
@@ -67,6 +67,22 @@
             </KCard>
 
             <KCard>
+              <search>
+                <form
+                  @submit.prevent
+                >
+                  <XInput
+                    placeholder="Filter by name..."
```
schogges, line 75: s.a.

src/app/common/filter-bar/FilterBar.vue
johncowen: Yeah your right. Right now, kinda mostly for "machine to human" purposes tho. There are a couple of issues relevant for an FYI:

- https://github.com/kumahq/kuma-gui/issues/2814

---

Separately, the code here comes from how it was previously, I just copy/pasted it elsewhere to move it for the fix, and additionally this PR was just a cherry-pick to port the fix we did on another branch so both commits match up in both `2.9.x` and `master`.

Would be fine moving these bits to our i18n strings here though if you think that would be better. lmk!

schogges: One the one hand I like to have one topic per PR, but sometimes including a little fix on the way is fine 😄 anyways, since there is an issue for that and you just moved some things around, I am fine tackling this with #2813 👍

schogges (approved):

lukidzi (approved):

PR #3135, opened 2024-10-31
chore: clean up some .env file vars

johncowen:
Since a long time ago we've had `.env` file support. The `.env` file used to be used for 2 things:

1. Changing several `env` vars at build-time (previous to having support for changing `env` vars at runtime via dev-time browser cookies)
2. Storing strings (sometimes test/copy, sometimes URLs) used throughout the application that we may need to change at build time (previous to having i18n support)

Since then, we added both support for setting env vars at runtime via cookies (removing the need to restart the dev server), and i18n support for storing maps of strings in a KV store mainly for localizing from "machine to human".

Both these additions completely remove the need for storing most of these values in an `.env` file so I've slowly been moving things to not use the `.env` file and then removing the env vars from `.env`.

This PR:

- Removes reference to `VITE_NAMESPACE`, we've used `t('common.product.name')` for quite a while now instead of this.
- Removes `KUMA_MESHSERVICE_ENABLED` we no longer use this.
- Removes `KUMA_ZONE_CREATION_FLOW` we no longer use this.

I also made some small amends to our `common.product` strings:

- I commented in `version` and `docs.base`. At some point I will uncomment and make the changes to also remove these from the env file.
- Moved the structure around slightly to put the `product` things at the very top of the file so they are visible at the top as soon as you open the file up.

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,3 @@
 VITE_VERSION_URL=https://kuma.io/latest_version
-VITE_NAMESPACE=Kuma
 VITE_KUMA_API_SERVER_URL=http://localhost:5681
 VITE_DOCS_BASE_URL=https://kuma.io/docs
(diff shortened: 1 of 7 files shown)


schogges (approved):
👍

bartsmykla (approved):

Issue #3123, opened 2024-10-29
remove sass usage


johncowen:
### Description

Whilst we need to use sass it for `@kong/design-tokens` support, we should remove as much sass usage as possible from our own code and just use native browser functionality, meaning less work to maintain (see https://github.com/kumahq/kuma-gui/pull/3122)

I think the only thing that uses sass is our `_mixins.scss` file.

Obvs we'll still need to use the sass variable supplied by `@kong/design-tokens`

lahabana:
@johncowen what's the scope and difficulty of this? Is it a lot of work?

PR #3136, opened 2024-10-31
chore: stop using sass to generate helper classes

johncowen:
Removes enough sass to prevent console errors due to a recent sass upgrade.

I also removed the deprecatedClasses file, which seems to only have been needed previous to kongponents v9. Now we are on v9 we no longer need this.

## Previously

Screenshot 2024-10-31 at 18 51 50

Related: https://github.com/kumahq/kuma-gui/issues/3123


schogges (approved):
👍

bartsmykla (approved):

PR #3137, opened 2024-11-01
fix: bump/override cookie dependency to ^0.7.1

johncowen:
`cookie` less than `0.7.0` has a low dev time security issue.

I've waited a few days in the hope that this would be resolved upstream, but its seems like the package that includes the dependency have bumped the version (3 weeks ago) but not actually made a release 😭 

Not sure when the release will happen and filter down to us, so I've used an override for now to mirror the version they have used (but not released) in the upstream repo.

Hopefully this should resolve the notification we have about this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1050,6 +1050,15 @@
         "cookie": "^0.5.0"
       }
     },
+    "node_modules/@bundled-es-modules/cookie/node_modules/cookie": {
+      "version": "0.7.2",
+      "resolved": "https://registry.npmjs.org/cookie/-/cookie-0.7.2.tgz",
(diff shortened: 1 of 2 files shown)


schogges:
> I've waited a few days in the hope that this would be resolved upstream, but its seems like the package that includes the dependency have bumped the version (3 weeks ago) but not actually made a release 😭

There is a PR open to automate releases, but this is not merged yet. Looking forward that `msw` has a direct dependency to `cookie`, but this is blocked atm. 🙁

johncowen:
Oh good spot! I didn't notice that, yeah would be good if that upstream PR goes in and they release, then we can remove this extra override

bartsmykla (approved):

PR #3150, opened 2024-11-05
fix: resolve sass problem in onboarding

johncowen:
Resolve another Sass issue in the onboarding area

## Previous

Screenshot 2024-11-05 at 12 28 25

I also added a commit here to remove a file we no longer need which unfortunately was missed in https://github.com/kumahq/kuma-gui/pull/3136. Its all "stop using sass related"


schogges (approved):
👍

michaelbeaumont (approved):

Issue #2921, opened 2024-09-12
Move zone proxy subscriptions to use table/drawer UX


johncowen:
### Description

See https://github.com/kumahq/kuma-gui/pull/2664 for more info, zone proxy (ingress/egress) subscriptions should also use this UX

#### Related PRs

- [x] https://github.com/kumahq/kuma-gui/pull/3151
- [x] https://github.com/kumahq/kuma-gui/pull/3153
- [x] https://github.com/kumahq/kuma-gui/pull/3159

johncowen:
Closed by PRs detailed in the above description

Issue #2430, opened 2024-04-12
Add auto i18n key prefixing to RouteView::t


johncowen:
### Description

This issue is split off from https://github.com/kumahq/kuma-gui/issues/1478

> Use auto-prefixes (so we don't need to pass long keys to i18n).

---

Our [`RouteView` component](https://github.com/kumahq/kuma-gui/tree/e865b09b5ff268836074028099223cc2325ea9f7/src/app/application/components/route-view) exports a `t` function for translating i18n keys to human text.

We generally uses a pattern of prefixing every key in a template with `module-name.route.route-descrirptor.key` for example `dataplanes.route.items.page-title`.

RouteView knows its own route name, and the route name includes the name of the module, so we could pre configure RouteViews exported `t` to automatically prefixing that information. 

This would mean: `t('page-title')` would access the `dataplanes.route.items.page-title` i18n string.

---

Things to consider before doing this.

1. Typescript. Our i18n string don't currently auto complete due to our label based DI container injection of 18n strings. This is solvable at some point. But being able to typecheck/autocomplete a call to `t` to automatically know that you have any keys of `dataplanes.route.items...` available adds a further complication. Considering we don't currently typecheck our i18n keys I'm not entirely sure given a choice which feature we'd like most - typechecked keys or not having to write the entire path all the time.

2. In order to add this feature, we would use the `RouteView::name` attribute. We currently purposefully hardcode in the name of the route taken from the routing config, both string should be exactly equal:

https://github.com/kumahq/kuma-gui/blob/e865b09b5ff268836074028099223cc2325ea9f7/src/app/data-planes/views/DataPlaneDetailView.vue#L9
https://github.com/kumahq/kuma-gui/blob/e865b09b5ff268836074028099223cc2325ea9f7/src/app/data-planes/routes.ts#L14

There are two things to consider on this point:
  - Ideally we wouldn't have to hardcode this string exactly the same in two places, and make sure we get it right. I think it would be pretty trivial to add a vite plugin to read the value from our routing config and compile the correct route name onto the `RouteView` at build time. I did see a fairly official docs link more or less explaining how to do this, but typically can't find it right now.
[...]

PR #3151, opened 2024-11-05
feature(routing): adds route config walking, plus several walkers

johncowen:
This PR introduces the concept of "route config walkers" to our application.

These walkers allow us to decorate individual route configs in a centralized place. They purposefully mutate the config rather than return the a copy of the config. I've kept the code for this in one place for now, but once I'm totally sure of the API I'll probably re-shuffle things slightly to use our service container labels. I've added code comments/TODOs to this effect in the PR.

---

The walkers/decorators I added here:

### `$routeName`

Currently we have/use `<RouteView name="the-name-of-the-route">` i.e. we hard code the name of the route associated with this file into the RouteView's name. This means we can't then reuse, or 'mount', the same route component file at another URL and at the very least its error/typo prone having to write the thing in two places. I actually wanted to do something like this when I added `RouteView` orignally but my Vue chops weren't good enough to think of a way to do it 😅  

For example, in relation to https://github.com/kumahq/kuma-gui/issues/2921, I'd like to 'mount' the `subscription/` module in various places: `zones/`, `zone-ingresses/`, `zone-egresses/` and `data-planes/`. I can't hardcode in `zone-cps-subscription-summary-view` as the name for the route when its under zones, because that will be incorrect when I re-use it for say `zone-ingress-subscription-summary-view`. We need to know the name of the route dynamically.

`addRouteName` adds a dynamically assigned `$routeName` global var to every single route component template rendering context:

- whilst following vue pseudo-standard `$` prefixing for global vars
- whilst avoiding the need for the template author to configure a prop all the time (by using the global)
- I don't think there is a way via typescript to say "this var exists in a route component, but not in any other types of components" so even though this is always set for route components, I set it to possibly undefined so that its correct if someone accidentally thinks its available in non-route components. This also explains why I'm ok using `$routeName!` (with `!`) here. I want to make sure `:name` is required, but I'm totally sure that in a route component `$routeName` is set.

### meta.module

For quite a while now our application is split into several microservices/micro-frontends/modules (or whatever we want to call them). It would be handy sometimes to know the module that a route belongs to. For example in relation to https://github.com/kumahq/kuma-gui/issues/2430. We would eventually like to avoid having to type `t('name-of-the-module.name-of-the-route.title')` and just write `t('title')` from within route components. If `RouteView::t` knows which module and route it is in, we can avoid having to repeat the prefix/namespace 10s of times throughout the route component, hopefully making it less of a chore to use i18n for anything not super essential.

I would guess there might be other places where we would use this.

`addMeta` lets you add a `meta.module` to the very top route in a modules route configuration. This then flows down to every single child route in that module so we don't have to hardcode that ourselves. We add it once, and it flows down. Note: I haven't added any top-level `meta.module` for this to use as yet.

### meta.path

I also think it will be useful at some point to have a full path for a routing configuration as well as just the name.

`addPath` automatically compiles a `meta.path` on every single route config containing the full path of a route. For example a route called `zone-ingress-subscription-summary-view` would have a `meta.path` similar to `zone-cp-detail-view.zone-ingress-detail-view.zone-ingress-subscription-summary-view`.

I hope at some point we'll be able to reduce some more redundant namespacing in the future and expand on the concept of 'mounting' modules at several different places.

For example, instead of a name of `zone-ingress-subscription-summary-view` I could have a route path of `zone-cp.zone-ingress.subscription.summary`. I might also want to link to simply `.summary.config` (the `config` of whereever I'm mounted) instead of `zone-cp.zone-ingress.subscription.summary.config`

---

Lastly I only need `$routeName` right now for https://github.com/kumahq/kuma-gui/issues/2921. Both `meta.module` and `meta.path` are features for other further work, but it was useful to add several functions to check the API (for example I didn't realise at first I might need access to `parent: RouteRecordRaw`)

### DIFF
--- a/src/app/vue/index.ts
+++ b/src/app/vue/index.ts
@@ -31,25 +31,26 @@ const $ = {
 // similar to navigationGuards. We'd then do this specific functionality in
 // `@app/application` as it relates to RouteView
 const addRouteName = (item: RouteRecordRaw, _parent?: RouteRecordRaw) => {
-  if (typeof item.component !== 'undefined' && typeof item.component === 'function') {
-    // @ts-ignore at this point item.component has to be a function because we checked it above
-    const component = item.component.bind(item)
-    item.component = async () => {
-      // we need to create a new instance of the component as import is
-      // natively cached we could re-cache this in a WeakMap if it becomes
-      // necessary (I don't think it will)
-      const cached = (await component()).default
-      const route = {
-        default: {
-          ...cached,
-          render: (...args: Record<string, unknown>[]) => {
-            args[0].$routeName = item.name
-            return cached.render(...args)
-          },
+  if (typeof item.component !== 'function') {
+    return
+  }
+  // @ts-ignore at this point item.component has to be a function because we checked it above
+  const component = item.component.bind(item)
+  item.component = async () => {
+    // we need to create a new instance of the component as import is
+    // natively cached we could re-cache this in a WeakMap if it becomes
+    // necessary (I don't think it will)
+    const cached = (await component()).default
+    const route = {
+      default: {
+        ...cached,
+        render: (...args: Record<string, unknown>[]) => {
+          args[0].$routeName = item.name
+          return cached.render(...args)
         },
-      }
-      return route
+      },
     }
+    return route
   }
 }
 const addModule = (item: RouteRecordRaw, parent?: RouteRecordRaw) => {
--- a/src/app/vue/index.ts
+++ b/src/app/vue/index.ts
@@ -26,6 +26,62 @@ const $ = {
   routesLabel: token<RouteRecordRaw[]>('vue.routes.label'),
   navigationGuards: token<NavigationGuard[]>('vue.routes.navigation.guards'),
 }
+
+// @TODO at some point we should expose this as an extendable thing
+// similar to navigationGuards. We'd then do this specific functionality in
+// `@app/application` as it relates to RouteView
+const addRouteName = (item: RouteRecordRaw, _parent?: RouteRecordRaw) => {
+  if (typeof item.component !== 'undefined' && typeof item.component === 'function') {
+    // @ts-ignore at this point item.component has to be a function because we checked it above
+    const component = item.component.bind(item)
+    item.component = async () => {
+      // we need to create a new instance of the component as import is
+      // natively cached we could re-cache this in a WeakMap if it becomes
+      // necessary (I don't think it will)
+      const cached = (await component()).default
+      const route = {
+        default: {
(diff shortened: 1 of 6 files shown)


src/app/vue/index.ts
```diff
@@ -26,6 +26,62 @@ const $ = {
...
+          ...cached,
+          render: (...args: Record<string, unknown>[]) => {
+            args[0].$routeName = item.name
+            return cached.render(...args)
+          },
+        },
+      }
+      return route
+    }
+  }
```
schogges, line 53: Minor thing, but I'm fan of early exits to reduce nesting 🙂

schogges (approved):
👍

src/app/vue/index.ts
johncowen: At first I wasn't totally sure whether reversing the condtional in order to return early would make the conditional harder to follow at a glance. But I tried it out, and I simplified the conditional in doing this also 🎉 

Thanks for the suggestion! 👍 

I actually just thought I wonder if you can pass `component: import()` 🤔 👀 , gonna look it up even though we never do this.

johncowen:
Just to note here also, still looking at this PR because we saw that this wouldn't cope with something like `component: MyRouteComponent`, so just gonna add something to make sure it copes with that also

johncowen:
Add a bit more code here to cope with static and dynamic components, worth another look, lmk if you have any more comments. I'll leave this hanging for a bit, but if I hear nothing else I'll go ahead and merge.. We can always loop back on anything.

johncowen:
Oh! Looks like whatever I did blew everything up (all the browser tests are red) 😭 

Will take better look this morning

johncowen:
Oh wow, so glad again that we run a full production build for testing.

This works fine locally, and all the tests pass if you run it against a local dev instance, but if you build it for prod and run, it blows up.

I'm gonna pull this back to the version that was fine, it won't have support for static components (yet) but as we don't use those (yet) it doesn't matter, we can circle back at a later date and add support. For now I have a bunch of work that I'd like to PR that depends on the functionality here.

src/app/vue/index.ts
johncowen: OK I've kept the conditional simplification, but hit some sort of issue when I started messing with static vs dynamic components (something related to the difference between dev-time and built JS), so I left the static vs dynamic to one side for now, also see other PR comments below.

Gonna resolve this convo.

johncowen:
K I pulled this back to the state at which @schogges gave me a thumbs up, and everything is green again, so we are good to go here.

lukidzi (approved):

Issue #2431, opened 2024-04-12
Use global i18n strings instead of global env vars in our i18n strings


johncowen:
### Description

This issue is split off from https://github.com/kumahq/kuma-gui/issues/1478

>  Instead of using env for "i18n global variables", we should just use i18n for a group of select i18n keys that we define as being available "globally". We can collect these "globals" at i18n service start up (using @kong-ui-public/i18n#t and then always feed them into our own t). In doing this we remove the dependency on env and then we should be closer to only having a single place for defining application strings.

---

TLDR;

https://github.com/kumahq/kuma-gui/blob/e865b09b5ff268836074028099223cc2325ea9f7/src/app/application/services/i18n/I18n.ts#L32-L36

See the env usage above? We should try and reduce that and replace it with the `t` usage shown above.

This code makes these `GLOBAL_VARIABLES` available within i18n files.

---

Coming back to look at this after some time I think we've actually done most of this, some examples:

- `KUMA_VERSION` should be an `env` var as its not static
- `KUMA_DOCS_URL` is weirder, as it should use a static i18n derived URL, appended with `KUMA_VERSION`
- `KUMA_PRODUCT_NAME` can probably be removed from `env` entirely (but kept as being i18n derived)

Ideally all i18n 'globals' should live in `global.NAME_OF_VAR` or similar so its super obvious how to make a new i18n global var.

The final aim is to finally have `env` not deal with static variables (we only use `env` for historical reasons because we didn't have i18n when we originally needed injectable strings).

johncowen:
I just took a look at this when I was removing a stale label. It's mostly done, almost finished, I _think_ it just needs an extra little bit of effort to finish it up, which is mostly just cleanup. So we should probably get on and do this so we can close this up.

lahabana:
Let's finish this apparently it's mostly done

PR #3153, opened 2024-11-07
chore: uses new subscriptions table/drawer layout for zoneIngress

johncowen:
Adds new Subscription Table/Drawer UI to ZoneIngresses (whilst maintaining old-style for Egress and Dataplanes until we change those also)

Part of: https://github.com/kumahq/kuma-gui/issues/2921

---

~Lots of failing tests here, so keeping in draft until I've sorted that. I would guess it's probably chore-like updating of tests but we'll see!~

Turns out we can't use `addRouteName` from https://github.com/kumahq/kuma-gui/pull/3151 due to things being slightly different and weird between dev source and compiled/built source.

Instead I reverted to use route config `props` here, which was second favourite option. I plan to improve on this a little more, and if I ever figure out a more reliable way to add a dynamic global like `$routeName` I can add it at a later date. In the meantime I'd rather keep this table/drawers feature on track. Apologies for the noisy PR!

### DIFF
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -198,9 +198,9 @@ watch(() => {
 }, { immediate: true })
 
 watch(() => props.name, () => {
-  if (props.name.startsWith('$routeName')) {
-    throw new Error('`$routeName` has been used for a name, please use `:name` not `name`')
-  }
+  // if (props.name.startsWith('$routeName')) {
+  //   throw new Error('`$routeName` has been used for a name, please use `:name` not `name`')
+  // }
   // we only want query params here
   const params = Object.entries(routeParams || {}).reduce((prev, [key, value]) => {
     if (typeof route.params[key] === 'undefined') {
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -198,9 +198,6 @@ watch(() => {
 }, { immediate: true })
 
 watch(() => props.name, () => {
-  if (typeof props.name === 'undefined' || props.name.startsWith('$routeName')) {
-    throw new Error('`$routeName` has been used for a name, please use `:name` not `name`')
-  }
   // we only want query params here
   const params = Object.entries(routeParams || {}).reduce((prev, [key, value]) => {
     if (typeof route.params[key] === 'undefined') {
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -198,9 +198,9 @@ watch(() => {
 }, { immediate: true })
 
 watch(() => props.name, () => {
-  // if (props.name.startsWith('$routeName')) {
-  //   throw new Error('`$routeName` has been used for a name, please use `:name` not `name`')
-  // }
+  if (typeof props.name === 'undefined' || props.name.startsWith('$routeName')) {
+    throw new Error('`$routeName` has been used for a name, please use `:name` not `name`')
+  }
   // we only want query params here
   const params = Object.entries(routeParams || {}).reduce((prev, [key, value]) => {
     if (typeof route.params[key] === 'undefined') {
--- a/src/app/application/components/route-view/RouteView.vue
+++ b/src/app/application/components/route-view/RouteView.vue
@@ -198,6 +198,9 @@ watch(() => {
 }, { immediate: true })
 
 watch(() => props.name, () => {
+  if (props.name.startsWith('$routeName')) {
+    throw new Error('`$routeName` has been used for a name, please use `:name` not `name`')
+  }
   // we only want query params here
   const params = Object.entries(routeParams || {}).reduce((prev, [key, value]) => {
     if (typeof route.params[key] === 'undefined') {
--- a/src/app/kuma/locales/en-us/index.yaml
+++ b/src/app/kuma/locales/en-us/index.yaml
@@ -94,10 +94,13 @@ http:
       tls: TLS
       mtls: mTLS
       mTLS: mTLS
+      id: ID
       globalInstanceId: Global instance ID
       controlPlaneInstanceId: CP instance ID
+      zoneInstanceId: Zone Leader Instance ID
       connectTime: Connected
       disconnectTime: Disconnected
+      version: Version
       cds: CDS
       eds: EDS
       lds: LDS
--- a/src/app/subscriptions/components/SubscriptionList.vue
+++ b/src/app/subscriptions/components/SubscriptionList.vue
@@ -28,12 +28,12 @@ import { computed } from 'vue'
 
 import SubscriptionDetails from './SubscriptionDetails.vue'
 import SubscriptionHeader from './SubscriptionHeader.vue'
+import type { Subscription } from '../data'
 import AccordionItem from '@/app/common/AccordionItem.vue'
 import AccordionList from '@/app/common/AccordionList.vue'
-import type { DiscoverySubscription, KDSSubscription } from '@/types/index.d'
 
 const props = defineProps<{
-  subscriptions: Array<KDSSubscription | DiscoverySubscription>
+  subscriptions: Subscription[]
 }>()
 
 const reversedSubscriptions = computed(() => {
--- a/src/app/subscriptions/data/index.ts
+++ b/src/app/subscriptions/data/index.ts
@@ -1,32 +1,77 @@
 import type {
   Version as PartialVersion,
   DiscoverySubscription as PartialDiscoverySubscription,
+  SubscriptionStatus as PartialSubscriptionStatus,
+  Subscription as ProtoSubscription,
 } from '@/types/index.d'
 
-export type Version = PartialVersion
-export type Subscription = {
-  disconnectTime?: string
-  status: {
-    lastUpdateTime: string
-  }
-  version?: Version
-}
-export type DiscoverySubscription = PartialDiscoverySubscription
-export type SubscriptionCollection<T extends { version?: any }> = {
-  subscriptions: T[]
-  connectedSubscription?: T
-  version?: T['version']
-}
-export type DiscoverySubscriptionCollection = SubscriptionCollection<DiscoverySubscription>
+const acknowledgements = [
+  'responsesSent',
+  'responsesAcknowledged',
+  'responsesRejected',
+] as const
 
+export type Version = PartialVersion
+
+export type PartialSubscription = {
+  status: PartialSubscriptionStatus
+  version?: Version
+} & ProtoSubscription
+
+export type DiscoverySubscription = PartialDiscoverySubscription
+
+export const Subscription = {
+  fromObject<T extends PartialSubscription>(item: T) {
+    return {
+      $raw: item,
+      ...item,
+      status: ((item) => {
+        const { total, lastUpdateTime, stat, ...rest } = { stat: {}, ...item }
+        const stats = Object.keys(stat).length > 0 ? stat : rest
+        return {
+          ...item,
+          // make sure we default to zero for all acknowledgement properties
+          total: {
+            ...total,
+            ...acknowledgements.reduce((prev, prop) => {
+              prev[prop] = prev[prop] ?? 0
+              return prev
+            }, item.total as {
+              [key in (typeof acknowledgements)[number]]: number
+            }),
+          },
+          acknowledgements: {
+            ...Object.fromEntries(
+              Object.entries(stats).map(([key, value]) => {
+                return [key, acknowledgements.reduce((prev, prop) => {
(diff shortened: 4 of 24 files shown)


src/app/application/components/route-view/RouteView.vue
```diff
@@ -198,6 +198,9 @@ watch(() => {
 }, { immediate: true })
 
 watch(() => props.name, () => {
+  if (props.name.startsWith('$routeName')) {
+    throw new Error('`$routeName` has been used for a name, please use `:name` not `name`')
+  }
```
johncowen, line 203: I bumped into a thing were I used `name="$routeName!"` instead of `:name="$routeName!"`, and it took me a good few seconds/minutes to realize.

This should prevent that from happening again, and seeing as we are already watching `props.name` it was no hassle to add.

src/app/kuma/locales/en-us/index.yaml
```diff
@@ -94,10 +94,13 @@ http:
       tls: TLS
       mtls: mTLS
       mTLS: mTLS
+      id: ID
       globalInstanceId: Global instance ID
       controlPlaneInstanceId: CP instance ID
+      zoneInstanceId: Zone Leader Instance ID
       connectTime: Connected
       disconnectTime: Disconnected
+      version: Version
```
johncowen, line 103: These are now part of the `http.api.property` i18n string. Means 2 things:

1. We don't have to add them in other modules as these are commonly used across the entire app.
2. The terms are consistent through out the entire app (see point 1)

This is also related to https://github.com/kumahq/kuma-gui/issues/2814 and https://github.com/kumahq/kuma-gui/issues/2431.

You'll notice I possibly seem overly cautious about adding i18n at the moment even though we really want our entire app to be 100% i18n'ed. I'd just really like to find/figure a good pattern/tooling so we don't just keep digging the current hole deeper. The hole is currently on the wrong place, and once we find the right place, we are gonna have to fill this one back in again. Possibly bad analogy! 😆

src/app/subscriptions/components/SubscriptionList.vue
```diff
@@ -28,12 +28,12 @@ import { computed } from 'vue'
...
 import SubscriptionDetails from './SubscriptionDetails.vue'
 import SubscriptionHeader from './SubscriptionHeader.vue'
+import type { Subscription } from '../data'
 import AccordionItem from '@/app/common/AccordionItem.vue'
 import AccordionList from '@/app/common/AccordionList.vue'
-import type { DiscoverySubscription, KDSSubscription } from '@/types/index.d'
 
 const props = defineProps<{
-  subscriptions: Array<KDSSubscription | DiscoverySubscription>
+  subscriptions: Subscription[]
```
johncowen, line 36: I had to amend the current Subscription{Header, List, Details} components in order to keep supporting subscriptions for Dataplanes seeing as I haven't upgrade that yet (thats coming next)

src/app/subscriptions/data/index.ts
```diff
@@ -49,3 +94,4 @@ export const SubscriptionCollection = {
     }
   },
 }
+export type SubscriptionCollection = ReturnType<typeof SubscriptionCollection.fromArray>
```
johncowen, line 97: Changes here to our transformation layer, partly an upgrade to the style here, but also brings parts of KDS and DiscoverySubscriptions together so we can just refer to a `Subscription` instead of a union of the two.

src/app/subscriptions/routes.ts
```diff
@@ -1,20 +1,20 @@
...
         {
           path: 'overview',
-          name: 'subscription-summary-overview-view',
+          name: `${prefix}-subscription-summary-overview-view`,
           component: () => import('@/app/subscriptions/views/SubscriptionSummaryOverviewView.vue'),
         },
         {
           path: 'config',
-          name: 'subscription-summary-config-view',
+          name: `${prefix}-subscription-summary-config-view`,
```
johncowen, line 17: Changes relating to "mounting" a modules routes at different "mount points".

src/app/subscriptions/views/SubscriptionSummaryConfigView.vue
```diff
@@ -6,36 +6,28 @@
...
+        language="yaml"
+        :code="YAML.stringify(props.data.$raw)"
+        is-searchable
+        :query="route.params.codeSearch"
+        :is-filter-mode="route.params.codeFilter"
+        :is-reg-exp-mode="route.params.codeRegExp"
+        @query-change="route.update({ codeSearch: $event })"
+        @filter-mode-change="route.update({ codeFilter: $event })"
+        @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+      />
```
johncowen, line 22: I noticed the SubscriptionConfig tab content had a header and a KCard which is not how we render this in other areas of the GUI so I removed those things

src/app/zones/data/index.ts
```diff
@@ -114,3 +103,4 @@ export const ZoneOverview = {
     }
   },
 }
+export type ZoneOverview = ReturnType<typeof ZoneOverview.fromObject>
```
johncowen, line 106: Upgraded this to use the slightly different new inference based approach to our data transformers

src/test-support/mocks/src/zone-ingresses/_/_overview.ts
```diff
@@ -57,27 +57,41 @@ export default ({ fake, env }: EndpointDependencies): MockResponder => (req) =>
...
+                            responsesAcknowledged: Number,
+                            responsesRejected: Number,
+                          }, value,
+                        )
+                        return prev
+                      },
+                      {} as Record<string, {}>,
+                    ),
+                  }
+                })(),
```
johncowen, line 94: Much better realistic randomness for the ack/nacks mocks. This will eventually looks slightly better if we add deep functionality to `partitionInto` (see TODO comment I left a little further up)

src/types/index.d.ts
```diff
@@ -238,33 +215,59 @@ export interface EnvoyVersion {
...
-  id: string
   controlPlaneInstanceId: string
-  connectTime?: string
-  disconnectTime?: string
-  status: DiscoverySubscriptionStatus
   generation?: number
-  version?: Version
-}
+  status: DiscoverySubscriptionStatus
+} & Subscription
```
johncowen, line 270: These are the old handwritten types that we still need to support until we get these types OpenAPI'ed.

I found these to be slightly wrong, and I also moved them together in the file so I could look at the in the same place without a split. Whenever you are working on KDSSubscriptions you invariably need to work on DiscoverySubscription , much better if they are next to each other.

johncowen (commented):
This PR ended up needing to be a bit bigger than I'd wanted, so I left some inline notes also:

src/app/application/components/route-view/RouteView.vue
johncowen: Following my discovery that we can't use a global `$routeName` I remove this check

src/app/kuma/locales/en-us/index.yaml
schogges: I like the idea to have the entire app i18n'ed 👍 but you are right that we have to find a good pattern and structure 🙂 what do you mean with "the hole is currently on the wrong place"?

src/app/zone-ingresses/views/ZoneIngressDetailView.vue
```diff
@@ -62,7 +66,7 @@
                 <template
                   v-if="props.data.zoneIngress.advertisedSocketAddress.length > 0"
                 >
-                  <TextWithCopyButton
+                  <XCopyButton
```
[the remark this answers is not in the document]
schogges: Nice! 👍 😄

src/app/subscriptions/data/index.ts
```diff
@@ -1,32 +1,76 @@
...
-  subscriptions: T[]
-  connectedSubscription?: T
-  version?: T['version']
-}
-export type DiscoverySubscriptionCollection = SubscriptionCollection<DiscoverySubscription>
 
+export const Subscription = {
+  fromObject<T extends PartialSubscription>(item: T) {
+    return {
+      $raw: item,
```
schogges, line 29: We usually name this key `config`, correct?

src/test-support/mocks/src/zone-ingresses/_/_overview.ts
```diff
@@ -57,27 +62,41 @@ export default ({ fake, env }: EndpointDependencies): MockResponder => (req) =>
...
+                  const stats = Object.entries(fake.kuma.partitionInto(
+                    {
+                      cds: Number,
+                      eds: Number,
+                      lds: Number,
+                      rds: Number,
+                    }, xcks,
+                  ))
+                  return {
+                    lastUpdateTime: '2021-07-13T09:03:11.614941842Z',
```
schogges, line 77: Should we use a random fake date in the past? (same for the other cases)

schogges (approved):
👍 
left some Qs 🙂

src/app/kuma/locales/en-us/index.yaml
johncowen: A while back we put i18n in place (we bought shovels), we just went crazy digging everywhere instead of finding the right place to dig first. Once we've found the right place, we can dig there, but we'll have to also fill up all the places where we dug in the wrong place i.e. the refactor. Not sure if its a good idea to continue with the analogy! 😆 

---

To give some history: I was really keen to get i18n in place, but at first **purely so we could do "machine to human" translations**. So for example we use it a lot for our tabs where we use the `route.name` as the i18n key for the text in the tab, it translates the "machine" `route.name` to "human" text (not necessarily of any particular language). A lot of this "machine" to "human" was converting HTTP API properties/values to "human" (e.g. `mtls` to `mTLS`). I think we did this right by having `http.api.property`, `http.api.value` keys to have centrally reference-able human texts for these things.

Thing is, once we added i18n, we then went ahead and did lots of localisation on things that weren't to do with "machine" to "human" translation, and we did the i18n strings using completely different conventions, so there is no consistency anywhere. In some places we didn't think about using the `http.api.property` section of our i18n strings. i.e. we dug lots of different holes all over the place, with very few in the right place. i.e. the convention (or lack of convention) we have is all wrong. We are going to have to go back and change it all, the more there is the more we'll have to change.

So basically exactly what you said "we have to find a good pattern and structure", a lot of it doesn't have that as yet, so once we find that pattern (the sooner the better) we'll have to go through and apply that pattern to all of the places where we don't use it.

Hope that helps! Analogies are sometimes a bad idea 😬 😆

src/app/subscriptions/data/index.ts
johncowen: Yeah I bumped into this during the work. A `subscription` already has a `config` property, so we can't use that name 😬 .

We wrote about this eventuality somewhere and said we could use `$raw` if we ever bumped into this (and in other places we already use `$raw`). We could change the places where we use `config` for this so that they all use `$raw` really so its all consistent.

Lemme see if I can find where we wrote about this

src/app/zone-ingresses/views/ZoneIngressDetailView.vue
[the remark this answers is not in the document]
johncowen: 👍 I made an issue for this work https://github.com/kumahq/kuma-gui/issues/3157, gonna resolve this convo

src/test-support/mocks/src/zone-ingresses/_/_overview.ts
johncowen: Yeah defo, I think (from vague recollection) the reason dates have been an outlier in the mocks is that faker doesn't offer a helper that formats dates like these golang formatted dates. Lemme take a quick look and see if thats correct. We should add one of our own formatters if so and use it everywhere.

johncowen: Oh! I just looked so there's code for this in that `fake.kuma.connection` thing we were chatting over the other day. I can extract that so we can use it here also, will be much better. I'll PR that separately in a sec.

Thanks for the suggestion! 🎉

johncowen: I made this https://github.com/kumahq/kuma-gui/pull/3158, gonna resolve this convo

schogges: Perfect 👍 🙂

lahabana (approved):

src/app/subscriptions/data/index.ts
schogges: I see 🙂 I like consistency, so I'd vote for making use of `$raw` as this key is more likely not going to appear in the api payloads 🙂

src/app/kuma/locales/en-us/index.yaml
schogges: Alright, got it, thank you 🙂 👍

PR #3158, opened 2024-11-08
feature(mocks): add new nanodate and date formats to FakeKuma

johncowen:
I made `.kuma.nanodate` and `.kuma.date` for the slightly different date formats we have in the mocks.

I only used them in one mock to prove the outputs, and I also updated the same mock to use `semver` and `commitSha`, but we should try to gradually use these as we work on related features.

Note this mock along with changes in https://github.com/kumahq/kuma-gui/pull/3153 mean this mock is now entirely random 🎉


src/test-support/mocks/src/zone-ingresses/_/_overview.ts
```diff
@@ -80,15 +82,15 @@ export default ({ fake, env }: EndpointDependencies): MockResponder => (req) =>
...
-                    version: '1.18.3',
-                    build: '98c1c9e9a40804b93b074badad1cdf284b47d58b/1.18.3/Clean/RELEASE/BoringSSL',
+                    version: fake.system.semver(),
+                    gitTag: fake.system.semver(),
+                    gitCommit: fake.git.commitSha(),
+                    buildDate: fake.kuma.date(),
                   },
+                  envoy: ((sha, version) => ({
+                    version,
+                    build: `${sha}/${version}/Clean/RELEASE/BoringSSL`,
```
johncowen, line 92: I tried using `fake.system.directoryPath` here but it didn't work for some reason, don't think its important for this to be random though 🤷

schogges (approved):
👍

lukidzi (approved):

PR #3159, opened 2024-11-08
chore(subscriptions): implement new table/drawer subscriptions views in egress and dpps

johncowen:
Similar to https://github.com/kumahq/kuma-gui/pull/3153 implements new table/drawer layouts for subscriptions in both:

- ZoneEgress
- Dataplanes

... but seeing we now no longer use the old Subscription components, I deleted those here also.

---

I noticed we could use some new style kong icons for the rows here, so I'm going to look at that next.


features/mesh/dataplanes/DataplaneDetailsStandardProxy.feature
```diff
@@ -23,84 +18,27 @@ Feature: Dataplane details for standard Data Plane Proxy
...
-      |     3 |
-      |     4 |
-    And the "$status-lds" element contains
-      | Value |
-      | LDS   |
-      |     6 |
-    And the "$status-rds" element contains
-      | Value |
-      | RDS   |
-      |     0 |
```
johncowen, line 103: I deleted this test as it wasn't really adding much value (pretty much just checking Vue renders HTML), plus I'd like to move any subscription related tests into their own place (currently there are some dotted around various other modules such as zones/ingress/egress and here)

features/mesh/dataplanes/DataplaneDetailsStandardProxy.feature
```diff
@@ -23,84 +18,27 @@ Feature: Dataplane details for standard Data Plane Proxy
...
+            - connectTime: 2021-02-17T07:33:36.412683Z
               disconnectTime: 2021-02-17T07:33:36.412683Z
-              status:
-                lastUpdateTime: 2021-02-17T07:33:36.412683Z
-              version:
-                kumaDp:
-                  version: 1.0.7
-                  kumaCpCompatible: false
-                envoy:
-                  kumaDpCompatible: false
```
johncowen, line 55: These overwrites weren't used for assertions, so they are pointless

features/zones/zone-cps/egresses/Item.feature
```diff
@@ -31,8 +34,8 @@ Feature: zones / egresses / item
...
-    Then the "$egress-detail-tabs-view" element contains "item-1"
-    Then the "$egress-detail-view" element contains "166.197.238.26:20555"
-    Then the "$egress-detail-view" element contains "Connected: Jul 28, 2020, 4:18 PM"
-    When I click the "$egress-config-tab" element
-    Then the "$egress-config-view" element contains "type: ZoneEgress"
+    Then the "$header" element contains "item-1"
+    Then the "$overview-view" element contains "166.197.238.26:20555"
+    Then the "$subscription" element contains "Jul 28, 2020, 4:18 PM"
+    When I click the "$config-tab" element
+    Then the "$config-view" element contains "type: ZoneEgress"
```
johncowen, line 41: This is testing exactly the same as in ingresses, so I made the code the same also, eventually I'd like to move these tests in a separate subscriptions set of tests and see if we can run them in all 4 areas.

johncowen (commented):
Notes:

src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -504,10 +505,62 @@
...
+            <AppCollection
+              :headers="[
+                { ...me.get('headers.instanceId'), label: t('http.api.property.instanceId'), key: 'instanceId' },
+                { ...me.get('headers.version'), label: t('http.api.property.version'), key: 'version' },
+                { ...me.get('headers.connected'), label: t('http.api.property.connected'), key: 'connected' },
+                { ...me.get('headers.disconnected'), label: t('http.api.property.disconnected'), key: 'disconnected' },
+                { ...me.get('headers.responses'), label: t('http.api.property.responses'), key: 'responses' },
+              ]"
+              :is-selected-row="item => item.id === route.params.subscription"
+              :items="props.data.dataplaneInsight.subscriptions.map((_, i, arr) => arr[arr.length - (i + 1)])"
```
schogges, line 517: If I am not mistaken, you are reversing `subscriptions`. What about using `Array.reverse`? 🙂 
```suggestion
              :items="props.data.dataplaneInsight.subscriptions.reverse()"
```

src/app/zone-egresses/views/ZoneEgressDetailView.vue
```diff
@@ -51,15 +57,85 @@
...
+          <AppCollection
+            :headers="[
+              { ...me.get('headers.instanceId'), label: t('http.api.property.instanceId'), key: 'instanceId' },
+              { ...me.get('headers.version'), label: t('http.api.property.version'), key: 'version' },
+              { ...me.get('headers.connected'), label: t('http.api.property.connected'), key: 'connected' },
+              { ...me.get('headers.disconnected'), label: t('http.api.property.disconnected'), key: 'disconnected' },
+              { ...me.get('headers.responses'), label: t('http.api.property.responses'), key: 'responses' },
+            ]"
+            :is-selected-row="item => item.id === route.params.subscription"
+            :items="props.data.zoneEgressInsight.subscriptions.map((_, i, arr) => arr[arr.length - (i + 1)])"
```
schogges, line 71: s.a.
```suggestion
            :items="props.data.zoneEgressInsight.subscriptions.reverse()"
```

src/test-support/mocks/src/meshes/_/dataplanes/_/_overview.ts
```diff
@@ -74,29 +74,44 @@ export default ({ env, fake }: EndpointDependencies): MockResponder => (req) =>
...
+              const stats = Object.entries(fake.kuma.partitionInto(
+                {
+                  cds: Number,
+                  eds: Number,
+                  lds: Number,
+                  rds: Number,
+                }, xcks,
+              ))
+              return {
+                lastUpdateTime: '2021-07-13T09:03:11.614941842Z',
```
schogges, line 92: What about using a random date-time?
```suggestion
                lastUpdateTime: fake.date.past().toISOString(),
```

src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: `reverse` mutates the original array, which we don't want to do. We probably want `toReversed` but at the time I figured it was a bit early to be using that so went with this. I normally aim for a rough ~2 years of support for browser stuff.

---

FYI: We _might_ eventually order these ourselves (but in the data layer), but I usually keep data in the same order as what the backend gives us and then if we need to re-order for the specific view then do it in the specific view itself.

src/test-support/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: I don't think this gives us the exact format for the dates this seems to use. I think the backend gives us dates to nano second precision. Ideally we'd use https://github.com/kumahq/kuma-gui/pull/3158 here for that.

Lemme see if I can get that other PR merged in and we can circle back on these all in one go.

src/app/data-planes/views/DataPlaneDetailView.vue
schogges: Oh you are right, should be `toReversed` 👍 [caniuse](https://caniuse.com/?search=toReversed) reports >91% support (all users), which is good but probably not enough.

johncowen: Resolving this one

src/app/zone-egresses/views/ZoneEgressDetailView.vue
johncowen: see above resolving

johncowen:
@schogges I think we resolved all the inline convos here, not sure if you meant to approve this or whether you had anything else to comment here? lmk!

schogges (approved):
👍

michaelbeaumont (approved):

PR #3160, opened 2024-11-08
feat: introduce hostname-generators

### LINKED ISSUE #2956: Introduce HostnameGenerators
### Description

Build out initial views for HostnameGenerators. This should be built as a new 'module'. Installation of this new module should be done inside the existing `service-mesh` module, seeing as its a fundamental/essential part of our service-mesh software.

Views/frontend needed:

- [ ] A way to get there (possibly some sort of view or widget on the home page)
- [ ] ListView
- [ ] SummaryView
- [ ] DetailView

HTTP client types and data types already exist our OpenAPI generated typescript, so data interactions (`sources.ts` should be able to entirely use a OpenAPI generated HTTP client.

Patterns should probably follow one of the new service types i.e. MeshExternalService (a.k.a. if there's copy/pasta to be done MeshExternalService would be a good starting point)

Importantly, HostnameGenerators are a "top-level" construct similar to Meshes and Zones i.e. they don't have a parent. Despite this we don't want to put them in the main side menu just yet (or maybe never). Users will mostly just use the default HostnameGenerators and therefore barely need to know that they exist/are a thing. But for "power users" we still need a way for them to get to them and view them. Therefore we'd like to put a link and or view somewhere on the "home"/"control plane view" page so users can get to a HostnameGenerator listing page. This way they are as visible/accessible as the main side menu items, but are still visible/accessible when/if you need to get to them.

Contents of the detail and summary pages will be dictated by the contents of the HTTP API responses. The important thing will be to get the data on the page following existing patterns and we can then tweak and refine what's best to show and where to show it with the rest of the team.

schogges:
This change introduces the views for `HostnameGenerator`s, including `ListView`, `SummaryView` and `DetailView`. With the next iterations we will add more information to the views.
Currently there is no navigation element on the home page that brings the user to the views. For now the path `/hostname-generators` has to be entered manually.

**ListView**
Screenshot 2024-11-08 at 19 13 40

**SummaryView**
Screenshot 2024-11-08 at 19 13 56

**DetailView**
Screenshot 2024-11-12 at 10 10 33

Closes #2956

### DIFF
--- a/src/app/hostname-generators/data/HostnameGenerator.ts
+++ b/src/app/hostname-generators/data/HostnameGenerator.ts
@@ -1,0 +1,71 @@
+import type { components, paths } from '@/types/auto-generated'
+
+// TODO(schogges): Remove `creationTime` and `modificationTime` once the api specs are complete
+export type HostnameGeneratorList = Omit<components['responses']['HostnameGeneratorList']['content']['application/json'], 'items'> & {
+  items: NonNullable<components['responses']['HostnameGeneratorList']['content']['application/json']['items']>[number] & {
+    creationTime: string
+    modificationTime: string
+  } | undefined
+}
+export type HostnameGeneratorItem = components['responses']['HostnameGeneratorItem']['content']['application/json'] & {
+  creationTime: string
+  modificationTime: string
+}
+export type HostnameGeneratorGetParams = paths['/hostnamegenerators/{name}']['get']['parameters']
+
+export interface HostnameGenerator extends HostnameGeneratorItem {
+  id: string
+  namespace: string
+  zone: string
+  mesh: string
+  selector?: {
+    routeName: string
+    label: string
+  } | undefined
+  $raw: HostnameGeneratorItem
+}
+
+export const HostnameGenerator = {
+  fromObject(item: HostnameGeneratorItem) {
+    const labels = item.labels ?? {}
+    const name = labels['kuma.io/display-name'] ?? item.name
+    const namespace = labels['k8s.kuma.io/namespace'] ?? ''
+    return {
+      ...item,
+      id: item.name,
+      name,
+      namespace,
+      zone: labels['kuma.io/origin'] === 'zone' && labels['kuma.io/zone'] ? labels['kuma.io/zone'] : '',
+      mesh: labels['kuma.io/mesh'] || 'default',
+      $raw: item,
+      selector: ((selector: string | undefined) => {
+        if (!selector) return undefined
+
+        let routeName = 'mesh-service-list-view'
+        switch (selector) {
+          case 'meshExternalService': {
+            routeName = 'mesh-external-service-list-view'
+            break
+          }
+          case 'meshMultiZoneService': {
+            routeName = 'mesh-multi-zone-service-list-view'
+          }
+        }
+
+        return {
+          routeName,
+          label: `${selector.charAt(0).toUpperCase()}${selector.slice(1, selector.length)}`,
+        }
+      })(Object.keys(item.spec.selector ?? {})?.[0]),
+    } satisfies HostnameGenerator
+  },
+
+  fromCollection(collection: HostnameGeneratorList) {
+    const items = Array.isArray(collection.items) ? collection.items.map(HostnameGenerator.fromObject) : []
+    return {
+      ...collection,
+      items,
+      total: collection.total ?? items.length,
+    }
+  },
+}
--- a/src/app/hostname-generators/data/HostnameGenerator.ts
+++ b/src/app/hostname-generators/data/HostnameGenerator.ts
@@ -1,26 +1,8 @@
 import type { components, paths } from '@/types/auto-generated'
 
-// TODO(schogges): Remove `creationTime` and `modificationTime` once the api specs are complete
-export type HostnameGeneratorList = Omit<components['responses']['HostnameGeneratorList']['content']['application/json'], 'items'> & {
-  items: NonNullable<components['responses']['HostnameGeneratorList']['content']['application/json']['items']>[number] & {
-    creationTime: string
-    modificationTime: string
-  } | undefined
-}
-export type HostnameGeneratorItem = components['responses']['HostnameGeneratorItem']['content']['application/json'] & {
-  creationTime: string
-  modificationTime: string
-}
+export type HostnameGeneratorList = components['responses']['HostnameGeneratorList']['content']['application/json']
+export type HostnameGeneratorItem = components['responses']['HostnameGeneratorItem']['content']['application/json']
 export type HostnameGeneratorGetParams = paths['/hostnamegenerators/{name}']['get']['parameters']
-
-export interface HostnameGenerator extends HostnameGeneratorItem {
-  id: string
-  namespace: string
-  zone: string
-  mesh: string
-  selector?: keyof NonNullable<HostnameGeneratorItem['spec']['selector']> | undefined
-  $raw: HostnameGeneratorItem
-}
 
 export const HostnameGenerator = {
   fromObject(item: HostnameGeneratorItem) {
@@ -34,10 +16,28 @@ export const HostnameGenerator = {
       name,
       namespace,
       zone: labels['kuma.io/origin'] === 'zone' && labels['kuma.io/zone'] ? labels['kuma.io/zone'] : '',
-      mesh: labels['kuma.io/mesh'] || 'default',
+      creationTime: item.creationTime ?? '',
+      modificationTime: item.modificationTime ?? '',
+      spec: ((item = {}) => {
+        return {
+          ...item,
+          selector: ((item = {}) => {
+            return {
+              meshService: {
+                matchLabels: item.meshService?.matchLabels ?? {},
+              },
+              meshExternalService: {
+                matchLabels: item.meshExternalService?.matchLabels ?? {},
+              },
+              meshMultiZoneService: {
+                matchLabels: item.meshMultiZoneService?.matchLabels ?? {},
+              },
+            }
+          })(item.selector),
+        }
+      })(item.spec),
       $raw: item,
-      selector: item.spec.selector ? Object.keys(item.spec.selector)[0] as keyof typeof item.spec.selector : undefined,
-    } satisfies HostnameGenerator
+    }
   },
 
   fromCollection(collection: HostnameGeneratorList) {
@@ -49,3 +49,5 @@ export const HostnameGenerator = {
     }
   },
 }
+
+export type HostnameGenerator = ReturnType<typeof HostnameGenerator.fromObject>
--- a/src/app/hostname-generators/data/HostnameGenerator.ts
+++ b/src/app/hostname-generators/data/HostnameGenerator.ts
@@ -18,10 +18,7 @@ export interface HostnameGenerator extends HostnameGeneratorItem {
   namespace: string
   zone: string
   mesh: string
-  selector?: {
-    routeName: string
-    label: string
-  } | undefined
+  selector?: keyof NonNullable<HostnameGeneratorItem['spec']['selector']> | undefined
   $raw: HostnameGeneratorItem
 }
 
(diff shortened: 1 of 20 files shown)


src/app/hostname-generators/data/HostnameGenerator.ts
```diff
@@ -0,0 +1,71 @@
...
+            routeName = 'mesh-external-service-list-view'
+            break
+          }
+          case 'meshMultiZoneService': {
+            routeName = 'mesh-multi-zone-service-list-view'
+          }
+        }
+
+        return {
+          routeName,
```
johncowen, line 56: Ah glad I spotted this. We should avoid putting `routeNames` in the data layer if possible. This is one of those things of getting into the habit of where/what the boundaries are. I/we should be more explicit in the Eng Docs about this also

src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
```diff
@@ -0,0 +1,104 @@
...
+                  <XAction
+                    :to="{
+                      name: data.selector?.routeName,
+                      params: {
+                        mesh: data.mesh,
+                      },
+                    }"
+                  >
+                    {{ data.selector?.label }}
+                  </XAction>
```
johncowen, line 60: nit: If selector is definitely set from what I remember from your data transformer above both routeName and label are guaranteed to be set? Actually this made me realise something else (we should avoid having routeNames in the data layer), I made a comment above on that.

src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
```diff
@@ -0,0 +1,104 @@
...
+
+import { sources } from '../sources'
+import DefinitionCard from '@/app/common/DefinitionCard.vue'
+import ResourceCodeBlock from '@/app/x/components/x-code-block/ResourceCodeBlock.vue'
+</script>
+
+<style lang="scss" scoped>
+:deep(.kong-ui-app-about-section .about-section-content) {
+  display: block;
+}
```
johncowen, line 103: Do you know what this was for? (not sure if I did this somewhere else and its just copy/pasta'ed or whether you needed to add this for something else)

src/app/hostname-generators/views/HostnameGeneratorListView.vue
```diff
@@ -0,0 +1,126 @@
...
+                          },
+                          query: {
+                            page: route.params.page,
+                            size: route.params.size,
+                          },
+                        }"
+                      >
+                        {{ item.name }}
+                      </XAction>
+                    </TextWithCopyButton>
```
johncowen, line 72: Not necessary in the slightest, but if we need to push another commit or so onto this PR we could switch this to use `XCopyButton` while we are here.

src/test-support/mocks/src/hostname-generators/_/_overview.ts
```diff
@@ -0,0 +1,49 @@
+import type { HostnameGeneratorItem } from '@/app/hostname-generators/data'
+import type { EndpointDependencies, MockResponder } from '@/test-support'
+
+export default ({ fake, env }: EndpointDependencies): MockResponder => (req) => {
+  const { params } = req
+  const name = params.name as string
+  const [displayName = '', namespace = ''] = name.split('.')
+  const meshServiceTypeSelector = fake.kuma.meshServiceTypeSelector()
```
johncowen, line 8: Oh! I remember walking through this with you, I see that here in the end we didn't even need the CapitalCase form of the meshServiceType in the end 🎉 . Perfect!

src/test-support/mocks/src/hostname-generators/_/_overview.ts
```diff
@@ -0,0 +1,49 @@
+import type { HostnameGeneratorItem } from '@/app/hostname-generators/data'
+import type { EndpointDependencies, MockResponder } from '@/test-support'
+
+export default ({ fake, env }: EndpointDependencies): MockResponder => (req) => {
+  const { params } = req
+  const name = params.name as string
+  const [displayName = '', namespace = ''] = name.split('.')
```
johncowen, line 7: Ah! I really like this form. I think we could use this everywhere in the mocks to replace the horrible `.pop`ing thing I've been doing 😆  🎉  Nice!

src/app/hostname-generators/views/HostnameGeneratorListView.vue
```diff
@@ -0,0 +1,126 @@
...
+          <RouteTitle
+            :title="t('hostname-generators.routes.items.title')"
+          />
+        </h1>
+      </template>
+      <div class="stack">
+        <div v-html="t('hostname-generators.routes.items.intro', {}, { defaultMessage: '' })" />
+        <KCard>
+          <DataLoader
+            :src="uri(sources, '/hostnamegenerators', {}, {
```
johncowen, line 25: This is a super big nit, but! In our application we use the "machine-name" of `hostname-generator{s}` i.e. with a hyphen. We only use it without a hyphen here.

One of the purposes of having the `sources.ts` file with our own URIs is that we can totoally ignore whatever is made in the backend for whatever reason and define our own URIs so they work well for us. i.e. we have a thin proxy layer over the HTTP API.

Ideally we'd use `/hostname-generators` I would imagine? I've no idea why its `hostnamegenerators` in the backend but I would imagine the backend has a very good reason to make it like that (probably kube related). But I would also guess that reason doesn't apply to us, and seeing we use the hypenated version everywhere else in the frontend we might want to proxy `/hostname-generators` to `/hostnamegenerators` in the sources file. wdyt?

johncowen (commented):
This looks super good 👏 

I'd be more than fine to approve/merge as is. I left a couple of nit comments that aren't essential, mostly for an opportunity to talk through some things that aren't super obvious (I need to do a better job in our Eng Docs really)

I think the one thing I wanted to triple check was the work in the data transformer, mostly around the how HostnameGenerator's relate with meshes (which would be mostly for my own knowledge)

src/app/hostname-generators/data/HostnameGenerator.ts
```diff
@@ -0,0 +1,71 @@
...
+    const labels = item.labels ?? {}
+    const name = labels['kuma.io/display-name'] ?? item.name
+    const namespace = labels['k8s.kuma.io/namespace'] ?? ''
+    return {
+      ...item,
+      id: item.name,
+      name,
+      namespace,
+      zone: labels['kuma.io/origin'] === 'zone' && labels['kuma.io/zone'] ? labels['kuma.io/zone'] : '',
+      mesh: labels['kuma.io/mesh'] || 'default',
```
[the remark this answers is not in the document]
schogges: I am also not too sure about that, but the main driver for this decision was because I added a link to the mesh service on the details page. Iirc somewhere in the docs that `default` is the default for a `mesh`. Another thing is that if I am not mistaken, the hostname generators are only for `MeshService`, `MeshExternalService` and `MeshMultiZoneService`, which means that there is always a mesh, right?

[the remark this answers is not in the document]
johncowen: As far as I understood, HostnameGenerators are a top-level thing, i.e. they don't belong to a mesh they span the entire install/cluster.

> I am also not too sure about that, but the main driver for this decision was because I added a link to the mesh service on the details page (mesh is a param for the URL).

Oh! Glad you pointed this out 👍  not totally sure if we can link into a specific mesh from a HostnameGenerator 🤔

src/app/hostname-generators/data/HostnameGenerator.ts
```diff
@@ -0,0 +1,71 @@
...
+            routeName = 'mesh-multi-zone-service-list-view'
+          }
+        }
+
+        return {
+          routeName,
+          label: `${selector.charAt(0).toUpperCase()}${selector.slice(1, selector.length)}`,
+        }
+      })(Object.keys(item.spec.selector ?? {})?.[0]),
+    } satisfies HostnameGenerator
```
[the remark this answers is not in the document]
schogges: I see your point and I agree with it, we should infer the return type and not specify it 🙂 but I also think that in this case here it helps with code suggestions when using `satisfies`. The type is different from what is being used in the mocks. The type here is what should be returned from `fromObject` -> `HostnameGenerator extends HostnameGeneratorItem`. So with `satisfies` we are also not returning a specific type, but just make sure that during development we have proper feedback.
On the other hand we could also remove the type `HostnameGenerator` and wherever we need that type we use `ReturnType<typeof HostnameGenerator.fromObject>` (or `.fromCollection`). 🤔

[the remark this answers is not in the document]
johncowen: >On the other hand we could also remove the type HostnameGenerator and wherever we need that type we use ReturnType<typeof HostnameGenerator.fromObject> (or .fromCollection).

This ^ is exactly what the "New Way" involves, take a look in say `services/data/MeshService.ts`

src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
```diff
@@ -0,0 +1,104 @@
...
+
+    <DataLoader
+      :src="uri(sources, '/hostnamegenerators/:name', {
+        name: route.params.name,
+      })"
+      v-slot="{ data }"
+    >
+      <AppView :docs="t('hostname-generators.href.docs')">
+        <template
+          v-if="data"
```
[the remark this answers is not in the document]
schogges: Oh yeah 👍 🙂

src/app/hostname-generators/data/HostnameGenerator.ts
schogges: You are right 👍 🙂

src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
schogges: I've addressed this with the changes related to your suggestions regarding the routeNames in data layer 🙂

src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
schogges: I copied this over from another place that uses `AppAboutSection`. Thought it's required to display the inner content properly 🤔

src/app/hostname-generators/views/HostnameGeneratorListView.vue
schogges: Ah I remember 💡 thanks 👍

src/app/hostname-generators/views/HostnameGeneratorListView.vue
schogges: That's a very good point 👍 I like to have the consistency and go for a unified use of the hyphenated version in the GUI 🙂

src/app/hostname-generators/data/HostnameGenerator.ts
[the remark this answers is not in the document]
schogges: The selector `meshService`, `meshExternalService` or `meshMultiZoneService` together with a `mesh` would allow to link to the respective service type view 🤔

src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
schogges: So I've checked this again and with just one item in the view this can be omitted. As soon as we add more stuff to the `AppAboutSection` the layout is better with that `display: block` on the element.

src/app/hostname-generators/data/HostnameGenerator.ts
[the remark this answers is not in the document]
johncowen: We chatted about this, and saw that we can't use the top-level `kuma.io/mesh` for linking, if anything we'd generate any links form `matchLabels` instead. If we decide to do that this will be a future PR. We also removed the defaulting to `default` for `kuma.io/mesh`. In the future, from what I understand this top-level label will be removed entirely from HGs.

src/app/hostname-generators/data/HostnameGenerator.ts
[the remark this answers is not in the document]
johncowen: We changed this to use the "New Way"

johncowen (approved):
Top work, thanks! 🎉

PR #3175, opened 2024-11-12
fix: remove decorative attribute from all icons to fix tooltipping

johncowen:
Hardcodes `decorative="false"` in order to guarantee that `@kong/icons` show (`kong/icons` PR that introduced the issue)

---

Not totally sure how to go about resolving the issue as yet, ideally I'd like to use CSS for all our icons as all icons by their very nature are "decorative". Once we've decided exactly what to do we can revisit this. In the meantime this fix resolves things for our users whilst dodging/avoiding any issues over implementation.


schogges (approved):
👍

michaelbeaumont (approved):

PR #3176, opened 2024-11-12
chore: Migrate kuma-gui to use workspaces

johncowen:
We decided that moving this repository to use workspaces would give us more flexibility moving forwards. For example having distinct isolated npm modules instead of the more informal sub-folders allows us to expose specific interfaces for modules, plus workspaces/modules are a more recognised way to isolate modules, as opposed to just using sub-folders.

---

The first commit contains the filesystem moving, the rest of the commits contain the config changes I had to make as a result of that, so you might want to review by commit to only see the config changes.

---

I tried to keep all the changes (apart from the filesystem move) to a bare minimum, touching as little as possible to avoid any changes that might cause unforeseen problems. Once this filesystem change is merged in it will be easier to make some more incremental changes moving forwards (with not as much priority once we have a shared config over here)


schogges (approved):
🚀

michaelbeaumont (approved):
cool!

PR #3184, opened 2024-11-13
chore: adds new @kumahq/config module to house shareable config

johncowen:
TBD


packages/config/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "types": "./src/index.ts",
+  "exports": {
+    ".": {
+      "require": "./src/index.cjs",
+      "import": "./src/index.ts"
+    }
+  },
+  "dependencies": {
+    "@bahmutov/cypress-esbuild-preprocessor": "^2.2.3",
+    "esbuild": "^0.24.0"
```
schogges, line 19: I think `esbuild` should go into `devDependencies`? 🤔

schogges (commented):
I think we need to specify more dependencies in the `config/package.json` and if we don't use them in `kuma-gui` code remove them respectively from `kuma-gui/package.json` 🤔 

- `@badeball/cypress-cucumber-preprocessor`
- eslint plugins
- stylelint plugins

packages/config/package.json
johncowen: This is a dependency of `@kumahq/config`, its required to use the module if you are depending on it i.e. its a runtime requirement.

If we only need a dependency for the development of `@kumahq/config` specifically, then yes it should go in `devDepenendencies` here, but that is not the case for `esbuild`, its a dependency required by `@kumahq/kuma`.

For example, if we were to move `esbuild` to `devDependencies` and I `npm install`ed `@kumahq/kuma` from another project, then `esbuild` wouldn't be installed, because we've said "this is a dependency that is only required if you are developing `@kumahq/kuma`"

johncowen:
> I think we need to specify more dependencies in the config/package.json and if we don't use them in kuma-gui code remove them respectively from kuma-gui/package.json 🤔

> @badeball/cypress-cucumber-preprocessor
> eslint plugins
> stylelint plugins

I think I saw that `@badeball/cypress-cucumber-preprocessor` is ued elsewhere, but let me double check. But definitely can move the eslint and stylelint ones over, let me look at that 👍

johncowen:
I made a few tiny tweaks to the dependencies in that last commit (also fixed the tsconfig with the minimum I could that gave me no squiggly lines). lmk what you think!

Issue #3182, opened 2024-11-13
add xds config tab to inbound tray


lahabana:
This view:

<img width="1661" alt="image" src="https://github.com/user-attachments/assets/9ef3cf41-7377-47d0-919e-3d9a5bf71e3b">

Should have a new tab calls "XDS Configuration"

These come from `dynamic_active_clusters` and `dynamic_listeners` and have the name for listener `inbound:{{networking.address}}:{{inbound[].port}}` and `localhost:{{inbound[].port}}`. 

Should check with backend that it's always `{{networking.address}}` or not.

Here's the extracted data for the demo:

- for redis:

listeners:
```
        {
          "name": "inbound:10.244.0.6:6379",
          "active_state": {
            "version_info": "79d858b7-8577-4316-92fe-4babf4f5453d",
            "listener": {
              "@type": "type.googleapis.com/envoy.config.listener.v3.Listener",
              "name": "inbound:10.244.0.6:6379",
              "address": {
                "socket_address": {
                  "address": "10.244.0.6",
                  "port_value": 6379
                }
              },
              "filter_chains": [
                {
                  "filters": [
                    {
                      "name": "envoy.filters.network.tcp_proxy",
                      "typed_config": {
                        "@type": "type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy",
                        "stat_prefix": "localhost_6379",
                        "cluster": "localhost:6379",
                        "idle_timeout": "7200s"
                      }
                    }
                  ]
                }
              ],
              "metadata": {
                "filter_metadata": {
                  "io.kuma.tags": {
                    "pod-template-hash": "54754f5b57",
                    "k8s.kuma.io/service-port": "6379",
                    "app": "redis",
                    "kuma.io/zone": "default",
                    "kuma.io/protocol": "tcp",
                    "k8s.kuma.io/service-name": "redis",
                    "kubernetes.io/hostname": "mesh-zone",
                    "k8s.kuma.io/namespace": "kuma-demo",
                    "kuma.io/service": "redis_kuma-demo_svc_6379"
                  }
                }
              },
              "traffic_direction": "INBOUND",
              "bind_to_port": false,
              "enable_reuse_port": false
            },
            "last_updated": "2024-11-13T17:48:29.270Z"
          }
        },
```

cluster:
```
        {
          "version_info": "cec466d9-ad97-4258-85cc-8d80b87e1383",
          "cluster": {
[...]

Issue #3181, opened 2024-11-13
☔ Add XDS config tab to outbound and inbound view


lahabana:
### Description

For outbound it should work only with `MeshService`.

Now that we have predictable names for Clusters we should be able to show the XDS config for a specific outbound.

We would have a new tab next to config/stats/clusters and show just the data we are interested in. This make it a lot easier for folks to understand what the XDS config for an outbound is and to understand what could be happening. 

For only do DPPs (no gateway proxies).

- [x] #3182
- [ ] add xds config tab to outbound tray with only xds cluster
- [ ] add xds config tab to outbound tray with xds cluster and listeners
- [ ] #3236
- [ ] #3237

lahabana:
Here's some XDS config for demo-app and redis:
[demo-app-xds.json](https://github.com/user-attachments/files/17737129/demo-app-xds.json)
[redis-xds.json](https://github.com/user-attachments/files/17737130/redis-xds.json)

PR #3186, opened 2024-11-14
feature(dataplanes): add filtered XDS config to drawers

johncowen:
Adds XDS configuration drawers to inbounds and outbounds so show only the listeners/clusters relevant to that inbound/outbound.

Closes https://github.com/kumahq/kuma-gui/issues/3182
Closes https://github.com/kumahq/kuma-gui/issues/3181

You can run this against a local cluster by setting a cookie like the following:

Screenshot 2024-11-25 at 11 28 49

note: I removed an annoying ESLint rule here https://eslint.org/docs/latest/rules/multiline-ternary#when-not-to-use-it . We don't have any conventions in regards to ternaries, do whatever works for you. Generally you've decided to use a ternary because the format you've personally chosen reads nicer - furthermore its pointless having a eslint rule to force wrapping of something that is designed for single lines. Not only that but the eslint rule we had makes these super hard to read.

---

~Note: This will not pass lint tests and is purposefully an early WIP PRed for preview site purposes.~

---

Notes taken during work:

Adding some pseudo code to help explain the filtering/selecting we are doing:

### Inbounds 

I use `'127.0.0.1:0000'` below in place of the actual `dataplane.networking.address`

- `dynamic_listeners[].name === '127.0.0.1:0000'`
- `{dynamic_active_clusters, static_clusters}.cluster.load_assignment.endpoints[].lb_endpoints[].endpoint.address.socket_address.{address:port_value} === '127.0.0.1:0000'` (I'm not totally sure we should be using this one)

Note for an inbound we do not know the type of the service (i.e. is it a `_msvc_` a `_mzsvc_` etc) so we can't filter by cluster name.

---

Improvements: we should look for `inbound:ignore.ip.address:0000` use prefix/suffix because we need to be careful with ipv6.

- `dynamic_listeners[].name === 'inbound:ignore.ip.address:0000'`
- `dynamic_active_clusters[].cluster.name === 'ignore.host:0000'`

### Outbounds

I use `outbound` below in place of the actual Envoy stats name (e.g. `default_demo-app_kuma-demo_default_msvc_5000`)
 
- `dynamic_active_clusters[].cluster.name === <outbound>`
- `dynamic_endpoint_configs[].endpoint_config.cluster_name === <outbound>`

Note we don't have any information about a specific outbound apart from the clusterName/envoy stats prefix. We can take information from the dataplane associated with the outbound, but not the outbound itself.

Improvements:

- `dynamic_listeners[].name === 'outbound:<outbound>'` to be future proof with https://github.com/kumahq/kuma/issues/12093.

If we need to add more instance of filtering/selecting it would be helpful to express what needs to be done using the above examples


lahabana:
Looks great! There's only 1 bug with clusters

https://konghq.zoom.us/clips/share/A2F3MRZnZDRPSjZfT1E5bVhNRTRERWl1MnlnAQ

johncowen:
PR to fix up the issue with clusters is here:

https://github.com/kumahq/kuma-gui/pull/3239

schogges:
> note: I removed an annoying ESLint rule here https://eslint.org/docs/latest/rules/multiline-ternary#when-not-to-use-it . We don't have any conventions in regards to ternaries, do whatever works for you. Generally you've decided to use a ternary because the format you've personally chosen reads nicer - furthermore its pointless having a eslint rule to force wrapping of something that is designed for single lines. Not only that but the eslint rule we had makes these super hard to read.

I agree with this 👍 since you speak about readability, I'd like to bring up [no-nested-ternary](https://eslint.org/docs/latest/rules/no-nested-ternary#rule-details) 🙂 I think this might be a good addition to our lint rules as it enforces short and one-lined ternary expressions, which are great. But when it comes to more complex flows, I'd rather use `if-else` or a mix of ternary and `if-else`. Nesting ternaries hurts readability IMO. (Have not seen any nested ones yet tbh 😄 )

johncowen:
> I agree with this 👍 since you speak about readability, I'd like to bring up [no-nested-ternary](https://eslint.org/docs/latest/rules/no-nested-ternary#rule-details) 🙂

More than happy to take look at it.

The only thing is ternaries are about more then readability, they allow you to have type safety at runtime in javascript, which while I don't think is as important as the type safety, I believe is also a micro-performance improvement in javascript engines (but lets not get into micro-optimizations, thats side benefit)

Its not so much using the ternaries, its more the avoidance of using `let`.

```javascript
let a = '1'
if(true) {
  a = '2'
}
// a is a string
```

vs

```javascript
const a = true ? '2' : '1'
// a is '1' | '2'
```

The second form:

- The JS engine knows `a` can only ever be `'1'` or `'2'` , not any-string (in TS) or anything (in JS), and therefore optimize.
- The JS engine knows that `a` will never be set again, and therefore optimize.

So I use these inline forms a lot (and did previous to Typescript) via ternaries and inline functions for `try/catch` and `switch` among other constructs for these reasons, not necessarily for readability. If there's a nested ternary somewhere there's a chance its for the above reasons.

I have an example of a change a made recently that outlines this, lemme try and find it.

packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_/xds.ts
```diff
@@ -1,521 +1,717 @@
 import type { EndpointDependencies, MockResponder } from '@/test-support'
-export default (_deps: EndpointDependencies): MockResponder => (_req) => {
+export default ({ env, fake }: EndpointDependencies): MockResponder => (req) => {
+  const query = req.url.searchParams
+  const { name } = req.params
+  const eds = query.get('include_eds') === 'true'
+
+  // use a seed based on the name to keep ports and ip address the same across
+  // _overview, stats and rules
+  fake.kuma.seed(name as string)
```
schogges, line 9: 💡 👍

packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_/xds.ts
schogges, line 1: There is a lot of hardcoded values in this mock. Is this on purpose?

johncowen: Not on purpose as such, just there's no need to do that as yet at least.

So we don't _use_ this data at all we just show it, it could be any old JSON.

I guess also, there's no documentation that I know of for this format, so we could go through and try to randomise all the types but we could get it completely wrong.

That being said, I have randomized the bits that I do know about, which are also the bits we depend on for the filtering.

So I suppose yes 😄 , I purposefully only randomized the bits we depend on because I don't want to prematurely randomize the rest when:

1. There's no need right now.
2. I'm not knowledgable enough on the format right now to do that correctly.

packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_/xds.ts
johncowen: 👍 I should spend some time documenting this a little more. This is how we do state in a stateless mocking framework. The state is in the URL.

schogges:
> > I agree with this 👍 since you speak about readability, I'd like to bring up [no-nested-ternary](https://eslint.org/docs/latest/rules/no-nested-ternary#rule-details) 🙂
> 
> More than happy to take look at it.
> 
> The only thing is ternaries are about more than readability, they allow you to have type safety at runtime in javascript, which while I don't think is as important as the type safety, I believe is also a micro-performance improvement in javascript engines (but lets not get into micro-optimizations, thats side benefit)
> 
> Its not so much using the ternaries, its more the avoidance of using `let`.
> 
> ```js
> let a = '1'
> if(true) {
>   a = '2'
> }
> // a is a string
> ```
> 
> vs
> 
> ```js
> const a = true ? '2' : '1'
> // a is '1' | '2'
> ```
> 
> In the second form:
> 
> * The JS engine knows `a` can only ever be `'1'` or `'2'` , not any-string (in TS) or anything (in JS), and therefore optimize.
> * The JS engine knows that `a` will never be set again, and therefore optimize.
> 
> So I use these inline forms a lot (and did previous to Typescript) via ternaries and inline functions for `try/catch` and `switch` among other constructs for these reasons, not necessarily for readability. If there's a nested ternary somewhere there's a chance its for the above reasons.
> 
> I have an example of a change a made recently that outlines this, lemme try and find it.
> 
> edit:
> 
> Not a ternary but its the same idea, if you take a look at this code from a recent PR (😆 oh its actually this PR):
> 
> https://github.com/kumahq/kuma-gui/blob/765d9f5f2ca3fdc44b5998f946064ed9aa725c7b/packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_/_overview.ts#L21-L26
> 
> In #3186 I'm changing it to this:
> 
> https://github.com/kumahq/kuma-gui/blob/c42fe8de8a10144cd1104567d394162eecc6314f/packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_/_overview.ts#L24-L35
> 
> Using the inline function gives me automatic type safety both in TS at build time and JS at runtime, and honestly in this case, in avoiding the multiple if conditionals the inline-function/switch is far easier to read to my eyes at least.

Yeah, I like the `switch-case` 👍 😄 another great way of avoiding nested (or chained) ternaries, having the optimization benefits and it's a lot easier to read and understand.

packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_/xds.ts
schogges: I see 👍 🙂

packages/kuma-gui/src/app/data-planes/sources.ts
```diff
@@ -30,7 +30,37 @@ export type MeshGatewayDataplaneSource = DataSourceResponse<MeshGatewayDataplane
 const includes = <T extends readonly string[]>(arr: T, item: string): item is T[number] => {
   return arr.includes(item as T[number])
 }
+const prop = <K extends PropertyKey>(obj: unknown, key: K | null | undefined): obj is Record<K, unknown> => {
```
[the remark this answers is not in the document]
johncowen: From what I've seen most TS folks use a single letter (mostly `T`, but sometimes `K` or others especially when there are multiple generic args). Our application generally follows this convention also.

🤔  I think actually at one point I was asked _not_ to use words and use letters instead, so I think I'd rather stick to that unless its an overly complicated instance.

Happy to change `K` to `T` if you like? I don't mind either way though, I'm more than happy to stick to `K` seeing as its a key, but I'm happy to change. lmk

johncowen:
Oh look I found a nested-ternary:

https://github.com/kumahq/kuma-gui/blob/61c99f3df9964a9a4e23da9cbb64c28197b338c6/packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts#L92

🤔 when I think about what the above file used to be like, I kinda prefer the direction it's going.

I think that, the fact that we have them already, and the type/js benefits I wouldn't be super keen on adding that rule, plus the more i think about it its very similar to saying "don't use multiple conditionals inside an if statement" (i.e. `&&`'s and `||`'s)

If you feel super strong about it happy to reconsider, but at the moment it doesn't feel like a super important error causing problem that we might want to add a lint rule for.

packages/kuma-gui/src/app/data-planes/sources.ts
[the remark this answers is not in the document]
schogges: I know that there is a lot of single character generics and `T` is definitely taking the lead 😄 but I've also seen a lot what I've described above and I like the readability.
I'm fine either way and nothing to do here, just wanted to know your opinion 🙂

schogges (approved):
LGTM! 👍

schogges:
> Oh look I found a nested-ternary:
> 
> https://github.com/kumahq/kuma-gui/blob/61c99f3df9964a9a4e23da9cbb64c28197b338c6/packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts#L92
> 
> 🤔 when I think about what the above file used to be like, I kinda prefer the direction it's going.
> 
> I think that, the fact that we have them already, and the type/js benefits I wouldn't be super keen on adding that rule, plus the more i think about it its very similar to saying "don't use multiple conditionals inside an if statement" (i.e. `&&`'s and `||`'s)
> 
> If you feel super strong about it happy to reconsider, but at the moment it doesn't feel like a super important error causing problem that we might want to add a lint rule for.

I think ternaries are getting unreadable a lot quicker when nesting than a complex boolean expression. But yeah, at some point it probably makes sense to outsource and group them into variables.
Anyways, will think about it and will probably bring it up again 🙂

johncowen:
One last thing I wanted to do here thats vaguely related to above convo, I added `fake.kuma.certificate` for various reasons here. Then typically I had to stabilise a test because of that. I did it in separate commits in case.

I'll leave this hanging for a little while today, but if I hear no more I'll go ahead and merge seeing as I already have approval on everything else. But lmk if there's anything else

PR #3189, opened 2024-11-15
feat(kuma-gui/control-planes): add ControlPlaneActionGroup with link to HostnameGenerators

### LINKED ISSUE #3174: Add link to `HostnameGenerator` list view to home
### Description

This is a follow-up to #2956.

We need to find a good spot to put the link for the `HostnameGenerator` list view on the home view or into the side-/topbar.
`HostnameGenerator`s are not specific to a `mesh` or a `zone` and are a top level resource. They are not as important as `mesh` and `zone` resources and therefore should not be on the same level of navigation options. We expect that only a few users will check the views.

schogges:
This change introduces a new `XActionGroup` to the root overview (`ControlPlaneDetailView`) and includes a link to the `HostnameGenerators`.

image

Closes #3174


packages/kuma-gui/src/app/x/components/x-action-group/XActionGroup.vue
```diff
@@ -13,7 +13,7 @@
         :kpop-attributes="{
           placement: 'bottom-end',
         }"
-        width="150"
+        :width="width"
```
schogges, line 16: Note: The width was hardcoded to 150. This caused that long labels of action items overflow the borders of the action group dropdown.

packages/kuma-gui/src/app/control-planes/locales/en-us/index.yaml
```diff
@@ -18,3 +18,7 @@ main-overview:
       title: 'Zones'
     meshes:
       title: 'Meshes'
+  action_menu:
+    toggle_button: 'Actions'
+    items:
+      hostname_generators: HostnameGenerators
```
schogges, line 24: Is the label fine?

johncowen: If we are staying consistent, I think this needs a space at least for now. We use `Mesh Services`/`Mesh External Services` etc elsewhere. TBH I kinda like using the actual config name (no space), but if we want to change that we should do it everywhere in one go.

packages/kuma-gui/src/app/x/components/x-action-group/XActionGroup.vue
johncowen: Do you know what happens if we use `auto` for the default, and then don't use the 150 width at all anywhere? 

I'm guessing the 150 might have just come from where we took this from originally. Actually, hold on let me have a little snoop around I might be able to see why we did this. brb

Ideally we wouldn't have to specify sizes and the menu would just shrink/grow to contain the content.

johncowen (commented):
Looks good to me!

Couple of little nits, one of them is for me to look at quickly, brb

packages/kuma-gui/src/app/x/components/x-action-group/XActionGroup.vue
johncowen: hmmm yeah ideally we need a min-width of `150`

## with 150
Screenshot 2024-11-18 at 17 31 20

## without 150
Screenshot 2024-11-18 at 17 31 31

schogges: > hmmm yeah ideally we need a min-width of `150`

I don't think that this is a solution that is generic enough. If we would, for whatever reason, decide to rephrase the content to just `View` or add another dropdown somewhere with a very short action item label, the `150px` would probably too big again 🤔

johncowen: Ah think we crossed over, I just tried out your point 3 also:

```css
.x-action-group :deep(.popover-container) {
  min-width: 150px;
}
```

I also had to keep `width="auto"` on the KDropdown otherwise it seems to get set to `width: 200px`
 even though "The width of the dropdown body (defaults to auto)." (https://kongponents.konghq.com/components/dropdown.html#width)

I think we should go with that. There is no chance we'll have a i18n'ed GUI if we have hardcoded sizes (I think this originally came from elsewhere). I think if we go with point 3, we stay consistent with elsewhere, but things grow once we start properly i18n'ing things. WDYT?

johncowen (approved):
This LGTM! Nice work on the more scalable menus! 🎉

Issue #2922, opened 2024-09-12
Add YAML to summary drawers everywhere


johncowen:
### Description

IMO lists of things in GUIs should be either:

1. A way to get somewhere
2. A way to compare things

NOT: A way to show "all-the-informations".

Point 2 above, along with our summary drawers make it easier to compare things such as YAML config. Not necessarily "at a single glance" as we can't display the YAML in every single row. But by single clicking we can show the YAML in each summary drawer for the row that you click on making it easy to click on several rows after each other to compare YAMLs.

This make it really easy to compare resources without having to navigate backwards and forwards between different pages and tabs.

We already have several places where we show the YAML in the summary drawer, sometimes in a "sticky tab" in a drawer, sometimes in a drawer with a `[Structured]/[YAML]` dropdown, and sometimes just the YAML on its own.

I think in a lot of places we can add the YAML on its own, for the summaries that are already quite busy we can decide how to add the YAML on a case by case basis. We may later decide for the sake of consistency to choose one approach whether we have a busy summary or not.

johncowen:
Wanted to note that there are at least 2 places where we might _not_ want to show a yaml or drawer at all:

1. Legacy kuma.io services (aka internal services). Reason: These don't have a config to show, so we can't show any YAML. The listing is generally "A way to get somewhere", and that somewhere is generally the list of dataplanes for the service.
2. Meshes. This is a bit more tricky as Meshes _do_ have a YAML config, but! We added the mesh listing page as a replacement for the Mesh selector that we used to have in the left hand main nav (it was a native select menu, which when selected would take you straight to that mesh) i.e. this listing is primarily "A way to get somewhere".

Wanted to leave that info here. I suppose we could still show drawers for both, and in the case of Meshes the YAML also, but we should be intentional about doing _or_ not doing that taking the above points into account.

I don't think there are any other resources that have these question marks over them, for example I know that Dataplane and Gateways do not have a drawer with their YAML in them.

Hope this helps, please shout if the above isn't clear or if anyone had more to add/discuss.

lahabana:
One thing is missing.

In https://deploy-preview-3262--kuma-gui.netlify.app/gui/meshes/default/data-planes/feed-proxy-8/summary-overview

- For dataplane overview remove inbounds parsed list and inline yaml (don't have tabs)
- For subscription use the Konnect structured/json view  (see RBAC in KM)

johncowen:
Lets make a new issue for this:

[...]

PR #3202, opened 2024-11-18
feat(zone-ingress/summary-view): add config

schogges:
Adds the `YAML` config to the `ZoneIngressSummaryView`.
I've also removed some of the custom types that are added in the data layer to move further to the point of inferring the type from the return value.

Part of #2922

### DIFF
--- a/packages/kuma-gui/src/app/zone-ingresses/data/index.spec.ts
+++ b/packages/kuma-gui/src/app/zone-ingresses/data/index.spec.ts
@@ -281,4 +281,14 @@ describe('ZoneIngressOverview', () => {
       },
     )
   })
+  describe('zoneIngressOverview.$raw', () => {
+    test('$raw is the same as original API response', async ({ fixture }) => {
+      let expected
+      const actual = await fixture.setup((item) => {
+        expected = item
+        return item
+      })
+      expect(actual.$raw).toStrictEqual(expected)
+    })
+  })
 })
--- a/packages/kuma-gui/src/app/zone-ingresses/data/index.spec.ts
+++ b/packages/kuma-gui/src/app/zone-ingresses/data/index.spec.ts
@@ -281,14 +281,45 @@ describe('ZoneIngressOverview', () => {
       },
     )
   })
-  describe('zoneIngressOverview.$raw', () => {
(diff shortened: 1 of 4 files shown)


johncowen:
> I've also removed some of the custom types that are added in the data layer to move further to the point of inferring the type from the return value.

Thanks for doing this ☝️ 🎉 

---

Separately, something I forgot to note before. This YAML is showing the `ZoneIngressOverview` which isn't an actual resource, its an HTTP API only thing.

Ideally here we want to show the config for the `ZoneIngress` I think, so if the aim is to show the copyable config of the thing.

It would be the equivalent of:

Screenshot 2024-11-18 at 11 39 55

I thought this little weirdness only happened for Dataplanes, but just realizing myself now that because ZoneIngress/Egresses are just Envoy instances (i.e. Dataplanes) we also have this weirdness here.

schogges:
> Ideally here we want to show the config for the ZoneIngress I think, so if the aim is to show the copyable config of the thing.

Ah, thank you 👍 🙂

schogges:
@johncowen I've reworked the part in the data layer and now include the `config` which is derived from `item.zoneIngress` and some more item metadata. Doing this gives the same result as and saves us making a request to `/zone-ingresses/:name` in the summary view. Please let me know what you think 🙂

johncowen:
> Looks good to me, I think we just need to lose the availableServices which I think are added in the backend and therefore not part of the config.

Actually, we need to be intentional about what we are showing here, and then keep it consistent. I suppose the question is are we showing:

1. The `config` or;
2. The "_result_ of what your `config` has done".

Now I think about it I'm not totally sure which we want to do.

johncowen:
Just to note down what we chatted through offline:

---

For the moment at least we decided that we would go with showing:

> The config  that is currently applied, including all the stuff that gets added by the backend (and potentially k8s)

Reasons being:

1. I think this is what we currently do everywhere else
2. Chances are the user already has the config because they just applied it (although there is a usecase of using the GUI to share the config)
3. The interesting thing is the representation of what that config has done, which the user doesn't have

We can always come back and revisit if we decide different, but we should try to be consistent as to what these panels are showing.

PR #3203, opened 2024-11-18
feat(zone-egresses/summary-view): add config

schogges:
Adds the `YAML` config to the `ZoneEgressSummaryView`.
I've also removed some of the custom types that are added in the data layer to move further to the point of inferring the type from the return value.

The `config` is aligned with what was added in #3202:

> The `config` that we add to the data for now represents what is currently applied, including all the stuff that gets added by the backend (and potentially k8s). We might revisit this in the future and add only what the user would apply and exclude everything that is added by the backend (and potentially k8s).

Part of #2922


johncowen (approved):
LGTM!

PR #3209, opened 2024-11-19
feat(data-planes/data-plane-summary-view): add config

schogges:
This change introduces a tab navigation to the `DataPlaneSummaryView` offering an `Overview` and `Config` tab. The reason for the separation into tabs was because the summary view could get very crowded already. The `Overview` tab shows everything that was previously shown on the summary view. The `Config` tab shows the config, which is based on some metadata and `dataplane.network` data of a `DataPlaneOverview` item (similar to what was done in #3202 and #3203).

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
+++ b/packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
@@ -1,3 +1,4 @@
+import { Dataplane } from './Dataplane'
 import { DataplaneInsight } from './DataplaneInsight'
 import { DataplaneNetworking } from './DataplaneNetworking'
 import type { PaginatedApiListResponse } from '@/types/api.d'
@@ -6,21 +7,6 @@ import type {
   DataplaneWarning,
   LabelValue,
 } from '@/types/index.d'
-export type DataplaneOverview = PartialDataplaneOverview & {
-  dataplane: {
-    networking: DataplaneNetworking
-  }
-  id: string
-  namespace: string
-  labels: Exclude<PartialDataplaneOverview['labels'], undefined>
-  dataplaneInsight: DataplaneInsight
-  dataplaneType: 'standard' | 'builtin' | 'delegated'
-  status: 'online' | 'offline' | 'partially_degraded'
-  warnings: DataplaneWarning[]
-  isCertExpired: boolean
-  zone: string | undefined
-  services: string[]
-}
 
 // any collection of non-spaces followed by a `:` or `: ` followed by a
 // collection of non-spaces
@@ -39,6 +25,17 @@ const searchShortcuts: Record<string, string> = {
   zone: 'kuma.io/zone',
   protocol: 'kuma.io/protocol',
 }
+const states = {
+  online: 'online',
+  offline: 'offline',
+  partiallyDegraded: 'partially_degraded',
+} as const
+const dpTypes = {
+  builtin: 'builtin',
+  delegated: 'delegated',
+  standard: 'standard',
+} as const
+
 export const DataplaneOverview = {
   search(search: string) {
     const terms = [...search.matchAll(searchRe)].filter(item => item[0].length > 0).map(item => item[0].trim())
@@ -61,7 +58,7 @@ export const DataplaneOverview = {
       })(prev, item)
     }, { tag: [] }) || {}
   },
-  fromObject(item: PartialDataplaneOverview, canUseZones: boolean): DataplaneOverview {
+  fromObject(item: PartialDataplaneOverview, canUseZones: boolean) {
     const dataplaneInsight = DataplaneInsight.fromObject(item.dataplaneInsight)
 
     const networking = DataplaneNetworking.fromObject(item.dataplane.networking)
@@ -73,6 +70,15 @@ export const DataplaneOverview = {
     const zone = tags.find((tag) => tag.label === 'kuma.io/zone')?.value
     const labels = typeof item.labels !== 'undefined' ? item.labels : {}
 
+    const { config } = Dataplane.fromObject({
+      type: 'Dataplane',
+      name: item.name,
+      mesh: item.mesh,
+      creationTime: item.creationTime,
+      modificationTime: item.modificationTime,
+      networking,
+    })
+
     return {
       ...item,
       id: item.name,
@@ -83,9 +89,9 @@ export const DataplaneOverview = {
       },
       labels,
       dataplaneInsight,
-      dataplaneType: networking.type === 'gateway' ? 'builtin' : typeof networking.gateway !== 'undefined' ? 'delegated' : 'standard',
+      dataplaneType: networking.type === 'gateway' ? dpTypes.builtin : typeof networking.gateway !== 'undefined' ? dpTypes.delegated : dpTypes.standard,
       status: (() => {
-        const state = typeof dataplaneInsight.connectedSubscription !== 'undefined' ? 'online' : 'offline'
+        const state = typeof dataplaneInsight.connectedSubscription !== 'undefined' ? states.online : states.offline
         if (networking.gateway) {
           return state
         }
@@ -94,10 +100,10 @@ export const DataplaneOverview = {
         switch (true) {
           case unhealthyInbounds.length === networking.inbounds.length:
             // All inbounds being unhealthy means the Dataplane is offline.
-            return 'offline'
+            return states.offline
           case unhealthyInbounds.length > 0:
             // Some inbounds being unhealthy means the Dataplane is partially degraded.
-            return 'partially_degraded'
+            return states.partiallyDegraded
           default:
             // All inbounds being healthy means the Dataplane’s status is determined by whether it’s connected to a control plane.
             return state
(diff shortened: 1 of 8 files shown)


packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -39,6 +25,17 @@ const searchShortcuts: Record<string, string> = {
...
+const states = {
+  online: 'online',
+  offline: 'offline',
+  partiallyDegraded: 'partially_degraded',
+} as const
+const dpTypes = {
+  builtin: 'builtin',
+  delegated: 'delegated',
+  standard: 'standard',
+} as const
```
schogges, line 37: Note: I added these to avoid writing too many `... as const`

packages/kuma-gui/src/app/data-planes/routes.ts
```diff
@@ -67,7 +67,29 @@ export const routes = () => {
...
               path: ':dataPlane',
               name: 'data-plane-summary-view',
+              props: () => ({
+                routeName: 'data-plane-summary-view',
+              }),
+              redirect: { name: 'data-plane-summary-overview-view' },
               component: () => import('@/app/data-planes/views/DataPlaneSummaryView.vue'),
+              children: [
+                {
+                  path: 'summary-overview',
```
schogges, line 77: Note: I went for `/summary-overview` route name as `/overview` is already taken by the `DataPlaneDetailView`

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -73,6 +70,15 @@ export const DataplaneOverview = {
...
 
+    const { config } = Dataplane.fromObject({
+      type: 'Dataplane',
+      name: item.name,
+      mesh: item.mesh,
+      creationTime: item.creationTime,
+      modificationTime: item.modificationTime,
+      networking,
+    })
+
```
johncowen, line 81: Is this the `config` we eventually show in the code editor?

If so, if we where to add some of our own custom properties within `Dataplane.fromObject` I think that would mean that we might then mistakenly display these custom properties in the code editor.

I suppose the TLDR; is do we want to use `Dataplane.fromObject` or just fix up the properties that are missing using a bit of:

```js
{
  ...item.dataplane,
  name: item.name,
  ...etc
}
```

Am I making sense? Let me know if not!

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -191,3 +198,5 @@ function getWarnings({ version }: DataplaneInsight, tags: LabelValue[], canUseZo
 
   return warnings
 }
+
+export type DataplaneOverview = ReturnType<typeof DataplaneOverview.fromObject>
```
johncowen, line 202: Thanks! 👏

packages/kuma-gui/src/app/data-planes/views/DataPlaneSummaryView.vue
```diff
@@ -1,15 +1,14 @@
 <template>
   <RouteView
-    name="data-plane-summary-view"
+    :name="props.routeName"
```
johncowen, line 3: Awesome 👏

packages/kuma-gui/src/app/services/routes.ts
```diff
@@ -74,13 +63,7 @@ export const routes = (can: Can) => {
...
                     name: 'mesh-multi-zone-service-detail-view',
                     component: () => import('@/app/services/views/MeshMultiZoneServiceDetailView.vue'),
-                    children: [
-                      {
-                        path: ':dataPlane',
-                        name: 'mesh-multi-zone-service-data-plane-summary-view',
-                        component: () => import('@/app/data-planes/views/DataPlaneSummaryView.vue'),
-                      },
-                    ],
+                    children: dataPlanes().summary('mesh-multi-zone-service'),
```
johncowen, line 66: Just wanted to make a note here to share thinking/ideas. Just a note for the future nothing to be done now.

---

Eventually I would like this sort of stuff to be done from The Outside i.e. something like:

```js
// probably from within main.ts, or possibly `module-name/index.ts`
mount(summaryRoutes, 'mesh.dataplanes.summary')
mount(summaryRoutes, 'mesh.service.dataplanes.summary')
```

i.e. "mount these routes at this point in the routing tree please"

johncowen (commented):
A did a code scan and left one comment/question about Dataplane.fromObject usage, lemme know on that one.

Separately I had another comment on the UX:

If you look at the current view it "almost" has all the same information that we can see in the YAML, just the YAML is more compact (and for some, including me, easier to parse/read). Not only that but it can have information that I don't have in the current overview (`outbounds` for example). The bad thing about the YAML is, we don't link or provide copy buttons for inside the YAML.

I wonder if we should have the YAML as the first tab (staying consistent with the other places that we are adding YAML/already have the YAML), and move the Inbounds to a separate tab (if we think its important to keep the non-YAML inbounds rendering).

Then we can maybe lose one or two lines of this stuff maybe:

Screenshot 2024-11-20 at 11 15 07

wdyt? You see my preference of having the YAML on the first tab? actually let me know what you think there, I just had another thought brb

johncowen:
P.S. I just spotted this (the `&ref`) I think not using `Dataplane.fromObject` could solve this, lemme know if not.

Screenshot 2024-11-20 at 11 18 03

johncowen:
> I just had another thought brb

Totally shouldn't do all of the below in this PR (if at all). But, what do you think of this idea:

Untitled

My thoughts behind this:

- YAML is on the first tab, like all our other drawers.
- Inbounds is in the second tab in "UI format" for if you find it easier instead of seeing it in the YAML
- "state" is underneath the title of the thing. This is important information, and is not directly visible from looking at the YAML (we derive/combine/calcuate information to show this "state"). Having it up here also means its visible no matter what tab you are looking at, which is good as its probably the most important piece of information.
- We move "Type: Proxy" to the icon w/tooltip. This is consistent with the icons in the listings.

Just to reiterate, should definitely **not** do all of this in this PR, but maybe just move the YAML to the first tab, and put the inbounds on the second, which feels like its within scope of this line of work. For the rest we should probably check with others on this idea and do a completely separate PR

Lmk what you think!

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
schogges: Using `Dataplane.fromObject` and only taking the `config` entry makes sure that we always show that same stuff in the config. We already show the config in the tab in the details view 🤔

packages/kuma-gui/src/app/services/routes.ts
schogges: Definitely that would be a great to have 👍 🙂

schogges:
> P.S. I just spotted this (the `&ref`) I think not using `Dataplane.fromObject` could solve this, lemme know if not.
> 
> Screenshot 2024-11-20 at 11 18 03

It's weird because it's supposed to show the exact same stuff as in the tab of the detail view 🤔

schogges:
> > I just had another thought brb
> 
> Totally shouldn't do all of the below in this PR (if at all). But, what do you think of this idea:
> 
> Untitled
> My thoughts behind this:
> 
> * YAML is on the first tab, like all our other drawers.
> * Inbounds is in the second tab in "UI format" for if you find it easier instead of seeing it in the YAML
> * "state" is underneath the title of the thing. This is important information, and is not directly visible from looking at the YAML (we derive/combine/calcuate information to show this "state"). Having it up here also means its visible no matter what tab you are looking at, which is good as its probably the most important piece of information.
> * We move "Type: Proxy" to the icon w/tooltip. This is consistent with the icons in the listings.
> 
> Just to reiterate, should definitely **not** do all of this in this PR, but maybe just move the YAML to the first tab, and put the inbounds on the second, which feels like its within scope of this line of work. For the rest we should probably check with others on this idea and do a completely separate PR
> 
> Lmk what you think!

Hmmm 🤔 here are my thoughts 🙂 

* I've checked two other places and in there `YAML` is the second tab (see screen grabs below). But definitely for consistency I will rename `Config` to `YAML` 👍 
* From my point of view both tabs can duplicate information. They are just different ways of presentation. One user might prefer the `YAML`, the other user might prefer the separation. But we should probably check what information is most important and keep the summary short. The user can go to the details for more information.
* Adding the `state` a hierarchy level up sounds good to me 👍 

Screenshot 2024-11-20 at 16 46 38
Screenshot 2024-11-20 at 16 45 55

johncowen (commented):
Cool! I saw you managed to get to the bottom of the `&ref` thing, and I didn't see that anymore in the PR preview.

I think this is good to go! I think I saw one test fail that needs fixing, lemme know when you are green.

packages/kuma-gui/src/app/data-planes/routes.ts
schogges: Generally it worked for me having both using `/overview`, but when it comes to deeplinking it always went for the closest route, which is the summary. So in order to allow deeplinking to the details overview I had to use different path names 🤔 I might have missed something, so I am looking forward to discuss this further 🙂

johncowen:
Oh just realised something else, but no changes needed for this here. Probably worth discussing a little more and if we change then we do it in a later PR.

> I've checked two other places and in there YAML is the second tab (see screen grabs below).

I think these are the only places where there is some sort of 'YAML' in the second tab, but these are for subscriptions (so one 'type'). I think (but not totally sure) that everywhere else (Mesh*Services, Zone*gresses i.e. our configurable resources) its in the first tab.

Important thing to note is, subscriptions aren't configurable like all our resources so its not quite the same thing. So whilst we could also put these in the first tab (or maybe leave them as they are), I would treat that decision as different to this one due to the fact that subscriptions aren't quite the same thing as our other resources.

Anyway wanted to share thoughts here, happy for this PR to stay as is, I've just seen whilst writing this the CI is currently running, so should be able to approve this in a sec 👀

PR #3210, opened 2024-11-20
chore(ci): installs cypress-terminal-report for surfacing console.warns

johncowen:
There are places where we can sometimes have certain HTTP errors swallowed and converted to `console.warn`, currently its up in the air whether we leave them alone or convert them to console.errors (so they fail our tests using `cypress-fail-on-console-error` that we already have installed).

This PR install a further dependency that allows us to log `console.warn`s.

Note there are other things that it allows us to log, but for the moment I'm looking for low noise and just output `console.warn`s. Also, we can't fail on console.warn because we have 3rd party dependencies that use `warn` correctly and for good reason.

---

I confirmed we are seeing console.warns in the tests in GH logs. You can apparently get the plugin to output a txt file so you can make an artifact, but I plan on adding this later if this plugin proves useful. For now it will be good to merge this as is so we can start watching for console.warn.

Screenshot 2024-11-20 at 12 45 38

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5277,6 +5277,11 @@
         "node": ">=4.0.0"
       }
     },
+    "node_modules/compare-versions": {
+      "version": "6.1.1",
+      "resolved": "https://registry.npmjs.org/compare-versions/-/compare-versions-6.1.1.tgz",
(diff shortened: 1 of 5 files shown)


schogges (approved):
That's super helpful 👍 🙂

PR #3213, opened 2024-11-21
refactor(app/navigation): split sidebar navigation, move diagnostics to configuration

schogges:
This change splits up the sidebar navigation into a top nav including the nav links to the most important routes. The new bottom nav offers a place for other routes. Both navigation lists are separated by a visual separator. This ensures that the user finds all nav links to top level resources in the same area of the UI. Both nav lists are only split visually and remain within a single `nav` element.
To further beautify the nav links I've added icons to each of them.

I've moved and renamed everything from `diagnostics` to `configuration`. This makes it more clear that this part of the UI shows the CP configuration.


packages/kuma-gui/src/assets/images/navigation/icon-zone-egresses.svg
schogges, line 1: Note: I'm if this icon is good for zone egresses 🤔

packages/kuma-gui/src/app/application/components/app-navigator/AppNavigator.vue
```diff
@@ -31,7 +31,6 @@ const props = withDefaults(defineProps<{
   &::before {
     content: '';
     mask-image: var(--icon);
-    -webkit-mask-image: var(--icon);
```
schogges: Checked and the autoprefixer takes care of this 🙂

packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
```diff
@@ -117,33 +117,38 @@
...
+          <nav
+            v-if="$slots.bottomNavigation"
+            aria-label="Secondary"
+          >
+            <ul>
+              <slot name="bottomNavigation" />
+            </ul>
+          </nav>
+        </div>
+      </aside>
```
johncowen, line 151: I have a few comments on this, most of them are "ideally":

- I'd say we shouldn't use an `aside` here. `aside` is generally used for content "indirectly related to the documents content". The important semantic tag here is the `nav` i.e. navigational links, I don't think we should re-wrap that again in something that says "this is content indirectly related to the documents content". Whilst its tempting to see a "sidebar" and think "aside", that's kinda mixing visuals/decorative concerns with meaning/semantic concerns.
- The second nav shouldn't be the "Secondary Navigation", "ideally" it would be part of the "Main Navigation", its just one more item in the "Main Navigation" the fact that visually its in a different place is kinda mixing the decorative/semantic concerns similar to point 1.

I think for the above two, sometimes its helpful to try to ignore the designs completely, just look at the HTML document and attempt to describe what the content is (not what it looks like) using HTML tags.

To try and give some alternative options:

1. "ideally" there is no difference between Home, Zones, Meshes and Configuration semantically. They are all part of the main navigation, the only reason you've split them here is because visually they are in a different place. Therefore "ideally" all of these items would be in the same `<nav aria-label="Main"><ul>` as separate `li`s. Then I would guess you "might" be able to apply similar CSS to the `ul` and the Configuration `li` to move that to the bottom (this might be overly difficult, if so see next idea)
2. We could also try keeping all the items in the `<nav aria-label="Main">` but putting Configuration in a separate `ul` to the other items, and again trying to apply similar CSS. This is cheating a little, but not as much as having 2 nav's.

Hopefully one of the above works, if not we will have to think of a different option that doesn't have a "Secondary Navigation" that isn't has links that aren't children of the "Main Navigation". Maybe we just name it something else.

Anyway let me know what you think, the entire thing is a bit of a nit really, but if we are aiming to do semantics we should try and do the best we can. If it proves problematic, we can just say "ok we tried to be super semantic here, but realistically it was overly difficult so we cheated"

Gonna submit this comment while I keep reviewing.

packages/kuma-gui/src/app/App.vue
```diff
@@ -103,4 +119,12 @@ router.afterEach(() => {
...
   max-height: 36px;
 }
+
+:deep(.app-sidebar) {
+  --icon-home: url('@/assets/images/navigation/icon-home.svg');
+  --icon-zones: url('@/assets/images/navigation/icon-zones.svg');
+  --icon-meshes: url('@/assets/images/navigation/icon-meshes.svg');
+  --icon-configuration: url('@/assets/images/navigation/icon-configuration.svg');
+  --icon-zone-egresses: url('@/assets/images/navigation/icon-zone-egresses.svg');
+}
```
johncowen, line 129: Just to leave some sort of note on what we discussed yesterday:

Ideally whilst we have an `XIcon` component we should try and use it for all our icons, instead of "all our Iconography uses XIcon... apart from this little bit over there"

Personally, I would rather all our icongraphy used CSS, its DOM node friendly, more accessible and more flexible. But not something we are in a position to change right now tho.

---

That said, lets stick with what you have here for now, not all of these icons are available as kong/icons, but just to make a note that we may circle back and change it to use XIcon here at some point (or the other way around and make all our icons use CSS). i.e. how to do iconography should be consistent everywhere.

johncowen (commented):
Looking good, I think it looks much neater now we've lined the icons up with the Kuma bear's nose, thanks for sorting that!

If we can sort the semantic comment I made then great, if not then let me know and we can get this in anyway, I think its more than good enough as is also 👍

johncowen:
P.S. I also just noticed there is something functionality wrong with the Configuration page, I think there's a test failing that could be related to that but currently getting this when searching when using the PR preview:

Screenshot 2024-11-22 at 09 55 33

packages/kuma-gui/src/assets/images/navigation/icon-zone-egresses.svg
johncowen: 🤔 yeah I getcha, not sure either, its almost like we need this thing in reverse, so an arrow pointing outwards:

https://kong.github.io/icons/?q=portalicon

Screenshot 2024-11-22 at 10 00 44

I guess lets leave it with what you have for now, we can discuss later and switch it out before we release 2.10

packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
schogges: > I have a few comments on this, most of them are "ideally":
> 
> * I'd say we shouldn't use an `aside` here. `aside` is generally used for content "indirectly related to the documents content". The important semantic tag here is the `nav` i.e. navigational links, I don't think we should re-wrap that again in something that says "this is content indirectly related to the documents content". Whilst its tempting to see a "sidebar" and think "aside", that's kinda mixing visuals/decorative concerns with meaning/semantic concerns.

hmm 🤔 I think that `aside` is also frequently used for sidebars. If we keep the semantics wherever possible I'd say they are more stable when it comes to refactorings. The `aside` is a container for anything that is, as you said, indirectly related to the documents context. Which is the case for nav links, but could also be other things. So if we decide to move the nav links somewhere else, the sidebar may stay there.

> * The second nav shouldn't be the "Secondary Navigation", "ideally" it would be part of the "Main Navigation", its just one more item in the "Main Navigation" the fact that visually its in a different place is kinda mixing the decorative/semantic concerns similar to point 1.

Yes, I fully agree. This makes more sense 👍

packages/kuma-gui/src/app/App.vue
schogges: Yes, sounds good 👍 🙂

packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
schogges: > To try and give some alternative options:
> 
> 1. "ideally" there is no difference between Home, Zones, Meshes and Configuration semantically. They are all part of the main navigation, the only reason you've split them here is because visually they are in a different place. Therefore "ideally" all of these items would be in the same `<nav aria-label="Main"><ul>` as separate `li`s. Then I would guess you "might" be able to apply similar CSS to the `ul` and the Configuration `li` to move that to the bottom (this might be overly difficult, if so see next idea)
> 2. We could also try keeping all the items in the `<nav aria-label="Main">` but putting Configuration in a separate `ul` to the other items, and again trying to apply similar CSS. This is cheating a little, but not as much as having 2 nav's.

I went for option 2 🙂 👍

johncowen: > hmm 🤔 I think that aside is also frequently used for sidebars.

Would you say that if we have a different style of navigation, say a top horizontal nav like this:

Screenshot 2024-11-22 at 10 24 18

With `[Home] [Meshes] [Zones] [Configuration]` along the top horizontally,  then you wouldn't use an `aside` as its not at the side of the page?

johncowen (approved):
Lets go with this! 🎉

PR #3219, opened 2024-11-25
fix: Use XAction for docs links

johncowen:
Uses `XAction` as opposed to `KButton` for our docs links, which means all our docs links automatically open in a new tab


schogges (approved):
👍

PR #3220, opened 2024-11-25
fix(zones): correct text of Zone CP empty state

johncowen:
### Before

Screenshot 2024-11-25 at 15 50 40

### After

Screenshot 2024-11-25 at 15 50 05


schogges (approved):
👍

PR #3221, opened 2024-11-25
chore: Formalize 'route walkers' a little more

johncowen:
Back in https://github.com/kumahq/kuma-gui/pull/3151 we introduced the concept of 'route walkers'. These walk over every single route in your entire routing tree, giving you the opportunity to mutate or otherwise tweak a piece of routing config from The Outside.

When we made https://github.com/kumahq/kuma-gui/pull/3151, we noted we'd added the 'route walker' concept plus 2 walkers at our 'vue' level (i.e. in our `vue` module), even though the 2 walkers are specific to our `application` module.

This PR splits the concept and the application level implementations using a `$.routeWalkers` dependency injection label. 

This means we can dynamically add/remove different routeWalkers at the service container level, which will be very useful moving forwards.

---

I also added one more walker than automatically adds a 'props.routeName' property to all route components. You will still need to use `defineProps{ routeName: string}()` in order to access it, but at least you no longer need to manually add repetitive routing configuration to do this.

### DIFF
--- a/packages/kuma-gui/src/app/application/index.ts
+++ b/packages/kuma-gui/src/app/application/index.ts
@@ -103,22 +103,23 @@ const addPath = (item: RouteRecordRaw, parent?: RouteRecordRaw) => {
   }
 }
 const addRouteName = (item: RouteRecordRaw) => {
-  if (typeof item.name !== 'undefined') {
-    const props = ((props) => {
-      switch (true) {
-        case typeof props === 'function':
-          return props
-        case typeof props === 'undefined':
-          return () => ({})
-        default:
-          return () => props
-      }
-    })(item.props)
-    item.props = (...args) => {
-      return {
-        ...props(...args),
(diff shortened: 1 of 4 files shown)


packages/kuma-gui/src/app/application/index.ts
```diff
@@ -83,6 +84,44 @@ const $ = {
...
+      }
+    })(item.props)
+    item.props = (...args) => {
+      return {
+        ...props(...args),
+        routeName: item.name,
+      }
+    }
+  }
+}
```
schogges, line 124: Just a nit 🙂 would turn around the `if`-condition and exit early

johncowen: Oh! Sorry completely missed this, this is done now, lmk!

schogges (approved):
👍

PR #3222, opened 2024-11-25
feat: default service hosts to period-delimited

### LINKED ISSUE #3098: Dot instead of underscores for Internal Services hostnames
### Description

The problem is if you want to use feature of DNS hostname with dots instead of `_` you need to replace it by hand. There should be a way for user to copy from the GUI hostnames with dots.

This is only for internal service tag services, not for mesh services

xref https://github.com/kumahq/kuma/issues/11869

johnharris85:
DNS is created in the mesh for `_`-delimited and `.`-delimited hostnames. This PR makes the `.`-delimited the default (as it is the only one in conformance with the RFC). This is a QOL improvement for folks copying/pasting from the UI, and both versions should work identically. The API returns `_`-delimited so in the case anyone was using any kind of automation, this shouldn't break them anyway.

fixes #3098


johncowen (approved):
🎉 Thanks!

johncowen:
I gonna go ahead and merge this, thanks for the contribution! 🎉

PR #3232, opened 2024-11-26
feature: enable delayed teleporting

johncowen:
Sometimes when you add a teleport to a page, the place where it is supposed to go hasn't quite been rendered yet (the slot) which is just down to Vue/DOM intricacies.

Previously we would just error out if we hit this, but there are some use cases I've found where this is always going to throw. Ideally, it should wait for the corresponding template slot to be available and only then teleport into it, and that is what this PR does.

Just to note: timings wise we are talking milliseconds here.

### DIFF
--- a/packages/kuma-gui/src/app/x/components/x-teleport/XTeleportTemplate.vue
+++ b/packages/kuma-gui/src/app/x/components/x-teleport/XTeleportTemplate.vue
@@ -22,8 +22,8 @@ onMounted(() => {
   if (document.querySelector(`[data-x-teleport-id='${props.to.name}']`) !== null) {
     ready.value = true
   } else {
-    window.addEventListener('x-teleport-slot:mounted', (e: Event) => {
-      if (isCustomEvent(e) && e.detail?.name === props.to.name) {
+    window.addEventListener('x-teleport-slot:mounted', (e: CustomEvent<{ name: string }> | Event) => {
+      if ('detail' in e && e.detail?.name === props.to.name) {
         ready.value = true
       }
     }, { signal: controller.signal })
--- a/packages/kuma-gui/src/app/x/components/x-teleport/XTeleportTemplate.vue
+++ b/packages/kuma-gui/src/app/x/components/x-teleport/XTeleportTemplate.vue
@@ -7,19 +7,30 @@
   </Teleport>
 </template>
 <script lang="ts" setup>
-import { onMounted, ref } from 'vue'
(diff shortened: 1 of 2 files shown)


packages/kuma-gui/src/app/x/components/x-teleport/XTeleportTemplate.vue
```diff
@@ -7,19 +7,30 @@
...
 onMounted(() => {
   if (document.querySelector(`[data-x-teleport-id='${props.to.name}']`) !== null) {
     ready.value = true
   } else {
-    throw new Error(`The '[data-x-teleport-id='${props.to.name}']' element could not be found to teleport to`)
+    window.addEventListener('x-teleport-slot:mounted', (e: Event) => {
+      if (isCustomEvent(e) && e.detail?.name === props.to.name) {
+        ready.value = true
+      }
+    }, { signal: controller.signal })
```
schogges, line 29: Really just a nit and I'll approve anyways, but I think we can save some lines and the type guard here:

```suggestion
    window.addEventListener('x-teleport-slot:mounted', (e: CustomEvent<{ name: string }> | Event) => {
      if ('detail' in e && e.detail?.name === props.to.name) {
        ready.value = true
      }
    }, { signal: controller.signal })
```

schogges (approved):

packages/kuma-gui/src/app/x/components/x-teleport/XTeleportTemplate.vue
```diff
@@ -7,19 +7,30 @@
...
 const props = defineProps<{
   to: {
     name: string
   }
 }>()
-
+const controller = new AbortController()
 const ready = ref<boolean>(false)
+
+const isCustomEvent = (e: Event | CustomEvent): e is CustomEvent => 'detail' in e
```
schogges, line 19: Sorry, could not add code suggestion over untouched lines 😓 

```suggestion
```

packages/kuma-gui/src/app/x/components/x-teleport/XTeleportTemplate.vue
```diff
@@ -7,19 +7,30 @@
...
+
```
schogges, line 20: Sorry, could not add code suggestion over untouched lines 😓 

```suggestion
```

johncowen: Oh 🤦 , one sec

Issue #2879, opened 2024-09-09
upgrade eslint to v9


johncowen:
### Description

See title

PR #3235, opened 2024-11-27
chore: lint package.json versions to only allow caret semvers

johncowen:
Honestly I'm not a huge fan of linting, a lot of rules that people use are either unnecessary or over-zealous. Not only that but it always means depending on lots of things like `eslint`, which you then have to maintain (dependabot PRs etc etc), just for the sake of not caring too much about things like having alphabetically ordered import statements.

Now I've got that off my chest 😅 , I do agree that there are sometimes use cases for specific rules to prevent mistakes that can lead to genuine bugs.

This use case:

Over the past couple of years we've had several issues with duplicate dependencies appearing in our package lock file. They are easy to miss in PR reviews, and a lot of the time people don't realise that hainvg duplicate dependencies can cause very hard to spot issues (in our case having a duplicate install of kongponents has led to CSS issues that are not caught by CI as we don't have CSS regression tests).

The reason we've ended up with a duplicate install of a dependency is because somebody somewhere perhaps mistakenly has used an exact version to specify a package version in their consumable package. This goes for _at least_ `dependencies` and `peerDependencies`, which effect consuming applications.

I think there is no argument/discussion about the need to use range constraints rather than exact constraints in consumable packages (as oppose to consuming applications). In fact we should really go the extra mile even if there is a problem and the temptation to fix a version exactly in a consumable package. (Just so its said, IMO, we should also only use range constraints for consuming applications unless its absolutely necessary - but we are not here to talk about that specific use case today.)

This PR:

- Installs `"eslint-plugin-json-schema-validator"` so we can validate JSON files (I know I know another dependency...)
- Uses a minimal custom json schema to validate our package.json file, only allowing caret versions i.e. `^1.0.0`
- Adds an exception for `@kuma/config` to only allow `*`
- Comments in examples of ignoring our caret rule so we have an escape-hatch/ignore for our rule.
- Includes the rule in our `make lint/js` **for '@kumahq/kuma-gui' only**. We don't yet apply all out linting to our entire workspace, but we will soon.

None of this is an issue in our repository as yet, (we always use `^` versions because as a team we've agreed to so we don't need a linter), but we have flip-flopped in the past. Also in our case we have comparatively little to manage and a small enough team to get away with no hard-defined lint rules, others might not be so lucky.

Kinda unrelated but interesting:

- Also adds a rule to disallow `engines`. Eventually I would like to use this to have a single source of truth for our `engines` values in our workspace `package.json`, whilst we can't include files from `@kumahq/config` into our root package.json file, at least we can lint it to make sure it has specific values using a single source of truth linting rule.

Lastly:

I started with this https://github.com/tclindner/npm-package-json-lint, but decided against it seeing as its a standalone tool (we already need/require eslint so best if we can continue to use that rather than add another tool). I also found it to be fairly rigid, i.e. whilst there is a rule to enforce carets on `devDependencies` and `dependencies` there strangely isn't one for `peerDependencies` and it looks like they don't support adding custom rules. I then thought about looking for a more generic solution where I could define my own rules - a generic JSON schema linter fits that problem, plus you never know as its generic we might find other uses for it.

### DIFF
--- a/packages/config/src/eslint.cjs
+++ b/packages/config/src/eslint.cjs
@@ -1,3 +1,6 @@
+const { dirname } = require('path')
+const $config = dirname(require.resolve('@kumahq/config'))
+
 // Taken from https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/utils/inline-non-void-elements.json.
 const INLINE_NON_VOID_ELEMENTS = [
   'a',
@@ -63,7 +66,15 @@ function createEslintConfig(
       'src/types/auto-generated.d.ts',
     ],
     plugins: ['vue', 'import', '@typescript-eslint'],
-    extends: ['eslint:recommended', 'plugin:vue/vue3-recommended', 'standard', '@vue/typescript', 'plugin:import/recommended', 'plugin:import/typescript'],
+    extends: [
+      'eslint:recommended',
+      'plugin:vue/vue3-recommended',
+      'standard',
+      '@vue/typescript',
+      'plugin:import/recommended',
+      'plugin:import/typescript',
+      'plugin:json-schema-validator/recommended'
+    ],
     settings: {
       'import/resolver': {
         typescript: {
@@ -94,6 +105,17 @@ function createEslintConfig(
       'no-console': ['error', { allow: ['info', 'warn', 'error'] }],
       'padded-blocks': 'off',
       'no-unreachable': 'error',
+      'json-schema-validator/no-invalid': ['error', {
+        useSchemastoreCatalog: false,
+        schemas: [
+          {
+            fileMatch: ['package.json'],
+            // our schema allows for ignoring individual dependencies if required
+            // see ./package.schema.json patternProperties examples
+            schema: `${$config}/package.schema.json`
+          }
+        ]
+      }]
     },
     overrides: [
       {
--- a/packages/config/src/eslint.cjs
+++ b/packages/config/src/eslint.cjs
@@ -1,5 +1,8 @@
 const { dirname } = require('path')
+const { readFileSync: read } = require('fs')
 const $config = dirname(require.resolve('@kumahq/config'))
+
+const packageSchema = JSON.parse(read(`${$config}/package.schema.json`).toString())
 
 // Taken from https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/utils/inline-non-void-elements.json.
 const INLINE_NON_VOID_ELEMENTS = [
@@ -44,12 +47,30 @@ const INLINE_NON_VOID_ELEMENTS = [
 ]
 
 /**
- * @param {{ tsConfigPath?: string, componentIgnorePatterns?: string[] }} [options]
+ * @param {{ tsConfigPath?: string, componentIgnorePatterns?: string[], versionIgnorePatterns: Record<string, unknown> }} [options]
  * @returns {import('eslint').ESLint.ConfigData}
  */
 function createEslintConfig(
-  { tsConfigPath = 'tsconfig.json', componentIgnorePatterns = [] } = { tsConfigPath: 'tsconfig.json', componentIgnorePatterns: [] }
+  {
+    tsConfigPath = 'tsconfig.json',
+    componentIgnorePatterns = [],
+    versionIgnorePatterns = {}
+  } = {
+    tsConfigPath: 'tsconfig.json',
+    componentIgnorePatterns: [],
+    versionIgnorePatterns: {}
+  }
 ) {
+
+  ((properties) => {
+    ['dependencies', 'devDependencies', 'peerDependencies'].forEach((item) => {
+      properties[item].patternProperties = {
+        ...properties[item].patternProperties,
+        ...versionIgnorePatterns[item] ?? {}
+      }
+    })
+  })(packageSchema.properties)
+
   return {
     root: true,
     parserOptions: {
@@ -107,12 +128,13 @@ function createEslintConfig(
       'no-unreachable': 'error',
       'json-schema-validator/no-invalid': ['error', {
         useSchemastoreCatalog: false,
+        mergeSchemas: true,
         schemas: [
           {
             fileMatch: ['package.json'],
             // our schema allows for ignoring individual dependencies if required
             // see ./package.schema.json patternProperties examples
-            schema: `${$config}/package.schema.json`
+            schema: packageSchema
           }
         ]
       }]
--- a/packages/config/src/package.schema.json
+++ b/packages/config/src/package.schema.json
@@ -7,11 +7,11 @@
     },
     "workspaceOnlyVersion": {
       "type": "string",
-      "pattern": "(^\\*$)|(^file:)"
+      "pattern": "(^\\*$)|(^file:)|(^workspace:)"
     },
     "caretOrWorkspaceOnlyVersion": {
       "type": "string",
-      "pattern": "(^\\^)|(^\\*$)|(^file:)"
+      "pattern": "(^\\^)|(^\\*$)|(^file:)|(^workspace:)"
     },
     "rangeOnlyVersion": {
       "type": "string",
--- a/packages/config/src/package.schema.json
+++ b/packages/config/src/package.schema.json
@@ -5,9 +5,17 @@
       "type": "string",
       "pattern": "^\\^"
     },
-    "workspaceVersion": {
+    "workspaceOnlyVersion": {
       "type": "string",
       "pattern": "(^\\*$)|(^file:)"
+    },
+    "caretOrWorkspaceOnlyVersion": {
+      "type": "string",
+      "pattern": "(^\\^)|(^\\*$)|(^file:)"
+    },
+    "rangeOnlyVersion": {
+      "type": "string",
+      "pattern": "^(?!\\d+\\.\\d+\\.\\d+).*"
     }
   },
   "properties": {
@@ -23,7 +31,7 @@
         }
       },
       "additionalProperties": {
-        "$ref": "#/definitions/caretOnlyVersion"
+        "$ref": "#/definitions/caretOrWorkspaceOnlyVersion"
       }
     },
     "devDependencies": {
@@ -32,13 +40,10 @@
       "patternProperties": {
         "@kumahq/ignored-example": {
           "type": "string"
-        },
-        "@kumahq/config": {
-          "$ref": "#/definitions/workspaceVersion"
         }
       },
       "additionalProperties": {
-        "$ref": "#/definitions/caretOnlyVersion"
+        "$ref": "#/definitions/caretOrWorkspaceOnlyVersion"
       }
     },
     "peerDependencies": {
@@ -50,7 +55,7 @@
         }
       },
       "additionalProperties": {
-        "$ref": "#/definitions/caretOnlyVersion"
+        "$ref": "#/definitions/rangeOnlyVersion"
       }
     }
   }
--- a/packages/config/src/package.schema.json
+++ b/packages/config/src/package.schema.json
@@ -7,7 +7,7 @@
     },
     "workspaceVersion": {
       "type": "string",
-      "pattern": "^\\*$"
+      "pattern": "(^\\*$)|(^file:)"
     }
   },
   "properties": {
--- a/packages/config/src/package.schema.json
+++ b/packages/config/src/package.schema.json
@@ -1,0 +1,57 @@
+{
+  "type": "object",
+  "definitions": {
+    "caretOnlyVersion": {
+      "type": "string",
+      "pattern": "^\\^"
+    },
+    "workspaceVersion": {
+      "type": "string",
+      "pattern": "^\\*$"
+    }
+  },
+  "properties": {
+    "engines": {
+      "not": {}
+    },
+    "dependencies": {
+      "type": "object",
+      "$comment": "To ignore a dependency to use a non-caret version follow the ignored-example below",
+      "patternProperties": {
+        "@kumahq/ignored-example": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": {
+        "$ref": "#/definitions/caretOnlyVersion"
+      }
+    },
+    "devDependencies": {
+      "type": "object",
+      "$comment": "To ignore a dependency to use a non-caret version follow the ignored-example below",
+      "patternProperties": {
+        "@kumahq/ignored-example": {
+          "type": "string"
+        },
+        "@kumahq/config": {
+          "$ref": "#/definitions/workspaceVersion"
+        }
+      },
+      "additionalProperties": {
+        "$ref": "#/definitions/caretOnlyVersion"
+      }
+    },
+    "peerDependencies": {
+      "type": "object",
+      "$comment": "To ignore a dependency to use a non-caret version follow the ignored-example below",
+      "patternProperties": {
+        "@kumahq/ignored-example": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": {
+        "$ref": "#/definitions/caretOnlyVersion"
+      }
+    }
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -2598,6 +2598,17 @@
         "node": ">=14"
       }
     },
+    "node_modules/@pkgr/core": {
+      "version": "0.1.1",
+      "resolved": "https://registry.npmjs.org/@pkgr/core/-/core-0.1.1.tgz",
+      "integrity": "sha512-cq8o4cWH0ibXh9VGi5P20Tu9XF/0fFXl9EUinr9QfTM7a7p0oTA4iJRCQWppXR1Pg8dSM0UCItCkPwsk9qWWYA==",
+      "engines": {
+        "node": "^12.20.0 || ^14.18.0 || >=16.0.0"
+      },
+      "funding": {
+        "url": "https://opencollective.com/unts"
+      }
+    },
     "node_modules/@popperjs/core": {
       "version": "2.11.8",
       "resolved": "https://registry.npmjs.org/@popperjs/core/-/core-2.11.8.tgz",
@@ -6627,7 +6638,6 @@
       "version": "0.5.1",
       "resolved": "https://registry.npmjs.org/eslint-compat-utils/-/eslint-compat-utils-0.5.1.tgz",
       "integrity": "sha512-3z3vFexKIEnjHE3zCMRo6fn/e44U7T1khUjg+Hp0ZQMCigh28rALD0nPFBcGZuiLC5rLZa2ubQHDRln09JfU2Q==",
-      "peer": true,
       "dependencies": {
         "semver": "^7.5.4"
       },
@@ -6642,7 +6652,6 @@
       "version": "7.6.3",
       "resolved": "https://registry.npmjs.org/semver/-/semver-7.6.3.tgz",
       "integrity": "sha512-oVekP1cKtI+CTDvHWYFUcMtsK/00wmAEfyqKfNdARm8u1wNVhSgaX7A8d4UuIlUI5e84iEwOhs7ZPYRmzU9U6A==",
-      "peer": true,
       "bin": {
         "semver": "bin/semver.js"
       },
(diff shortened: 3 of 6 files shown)


packages/config/src/eslint.cjs
```diff
@@ -63,7 +66,15 @@ function createEslintConfig(
...
-    extends: ['eslint:recommended', 'plugin:vue/vue3-recommended', 'standard', '@vue/typescript', 'plugin:import/recommended', 'plugin:import/typescript'],
+    extends: [
+      'eslint:recommended',
+      'plugin:vue/vue3-recommended',
+      'standard',
+      '@vue/typescript',
+      'plugin:import/recommended',
+      'plugin:import/typescript',
+      'plugin:json-schema-validator/recommended'
+    ],
```
adamdehaven, line 77: We have a shared ESLint v9+ config for the frontend ecosystem if you'd like to adopt it: https://github.com/Kong/eslint-config-kong-ui

johncowen: At the moment I think that only supports eslint 9 (correct me if I'm wrong!) we are still on 8. We have an issue to upgrade it tho - https://github.com/kumahq/kuma-gui/issues/2879

johncowen:
>> just for the sake of not caring too much about things like having alphabetically ordered import statements.

> This is actually super important to avoid conflicts when folks modify the same files.
> Linting in generally helps a lot with auto formatting and killing useless convs like "Space vs Tab"

Oh don't get me wrong, (and this is the balance I was trying to strike in the description). Linting is good for certain usecases, like to prevent actual mistakes, and as you pointed out to enforce agreed upon important formatting for folks who submit PRs.

Not to get too into it, but when it comes to things like alphabetically order import statements I'd rather just say "don't mess with the order if you don't need to, but if you need to thats fine also"

e.g. we have things like this happening because of a lint rule:

https://github.com/kumahq/kuma-gui/blob/765d9f5f2ca3fdc44b5998f946064ed9aa725c7b/packages/kuma-gui/src/test-support/mocks/fs.ts#L1-L20

I'd rather just say, "don't mess with the order if you don't need to, but if you need to thats fine also" I really don't mind if someone found it easier to work with the code by changing the order of an import statement. I'd rather they found it easier.

Anyway, sorry I didn't want this to turn into a discussion on the usefulness of linters, this PR is a good example of usefulness in itself 👍

johncowen:
Wanted to underline, note that this was more of an example to solve an external problem, we might not end up merging this.

> None of this is an issue in our repository as yet, (we always use ^ versions because as a team we've agreed to so we don't need a linter), but we have flip-flopped in the past. Also in our case we have comparatively little to manage and a small enough team to get away with no hard-defined lint rules, others might not be so lucky.

That said, I'm more inclined to leave it around due to:

> Also adds a rule to disallow engines. Eventually I would like to use this to have a single source of truth for our engines values in our workspace package.json, whilst we can't include files from @kumahq/config into our root package.json file, at least we can lint it to make sure it has specific values using a single source of truth linting rule.

>  I then thought about looking for a more generic solution where I could define my own rules - a generic JSON schema linter fits that problem, plus you never know as its generic we might find other uses for it.

(quotes are from the PR description)

johncowen:
Just to note as I'd mentioned offline that I'm indifferent on merging this seeing as I think we are pretty good with our dependencies and we probably don't need a linter to check anything for us. I'm inclined to merge it though unless there are any major concerns, if only due to the other things we might want to use it for (e.g. `engines` plus other)

schogges (commented):
I agree that we should configure our linters based on our needs and adopt the common kong config.
It's surely not helpful to have all rules applied and different projects have different requirements, but we should share the same base 👍 

Regarding this new rule, I have some concerns:
1. The schema only recognizes `^`, but no other range specifications like `~`
2. Only the `config` package is considered as a workspace, which means this rule has to be changed whenever we add/remove a workspace package
3. If we have to pin a version for whatever reason for some time, there is no way to add a eslint disable directive comment into the `package.json` (not sure, but I doubt that there is another way to ignore a specific line in a file via eslint config)

In case we want to merge this I would probably turn it around and check the versions for not being pinned. Pinned version is major.minor.patch `{number}.{number}.{number}` (we should probably just check for the first part to allow pre-release versions). If there is any char before that pattern in the version, it's not pinned.

---

> I started with this https://github.com/tclindner/npm-package-json-lint, but decided against it seeing as its a standalone tool (we already need/require eslint so best if we can continue to use that rather than add another tool). I also found it to be fairly rigid, i.e. whilst there is a rule to enforce carets on devDependencies and dependencies there strangely isn't one for peerDependencies and it looks like they don't support adding custom rules. I then thought about looking for a more generic solution where I could define my own rules - a generic JSON schema linter fits that problem, plus you never know as its generic we might find other uses for it.

The thing with peers is that they can also be specified like `>= 1.5.2 < 2`. So also here it would be beneficial to just check for the versions **not** being pinned 🤔

johncowen:
Hey, thanks for documenting those concerns, I agree some of them if I understand correctly I think, but not totally sure. 

---

This one is the easiest to check my understanding I think:

> - If we have to pin a version for whatever reason for some time, there is no way to add a eslint disable directive comment into the package.json (not sure, but I doubt that there is another way to ignore a specific line in a file via eslint config)

Yeah for sure, typically JSON won't let us add comments, but did you notice ignoring is done by adding an different constraint for the package?:

https://github.com/kumahq/kuma-gui/pull/3235/files#diff-89be3878e6a412c779ce56d885876b812028f26510f306c4e6514eb883777e4eR36

I have a better example of this elsewhere, but let me grab a link for that

---

> Only the config package is considered as a workspace, which means this rule has to be changed whenever we add/remove a workspace package

Not sure if I understood this one, I read/understood this as "whenever we add a new package thats in our workspace, we need to update the `@kumahq/config` package constraint" ? Whereas in reality it will be more "whenever we add a new package thats in our workspace we need to add a new ignore to the schema" (see above comment on how to add ignores). Let me know on this one.

Semi-related to this point, I think (99% sure) I want to make a change to the code here to make the default constraint "caretOrWorkspace", which I think is the constraint we commonly want i.e. "as long as its tracking minors+patches, or its one of our own, then you are all good"


---

> The schema only recognizes ^, but no other range specifications like ~

I think this is the one where I agree less. I'd suggest not being able to add `~` versions is a feature. We are making it easy to do the right thing and making it hard to do the wrong thing. Mostly consumable dependencies should always use `^`, unless there is a good reason. If there is a good reason it should be noted, and the best way to do that is via a PR (see JSON reason in a sec), and its more likely to be noticed and then noted if a ignore/exception is added in that PR. i.e. the linter is forcing us to make a documented note of why we are pinning a dependency differently. I think the even better thing about this is because JSON again won't let us add this reasoning in the code itself, so the git blame of that line of code would lead is to the PR and the reason.

---

> The thing with peers is that they can also be specified like >= 1.5.2 < 2.

Completely see what you are saying here, and agree with the use case. I suppose the answer would be, we write a `definition/somethingSomethingVersion` that allows this and we add this exception to the corresponding pacakge. The important thing to point out here is that for our project (and others I guess) these types of constraints are not the norm (although perhaps should be, but lets not go there just yet). Actually, come to think of it, we could add a different constraint just to `peerDependencies` that makes wider ranges like this the norm but just for peers. I think first we would need to decide what the norm is i.e. what do we want to make it easy to do and hard to get wrong.

schogges:
> Hey, thanks for documenting those concerns, I agree some of them if I understand correctly I think, but not totally sure.
> 
> This one is the easiest to check my understanding I think:
> 
> > * If we have to pin a version for whatever reason for some time, there is no way to add a eslint disable directive comment into the package.json (not sure, but I doubt that there is another way to ignore a specific line in a file via eslint config)
> 
> Yeah for sure, typically JSON won't let us add comments, but did you notice ignoring is done by adding an different constraint for the package?:
> 
> https://github.com/kumahq/kuma-gui/pull/3235/files#diff-89be3878e6a412c779ce56d885876b812028f26510f306c4e6514eb883777e4eR36
> 
> I have a better example of this elsewhere, but let me grab a link for that

Ah right, gotcha 🙂 

> > Only the config package is considered as a workspace, which means this rule has to be changed whenever we add/remove a workspace package
> 
> Not sure if I understood this one, I read/understood this as "whenever we add a new package thats in our workspace, we need to update the `@kumahq/config` package constraint" ? Whereas in reality it will be more "whenever we add a new package thats in our workspace we need to add a new ignore to the schema" (see above comment on how to add ignores). Let me know on this one.
> 
> Semi-related to this point, I think (99% sure) I want to make a change to the code here to make the default constraint "caretOrWorkspace", which I think is the constraint we commonly want i.e. "as long as its tracking minors+patches, or its one of our own, then you are all good"

Sorry, I wasn't clear enough. But I think we are on the same page 👍 so, I think it would be better if we wouldn't need to touch the schema whenever we add/remove a package to/from our workspace.

> > The schema only recognizes ^, but no other range specifications like ~
> 
> I think this is the one where I agree less. I'd suggest not being able to add `~` versions is a feature. We are making it easy to do the right thing and making it hard to do the wrong thing. Mostly consumable packages like our own should always use `^`, unless there is a good reason. If there is a good reason it should be noted, and the best way to do that is via a PR (see JSON reason in a sec), and its more likely to be noticed and then noted if a ignore/exception is added in that PR. i.e. the linter is forcing us to make a documented note of why we are pinning a dependency differently. I think the even better thing about this is because JSON again won't let us add this reasoning in the code itself due to lack of comments, so the git blame of that line of code would lead us to the PR and the reason.

Yeah, I think it's very very rare to use `~` and mostly use `^`. So, if we want to achieve that all versions use `^` it's fine. But I thought we want to achieve that pinned versions are getting catched?

> > The thing with peers is that they can also be specified like >= 1.5.2 < 2.
> 
> Completely see what you are saying here, and agree with the use case. I suppose the answer would be, we write a `definition/somethingSomethingVersion` that allows this and we add this exception to the corresponding package. The important thing to point out here is that for our project (and others I guess) these types of constraints are not the norm (although perhaps should be, but lets not go there just yet). Actually, come to think of it, we could add a different constraint just to `peerDependencies` that makes wider ranges like this the norm but just for peers. I think first we would need to decide what the norm is i.e. what do we want to make it easy to do and hard to get wrong.

I think this is getting very complex. I guess it would be much easier if we focus on what was causing issues and just catch pinned versions 🤔

schogges:
I just realized that it's possible to define a version in `dependencies` and `devDependencies` (not only in `peerDependencies`) via `>= x.x.x` or similar. This means it would also be possible to pin a version like this: `>= 1.17.1 < 1.17.2`.

johncowen:
> I just realized that it's possible to define a version in dependencies and devDependencies (not only in peerDependencies) via >= x.x.x or similar. This means it would also be possible to pin a version like this: >= 1.17.1 < 1.17.2.
This means that if we want to bring this in, it's probably better to follow what you've started and only allow caret ranges, workspaces, ... 🤔

If I understand you correctly, yes I completely agree. If this PR is something we want to add (linting for package versions) then we should say "the default is: for devDeps and deps is caretOrWorkspace, peerDeps any non-exact version, if you need to do anything else then add an ignore". Then the fact that an ignore is added would probably be questioned/commented in the PR and we leave a trail of reason.

adamdehaven:
> > I just realized that it's possible to define a version in dependencies and devDependencies (not only in peerDependencies) via >= x.x.x or similar. This means it would also be possible to pin a version like this: >= 1.17.1 < 1.17.2.

> If I understand you correctly, yes I completely agree. 

The way this PR is headed is a very non-standard way of managing dependencies.

johncowen:
Just incase you misunderstood @adamdehaven 

> This means it would also be possible to pin a version like this: >= 1.17.1 < 1.17.2.
> If I understand you correctly, yes I completely agree.

When I said I completely agree, I meant that I agree that _without this PR_ it would be possible to pin a version like `>= 1.17.1 < 1.17.2`.

With this PR you absolutely have to use either a caret or a workspace for devDeps or deps.

Is there a problem with enforcing that do you think?

adamdehaven:
> When I said I completely agree, I meant that I agree that _without this PR_ it would be possible to pin a version like `>= 1.17.1 < 1.17.2`.

This is my point: there’s nothing wrong with a version range as declared here. It’s valid. 

Im not sure if there’s a misunderstanding of version management or if you just have very opinionated view on the subject, but I’d recommend not going the path this PR is headed. 

I’m going to take myself out of any further discussion; I’m just not sure I understand the way this is going

schogges:
I have mixed feelings about this 🙂 

First of all, I have no big issues with pinning a version in an application (downstream). And as far as I know we had no issues with a pinned version of a direct dependency in this project yet. We are adding more stuff (including another dependency) that attempts to solve a problem we haven't had yet. So, right now I see no real needs in adding this.

On the other hand, some linting that helps us to enforce the use of version ranges for our direct dependencies and leave the rest to the lock-file, can be beneficial.

johncowen:
Ok thanks both for your input!

I'm going to split replies into two comments so its clearer what I'm addressing.

Firstly to @adamdehaven 's points:

>> When I said I completely agree, I meant that I agree that without this PR it would be possible to pin a version like >= 1.17.1 < 1.17.2.

> This is my point: there’s nothing wrong with a version range as declared here. It’s valid.

Completely agree that it is a valid version constraint, otherwise it just wouldn't work at all, and we wouldn't have an issue because our tests would fail and CI wouldn't let us merge a PR if the value used for the version wasn't valid.

The thing is, using `>= 1.17.1 < 1.17.2` instead of just using `1.17.1` (which is what I assume @schogges was pointing out) and saying its ok to use `>= 1.17.1 < 1.17.2` instead of just `1.17.1` I would say is very similar to saying its ok to use Tabs in a Spaces project.

Tabs are perfectly valid, but would you submit/merge a PR using Tabs in a project that was clearly using Spaces, I don't think so. You'd probably ask the person who did that to change it:

> Linting in generally helps a lot with auto formatting and killing useless convs like "Space vs Tab"

_It is my opinion_ that `>= 1.17.1 < 1.17.2` is harder to grok at a glance also, even in this conversation I looked at that and thought "Wait!? Wat!? - oh its pretty much an exact version range i.e. `1.17.1`". This PR I've done prevents someone even trying to do that. (Worth saying I don't think anyone would try)

The above doesn't even get into the fact that we don't want people using exact version constraints unless there is a good reason, which I'll touch on next:

> Im not sure if there’s a misunderstanding of version management or if you just have very opinionated view on the subject, but I’d recommend not going the path this PR is headed.

Of the 4 people discussing this here, I believe at least 3 of us hold this opinion:

"Dependencies of libraries should not be exact pinned"

At least 3 of us here believe that in regards to version management, dependencies of libraries (i.e. consumed dependencies) should not be exact pinned. I am one of these 3 people and if my view is an opinionated view, then it seems to be shared by at least 2 others.

From what I understand, most folks in the JS ecosystem also think that consumed dependencies shouldn't use exact versioning, ie. "its a bad idea".

There is of course the question of whether we are a "consumed dependency". I think we'd all say that all the sub-packages in our workspace are "consumed dependencies".

Lastly, I'd really like to understand:

>  I’d recommend not going the path this PR is headed.

If you could point out to me what I'm missing here, I'd gladly close this PR down. If you look back at what I've already said in this PR I've explained that I'm not a great fan of linting _unless its providing actual technical benefit_, if there is no technical benefit then it can just lead to endless discussions over who's "style" is "best", i.e. which rules should we adopt.

The more I talk with you folks about this the more I think that it does have actual technical benefit, it's not just opinion nor code-style preference etc. and is could actually prevent bugs and help us document decisions made in our project as noted by @schogges (completely get that it comes with mixed feelings)

> On the other hand, some linting that helps us to enforce the use of version ranges for our direct dependencies and leave the rest to the lock-file, can be beneficial.

I'll talk about that more in my next comment. But, if you could explain why you wouldn't recommend going down the path that this PR is headed, I would really like to hear it as it might change my mind. Maybe it is just that I'm not being clear as to what we are trying to achieve here? But also appreciate that you've taken yourself out of this discussion, so thats fine also. 

Happy to talk offline also, you know where to find me 👍

johncowen:
> I have mixed feelings about this 🙂

Same here! 😆 I'm actually now really glad I wrote the PR description as I did as hopefully you can see my uneasy-ness over adding _another_ linting rule.

Honestly though, the more we've discussed this the more I feel there is benefit here. Despite this, I'm still more than happy to close this PR down if people think there is no benefit here at all.

> First of all, I have no big issues with pinning a version in an application (downstream). And as far as I know we had no issues with a pinned version of a direct dependency in this project yet. We are adding more stuff (including another dependency) that attempts to solve a problem we haven't had yet. So, right now I see no real needs in adding this.

Yes definitely, in some sense I can agree with this entirely. We ourselves **in this repository have had no issues with an exact pinned direct dependency** _as yet_ (actually 🤔  I'm not even sure thats true, I'd have to go back to when we had some pinned dependencies tho). I also have no doubt that neither I nor yourself would have to be nudged to ensure we use a range version over an exact version. And also "I know I know another dependency!" 😆 These facts make me think this PR isn't worth merging.

But! (there's always a but! 😆 ). What about external contributors? What about future contributors? Is this an actual usecase, - letting maintainers of a project define what version range folks should use? 

@adamdehaven reminded me of this:

> Dependencies saved to package.json will be configured with an exact version rather than using npm's default semver range operator.

https://docs.npmjs.com/cli/v8/using-npm/config#save-exact

"rather than using npm's **default semver range** operator" i.e. npm wants you to use a range by default, and it will use a caret range by default if you don't specify a range when using `npm install`.

Problem solved right? Why do we need a lint rule? Just always use `npm install` and lets not set `save-exact` in our npmrc file.

Because we all know people quite often just edit package.json by hand. So `npm`s intention of defaulting to always using a caret range for a version isn't a strong enough constraint. This PR adds that constraint (with ignorability). 

This also leads me to the next and perhaps more important benefit, I have no idea _why_ someone exact pinned a dependency.

---

I've split this in two as there is another reason I've realised that I like this during our discussion in this very PR.

Lets forget about bugs and whats right and whats wrong technically. Lets talk about documenting our decisions, architectural decision records, in this case small ones, but still important ones.

If I see an exact pinned dependency in a project I maintain, I really want to know why its exact pinned. At what point can I upgrade it? Why was it pinned in the first place? Thing is package.json files are JSON, so we all know we can't add comments! Damn JSON! 😆 

But if we are forced to add an ignore statement, its likely to be noticed and commented upon in a PR and probably easy-ish to find via git blame (as I suggested above). The further thing I realised after suggesting that is, not only do you have to add an ignore statement to exact pin a dependency, but you have to add the ignore statement _in a comment-able Javascript file_. Bingo! I don't care about your comment-less format JSON! 😅 

There's my mini-ADR for why it was pinned, future us will thank ourselves.

Same thing stands, happy to shut this down if you feel there is no benefit at all, also happy to chat offline.

But whichever I think we should make a decision on this sooner rather than later, so we can move forwards with it one way or another.

P.S. I didn't even talk about the other benefits here like using the json schema to make sure our `engines` property is in line with our `.nvm` file, and probably a couple of other things.

johncowen:
Sorry for yet another comment here, I wouldn't do this if I didn't feel it was important.

> We ourselves in this repository have had no issues with an exact pinned direct dependency as yet (actually 🤔 I'm not even sure thats true, I'd have to go back to when we had some pinned dependencies tho).

It wasn't too long ago that we were exact pinning dependencies here even though we are a consumable package, and I'm pretty sure that did cause issues. I don't want to go back there.

Take this perhaps this hypothetical situation:

- Somebody thinks "oh this is kuma-gui application, therefore its not a consumable package" (quite easy to assume if you are an external contributor), or somebody either knows or notices we _are_ a consumable package but is of the opinion that its ok to use exact versioning. Whichever, they are going to exact pin a dependency here.
- They want to exact pin `@kong/kongponents` and pin that to say "9.9.9"
- Upstream we have another dependency that range pins to `@kong/kongponents` to `^9.10.0`
- We suddenly end up with a duplicate dependency upstream and unnoticed CSS problems that block our users from getting their job done.

This is why its important to always range pin inside consumable packages, and it can cause a technical issue if you don't.

I'm pretty sure the above situation was happening in the past.

This is why in this case I think a lint rule is a good idea for this because it has technical benefits, not just an opinion thing.

Would be good to make a decision on this soon (i.e. prioritise action), so we can either merge this or close it out.

Either way we can always bring it back, or remove it at a later date, its not like we can't back out of any decision we make.

schogges:
> > First of all, I have no big issues with pinning a version in an application (downstream). And as far as I know we had no issues with a pinned version of a direct dependency in this project yet. We are adding more stuff (including another dependency) that attempts to solve a problem we haven't had yet. So, right now I see no real needs in adding this.
> 
> Yes definitely, in some sense I can agree with this entirely. We ourselves **in this repository have had no issues with an exact pinned direct dependency** _as yet_ (actually 🤔 I'm not even sure thats true, I'd have to go back to when we had some pinned dependencies tho). I also have no doubt that neither I nor yourself would have to be nudged to ensure we use a range version over an exact version. And also "I know I know another dependency!" 😆 These facts make me think this PR isn't worth merging.

Fair point 😄 I have to correct myself that I couldn't think of a situation where I faced issues with a pinned version in an app (downstream). In some project we actually used to pin all versions and it worked fine (`save-exact=true`). But no other project depended on it 🙂 

> But! (there's always a but! 😆 ). What about external contributors? What about future contributors? Is this an actual usecase, - letting maintainers of a project define what version range folks should use?
> 
> @adamdehaven reminded me of this:
> 
> > Dependencies saved to package.json will be configured with an exact version rather than using npm's default semver range operator.
> 
> https://docs.npmjs.com/cli/v8/using-npm/config#save-exact
> 
> "rather than using npm's **default semver range** operator" i.e. npm wants you to use a range by default, and it will use a caret range by default if you don't specify a range when using `npm install`.
> 
> Problem solved right? Why do we need a lint rule? Just always use `npm install` and lets not set `save-exact` in our npmrc file.

That's a good point. We actually might want to explicitely set `save-exact=false` in this project's `.npmrc` to overrule personal settings.

> This also leads me to the next and perhaps more important benefit, I have no idea _why_ someone exact pinned a dependency.
> 
> I've split this in two as there is another reason I've realised that I like this during our discussion in this very PR.
> 
> Lets forget about bugs and whats right and whats wrong technically. Lets talk about documenting our decisions, architectural decision records, in this case small ones, but still important ones.
> 
> If I see an exact pinned dependency in a project I maintain, I really want to know why its exact pinned. At what point can I upgrade it? Why was it pinned in the first place? Thing is package.json files are JSON, so we all know we can't add comments! Damn JSON! 😆
> 
> But if we are forced to add an ignore statement, its likely to be noticed and commented upon in a PR and probably easy-ish to find via git blame (as I suggested above). The further thing I realised after suggesting that is, not only do you have to add an ignore statement to exact pin a dependency, but you have to add the ignore statement _in a comment-able Javascript file_. Bingo! I don't care about your comment-less format JSON! 😅
> 
> There's my mini-ADR for why it was pinned, future us will thank ourselves.
> 
> Same thing stands, happy to shut this down if you feel there is no benefit at all, also happy to chat offline.
> 
> But whichever I think we should make a decision on this sooner rather than later, so we can move forwards with it one way or another.

I agree and we should come to a decision. I think you convinced me with the aspect of documentation. The project's and npm default is caret version range and this PR enforces that. Any reason to change that for a dependency will be documented, as you pointed out. Therefore I'd go with it and give it a try. If we face any issues, we can always remove it.

> P.S. I didn't even talk about the other benefits here like using the json schema to make sure our `engines` property is in line with our `.nvm` file, and probably a couple of other things.

Yeah, this is nice 👍 🙂

johncowen:
Ok let me rebase this and we can go give it a go, I think we are both on the same page here, some reservations but overall think there are some benefits here.

Also importantly:

> Therefore I'd go with it and give it a try. If we face any issues, we can always remove it.

🙌 my thoughts exactly!

Lemme rebase, and I'll ping for a final review

packages/config/src/package.schema.json
```diff
@@ -0,0 +1,62 @@
...
+      "pattern": "^\\^"
+    },
+    "workspaceOnlyVersion": {
+      "type": "string",
+      "pattern": "(^\\*$)|(^file:)"
+    },
+    "caretOrWorkspaceOnlyVersion": {
+      "type": "string",
+      "pattern": "(^\\^)|(^\\*$)|(^file:)"
+    },
```
schogges, line 15: We use `npm` and probably stay with it, so this is not needed right now, but could be worth to add here sooner or later. At least I leave a comment for future reference 🙂 

With `pnpm` and `yarn` we would need to add `^workspace:`

```suggestion
    "workspaceOnlyVersion": {
      "type": "string",
      "pattern": "(^\\*$)|(^file:)|(^workspace:)"
    },
    "caretOrWorkspaceOnlyVersion": {
      "type": "string",
      "pattern": "(^\\^)|(^\\*$)|(^file:)|(^workspace:)"
    },
```

johncowen:
Ok I think we are good here for final review 👍

schogges (approved):
👍

johncowen:
Ok time to merge this, just wanted to say thanks all for your input, its very much appreciated.

Just to note its never too late to add comments/ask questions here (or elsewhere) if anyone thinks of any other reasons for or against this. Its never too late to back out if we are convinced otherwise.

PR #3239, opened 2024-11-28
fix(dataplanes): Correct Dataplane Inbound cluster tab filter

johncowen:
Previously we were finding lines for inbounds in the cluster response by using the `kuma.io/service` name. We should just be using `localhost_<port>`.

There are few more semi-related things here which I've also noted inline:

- Altered a e2e test step so we can look for non-existence of text
- Upgraded DataSource usage to use `uri` for better type checking.
- Made both Inbound and Outbound views as consistent as possible.

### DIFF
--- a/packages/kuma-gui/cypress/support/step_definitions/index.ts
+++ b/packages/kuma-gui/cypress/support/step_definitions/index.ts
@@ -248,10 +248,12 @@ Then(/^the "(.*)" element(s)? contain[s]?$/, (selector: string, multiple = '', t
     })
   }
 })
-Then(/^the "(.*)" element contains "(.*)"$/, (selector: string, value: string) => {
+Then(/^the "(.*)" element( doesn't | )contain[s]? "(.*)"$/, (selector: string, assertion: string, value: string) => {
   $(selector).then(($el) => {
+    const negative = assertion !== ' '
+    const prefix = negative ? 'not.' : ''
     const chainer = $el[0].tagName === 'INPUT' ? 'have.value' : 'contain'
-    $(selector).should(chainer, value)
+    $(selector).should(`${prefix}${chainer}`, value)
   })
 })
 
--- a/packages/kuma-gui/src/app/connections/views/ConnectionInboundSummaryClustersView.vue
+++ b/packages/kuma-gui/src/app/connections/views/ConnectionInboundSummaryClustersView.vue
@@ -9,7 +9,7 @@
       connection: '',
     }"
     name="connection-inbound-summary-clusters-view"
-    v-slot="{ route }"
+    v-slot="{ route, uri }"
   >
     <RouteTitle
       :render="false"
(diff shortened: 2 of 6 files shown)


packages/kuma-gui/src/app/connections/views/ConnectionOutboundSummaryClustersView.vue
```diff
@@ -9,48 +9,56 @@
...
+            </XCodeBlock>
+          </DataCollection>
+        </template>
       </DataLoader>
     </AppView>
   </RouteView>
 </template>
 <script lang="ts" setup>
-import type { ClustersDataSource } from '@/app/data-planes/sources'
+import { sources } from '@/app/data-planes/sources'
```
johncowen, line 63: The changes in this file are just to make it as consistent as possible with inbounds, the only things that changes is the setting of the `prefix` property.

packages/kuma-gui/cypress/support/step_definitions/index.ts
```diff
@@ -248,10 +248,12 @@ Then(/^the "(.*)" element(s)? contain[s]?$/, (selector: string, multiple = '', t
...
   }
 })
-Then(/^the "(.*)" element contains "(.*)"$/, (selector: string, value: string) => {
+Then(/^the "(.*)" element( doesn't | )contain[s]? "(.*)"$/, (selector: string, assertion: string, value: string) => {
   $(selector).then(($el) => {
+    const negative = assertion !== ' '
+    const prefix = negative ? 'not.' : ''
     const chainer = $el[0].tagName === 'INPUT' ? 'have.value' : 'contain'
-    $(selector).should(chainer, value)
+    $(selector).should(`${prefix}${chainer}`, value)
```
johncowen, line 256: This enables us to do "doesn't contain" in our tests.

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -94,7 +93,7 @@ export const DataplaneNetworking = {
             // inbound address, advertisedAddress, networkingAddress because externally accessible address
             addressPort: `${item.address ?? networking.advertisedAddress ?? networking.address}:${item.port}`,
             // inbound serviceAddress, inbound address, networkingAddress because the internal services accessible address
-            serviceAddressPort: `${item.serviceAddress ?? address}:${port}`,
+            serviceAddressPort: `${item.serviceAddress ?? address}:${item.servicePort ?? item.port}`,
```
johncowen, line 96: I don't think we should ever use the `servicePort` for when displaying/rendering/calculating inbounds

packages/kuma-gui/src/app/connections/views/ConnectionInboundSummaryClustersView.vue
```diff
@@ -9,53 +9,56 @@
...
+        })"
+        v-slot="{ data , refresh }"
       >
-        <DataCollection
-          :items="clusters!.split('\n')"
-          :predicate="item => item.startsWith(`${props.data.service}::`)"
-          v-slot="{ items: lines }"
+        <template
+          v-for="prefix in [route.params.connection.replace('_', ':')]"
+          :key="typeof prefix"
```
schogges, line 28: Just a nit because there is only one item in the loop, but I think using `typeof` can easily lead to non-unique keys. So do we need a `key` at all? Can it be the `prefix` itself or maybe the `connection`?

johncowen: > So do we need a key at all?

Yeah, unfortunately we do because of the linter 😭 . If we didn't have the linter I wouldn't do this, but in some case with `for` loops the `key` is important.

Moreover, in order to satisfy the linter you need to use a variable from the for statement itself.

I found this trick a while back, in the case of where you specifically don't want Vue to use its detach/reattach behaviour in a for loop (instead you want it to just change the properties of existing components), you can use `:key="typeof variableFromTheLoop"`. This means:

1. `key` will never change, so Vue will never attach/detach, and instead reset properties of existing components if the data changes.
2. We are using a variable from the for loop so this satisfies the linter.

I only found this out a while ago, so there are probably loads of places where we should use this form to prevent unwanted `mount/unmount` hooks being called unnecessarily.

Let me know if you have more questions around this, I not its a little strange looking, I can try and find the PR where this was super important - I remember also making a note on that PR so that could be useful.

johncowen: Ah here we go (easier to find than I thought 😄 ), take a look at the description https://github.com/kumahq/kuma-gui/pull/2878

johncowen: Actually, I wonder if I could add this info somewhere in our engineering docs 🤔

schogges: Hmmm, I see 🤔 I agree that we shouldn't disable the rule on a global level, but probably a simple eslint-disable directive would have worked aswell? Would mean less confusion 😄

schogges: I'll approve and unblock this 🙂

schogges (approved):

Issue #3238, opened 2024-11-28
rules for outbounds on builtin gateway are broken


lahabana:
### What happened?

While investigating #3237 we noticed outbound rules do not work on gateway dataplanes.

My guess is that the matching on names doesn't handle the hash suffix present on gateway clusters.
Also as #3237 points out we seem to be misclasifying the type of the outbound cluster to be an old service instead of a new one.

<img width="669" alt="image" src="https://github.com/user-attachments/assets/d277173c-aa5f-4d46-b4f7-26da087f5384">

rules output (really only the MeshTimeout is relevant here)

```json
{
 "httpMatches": [
  {
   "hash": "9Zuf5Tg79OuZcQITwBbQykxAk2u4fRKrwYn3//AL4Yo=",
   "match": [
    {
     "path": {
      "value": "/",
      "type": "PathPrefix"
     }
    }
   ]
  }
 ],
 "resource": {
  "labels": {
   "k8s.kuma.io/namespace": "kuma-demo",
   "kuma.io/display-name": "edge-gateway-69c4dfc4d4-c2kzg",
   "kuma.io/env": "kubernetes",
   "kuma.io/mesh": "default",
   "kuma.io/origin": "zone",
   "kuma.io/zone": "default"
  },
  "mesh": "default",
  "name": "edge-gateway-69c4dfc4d4-c2kzg.kuma-demo",
  "type": "Dataplane"
 },
 "rules": [
  {
   "fromRules": [],
   "toResourceRules": [
    {
     "conf": [
      {
       "rules": [
        {
         "matches": [
          {
           "path": {
            "value": "/",
            "type": "PathPrefix"
           }
          }
         ],
         "default": {
          "backendRefs": [
           {
            "kind": "MeshService",
            "name": "demo-app",
            "namespace": "kuma-demo",
            "weight": 1,
            "port": 5050
           }
          ]
         }
        }
       ]
      }
     ],
     "origin": [
      {
       "resourceMeta": {
        "labels": {
         "k8s.kuma.io/namespace": "kuma-system",
         "kuma.io/display-name": "demo-app-edge-gateway",
         "kuma.io/env": "kubernetes",
         "kuma.io/mesh": "default",
         "kuma.io/origin": "zone",
         "kuma.io/policy-role": "system",
         "kuma.io/zone": "default"
        },
        "mesh": "default",
        "name": "demo-app-edge-gateway.kuma-system",
        "type": "MeshHTTPRoute"
       },
[...]

Issue #3237, opened 2024-11-28
Reflect backend decision on stats line format separator


lahabana:
This mostly works with #3186.

The few things that needs improvement:

1. We still show kuma.io/service even though we use `MeshService: Exclusive`
<img width="859" alt="image" src="https://github.com/user-attachments/assets/51a2cd01-6888-496a-be43-fc573a0f5f3b">

2. Worth understanding that `default_demo-app_kuma-demo_default_msvc_5050-f2d182aa9efc4e85` has a suffix after the port which relates to the listener.
   1. work with backend on precisely explaining this suffix
   2. should we use `_` as separator for the suffix so it's compliant or do we do more complex regex matching?

johncowen:
Point one here is dealt with in https://github.com/kumahq/kuma-gui/pull/3240, gonna chase up with backend and see if there is anywhere that describes the format of these names with all the potential differences.

PR #3240, opened 2024-11-28
fix(dataplanes): correct MeshService regexp based matching

johncowen:
Sometimes clusterNames can have hashes after the port. This means some of our clusterName matching is a little too strict.

This PR remove the "end of line" match from our `_msvc_` regexp matching to make sure we also match services with hashes after the port.

Closes https://github.com/kumahq/kuma-gui/issues/3238

### DIFF
--- a/packages/kuma-gui/src/app/connections/data/index.ts
+++ b/packages/kuma-gui/src/app/connections/data/index.ts
@@ -4,7 +4,7 @@ const appProtocols = ['http', 'tcp', 'grpc'] as const
 // `_` followed by 1 to 5 digits followed by a `.`
 const trailingPortRe = /_\d{1,5}\./
 const trailingPortRe2 = /_\d{1,5}/
-const meshServiceRe = /_(mz|m|ext){1}svc_\d{1,5}$/
+const meshServiceRe = /_(mz|m|ext){1}svc_\d{1,5}/
 
 export const Stat = {
   fromCollection(items: string) {
--- a/packages/kuma-gui/src/app/connections/data/index.ts
+++ b/packages/kuma-gui/src/app/connections/data/index.ts
@@ -4,7 +4,7 @@ const appProtocols = ['http', 'tcp', 'grpc'] as const
 // `_` followed by 1 to 5 digits followed by a `.`
 const trailingPortRe = /_\d{1,5}\./
 const trailingPortRe2 = /_\d{1,5}/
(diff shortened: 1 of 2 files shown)


packages/kuma-gui/src/app/connections/data/index.ts
```diff
@@ -4,7 +4,7 @@ const appProtocols = ['http', 'tcp', 'grpc'] as const
 // `_` followed by 1 to 5 digits followed by a `.`
 const trailingPortRe = /_\d{1,5}\./
 const trailingPortRe2 = /_\d{1,5}/
-const meshServiceRe = /_(mz|m|ext){1}svc_\d{1,5}$/
+const meshServiceRe = /_(mz|m|ext){1}svc_\d{1,5}/
```
lahabana, line 7:
```suggestion
const meshServiceRe = /_(mz|m|ext){1}svc_\d{1,5}(-[a-z0-9]+)?$/
```

I think this would be the very correct version of the regex

johncowen: Ah even better 🎉  thanks!

packages/kuma-gui/src/app/connections/data/index.ts
```diff
@@ -4,7 +4,7 @@ const appProtocols = ['http', 'tcp', 'grpc'] as const
 // `_` followed by 1 to 5 digits followed by a `.`
 const trailingPortRe = /_\d{1,5}\./
 const trailingPortRe2 = /_\d{1,5}/
-const meshServiceRe = /_(mz|m|ext){1}svc_\d{1,5}$/
+const meshServiceRe = /_(mz|m|ext){1}svc_\d{1,5}(-[a-z0-9]+)?$/
```
johncowen, line 7: @lahabana should we make this forwards compatible now:

> should we use _ as separator for the suffix so it's compliant or do we do more complex regex matching?

https://github.com/kumahq/kuma-gui/issues/3237#issue-2701729992

Wasn't there a MADR somewhere outlining a "standard" format for these names?

schogges (approved):
LGTM 👍

packages/kuma-gui/src/app/connections/data/index.ts
johncowen: Also see https://github.com/kumahq/kuma/issues/12147 which we just made in relation to this

PR #3261, opened 2024-12-04
fix: Use CSS for decoration, fixes CodeQL error

johncowen:
Also see https://github.com/kumahq/kuma-gui/pull/1803#pullrequestreview-1744286128

This should fix an over-zealous CodeQL linting security issue/error by using CSS (a super powerful, declarative and therefore bug free "from The Outside" language) instead of Javascript.

I think this should probably be an almost global type of rule (I would like to delete `TagList` at some point), and I think it would be a good candidate for something similar to `XLayout`, probably called `XTheme` which we would primarily use at the root of our application.

Finally we won't know for sure if this fixes the issue without merging to master, but I think its a better approach even if it doesn't fix the issue.

Screenshot 2024-12-04 at 10 28 40


schogges (approved):
👍

johncowen:
Sorry I wanted to push a little amend here after reading my own comment from way back when. Been trying to get internet so missed the approval 😅 . I'm sure the addition is fine but let me know if not, otherwise I'll leave this hanging for a little while and then go ahead and merge

johncowen:
FYI this fixed the security alert, hopefully for good 🤞 

Screenshot 2024-12-04 at 12 50 00

PR #3262, opened 2024-12-04
fix(app-shell): duplicate slot notifications

schogges:
This removes a duplicate slot `#notifications` from the `ApplicationShell` component. This caused that when using the slot, the notification would appear twice.
This is probably a leftover from an attempt to resolve conflicts of merge or rebase.


johncowen (approved):
Thanks for this, super good spot!

I looked again at the original PR and even though it was up for a very long time, and probably rebased lots, I think it could have been a genuine mistake 😅 so thank you for spotting!

PR #3267, opened 2024-12-06
refactor(dataplane-summary-view): remove tab navigation

schogges:
In order to improve the user experience by reducing the navigation efforts to show the most important information, we have decided to remove the tab navigation from the dataplane summary view.

As pointed out in #3209

> The reason for the separation into tabs was because the summary view could get very crowded already.

I have removed the `inbounds` section. Both `inbounds` and `outbounds` are part of the config, which is now part of the summary view.

Part of #2922


packages/kuma-gui/src/app/data-planes/routes.ts
```diff
@@ -66,26 +66,7 @@ export const routes = () => {
         props: () => ({
           routeName: `${fullPrefix}data-plane-summary-view`,
         }),
-        redirect: { name: `${fullPrefix}data-plane-summary-overview-view` },
         component: () => import('@/app/data-planes/views/DataPlaneSummaryView.vue'),
-        children: [
-          {
-            path: 'summary-overview',
```
johncowen, line 73: So we avoid this for the moment also 🎉 

Good that we avoid it here, that said I think we are going to have to change things slightly in the future so we can avoid such URL clashes totally

(the reason we where using `summary-overview` as a solution was in explained in https://github.com/kumahq/kuma-gui/pull/3209#discussion_r1851696117 and we have to do this because of the unfortunate way our URLs are currently)

johncowen (approved):
Looks good!

I think the GUI is starting to look much better overall now. If I'm just having a glance over my install I can do that without having to leave any list/drawer views at all. I can just click rows and see config without having to dive in and out or more and more GUI pages.

Only if I need to dive deeper do I need to dive deeper into deeper pages.

The only place left where I can't do this is Gateways (both built in a delegated lists) and legacy external services (and there is an argument to say we don't need to bother doing legacy external services)

packages/kuma-gui/src/app/data-planes/routes.ts
schogges: Yes, exactly 😄 👍

johncowen: I think this is a good PR to cross reference here https://github.com/kumahq/kuma-gui/pull/1459

Not sure if its super clear from the conversation there but I was always very keen to use plurals for listings/summaries and singular for detail pages:

Something like:

- `/dataplanes` <- listing no summary
- `/dataplanes/dataplane/my-dataplane/overview` <- listing with summary overview for my-dataplane
- `/dataplane/my-dataplane/overview` <- detail page overview for my-dataplane

PR #3284, opened 2024-12-09
feat(gateways/builtin-gateway-list-view): add summary drawer with config

schogges:
As part of the efforts to show the config "everywhere" I've now also added a summary drawer for the `builtin` gateway. I could only add it for the `builtin` gateway because this can be configured via `MeshGateway`. For the `delegated` gateway this is not possible because the config is part of other resources like `Dataplane` (refer to [kuma.io/managing-ingress-traffic/delegated](https://kuma.io/docs/2.9.x/using-mesh/managing-ingress-traffic/delegated/)).

Part of #2922


packages/kuma-gui/src/app/gateways/routes.ts
```diff
@@ -62,6 +62,16 @@ export const routes = () => {
               path: 'builtin',
               name: 'builtin-gateway-list-view',
               component: () => import('@/app/gateways/views/BuiltinGatewayListView.vue'),
+              children: [
+                {
+                  path: ':gateway',
+                  name: 'builtin-gateway-summary-view',
+                  props: () => ({
+                    routeName: 'builtin-gateway-summary-view',
+                  }),
```
johncowen, line 71: Oh sorry missed this, I think I added a PR that made it so you no longer need to add these here, they are added automatically. Unfortunately you still need the `defineProps<{routeName: string}>()` in the route component for the moment. (gonna fire off a single comment, and keep reviewing but I think this is the last one and I can approve)

johncowen (commented):
Yep just that one last comment on the props, and I can then give this a 👍

packages/kuma-gui/src/app/gateways/routes.ts
schogges: Ah, right 🙂

PR #3296, opened 2024-12-10
feat(zones): Add zone-ingress traffic view

johncowen:
I've been working on a new traffic visualisation for zone-ingresses similar to our visualisation that we have for Dataplanes.

I'm made some progress, but felt like it was a good time to put some work up for review, partly to preview progress and also to PR some scattered/tiny but important refactors.

Things here:

1. ~Added a temporary feature flag that can be used to preview the work, `KUMA_ZONE_TRAFFIC`.~ I took this out. I think we are ok to ship as is.
2. Moved the exclusion filter that we apply when parsing `/stats` deeper into the data layer, this bit will apply everywhere we parse stats.
3. Added a new `/connections/stats/...` source modelled on the one we have for dataplanes. Note: I plan to move all the sources we have in this file to use a `/connections/...` prefix (the same name as the module)
4. When building up to do this work, I went over our existing Dataplane viz, mainly to refresh my memory and stare at it a bit to figure out how to re-use or re-implement this for zones. Whilst doing that I made some very small amends. This should only be upgrading components, and maybe a bit of a rename/re-nest. There should be no functionality changes in this file.
5. Finally, new work in the Zone Ingress Detail page. Note: The cards don't currently link to drawers, that will be the next PR.

If you want to try this out yourself with a local install see screengrab below for relevant cookies:

Screenshot 2024-12-13 at 12 56 30

### DIFF
--- a/packages/kuma-gui/src/app/connections/sources.ts
+++ b/packages/kuma-gui/src/app/connections/sources.ts
@@ -73,6 +73,11 @@ export const sources = (source: Source, api: KumaApi) => {
         : Object.fromEntries(Object.entries(connections.listener).filter(([key, value]) => key.startsWith(`${params.address}_`) && !value.$clusterName.startsWith('_')))
 
       const outbounds = Object.fromEntries(Object.entries(connections.cluster).filter(([key, _value]) => ![
+        // if we don't exclude localhost_ we end up with  a `localhost_`
+        // outbound, which is the cluster of the inbound. Whilst we don't want
+        // to show this now, we might want to later as it can show how envoy
+        // might be interacting with the traffic vs the cluster/service itself
+        'localhost_',
         'inbound_passthrough_',
         'outbound_passthrough_',
       ].some(item => key.startsWith(item))))
--- a/packages/kuma-gui/src/app/connections/sources.ts
+++ b/packages/kuma-gui/src/app/connections/sources.ts
@@ -43,6 +43,20 @@ const filter = (data: Record<string, unknown>, cb: (key: string, arr: unknown[])
 }
 export const sources = (source: Source, api: KumaApi) => {
   return defineSources({
+    '/connections/stats/for/zone-ingress/:name/:socketAddress': async (params) => {
+      const { name, socketAddress } = params
+      const res = await api.getZoneIngressData({
+        zoneIngressName: name,
+        dataPath: 'stats',
+      })
+      const connections = ConnectionCollection.fromObject(Stat.fromCollection(res))
+      return {
+        inbounds: Object.fromEntries(Object.entries(connections.listener).filter(([key, _value]) => key.startsWith(socketAddress.replace(':', '_')))),
+        outbounds: connections.cluster,
+        $raw: res,
+        raw: res,
+      }
+    },
     '/meshes/:mesh/dataplanes/:name/stats/:address': async (params) => {
       const { mesh, name } = params
       const res = await api.getDataplaneData({
@@ -57,26 +71,10 @@ export const sources = (source: Source, api: KumaApi) => {
       const inbounds = params.address === 'localhost'
         ? Object.fromEntries(Object.entries(connections.cluster).filter(([key, _value]) => key.startsWith('localhost_')))
         : Object.fromEntries(Object.entries(connections.listener).filter(([key, value]) => key.startsWith(`${params.address}_`) && !value.$clusterName.startsWith('_')))
-      // pick out the outbounds which aren't internal outbounds
+
       const outbounds = Object.fromEntries(Object.entries(connections.cluster).filter(([key, _value]) => ![
-        // if we don't exclude localhost_ we end up with  a `localhost_`
-        // outbound, which is the cluster of the inbound. Whilst we don't want
-        // to show this now, we might want to later as it can show how envoy
-        // might be interacting with the traffic vs the cluster/service itself
-        'localhost_',
-        //
-        '_', // most internal names will be prefixed by `_` the rest will become legacy internal names
-        'admin',
-        'async-client',
-        'kuma_readiness',
-        'kuma_envoy_admin',
-        'probe_listener',
         'inbound_passthrough_',
         'outbound_passthrough_',
-        'access_log_sink',
-        'ads_cluster',
-        'meshtrace_zipkin',
-        'meshtrace_opentelemetry',
       ].some(item => key.startsWith(item))))
 
       // pick out outbounds passthrough
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -167,11 +167,13 @@
             </template>
           </XAboutCard>
 
-          <KCard
+          <XCard
             class="traffic"
             data-testid="dataplane-traffic"
           >
-            <div class="columns">
+            <XLayout
+              type="columns"
+            >
               <ConnectionTraffic>
                 <template #title>
                   <ForwardIcon
@@ -181,34 +183,34 @@
                   />
                   Inbounds
                 </template>
-                <ConnectionGroup
-                  type="inbound"
-                  data-testid="dataplane-inbounds"
+                <!-- if we are a builtin gateway proxy i.e. a 'gateway' proxy -->
+                <!-- use its first and only inbounds as a template  -->
+                <!-- and fill the inbounds out from the stats once they are loaded -->
+                <template
+                  v-for="inbounds in [props.data.dataplane.networking.type === 'gateway' ? Object.entries<any>(traffic?.inbounds ?? {}).reduce<DataplaneInbound[]>((prev, [key, value]) => {
+                    // As we are just 'finding' inbounds from stats without knowing them
+                    // from the dataplane overview API request, we will wrongly 'find'
+                    // the envoy admin inbound that is used for kumas /stats API. Ignore
+                    // the envoy admin inbound when we find it
+                    const port = key.split('_').at(-1)
+                    if(port === (props.data.dataplane.networking.admin?.port ?? '9901')) {
+                      return prev
+                    }
+                    return prev.concat([
+                      {
+                        ...props.data.dataplane.networking.inbounds[0],
+                        name: key,
+                        port: Number(port),
+                        protocol: ['http', 'tcp'].find(item => typeof value[item] !== 'undefined') ?? 'tcp',
+                        addressPort: `${props.data.dataplane.networking.inbounds[0].address}:${port}`,
+                      },
+                    ])
+                  }, []) : props.data.dataplane.networking.inbounds]"
+                  :key="inbounds"
                 >
-                  <!-- if we are a builtin gateway proxy i.e. a 'gateway' proxy -->
-                  <!-- use its first and only inbounds as a template  -->
-                  <!-- and fill the inbounds out from the stats once they are loaded -->
-                  <template
-                    v-for="inbounds in [props.data.dataplane.networking.type === 'gateway' ? Object.entries<any>(traffic?.inbounds ?? {}).reduce<DataplaneInbound[]>((prev, [key, value]) => {
-                      // As we are just 'finding' inbounds from stats without knowing them
-                      // from the dataplane overview API request, we will wrongly 'find'
-                      // the envoy admin inbound that is used for kumas /stats API. Ignore
-                      // the envoy admin inbound when we find it
-                      const port = key.split('_').at(-1)
-                      if(port === (props.data.dataplane.networking.admin?.port ?? '9901')) {
-                        return prev
-                      }
-                      return prev.concat([
-                        {
-                          ...props.data.dataplane.networking.inbounds[0],
-                          name: key,
-                          port: Number(port),
-                          protocol: ['http', 'tcp'].find(item => typeof value[item] !== 'undefined') ?? 'tcp',
-                          addressPort: `${props.data.dataplane.networking.inbounds[0].address}:${port}`,
-                        },
-                      ])
-                    }, []) : props.data.dataplane.networking.inbounds]"
-                    :key="inbounds"
+                  <ConnectionGroup
+                    type="inbound"
+                    data-testid="dataplane-inbounds"
                   >
                     <!-- don't show a card for anything on port 49151 as those are service-less inbounds -->
                     <DataCollection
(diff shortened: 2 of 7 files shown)


packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -177,34 +183,35 @@
...
+                      {
+                        ...props.data.dataplane.networking.inbounds[0],
+                        name: key,
+                        port: Number(port),
+                        protocol: ['http', 'tcp'].find(item => typeof value[item] !== 'undefined') ?? 'tcp',
+                        addressPort: `${props.data.dataplane.networking.inbounds[0].address}:${port}`,
+                      },
+                    ])
+                  }, []) : props.data.dataplane.networking.inbounds]"
+                  :key="inbounds"
```
schogges, line 210: Would it make sense to move this out of the template? It's hard to read because of ternary and nested logic. I know that this just pops up due to some surrounding changes and this haven't been touched, but I still want to mention it and see what you think about it 🙂

packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressDetailView.vue
```diff
@@ -4,171 +4,271 @@
...
+          type="stack"
+        >
+          <XTimespan
+            :start="t('common.formats.datetime', { value: Date.parse(props.data.creationTime) })"
+            :end="t('common.formats.datetime', { value: Date.parse(props.data.modificationTime) })"
+          />
+
           <XLayout
-            type="stack"
+            class="columns"
```
schogges, line 22: Shouldn't this be `type="columns"`? 🤔

packages/kuma-gui/src/app/connections/sources.ts
```diff
@@ -43,6 +43,20 @@ const filter = (data: Record<string, unknown>, cb: (key: string, arr: unknown[])
...
+      const res = await api.getZoneIngressData({
+        zoneIngressName: name,
+        dataPath: 'stats',
+      })
+      const connections = ConnectionCollection.fromObject(Stat.fromCollection(res))
+      return {
+        inbounds: Object.fromEntries(Object.entries(connections.listener).filter(([key, _value]) => key.startsWith(socketAddress.replace(':', '_')))),
+        outbounds: connections.cluster,
+        $raw: res,
+        raw: res,
```
schogges, line 57: Looks like this is duplicate? For the naming I think we decided for `$raw` 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: > Would it make sense to move this out of the template? It's hard to read because of ternary and nested logic.

I would say no 😅 

Ummm, I would encourage you to read https://copyconstruct.medium.com/small-functions-considered-harmful-91035d316c29. Whilst you know I don't like it when people quote things from "somebody somewhere on the internet", but this person I happen to highly respect, I think a lot of folks do.

Cindy's intro and quite a useful TLDR;

> Some people seem so enamored with small functions that the idea of abstracting any and every piece of logic that might seem even nominally complex into a separate function is something that is passionately advocated for.

> I’ve worked on codebases inherited from folks who’d internalized this idea to such an unholy extent that the end result was pretty hellish and entirely antithetical to all the good intentions the road to it was paved with. 

I think the sections headed "Loss of Locality" and "Hard to Read" are particularly relevant here. Both specifically argue against moving things out of the place where its being executed (more or less) in the name of making things less hard to read.

I have also worked on codebases inherited from folks who’d internalized this idea to such an unholy extent that the end result was pretty hellish and entirely antithetical to all the good intentions the road to it was paved with. It has taken us a long while and a lot of extremely hard work (not just typing code into a keyboard) to get to the more solid foundations of where we are today.

Just to give you a heads up, I would guess that unless there is something technically incorrect here (let me know if there is!), I'm inclined not to change anything especially within the scope of this work (as you've pointed out), but let me know what you think of the article, I think thats the most important point here 👍

packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressDetailView.vue
johncowen: Ohh! good spot lemme check

packages/kuma-gui/src/app/connections/sources.ts
johncowen: Yeah I wasn't totally sure about this, I just copied how it was elsewhere to remain the same. Let me see if I can remember why I originally did both, I think it was a similar things to the fact that we have `config` and `$raw` only here its not `config` but i had to name it something so it ended up being `raw` (`$raw` was originally supposed to be a non-obvious to access property)

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
schogges: Thanks for sharing this article 🙂 I really like it and I fully agree with it and your position. I also think that too many abstractions and outsourcing is counter productive and actually makes it harder to understand the code. Especially if I have to open a lot of different files for the full picture.

> Just to give you a heads up, I would guess that unless there is something technically incorrect here (let me know if there is!), I'm inclined not to change anything especially within the scope of this work (as you've pointed out), but let me know what you think of the article, I think thats the most important point here 👍

There is nothing wrong with this on the technical perspective, it just popped up due to the surrounding changes and I don't expect any changes to this within the scope of this PR.

However, my point was not to create abstractions for this, I just meant to ask if there is a better location for this other than the template. If I understand it correctly, we are partially shaping some nested data here, which I think should be part of the data layer. Please correct me if I'm wrong 🙂 

---

Feel free to resolve this discussion any time. This is not meant to block the progress of the PR.

johncowen: > Thanks for sharing this article 🙂 I really like it and I fully agree with it and your position. I also think that too many abstractions and outsourcing is counter productive and actually makes it harder to understand the code. Especially if I have to open a lot of different files for the full picture.

🙌  awesome, glad you like the article.

> There is nothing wrong with this on the technical perspective, it just popped up due to the surrounding changes and I don't expect any changes to this within the scope of this PR.

Cool yeah, sorry I totally think we are on the same page here by the way (not technically problematic and anything else would be out of scope), it's not my favourite piece of code but there are reasons that it is like it is. Its a bit difficult to explain the reasons why it is like it is, and its probably a longer story than this but 1 sec lemme try:

---

> If I understand it correctly, we are partially shaping some nested data here, which I think should be part of the data layer.

This is almost 100% correct! But! (yes, you guessed it there's always a but! 😆 ). We are merging two pieces of data here and this fact is very important.

`DataSource` is designed to be composable, and its also designed to be used declaratively inside templates so that complex loading/erroring displays are also composable. In order to make displayable composable data sourcing/loading/erroring the best place to do that is in the template because its all about the visuals. I know you know quite a bit about how DataSource works so I'm pretty sure you follow the approach here.

In this particular instance, we really wanted to have the two HTTP calls we use here be **independently** loaded/error'ed, so the option of using say `Promise.all([firstCall, secondCall]).then(doSomeMerging)` from within the `sources.ts` file, and therefore then use our data layer to do data merging shaping is not ideal. If we did this the loading/erroring in the template would have to be one single thing, each call (with its loading/erroring) wouldn't be an "independent display".

Hopefully that last paragraph ^ makes sense. Let me know if not because its really important and what everything else hangs on that.

If I've explained the above well enough, you'll see that the merging is best done in the the template. Granted, we could make a function and move the code elsewhere and just have the call site in the template. But we are risking some of the points pointed out in the above article.

Lastly, all of this said, I'm a firm believer in making sure we understand what's required from a piece of code before we start to try to optimize it or make it reusable (or even move it elsewhere if that makes sense, despite the above article).

The piece of work here (so the old dataplane work, not the current scope of zones) had very many unknowns all the way from start to finish. I specifically wanted to treat it like prototyping work until we knew what we wanted. I think only now I'm starting to get a feel for what I think we want because I'm doing similar functionality elsewhere (zones). I'm so glad I didn't try to prematurely optimize or worry too much about small pieces of code like this because I've found out over the past few days that zones is very different in subtle ways and if I'd tried to organize this a little better already it most definitely would have been premature.

Anyway, thanks for asking the questions, its really nice to have the reason and opportunity to try and explain these things in PRs. Not only is it good to share thinking, and validate or (or sometimes not 😅 ) reasoning, but it acts as good documentation for future us 🙌

schogges: Ohhh, now this makes sense to me 😄 I've completely missed the `DataSource` and just saw the `props.data`, so I thought there is just one data set 🙈 
I might be wrong, but with the declarative data fetching approach (which I like), I think there is currently no good way of merging two data sets from different sources in the data layer (or outside of the template) 🤔 

> Lastly, all of this said, I'm a firm believer in making sure we understand what's required from a piece of code before we start to try to optimize it or make it reusable (or even move it elsewhere if that makes sense, despite the above article).
>
> The piece of work here (so the old dataplane work, not the current scope of zones) had very many unknowns all the way from start to finish. I specifically wanted to treat it like prototyping work until we knew what we wanted. I think only now I'm starting to get a feel for what I think we want because I'm doing similar functionality elsewhere (zones). I'm so glad I didn't try to prematurely optimize or worry too much about small pieces of code like this because I've found out over the past few days that zones is very different in subtle ways and if I'd tried to organize this a little better already it most definitely would have been premature.
>
> Anyway, thanks for asking the questions, its really nice to have the reason and opportunity to try and explain these things in PRs. Not only is it good to share thinking, and validate or (or sometimes not 😅 ) reasoning, but it acts as good documentation for future us 🙌

Yes, fully agree 💯 Thanks for the patience and answering my questions 🙂

packages/kuma-gui/src/app/connections/sources.ts
johncowen: Ah yeah, I just went back to look at this.

The idea of `$raw` is that it shouldn't be amended or used unless we absolutely need to. It should be the exact response from the endpoint and we should avoid using it unless we are in a sticky situation (maybe as a temporary escape hatch).

`/stats` is unlike most other endpoints in that its plain text. We put this context in `raw` in order to display it. We also then add parsed and reshaped JSON as sibling properties for a when we display it differently to plain text (using GUI cards).

Essentially, right now at least, `raw` is the same as `$raw`, but we should avoid displaying `$raw` or potentially mutating `$raw`.

Does that help here? Do you think that thinking makes sense? i.e. keeping a semi-private `$raw` as a reference to the original HTTP response?

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: >I might be wrong, but with the declarative data fetching approach (which I like), I think there is currently no good way of merging two data sets from different sources in the data layer (or outside of the template) 🤔

Well there is if you don't want to have independent loading/error states. You can use `Promise.all` then merge from within `sources.ts`, so thats possible if you need to do this in the future.

There are also other approaches but what we do here is by far the simplest and keeps the independent loading/errors states in a place where you expect to see them ("expect to see them" both from a user perspective and a code perspective)

packages/kuma-gui/src/app/connections/sources.ts
schogges: hmm 🤔 I don't really see a benefit in having both identical entries. Also, if `$raw` is not meant to be accessed/used anywhere, why should we add it then? If this is a preparation for a sticky situation we might or might not meet in the future, I would see it as a premature optimization.
However, I do see a benefit in having the exact response from the API somewhere next to the sibling entries that we add in the data layer. So, I would definitely keep one of them 🙂

johncowen: > Also, if $raw is not meant to be accessed/used anywhere, why should we add it then? If this is a preparation for a sticky situation we might or might not meet in the future, I would see it as a premature optimization.

We've had times in the past where we've mistakenly added to the actual HTTP response (by adding our own custom props), which have then ended up being displayed in config. `raw` is supposed to be the actual content we want to show to users (in other places with this called `config`, its not called that here because its not really "config"), importantly without the additional custom props.

It might not be super clear why we then have another additional `$raw` property. This is the content without anything added to it, but also without anything _removed_ from it i.e. the original HTTP response. We sometimes also remove things from the thing we show to users. If we remove it from `raw/config` and its not contained in any custom properties, then we've lost that information completely.

As `$raw` is the actual unedited response, it continues to contain that information everywhere, and aswell as being an escape hatch is also sometimes useful when debugging/console.logging so I guess that last point is an extra benefit aswell as the escape hatch.

Let me know if this is clear, completely get it's weird that its seem that there are two, but the TLDR is:

1. `raw` (sometimes `config`) contains the thing we want to show to the user. Without any custom props and with some non-custom props removed.
2. `$raw` (which I think we wanted to make consistent form a naming point of view), is the actual HTTP response. No custom props, and _still containing some things that we might remove from `raw`_.

Again, I'm glad you asked this as its helped my think about why we do this and write it down somewhere!

Apart from that, and have re-understood myself why we do this, I'm inclined to keep both, and continue to add `$raw` properties if we see anywhere where there currently isn't one. wdyt?

schogges: > Let me know if this is clear, completely get it's weird that its seem that there are two, but the TLDR is:
>
> raw (sometimes config) contains the thing we want to show to the user. Without any custom props and with some non-custom props removed.
$raw (which I think we wanted to make consistent form a naming point of view), is the actual HTTP response. No custom props, and still containing some things that we might remove from raw.

Ok, I think I got your point. Probably it makes more sense to me when I face such a case in which I want to show a `config` or (`raw`) but remove something from it.
But, am I right that we are currently neither referencing `raw` nor `$raw`?

johncowen: > But, am I right that we are currently neither referencing `raw` nor `$raw`?

Oh yes totally correct! 🤦 oh I see why this is even more confusing now!

Umm, if you look at the dataplane equivalent of this, you'll see where we reference `raw` (not `$raw`) and we'll eventually do the same here for zones (the "entire proxy" `Stats` tab will need changing to use the new source (which exposes `raw`), and we we add the "per 'bound" drawer Stats tab will also use it.

schogges: I see 🙂 so this is also a preparation for further additions/changes. I still think that it would be better to add things when they are actually needed, just for reference. But I guess that's just a nit.
So, lets move on then 😄

schogges (approved):

johncowen:
FYI: I'm gonna leave this up a little while more seeing as we asked for feedback on this in todays meeting with the team.

PR #3298, opened 2024-12-13
fix(dataplanes): fixes layout of Status: info tooltip

johncowen:
## Before

Screenshot 2024-12-13 at 11 51 38

## After

Screenshot 2024-12-13 at 11 54 38

I also took the opportunity while I was here to delete some left over files and refactor this template to use our `X*` components.


schogges (approved):
Thanks for the cleanup 🧹 🙂

PR #3300, opened 2024-12-13
refactor: add and use XAboutCard component in detail views

schogges:
These changes introduce a new global component `XAboutCard` that is used to improve the layout of the formerly used `KCard` in the detail views "About" section. It is based on the `AppAboutSection` exported by `@kong-ui-public/app-layout`. The layout in the content slot changed to a horizontal layout with badges (`XBadge`) instead of the column based layout for a general improvement of readability. All `XAboutSection` show the date times of creation and potential modification of the resource.

image

For better visual separation of the `ResourceStatus` entries in the `ControlPlaneStatus` I have introduced a new class `.columns-with-borders` which works similarly to `.stack-with-borders` but for a horizontal layout.

Screenshot 2024-12-13 at 18 28 17

Furthermore I have replaced several occurrences of `class="column"` or `class="stack"` in the files I touched with `XLayout` and replaced static text with i18n.


johncowen:
Would it be much effort to split this PR in two:

1. a PR with the changes related to the innards of `AppAboutSection`
2. a PR containing nothing but the re-introduction of `AppAboutSection` itself.

The reason I'm checking is, if its not too much effort it would be good to do the `AppAboutSection` as a second isolated PR, mostly because we have very recently flip-flopped between using this or not using this. I'd like it if we explicitly made a PR that does this and explains the reasoning behind the flip-flop and almost acts as a mini-ADR. We then have that for a clean reference for wants to know the reasons for the "flip" back.

Also there are a couple of things that I'd like to do with the design of `XAppAboutSection` (seeing as we have that to wrap over `AppAboutSection`), that I'd like to pass by you so it would be good to have a separate PR to do that. lmk!

johncowen:
Ok, I understood that you are going to keep the `XAppAboutSection` and its innards will temporarily contain `KCard` and `XTimespan`?

I guess thats mixing concerns still a little. The reason ideally we'd have it all in one PR is so that would mean the re-introduction of `XAppAboutSection` (or whatever we end up calling it) and its implementation is in one single PR that we can point/link to and not mixed in with other work. Ideally it would all be contained in one PR.

If `XAppAboutSection` can't all go in its own PR, I don't think there a massive point in splitting, it would be splitting for the sake of making two PRs rather than splitting for the sake of clarifying things and having a single PR to point/link to. lmk either way and I can start reviewing this PR or not.

schogges:
> Ok, I understood that you are going to keep the `XAppAboutSection` and its innards will temporarily contain `KCard` and `XTimespan`?
> 
> I guess thats mixing concerns still a little. The reason ideally we'd have it all in one PR is so that would mean the re-introduction of `XAppAboutSection` (or whatever we end up calling it) and its implementation is in one single PR that we can point/link to and not mixed in with other work. Ideally it would all be contained in one PR.
> 
> If `XAppAboutSection` can't all go in its own PR, I don't think there a massive point in splitting, it would be splitting for the sake of making two PRs rather than splitting for the sake of clarifying things and having a single PR to point/link to. lmk either way and I can start reviewing this PR or not.

The main point of adding `XAboutSection` is to add some stylings (e.g. for the `DescriptionCard`). This would still be required when using `KCard` and `XTimestamp` instead, because I would not like to repeat that in all the places where `DescriptionCard` is being used. So regardless of using `AppAboutSection` as a base, I would want to add a component.

packages/kuma-gui/src/app/external-services/views/ExternalServiceDetailView.vue
```diff
@@ -21,31 +21,38 @@
           })"
           v-slot="{ data }"
         >
-          <KCard
+          {{ console.log('test', data) }}
```
johncowen, line 24: 👀  just so you notice this, interesting it wasn't caught by a our linter tho 🤔

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -38,130 +38,132 @@
...
-                  <template #body>
-                    <TextWithCopyButton :text="`${props.data.dataplane.networking.address}`" />
-                  </template>
-                </DefinitionCard>
-              </div>
-            </div>
-          </KCard>
+              <template #body>
+                <XBadge appearance="decorative">
+                  <TextWithCopyButton :text="`${props.data.dataplane.networking.address}`" />
```
johncowen, line 148: Just a note, if we are editing this PR a bit more may aswell update this to `XCopyButton` while we are here.

packages/kuma-gui/src/app/x/components/x-about-section/XAboutSection.vue
```diff
@@ -0,0 +1,40 @@
...
+      gap: $kui-space-20;
+      font-size: $kui-font-size-20;
+      font-weight: normal;
+    }
+
+    .x-layout.separated {
+      align-items: center;
+    }
+  }
+}
```
johncowen, line 39: These are the bits I'm mostly concerned about. I believe (but not totally sure) these are all overwrites, meaning we potentially fall out of sync with upstream, and I think the reasoning behind flipping this component back in is to prevent us from falling out of sync with upstream?

We had a little chat offline about this, specifically we noted that the important things that AppAboutSection gives us is a font-size and a font-color, yet here we overwrite a font-size. That doesn't make sense to me as we then diverge from upstream.

It would be great if we could find an option that meant we can use AppAboutSection without any overwrites at all. I don't get why we need overwrites if nobody else using the same component has overwrites.

packages/kuma-gui/src/app/x/index.ts
```diff
@@ -101,6 +103,7 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
           ['XTeleportSlot', XTeleportSlot],
           ['XTimespan', XTimespan],
           ['XDisclosure', XDisclosure],
+          ['XAboutSection', XAboutSection],
```
johncowen, line 106: ⚠️  Large warning: naming discussion ⚠️  😅 

We have an opportunity to redesign this interface (inc the name) here.

Do you think its weird to name a component after a place that you put it, especially when you could put it anywhere? Would it be better to name it after what it is? To me this component is an `XCard` that has a `Timespan` (or just `Date`).

i.e. `XDateCard`, which happens to fit with other components we have such as `DefinitionCard`, `DataCard` etc, and means our "generic component set" feels the same throughout all its components (similar to using say `:variant` whenever we have a type-like setting)

tiny thing, but I think we are going to make a a few changes here anyway, wdyt?

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
```diff
@@ -18,53 +18,59 @@
...
+            mesh: route.params.mesh,
+            name: route.params.service,
+          })"
+          v-slot="{ data }"
+        >
+          <XAboutSection
+            :title="t('services.internal-service.about.title')"
+            :created="data.creationTime ? t('common.formats.datetime', { value: Date.parse(data.creationTime) }) : undefined"
+            :modified="data.modificationTime ? t('common.formats.datetime', { value: Date.parse(data.modificationTime)}) : undefined"
           >
```
johncowen, line 32: Things would be a lot easier to use if we didn't have to do this. Just to be clear ,I know this PR removes usage of `XTimespan`, but seeing as `XTimespan` only ever displays a span of two dates, the date formatting is contained inside the component, meaning less work at the call site and a more ergonomic component. This is why XTimespan is built as it is.

Actually that made me think, should we be deleting the XTimespan component in this PR or do you think we might need it to show dates somewhere else?

johncowen (commented):
I made an initial pass on this, mostly concentrating on `XAboutAppSection`. Lmk know on those comments, I'd be keen to get this work in by tomorrow/Wednesday preferably 👍 

I was about to look and see why we have to overwrite things and others don't but wondered if you'd checked that before doing this? lmk if so, if not I'll take look.

packages/kuma-gui/src/app/external-services/views/ExternalServiceDetailView.vue
schogges: Oh wow, nice catch! I'll check why it wasn't catched 😄

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
schogges: Yes, absolutely 👍

packages/kuma-gui/src/app/x/components/x-about-section/XAboutSection.vue
schogges: These overwrites are not for the upstream component, but for our internal `DefinitionCard`. There is no "standard" for the content (`#body` slot of `AppAboutSection`) and we could add anything in there. This styling just makes sure that we are (for now) aligned with how others are using this component.

packages/kuma-gui/src/app/x/index.ts
schogges: That's a very good point 🤔 I think there are some (very rare) cases in which I would name the component based on the location and only if the component is not meant to be placed anywhere else (still more than one place in the app). So, in this case I'd be fine either way and I do see benefits in both approaches:

1. Name the component based on the location: should not be used anywhere else and the name does not require any changes in the future. This means we are flexible in how the component evolves, but not flexible in where we can put it. The downside is that if we need to add this component anywhere else, we might create a similar one or come back to this discussion.
2. Name the component based on what it is: allows to put the component anywhere and the name probably needs to change if the content of the component changes (e.g. at some point we add a description text). This means we are flexible in where we put the component, but probably not so flexible in how the component evolves. Also here, we would probably come back to this discussion when the component evolves.

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
schogges: Yeah, I fully agree and dislike that we need to call it like this. This should be handled by the component, also because of consistency, because it does not enforce a common date-time format in all the various places this component is used.

I quickly went through the code and found that I missed to refactor the `HostnameGeneratorDetailView` accordingly. This is the only place left where we use `XTimespan`. So, I guess we can remove it. Once we need it again, we can add it back.

packages/kuma-gui/src/app/x/components/x-about-section/XAboutSection.vue
johncowen: Oh I see! 🤦 Gotcha I missed that it was not really styling for this component but defined in this component, thanks!

packages/kuma-gui/src/app/external-services/views/ExternalServiceDetailView.vue
schogges: Removed and opened https://github.com/kumahq/kuma-gui/pull/3311

packages/kuma-gui/src/app/x/components/x-about-section/XAboutSection.vue
johncowen: You shouldn't need the `x-layout.separated` here anymore a similar rule was added just before you put this PR up I think https://github.com/kumahq/kuma-gui/blob/698c2a75cd3a46f4cfb4363f8199dab77f5438b3/packages/kuma-gui/src/app/x/components/x-layout/XLayout.vue#L27

Actually, thats made me think of something else 🤔

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -38,130 +38,132 @@
...
-                  </template>
-                </DefinitionCard>
+          <XAboutSection
+            :title="t('data-planes.routes.item.about.title')"
+            :created="t('common.formats.datetime', { value: Date.parse(props.data.creationTime) })"
+            :modified="t('common.formats.datetime', { value: Date.parse(props.data.modificationTime) })"
+          >
+            <DefinitionCard layout="horizontal">
+              <template #title>
+                {{ t('http.api.property.status') }}:
```
johncowen, line 52: Could we use something similar to thi approach for colons?

https://github.com/kumahq/kuma-gui/blob/698c2a75cd3a46f4cfb4363f8199dab77f5438b3/packages/kuma-gui/src/app/common/data-card/DataCard.vue#L59-L60

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
```diff
@@ -18,53 +18,57 @@
...
+              </template>
 
-                  <template v-else>
-                    {{ t('common.detail.none') }}
-                  </template>
+              <template #body>
+                <template v-if="data.addressPort">
+                  <XBadge appearance="decorative">
+                    <XCopyButton :text="data.addressPort" />
+                  </XBadge>
```
johncowen, line 52: I believe these shouldn't use `XBadge` at all. They should just use `XCopyButton` in order to achieve what we are trying to do here.

It would be good to add a `:variant="badge"` to `XCopyButton`. It already has a `format` but I think that refers to the format of the text you are about to copy, not the 'variant'/look and feel.

packages/kuma-gui/src/app/x/components/x-badge/XBadge.vue
```diff
@@ -14,3 +14,10 @@ const props = withDefaults(defineProps<{
   maxWidth: 'auto',
 })
 </script>
+
+<style lang="scss" scoped>
+:deep(a) {
+  text-decoration: underline;
+  color: currentColor;
+}
+</style>
```
johncowen, line 23: Is this the reason why we have underlines here now:

Screenshot 2024-12-17 at 12 40 38

Is that intentional?

schogges: Yes, I thought it would help the user to identify that there is a link behind. That was a quick decision and I know that it affects all the places. Lmk if this is completely off design 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
schogges: Oh, that's nice 🙂 thanks for the pointer 👍

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
johncowen: I also completely agree with what you are saying i.e:

>  I'd prefer the composable strategy, which separates concerns and allows more flexibility.

Thing is, if someone changes KCopy's to use a different type of badge (not appearance="decorative") then we may hit issues because our "AppAboutSection" will still use appearance="decorative" i.e. this purple look, while everywhere else it will use the "updated upstream look" maybe I dunno blue.

This my main concern, as from what I understand it goes against the reasonings behind this PR. I'm not sure if flexibility is what we are trying to achieve here.

Honestly I don't know what to suggest here myself as yet, still thinking.

packages/kuma-gui/src/app/x/components/x-badge/XBadge.vue
johncowen: Gotcha, this is definitely something I wouldn't add to this PR then, feels like it changes things all over the place instead of just in the places we are working on here. Umm, is this something you definitely need in this PR?

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
schogges: Aah 😅 I've checked and realized that `KCopy` has a prop `badge` already. I thought this was a custom implementation further downstream.
Now after I realized that, I think I'd go for the `:variant` to stay consistent with upstream changes.

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
schogges: Removed `XTimespan`. For future reference: https://github.com/kumahq/kuma-gui/pull/3300/commits/2d98182a1323b9bd1b627f6af11fa65e5e93176e

packages/kuma-gui/src/app/x/components/x-badge/XBadge.vue
schogges: Nope, we can discuss this separately 👍 Another option could be to add an icon to indicate the link. It's just that without any indication, as a user, I don't know that there is a link behind.

packages/kuma-gui/src/app/x/components/x-about-section/XAboutSection.vue
schogges: Ah, thanks 👍 I've removed it 🙂

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
schogges: I've added the `variant` prop for the `badge` appearance and applied it to the places in `XAboutSection`.

packages/kuma-gui/src/app/x/components/x-badge/XBadge.vue
schogges: Removed 🙂

johncowen:
> Yeah, this is tricky. I am not really sure what the best solution is. Ideally we can get rid of the mixture. At least for the mTLS one we could probably move it to the AboutSection?

Yeah I think that might be best, give it a try and see what we think? We can always revert. Not sure about the DataPlane one those, thats unfortunate, not quite sure 🤔

packages/kuma-gui/src/app/gateways/views/DelegatedGatewayDetailView.vue
```diff
@@ -12,50 +12,56 @@
...
-                :total="data.dataplanes?.total ?? 0"
-              >
-                <template #title>
-                  {{ t('http.api.property.dataPlaneProxies') }}
-                </template>
-              </ResourceStatus>
-            </div>
-          </KCard>
+            <ResourceStatus
+              layout="horizontal"
```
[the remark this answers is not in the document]
schogges: Hmmmm 🤔 I am little confused now 😅 because:

1. Why is this not catched?
2. How does it work still work?

Regarding 2.: The default value of `layout` in `DefinitionCard` is `vertical`. Under the hood `ResourceStatus` uses `DefinitionCard`, without setting `layout`, because the default is `vertical` (what we want for `ResourceStatus`). However, when I set `layout="horizontal"` on the `ResourceStatus` it's somehow being passed down to the child `DefintionCard` 🤔

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
schogges: Oh, sorry this got lost in my mind 😅

packages/kuma-gui/src/app/gateways/views/DelegatedGatewayDetailView.vue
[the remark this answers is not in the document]
schogges: I've just realized that I did the same with `XAboutSection`. I did not specify any prop, but since `AppAboutSection` is the top level component that wraps everything, I guess the props will be just passed further down there 🤔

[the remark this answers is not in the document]
johncowen: oh wait, sorry I didn't notice that `ResourceStatus` uses `DefinitionCard`

Because `DefinitionCard` is the first child of `ResourceStatus` it automatically gets another props pass to `ResourceStatus` passed down to it. SO this is why it still works and TS doesn't say anything.

I think this is still confusing though, I guess its a consequence of our other conversation about this horizontal style of ResourceStatus even being a thing.

Umm, not sure what to suggest here, if anything. 🤔 I think we can probably resolve this, tiny chance I might have another thought here, plus we might think of another way to do this (see other convo on the `Dataplanes: 0/20` thing)

packages/kuma-gui/src/app/x/index.ts
johncowen: I left a comment here about another potential name - `XAboutCard`, I think it coinciding with a GH blip tho. Not sure from your comment whether we are looking to rename or not, lmk

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
johncowen: Well to be fair I had in my mind that I did this in XTimespan, but apparently not! 😆 Ah now I remember 🤦 thanks for that link!

https://github.com/kumahq/kuma-gui/blob/484769bc4ff0113b81aff9cd0c8f0eb164923022/packages/kuma-gui/src/app/x/components/x-timespan/README.md?plain=1#L15-L16

Umm, I guess that makes me still on the fence with this one 😆 , I'll leave it with you if you want to change then fine or you want to leave as is, lets resolve. I think apart from this theres only one more left.

johncowen (commented):
Ok I think I've covered everything, there are a couple of outstanding things that we just want to cover off either way. But wanted to note that I don't think theres anything else here.

Appreciate I had quite a few comments and concerns, I was being extra careful considering everything. Also appreciate how you dealt with my comments with care and patience, so thanks! We've both said ideally we don't want to be coming back here again at any point soon, if at all, and I think this is in the best state we can get it now 🎉 

Just waiting to see if there's anything else to be done on the unresolved comments, then I can come back here with a 👍 I think

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
schogges: Ah 😄 anyways, I've reworked it and moved the formatting to the component.

packages/kuma-gui/src/app/gateways/views/DelegatedGatewayDetailView.vue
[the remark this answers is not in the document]
schogges: I'd resolve it then for now and revisit this once we discuss how we want to show the `ResourceDefinition` next to horizontal `DescriptionCards`.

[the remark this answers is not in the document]
johncowen: 😆 all these different name forms are making my head spin!

schogges:
> Ok I think I've covered everything, there are a couple of outstanding things that we just want to cover off either way. But wanted to note that I don't think theres anything else here.

Just pushed the last changes and resolved some conflicts. I think it's ready to go now 🙂 

> Appreciate I had quite a few comments and concerns, I was being extra careful considering everything. Also appreciate how you dealt with my comments with care and patience, so thanks! We've both said ideally we don't want to be coming back here again at any point soon, if at all, and I think this is in the best state we can get it now 🎉
> 
> Just waiting to see if there's anything else to be done on the unresolved comments, then I can come back here with a 👍 I think

Thanks a lot for your thoughts, comments and patience 🙂 this is much better now! 🙌

johncowen (approved):
Ok lets get this in!

Issue #2667, opened 2024-06-20
GlobalCP Health Card: Make numbers of things more clear/explanatory


johncowen:
### Description

In the below screenshot we use various colors for certain numbers, but its not super clear why.

Maybe we should use tooltips or similar to explain what these numbers are/why they are colored.

![Screenshot 2024-06-20 at 16 31 35](https://github.com/kumahq/kuma-gui/assets/554604/d7420966-60dc-4aa5-81d7-ebe7bb446299)

johncowen:
On re-reading this again, do we even want to use coloring here? Does it actually "mean" anything?

Looking at the code it looks to use a simple ratio of "if less than 50% are `online` use red, otherwise if less than 100% are `online` use yellow".

The thing is, who's to say that my cluster is only "in danger" if less than half of my services are online? Surely I might want to be "in danger" if any single service is not online?

Alternatively, if I have 5 services with 10 dataplanes/instances, I'm probably running perfectly fine if half of my dataplanes/instances are offline because each still has 5 instances available.

I'd almost be fine using (or something similar)

```
Zones             | Meshes | Services       | Data Plane Proxies
2 of 11 online    | 93     | 1 of 6 online  | 17 of 23 online
```

We have the space to do that here. 

Anyway mainly food for thought.

PR #3301, opened 2024-12-16
refactor(common/resource-status): prefer undefined over null

schogges:
Generally we want to avoid handling `null` anywhere else than the data layer. In the data layer we make sure to transform all values to something that we can safely access in the view layer (e.g. no conditional chaining required). As a last resort we prefer `undefined` over `null` if something is just not set.
To meet these requirements I've refactored this component. It's important to mention that we specifically need the case of `online` being `undefined`, because there are cases in which we only want to show the `total`.


packages/kuma-gui/src/app/common/ResourceStatus.vue
```diff
@@ -33,23 +31,23 @@ import DefinitionCard from './DefinitionCard.vue'
...
+  const ratio = props.online / props.total
+
+  if (ratio <= 0.5) {
+    return 'danger'
+  } else if (ratio < 1) {
+    return 'warning'
   }
 
-  return null
+  return undefined
```
johncowen, line 50: Feels like we could maybe make this a tiny bit better whilst we are here. Feels weird that we need an `else` in here 🤔 

I guess you want to keep your preferred style of exiting early? (I feel its a bit weird that we return `undefined` in two places now!).

Umm, if we make `ratio` a `max` of `1` would we find a compromise here? i.e. keep the exit early style, and remove the double `undefined`? I'm not totally sure myself btw. lmk

johncowen (commented):
Great improvement, thanks for the detailed description also 👍 

I had one comment, let me know what you think

packages/kuma-gui/src/app/common/ResourceStatus.vue
```diff
@@ -13,14 +13,12 @@
...
-          class="status-online"
-          :class="{ [`status-online--${statusAppearance}`]: statusAppearance !== null }"
-        >{{ props.online }}</span><span
-          v-if="props.online !== null"
-          class="status-separator"
-        >/</span><span class="status-total">{{ props.total }}</span>
+        <template v-if="typeof props.online !== 'undefined'">
+          <span
+            class="status-online"
+            :class="{ [`status-online--${statusAppearance}`]: statusAppearance !== null }"
```
johncowen, line 19: Ah, sorry! P.S. I just spotted there's a `null` here also, actually I have another idea 🤔 let me know on the [below](https://github.com/kumahq/kuma-gui/pull/3301/files#r1886686583) first in the meantime.

schogges: oh, great spot 😄 this check for `null` was actually meant to be a check for `undefined`. Went a little back and forth with this and forgot to commit that.

---

By also referring to your other comment above and the duplicate return of `undefined`, I'd like to explain that there are generally two cases here:
1. `online` is `undefined`, which leads to early exit, because we don't have to calculate the `ratio`.
2. The second return of `undefined` is meant to for the class modifier. The case for "all fine" is the default.

packages/kuma-gui/src/app/common/ResourceStatus.vue
johncowen: Explained below, resolving

packages/kuma-gui/src/app/common/ResourceStatus.vue
schogges: > But also means we can completely remove the Javascript in the bottom <script> tag (avoiding Loss of Locality), avoids fiddling with string | undefined aswell as avoiding with string | null, avoids checking for the undefined-ness of props.online twice because Vue+Typescript, avoids checking for existence of statusAppearance and finally because its all local/inline and things are only checked once, it would avoid the accidental oversight we spotted here.

Generally I like to simplify things. Since we still need a check in the template, we can use it as a guard for the rest. I like that 👍  But honestly, whenever I come across a `v-for` I still expect a loop and it takes a moment or two to realize that in these cases it's just one iteration. However it's a great workaround to make variables available further down the tree (at least that's my understanding).
I was playing around with a custom vue directive to make to possible without using `v-for`, but yet no success. I still want to look further into this, but not in the scope of this PR 🙂 

> I know this adds a nested ternary which we've spoken about before, but I like this solution as it shows all the benefits of sometimes using such constructs that "somebody somewhere on the internet" would probably tell you not to use and I think its a very good example of being sure to question such things depending on your specific situation. In the same respect if you notice something I've missed here do let me know and I'll eat my hat! 😆

The approach I now went for somewhere meets on the middleground 😄 I used the `v-for` directive and the binding for class, that only sets the class modifier when `ratio < 1`. That still uses a ternary, but no nested. I hope that is fine.

> which doesn't look quite right and I think would be a regression we've made in making sure we always show a fraction if there are two numbers. I think both numbers and the separator should be black I think, do you see what I mean? Its a bit weird that meshes is always black, and then for some reason even if all my dataplanes are online they are grey?

Yeah, you are right. I think we should sit together and talk about this in more detail.

> In spending time on this it reminded me of another issue we have, just to be clear IMO this issue is totally not in the scope of this PR. But linking here as its totally related, whilst I made the issue I'm pretty sure it was originally from @lahabana
>
> https://github.com/kumahq/kuma-gui/issues/2667
>
> At the very least its just to xref the two things together.

Ah, that's nice. Thanks for the pointer. This issue would be great opportunity to revisit and redesign the whole component and view 👍

johncowen: > But honestly, whenever I come across a v-for I still expect a loop and it takes a moment or two to realize that in these cases it's just one iteration. 
> I was playing around with a custom vue directive to make this possible without using v-for, but yet no success. I still want to look further into this, but not in the scope of this PR 🙂

I getcha, I don't totally like that we have to use a `v-for` for this really either, but the advantages of doing this far outweigh this little sharp edge, I guess its just a naming thing.

I also had a good look at seeing how straightforwards it was to add a `v-ref` or `v-assign` or something by more or less aliasing `v-for`. I found that `v-for` seemed to be very baked into the innards of Vue, almost like a magical directive (i.e. it can do stuff other directives can't). It was a while ago now and my Vue knowledge at the time was less than it is now so if you find a way to do it that would be awesome!

Back to today "it takes a moment or two to realize that in these cases it's just one iteration" I'd guess this probably happens at first? But then once you understand why its like this, and you use it/see it a couple of times this is no longer the case? You'd then understand and learn about it right? It just becomes natural.

---

> The approach I now went for somewhere meets on the middleground 😄 I used the v-for directive and the binding for class, that only sets the class modifier when ratio < 1. That still uses a ternary, but no nested. I hope that is fine, lmk 🙂

This looks much better to me! I much prefer what you've done here, I think that's much clearer as to what's happening and less checking for emptiness (granted, because of how TS+Vue works). Thanks!

The only thing I think is missing is the regression type-thing I mentioned about colouring, here's a a better example of what I'm talking about.

Screenshot 2024-12-17 at 09 40 28

In the above screenshot, notice Meshes uses black, Services uses black for its online but grey for its total. I think that if nothing is wrong with any services (i.e. online and total are the same) then it should also be in black. Then if everything is online in your setup, you aren't thinking "what are these colours for?" (as in https://github.com/kumahq/kuma-gui/issues/2667) , or "why is services greyed out?".

If everything is _not_ fine in your setup, then I suppose I kinda want you to be asking "what are these colours for?" 🤔

schogges: > > But honestly, whenever I come across a v-for I still expect a loop and it takes a moment or two to realize that in these cases it's just one iteration.
> > I was playing around with a custom vue directive to make this possible without using v-for, but yet no success. I still want to look further into this, but not in the scope of this PR 🙂
> 
> I getcha, I don't totally like that we have to use a `v-for` for this really either, but the advantages of doing this far outweigh this little sharp edge, I guess its just a naming thing.
> 
> I also had a good look at seeing how straightforwards it was to add a `v-ref` or `v-assign` or something by more or less aliasing `v-for`. I found that `v-for` seemed to be very baked into the innards of Vue, almost like a magical directive (i.e. it can do stuff other directives can't). It was a while ago now and my Vue knowledge at the time was less than it is now so if you find a way to do it that would be awesome!
> 
> Back to today "it takes a moment or two to realize that in these cases it's just one iteration" I'd guess this probably happens at first? But then once you understand why its like this, and you use it/see it a couple of times this is no longer the case? You'd then understand and learn about it right? It just becomes natural.

Yeah, I think the more I work with it, the more natural it becomes. However, I think I'd always need to double check what the loop is about. So, I'd like to spend some time to find a replacement for this, also to improve my knowledge about Vue. And I have to admit that `v-for` really feels magical 😄 

> The only thing I think is missing is the regression type-thing I mentioned about colouring, here's a a better example of what I'm talking about.
> 
> Screenshot 2024-12-17 at 09 40 28
> 
> In the above screenshot, notice Meshes uses black, Services uses black for its online but grey for its total. I think that if nothing is wrong with any services (i.e. online and total are the same) then it should also be in black. Then if everything is online in your setup, you aren't thinking "what are these colours for?" (as in #2667) , or "why is services greyed out?".
> 
> If everything is _not_ fine in your setup, then I suppose I kinda want you to be asking "what are these colours for?" 🤔

Yeah, I think it can be confusing with this coloring. So, I'll just remove the grey scale and make sure that `total` is always black.

schogges: Applied that change and `total` is the same as `online`

johncowen (approved):
thanks for this this one!

PR #3302, opened 2024-12-16
chore(deps): remove standard

schogges:
Since we are not using `standard` but `eslint` and `eslint-config-standard`, we can remove the dependency to `standard`.


johncowen (approved):
top spot! 👏

PR #3311, opened 2024-12-17
chore(config/eslint): add vue/no-console

schogges:
We recently noticed that `console` statements are not catched in `.vue` files inside the `template` (although catched in `script`) (xref: https://github.com/kumahq/kuma-gui/pull/3300#discussion_r1887136806).

### DIFF
--- a/packages/config/src/eslint.cjs
+++ b/packages/config/src/eslint.cjs
@@ -143,6 +143,7 @@ function createEslintConfig(
       {
         files: ['*.vue'],
         rules: {
+          'vue/no-console': 'error',
           'vue/attributes-order': ['error', {
             order: [
               'DEFINITION',
--- a/packages/kuma-gui/src/app/x/components/x-code-block/ResourceCodeBlock.vue
+++ b/packages/kuma-gui/src/app/x/components/x-code-block/ResourceCodeBlock.vue
@@ -36,7 +36,7 @@
                 if (expanded) {
                   toggle()
                 }
-                cb((text: Object) => copy(toYamlRepresentation(text)), (e: unknown) => console.error(e))
+                cb((text: Object) => copy(toYamlRepresentation(text)), onCopyReject)
               }"
               :copying="expanded"
             />
@@ -88,4 +88,6 @@ function toYamlRepresentation(resource: Object): string {
   }
   return YAML.stringify(resource)
 }
+
+const onCopyReject: Parameters<CopyCallback>[1] = (e: unknown) => console.error(e)
 </script>


johncowen:
🤔 you might be able to help me here, as I'm not totally sure of why this wasn't working in the first place, and I have a small curiosity off the back of that. Maybe I'm just not looking at this properly, let me know either way.

We have this rule already in the top-level rules, do you know why we have to add it again in the overrides? The follow up I'm thinking about is "are there any other rules we have to add twice to get them to work?"

🤔 Do you think its because in order to get eslint rules to work in Vue templates they _also_  need to be explicitly added to `*.vue` overrides? (i.e. its just a wrinkle in how the Vue eslint support works)

schogges:
> 🤔 you might be able to help me here, as I'm not totally sure of why this wasn't working in the first place, and I have a small curiosity off the back of that. Maybe I'm just not looking at this properly, let me know either way.
> 
> We have this rule already in the top-level rules, do you know why we have to add it again in the overrides? The follow up I'm thinking about is "are there any other rules we have to add twice to get them to work?"
> 
> 🤔 Do you think its because in order to get eslint rules to work in Vue templates they _also_ need to be explicitly added to `*.vue` overrides? (i.e. its just a wrinkle in how the Vue eslint support works)

I've played around with it and it seems that the `no-console` is only checking in scripts (`js|ts`-files, `<script>`-tags, ...), but no in templates 🤔 so my PR description is not 100% correct and I'll correct myself in a sec.
I've looked up the [rule description](https://eslint.vuejs.org/rules/no-console#rule-details) and it looks like my assumption is correct.

johncowen:
Ah great, thanks for the link:

> This rule is the same rule as core [no-console](https://eslint.org/docs/latest/rules/no-console) rule but it applies to the expressions in <template>.

That's what I was hoping ^. In this case I don't think we should allow any `console.*` usage in the template tags - the only one I'm only a tiny bit unsure about is `.error`, but as we haven't needed that in templates at all up until now. I'd be inclined to leave that out also. wdyt?

Just to share thinking again: My thinking is `console.` is a "development" console and ideally we'd have no "development" code in our production code at all. We do right now "for reasons", but only in script, ideally we should throw actual errors when there are errors not just `console.error` them out.

schogges:
> Ah great, thanks for the link:
> 
> > This rule is the same rule as core [no-console](https://eslint.org/docs/latest/rules/no-console) rule but it applies to the expressions in `<template>`.
> 
> That's what I was hoping ^. In this case I don't think we should allow any `console.*` usage in the template tags - the only one I'm only a tiny bit unsure about is `.error`, but as we haven't needed that in templates at all up until now. I'd be inclined to leave that out also. wdyt?

Actually there is [one case for error](https://github.com/kumahq/kuma-gui/blob/698c2a75cd3a46f4cfb4363f8199dab77f5438b3/packages/kuma-gui/src/app/x/components/x-code-block/ResourceCodeBlock.vue#L39). But definitely we can disallow `warn` and `info`.

> Just to share thinking again: My thinking is `console.` is a "development" console and ideally we'd have no "development" code in our production code at all. We do right now "for reasons", but only in script, ideally we should throw actual errors when there are errors not just `console.error` them out.

Fully agree 💯

johncowen:
Ah I see, what about dropping that last `console.error` into the script tags for now? That means we can blanket disallow any `console.*` in any templates. I guess there are no more in-template `console` apart from that one?

After looking at the above link (thanks for that!), I think the only reason the `console.error` is there is because it's overly-complicated to change the text of the tooltip to an error message when there is an error 🤦 . Ideally this it what we would do.

Short term tho shall we just drop the `console.error` into a callback in script tag and call that?

johncowen (approved):
🎉 Awesome thanks!

(also wanted to xref a past PR here https://github.com/kumahq/kuma-gui/pull/1140)

PR #3315, opened 2024-12-17
chore: remove forms related CSS

johncowen:
I've been looking over all our custom "global" CSS with an eye to reducing it as much as possible.

This `_forms.scss` seems to be unused here now. There is one selector that we still use in onboarding `.radio-group`, but the CSS rule in this removed file (`display: flex`) is also declared where we use it over in onboarding, so looks safe to remove.


schogges (approved):
👍

PR #3319, opened 2024-12-18
chore(docs): add name of mock disabling env var

johncowen:
Adds name of mock disabling env var/cookie to our docs so it remains visible/usable outside of our docs site (ie. in GH markdown)

Usage is here on the right:

395566909-13a477d7-0b0c-4a44-8209-ab558a5fd08f

I also made https://github.com/kumahq/kuma-gui/issues/3320


schogges (approved):
👍

PR #3323, opened 2024-12-19
fix: globally fix text color of buttons in alerts

johncowen:
Fixes text color of all primary buttons within alerts.

See https://github.com/Kong/kongponents/commit/84a4e9f75c7bff9b136ae3e89d7cc2be51dbcc3f 

I previously "fixed" this in https://github.com/kumahq/kuma-gui/pull/3317 just for our upgrade button, thinking this was the only instance 

This PR undoes that ^ PR and fixes it globally instead, which will fix the issue for all instances whether I remember about them or not.

### Before

Screenshot 2024-12-19 at 11 22 19

### After

Screenshot 2024-12-19 at 11 08 31


schogges (approved):
Nice catch 👍 also happy to see further adoption of `variant` 🙂

johncowen:
FYI I had/still have another branch which takes a completely different approach and makes it impossible for these things to happen at all (this PR only fixes primary buttons within a KAlert).

It's an interesting branch/piece of work, so might put it up anyway to share. Would be good to chat with someone about it whether we wanted to use it or not (I'm not totally convinced yet myself).

johncowen:
Ah, I found out how to access existence of `@event`s a couple of weeks back and it was something I wanted to use here but couldn't figure out how at the time (events are also present on `attrs` prefixed with `on` is seems). I added a commit here also so that just by adding a `@dismiss` handler it makes the alert `:dismissable`.

Will leave this hanging a little while longer here in case, but will merge later today if I hear no more.

schogges:
> Ah, I found out how to access existence of `@event`s a couple of weeks back and it was something I wanted to use here but couldn't figure out how at the time (events are also present on `attrs` prefixed with `on` is seems). I added a commit here also so that just by adding a `@dismiss` handler it makes the alert `:dismissable`.
> 
> Will leave this hanging a little while longer here in case, but will merge later today if I hear no more.

Ah, yeah that is good to know 👍

Issue #2977, opened 2024-09-19
ZoneEgress view using Envoy data


lahabana:
### Description

Similar to #2976 but for ZoneEgress

PR #3325, opened 2024-12-19
feature(zones): adds a traffic view for zone egress

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/2977

Related:

- https://github.com/kumahq/kuma-gui/pull/3296

Screenshot 2024-12-19 at 12 35 20

### DIFF
--- a/packages/kuma-gui/src/app/connections/locales/en-us/index.yaml
+++ b/packages/kuma-gui/src/app/connections/locales/en-us/index.yaml
@@ -1,3 +1,9 @@
+inbounds:
+  x-empty-state:
+    title: No inbounds
+outbounds:
+  x-empty-state:
+    title: No active outbounds
 connections:
   include_endpoints: Include Endpoints
   routes:
--- a/packages/kuma-gui/src/app/connections/locales/en-us/index.yaml
+++ b/packages/kuma-gui/src/app/connections/locales/en-us/index.yaml
@@ -3,6 +3,9 @@ inbounds:
     title: No inbounds
 outbounds:
   x-empty-state:
+    title: No outbounds
+activeOutbounds:
+  x-empty-state:
     title: No active outbounds
 connections:
   include_endpoints: Include Endpoints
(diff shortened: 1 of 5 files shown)


schogges (approved):
👍

schogges:
Oh, just noticed that `No data` is not centered

Screenshot 2025-01-06 at 12 17 16

johncowen:
Oh! Thanks! I vaguely remember seeing that also but forgot about it! Thanks for the nudge, lemme add something here.

johncowen:
I did it like this in the end seeing as you can have an empty state 'per column'

Screenshot 2025-01-06 at 12 11 54

Screenshot 2025-01-06 at 12 11 59

packages/kuma-gui/src/app/connections/locales/en-us/index.yaml
```diff
@@ -1,3 +1,9 @@
+inbounds:
+  x-empty-state:
+    title: No inbounds
+outbounds:
+  x-empty-state:
+    title: No active outbounds
```
schogges, line 6: Just to confirm, there are always outbounds? At least inactive ones?

schogges (commented):
Just a Q

packages/kuma-gui/src/app/connections/locales/en-us/index.yaml
johncowen: From what I understand, mostly there are always outbounds. I wouldn't be able to 100% confirm that though. I can remove the `active` word maybe to make it more general?

johncowen: Ok so I ended up with this:

Screenshot 2025-01-06 at 13 35 50

Screenshot 2025-01-06 at 13 36 01

lmk!

schogges: LGTM 🙂

schogges (approved):
👍

PR #3326, opened 2024-12-19
fix: Fixes an issue where DataCollection can cause extra elements

johncowen:
TBD

### DIFF
--- a/packages/kuma-gui/src/app/application/components/data-collection/DataCollection.vue
+++ b/packages/kuma-gui/src/app/application/components/data-collection/DataCollection.vue
@@ -1,53 +1,53 @@
 <template>
-  <XLayout
-    type="stack"
+  <template
+    v-if="slots.item"
   >
     <template
-      v-if="slots.item"
+      v-for="item in [props.items.find(props.predicate)]"
+      :key="item"
     >
-      <div>
-        <template
-          v-for="item in [props.items.find(props.predicate)]"
-          :key="item"
-        >
-          <slot
-            v-if="item"
-            name="item"
-            :item="item as T"
-          />
-          <slot
-            v-else
-            name="empty"
-            :items="items"
-          >
-            <XEmptyState
-              v-if="props.empty"
-              :type="props.type"
-            />
-          </slot>
-        </template>
-      </div>
-    </template>
-    <template
-      v-else
-    >
-      <div>
-        <slot
-          v-if="items.length === 0"
-          name="empty"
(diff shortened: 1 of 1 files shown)


packages/kuma-gui/src/app/application/components/data-collection/DataCollection.vue
```diff
@@ -1,53 +1,53 @@
...
-              v-if="props.empty"
-              :type="props.type"
-            />
-          </slot>
-        </template>
-      </div>
+      <slot
+        v-if="item"
+        name="item"
+        :item="item as T"
```
schogges, line 12: I don't think we need `as T` here? `v-if` already rules out `undefined` AFAICS. I know this got only highlighted due to surrounding changes, but still wanted to mention it 🙂

schogges (approved):
Just that nit above, but I want to unblock this anyways as it's just a super minor nit 🙂

packages/kuma-gui/src/app/application/components/data-collection/DataCollection.vue
johncowen: 🤔 yeah not totally sure, let me remove and see what happens. brb

johncowen: I looked at this bit of code to see what happens without the `T` cast:

https://github.com/kumahq/kuma-gui/blob/37a68b227fd9a957d22b7384d7cf786eae51d468/packages/kuma-gui/src/app/subscriptions/views/SubscriptionSummaryView.vue#L13-L15

With `T`:

Screenshot 2025-01-06 at 11 38 39

Without `T`:

Screenshot 2025-01-06 at 11 38 30

All in all I think it results in the same, just using `T` make it look less weird.

wdyt? I'd be inclined to leave it as is, unless you can see a way where the casting might trip us up.

schogges: I think in this case I'd be fine either way 🙂 these casts just always pop to my eye as I generally try to avoid them.

johncowen: Gotcha, yeah totally usually try and avoid.

I think this is one of the times when its ever so slightly useful, plus I don't think it would ever trip us up. We are pretty much casting it to what it already is, just so happens that Vue's TS plugin means this way we get a slightly less weird output.

PR #3327, opened 2024-12-19
chore(mocks): improves mocking for dataplane.networking

johncowen:
We've had a couple of minor CSS bugs lately that have arisen because the randomized mocks weren't ever giving us certain states (generally we try to make the mocks return all types or data randomly so we see things like these bugs happen during development).

See:

- https://github.com/kumahq/kuma-gui/pull/3298
- https://github.com/kumahq/kuma-gui/pull/3326

This improves the mocks by:

- Further removing usage of the `.dataplaneNetworking()` function, that does too much and hides things away from where you are (i.e. Loss of Locality) This also makes it easier to use synced state across multiple HTTP requests using the faker seed.
- Adds missing variations of data so the variation that caused these issue becomes noticeable.

### DIFF
--- a/packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_overview.ts
+++ b/packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_overview.ts
@@ -8,7 +8,9 @@ export default ({ fake, pager, env }: EndpointDependencies): MockResponder => (r
   const _tags = query.get('tag') ?? ''
 
   const k8s = env('KUMA_ENVIRONMENT', 'universal') === 'kubernetes'
-  const inbounds = parseInt(env('KUMA_DATAPLANEINBOUND_COUNT', `${fake.number.int({ min: 1, max: 5 })}`))
+  const defaultType = env('KUMA_DATAPLANE_TYPE', '')
+  const inboundCount = parseInt(env('KUMA_DATAPLANEINBOUND_COUNT', `${fake.number.int({ min: 1, max: 5 })}`))
+  const outboundCount = parseInt(env('KUMA_DATAPLANEOUTBOUND_COUNT', `${fake.number.int({ min: 1, max: 10 })}`))
   const subscriptionCount = parseInt(env('KUMA_SUBSCRIPTION_COUNT', `${fake.number.int({ min: 1, max: 10 })}`))
   const isMtlsEnabledOverride = env('KUMA_MTLS_ENABLED', '')
   const { offset, total, next, pageTotal } = pager(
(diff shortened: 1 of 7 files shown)


packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_overview.ts
```diff
@@ -23,20 +25,23 @@ export default ({ fake, pager, env }: EndpointDependencies): MockResponder => (r
...
+      case _gateway === 'builtin':
+        return 'BUILTIN'
+      case defaultType === 'delegated':
+      case _gateway === 'delegated':
+      case tags['kuma.io/service']?.includes('-delegated'):
+        return 'DELEGATED'
+      case tags['kuma.io/service']?.includes('-builtin'):
+        return 'BUILTIN'
+      default:
+        return ''
```
schogges, line 42: Just reordering a bit to have one return of `BUILTIN` 🙂 

```suggestion
  const filterType = (() => {
    switch (true) {
      case _gateway === 'false':
        return 'STANDARD'
      case defaultType === 'builtin':
      case _gateway === 'builtin':
      case tags['kuma.io/service']?.includes('-builtin'):
        return 'BUILTIN'
      case defaultType === 'delegated':
      case _gateway === 'delegated':
      case tags['kuma.io/service']?.includes('-delegated'):
        return 'DELEGATED'
      default:
        return ''
```

schogges: The case `_gateway === 'false'` used to return `'proxy'`. Just want to verify that we are going with `'STANDARD'` now?

johncowen: > Just reordering a bit to have one return of BUILTIN 🙂

This means the tags containing `-builtin` is more important than the actual type being `DELEGATED`. This was the reason why it's in the order that it is. I don't think its massively important, but I wanted to keep the previous functionality (which also had this priority/ordering)

> The case _gateway === 'false' used to return 'proxy'. Just want to verify that we are going with 'STANDARD' now?

Yep, going with `STANDARD` so its more similar to `DELEGATED` and `BUILTIN`, from what I remember we don't actually print out `STANDARD` anywhere we just follow the existing pattern of uppercasing. i.e. both the string `proxy` and `STANDARD` aren't anything in the backend (whereas `DELEGATED` and `BUILTIN` are).

schogges: > > Just reordering a bit to have one return of BUILTIN 🙂
> 
> This means the tags containing `-builtin` is more important than the actual type being `DELEGATED`. This was the reason why it's in the order that it is. I don't think its massively important, but I wanted to keep the previous functionality (which also had this priority/ordering)

Ah, gotcha 😄 nvm then

> > The case _gateway === 'false' used to return 'proxy'. Just want to verify that we are going with 'STANDARD' now?
> 
> Yep, going with `STANDARD` so its more similar to `DELEGATED` and `BUILTIN`, from what I remember we don't actually print out `STANDARD` anywhere we just follow the existing pattern of uppercasing. i.e. both the string `proxy` and `STANDARD` aren't anything in the backend (whereas `DELEGATED` and `BUILTIN` are).

Alright, thanks for the explanation 👍

schogges (approved):
👍

PR #3329, opened 2024-12-23
chore(dataplanes): add bundle downloading functionality

johncowen:
Adds functionality to allow downloading of a bundle of config/logs for a Dataplane.

Closes https://github.com/kumahq/kuma-gui/issues/3263

Screenshot 2024-12-23 at 15 09 54

~The PR currently works, but I feel I might tweak the code slightly, wanted to put it up incase anyone wanted to try it out.~

~Its also waiting on https://github.com/kumahq/kuma-gui/pull/3330~

Lastly, policy downloading isn't added here as yet, I added an issue to cover that separately https://github.com/kumahq/kuma-gui/issues/3356

PR #3330, opened 2024-12-23
feature: adds x-progress component

johncowen:
Replaces our old `LoadingBlock` component (and several others) with a single `XProgress` component with several `:variant`s.

I'll be using this for https://github.com/kumahq/kuma-gui/pull/3329


packages/kuma-gui/src/app/x/components/x-progress/XProgress.vue
```diff
@@ -0,0 +1,43 @@
...
+  />
+  <KSkeleton
+    v-else-if="props.variant === 'list'"
+    data-testid="list-skeleton"
+    type="table"
+  />
+  <KSkeletonBox
+    v-else-if="props.variant === 'line'"
+    data-testid="line-skeleton"
+    width="10"
```
schogges, line 32: Would it make sense to make `width` configurable? Or do we always want to go with `10`?

schogges (commented):
Just a Q 🙂

packages/kuma-gui/src/app/x/components/x-progress/XProgress.vue
johncowen: I took this from the only place we use a KSkeletonBox where it was `10` (I think its supposed to be a percentage of its parent, but I'm not sure it works like that in reality).

If we wanted to make this more configurable, I think I'd set it to 100% and then require a wrapper of some sort to set the width.

🤔 I suppose its not something I want to mess with right now, we only use this in one place right now, so we might even get rid of it. I'd say in the future if we need to use it more we should consider how we make it configurable then, but I think I'd avoid giving this component a width prop seeing as it doesn't make sense for the other loader components in here, yet at least.

I suppose the tldr is, lets leave it as `10` for now (seeing as that was how it was) and if we need to change in future then lets do that, but at that point (if it arises) lets consider the 100% thing (even if its not the route we go). Not very TLDR 😆 sorry!

schogges: Alright, makes sense 👍 🙂

schogges (approved):

PR #3333, opened 2024-12-23
chore: merges KTruncate into XLayout

johncowen:
We recently found out that `KTruncate` also applies a margin/gap the same as `XLayout` (and we want to keep the same value for gap whether we are using `KTruncate` or not).

Therefore I felt it made sense to combine `KTruncate` with `XLayout` via a `:truncate` property.

Therefore whether you have elements require truncation or not, you use the same component and they will have the same gap spacing


schogges (approved):
👍

PR #3354, opened 2025-01-06
chore: removes TextWithCopyButton

johncowen:
We made `XCopyButton` a while back which was to replace our multiple copy button type components.

I've replaced all usages of `TextWithCopyButton` with our now only copy button component `XCopyButton`


schogges (approved):
👍

PR #3373, opened 2025-01-07
chore: uprade eslint to v9

### LINKED ISSUE #2879: upgrade eslint to v9

schogges:
This change is going to upgrade eslint to the current latest version of `9.17.0` and related modules to their latest version. This is the first part and more efforts will follow to further simplify and cleanup the config in favor of recommended rules.
I also took the chance to move eslint related modules to the dependency list of `@kumhq/config` (`@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`).

- Some `@typescript-eslint` rules have been moved to `@stylistic` to move maintenance efforts to the community (e.g. `@stylistic/indent`)
- I've started to group related configs:
  - `@vue/eslint-config-typescript` enables [the essential rules for Vue 3](https://eslint.vuejs.org/rules/#priority-a-essential-error-prevention) and [the recommended rules for TypeScript](https://typescript-eslint.io/rules/?=recommended). So there is no need to add `@typescript-eslint` rules separately. We are using `flat/recommended` ruleset, which includes `base`, `essential` and `strongly-recommended` rulesets.
  - `eslint-plugin-import` recommended rules
  - `eslint-plugin-json-schema-validator` recommended rules
  - `@stylistic/eslint-plugin` recommended stylistic rules and custom config
  - custom rules that we can revisit and decide to keep or remove them

I've compared the rulesets and although we already have most of the rules applied, it looks like by adding the recommended rules we are more strict.

---

I had to disable certain rules to stay in the scope of this PR. The issues will be fixed separately. The rules are the following
- `@typescript-eslint/no-explicit-any`
- `@typescript-eslint/ban-ts-comment`
- `@typescript-eslint/no-empty-object-type`

---

Initially I wanted to move our config files over to esm, but reverted that for now. The reason is that [`eslint.config.ts` (TypeScript) is not fully supported yet](https://eslint.org/docs/latest/use/configure/configuration-files#typescript-configuration-files) and we would have to stay with `eslint.config.js`. As our config module entry file is a `.ts`file, it can't be used for imports in a `.js` file. Therefore I added eslint support for `.cjs` files.

Closes #2879


packages/config/package.json
```diff
@@ -33,7 +35,7 @@
     "cypress": "^13.15.2",
     "cypress-fail-fast": "^7.1.1",
     "cypress-terminal-report": "^7.0.4",
-    "eslint": "^8.57.1",
+    "eslint": ">=9",
```
johncowen, line 38: Lets consider using `^9` which I think covers from 9 up until 10 (not including 10)

packages/config/src/eslint.config.cjs
johncowen, line 1: Should we not use `.config.` here. Origanally I considered using `.config.` but decided not to because I didn't want them to look like an actual "root config file" and look like a dependency instead.

packages/config/package.json
schogges: Yes, makes sense 👍

packages/config/src/eslint.config.cjs
schogges: Good point 🙂 👍

packages/kuma-gui/src/app/application/services/data-source/index.ts
```diff
@@ -75,6 +75,7 @@ type Hideable = EventTarget & { hidden: boolean }
 export const getSource = (doc: Hideable) => {
   return (cb: (source: RetryingEventSource) => Promise<unknown>, config: Configuration = {}) => {
     return new CallableEventSource<Configuration>(async function * (this: RetryingEventSource) {
+      // eslint-disable-next-line @typescript-eslint/no-this-alias
       const self = this
```
schogges, line 79: Worth to mention this one here. I wasn't totally sure what the reasoning was here to alias `this` or why the param is named `this`.

>Assigning a variable to this instead of properly using arrow lambdas may be a symptom of pre-ES6 practices or not managing scope well.
It's a generator function, so we have to keep the function syntax.

packages/kuma-gui/src/app/data-planes/data/SidecarDataplane.ts
```diff
@@ -71,12 +71,16 @@ function getPolicyTypeEntryConnections(policy: MatchedPolicyType, sidecarDatapla
...
+      // @ts-expect-error type mismatch config and policy.conf: {} vs string. PolicyTypeEntryConnection expects config as string | undefined
       const connection: PolicyTypeEntryConnection = { sourceTags, destinationTags, name, config, origins }
 
       policyTypeEntryConnections.push(connection)
     }
   } else {
     const sourceTags: LabelValue[] = []
+    // FIXME: below ts error
+    // @ts-expect-error type mismatch config and policy.conf: {} vs string. PolicyTypeEntryConnection expects config as string | undefined
     const connection: PolicyTypeEntryConnection = { sourceTags, destinationTags, name, config, origins }
```
schogges, line 84: Note these. Will have to fix them separately 🙂

packages/kuma-gui/src/app/application/services/data-source/index.ts
```diff
@@ -75,6 +75,7 @@ type Hideable = EventTarget & { hidden: boolean }
...
       let attempts = 0
       let iterations = 0
```
schogges, line 81: Worth to mention this one here. I wasn't totally sure what the reasoning was here to alias `this` or why the param is named `this`.

>Assigning a variable to this instead of properly using arrow lambdas may be a symptom of pre-ES6 practices or not managing scope well.

It's a generator function, so we have to keep the function syntax.

johncowen: From what I remember back in the mists of time it had to be added due to some sort of overzealous TS/Lint rule, I blamed it back and found this:

https://github.com/kumahq/kuma-gui/pull/1634/files#r1376243481

Love Github 😆 

I'm guessing we no longer have the node rule running on our browser code so I guess we can just remove this and refer to `this` instead.

As this is deep in app code tho, I would probably keep the ignore here and do this in a separate PR. Might sound a bit over cautious but would be good for keeping non-config code changes out of this PR.

packages/kuma-gui/src/app/x/components/x-tabs/XTabs.vue
```diff
@@ -67,7 +67,10 @@ const rewrite = () => {
   const $tabindexed = $el.querySelectorAll('[tabindex="0"]')
   const $id = $el.querySelectorAll('[id]')
 
-  $tablist && ['role', 'aria-label'].forEach(attr => $tablist.removeAttribute(attr))
+  if($tablist) {
+    ['role', 'aria-label'].forEach(attr => $tablist.removeAttribute(attr))
+  }
+
```
johncowen, line 73: Same as above, looks like:

```
'@typescript-eslint/no-unused-expressions': ['error', {
  allowTernary: true,
  allowShortCircuit: true
}],
```

would mean we wouldn't have to change non-config code in this PR.

packages/kuma-gui/src/types/index.d.ts
```diff
@@ -488,7 +488,7 @@ export interface MatchedPolicyType extends MeshEntity {
   sources?: PolicyMatch[]
   destinations?: PolicyMatch[]
   selectors?: Array<{ match: Record<string, string> }>
-  conf?: any
+  conf?: unknown
```
johncowen, line 491: bit of a nit but: Did we disable `any` in this PR? I remember there were a load more so I guess not. If not we should probably/maybe not do these in this PR.

Two things:

- I think we use `Record<string, unknown>` for `conf` elsewhere, I saw there's even a `Conf` type.
- I'm not totally sure that `KDSSubscription.version` is the same as `Version` 99.99% sure it is looking at the code now, but still I'm thinking we should maybe do that in a separate PR if poss.

johncowen (commented):
I left quite a few questions/comments, I think otherwise we are good to go here though.

Lemme know on the q's!

packages/kuma-gui/src/app/application/services/data-source/index.ts
schogges: Ah, I see 👍 thanks for tracing this down 😃 
Will add a TODO and resolve it for now.

packages/config/package.json
schogges: Done 🙂

packages/config/src/eslint.config.cjs
schogges: Done 🙂

packages/config/src/eslint.cjs
```diff
@@ -1,8 +1,14 @@
...
+const vuePlugin = require('eslint-plugin-vue')
+const { defineConfig, createConfig: vueTsEslintConfig } = require('@vue/eslint-config-typescript')
+const importPlugin = require('eslint-plugin-import')
+const stylistic = require('@stylistic/eslint-plugin')
+const eslint = require('@eslint/js')
+const jsonSchemaValidatorPlugin = require('eslint-plugin-json-schema-validator')
+const globals = require('globals')
 
-const packageSchema = JSON.parse(read(`${$config}/package.schema.json`).toString())
+const packageSchema = JSON.parse(read(resolve(__dirname, `./package.schema.json`)).toString())
```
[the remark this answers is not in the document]
schogges: But I'm wondering if it makes a difference 🤔 both resolve to the same path regardless of cwd or am I completely wrong? 😅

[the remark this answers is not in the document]
johncowen: Yeah I _think_ they always would return the same path, just I'm planning on using the `const $config = dirname(require.resolve('@kumahq/config'))` construct in other places and I was hoping to use the same construct everywhere.

FYI:

I hope to use this in our Makefiles to find out where `@kumahq/config` is i.e. `node -e 'console.log(require("path").dirname(require.resolve("@kumahq/config")))'` so wanted to use the same approach everywhere we need to ask the question "where is @kumahq/config?". `__dirname` won't work everywhere, especially not if there is no file being executed.

It's a tiny thing, but I'd rather use the same approach everywhere if poss. Plus (even tinier) `__dirname` is a little magical

packages/config/src/eslint.cjs
```diff
@@ -137,133 +101,213 @@ function createEslintConfig(
...
+      rules: {
+        '@typescript-eslint/no-require-imports': 'off'
+      }
+    }
+  ]
 }
 
 module.exports = {
-  createEslintConfig,
+  createEslintConfig
```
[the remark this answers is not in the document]
schogges: Nice catch 👍 Regarding `eslint` the rule has been deprecated with `eslint@8.53.0` [(docs)](https://eslint.org/docs/latest/rules/comma-dangle) and the rule is part of `@stylistic`. Same for the quotes. Will add them to the stylistic config.

This has not been catched because we are not linting our config package 🤔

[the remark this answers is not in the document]
johncowen: > This has not been catched because we are not linting our config package 🤔

Oh of course 🤦 , when I moved to the workspace I kept things as is, i.e. the linting only ran on `/packages/kuma-gui`, I guess we need to add a Makefile here also 🤔 . Might take a bit more thinking tho, have a look and see what you think. If its problematic, I'd be fine skipping this and making an issue to cover, I think it could turn into a bigger thing. Lemme know what you think.

packages/config/src/stylelint.cjs
```diff
@@ -1,26 +1,26 @@
...
-        'no-duplicate-selectors': [null],
-      },
-    }
-  }
+          severity: "error",
+        },
+      ],
+      "no-duplicate-selectors": [null],
+    },
+  };
```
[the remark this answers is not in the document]
johncowen: This will be the same as the above, i.e. we aren't running a linter here yet.

packages/config/src/eslint.cjs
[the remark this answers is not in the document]
schogges: Looks like this was not consistent. I added `@stylistic/quotes` and `@stylistic/semi` and this needed to be fixed in several places 🤔

packages/kuma-gui/eslint.config.cjs
```diff
@@ -0,0 +1,18 @@
...
+
+const config = [
+  ...eslint({}),
+  {
+    ignores: [
+      'dist/*',
+      'node_modules/*',
+      'src/types/auto-generated.d.ts',
+      'public/mockServiceWorker.js',
+      '**/.*'
```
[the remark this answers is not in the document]
johncowen: This will be the same as the above, i.e. we aren't running a linter here yet.

packages/config/src/stylelint.cjs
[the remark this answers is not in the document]
schogges: Let's move the discussion about semi and quotes to the trailing commas discussion above. It's all related to `@stylistic` 🙂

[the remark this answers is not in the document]
johncowen: Sounds good 👍

packages/kuma-gui/eslint.config.cjs
```diff
@@ -0,0 +1,18 @@
...
+    ]
```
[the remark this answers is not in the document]
schogges: > Should these (or at least some of these) be part of a shared config? I feel like these are common ignores that would be shared between apps.

Initially I thought that ignoring files and folders is more a project specific thing. But I think we can move at least `dist` and `node_modules` to the shared config 🙂 

> Also could you explain the last entry, is that to ignore all dot files? If so would that include a .name.js file? (ie. a javascript dotfile)

Ah, I added that because I got a lot of issues reported from `.vitepress/cache`. I'd leave `.vitepress` for now as it would fix a lot of issues in there (not only in cache). In a next step I'd scope it to only ignore `.vitepress/cache`. Wdyt?

packages/kuma-gui/src/app/kuma/services/kuma-api/makeRequest.spec.ts
```diff
@@ -94,7 +94,7 @@ describe('makeRequest', () => {
     [
       function () {
         // We specifically want to test this edge case.
-        // eslint-disable-next-line prefer-promise-reject-errors
```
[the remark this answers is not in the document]
schogges: It seems that it's not part of the recommended rules (anymore) 🤔 
I've added it back 👍

packages/kuma-gui/src/app/common/ErrorBlock.spec.ts
```diff
@@ -86,12 +86,12 @@ describe('ErrorBlock.vue', () => {
         invalidParameters: [
           {
             field: 'name',
-            reason: "invalid characters. Valid characters are numbers, lowercase latin letters and '-', '_' symbols.",
+            reason: 'invalid characters. Valid characters are numbers, lowercase latin letters and \'-\', \'_\' symbols.',
```
johncowen, line 89: Ugh this is unfortunate (sorry, I know this was off the back of another comment of mine!)

Is there some sort of option for only enforcing single quotes when it doesn't lead to awkward escaping like this? I vaguely feel like I've seen that somewhere for this sort of lint rule?

packages/kuma-gui/src/types/index.d.ts
```diff
@@ -822,7 +822,7 @@ export interface TlsConfDataSource {
   inlineString?: string
 }
 
-export interface TlsConfOptions { }
+export type TlsConfOptions = Record<string, unknown>
```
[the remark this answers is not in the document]
schogges: Yes, it's related to `@typescript-eslint/no-empty-object-type`. But since I've disabled this rule for now I'll roll it back and we deal with it as a separate step 🙂

packages/kuma-gui/src/services/utils.ts
```diff
@@ -1,11 +1,11 @@
...
-  Token,
-  TokenType,
-  TokenValue,
-  RequiredToken,
-  OptionalToken,
+  type Token,
+  type TokenType,
+  type TokenValue,
+  type RequiredToken,
+  type OptionalToken,
```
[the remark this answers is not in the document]
schogges: Done 🙂

packages/kuma-gui/eslint.config.cjs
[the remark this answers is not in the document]
johncowen: Sounds fine to me! I think we can resolve this one

packages/kuma-gui/eslint.config.cjs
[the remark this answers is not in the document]
schogges: Should work now after adding `@stylistic/comma-dangle` 🙂

packages/kuma-gui/src/app/common/ErrorBlock.spec.ts
schogges: Ah 🙈  yes, I've added `avoidEscape: true`

schogges: Done 🙂

packages/kuma-gui/src/app/application/services/data-source/DataSourcePool.spec.ts
```diff
@@ -56,7 +56,8 @@ describe('DataSourcePool', () => {
     // acquire 10 times, with the last one being a different URI
     // they should all be the same instance apart from the last one
     Array.from({ length: 10 }).map((_, i) => i !== 9 ? data.source('/one') : data.source('/two')).reduce((prev, item, i) => {
-      i !== 9 ? expect(prev === item).toBe(true) : expect(prev === item).toBe(false)
+      const isSame = i !== 9
+      expect(prev === item).toBe(isSame)
```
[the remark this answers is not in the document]
schogges: Yes, makes sense 👍 🙂

packages/kuma-gui/src/types/index.d.ts
schogges: Yes, this is a leftover from the short time before we said to allow any for now. I rolled it back and this also resolved the ts errors I added an expect for. So we are going to deal with this later 👍

johncowen: Oh! Gotcha, wow I did wonder how this was changed with no consequences, sounds good. Glad we are looking at these little app code changes separately.

packages/kuma-gui/src/app/me/sources.ts
```diff
@@ -32,7 +32,7 @@ export const sources = (prefix: string = 'me', storage: Storage = window.localSt
     },
     '/me/:route/:data': async (params) => {
       const json = JSON.parse(params.data)
-      const res = merge(await get(params.route), json)
+      const res = merge<Record<string, unknown>>(await get(params.route), json)
```
johncowen, line 35: Super nit pick this one, wondering if this should be `object` also? Kinda weird that everything else here deals with `object` (apart from JSON.parse's `any`) and just this one thing uses `Record<string, unknown>`.

I think everything we pass in here happens to be up being more or less a `Record<string, unknown>`, but I think its probably better to have everything in here all use the same type.

Again, super nit pick, but it we are editing types for absolute correctness then I guess we should tweak this also, especially seeing as it didn't have a type associated with it before.

---

I guess this comment is "these types should all be the same" rather than "this one should use `<specific-type>`", so whichever is easier to switch so they all use the same type.

johncowen (commented):
1 last comment which I don't mind too much about, lmk and then I can give this a 👍 either way.

Thanks for the work here, especially liked the detailed description 👏

johncowen (approved):
LGTM!

packages/kuma-gui/src/app/me/sources.ts
schogges: Yeah, I think we can go with `object` also, although I think `Record<string, unknown>` is generally better for objects as functions and arrays can also be assigned to type `object`.

PR #3375, opened 2025-01-07
chore(ci): use GITHUB_REF_NAME for dispatch branch

johncowen:
Follow up of follow up of https://github.com/kumahq/kuma-gui/pull/3355

This is the same approach that kuma uses https://github.com/kumahq/kuma/blob/67912cfb66806a5157bf03122f60e82f8c6dc2fd/.github/workflows/pr-merged.yaml#L22

See:

- https://github.com/kumahq/kuma-gui/pull/3374
- https://github.com/kumahq/kuma-gui/pull/3355


schogges (approved):
LGTM 👍

PR #3378, opened 2025-01-08
chore(dataplanes): improve bundle downloading checkboxes

johncowen:
Few tiny UX improvements to the "Download Bundle" modal (see https://github.com/kumahq/kuma-gui/pull/3329):

- Disable the download button until at least one checkbox is clicked.
- Only allow checking "Include EDS" is XDS is also checked.
- Close modal when downloading

Best viewed with GH's "Hide Whitespace" option enabled for the diff


schogges (approved):
LGTM 👍

Issue #3356, opened 2025-01-06
Add policies rules to the Download Bundle modal


johncowen:
### Description

Continuation of https://github.com/kumahq/kuma-gui/pull/3329 to finish up https://github.com/kumahq/kuma-gui/issues/3263

PR #3380, opened 2025-01-08
feat(dataplanes): include rules in dataplane bundle download

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/3356

Continuation of https://github.com/kumahq/kuma-gui/pull/3329

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
+++ b/packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
@@ -60,7 +60,7 @@ data-planes:
           xds: XDS Configuration
           eds: Include EDS
           dataplane: Dataplane Configuration
-          policies: Poilicies
+          policies: Policies
           clusters: Envoy Clusters Logs
           stats: Envoy Stats Logs
       about:
--- a/packages/kuma-gui/src/app/data-planes/sources.ts
+++ b/packages/kuma-gui/src/app/data-planes/sources.ts
@@ -1,4 +1,5 @@
 import { TarWriter } from '@gera2ld/tarjs'
+import createClient from 'openapi-fetch'
 
 import {
   Dataplane,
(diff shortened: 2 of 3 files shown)


packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
```diff
@@ -60,7 +60,7 @@ data-planes:
           xds: XDS Configuration
           eds: Include EDS
           dataplane: Dataplane Configuration
-          policies: Poilicies
+          policies: Policies
```
schogges, line 63: Seeing this and the fact I missed that during last review, I'd like to ask what you think about a spell checker? I find it useful, but it's also another thing to maintain especially because we have a lot of custom words that we'd have to ignore. But still wanted to hear your opinion 🙂

schogges (approved):
LGTM 🙂

packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
johncowen: I think on a project like ours with very minimal textual copy and very few changes to that very minimal copy, its a bit unnecessary. For something like a docs site then maybe its more useful.

Umm, I'd probably ask, how many times have we had an issue with a incorrect spelling? Considering this one was commented out until now, it's not really one we can count. (I would guess that, because it was commented out, this is the reason why neither of us spotted this one in the last PR. As soon as I uncommented it I spotted it straight away)

And then I'd ask how much effort would it be to support such tooling? i.e. dependabot PRs, upgrading support, security fixes etc (plus the maintenance you pointed out)

FYI: I think since I've been working on kuma I vaguely remember 1 minor spelling mistake that was spotted and fixed soon after, but I could be wrong that this was the only instance.

schogges: Every once in a while I see typos in different environments: docs, code, comments, ...
It's probably a personal preference that also code (like variables and function names) is spelled correctly, but that's not a reason to pull in another dependency and maintain it 😄 
Generally I think it's important that text in both an app and a docs page has no spelling issues. Like equally important. Anyways, nothing I'd want to add for now. Thanks for sharing your opinion 🙂

PR #3381, opened 2025-01-09
chore(config/deps): explicitely install globals, @eslint/js

schogges:
See title

Missed to install `@eslint/js` and `globals` in #3373


johncowen (approved):
🎉

PR #3382, opened 2025-01-10
chore: lint config package

schogges:
The main goal of this change is to add linting for the `config` package. While doing this I wanted to take the chance and add a few more things:

1. Root `Makefile` for root operations `install` and `clean` for easy cleaning and reinstallation of dependencies
2. Add rule `import/no-extraneous-dependencies` to error in case we make use of a module without installing it beforehand. This caused some issues already and a first fix was added with #3381. But also here I have to install `glob` as it wasn't listed in the config's dependency list, but is imported in [packages/config/scripts/ci.cjs#L1](https://github.com/kumahq/kuma-gui/blob/36911e6e113bdf823e98233d2eac4543acb588b4/packages/config/scripts/ci.cjs#L1).
3. Linting of staged files in `packages/config`
4. Auto-fixing some linting issues that got revealed after adding the linter here
5. A typo I found

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,7 @@ include $(MK)/install.mk
 include $(MK)/check.mk
 
 .PHONY: clean
-clean:
-	@echo "Recursively removing all node_modules/ directories in `pwd`..."; \
-		$(MAKE) .clean
+clean: .clean ## Dev: Remove all `node_modules` recursively
 
 .PHONY: install
 install: .install ## Dev: Install all dependencies
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,13 @@
 SHELL := /usr/bin/env bash
 MK := ./packages/kuma-gui/mk
 
-ROOT := ./mk
-
-include $(ROOT)/index.mk
 include $(MK)/install.mk
 include $(MK)/check.mk
 
 .PHONY: clean
-clean: .install/clean/workspace ## Dev: Delete all node_modules directories
+clean:
+	@echo "Recursively removing all node_modules/ directories in `pwd`..."; \
+		$(MAKE) .clean
 
 .PHONY: install
 install: .install ## Dev: Install all dependencies
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,14 @@
+SHELL := /usr/bin/env bash
+MK := ./packages/kuma-gui/mk
+
+ROOT := ./mk
+
+include $(ROOT)/index.mk
+include $(MK)/install.mk
+include $(MK)/check.mk
+
+.PHONY: clean
+clean: .install/clean/workspace ## Dev: Delete all node_modules directories
+
+.PHONY: install
+install: .install ## Dev: Install all dependencies
--- a/mk/index.mk
+++ b/mk/index.mk
@@ -1,0 +1,7 @@
+FILE_DIR := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
+
+.PHONY: .install/clean/workspace
+.install/clean/workspace:
+	@cd $(FILE_DIR).. && \
+		echo "Recursively removing all node_modules/ directories in `pwd`..."; \
+		$(MAKE) .clean
--- a/mk/index.mk
+++ b/mk/index.mk
@@ -1,7 +1,0 @@
-FILE_DIR := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
-
-.PHONY: .install/clean/workspace
-.install/clean/workspace:
-	@cd $(FILE_DIR).. && \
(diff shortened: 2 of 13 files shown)


mk/index.mk
```diff
@@ -0,0 +1,7 @@
+FILE_DIR := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
+
+.PHONY: .install/clean/workspace
+.install/clean/workspace:
+	@cd $(FILE_DIR).. && \
+		echo "Recursively removing all node_modules/ directories in `pwd`..."; \
+		$(MAKE) .clean
```
johncowen, line 7: Don't we already have clean target somewhere? If so, could that be reused somehow?

johncowen (commented):
Nice looks/sounds great 👍 

I left one q, but lemme just give it whirl also, brb

schogges:
Flake: https://github.com/kumahq/kuma-gui/actions/runs/12711384283/job/35434626365?pr=3382

Dataplane details for standard Data Plane Proxy -- Overview tab has expected content (failed)

Restarting...

mk/index.mk
schogges: Yes, there is the one I trigger with `$(MAKE) .clean`. Would you prefer to use that directly in the root Makefile and remove this index file?

johncowen: I think so 🤔 , then there's no other reason to have this file, is that right? Guessing using that other one works as expected?

Makefile
```diff
@@ -0,0 +1,13 @@
+SHELL := /usr/bin/env bash
+MK := ./packages/kuma-gui/mk
+
+include $(MK)/install.mk
+include $(MK)/check.mk
+
+.PHONY: clean
+clean:
+	@echo "Recursively removing all node_modules/ directories in `pwd`..."; \
```
johncowen, line 9: If we are adding this `echo` should it be in the implementation of `.clean` so it does the same everywhere?

packages/config/.lintstagedrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "*.{cjs,js,ts}": "make lint"
+}
```
johncowen, line 3: I wonder if there is anyway to move this to the workspace root and have it apply to the entire workspace? Doing something on the repo before staging applies to the entire workspace, so is probably more workspace configuration rather than individual project configuration, wdyt?

We might also want to move these to JS files: https://github.com/lint-staged/lint-staged?tab=readme-ov-file#using-js-configuration-files so we can have a config in `@kumahq/config` and then import that from each repo. Up to you whether you want to try this bit as a follow up PR at some point or here. I think this would help when we come to implement lint-staged elsewhere also, we'd just import this config from `@kumahq/config`

johncowen (commented):
Sorry for the wait, had to jump on a couple of other bits.

I tried it locally and I think this looks good to me, I left 2 more q's, lemme know on those and then I think we can 👍 this!

packages/config/.lintstagedrc.json
johncowen: P.S. If any of this ☝️ is in the slightest bit problematic, lets leave it for a follow up

Makefile
schogges: Yes, makes a lot more sense 👍

packages/config/.lintstagedrc.json
schogges: > I wonder if there is anyway to move this to the workspace root and have it apply to the entire workspace? Doing something on the repo before staging applies to the entire workspace, so is probably more workspace configuration rather than individual project configuration, wdyt?

I had the same thought first, but reverted that for now and would do this in a follow-up PR.

> We might also want to move these to JS files: https://github.com/lint-staged/lint-staged?tab=readme-ov-file#using-js-configuration-files so we can have a config in `@kumahq/config` and then import that from each repo. Up to you whether you want to try this bit as a follow up PR at some point or here. I think this would help when we come to implement lint-staged elsewhere also, we'd just import this config from `@kumahq/config`

Yes, sounds great 👍 will do that in the follow-up 🙂

packages/kuma-gui/mk/install.mk
```diff
@@ -8,5 +8,6 @@
 
 .PHONY: .clean
 .clean:
-	find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +
+	@echo "Recursively removing all node_modules/ directories in `pwd`..."; \
+		find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +
```
johncowen, line 12: Oh one thing I just noticed, ideally `find` would always execute in the workspace root, no matter where I use `make` from. I'd be happy doing this in any sort of follow up though, so no action here just a note.

PR #3384, opened 2025-01-10
chore(deps): upgrade reflect-metadata

johncowen:
FYI: You will only see a warning for the issue here if you run `make install` with completely clean `node_modules`

---

I chose to put the override in the root package.json, I also had to run `npm dedupe` several times to get it to "stick" so unfortunately quite a lot of churn in the package-lock.json.

I just looked at it myself, and there are lots of downgrades of some transitive dependencies here (as well as upgrades). I would assume these still satisfy the semver ranges for everything and are a consequence of the dedupe's.

Ideally I'd like less churn here though, so if there are any suggestions/ideas for trying to get a smaller changeset I'd be very grateful. Otherwise I think this is probably still good as it seems like the changed deps are mostly (at least) dev/tooling deps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "workspaces": [
     "packages/*"
   ],
+  "overrides": {
+    "reflect-metadata": "^0.2.2"
+  },
   "engines": {
     "node": ">=20.9",
     "npm": "^10.0.0"
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
     "packages/*"
   ],
   "overrides": {
-    "reflect-metadata": "^0.2.2"
+    "@cucumber/messages": "^27.0.2"
   },
   "engines": {
     "node": ">=20.9",
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,7 +4,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "kuma-gui",
       "workspaces": [
         "packages/*"
       ],
@@ -594,17 +593,6 @@
         "regexp-match-indices": "1.0.2"
       }
     },
-    "node_modules/@badeball/cypress-cucumber-preprocessor/node_modules/@cucumber/messages": {
-      "version": "27.0.2",
-      "resolved": "https://registry.npmjs.org/@cucumber/messages/-/messages-27.0.2.tgz",
-      "integrity": "sha512-jo2B+vYXmpuLOKh6Gc8loHl2E8svCkLvEXLVgFwVHqKWZJWBTa9yTRCPmZIxrz4fnO7Pr3N3vKQCPu73/gjlVQ==",
-      "dependencies": {
-        "@types/uuid": "10.0.0",
-        "class-transformer": "0.5.1",
-        "reflect-metadata": "0.2.2",
-        "uuid": "10.0.0"
-      }
-    },
     "node_modules/@badeball/cypress-cucumber-preprocessor/node_modules/glob": {
       "version": "10.4.5",
       "resolved": "https://registry.npmjs.org/glob/-/glob-10.4.5.tgz",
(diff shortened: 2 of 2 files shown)


schogges (commented):
I think the "problematic" dep is `@cucumber/gherkin-utils@9.0.0`, which has a dependency of `@cucumber/messages: ^24.0.0` (-> `24.1.0`). Because of `dedupe`  running on the entire tree, there is more deduplication happening. So I guess there is no way having less churn when using `dedupe` 🤔 
Looks like the deduplication caused that `@cucumber/messages` got downgraded to `24.1.0` (except the one that `@badeball/cypress-cucumber-preprocessor` depends on).
Another option would be to override `@cucumber/messages`: `"@cucumber/messages@<=24.1.0": ">=24.1.0"` and dedupe. That would cause upgrades instead of downgrades, but also a lot of churn in the lock-file.

johncowen:
> Another option would be to override @cucumber/messages: "@cucumber/messages@<=24.1.0": ">=24.1.0" and dedupe. That would cause upgrades instead of downgrades, but also a lot of churn in the lock-file.

Ah good call, lemme try that 👍

package.json
```diff
@@ -2,6 +2,9 @@
   "workspaces": [
     "packages/*"
   ],
+  "overrides": {
+    "reflect-metadata": "^0.2.2"
```
schogges, line 6: Can we foresee how long the override will stay? Would it make sense to leave the upwards range open? What I had in mind is: `"reflect-metadata@<=0.2.1": ">=0.2.2"`

johncowen: > Can we foresee how long the override will stay?

Kinda not sure, when I was looking at this I think I saw upstream issues for it. I think in any case, I think the caret is fine for this. You never know if might be gone in a week or two, I also doubt that we'll see `reflect-metadata` release a new version any time soon, but I could be wrong.

schogges: Ok, what about only overriding this specific range and have it like they specify it via `reflect-metadata@<1`: "^0.2.2"`? I think this would prevent us accidentally staying with version smaller than `v1` 🤔

johncowen:
I did the pin on `@cucumber/messages` like you suggested, looks much better i think, still quite a bit of churn to stare at, but it looks like it did what you said, all upgrades, no downgrades 👍

schogges:
> I did the pin on `@cucumber/messages` like you suggested, looks much better i think, still quite a bit of churn to stare at, but it looks like it did what you said, all upgrades, no downgrades 👍

Nice, LGTM 🙂 Although I would have probably used a less "aggressive" override like `"@cucumber/messages@<=24.1.0": ">=24.1.0"` which would allow anything above `24.1.0`.

johncowen:
> Nice, LGTM 🙂 Although I would have probably used a less "aggressive" override like "@cucumber/messages@<=24.1.0": ">=24.1.0" which would allow anything above 24.1.0.

Maybe I misunderstood what this would be achieving? Wouldn't using `^27.0.2` mean we get any non-breaking version of `27` which I think is what we want? Not sure if I follow your suggestion here? I would guess (but I could be wrong) by the time we get to `28` we've probably already removed the override.

If not I guess there are two outcomes:

1. With `>=27.0.2` `28` is released (a breaking version), and we still have the override in place, we are upgraded because of the `>=27.0.2`. Things break because we unexpectedly went to 28.
2. With `^27.0.2` `28` is released (a breaking version), whatever is depending on `@cucumber/messages` bumps to `28`, and we still have the `^27.0.2` override in place. Things break because we didn't go to 28.

Both outcomes depend on the breaking change hitting us i.e. in our code path (unlikely), both outcomes depend on us having the override still in place when the bump happens (unlikely). There are probably some other things, but all in all I'd say both outcomes are equally problematic/unlikely in theory.

Separate from that, the fact that we should be explicit with a override and try to be like "this is what the upstream package should be doing i.e. hey upstream, please use ^27.0.2" makes me think I'd prefer to stick to what we have here in this PR right now.

I guess we shold say, we should always default to using `^` for `deps` and `devDeps` unless there is a strong reason, that includes `deps` and `devDeps` applied via `overrides`

If you still feel super strongly about this let me know tho, I think technically they both have equal problems, its more about the meaning of it.

schogges:
> > Nice, LGTM 🙂 Although I would have probably used a less "aggressive" override like "@cucumber/messages@<=24.1.0": ">=24.1.0" which would allow anything above 24.1.0.
> 
> Maybe I misunderstood what this would be achieving? Wouldn't using `^27.0.2` mean we get any non-breaking version of `27` which I think is what we want? Not sure if I follow your suggestion here? I would guess (but I could be wrong) by the time we get to `28` we've probably already removed the override.
> 
> If not I guess there are two outcomes:
> 
> 1. With `>=27.0.2` `28` is released (a breaking version), and we still have the override in place, we are upgraded because of the `>=27.0.2`. Things break because we unexpectedly went to 28.
> 2. With `^27.0.2` `28` is released (a breaking version), whatever is depending on `@cucumber/messages` bumps to `28`, and we still have the `^27.0.2` override in place. Things break because we didn't go to 28.
> 
> Both outcomes depend on the breaking change hitting us i.e. in our code path (unlikely), both outcomes depend on us having the override still in place when the bump happens (unlikely). There are probably some other things, but all in all I'd say both outcomes are equally problematic/unlikely in theory.
> 
> Separate from that, the fact that we should be explicit with a override and try to be like "this is what the upstream package should be doing i.e. hey upstream, please use ^27.0.2" makes me think I'd prefer to stick to what we have here in this PR right now.
> 
> I guess we shold say, we should always default to using `^` for `deps` and `devDeps` unless there is a strong reason, that includes `deps` and `devDeps` applied via `overrides`
> 
> If you still feel super strongly about this let me know tho, I think technically they both have equal problems, its more about the meaning of it.

Generally I think we are good to go, but I don't think that using `>=24.1.0` would mean that we would jump on `28` once it's released 🤔 anyways with the dedupe all `@cucumber/messages` will end up with `27.0.2`. I thought using an open range would allow certain dependencies to stay with their version above `24.1.0`, but with dedupe this is not the case anymore. If we see issues upgrading `@cucumber/messages` for all, we probably have to be more specific in terms of which dependencies we want to override (e.g. `@cucumber/gherkin@30.0.4` depends on `"@cucumber/messages": ">=19.1.4 <=26"` -> `@cucumber/messages@26.0.1`).
Since all these are devDeps I think we can take the risk 🙂

PR #3385, opened 2025-01-13
chore(dataplanes): minor amends to dataplane download bundling

johncowen:
Adds another few minor amends to bundle downloading:

- Text amends
- Default most options to ticked
- Add JSON whitespace to downloaded XDS config file so you have a chance of reading it when you open it


schogges (approved):
👍

PR #3401, opened 2025-01-14
chore: lint-staged

schogges:
This adds `lint-staged` to all packages and the root to lint staged changes depending on the package (environment).

1. root: lint lock-file only
2. kuma-gui: lint styles and scripts only
3. config: lint scripts only

---

Apart from lint-staged all the shareable make targets and recipes have been moved to our config/tooling package.


Makefile
```diff
@@ -1,11 +1,18 @@
 SHELL := /usr/bin/env bash
-MK := ./packages/kuma-gui/mk
+MK := ./packages/config/src/mk
```
schogges, line 2: Note: I have tried to resolve the path to the shareable make targets via `node` using

```Makefile
MK := $(shell node -e "console.log(require('node:path').dirname(require.resolve('@kumahq/config')) + '/mk')")
```

and it worked as expected until I ran `make clean`. So without installing the config package it can't be resolved. That means we can only use make in ci when installing via `npm install` or `npm ci` first.

johncowen: Oh! 😆 classic!

Umm ok, nearly there with this one then, thanks for the note!

packages/kuma-gui/Makefile
```diff
@@ -28,9 +28,6 @@ install/sync: .install/sync
 .PHONY: help
 help: .help ## Display this help screen
 
-.PHONY: clean
-clean: .clean ## Dev: Delete all node_modules directories
-
```
johncowen, line 33: A note for the future, ideally you'd be able to call `make clean` from within a sub package dir also, I guess it doesn't work here quite yet, I think elsewhere there is something that would make it work, but we can add this later if so. No action here, just anote.

johncowen (approved):
Nice 👏  This is all really to coming together now!

Umm I suppose one more note here: As we said yesterday, seeing as the lint-stage config files are so small, if we can figure out how to add those to the package.json files without causing the annoying red config missing error at some point in the future that'd be perfect.

Makefile
johncowen: `npm query .workspace | jq -r '.[] | select(.name == "@kumahq/config") | .path'`

I think that would work, lets get this PR first tho we can circle back with another PR for that ^ (or similar) if we want to use it. Just leaving as a note.

packages/kuma-gui/Makefile
schogges: It would work and remove the node_modules from the kuma-gui package, but I thought it doesn't really make sense to call this from a sub package and therefore removed it.

schogges: Will move on and resolve this. We can add it back later if necessary 🙂

johncowen: I would recommend folks always work from within the project they are working on, our monorepo isn't a monolith application, is a collection of micro "service" applications, ideally each should always be able to stand on their own.

Obvs its fine if you want to be in root and use `make -C` to call things in the sub apps, but I think, if possible, primarily things should work as standalone micro "services"/applications not as a monolith. We use a monorepo just so we don't have a tonne of repos, not so we have a monolith application.

BTW, I'm fine that you can do both, and I get that you work from root I'm a-ok with that, in fact its important to me that you can do both (ie call `clean`/`install` etc from the application or the root)

---

Definitely one for future tho 👍

PR #3408, opened 2025-01-15
feat(application): source default URL params from storage

johncowen:
This PR allows us to source certain parameters from localStorage to use for defaults for a page. This functionality isn't used as yet, as in this PR we don't actually save any parameters to localStorage (we will add these later)

For example, if you set your page to `75` on one listing, chances are you probably want that to say as `75` throughout the application for all listings.

There are several interesting things that we backwards and forwards with a bit (even this morning), so I'll try detail a little bit.

### Design constraints

- Preferably, we want to always use declarative `DataSource` to access external data, this includes localStorage data which is external to the main Javascript execution.
- We want it to be as easy as possible to use as a consumer, ie. someone building features for our application.
- Related to the above point, we should continue to use the automatic defaulting that we already have in place that gives us query param defaults, query param normalisation and the engineering focussed type features that we have in `RouteView :params`

### Approach

- We have two types of localStorage "global for the app" and "per route". When you ask for user preferences/storage via `/me/:route` we retrieve both and merge, overwriting anything in app specific storage with route specific storage.
- We wait for the existing `/me/:route` DataSource in RouteView before we do any parameter "fiddling". Previous to this PR we would not render the default slot of the RouteView anyway, ~so from an application standpoint waiting do do this before parameter fiddling before doing anything shouldn't change any functionality here~ 👈 I found this was true with human usage but not in e2e test, so I reused `redirected` an extra guard in the template. Note: Whilst prototyping this, I exposed our storage via a `useStorage` accessor (i.e. I cheated and bypassed DataSource because we weren't sure if we could use this). The final version here, removes the `useStorage` accessor and reverts to using DataSource, which does work with this version. I kept the service injection re-arrangement I did to make the `useStorage` accessor, because I figured we may as well configure it like this, but I removed the `useStorage` accessor entirely as we want folks to always use DataSource to access storage, `useStorage` would make it easy to work around that.
- Whilst prototyping we found we had concurrency issues with the requirement to wait for our storage derived params, mainly around having two side-effect-y Vue `watch`es. I initially chose the second "one time only" `watch` to add the asynchronous code. In the end (for other reasons) I figured I could just merge both watches into one, and use a guard to only run the code in the second one once.
- Furthermore using a second `watch` to pull the storage defaults out of an async storage, and using that by default, meant that things would work fine if you didn't add a certain param to the URL, but if you did add a param to the URL manually (i.e. via a bookmark), this manual param would be overwritten by the one from localStorage. i.e. the defaults were specified in the wrong order. Moving to a single watch meant I could add the storage derived params in the same function along with the merge of URL params and query params, solving this issue.
- We had a bit of fun pairing through this, and figured I'd add some more comments while we were here. I think the code might benefit from some slight organisation, more around grouping related RouteView functionality parts together. i.e. not as much as what AppView probably could do with but _something_

### Manual Testing

Add the following to localStorage:

Screenshot 2025-01-15 at 12 19 33

With the above settings:

- Visit `http://localhost:8080/gui/zones?page=1`, `&size=75` should be added to the URL (note we hardcode in `size=50` everywhere in the app, so this proves its using the value from localStorage)
- Visit `http://localhost:8080/gui/zones?page=1&size=25`. The URL should remain with `&size=25` proving that we don't overwrite anything in the URL with the localStorage settings.
- Delete your localStorage, visit `http://localhost:8080/gui/zones?page=1`, `&size=50` should be added to the URL, proving that we use the hardcode value of `50` when there is no value in localStorage

I might look to add some tests here either in the PR or after, I wanted to get this up and merged asap so we can continue with the feature this enables (and I think there is still a bit of under-the-hood work to do before we can do that)

I might also add some inline PR comments brb.

### DIFF
--- a/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
+++ b/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
@@ -27,10 +27,11 @@
       :src="uri(sources, '/me/:route', {
         route: props.name,
       })"
+      @change="resolve"
       v-slot="{ data: me }"
     >
       <slot
-        v-if="me && submit"
+        v-if="me && submit && redirected"
         :id="UniqueId"
         name="default"
         :t="t"
@@ -72,6 +73,7 @@ import { sources } from '@/app/me/sources'
 import type { Ref } from 'vue'
 import type { RouteRecordRaw, RouteLocationNormalizedLoaded } from 'vue-router'
 
+
 export type RouteView = {
   name: string
   from: Ref
@@ -89,6 +91,11 @@ const win = window
 const env = useEnv()
 const can = useCan()
 const uri = useUri()
+
+
+let resolve: (resolved: object) => void
+const meResponse = new Promise((r) => resolve = r)
+
 const htmlAttrs = useAttrs()
 const { t } = useI18n()
 const route = useRoute()
@@ -180,13 +187,26 @@ watch(routeParams, (val) => {
 })
 
 // when any URL params change, normalize/validate/default and reset our actual application params
+const redirected = ref<boolean>(false)
 watch(() => {
   return Object.keys(props.params).map((item) => { return route.params[item] || route.query[item] })
-}, () => {
+}, async () => {
+  const stored = await meResponse
+
+  // merge params in order of importance/priority:
+  // 1. Anything stored by the user in storage
+  // 2. URL query params
+  // 3. URL path params
   const params = {
+    ...get(stored, 'params', {}),
     ...route.query,
     ...route.params,
   }
+
+  // normalize/validate/default all params using the RouteView :params
+  // 1. Ignore any `?param[]=0` type params, we just take the first one
+  // 2. Using normalizeUrlParam and the type information from RouteView :params convert things to the correct type i.e. null > false
+  // 3. Use RouteView :params to set any params that are empty, i.e. use RouteView :params as defaults.
   Object.entries({
     ...props.params,
   }).reduce((prev, [prop, def]) => {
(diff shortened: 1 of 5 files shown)


packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
```diff
@@ -27,10 +27,11 @@
       :src="uri(sources, '/me/:route', {
         route: props.name,
       })"
+      @change="resolve"
       v-slot="{ data: me }"
     >
       <slot
-        v-if="me && submit"
+        v-if="me && submit && redirected"
```
johncowen, line 34: This is the extra template guard I reused to make sure we don't show anything until we've set our default query params. This is only necessary during CI for some reason, probably due to the single tick used in the `watch` as a consequence of the `async/await`

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
```diff
@@ -89,6 +91,11 @@ const win = window
 const env = useEnv()
 const can = useCan()
 const uri = useUri()
+
+
+let resolve: (resolved: object) => void
+const meResponse = new Promise((r) => resolve = r)
```
johncowen, line 97: Boilerplate for setting up the JS accessible promise to wait for `me` to resolve.

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
```diff
@@ -180,56 +187,77 @@ watch(routeParams, (val) => {
...
+        query: cleanQuery(params, route.query),
+      })
     }
-    return prev
-  }, {} as Record<string, string | boolean | undefined>)
-  if (Object.keys(params).length > 0) {
-    router.replace({
-      query: cleanQuery(params, route.query),
-    })
+    redirected.value = true
```
johncowen, line 234: This is the two merged `watch`es, you might want to look at the old file/new file outside a diff to get a better idea.

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
```diff
@@ -180,56 +187,77 @@ watch(routeParams, (val) => {
...
-}, () => {
+}, async () => {
+  const stored = await meResponse
+
+  // merge params in order of importance/priority:
+  // 1. Anything stored by the user in storage
+  // 2. URL query params
+  // 3. URL path params
   const params = {
+    ...get(stored, 'params', {}),
```
johncowen, line 201: This is the super important bit that simplifies everything, we use the stored params first, then query, then URL params. i.e. we just continue to do as we did before. The problem we more fiddling around with Vue so we could use `await` here without hitting concurrency issues and making sure we always normalize route params, but only redirect once.

packages/kuma-gui/src/app/application/services/storage/index.ts
```diff
@@ -0,0 +1,21 @@
...
+      storage.setItem(`${prefix}:${key}`, JSON.stringify(value))
+      return value
+    } catch (e) {
+      console.error(e)
+    }
+    return {}
+  }
+
+  return { get, set }
+}
```
johncowen, line 21: If we want to get rid of `console.error`s entirely (and throw/catch errors or something else), we'd have to decide what to do here. Thinking about the type of error we'd get here, I'd be almost 100% on just lint-ignoring these ones.

packages/kuma-gui/src/app/me/sources.ts
```diff
@@ -2,38 +2,32 @@ import merge from 'deepmerge'
...
-          size: 50,
+      const [app, route] = await Promise.all([
+        get('/'),
+        get(params.route),
+      ])
+      return merge(
+        {
+          params: {
+            size: 50,
+          },
```
johncowen, line 22: I have a plan to use this as the "centralized 50". We currently set `size=50` in every single list view, I'd like to remove those and make it optional to specific an exact hardcoded value (you'd still have to specify `size` as a route param to get the type goodness etc)

johncowen (commented):
Notes:

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
schogges: Yes, that helps a lot and thanks for all the new comments 👍

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
schogges: Looks much cleaner now 👍

packages/kuma-gui/src/app/me/sources.ts
schogges: We could probably move these things to the global `me` as defaults 🤔

schogges (approved):
LGTM 👍

Issue #3283, opened 2024-12-09
Consider adding structured/YAML dropdown to summary drawers


schogges:
In order to track and decide on a new way of representing and navigation for the configuration in the summary drawer, I've created this issue.

> As of next I'd like to discuss if we really want to add the `Structured/YAML` dropdown to subscription. This introduces a whole new way of how we represent data. Instead it would probably make more sense to fully focus on the `YAML` representation and remove any duplicate information, that the user could also retrieve from the config itself. 

 _Originally posted by @schogges in [#2922](https://github.com/kumahq/kuma-gui/issues/2922#issuecomment-2523343141)_

Assuming that most of our users prefer the `YAML` representation of the config, we should make it as accessible as possible and the default type of representation (if there is a navigation option). In some places we extract some information from the config and show them directly next to the `YAML`. So we can also think of removing the duplicate information.
But also there are actual useful (computed) information (e.g. `status`) that we might want to show next to the `YAML`.

<h3>Tab Navigation</h3>
<img src="https://github.com/user-attachments/assets/5abdd52f-e81e-478f-9915-3b84d1212afc" width="200" height="400" />

<h3>No Navigation</h3>
<img src="https://github.com/user-attachments/assets/32452bd0-ddd4-4360-acdb-46d0472e802e" width="200" height="400" />
<img src="https://github.com/user-attachments/assets/0e59ce95-bb3a-41aa-8325-cbd97a2a18d0" width="200" height="400" />

<h3>Dropdown Navigation</h3>
<img src="https://github.com/user-attachments/assets/3760e50c-1d6e-4fe8-98d4-8bf48cd62749" height="200" width="500" />
<img src="https://github.com/user-attachments/assets/ef2fe066-bf6b-4977-89f4-e43857b4b0bc" height="200" width="500" />

<h2>Inventory list</h2>

- [x] [ZoneEgressSummaryView.vue](https://github.com/kumahq/kuma-gui/tree/master/packages/kuma-gui/src/app/zone-egresses/views/ZoneEgressSummaryView.vue)
- [x] [HostnameGeneratorSummaryView.vue](https://github.com/kumahq/kuma-gui/tree/master/packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue)
- [x] [PolicySummaryView.vue](https://github.com/kumahq/kuma-gui/tree/master/packages/kuma-gui/src/app/policies/views/PolicySummaryView.vue)
[...]

PR #3409, opened 2025-01-16
fix(route-view): non-boolean query search params not added in nested routes

schogges:
As part of the work for #3283 we need to add non-boolean query search params to nested routes like `*SummaryView`. Due to a race condition the `route.replace` in the eager watcher got cancelled when the `routeUpdate` in the other watcher finished earlier. This caused that non-boolean query search params would not been added to the URL as expected.

```shell
# Added props param `foo: 'bar'` (non-boolean) to the `MeshServiceSummaryView` and logged the result of `router.replace`
Error: Navigation cancelled from "/meshes/default/services/mesh-services/card-0.array?page=1&size=50" to "/meshes/default/services/mesh-services/card-0.array?page=1&size=50&foo=bar" with a new navigation.
```

---

Expected behaviour:

- Any primitive value can be added to the URL as a query search param (no matter if nested route or not)
- When `redirected` is `false` the change to the URL that adds query search params should **not** be pushed to the stack (use `router.replace`)
- When `redirected` is `true` the change to the URL that adds/removes/changes query search params should be pushed to the stack, such that a user can "go back" and stays on the same route (use `router.push`)
- A boolean query param that is set to `true` as per default can be toggled via component `@change`-event (e.g. `<XInputSwitch @change="route.update({ foo: $event}" ... />`

---

To fix this issue I changed some parts in the `RouteView`:

1. Track `route.params` in a `ref` to check if a navigation happened that didn't unmount the `RouteView`. This is a scenario that happens when a user navigates between different `*SummaryView`s from a listing without closing the summary view before navigating to the next one. If this kind of navigation is detected, we have to manually set back the refs `redirected` and `routeParams`
2. In order to better describe the different states/refs we use, I have renamed previous `routeParams` to `reactivePropsParams` as this is a state about `props.params`. The `ref` for `route.params` takes over the name `routeParams`.
3. The `routeUpdate` (which is a `router.push` under the hood) in the watcher for `reactiveRouteParams` will only be triggered once the initial `router.replace` (to update the route params, including query search params) is done. To achieve this I added the check for the `redirected.value` and also added an `await` to the `router.replace` to wait for the Promise to resolve. This ensures there is no more race condition and we can safely update the `redirected` afterwards.
4. I had to keep the watcher with the booleans only for the case of `KInputSwitch`.

### DIFF
--- a/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
+++ b/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
@@ -43,7 +43,7 @@
           name: props.name,
           update: routeUpdate,
           replace: routeReplace,
-          params: routeParams,
+          params: reactivePropsParams,
           back: routerBack,
           children,
           child,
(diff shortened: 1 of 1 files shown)


schogges:
> Do you have the work still for the structured/yaml dropdown in one of the summary views?
> 
> Wondering if we should PR that first if so (and then rebase this)? Then we have something to test/preview with. At the moment I don't think (??) there is anywhere we can test this right now seeing as its not something we've needed here in the past?

Yeah, I was thinking about adding tests with the PRs for #3283. I have started with this in a local branch and I can rebase to this and continue with it 👍

johncowen:
Nice sounds good! I'll wait for a PR on that one then, I figure it'll be a quick approve and we can get this onto it

schogges:
Saving this for the next iteration.

johncowen:
Can this one be closed up now since https://github.com/kumahq/kuma-gui/pull/3457?

schogges:
> Can this one be closed up now since #3457?

Yes 👍

PR #3410, opened 2025-01-16
chore(ci): various makefile improvements

johncowen:
I stumbled across quite a few things whilst fixing up our `clean` script so:

1. It can be run from any sub-package (like `make install`)
2. It didn't echo the warning message twice in certain environments

I have a longer term goal of making our "wiring root" Makefiles easier to maintain (there might end up being lots of them) and possibly creating them from a template. This goal means that the less path differences etc (`../` over here, `./` over there) the better. This has influenced some decisions here.

In order to run `make clean` from anywhere and it automatically do this from root, we need to know where root is, from wherever we are. This led me to using `npm prefix` for `NPM_WORKSPACE_ROOT`. This led me to realize a `NPM_WORKFLOW_CONFIG_PATH` that I'd not noticed in a previous PR, which led me to our `release` script. Getting here made me realize our `release` script wasn't working since we moved our `mk` files over to `@kumahq/config`.

Hopefully the above explains why this is "various" things rather than just the one.

### `make release/prune` fixes

Since we moved `mk` to `@kumahq/config`, and we also `rm` `@kumahq/config` during SBOM generation, this means we can no longer run any Makefiles after SBOM generation, doh 🤦 

I figured that we only want to delete the dependency configuration for the SBOM generation, not the entire `@kumahq/config` module. Additionally our SBOM generation ignores devDependencies, and `@kumahq/config` is a devDependency but its dependencies are not (ideally our SBOM generator would ignore devDependencies entirely but hey ho).

This PR runs a tiny script before SBOM generation that moves `@kumahq/config`s package config for `dependencies` to `devDependencies` and also deletes its `peerDependencies`. We then run `npm install --package-lock-only` as before. This I believe has the exact same effect as `rm`ing the entire @kumahq/config` module, but without also deleting our Makefile tooling/scripts.

## `make clean`

Can now be run from any repository in the workspace or from the root. I also added a confirmation (if not in CI) so that you have some kind of visibility into what its deleting and a chance to say no.

## make variables

I added some useful variables into `help.mk`, which are necessary for all scripts. I added them to `help.mk` because I would like to make it easy to add the `help` make target and make it hard not to. Adding them to make means you _must_ include `help.mk`

### copypasta ability

I made it so the top ~11 lines of every `Makefile` are exactly the same, this makes it easy to copy/paste them to make multiple ones and keep them in sync i.e. poor mans templating (we may add templating at some point but I doubt it)

### wait for a system/application to test

Lastly I added a wait for our test site to come up, cypress wasn't waiting long enough and even before this it would try 3 times before (sometimes) succeeding. `make` now tales a split second longer due to the new variables, so I added a quick check for our application being served on localhost:5681 before running Cypress.

I may add some inlines.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,23 @@
+# Please keep this file free of actual scripts
+# It should only be used for adding "non-dot" aliases and documentation
+
 SHELL := /usr/bin/env bash
-MK := ./packages/config/src/mk
+
+KUMAHQ_CONFIG := $(shell npm query .workspace | jq -r '.[] | select(.name == "@kumahq/config") | .path')
+MK := $(KUMAHQ_CONFIG)/src/mk
 
 ## make help: if you're aren't sure use `make help`
 .DEFAULT_GOAL := help
+include $(MK)/help.mk
 
 include $(MK)/install.mk
 include $(MK)/check.mk
-include $(MK)/help.mk
+
+.PHONY: help
+help: .help ## Display this help screen
 
 .PHONY: clean
 clean: .clean ## Dev: Remove all `node_modules` recursively
 
 .PHONY: install
 install: .install ## Dev: Install all dependencies
-
-.PHONY: help
-help: .help ## Display this help screen
--- a/packages/config/Makefile
+++ b/packages/config/Makefile
@@ -1,13 +1,26 @@
-MK := ./src/mk
+# Please keep this file free of actual scripts
+# It should only be used for adding "non-dot" aliases and documentation
+
+SHELL := /usr/bin/env bash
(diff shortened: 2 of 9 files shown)


johncowen:
@schogges I made an addition here in https://github.com/kumahq/kuma-gui/pull/3410/commits/1d81b99c458cc9ad595b91bde7af91a2f700ad82 to make the sbom exclusion reusable for when/if we need to exclude any more workspace modules from our SBOM. Figured I'd ping you for some 👀 just incase.

schogges (approved):
👍

PR #3414, opened 2025-01-17
feat(zone-ingress-summary-view): use dropdown to select config format

schogges:
This is part of the efforts to allow the user using the preferred format of showing the config: `structured` vs `yaml` and adds a dropdown menu to select between `structured` and `yaml` in the `ZoneIngressSummaryView`.

Screenshot 2025-01-17 at 18 31 58

Part of #3283

### DIFF
--- a/packages/kuma-gui/cypress/support/step_definitions/index.ts
+++ b/packages/kuma-gui/cypress/support/step_definitions/index.ts
@@ -150,6 +150,11 @@ When(/^I click the "(.*)" element(?: and select "(.*)")?$/, (selector: string, v
   }
 })
 
+When(/^I select "(.*)" from the "(.*)" element$/, (value: string, selector: string) => {
+  $(selector).click()
+  $(`[data-testid="select-item-${value}"] button`).click()
+})
+
 When(/^I (.*) on the "(.*)" element$/, (event: string, selector: string) => {
   switch (event) {
     case 'hover':
--- a/packages/kuma-gui/cypress/support/step_definitions/index.ts
+++ b/packages/kuma-gui/cypress/support/step_definitions/index.ts
@@ -150,11 +150,6 @@ When(/^I click the "(.*)" element(?: and select "(.*)")?$/, (selector: string, v
   }
 })
 
-When(/^I select "(.*)" from the "(.*)" element$/, (value: string, selector: string) => {
-  $(selector).click()
-  $(`[data-testid="select-item-${value}"] button`).click()
-})
-
 When(/^I (.*) on the "(.*)" element$/, (event: string, selector: string) => {
   switch (event) {
     case 'hover':
--- a/packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressSummaryView.vue
+++ b/packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressSummaryView.vue
@@ -84,9 +84,7 @@
                 </XLayout>
               </header>
               <template v-if="route.params.format === 'structured'">
-                <XLayout
-                  type="stack"
-                  size="none"
+                <div
                   class="stack-with-borders"
                   data-testid="structured-view"
                 >
@@ -160,7 +158,7 @@
                       </template>
                     </template>
                   </DefinitionCard>
-                </XLayout>
+                </div>
               </template>
               <template v-else>
                 <div>
--- a/packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressSummaryView.vue
+++ b/packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressSummaryView.vue
@@ -86,7 +86,7 @@
               <template v-if="route.params.format === 'structured'">
                 <XLayout
                   type="stack"
-                  size="unset"
+                  size="none"
                   class="stack-with-borders"
                   data-testid="structured-view"
                 >
--- a/packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressSummaryView.vue
+++ b/packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressSummaryView.vue
@@ -6,6 +6,7 @@
       codeSearch: '',
       codeFilter: false,
       codeRegExp: false,
+      format: 'structured',
     }"
     v-slot="{ route, t }"
   >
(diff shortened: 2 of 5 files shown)


packages/kuma-gui/cypress/support/step_definitions/index.ts
```diff
@@ -150,6 +150,11 @@ When(/^I click the "(.*)" element(?: and select "(.*)")?$/, (selector: string, v
   }
 })
 
+When(/^I select "(.*)" from the "(.*)" element$/, (value: string, selector: string) => {
+  $(selector).click()
+  $(`[data-testid="select-item-${value}"] button`).click()
+})
+
```
schogges, line 157: Note: this is very specific to how [KSelect](https://kongponents.konghq.com/components/select.html) works

schogges:
Flake

image

packages/kuma-gui/cypress/support/step_definitions/index.ts
johncowen: oh yeah, I get this is a little awkward, although we've usually just done two clicks for this to avoid anything project specific in the step definitions

https://github.com/kumahq/kuma-gui/blob/09a9318112583014e162ac6d64397f03e0c7dbdd/packages/kuma-gui/features/mesh/dataplanes/Index.feature#L137-L139

packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressSummaryView.vue
```diff
@@ -52,110 +53,142 @@
               </h2>
             </template>
 
-            <div
-              class="stack-with-borders"
+            <XLayout
+              type="stack"
```
johncowen, line 57: Did we lose the border here intentionally?

schogges: Oh, nice catch 👍 added it back 🙂

packages/kuma-gui/cypress/support/step_definitions/index.ts
schogges: Ah, right 💡 so the step definition is a little easier to use and if the `KSelect` changes how it works we would only need to change it in one place 🤔

johncowen: I'd be super wary of baking this sort of stuff into the steps, especially `data-testid` CSS selectors like this. I think at the least in this case I'm not sure what would happen if we had two KSelect components on the page with the same value.

I would guess thats easy to fix, but I'd still much prefer to try and avoid baking in application specific `data-testid` CSS selectors into the steps.

schogges: Good point, I'll remove this 👍

johncowen:
Something looks off to me here:

Screenshot 2025-01-20 at 14 56 20

compare the bottom padding with say a dataplane summary:

Screenshot 2025-01-20 at 14 56 31

lmk!

schogges:
> Something looks off to me here:
> 
> Screenshot 2025-01-20 at 14 56 20
> 
> compare the bottom padding with say a dataplane summary:
> 
> Screenshot 2025-01-20 at 14 56 31
> 
> lmk!

Argh 😄  It's because I switched to move `XLayout` and the `size` prop defaults to `normal` which adds a margin to each child element. I added another option `none`.

johncowen:
> Argh 😄 It's because I moved to XLayout and the size prop defaults to normal which adds a margin to each child element. I added another option none.

Ah gotcha, umm lets be a little bit more cautious with this, especially since its XLayout which is currently in like "try-out period".

🤔 An option of `none` would be the same as just not adding the `XLayout` at all right? So I don't think we should add that as an option, at least at the moment. Do you follow what I mean.

I think if anything `size="small"` is what you want, but I'm even a little cautious of doing:

```yaml
  <XLayout
    type="stack"
    size="small"
    class="stack-with-borders"
    data-testid="structured-view"
  >
```

This applies two lots of stacking styles, `stack-with-borders` and `XLayout`'s `stack`. I think this, coupled with my opinion that this layout is really a two column table with row scoped `th`'s, make me think we should just avoid changing this in this PR. 

Umm I suppose that aside, I'd suggest one of two options here:

1. Leave as is. Keep using `<div class="stack-with-borders">` as before. The feature/PR is adding the select functionality, and whilst it's great that we are upgrading things as we go, we've hit some complication here so we should be cautious.
2. If we want to upgrade to XLayout here/now, I'd go with `<XLayout type="stack" size="small" class="stack-with-borders">` but I really don't like mixing the two approaches to stacking just because we want borders so this would be my least fave option.

---

FYI: As its vaguely related to this point, I'd like to delete DefinitionCard entirely soon-ish.

schogges:
> > Argh 😄 It's because I moved to XLayout and the size prop defaults to normal which adds a margin to each child element. I added another option none.
> 
> Ah gotcha, umm lets be a little bit more cautious with this, especially since its XLayout which is currently in like "try-out period".
> 
> 🤔 An option of `none` would be the same as just not adding the `XLayout` at all right? So I don't think we should add that as an option, at least at the moment. Do you follow what I mean.
> 
> I think if anything `size="small"` is what you want, but I'm even a little cautious of doing:
> 
> ```yaml
>   <XLayout
>     type="stack"
>     size="small"
>     class="stack-with-borders"
>     data-testid="structured-view"
>   >
> ```
> 
> This applies two lots of stacking styles, `stack-with-borders` and `XLayout`'s `stack`. I think this, coupled with my opinion that this layout is really a two column table with row scoped `th`'s, make me think we should just avoid changing this in this PR.
> 
> Umm I suppose that aside, I'd suggest one of two options here:
> 
> 1. Leave as is. Keep using `<div class="stack-with-borders">` as before. The feature/PR is adding the select functionality, and whilst it's great that we are upgrading things as we go, we've hit some complication here so we should be cautious.
> 2. If we want to upgrade to XLayout here/now, I'd go with `<XLayout type="stack" size="small" class="stack-with-borders">` but I really don't like mixing the two approaches to stacking just because we want borders so this would be my least fave option.
> 
> FYI: As its vaguely related to this point, I'd like to delete DefinitionCard entirely soon-ish.

Aight gotcha 👍 I just though we are moving everything to `XLayout`, even `class="stack-with-borders"` at some point 🤔

johncowen:
> Aight gotcha 👍 I just though we are moving everything to XLayout, even class="stack-with-borders" at some point 🤔

Yeah we kinda are, but just not totally sure how yet, `borders` isn't a layout thing, so shouldn't be part of `XLayout`. This is partly why I think it should probably use a `Table` with row scoped `th`s. These `structured` views look and use the exact same design/CSS as our list tables. So I'm pretty sure I'd want them to somehow use the same CSS.

It would be great if our listing tables could also use row scoped `th`s (it would solve a few minor warts we have), just not quite sure how we can do that yet (if at all)

schogges:
> > Aight gotcha 👍 I just though we are moving everything to XLayout, even class="stack-with-borders" at some point 🤔
> 
> Yeah we kinda are, but just not totally sure how yet. `borders` isn't a layout thing, so shouldn't be part of `XLayout`. This is partly why I think it should probably use a `Table` with row scoped `th`s. These `structured` views look and use the exact same design/CSS as our list tables. So I'm pretty sure I'd want them to somehow use the same CSS.
> 
> It would be great if our listing tables could also use row scoped `th`s (it would solve a few minor warts we have), just not quite sure how we can do that yet (if at all)

Fair point 🙂 so we probably add another component? From my point of view that's the difference of a table vs list. A table has columns where each row shows the data related to that column. A list is just not that strict 🤔

johncowen:
> Fair point 🙂 so we probably add another component?

Not totally sure yet, this is why I think we should stick to adding the format menu here.

Ideally I want an `XTable` that we use here and in _our_ lists. But not quite sure yet if thats going to be possible because `KTable` won't let you add `th`'s in the `tbody` or add `scope` attributes to `th`s.

> A table has columns where each row shows the data related to that column.

I'd say this layout has columns. Two columns, the key and the value. The key column is left aligned the value column is right aligned. If you think about it like that they look exactly the same as our list tables.

---

Anyway, this is double-y why I think we just add the format menu here. I appreciate we are going through and upgrading things elsewhere as we go, I just think that this needs a tiny bit more thought first.

johncowen (approved):

PR #3416, opened 2025-01-20
refactor: for route state prefer events over v-model

schogges:
In #1966 a watcher has been added to sync boolean query search params. This watcher caused some issues in nested routes due to a race condition.

The reason for the watcher (from #1966)
>Whilst I'd prefer not use two way binding models we have some components that we use that uses these as required values, such as KInputSwitch

Turns out that the components `KSelect` and `KCheckBox` now support the `@change` events. When used alongside the `:model-value` we have our getter and setter separated. This means that we can remove the additional watcher.

From #3409 
> 4. I had to keep the watcher with the booleans only for the case of KInputSwitch.

### DIFF
--- a/packages/kuma-gui/src/app/application/components/route-view/README.md
+++ b/packages/kuma-gui/src/app/application/components/route-view/README.md
@@ -93,12 +93,10 @@ Defining these "dependencies" here has several advantages:
 4. Boolean route params are transformed correctly for the URL. In the above
    example the resulting query parameter will be the existence of `?checked`
    not `?checked=true` and `?checked=false`
-5. Route params can be used as `v-model` values. Ideally we usually use events
-   to change route properties, but for components that only support `v-model`
-   you can use the route.param as the v-model. For example `<KCheckBox
-   v-model="route.params.checked" />` will automatically sync the checkbox state
-   to the `?checked` query parameter. This aims to make it easier to maintain
-   state in the browser URL and harder not to
+5. Route params can be used as a state for input elements. In order to retrieve and update the route properties
+   we can use events. For example `<KCheckBox :model-value="route.params.checked" @change="(value) => route.update({ checked: value })" />`
+   will automatically sync the checkbox state to the `?checked` query parameter. This aims to make it easier to maintain
+   state in the browser URL and harder not to. Note that this does not work when using the bi-directional `v-model` attribute.
 
 ## Setting the `<title>` of the page
 
--- a/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
+++ b/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
@@ -174,18 +174,6 @@ const routeView = {
 }
 const routeParams = reactive<Params>(structuredClone(props.params) as Params)
 
-// Updates the URL for route params if used as a modelValue (for boolean props only)
-watch(routeParams, (val) => {
-  if (route.name === props.name) {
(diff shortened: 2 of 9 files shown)


packages/kuma-gui/src/app/application/components/route-view/README.md
```diff
@@ -93,12 +93,10 @@ Defining these "dependencies" here has several advantages:
...
-5. Route params can be used as `v-model` values. Ideally we usually use events
-   to change route properties, but for components that only support `v-model`
-   you can use the route.param as the v-model. For example `<KCheckBox
-   v-model="route.params.checked" />` will automatically sync the checkbox state
-   to the `?checked` query parameter. This aims to make it easier to maintain
-   state in the browser URL and harder not to
+5. Route params can be used as a state for input elements. In order to retrieve and update the route properties
+   we can use events. For example `<KCheckBox :model-value="route.params.checked" @change="(value) => route.update({ checked: value })" />`
+   will automatically sync the checkbox state to the `?checked` query parameter. This aims to make it easier to maintain
+   state in the browser URL and harder not to. Note that this does not work when using the bi-directional `v-model` attribute.
```
schogges, line 99: > Note that this does not work when using the bi-directional `v-model` attribute.

Ideally it wouldn't matter what to use, but when removing the "boolean watcher" it's not working with the `v-model` anymore 🤔 I think this is what we want for now, but leaving this as a note to check back if we can bring it back to work in next iterations.

schogges: Ok, that makes a lot more sense to me now too 😄 I'll add `:checked` to our  `XCheckBox` component. For the `KInputSwitch` I'd go for `:enabled` or would you prefer something else?

packages/kuma-gui/src/app/x/index.ts
```diff
@@ -110,6 +110,8 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
           ['XTeleportSlot', XTeleportSlot],
           ['XDisclosure', XDisclosure],
           ['XAboutCard', XAboutCard],
+          ['XInputSwitch', XInputSwitch],
+          ['XCheckbox', XCheckBox],
```
johncowen, line 114: Thanks for doing this detail!

packages/kuma-gui/src/app/application/components/route-view/README.md
```diff
@@ -93,12 +93,10 @@ Defining these "dependencies" here has several advantages:
...
-5. Route params can be used as `v-model` values. Ideally we usually use events
-   to change route properties, but for components that only support `v-model`
-   you can use the route.param as the v-model. For example `<KCheckBox
-   v-model="route.params.checked" />` will automatically sync the checkbox state
-   to the `?checked` query parameter. This aims to make it easier to maintain
-   state in the browser URL and harder not to
+5. Route params can be used as a state for input elements. In order to retrieve and update the route properties
+   we can use events. For example `<KCheckBox :model-value="route.params.checked" @change="(value) => route.update({ checked: value })" />`
+   will automatically sync the checkbox state to the `?checked` query parameter. This aims to make it easier to maintain
+   state in the browser URL and harder not to. Note that this does not work when using the two-way binding attribute `v-model`.
```
johncowen, line 99: If it was me I'd just remove point 5. I don't believe vue-router's params can be used as `v-model` values either

johncowen (approved):
LGTM! I left one comment about point 5 in the docs, but I'm easy either way.

PR #3417, opened 2025-01-20
fix(ci): fix release/prune amends

johncowen:
Last two little amends to the pruning we do for release:

- Prettify the JSON
- `rm` and replace package.json instead of using a single command (can cause issues elsewhere)


schogges (approved):
👍

PR #3419, opened 2025-01-20
fix: flaky tests due to mismatch of payload type

### LINKED ISSUE #3376: flake: expected '<div.x-layout.stack.normal>' to contain 'online'
### What happened?

I think we've got a new flake:

Screenshot 2025-01-07 at 13 11 34

e.g. https://github.com/kumahq/kuma-gui/actions/runs/12670762378/job/35311113869

Guessing its as a result of https://github.com/kumahq/kuma-gui/pull/3327

schogges:
Over the past weeks we have noticed some flakes of our browser tests where the assertion was mainly about a state being `online`. I have traced it down to a mismatch of our payload type. In our tests we always set the state that we want to assert for in `dataplane.inbound.health.state`, but according to our type in [data-planes](https://github.com/kumahq/kuma-gui/blob/26976933fb563768d5f09099f2d7d28c11545bcd/packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts#L13) there is no entry `health`. The `state` can be found in `dataplane.inbound.state`.

Example:
Screenshot 2025-01-20 at 11 11 05

Closes #3376

### DIFF
--- a/packages/kuma-gui/features/mesh/dataplanes/Index.feature
+++ b/packages/kuma-gui/features/mesh/dataplanes/Index.feature
@@ -80,7 +80,6 @@ Feature: mesh / dataplanes / index
                 inbound:
                   - tags:
                       kuma.io/service: !!js/undefined
-                    state: NotReady
             dataplaneInsight:
               mTLS:
                 certificateExpirationTime: !!js/undefined
--- a/packages/kuma-gui/features/mesh/dataplanes/Index.feature
+++ b/packages/kuma-gui/features/mesh/dataplanes/Index.feature
@@ -31,11 +31,10 @@ Feature: mesh / dataplanes / index
             networking:
               gateway: !!js/undefined
               inbound:
-                - health:
-                    ready: true
-                  tags:
+                - tags:
                     kuma.io/service: service-1
                     kuma.io/zone: zone-1
+                  state: Ready
           dataplaneInsight:
             mTLS:
               certificateExpirationTime: 2023-11-03T09:10:17Z
@@ -79,10 +78,9 @@ Feature: mesh / dataplanes / index
               networking:
                 gateway: !!js/undefined
                 inbound:
-                  - health:
-                      ready: true
-                    tags:
+                  - tags:
                       kuma.io/service: !!js/undefined
+                    state: NotReady
             dataplaneInsight:
               mTLS:
                 certificateExpirationTime: !!js/undefined
--- a/packages/kuma-gui/features/mesh/dataplanes/Index.feature
+++ b/packages/kuma-gui/features/mesh/dataplanes/Index.feature
@@ -80,6 +80,7 @@ Feature: mesh / dataplanes / index
                 inbound:
                   - tags:
                       kuma.io/service: !!js/undefined
+                    state: Ready
             dataplaneInsight:
               mTLS:
                 certificateExpirationTime: !!js/undefined
--- a/packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
+++ b/packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
@@ -10,6 +10,14 @@ Feature: dataplanes / no-subscriptions
       KUMA_SUBSCRIPTION_COUNT: 0
       KUMA_DATAPLANEINBOUND_COUNT: 1
       """
+    And the URL "/meshes/default/dataplanes/backend/_overview" responds with
+      """
+      body:
+        dataplane:
+          networking:
+            inbound:
+              - state: Ready
+      """
     When I visit the "/meshes/default/data-planes/backend/overview" URL
     And the "$detail-view" element contains "backend"
     And the "$overview-content" element contains "offline"
--- a/packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
+++ b/packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
@@ -10,9 +10,6 @@ Feature: dataplanes / no-subscriptions
       KUMA_SUBSCRIPTION_COUNT: 0
       KUMA_DATAPLANEINBOUND_COUNT: 1
       """
-    And the URL "/meshes/default/dataplanes/backend/_overview" responds with
-      """
-      """
     When I visit the "/meshes/default/data-planes/backend/overview" URL
     And the "$detail-view" element contains "backend"
     And the "$overview-content" element contains "offline"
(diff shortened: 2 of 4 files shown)


johncowen (approved):
Amazing thanks! 🎉

packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
```diff
@@ -16,8 +16,7 @@ Feature: dataplanes / no-subscriptions
         dataplane:
           networking:
             inbound:
-              - health:
-                  ready: true
+              - state: NotReady
```
johncowen, line 19: Oh wait is this one right? It used to be `health.ready: true` whereas we say `NotReady` here?

(FYI: we used to have `health.ready` but it got removed a while back)

schogges: Yeah, but this had no effect and according to the assertion a few lines below we need `NotReady` here

packages/kuma-gui/features/mesh/dataplanes/Index.feature
```diff
@@ -79,10 +78,9 @@ Feature: mesh / dataplanes / index
               networking:
                 gateway: !!js/undefined
                 inbound:
-                  - health:
-                      ready: true
-                    tags:
+                  - tags:
                       kuma.io/service: !!js/undefined
+                    state: NotReady
```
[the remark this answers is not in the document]
schogges: Answered on the comment above, it's the same thing 🙂

packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
johncowen: Ah wait! The title of the test is:

> When there are no subscription I don't get an error

So this is testing when the DPP is healthy but there is no subscription, so we need to make the DPP healthy so we can test that the offline-ness is due to no subscriptions rather than an unhealthy DPP. d'you follow?

schogges: Does it mean that since the other parts I fixed included both `health.ready: true` and `health.state: Ready` we probably want to test if no `state` is set at all? 🤔

packages/kuma-gui/features/mesh/dataplanes/Index.feature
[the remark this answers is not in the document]
johncowen: Same for this one, it purposefully makes the DPP healthy and then adds a single disconnected subscription, the lack of a connected subscription then makes this `offline` in the assertion (I think?)

packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
johncowen: Possibly separately, depending on whether `state` if ever empty (I think its an enum so probably not). I think at the least we should fix the tests here due to the change from `healthy` to `state`, and that should be by making the test exactly the same as it was previously. From looking closer it looks like these mocks and there values were like this on purpose.

schogges: According to our [dp mocks](https://github.com/kumahq/kuma-gui/blob/07fada0abe83773a81c748c4dcf8ed3a2eda55e3/packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_/_overview.ts#L81) it can be that `state` is not set. If it's set then with either `Ready`, `NotReady` or `Ignored`.

schogges: > (FYI: we used to have health.ready but it got removed a while back)

In that case I think we should remove it from the tests as it has no effect other than adding an entry with everything mocked inside 🤔

schogges: nvm, the entry is added via `KUMA_DATAPLANEINBOUND_COUNT: 1`. I removed the whole payload setting from the test and it still passes 🙂

johncowen: I think there are a couple of convos happening here:

1. It sounds like the flake was down to us using the older `healthy` prop in these tests. I think we should update the tests so they are mocked in the same way with the new prop i.e. `health.ready: true` > `state: Ready` not health.ready: true` > `state: NotReady`.
2. "In that case I think we should remove it from the tests as it has no effect other than adding an entry with everything mocked inside" yes exactly, and replace it with `state`, not sure if there are more of these.
3. "According to our [dp mocks](https://github.com/kumahq/kuma-gui/blob/07fada0abe83773a81c748c4dcf8ed3a2eda55e3/packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_/_overview.ts#L81) it can be that state is not set. If it's set then with either Ready, NotReady or Ignored." Maybe but according to the protobufs I think `state` is always supposed to be set https://github.com/kumahq/kuma/blob/93ee1540e5205211fd76cd91b4f54471bdc4ffea/api/mesh/v1alpha1/dataplane.proto#L144-L156. I think this is one of those where sometimes I'm not sure how reliable that is, but we also default to `Ready`in the fe code if its empty anyway https://github.com/kumahq/kuma-gui/blob/07fada0abe83773a81c748c4dcf8ed3a2eda55e3/packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts#L90.

packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
```diff
@@ -10,15 +10,6 @@ Feature: dataplanes / no-subscriptions
...
       """
-    And the URL "/meshes/default/dataplanes/backend/_overview" responds with
-      """
-      body:
-        dataplane:
-          networking:
-            inbound:
-              - health:
-                  ready: true
-      """
```
johncowen, line 21: This needs a `state: Ready` so that the assertion `And the "$overview-content" element contains "offline"` doesn't give us a false positive if the mocks randomly give us a `NotReady`

packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
schogges: Aaaaah, ok that makes sense 😄 thanks for the explanation 👍

packages/kuma-gui/features/mesh/dataplanes/NoSubscriptions.feature
schogges: Got it, thanks 😄 seems like I just mixed up the subscriptions and inbounds 😅

PR #3420, opened 2025-01-20
chore: XI18n improvements

johncowen:
Several improvements to our `XI18n` component (we currently only use this in a single place in our application)

- `XI18n` now exports `t`: `<XI18n v-slot="{ t }">{{ t('path' )}}</XI18n>`
- `XI18n` is now prefixable, prefixes apply to both component and exported `t` usage. Relative paths should begin with a `.` i.e. the following uses both relative and absolute paths: `<XI18n prefix="http.api.property" v-slot="{ t }">{{ t('.path' )}}{{ t('common.path' )}}</XI18n>`.
- Simpler (not using Teleport), slot-less, safe i18n-ing is available via `<XI18n path="path" :params="{ var: html }" />`. If used in this "component" form, `path`, `prefix` and `params` are always escaped via Vue's native `escapeHtml` function but the i18n content is not.
- Teleport based i18n slotting is still available for those times when you need to insert/interpolate rendered Vue components into a i18n string. Any defined slots will overwrite any `params` already set.

As we only use this currently in one place, please use `make run/docs` to view the sandboxed docs page for the various use cases you can see in the code here.

Followup of:

- https://github.com/kumahq/kuma-gui/pull/3293

Part of:

- https://github.com/kumahq/kuma-gui/issues/2430

The plan is to use this `X18n` component everywhere in our application.


packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
```diff
@@ -1,35 +1,88 @@
...
+  path: '',
+  params: () => ({}),
 })
+
+const i18n = typeof props.strings !== 'undefined' ? createI18n(typeof props.strings === 'function' ? props.strings(icuEscapeHtml) : props.strings, useEnv()) : useI18n()
+
+const t = (...rest: Parameters<typeof i18n['t']>) => {
+  rest[0] = `${rest[0].startsWith('.') ? props.prefix : ''}${rest[0]}`
+  return i18n.t(...rest)
+}
```
schogges, line 72: I'm not a big fan of directly manipulating parameters because depending on the parameter type it can have unexpected side effects. Therefore I just generally avoid doing it.

What about doing this instead?
```suggestion
const t: typeof i18n['t'] = (tKey, ...rest) => {
  const _tKey = `${rest[0].startsWith('.') ? props.prefix : ''}${rest[0]}`
  return i18n.t(_tKey, ...rest)
}
```

johncowen: I did something slightly different to avoid mutating any args here (and below) lmk!

packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
```diff
@@ -1,35 +1,100 @@
...
+
+const t = (key: Parameters<typeof i18n['t']>[0], ...rest: ParametersExceptFirst<typeof i18n['t']>) => {
+  return i18n.t(`${key.startsWith('.') ? props.prefix : ''}${key}`, ...rest)
+}
+const safeT = (
+  key: Parameters<typeof i18n['t']>[0],
+  params: Parameters<typeof i18n['t']>[1] = {},
+  ...rest: ParametersExceptFirstTwo<typeof i18n['t']>
+) => {
+  // escape any param values, if the value isn't a string, empty it out we
```
schogges, line 80: Just a nit, but I think we don't need to type each parameter individually via `Parameters<...>[...]` and `ParametersExcept...` if we just add a type to the whole function, like

```suggestion
const t: typeof i18n['t'] = (key, ...rest) => {
  return i18n.t(`${key.startsWith('.') ? props.prefix : ''}${key}`, ...rest)
}
const safeT: typeof i18n['t'] = (
  key,
  params = {},
  ...rest
) => {
  // escape any param values, if the value isn't a string, empty it out we
```

schogges (approved):
LGTM, just left a comment about a nit 🙂 thanks for the docs in the readme 👍

packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
johncowen: Ok gotcha, lemme take another look at this brb

johncowen: Ok, I aliased the `typeof i18n['t']` to TFunction (the space there made it look weird to me), but you;re right this is simpler 👍 . Lemme know

(I also fixed up the `defineSlots` changes)

schogges (approved):
👍

PR #3442, opened 2025-01-21
fix: don't remove left and right padding on KCodeBlock

johncowen:
Top/bottom padding is necessary.

Fixes https://github.com/kumahq/kuma-gui/pull/3438#pullrequestreview-2564284517


schogges (approved):
👍

Issue #3324, opened 2024-12-19
zone version should have a different color if outdated


johncowen:
### Before

<img width="713" alt="Screenshot 2024-12-19 at 11 39 07" src="https://github.com/user-attachments/assets/138fff3c-60b5-4c83-b4fe-5af9eea4cbcc" />

### After

<img width="933" alt="Screenshot 2024-12-19 at 11 39 21" src="https://github.com/user-attachments/assets/329877e5-e535-480f-a2a8-093346df44fb" />

Also see:

- https://github.com/kumahq/kuma-gui/pull/2621
- https://github.com/kumahq/kuma-gui/issues/2164

PR #3443, opened 2025-01-21
chore(zones): use warning state for displaying version when outdated

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/3324


schogges (approved):
👍

PR #3444, opened 2025-01-21
fix(deps): uses defineSlots macro over useSlots

johncowen:
Required for https://github.com/kumahq/kuma-gui/pull/3336

I got a bit more time to look at https://github.com/kumahq/kuma-gui/pull/3336 and it looks like Vue folks are moving more and more to macro-based `defineSlots` for defining slots.

Moving everything to use defineSlots worked in most areas and I only had to define the types also for areas where we actually need the type information, otherwise defineSlots is a `Record<string, any>` (see https://vuejs-language-tools.vercel.app/features/slots#how-to-handle-indeterminate-slot-types)

There 'may' be some other places where it would be advantageous to define our slots as more than `Record<string, any>` but for the moment this is enough to unblock the upgrade. We can add any more type information if required in later PRs.

Also see https://vuejs.org/api/sfc-script-setup#defineslots

### DIFF
--- a/packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
+++ b/packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
@@ -29,13 +29,12 @@
     <template
       v-for="key in Object.keys(slots)"
       :key="key"
-      #[key]="{ row, rowValue }"
+      #[key]="{ row }"
     >
       <slot
         v-if="(props.items ?? []).length > 0"
         :name="key"
         :row="row as Row"
-        :row-value="rowValue"
       />
     </template>
   </KTable>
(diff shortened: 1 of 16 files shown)


packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -29,21 +29,20 @@
     <template
       v-for="key in Object.keys(slots)"
       :key="key"
-      #[key]="{ row, rowValue }"
```
johncowen, line 32: We don't use `rowValue` anywhere so I removed it.

johncowen:
> In https://github.com/kumahq/kuma-gui/pull/3416 I now also went for defineSlots

Nice! I guess we could lint for these things (in this case `useSlots` usage instead of `defineSlots`) in future. I guess Vue folks might add something at somepoint. Although I can probably think of other things where we need to expose/use an import only in one place (so it needs to be public/exposed) but it shouldn't be used anywhere else.

schogges:
> > In #3416 I now also went for defineSlots
> 
> Nice! I guess we could lint for these things (in this case `useSlots` usage instead of `defineSlots`) in future. I guess Vue folks might add something at somepoint. Although I can probably think of other things where we need to expose/use an import only in one place (so it needs to be public/exposed) but it shouldn't be used anywhere else.

Sounds good 👍 🙂

PR #3445, opened 2025-01-22
feat: in summary drawers use dropdown to select config format

schogges:
As a continuation of #3414 this brings the dropdown to select the preferred config format to all the other summary drawers.

image

Screenshot 2025-01-22 at 10 47 11

Part of #3283


packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
```diff
@@ -41,7 +42,38 @@
           <PolicySummary
             v-if="data"
             :policy="data"
+            :format="route.params.format"
```
johncowen, line 45: Just as a reminder: Is PolicySummary the component we added a while back to share the same drawer in two places?

packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
```diff
@@ -66,7 +66,11 @@ data-planes:
       about:
         title: About this Data Plane Proxy
       gateway: 'Gateway'
-      config: YAML
+      config: Configuration
+      format: Format
+      formats:
+        yaml: YAML
+        structured: Structured
```
johncowen, line 73: I guess these are the kinds of things that we might want to put on some sort of `common` or similar. No action here tho

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
```diff
@@ -41,7 +42,38 @@
...
+            <template #header>
+              <header>
+                <XLayout
+                  type="separated"
+                  size="max"
+                >
+                  <h3>
+                    {{ t('policies.routes.item.config') }}
+                  </h3>
+                  <div v-if="data.spec">
```
johncowen, line 56: Just so I follow, how come we guard on `data.spec` here?

johncowen (approved):
Looks great! 🚀 

I had a couple of inline questions but just for me really.

Ah also I noticed that

https://deploy-preview-3445--kuma-gui.netlify.app/gui/meshes/default/policies/circuit-breakers/sensor-0?page=1&size=50

doesn't have a structured view (maybe this also answer my question about `data.spec`?). Should we add a structured view here so its consistent with all the rest?

Top work 👏

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: Yes, I have added it in https://github.com/kumahq/kuma-gui/pull/2565 

> - To keep the same style and layout for PolicySummaryView and DataPlanePolicySummaryView I have added two one new components by extracting them it from the current PolicySummaryView
>    - PolicySummaryTitle including the link to -> PolicyDetailView
>    - PolicySummary for the content with a default slot for the ResourceCodeBlock

We could maybe think of making this component more abstract and use it for the majority of summary drawers that have the same look and feel 🤔

packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
schogges: Yes, agree 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: This was already implemented like this in the shared component [here](https://github.com/kumahq/kuma-gui/blob/7fef92f51cb1474eda761b28218f8c1fdce39ad6/packages/kuma-gui/src/app/policies/components/PolicySummary.vue#L5). For the structured view we only use entries from the `spec` and afaik not all policies have a `spec` in their config.
Also if there is no `spec`, we only show the yaml format and don't render the select.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: My thought was to reuse the route component, like we do in other areas now. Back when you did #2565 we didn't have the full ability to reuse route components, but now we do.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: Cool gotcha thanks!

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: Ah, understood 🙂 👍

PR #3447, opened 2025-01-22
fix(policies): fixes padding for policy type list

johncowen:
This is actually a bug/incorrect configuration/usage of `XLayout` within `DataCollection`.

I made sure the immediate children of `XLayout` are within the `DataCollection`. i.e. I added two divs, the second one isn't necessary as KPagination is a single block level element, but I wanted to keep _both_ immediate children visible within `DataCollection`.

## Before

Screenshot 2025-01-22 at 11 06 31

## After

Screenshot 2025-01-22 at 11 06 17


schogges (approved):
Nice 👍

PR #3448, opened 2025-01-22
chore: moves AnonymousComponent to XAnonymous

johncowen:
This one feels like it should belong in `x` not `application`, seeing as its a generic-type UI component. The component came up this morning and it reminded me that I wanted to remove it. I also added a short `README.md`


schogges (approved):
👍 Thanks for the explanation in the readme 🙂

Issue #3386, opened 2025-01-13
remove onboarding


johncowen:
### Description

Remove all onboarding functionality

PR #3449, opened 2025-01-22
chore: Several small cleanup amends around K named components

johncowen:
I went through and moved (almost) any remaining K components to use X components instead.

We still use K components in two instances:

1. In Onboarding (we are removing onboarding soon https://github.com/kumahq/kuma-gui/issues/3386 so I didn't bother doing this here)
2. KTable usage in a view in the Policy module. There are ongoing ideas for how we go about possibly doing something with KTable and AppCollection which may or may not end up being `XTable` and `<XCollection :variant="table">`

I noticed a couple of other things whilst doing this and amended here seeing as this is one of those "random bits" PRs, please ask if anything isn't clear 👍

### DIFF
--- a/packages/kuma-gui/src/app/x/components/x-icon/XIcon.vue
+++ b/packages/kuma-gui/src/app/x/components/x-icon/XIcon.vue
@@ -1,8 +1,7 @@
 <template>
   <component
-    :is="slots.default ? KTooltip : XAnonymous"
+    :is="slots.default ? XTooltip : XAnonymous"
     :placement="props.placement"
-    max-width="400"
   >
     <!-- we hardcode decorative to false for the moment due to an issue with kong/icons -->
     <component
@@ -51,11 +50,12 @@ import {
   ProgressIcon,
   InfoIcon,
 } from '@kong/icons'
-import { KTooltip, PopPlacements } from '@kong/kongponents'
 import { useSlots, useAttrs } from 'vue'
 
 import { uniqueId } from '@/app/application'
 import XAnonymous from '@/app/x/components/x-anonymous/XAnonymous.vue'
+import XTooltip from '@/app/x/components/x-tooltip/XTooltip.vue'
+import type { PopPlacements } from '@kong/kongponents'
 
 defineOptions({
   inheritAttrs: false,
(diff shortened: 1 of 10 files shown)


packages/kuma-gui/src/app/x/components/x-icon/XIcon.vue
```diff
@@ -2,6 +2,7 @@
   <component
     :is="slots.default ? KTooltip : XAnonymous"
     :placement="props.placement"
+    max-width="400"
```
schogges, line 5: How come we set the `max-width=400`?

schogges (approved):
LGTM, left a Q

packages/kuma-gui/src/app/x/components/x-icon/XIcon.vue
johncowen: Oh glad you asked I was gonna inline comment on this.

I don't think we ever want an icon tooltip to be more than say `400px`. I noticed one of these "icon tooltips" had a a max-width set and figured it would be good for all of them

johncowen: Maybe yeah 🤔 wdyt? Should we put a 400 on all of them?

schogges: Yeah, I think that's much better than having tooltips covering the whole width of the page 🙂

johncowen: K I moved it to XTooltip via a new file for XTooltip instead of an aliased KTooltip 👍 , will merge this in a sec

schogges: I think we have to change [this](https://github.com/kumahq/kuma-gui/blob/f80108f0ebc5a4cb81da3e85fddb1bd142f862f8/packages/kuma-gui/src/app/x/components/x-icon/XIcon.vue#L3) to adopt `XIcon`, right?

johncowen: Oh good catch! 🙇 1 sec!

johncowen: K I think we are good to go now, thanks for that!

schogges (approved):
🚀

PR #3455, opened 2025-01-23
chore: uses XI18n component pt1

johncowen:
Part of https://github.com/kumahq/kuma-gui/issues/2813

This PR uses `<XI18n />` (https://github.com/kumahq/kuma-gui/pull/3420) throughout the application in place of `v-html`. 

I also turned our `vue/no-v-html` to `error` (it only warns by default)

This means we can start using "non-DNS safe" terms/vars for our i18n texts if we need to do that.

---

I also switched some usages of `innerHTML` to use `textContent` which, aside from other things, is apparently more performant, so may help further speed up our CodeBlock component.

Finally, I added `eslint-plugin-no-unsanitized`, i.e. if we are linting for `v-html` we should also lint for `innerHTML`, can't really depend on one without the other.

There'll be a separate PR to start using `<XI18n v-slot="{ t }" />` within non-route component (in the few places where we need that), and also using `<XI18n v-slot="{ t }" />` within `RouteView` to provide `RouteView::t`. Once that is done I'd like to somehow restrict usage of `useI18n()` so that the only way to translate/localize text is via `<XI18n />`.

---

(sidenote: our RouteAnnouncer functionality has been broken for quite some time, I'll fix that up in a separate PR)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7426,6 +7426,14 @@
         "node": ">=10"
       }
     },
+    "node_modules/eslint-plugin-no-unsanitized": {
+      "version": "4.1.2",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-no-unsanitized/-/eslint-plugin-no-unsanitized-4.1.2.tgz",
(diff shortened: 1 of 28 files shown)


schogges (approved):
Nice 👍

PR #3456, opened 2025-01-23
feature: adds dev-time CSP headers

johncowen:
Adds stricter CSP headers to our development vite server.

Whilst this is dev time only, it ensures that our GUI runs on a similarly configured server (such as https://github.com/kumahq/kuma/issues/12553)

---

As mentioned in other places, it would be good to add work so we can remove the `style-src 'unsafe-inline'`. This will require a globally available `v-style` directive which adds/removes/modifies styles imperatively behind the scenes.

Testing:

Using `make run`, add the following or similar anchor with an inline script and click it.

Untitled-1

### DIFF
--- a/packages/kuma-gui/vite.plugins.ts
+++ b/packages/kuma-gui/vite.plugins.ts
@@ -67,6 +67,16 @@ const server = (template: string = './index.html', vars: Partial<KumaHtmlVars> =
           }).filter(([_, value]) => typeof value !== 'undefined')),
         } satisfies KumaHtmlVars,
       )
+      res.setHeader('Content-Security-Policy', [
+        "default-src 'self'",
+        "script-src 'self'",
+        "script-src-elem 'self'",
+        "img-src 'self' data: ",
+        "style-src 'self' 'unsafe-inline'",
+        // in production connect-src would use kuma's environment variable for
+        // setting the location of the HTTP API (or just use the default)
+        "connect-src 'self' localhost:5681 https://kuma.io",
+      ].join(';'))
       res.end(body)
     } else {
       next()
--- a/packages/kuma-gui/vite.plugins.ts
+++ b/packages/kuma-gui/vite.plugins.ts
@@ -1,5 +1,5 @@
-import crypto from 'crypto'
 import { readFile, stat } from 'fs/promises'
+import crypto from 'node:crypto'
 import { dirname } from 'path'
 
 import type { KumaHtmlVars } from '@/app/application/services/env/Env'
--- a/packages/kuma-gui/vite.plugins.ts
+++ b/packages/kuma-gui/vite.plugins.ts
@@ -38,7 +38,11 @@ export const kumaIndexHtmlVars = (): Plugin => {
     transformIndexHtml: (template) => interpolate(template, htmlVars),
   }
 }
-const server = (template: string = './index.html', vars: Partial<KumaHtmlVars> = {}) => async (server: PreviewServer | ViteDevServer) => {
+const server = (
+  template: string = './index.html',
+  vars: Partial<KumaHtmlVars> = {},
+  csp: boolean = true,
+) => async (server: PreviewServer | ViteDevServer) => {
   server.middlewares.use('/', async (req, res, next) => {
     const url = req.originalUrl || ''
     const baseGuiPath = vars.baseGuiPath || '/gui'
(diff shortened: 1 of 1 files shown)


packages/kuma-gui/vite.plugins.ts
```diff
@@ -67,6 +67,16 @@ const server = (template: string = './index.html', vars: Partial<KumaHtmlVars> =
           }).filter(([_, value]) => typeof value !== 'undefined')),
         } satisfies KumaHtmlVars,
       )
+      res.setHeader('Content-Security-Policy', [
+        "default-src 'self'",
+        "script-src 'self'",
+        "script-src-elem 'self'",
```
schogges, line 73: Just a nit, but I think these are obsolete with `default-src` set.

packages/kuma-gui/vite.plugins.ts
```diff
@@ -67,6 +67,16 @@ const server = (template: string = './index.html', vars: Partial<KumaHtmlVars> =
...
+        "img-src 'self' data: ",
+        "style-src 'self' 'unsafe-inline'",
+        // in production connect-src would use kuma's environment variable for
+        // setting the location of the HTTP API (or just use the default)
+        "connect-src 'self' localhost:5681 https://kuma.io",
```
schogges, line 78: Another nit: Since this is the dev server, why do we need to add `connect-src`.? We would only need `self` which is covered by `default-src` 🤔

schogges (approved):
Just two nits 🙂 I think for both it doesn't hurt to add/remove them, so approving this anyways.

packages/kuma-gui/vite.plugins.ts
johncowen: Yes you're right, I suppose I wanted to set these explicitly, I think they'd be used if we ever needed to add some 3rd party URLs in here for "reasons"

packages/kuma-gui/vite.plugins.ts
johncowen: localhost:5681 seems to be needed which I was a bit 🤔 about myself, but I think its to do with how we use MSW. Actually that's just reminded me of something else I haven't checked, brb

johncowen: Ah yeah this is also needed in development if we want to disable the mocks and point to a local kuma

johncowen:
Cool, thanks! I addressed the inlines also, remember this is only dev-time so we can always come back and tweak things here. If you spot anything that we would usually do in development that we now can't for whatever reason lemme know

Gonna go ahead and merge 👍

johncowen:
Oh wait I just thought of something else I should check 😅 brb!

johncowen:
😅 I added a straight-forward guard so we can turn this off in places we might need to. Later on I'll probably expand this so we can inject the config from The Outside

johncowen:
> As mentioned in other places, it would be good to add work so we can remove the style-src 'unsafe-inline'. This will require a globally available v-style directive which adds/removes/modifies styles imperatively behind the scenes.

I made https://github.com/kumahq/kuma-gui/pull/3458 which hopefully will allow us to remove the `style-src 'unsafe-inline'` also, so I'm gonna hold off on merge here wait for that and rebase.

johncowen:
Gonna take this out of draft. Ideally I wanted to also restrict `style-src` but there are some experiments I wanted to try with that before tightening it up, and I haven't had time to do those as yet. Therefore I've set this to use `style-src 'unsafe-inline'` for the moment.

johncowen:
Oh it was approved a while back also, @schogges could you give this one one last look seeing as it was moved to draft and back out again 🙇‍♂️

packages/kuma-gui/vite.plugins.ts
```diff
@@ -1,3 +1,4 @@
+import crypto from 'crypto'
```
schogges, line 1: Just a nit, but could we add `node:` to the import to indicate that it's builtin?

packages/kuma-gui/vite.plugins.ts
```diff
@@ -67,6 +72,25 @@ const server = (template: string = './index.html', vars: Partial<KumaHtmlVars> =
           }).filter(([_, value]) => typeof value !== 'undefined')),
         } satisfies KumaHtmlVars,
       )
+
+      if (csp) {
+        const nonce = crypto.randomBytes(16).toString('base64')
+        body = body.replace('<meta charset="utf-8" />', `<meta charset="utf-8" /><meta property="csp-nonce" nonce="${nonce}">`)
```
schogges, line 78: Should we use [vite/options/html.cspNonce](https://vite.dev/config/shared-options.html#html-cspnonce)?

johncowen: I did look at this, I can't quite remember why we don't don't bother using it, lemme try and remember, brb

johncowen: Ah! I think it was because we totally override Vites output of `index.html` (which contains `<meta charset="utf-8" /><meta property="csp-nonce" nonce="${nonce}">`) and start afresh with our index.html template. Therefore setting `cspNonce` has zero effect here.

---

I'd rather not totally override Vites output of `index.html`, so if you know of a way we can get that output from within the middleware that would be great. From trying to remember from way back when I think this is the reason we totally override Vites index.html. I guess we might need to reorder the middlewares and use output from `next()` or something, but I might have tried that way back when and be totally mistaken also.

packages/kuma-gui/vite.plugins.ts
johncowen: This one it done, resolving!

packages/kuma-gui/vite.plugins.ts
schogges: > so if you know of a way we can get that output from within the middleware that would be great

Not without messing/fiddling myself 😄 
Maybe just left a todo and we can have a look separately?

schogges (approved):
👍

packages/kuma-gui/vite.plugins.ts
johncowen: Added a comment for this, resolving!

PR #3457, opened 2025-01-23
feat: store global user preferences size and format

### LINKED ISSUE #3413: Store user preferences in local storage
### Description

This is a follow-up to #3283 and #3408 

A user might have some preferences for the GUI that we should consider and store. An example can be how many entries the listings should show or how the config should be displayed. We can set defaults like 50 entries in listings or always use a structured view instead of the yaml representation.
When storing and retrieving the user preferences we should consider that the URL always has priority over what is stored as a preference (i.e. open a url `foo/bar?size=15` should render 15 entries even though the user preference is `size: 75`).

- [ ] store and retrieve user preference of `size` in listings
- [ ] store and retrieve user preference of `format`

schogges:
Stores user preferences for `size` of listings and `format` of configs globally in `localStorage`. Usually we store preferences like column sizes on a route basis. To store preferences "globally" we use the `/` route (still including the prefix).

Closes #3413


packages/kuma-gui/src/app/application/components/data-collection/DataCollection.vue
```diff
@@ -86,7 +87,9 @@
 </template>
 <script lang="ts" generic="T" setup>
 import { useThrottleFn } from '@vueuse/core'
-import { computed, useSlots } from 'vue'
+import { computed, inject, useSlots } from 'vue'
+
+const setMe = inject<(_args: any, useGlobal: boolean) => void>('setMe')
```
schogges, line 92: Note: I've used `provide` and `inject` to always set the `size` of listings globally. Is it better to do this explicitely in all places we use `DataCollection` with `size`?

johncowen: Ideally, DataCollection would not be aware of localStorage (in this case `me`) 🤔 

> Is it better to do this explicitely in all places we use DataCollection with size?

Ideally yes, I actually have another idea that might make this feel a little better

schogges:
@johncowen I think this is good to go for now, but there is still an issue I want to resolve separately. Storing the preferences works, but when trying to retrieve them there is still an issue. It looks like the `meResponse` in the watcher is not updated properly and even a manual refresh doesn't seem to help. Lmk 🙂

johncowen:
> It looks like the meResponse in the watcher is not updated properly and even a manual refresh doesn't seem to help.

This will be using the cache, if you add a `CacheControl: no-cache` (do a search in the codebase for that and you'll see how), this might resolve this issue (maybe).

Maybe we should pair on this as I'm not sure I understand the issue enough to replicate. Lemme finish looking over this and I'll give you a ping.

Initial look from me tho, I think it looks great!

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
```diff
@@ -172,7 +173,21 @@ const routeView = {
     setAttrs([...attributes.values()])
   },
 }
-const routeParams = reactive<Params>(structuredClone(props.params) as Params)
+const routeParams = reactive<Params>({} as Params)
+
+onBeforeRouteUpdate((to, from) => {
+  if(route.name === props.name) {
+    // Check for certain query search params to be stored as global user preference
+    for (const key of ['size', 'format']) {
```
johncowen, line 181: I guess you might have thought this yourself, but ideally we should avoid referencing specific properties by name. I would guess we should be able to figure out which ones to reference based on whether they use `Number` or `String`.

I think the other standout thing is that this code seems to just check if the route param has changed, something that `watch` is made for. 🤔 I'm actually unsure about this one myself, because I'm wondering if I like using `onBeforeRouteUpdate` better even though we have to manually check for changes.

Leave this one with me, and I suppose we'll chat through this if we pair. I think at the very least tho (and I'd still be more than happy for this PR to go in without this improvement) we should avoid referring to the names of the properties

johncowen: Oh wait, this bit of code is always automatically saving things to localStorage whether I've called `route.update` or not?

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
```diff
@@ -191,17 +206,24 @@ watch(() => {
     ...route.params,
   }
 
+  const propsParams = Object.entries(props.params).reduce((acc, [key, value]) => {
+    let param = params[key]
+    if(value === Number || typeof value === 'number') {
```
johncowen, line 211: Immediate thought here is that if the rest of the code is made to work with `Number` and `String` how come we aren't dealing with `String` here?

packages/kuma-gui/src/app/me/sources.ts
```diff
@@ -19,16 +19,18 @@ export const sources = ({ get, set }: Storage) => {
...
         },
         app,
         route,
       )
     },
     '/me/:route/:data': async (params) => {
-      const json = JSON.parse(params.data)
-      const res = merge<object>(await get(params.route), json)
-      set(params.route, res)
+      const { global: useGlobal = false, ...json } = JSON.parse(params.data)
```
johncowen, line 30: Ideally we wouldn't depend on a "magical" property called `global`, granted super unlikely this is gonna be overwritten by something else, but still. Maybe for the minute we call it `$global`, and maybe longer term we think of another way to hide this. 

We thought of adding a new `me` route, but also we are trying to avoid adding multiple possibly nested datasources.

I wondered if we could wrap the data in two different types of classes ( `new Local({data})` `new Global({data})` and we could tell which type it was with an instance of, but we can't do this as yet because the way DataSink works is that it JSON.stringify's and then JSON.parses the data, which would mean these classes would be lost.

packages/kuma-gui/src/app/application/utilities/index.ts
```diff
@@ -90,20 +91,24 @@ export const urlParam = function <T extends URLParamValue> (param: T | T[]): T {
...
+export const normalizeUrlParam = (param: URLParamValue, definition: URLParamDefault): URLParamValues => {
   switch (true) {
-    case typeof def === 'boolean':
-      return param === null ? true : def
-    case typeof def === 'number': {
-      const value = param === null || param.length === 0 ? def : Number(decodeURIComponent(param))
+    case definition === Number:
+      return Number(param)
+    case definition === String:
+      return String(param)
```
johncowen, line 99: Missed this one in our sync, looks like both of these need a `null` guard and need to be `decodeURIComponent(param)`/ `Number(decodeURIComponent(param))` same as below.

johncowen: I think I said in our sync to add a different `case`s, but I think that was because we were adding more conditionals inside the `case`s. But I guess it might be easier to use fallthroughs here. i.e.

```javascript
case definition === String:
case typeof definition === 'string':
```

etc

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
```diff
@@ -187,21 +200,19 @@ watch(() => {
   // 3. URL path params
   const params = {
     ...get(stored, 'params', {}),
-    ...route.query,
+    ...Object.fromEntries(Object.entries(route.query).filter(([_, val]) => val && val.length > 0)),
```
johncowen, line 203: Just thought, we added this in quickly, but could you change this up for me so that it only checks for strings. I think booleans can come in as `null` here and therefore be filtered out, thanks!

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
```diff
@@ -172,10 +184,11 @@ const routeView = {
     setAttrs([...attributes.values()])
   },
 }
-const routeParams = reactive<Params>(structuredClone(props.params) as Params)
```
johncowen, line 175: I just noticed we removed the structuredClone, which from what I remember was to make sure `RouteView::params` was static/hardcoded, if its not doing any harm otherwise could we keep this?

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
schogges: Good point 👍 🙂

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
schogges: Oh yeah, forgot to bring that back in 👍

schogges: Oh, but now I remember why I removed it. It's because `structuredClone` can't be used on `Number` and `String` constructors 🤔

schogges: Looks like `JSON.parse(JSON.stringify(props.params))` works 😄

packages/kuma-gui/src/app/application/utilities/index.ts
schogges: I think it's cleaner as is because I'd have to squeeze that into the other conditions and satisfy typescript 🤔

johncowen: Amended this to use decodeURIComponent so we can resolve this

packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
johncowen: We changed this to start as an empty object, so there's no use cloning anymore.

We change it because, the `Number/String` types would also have to be taken into account at this point, and at the end of the day its unnecessary to start with the `props.params` here because we only overwrite them again in the immediate watcher.

packages/kuma-gui/src/app/me/sources.ts
johncowen: We changed the name of this prop to `$global` to avoid any variable name clashing

schogges (commented):
@johncowen I think this is good to go in now? 🙂

johncowen (approved):
Yep LGTM!

PR #3458, opened 2025-01-24
feature: adds v-style directive

johncowen:
Adds a `v-style` directive to allow us to set occasional inline styles within a strict CSP environment.

Additionally makes it easy to add simple CSS transitions that often require waiting a tick before applying a rule.

Note: There are a couple of unrelated changes here that I bumped into whilst working on this. Happy to split those off if necessary.

~**P.S. I need to add some sort of `Set.difference` polyfill here, its a bit new at the moment.**~ 👈 this is done
~**P.P.S. Cypresses ~Chrome~ Electron (ooooh it uses Electron 🤔 ) doesn't support Set.difference either so it also failed our tests 🎉  **~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5258,6 +5258,7 @@
           "url": "https://github.com/sponsors/ai"
         }
       ],
+      "peer": true,
       "dependencies": {
         "caniuse-lite": "^1.0.30001669",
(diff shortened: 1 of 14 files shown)


schogges (approved):
LGTM 👍

PR #3459, opened 2025-01-24
chore(deps): remove autoprefixer

johncowen:
Removes autoprefixer and manually adds a prefix `-webkit-mask-image` in the only place where we weren't manually adding a `-webkit` prefix where potentially necessary.

Configuration for `autoprefixer` was removed back here https://github.com/kumahq/kuma-gui/pull/384/files#diff-05ed1c7f99e45b485708115502a1e0f28c8547e9a9a29c1b8664f79103cf7873, but the dependency wasn't, this means we've not been using autoprefixer for quite a while but [dutifully upgrading it in dependabot](https://github.com/kumahq/kuma-gui/pulls?q=is%3Apr+is%3Aclosed+author%3Aapp%2Fdependabot+autoprefixer).

All in all its less hassle to manually prefix than to support the dependency, especially considering that as prefixing is now generally frowned upon by browser vendors we are less and less likely to need to add vendor prefixes at all.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5122,43 +5122,6 @@
         "node": ">= 4.0.0"
       }
     },
-    "node_modules/autoprefixer": {
-      "version": "10.4.20",
-      "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-10.4.20.tgz",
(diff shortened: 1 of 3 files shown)


schogges (approved):
👍

PR #3461, opened 2025-01-24
chore(deps): remove @vitest/coverage-istanbul

johncowen:
We don't really use this, if we come to start needed/using this we can add it back in.

Also see https://github.com/kumahq/kuma-gui/pull/3428#issuecomment-2605074831


schogges (approved):
👍

PR #3463, opened 2025-01-27
fix: wrap RouteView::from in a function to prevent it causing a rerender

johncowen:
I noticed that our GH Star button was flickering again, which means we are re-rendering parts of the application that don't need re-rendering.

The fix here adds a wrapping function which needs calling in order to access the value, meaning we don't get unnecessary re-renders.


schogges (approved):
👍

PR #3464, opened 2025-01-27
fix: fixes detection of cacheControl setting for DataSources

johncowen:
Fixes detection of cacheControl setting for DataSources

A long time ago we used the form `?no-cache` to turn off caching for DataSources, but we changed it to use `cacheControl: no-cache` a while back. The piece of code here didn't cope for that change. Therefore in some instances turning off the cache wasn't working. We only turn off the cache in a small amount of places. 

Here's a good example of the issue in the PR which should have had this fix in it:

https://github.com/kumahq/kuma-gui/pull/2453/files#diff-529c3a751eb79cebe6b1e5ad6966b966887fb03638960a4c7fc1f2e32201f458L34


schogges (approved):
🚀

PR #3484, opened 2025-01-28
chore(dependabot): group all @typescript-eslint dependencies

schogges:
In order to reduce the amount of individual dependabot PRs and to update companion packages in one go, we can group them. With this we are now starting to group all `@typescript-eslint/*` dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,13 @@ updates:
       interval: "weekly"
     labels:
       - "dependencies"
+    groups:
+      dev-deps:
+        dependency-type: "development"
+        patterns:
+          - "@typescript-eslint/*"
+
+      
 
   - package-ecosystem: "github-actions"
     open-pull-requests-limit: 50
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -7,6 +7,13 @@ updates:
       interval: "weekly"
     labels:
       - "dependencies"
+    groups:
+      dev-deps:
+        dependency-type: "development"
```
johncowen, line 12: Does it matter that these dependencies are in `@kumahq/config` `.dependencies` not `.devDependencies` (although `@kumahq/config` itself is a devDep), lmk!

schogges: Oh, good point 👍 I'll change the grouping by ignoring the type of dependency, just focus on the group itself.

schogges: Updated it so it now doesn't matter if it's a prod or dev dependency.

PR #3486, opened 2025-01-28
feat(tools): adds "execute on all workspaces" Makefile decorator

johncowen:
This target calls a Makefile target on all packages within the workspace:

i.e. `make ws/lint` will run `make lint` on all packages. If the target doesn't exist, the command will fail only for that sub-package.


schogges (approved):
Nice 👍

PR #3488, opened 2025-01-29
fix: changes Xi18n's default-message prop to default-path

johncowen:
We noticed this:

image (3)

which is the thing we are fixing here, but its a bit of a longer story due to what Xi18n should be doing.

---

The bug is that we aren't using a `:` here:

https://github.com/kumahq/kuma-gui/blob/603c39571c34fcdfad9eeafccf02ee4453fdc17b/packages/kuma-gui/src/app/policies/views/PolicyListView.vue#L68

At first I was like "I can just add the `:` and job done", but then I was like "wait a second that means I can pass unencoded text into `default-message` using a normal `t` and `XI18n` will blindly render it, therefore bypassing any encoding.

This means we can't give `XI18n` a `default-message` prop, because we can't guarantee that its encoded correctly (i.e. sometimes we can use `t` which will encode it, but if we need some HTML in the string that `t` returns, then we can't guarantee and params passed to that are encoded correcly.

I chose to make `default-message` into `default-path` (which matches the `path` prop), i.e. you don't pass it a message to use for a default, you pass it a path, which is then looked up internally.

This means two things:

- You can't currently pass parameters to a default path (we don't need to, yet at least). We don't use a default very much so its no big deal later down the line to add a `default-path={ path 'common.something', params: {}}` or just an additional `default-params` props, or something.
- We need a i18n string that is literally a blank string i.e. `common.i18n.ignore-errors`. ... I'm actually wondering now if I should change this PR to make `ignoreErrors` an option instead 🤔 of the `common.i18n.ignore-errors` thing.

@schogges what do you think on this last bit, the more I think about it we should just make an `ignoreError` option or `ignoreMissingPath` or something, wdyt? Lmk if you don't follow the thinking here

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -20,7 +20,7 @@
     >
       <XI18n
         path="data-planes.routes.items.intro"
-        default-message=""
+        default-path="common.i18n.ignore-error"
       />
       <XCard>
         <search>
--- a/packages/kuma-gui/src/app/gateways/views/GatewayListTabsView.vue
+++ b/packages/kuma-gui/src/app/gateways/views/GatewayListTabsView.vue
@@ -43,7 +43,7 @@
 
         <XI18n
           :path="`gateways.routes.items.navigation.${route.child()?.name}.description`"
-          default-message=""
+          default-path="common.i18n.ignore-error"
         />
 
         <RouterView
--- a/packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorListView.vue
+++ b/packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorListView.vue
@@ -20,7 +20,7 @@
       </template>
       <XI18n
         path="hostname-generators.routes.items.intro"
-        default-message=""
+        default-path="common.i18n.ignore-error"
       />
       <XCard>
         <DataLoader
--- a/packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
+++ b/packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
@@ -1,4 +1,6 @@
 common:
+  i18n:
+    ignore-error: ""
   product:
     name: Kuma
     utm_query_params: utm_source=Kuma&utm_medium=Kuma
--- a/packages/kuma-gui/src/app/meshes/views/MeshListView.vue
+++ b/packages/kuma-gui/src/app/meshes/views/MeshListView.vue
@@ -21,7 +21,7 @@
 
       <XI18n
         path="meshes.routes.items.intro"
-        default-message=""
+        default-path="common.i18n.ignore-error"
       />
 
       <XCard>
--- a/packages/kuma-gui/src/app/policies/views/PolicyListView.vue
+++ b/packages/kuma-gui/src/app/policies/views/PolicyListView.vue
@@ -65,7 +65,7 @@
               </header>
               <XI18n
                 :path="`policies.type.${type.name}.description`"
-                default-message="t('policies.collection.description')"
+                default-path="policies.collection.description"
               />
             </XCard>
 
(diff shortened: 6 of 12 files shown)


schogges:
Nice catch! 👍 

> You can't currently pass parameters to a default path (we don't need to, yet at least). We don't use a default very much so its no big deal later down the line to add a `default-path={ path 'common.something', params: {}}` or just an additional `default-params` props, or something.

As we've talked offline, I'll leave my pov for future reference. We currently barely use `default-message` and we don't need to pass any options or params yet. We don't know if we ever need that in the future. Therefore I agree to just go with `default-path`. If we ever need to pass more than just a path as a fallback, I'd propose to have just one prop as a fallback instead of multiple ones (`default-path`, `default-options`, `default-params`, ...) and probably add something like:

```vue
:fallback="['path.to.default-message', { foo: 'bar' }]"
```

```typescript
type Fallback = Parameters<typeof i18n.t>
t(...props.fallback)
```

schogges:
> * We need a i18n string that is literally a blank string i.e. `common.i18n.ignore-errors`. ... I'm actually wondering now if I should change this PR to make `ignoreErrors` an option instead 🤔 of the `common.i18n.ignore-errors` thing.
> 
> @schogges what do you think on this last bit, the more I think about it we should just make an `ignoreError` option or `ignoreMissingPath` or something, wdyt? Lmk if you don't follow the thinking here

So, I assume with `ignoreError` you mean that to be used as the fallback when `default-path` (or `-message`) doesn't hit. I'm not 100% sure if we need that and can just use empty string in that case 🤔 From a user perspective, what is better: seeing blank or "t('non.matching.path')"? I would guess it's better to show blank and log a warn or error, which is hidden from the user but we get notified.

johncowen:
> Therefore I agree to just go with default-path. If we ever need to pass more than just a path as a fallback, I'd propose to have just one prop as a fallback instead of multiple ones (default-path, default-options, default-params, ...)

Yeah agree, one prop for this would be best I think, but I'm not sure if I'd prefer an array as an arg or an object with explicitly named props. But we can decide that further down the line.

> So, I assume with ignoreError you mean that to be used as the fallback when default-path (or -message) doesn't hit. I'm not 100% sure if we need that and can just use empty string in that case 🤔 From a user perspective, what is better: seeing blank or "t('non.matching.path')"? I would guess it's better to show blank and log a warn or error, which is hidden from the user but we get notified.

if would be `{{ t('path.that.might.not.exist', {}, { ignoreError: true }) }}` that would mean that i18n wouldn't throw on missing path error and just print nothing instead, i.e. the exact equivalent of `{{ t('path.that.might.not.exist', {}, { defaultMessage: ''}) }}`.

Thing is when you move over to the `XI18n` component, you can't define a default message, you have to define a default path, which then means we need to add a path which is just an empty string (`common.i18n.ignore-errors: ''`) which feels a little like a hack to say "ignore missing path errors here please".

Does that explain a little better?

schogges:
> > Therefore I agree to just go with default-path. If we ever need to pass more than just a path as a fallback, I'd propose to have just one prop as a fallback instead of multiple ones (default-path, default-options, default-params, ...)
> 
> Yeah agree, one prop for this would be best I think, but I'm not sure if I'd prefer an array as an arg or an object with explicitly named props. But we can decide that further down the line.
> 
> > So, I assume with ignoreError you mean that to be used as the fallback when default-path (or -message) doesn't hit. I'm not 100% sure if we need that and can just use empty string in that case 🤔 From a user perspective, what is better: seeing blank or "t('non.matching.path')"? I would guess it's better to show blank and log a warn or error, which is hidden from the user but we get notified.
> 
> if would be `{{ t('path.that.might.not.exist', {}, { ignoreError: true }) }}` that would mean that i18n wouldn't throw on missing path error and just print nothing instead, i.e. the exact equivalent of `{{ t('path.that.might.not.exist', {}, { defaultMessage: ''}) }}`.
> 
> Thing is when you move over to the `XI18n` component, you can't define a default message, you have to define a default path, which then means we need to add a path which is just an empty string (`common.i18n.ignore-errors: ''`) which feels a little like a hack to say "ignore missing path errors here please".
> 
> Does that explain a little better?

Ah, well I wasn't aware that `ìgnoreError` is an option of `i18n.t` 😅 in that case it seems that we actually might need a more configurable `default-message`. So, probably it makes sense to move to a more flexible prop than just `default-path`. Or we just set `ignoreError: true` and if we ever need something else we come back here.

schogges (approved):
I'll approve it such that we can ship the fix and we can iterate and improve in a follow-up.

PR #3490, opened 2025-01-29
feat(vcp-health): add policies, show totals only, hide border on wrapped items

### LINKED ISSUE #2667: GlobalCP Health Card: Make numbers of things more clear/explanatory

schogges:
Transforming the VCP health card into a VCP status card, that show total numbers. Additionally I've aligned the icons with the navigation and added the total number of policies to the group.

image

---

As part of this I've also fixed an issue where the borders (`.columns-with-borders`) would still be shown when the items wrap and the layout turns from row into column layout. Now when the container is smaller than `640px`, the borders will be hidden.

Closes #2667


lahabana:
1. icons should match sidebar.
2. only put totals (not unhealthy ones), remove tooltip
3. Add policies next to the count.

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
```diff
@@ -13,13 +13,16 @@
...
         :total="props.globalInsight.zones.controlPlanes.total"
-        :online="props.globalInsight.zones.controlPlanes.online"
         data-testid="zone-control-planes-status"
       >
         <template #icon>
-          <img src="@/assets/images/icon-location-on.svg?url">
+          <img
+            class="icon"
+            src="@/assets/images/navigation/icon-zones.svg?url"
+            alt="Zones Icon"
```
johncowen, line 22: nit: I'd just leave a blank `alt=""` here and add an `aria-hidden` for now (proper decorative iconography I hope will land later on at some point).

Otherwise I think this will end up reading out "Zones Icon Zones" (??) whereas we just want it to say "Zones" judging from the markup:

Screenshot 2025-01-31 at 14 22 59

---

This is super cool right now as we are just amending what is there, but as a side point so you can see where I'm going/what I'm thinking: The vast majority of iconography is usually (should be) accompanied by text. Either similar to this case with a title, or in other cases with a tooltip. Because of that (and probably other reasons) icons are nearly always decorative, and are likely better added via CSS. CSS icons have the nice feature, among other things, that they aren't in the DOM and due to that are kinda aria-hidden by default.

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
```diff
@@ -54,23 +64,38 @@
...
           {{ t('main-overview.detail.health.data_plane_proxies') }}
         </template>
       </ResourceStatus>
+
+      <ResourceStatus
+        :total="props.globalInsight.policies.total"
+        data-testid="policies-status"
+      >
+        <template #icon>
+          <XIcon name="service-document" />
```
johncowen, line 87: Can we change this name to `policy`? I want XIcon to act as an abstraction between "Kuma Noun" > "generic icon name".

So we can then do `<XIcon name="policy" />` everywhere, and then whatever we choose for the look of the icon (maybe `service-document` maybe `something-else`) will be in one place.

---

Another sidenote so you following my thinking: `x` is supposed to be generic so theoretically we shouldn't add kuma specific things there. If we were to build it out more I'd add a way to inject project specific terms/icons into XIcon, so in this case I'm happy to add kuma specific things in there.

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
```diff
@@ -84,8 +109,24 @@ const props = defineProps<{
...
+.icon {
+  height: $kui-icon-size-30;
+  width: $kui-icon-size-30;
+}
+
+@container health-card (max-width: 39.938rem) {
+  .columns-with-borders > * {
+    border-right: unset;
+  }
+}
```
johncowen, line 131: Nice 👍 .

Lemme know what that `rem` is equivalent to, I'm thinking we could make our own `--x-something-size` or `--kuma-something-size`

johncowen (commented):
Looks great, I think this looks so much better/cleaner that what we had before!

I left a question or 2 lmk know on those, I've not approved just yet because of these, but this looks fine to me!

schogges:
> Looks great, I think this looks so much better/cleaner that what we had before!
> 
> I left a question or 2 lmk know on those, I've not approved just yet because of these, but this looks fine to me!

Ah, thanks for the feedback ❤️ I missed to convert it to draft, as I wanted to cleanup some things, but that's even better now so I can already consider your suggestions 👍 🙂

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
schogges: Good point, that makes sense 👍

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
schogges: Oh yes, I really like that abstraction 👍 will change it 🙂

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
schogges: You mean adding a sass function or custom property?

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
schogges: I've removed the `service-document` from `XIcon` because I wanted to reuse the same icon in the policy summary. In summary views we use css icons only.

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
johncowen: 🤔 not totally sure. I realized after writing this its a custom query rather than a custom property, so you're right its not that straight forwards 🤦 .

Umm, I'm a little reluctant to add too much sass everywhere, so I guess for now a comment saying what that size is might help here.

packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
johncowen: gotcha 👍

johncowen (approved):

packages/kuma-gui/src/app/control-planes/locales/en-us/index.yaml
```diff
@@ -7,13 +7,14 @@ main-overview:
       message: !!text/markdown |
         **Welcome to {name}!** We've detected that you don't have any data plane proxies running yet. We've created an onboarding process to help you!
       get_started_link: 'Get started'
-    health:
-      title: 'Global Control Plane health'
+    vcp-stats:
+      title: 'Global Control Plane stats'
```
schogges, line 11: @johncowen I've change the wording from `health` to `stats` as it's now more about statistics than health.

johncowen: Oh nice, yeah makes sense 👍 

Although I'm not sure about using the term `vcp` here, we have one real Control Plane here, and it's not virtual. Kind of a nit this, but the sooner we standardize these i18n files the better really.

Umm maybe its just `.about.title` or `subtitle`.

(side note: I'd like to get rid of the term vcp everywhere its used and just used the term `control-plane`)

schogges: Ok, that's just something I picked up, so I'll drop that (also from my mind) 😄

johncowen (approved):

PR #3491, opened 2025-01-29
fix: prevents KSelect/XSelect from firing events twice

johncowen:
I found that KSelect `select` event would fire twice if you change the `items` as a result of the the first `select` event.

I would guess that KSelect could do something like firing a select event as a result of a non-user-interaction if you change the `selected` property of an `item`. So even though the menu has already been changed/selected by the user, if you change `items` immediately after to be the same as the user selection, it re-fires the same event. But kinda just guessing.

I moved XSelect to use a different approach which avoids using `items[].selected`, which fixes the issue, and I no longer see the double events.

From my investigation this issue doesn't seem to be affecting anything currently, but it'd best if our XSelect component didn't fire `change` twice.

### DIFF
--- a/packages/kuma-gui/src/app/x/components/x-select/XSelect.vue
+++ b/packages/kuma-gui/src/app/x/components/x-select/XSelect.vue
@@ -1,5 +1,6 @@
 <template>
   <KSelect
+    v-model="selected"
     :label="props.label"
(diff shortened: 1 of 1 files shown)


schogges (approved):
Nice catch! 👍

PR #3509, opened 2025-02-04
Feature/zone traffic drawers

johncowen:
Whilst there looks like there is a lot going on here, its mainly copy/pasted from the similar view over in Dataplanes and most of the code here was pre-existing.

---

This PR replicates the Dataplane Traffic drawers from over in the dataplanes section into both ZoneIngress and ZoneEgress.

I began by trying to make the code used for Dataplanes more reusable so I coulduse it for both ZoneIngress and ZoneEgress. I found that for several reasons this was overly complicated to do and would have required quite a lot of change within Dataplanes also.

For instance:

- explicitly named route params such as `params.dataPlane`, `params.zoneIngress` and `params.zoneEgress`
- explicitly named route props such as `props.zoneIngress` and `props.zoneEgress`
- explicitly named source URIs `zone-ingress`/`zone-egress` (this one was relatively easy to address and is addressed in this PR)
- Dataplane Proxy URIs require a `mesh` specifying whereas Zone Proxy URIs don't
- ...probably more edge cases

Therefore I chose to mostly copy/paste over the approach from Dataplanes to ZoneIngress, and then again to ZoneEgress (ZoneIngress and ZoneEgress modules are mostly copy pasted anyway).

I plan to make some follow up PRs to reduce the repetition here.

I'll try and add some inlines, but please ask questions if anything isn't clear.

### DIFF
--- a/packages/kuma-gui/src/app/zone-egresses/data/index.ts
+++ b/packages/kuma-gui/src/app/zone-egresses/data/index.ts
@@ -8,39 +8,21 @@ import type {
 
 type PartialInternalZoneEgress = PartialZoneEgressOverview['zoneEgress']
 
-// TODO(jc) Theres probably a better way to not copy/pasta this i.e. make `Entity` composable
-type InternalZoneEgress = {
-  socketAddress: string
-} & PartialInternalZoneEgress
-
-export type ZoneEgress = {
-  config: PartialZoneEgress
-  socketAddress: string
-} & PartialZoneEgress
-// end TODO
-
 export type ZoneEgressInsight = PartialZoneEgressInsight & DiscoverySubscriptionCollection & {}
 
-// TODO(jc) Theres probably a better way to not copy/pasta this i.e. make `Entity` composable
-const InternalZoneEgress = {
-  fromObject: (item: PartialInternalZoneEgress): InternalZoneEgress => {
-    return {
-      ...item,
-      socketAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
-    }
-  },
-}
-
 export const ZoneEgress = {
-  fromObject: (item: PartialZoneEgress): ZoneEgress => {
+  fromObject: (item: PartialZoneEgress) => {
     return {
       ...item,
       config: item,
       socketAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
+      networking: {
+        ...item.networking,
+        inboundAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
+      },
     }
   },
 }
-// end TODO
 export const ZoneEgressInsight = {
   fromObject: (item: PartialZoneEgressInsight | undefined): ZoneEgressInsight => {
     return {
@@ -49,6 +31,20 @@ export const ZoneEgressInsight = {
     }
   },
 }
+// TODO(jc) Theres probably a better way to not copy/pasta this i.e. make `Entity` composable
+const InternalZoneEgress = {
+  fromObject: (item: PartialInternalZoneEgress) => {
+    return {
+      ...item,
+      socketAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
+      networking: {
+        ...item.networking,
+        inboundAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
+      },
+    }
+  },
+}
+
 export const ZoneEgressOverview = {
   fromObject: (item: PartialZoneEgressOverview) => {
     const zoneEgressInsight = ZoneEgressInsight.fromObject(item.zoneEgressInsight)
@@ -86,3 +82,4 @@ export const ZoneEgressOverview = {
 }
 
 export type ZoneEgressOverview = ReturnType<typeof ZoneEgressOverview.fromObject>
+export type ZoneEgress = ReturnType<typeof ZoneEgress.fromObject>
--- a/packages/kuma-gui/src/app/zone-egresses/routes.ts
+++ b/packages/kuma-gui/src/app/zone-egresses/routes.ts
@@ -14,7 +14,59 @@ export const routes = (prefix = 'egresses') => {
             path: 'overview',
             name: 'zone-egress-detail-view',
             component: () => import('@/app/zone-egresses/views/ZoneEgressDetailView.vue'),
-            children: subscriptions('zone-egress'),
+            children: [
+              ...((prefix) => {
+                return [
+                  {
+                    path: 'inbound/:connection',
+                    name: `${prefix}-connection-inbound-summary-view`,
+                    component: () => import('@/app/zone-egresses/views/ConnectionInboundSummaryView.vue'),
+                    children: [
+                      {
+                        path: 'stats',
+                        name: `${prefix}-connection-inbound-summary-stats-view`,
+                        component: () => import('@/app/zone-egresses/views/ConnectionInboundSummaryStatsView.vue'),
+                      },
+                      {
+                        path: 'clusters',
+                        name: `${prefix}-connection-inbound-summary-clusters-view`,
+                        component: () => import('@/app/zone-egresses/views/ConnectionInboundSummaryClustersView.vue'),
+                      },
+                      {
+                        path: 'xds-config',
+                        name: `${prefix}-connection-inbound-summary-xds-config-view`,
+                        component: () => import('@/app/zone-egresses/views/ConnectionInboundSummaryXdsConfigView.vue'),
+                      },
+                    ],
+                  },
+
+                  {
+                    path: 'outbound/:connection',
+                    name: `${prefix}-connection-outbound-summary-view`,
+                    component: () => import('@/app/zone-egresses/views/ConnectionOutboundSummaryView.vue'),
+                    children: [
+                      {
+                        path: 'stats',
+                        name: `${prefix}-connection-outbound-summary-stats-view`,
+                        component: () => import('@/app/zone-egresses/views/ConnectionOutboundSummaryStatsView.vue'),
+                      },
+                      {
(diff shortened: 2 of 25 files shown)


packages/kuma-gui/src/app/zone-egresses/data/index.ts
```diff
@@ -8,39 +8,21 @@ import type {
...
-  fromObject: (item: PartialZoneEgress): ZoneEgress => {
+  fromObject: (item: PartialZoneEgress) => {
     return {
       ...item,
       config: item,
       socketAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
+      networking: {
+        ...item.networking,
+        inboundAddress: item.networking?.address && item.networking?.port ? `${item.networking.address}:${item.networking.port}` : '',
+      },
```
johncowen, line 22: I myself have questions over `socketAddress` and `inboundAddress`. Zones already had `socketAddress` but Dataplanes don't. Dataplanes use `networking.inboundAddress` for this feature, so I copied that over here. In a later PR I may change this slightly and try to organize it so it works for all "proxy types" (all proxy types being ingress, egress _and_ dataplane)

packages/kuma-gui/src/app/zone-egresses/routes.ts
```diff
@@ -14,7 +14,59 @@ export const routes = (prefix = 'egresses') => {
...
+                      {
+                        path: 'xds-config',
+                        name: `${prefix}-connection-outbound-summary-xds-config-view`,
+                        component: () => import('@/app/zone-egresses/views/ConnectionOutboundSummaryXdsConfigView.vue'),
+                      },
+                    ],
+                  },
+
+                ]
+              })('zone-egress'),
```
johncowen, line 67: Perhaps temporarily _not_ using the `connections` module routes and view.vue files.

johncowen (commented):
Notes:

packages/kuma-gui/src/app/zone-ingresses/views/ConnectionInboundSummaryClustersView.vue
```diff
@@ -0,0 +1,66 @@
...
+              @filter-mode-change="route.update({ codeFilter: $event })"
+              @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+            >
+              <template #primary-actions>
+                <XAction
+                  action="refresh"
+                  appearance="primary"
+                  @click="refresh"
+                >
+                  Refresh
```
schogges, line 51: Another i18n `common.refresh`? 🙂

packages/kuma-gui/src/app/zone-egresses/views/ConnectionOutboundSummaryView.vue
```diff
@@ -0,0 +1,63 @@
...
+    :params="{
+      connection: '',
+      inactive: false,
+    }"
+    v-slot="{ route, t }"
+  >
+    <AppView>
+      <template #title>
+        <h2>
+          Outbound {{ route.params.connection }}
```
schogges, line 13: Should we add this to i18n?

packages/kuma-gui/src/app/zone-ingresses/views/ConnectionInboundSummaryStatsView.vue
```diff
@@ -0,0 +1,68 @@
...
+            @filter-mode-change="route.update({ codeFilter: $event })"
+            @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+          >
+            <template #primary-actions>
+              <XAction
+                action="refresh"
+                appearance="primary"
+                @click="refresh"
+              >
+                Refresh
```
schogges, line 50: Another i18n `common.refresh`?

packages/kuma-gui/src/app/zone-ingresses/views/ConnectionOutboundSummaryClustersView.vue
```diff
@@ -0,0 +1,66 @@
...
+              @filter-mode-change="route.update({ codeFilter: $event })"
+              @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+            >
+              <template #primary-actions>
+                <XAction
+                  action="refresh"
+                  appearance="primary"
+                  @click="refresh"
+                >
+                  Refresh
```
schogges, line 51: Another i18n `common.refresh`?

packages/kuma-gui/src/app/zone-ingresses/views/ConnectionOutboundSummaryStatsView.vue
```diff
@@ -0,0 +1,65 @@
...
+            @filter-mode-change="route.update({ codeFilter: $event })"
+            @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+          >
+            <template #primary-actions>
+              <XAction
+                action="refresh"
+                appearance="primary"
+                @click="refresh"
+              >
+                Refresh
```
schogges, line 49: Another i18n `common.refresh`?

schogges (commented):
Nice! LGTM 👍 
just a few nits and minor things 🙂

packages/kuma-gui/src/app/zone-egresses/views/ConnectionInboundSummaryStatsView.vue
```diff
@@ -0,0 +1,68 @@
...
+      <DataLoader
+        :src="uri(sources, '/connections/stats/for/:proxyType/:name/:socketAddress', {
+          name: route.params.zoneEgress,
+          socketAddress: props.networking.inboundAddress,
+          proxyType: 'zone-egress',
+        })"
+        v-slot="{ data: stats, refresh }"
+      >
+        <DataCollection
+          :items="stats!.raw.split('\n')"
```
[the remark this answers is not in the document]
johncowen: Thanks for checking! Sounds like this is no longer needed, but as its copy/paste the very next PR after this one will be removing this in Zones and where I copied it from (Dataplanes), just trying to stay in scope of zones here.

packages/kuma-gui/src/app/zone-egresses/views/ConnectionInboundSummaryStatsView.vue
```diff
@@ -0,0 +1,68 @@
...
+            @filter-mode-change="route.update({ codeFilter: $event })"
+            @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+          >
+            <template #primary-actions>
+              <XAction
+                action="refresh"
+                appearance="primary"
+                @click="refresh"
+              >
+                Refresh
```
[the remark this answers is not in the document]
johncowen: We are gonna circle back on these i18n things, will resolve all these for now

johncowen:
Ok I've added a commit which begins to address some of the complications around duplication, the extra commit specifically addresses the first point in the PR description:

> - explicitly named route params such as params.dataPlane, params.zoneIngress and params.zoneEgress

We went through offline, but TLDR is this commit changes and `route.params.zoneIngress`/`route.params.zoneEgress` to instead use `route.params.proxy` which is generic enough to refer to both type of zone proxy.

There was a little extra plumbing required to move it to use `.proxy` but we've moved one step forwards in that we no have a set of 'connections' view files that are shared between ZoneIngress/ZoneEgress.

I'm going to continue working, but in my mind this PR is still fine to go in. If its still here when I get to the next stopping point, I can either add a commit on the end if not I'll just make a new PR.

Apologies that this PR takes a bit of a haphazard approach from a PR/commit perspective, I'm trying to keep the PR scoped, as small as possible but still provide the backstory to the changes being made!

packages/kuma-gui/src/app/connections/views/zones/ConnectionInboundSummaryClustersView.vue
```diff
@@ -0,0 +1,67 @@
...
+              @filter-mode-change="route.update({ codeFilter: $event })"
+              @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+            >
+              <template #primary-actions>
+                <XAction
+                  action="refresh"
+                  appearance="primary"
+                  @click="refresh"
+                >
+                  Refresh
```
schogges, line 52: Could we use i18n `common.refresh`? 🙂

schogges (approved):
LGTM and I really like the deduplication 👍 🙂 
There are still some of the i18n things, but I'll approve anyways to unblock this since these are just nits 🙂

packages/kuma-gui/src/app/connections/views/zones/ConnectionInboundSummaryClustersView.vue
johncowen: Honestly I would rather do these as a "lets i18n some remaining text" PRs. If I do them here, then we are likely to miss other places that this has been duplicated from, or just do it badly without taking other things into account.

If we had to i18n the entire app, it wouldn't take us long to do, and I believe there are things that would actually prevent us from having the entire app properly i18n'ed (text in other dependencies that we don't control)

So for me it still stands that I want to stop digging little holes in all the wrong places, and potentially making wide ranging mistakes.

We are a little further along than we were now, we have `<XI18n />` which also potentially gives us prefixed keys. But I'd still prefer to wait a little longer before we do some "lets i18n some remaining text" PRs.

If it comes to it that we need to have our app i18n'ed asap, then I'd be happy to go through and do the remaining texts right now, but I don't think we are at that stage yet.

johncowen: Let me know on this one, I'd be happy to change this bit of text if you felt strongly about it, but all in all I'd prefer to do a piece of work specifically scoped to https://github.com/kumahq/kuma-gui/issues/2813 (having thought about https://github.com/kumahq/kuma-gui/issues/2814 first)

schogges: Fair points, I'm fine either way 🙂

PR #3511, opened 2025-02-05
chore: remove some old ts exclamation marks from envoy related views

johncowen:
See https://github.com/kumahq/kuma-gui/pull/3509#discussion_r1942548612


schogges (approved):
👍

Issue #3452, opened 2025-01-23
Change "XDS Configurations" to say "XDS Connections"


johncowen:
### Description

Also add an infoicon/tooltip to say something like "Connections to the GlobalCP"

PR #3514, opened 2025-02-06
chore: XDS Configurations > XDS Connections

johncowen:
Changes the text of `XDS Configurations` titles to say `XDS Connections`. This is mainly the title for some nest tables that we have, but also I removed the word `Configuration` from the Subscription Summary Drawers and just left it blank.

Leaving it blank meant I needed a new right alignment `XLayout` so I following flexbox I added `start`, `around`, `between` and `end` for our `separated` layout, and also (importantly) made all `separated` layouts take up 100% of their parents width (similar to a native div).

Closes https://github.com/kumahq/kuma-gui/issues/3452

### DIFF
--- a/packages/kuma-gui/src/app/x/components/x-layout/XLayout.vue
+++ b/packages/kuma-gui/src/app/x/components/x-layout/XLayout.vue
@@ -1,7 +1,7 @@
 <template>
   <component
     :is="props.type === 'separated' && props.truncate ? KTruncate : 'div'"
-    :class="['x-layout', props.type, props.size]"
+    :class="['x-layout', props.type, props.size, props.justify]"
   >
     <slot name="default" />
   </component>
@@ -12,10 +12,12 @@ const props = withDefaults(defineProps<{
   // TODO(jc) :variant
   type?: 'stack' | 'separated' | 'columns'
   size?: 'small' | 'normal' | 'max'
+  justify?: 'start' | 'around' | 'between' | 'end'
   truncate?: boolean
 }>(), {
   type: 'stack',
   size: 'normal',
+  justify: 'start',
   truncate: false,
 })
 </script>
@@ -26,17 +28,25 @@ const props = withDefaults(defineProps<{
 .stack.small > * + * {
   margin-block-start: $kui-space-40;
 }
-.max {
-  width: 100%;
-}
 .separated:not(.k-truncate) {
   display: inline-flex;
   flex-wrap: wrap;
   align-items: center;
   gap: $kui-space-40;
+  width: 100%;
 
-  &.max {
+  &.start {
+    justify-content: flex-start;
+  }
+  &.max,
+  &.between {
     justify-content: space-between;
+  }
+  &.around {
+    justify-content: space-around;
+  }
+  &.end {
+    justify-content: flex-end;
   }
 }
 .columns {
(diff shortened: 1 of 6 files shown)


packages/kuma-gui/src/app/x/components/x-layout/XLayout.vue
```diff
@@ -12,10 +12,12 @@ const props = withDefaults(defineProps<{
   // TODO(jc) :variant
   type?: 'stack' | 'separated' | 'columns'
   size?: 'small' | 'normal' | 'max'
+  justify?: 'start' | 'around' | 'between' | 'end'
```
schogges, line 15: Hmmm, I am wondering if we should continue with the approach of adding props that we turn into utility classes and combine them in the css. At some point this might become hard to understand and hard to maintain. Probably the `XLayout` component has to deal with too many cases? Should we rather go with the tailwind approach and directly add things into the `class` prop and keep the utility classes in `XLayout` separated? 🤔 When I added `max` for `size` my gut feeling already told me that this is going into a unsatisfactory situation 😅

schogges: > Totally want to avoid Tailwind, if anything makes something hard to maintain its Tailwind.

Don't get me wrong, I don't want to start using tailwind itself, just thought about adopting the approach of putting things together like tailwind just for `XLayout`. Meaning that I'd like to avoid the having the conditions in the css part and stop nesting things. Maybe the tailwind approach is not what we want, but then maybe something like a single `variant` prop that offers some layouts.

But yeah, maybe it's too early to judge 🙂 

> One good thing about this approach is that we get autocompletion, all our layout classes are contained in a centralized but non-globally scoped place also, which is probably one of the biggest reasons for doing this

💯 on this

> Honestly, I think XLayout is going reasonably well. My only qualm at the moment is its a little more difficult to add spacing to `ul li`s because we can't put a `div` in a `ul` (solvable if necessary with a `tagName` prop). At the moment it kinda looks like XLayout is pretty much turning into componentized flex-box which I'm more than fine with.

Yeah, I like to have this in one place and we can deal with that there 👍 and I think at the moment `XLayout` is going well. Just wanted to raise concerns early 🙂

schogges (approved):
LGTM 👍

packages/kuma-gui/src/app/x/components/x-layout/XLayout.vue
johncowen: Ok cool, gotcha, thanks for the expanding.

I'm guessing you are also in two minds about it as I feel like these two contradict each other a little.

> just thought about adopting the approach of putting things together like tailwind just for XLayout
> but then maybe something like a single variant prop that offers some layouts.

---

I think we do have two 'main' variants here and I have considered doing `hbox`/`vbox` like old school XUL. Our `stacked` is pretty much `vbox` and `separated` is pretty much `hbox`. If we split this into two components in a similar way we would have these `<hbox variant="end">` with a single variant etc. But do we want two components to signify direction?

Alternatively we could have `<XLayout variant="vertical-end">` and `<XLayout variant="horizontal-end"`>, but we may as well have two props one for direction and one for spacing, just like flex-box.

Thing is, and this is the thing with flexbox also, most layouts whether they are vertical or horizonal (i.e. columns or rows) also use justification and they apply equally to both just in different directions.

So for our `stacked`, `separated` these justify settings could apply equally to both (we only really need them for `separated` tho at least for now), to be honest it wouldn't surprise me if they could be relevant to `columns` also, just I would guess we don't really need them there.

I guess all in all having just a single `XLayout` for now to me is fine, its simple to use as its a single component for layout and at the moment at least its just as simple on the inside, its just a bit of CSS (which is centralized and non-global).

All in all, totally agree its early to judge, but the only significant downside I see as of now is having to use a `div` (currently), we may find more as we go.

PR #3515, opened 2025-02-06
chore(zones): reuse Clusters/XDS Configuration views from `connections`

johncowen:
Reuses the `Clusters` and `XDSConfig` views from `connections`, meaning we can then delete these views from both `zone-ingresses` and `zone-egresses`.

- `Stats` views are more complicated to reuse and will require more changes within `data-planes` so I left this for a subsequent PR (coming soon)
- There is a tiny but of change inside of the Clusters summary views, but just enough so I could get this stage done. I'm hoping there will be an upcoming PR the removes the recently added `connections/views/zones/*` files and reuses the ones in `connections/views/*` instead - things are becoming clearer and looking more possible the more I make progress on this piece of work.

Probably the biggest thing to point out here is, we want to reuse the sources between Dataplanes and ZoneProxies. As ZoneProxies aren't 'namespaced' within a mesh there is a small sharp edge in making them use the same thing (the same URI). Therefore we say ZonesProxies apply across all meshes i.e. `*` e.g. `mesh: '*'`


schogges (approved):
Nice! LGTM 👍

PR #3517, opened 2025-02-06
chore(zones): dedupe ZoneStats views and use ConnectionStats instead

johncowen:
Continuation of https://github.com/kumahq/kuma-gui/pull/3515

This PR removes the final `Stats` 'networking' type view from Zone*gresses and reuses the one in `connections` instead.

This one required some minor changes related to dataplanes, just so dataplanes and ZoneProxies could all share the same `source` URIs.

There should be no logic changes here in `sources` but I just included some minor conditionals in the final shared `source`/URI to cope with the slight differences between dataplanes and zoneproxies. I didn't pay too much attention to why they are different right now, but there is a chance I might be able to reduce whatever the differences are for at a later date.

Once this PR goes in, all _top level_ Stats/Clusters/XDSConfig views all use the same View.vue files. I'm going to try to reduce the duplication of the similar summary drawers views in `connections/views/*View.vue` / `connections/views/zones/*View.vue` next. I'm hoping when I stare at it some more now it will seem less complex.


schogges (approved):
Nice 🙂  LGTM 👍

PR #3518, opened 2025-02-06
feat: wrap nested route features in x-card

schogges:
In addition to wrap tables nested in routes in `XCard`, I decided to wrap all nested features in `XCard`. Also some of the titles of these features where not part of the `XCard` and I moved them to the `#title` slot. This contributes to overall consistency.

---

Apart from moving things to `XCard` I moved the warning about missing tls in the `DataPlaneDetailView` to the other warnings, which improves visibility on that as it's now at the top of the page.

image

Some more examples:

image
image

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -428,8 +428,6 @@
           </RouterView>
 
           <div data-testid="dataplane-mtls">
-            <h2>{{ t('data-planes.routes.item.mtls.title') }}</h2>
-
             <template
               v-if="props.data.dataplaneInsight.mTLS"
             >
@@ -442,6 +440,10 @@
                 <XCard
                   class="mt-4"
                 >
+                  <template #title>
+                    {{ t('data-planes.routes.item.mtls.title') }}
+                  </template>
+
                   <div class="columns">
                     <DefinitionCard>
                       <template #title>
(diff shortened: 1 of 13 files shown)


packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -502,82 +504,74 @@
...
-              >
-                {{ item.version?.kumaDp?.version ?? '-' }}
-              </template>
-              <template
-                #connected="{ row: item }"
-              >
-                {{ t('common.formats.datetime', { value: Date.parse(item.connectTime ?? '') }) }}
+            <XCard>
+              <template #title>
+                {{ t('data-planes.routes.item.subscriptions.title') }}
```
johncowen, line 515: Did I understand right that moving these here means we lose the semantic `h2` tag?

packages/kuma-gui/src/app/gateways/views/DelegatedGatewayDetailView.vue
```diff
@@ -64,181 +64,181 @@
           </XAboutCard>
         </DataLoader>
 
-        <div>
-          <h3>{{ t('delegated-gateways.detail.data_plane_proxies') }}</h3>
+        <XCard
+          class="mt-4"
```
johncowen, line 68: Is XLayout::stack not working here?

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
```diff
@@ -37,6 +37,8 @@ common:
   warnings:
     CERT_EXPIRED: !!text/markdown |
       The certificate for this dataplane has expired
+    DPP_NO_MTLS: !!text/markdown |
+      This Data Plane Proxy does not have mTLS configured, yet  — <a href="{KUMA_DOCS_URL}/policies/mutual-tls?{KUMA_UTM_QUERY_PARAMS}">Learn about certificates in {KUMA_PRODUCT_NAME}</a>
```
johncowen, line 41: How come this is in the kuma module and not the data-planes module? The key specifically mentions DPP so I'm guessing we aren't using this text for anything but DPPs?

packages/kuma-gui/src/app/policies/views/PolicyDetailView.vue
```diff
@@ -99,124 +99,124 @@
           </template>
         </DefinitionCard>
       </XAboutCard>
-      <div>
-        <h3>
+      
+      <XCard
+        class="mt-4"
+      >
+        <template #title>
```
johncowen, line 106: I think the `mt-4` here is no longer needed, if I'm right could you remove that while we are here? ty! (lemme know if its still needed tho)

packages/kuma-gui/features/mesh/dataplanes/DataplaneDetailsDelegatedGateway.feature
```diff
@@ -53,4 +53,4 @@ Feature: Dataplane details for delegated gateway
       |       193.107.134.106 |
       | kuma.io/protocol:http |
       | kuma.io/zone:zone-1   |
-    And the "$warnings" element doesn't exist
+    And the "$warnings" element exists
```
johncowen, line 56: Could you explain to me why these two assertions have reversed even though we've changed no test data? I'm not quite seeing what's happening here, ty!

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
schogges: Oh, yes that was not intended. Thanks for catching this 👍

packages/kuma-gui/src/app/gateways/views/DelegatedGatewayDetailView.vue
schogges: I'll just remove that class. It's not needed because `XCard` has it's own margin set.

packages/kuma-gui/src/app/policies/views/PolicyDetailView.vue
schogges: Yes, you are right 👍 `XCard` brings its own margin 🙂

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
schogges: Previously this warning was shown in the TLS section of the `DataPlaneDetailView`. I've moved it up and located it next to the other warnings. I had to move this to the `common.warnings` because the `#notifications` template loops over the warnings and references the translations like `common.warnings.${warning.kind}`. I'll try to use `:default-path`.

schogges: Hmmm, when I moved `DPP_NO_MTLS` to `data-planes.warnings.DPP_NO_MTLS` and used `:default-path="`data-planes.warnings.${warning.kind}`"` I get an error for the other `warning.kind`s which are not defined in `data-planes.warnings` 🤔

packages/kuma-gui/features/mesh/dataplanes/DataplaneDetailsDelegatedGateway.feature
schogges: Because we've set `mTLS: !!js/undefined` as a response, the warning `DPP_NO_MTLS` is shown in the `$warnings` section. To be able to keep the old assertion, I'd have to mock mtls here.

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
johncowen: Oh I see! I think 🤔 ..

Umm, not totally sure on this I'd have to look closer, but maybe we need to move those other warnings out of common and into data-planes so we have a common 'root' i.e. `dataplane.warnings`

Umm maybe we should leave this one as is for now and resolve this as it was and circle back?

packages/kuma-gui/features/mesh/dataplanes/DataplaneDetailsDelegatedGateway.feature
johncowen: Ohhh, I see! Because now the $warnings DOM element we are using for the assertion potentially has a new warning in it (previously it was in a totally different area of the page).

Ok, umm

I suppose my question is, if this test was originally supposed to be asserting that the warnings weren't shown when we are in a specific state, I would say we should change the state so that it requires no warnings, and then continue to assert that there are no warnings. i.e. keep the original assertion

Then, if we want an assertion to test that a warning does show when `mTLS: !!js/undefined` we should add that separately.

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
johncowen: Just to be clear on this one, I'd say do whatever works on this one for now, we can mess with the "where things should live ideally" later on.

packages/kuma-gui/src/app/zones/views/ZoneDetailView.vue
```diff
@@ -113,10 +113,13 @@
             </DefinitionCard>
           </XAboutCard>
 
-          <div
+          <XCard
```
johncowen, line 116: I love that we are slowly getting to an application with no `div`s at all. 🎉

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
schogges: Yeah I'd agree. It's not ideal, but this causes the least amount of changes for now. And honestly I was surprised how `:default-path` (`defaultMessage` under the hood) caused these issues. I'd expect that `defaultMessage` is only evaluated if there is no match in `:path`.

packages/kuma-gui/features/mesh/dataplanes/DataplaneDetailsDelegatedGateway.feature
schogges: Ok makes sense 🙂 these tests are not specifically testing for the `$warning` not to be shown, but just asserting that the page shows what's expected. From the given response this is what we expect to see now, i.e. the warning is shown. I'll try to mock the `mTLS` field, such that we can assert for no warnings shown, and add another test that asserts for the warnings.

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
schogges: I think that in the `XI18n` component internally the `defaultMessage` with `safeT` is getting evaluated before the wrapping `safeT`. The expected behaviour would be the other way around. For repro, I've tried to use the following:

```vue
<XI18n
  :path="`common.warnings.${warning.kind}`"
  :params="warning.payload"
  :default-path="`data-planes.warnings.${warning.kind}`"
/>
```
This should crash even when leaving all the locales in the current places.

schogges: After chatting offline, we agreed that this is expected behaviour and `:default-path` SHOULD throw if there is no match.

schogges: We also agreed to leave `DPP_NO_MTLS` in place where it is now and circle back later when we restructure the locales.

johncowen (approved):

PR #3519, opened 2025-02-06
chore: add boolean default storage route.params

johncowen:
Builds on the work we did in https://github.com/kumahq/kuma-gui/pull/3457 to allow us to persist default values of a `boolean` type.

I then used this for `inactive` (used in traffic views) and `includeEds` (used in XDSConfig views)


schogges (approved):
Nice 🙂 👍

PR #3520, opened 2025-02-06
feat(summary-view): use correct icons in dpp, in- and egress summary views

### LINKED ISSUE #3454: Change ZoneProxy icons to use a world icon instead of a location icon
### Description

See title. Note: This is about ZoneProxies not ZoneCPs which use either a kubernetes or "box" icon.

schogges:
This updates the icons we use in some of our summary drawers according to what we use in other areas:

- zone icon in zone ingress and zone egress summary views
- gateway icon in data plane summary views

image
image
image
image

Closes #3454


johncowen (approved):
Looks great. I had a few questions/suggestions as potential follow ups, I think we briefly discussed one at the end of last week:

1. Similarly to sidecar dpps and gateway dpps having slightly different icons to illustrate their type, zones have a type of "kubernetes"/"universal" that have different iconography. Should we use that in the summaries also? (I guess we should switch out the universal "box" icon for the world looking thing). ~I'm just checking with backend to make sure we can definitely assume a zone proxy in a kubernetes zone is definitely also classed as kubernetes.~ Confirmed that we can assume this.
2. How easy is it to add a tooltip to the icons in the summary, similar to what we do with the icons in the list views.
3. I've no idea why these summaries were ever grey looking, we should probably check that they should be grey and if not make all our icons consistently colored.

I've approved anyway as I think this is a good improvement in itself, and the above can be additions to that (if we want to act on those)

PR #3543, opened 2025-02-11
chore(networking): make all zones views use the same views as dataplanes

johncowen:
Continuation of:

- https://github.com/kumahq/kuma-gui/pull/3509
- https://github.com/kumahq/kuma-gui/pull/3512
- https://github.com/kumahq/kuma-gui/pull/3515
- https://github.com/kumahq/kuma-gui/pull/3517

Finally removes all duplicated views in `connections/views/zones/*.View.vue`

There is still a tiny thing I would like to resolve and that is the difference/usage/duplication of `socketAddress` and `listenerAddress`. ZoneProxies have always had the concept of a `socketAddress` and Dataplanes have always had `listenerAddress` they are pretty much the same thing. This tiny thing is now mainly a data layer thing now though, so happy to circle back on this now that these data types (Dataplanes/ZoneProxies) are now all using the same views for `networking`/`connections`.


schogges (approved):
LGTM 👍

PR #3545, opened 2025-02-12
feat(ci): adds centralized engines workspace property

johncowen:
`kumahq/kuma-gui/package.json` now holds the centralized versions for `engines`.

Using any other versions within any other workspace which uses `@kumahq/config` and different  `engines` will throw a linting error.

We also now check the npm version during our preinstall checks and suggest a command to use to update it

I also bumped our npm version to `^10.8.2`, and our node version to `20.17`

---

Just incase happy to split this up into 2 PRs if we want the version bump separate to the linting code.


packages/config/src/mk/check.mk
```diff
@@ -1,11 +1,16 @@
...
 .PHONY: check/node
 check/node:
 	@node -v | grep $(NODE_VERSION) &> /dev/null || ( \
 		echo "Make sure node-$(NODE_VERSION) is installed (see .nvmrc)"; \
 		exit 1; \
 	)
+	@npm ls -g "npm@$(NPM_VERSION)" | grep "empty" > /dev/null && ( \
+		echo "Make sure npm@$(NPM_VERSION) is installed. Try npm install -g npm@$(NPM_VERSION)"; \
+		exit 1; \
+	) || (exit 0)
```
schogges, line 13: Just double checking if we still need this?

johncowen: Well I guess not right now maybe, but do we want to track a non-breaking versions of npm at any point? Or we want to make sure that you haven't got a downgraded your `npm` even though you are on node-v20.17

schogges: Yeah, engines is not really enforcing the versions it specifies. It just throws a warning and people can just ignore it. This is more of an enforcement. I'd say we can leave it, but I don't really agree with the suggestion `Try npm install -g ...` 😅

johncowen: I get why you are concerned about installing dependencies globally (in this case if you are using nvm its not global) but have you ever heard of an instance where when not using something like nvm or similar npm is not installed globally?

johncowen: The other thing to note is, this is half the reason I don't want to execute this command for you (similar to why we don't execute `nvm use` for you). It's totally up to you how you install the version of npm that we require you to have, we just give a suggestion.

schogges: > The other thing to note is, this is half the reason I don't want to execute this command for you (similar to why we don't execute `nvm use` for you). It's totally up to you how you install the version of npm that we require you to have, we just give a suggestion.

Yeah, that's the point. It's totally up to the user how to align the node and npm versions. Similarly to a node version mismatch where we point to the .nvmrc, we could point to the engines 🤔 
But fine, it's probably not too important. I'll leave it up to you 🙂

johncowen: Well I was about to say, we could also just add it in next time we consider bumping our npm past the version that comes with nvm, its at that point that we'd need this.

If its not doing any harm I'll leave it here, its not like we'll see it at all.

schogges: Yeah 😄 👍 feel free to leave it and resolve this 🙂

johncowen:
I noticed I forgot to actually run the thing in CI 😅 . I spotted a couple of other bits whilst looking at that here, so I'll do a quick follow up PR in sec

PR #3546, opened 2025-02-12
chore(ci): integrates further package linting

johncowen:
Integrates work from https://github.com/kumahq/kuma-gui/pull/3545 into CI.

Also, while I was there I did a few nitty cleanup things:

- made 99.9% of arrays a uses linewise YAML arrays consistently (i.e. `-` vs `[]`)
- consistently moved `concurrency:` to the same places in all workflows
- changed the name of the `main.yml` workflow to `main` (the name is used for cross-referencing in another workflow). It seemed to me to be a good idea to keep the name of the workflow the same as the name of the file.
- ...possibly another couple of teensy tiny nitty things.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -7,9 +7,13 @@ concurrency:
 
 on:
   push:
-    branches: [master, 'release-[0-9]+.[0-9]+']
+    branches:
+      - master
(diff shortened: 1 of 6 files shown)


schogges (approved):
LGTM 👍

Issue #3523, opened 2025-02-10
Add [Download bundle] button similar to Dataplanes to ZoneProxies also


johncowen:
### Description

See title

PR #3547, opened 2025-02-12
feat(zones): Add downloadable bundles to zone proxies

johncowen:
Similar to https://github.com/kumahq/kuma-gui/pull/3329 this PR adds downloadable "bundles" to both Zone Ingresses and Zone Egresses.

Pretty much a copy/pasta of the Dataplanes view. But I changed the Dataplanes view to use the key `proxies` instead of `dataplane` for consistency. Also, zone proxies don't have policies as an option to download.

Closes https://github.com/kumahq/kuma-gui/issues/3523


schogges (approved):
LGTM 👍

PR #3548, opened 2025-02-13
feat: add XDownload component for downloading blobs

johncowen:
Adds a new `XDownload` component for downloading file blobs easily.

I've currently used it/provided an example in our Dataplanes downloading functionality, but it will also be rolled out to ZoneProxies (also see https://github.com/kumahq/kuma-gui/pull/3547)

This PR can go in before (and https://github.com/kumahq/kuma-gui/pull/3547 be rebased) or after https://github.com/kumahq/kuma-gui/pull/3547.

---

One thing I was unsure of was the name of the `start` event. I went through MDN looking for a name that felt HTML-like, and came up with `start` if anyone has suggestions for a name, that I like better than `start`, please lemme know!

### DIFF
--- a/packages/kuma-gui/src/app/x/components/x-download/XDownload.vue
+++ b/packages/kuma-gui/src/app/x/components/x-download/XDownload.vue
@@ -1,0 +1,47 @@
+<template>
+  <slot
+    :download="download"
+    name="default"
+  />
+</template>
+<script lang="ts" setup>
+
+const props = withDefaults(defineProps<{
+  revokeAfter?: number
+}>(), {
+  revokeAfter: 60000,
+})
+
+defineSlots<{
+  default(props: {
+    download: typeof download
+  }): any
+}>()
+
+const emit = defineEmits<{
+  (event: 'start'): void
+}>()
+
+const download = async (
+  {
+    url,
+    name = `download_${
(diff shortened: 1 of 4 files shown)


packages/kuma-gui/src/app/x/components/x-download/XDownload.vue
```diff
@@ -0,0 +1,47 @@
...
+  (event: 'start'): void
+}>()
+
+const download = async (
+  {
+    url,
+    name = `download_${
+      new Date().toLocaleString('en-US', { year: 'numeric', month: '2-digit', day: '2-digit' }).split('/').reverse().join('-')
+    }_${
+      new Date().toLocaleString('en-US', { hour: '2-digit', hour12: false, minute: '2-digit', second: '2-digit' }).split(':').join('-')
```
schogges, line 31: Nothing important, but I was wondering if it makes sense to use `window.navigator.language` and just fall back to `en-US` 🤔

johncowen: Ah, I nearly fell into the same trap! We use `en-US` to make sure I get `/` which I then split on. other languages might not use `/`

also fyi, I might be wrong but I think passing `undefined` takes the local of the browser, which is what I was using at first.

schogges: Ah good point 😄 👍

schogges (approved):

PR #3549, opened 2025-02-13
chore(deps): overrides esbuild and runs npm dedupe

johncowen:
Adds an override for esbuild, then runs `npm dedupe` which dupes quite a few deps most importantly dedupes `esbuild` to 0.25.0.

I tried just using `npm install` but even with the override it wasn't sticking, `npm dedupe` on the otherhand did the trick.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
     "packages/*"
   ],
   "overrides": {
-    "@cucumber/messages": "^27.0.2"
+    "@cucumber/messages": "^27.0.2",
+    "esbuild": "^0.25.0"
   },
   "dependencies": {
     "@kumahq/config": "*"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1155,9 +1155,10 @@
       }
     },
     "node_modules/@cucumber/html-formatter": {
-      "version": "21.7.0",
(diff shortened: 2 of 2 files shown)


schogges:
> Dedupes quite a few deps most importantly dedupes `esbuild` to 0.25.0

Hmm 🤔 between the deduplications I can't find anything about `esbuild` being deduped. Is this correct?

johncowen:
Oh that was weird, I thought I"d managed to get it to bump esbuild without the override, but obviousy not. I've added the override again and also amended the title and desc of the PR. lmk!

package.json
```diff
@@ -3,7 +3,8 @@
     "packages/*"
   ],
   "overrides": {
-    "@cucumber/messages": "^27.0.2"
+    "@cucumber/messages": "^27.0.2",
+    "esbuild": "0.25.0"
```
schogges, line 7: Is it intended to set `esbuild` to fixed version `0.25.0`?

johncowen: 🤦 , thanks 😅  (need to update our linter to include overrides methinks!)

schogges (approved):
👍

Issue #2705, opened 2024-06-28
redesign/rebuild our warnings/notifications


johncowen:
### Description

![Screenshot 2024-06-28 at 11 45 05](https://github.com/kumahq/kuma-gui/assets/554604/b906f4f0-1855-4b69-92c5-1dfffa2dc7b9)

Automaat:
waiting on design

lahabana:
Let's go with: https://kongponents.konghq.com/components/alert.html

For the moment we know how to do external links (example in slot section).

Apparently internal links are underlined: https://github.com/Kong/kongponents/pull/2528

Design to provide a clear figma with all possible options (internal/external links, different level of alerts and multiple alerts within a page...)

lahabana:
Let's first ship without dismissal and work on dismissal as a follow up

PR #3566, opened 2025-02-18
feat: add XNotification component

johncowen:
This PR completely rebuilds our notifications and replaces all instances of our current ones to use the new approach. (I would guess there will be a further PR to amend where we put certain notifications, user-wise I've tried to keep things exactly the same for now)

### Different variants

Screenshot 2025-02-18 at 14 08 31

### Different "areas"

Screenshot 2025-02-18 at 14 08 43

### Multiple notifications of the same type

Screenshot 2025-02-18 at 14 11 00

---

The biggest question I have is surround notification dismissal.

For example, if I dismiss a notification about using the memory store for the `default` Mesh, I would guess I still would need to see a warning if I then created a new mesh called `johns-mesh`? If you carry this on, what about dismissal of notifications for a specific MeshExternalService?

I've tried to not to overthink it too much for now, and namespaced certain warnings (for example the Mesh example I gave above)

Closes https://github.com/kumahq/kuma-gui/issues/2705

### DIFF
--- a/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
+++ b/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
@@ -175,7 +175,7 @@
                     type="stack"
                   >
                     <template
-                      v-for="[variant, value] in hub.notifications.entries()"
+                      v-for="[variant, value] in hub.notifications"
                       :key="variant"
                     >
                       <XAlert
@@ -192,8 +192,11 @@
                           <li
                             v-for="notification in value"
                             :key="notification"
+                            :data-testid="`notification-${notification}`"
                           >
-                            <XNotification :uri="notification" />
+                            <XNotification
+                              :uri="notification"
+                            />
                           </li>
                         </ul>
                       </XAlert>
--- a/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
+++ b/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
@@ -7,7 +7,7 @@
     />
     <DataSource
       :src="`/me/~notifications`"
-      v-slot="{ data: dismissed, refresh }"
+      v-slot="{ data: dismissed }"
     >
       <XNotificationHub
         v-if="dismissed"
@@ -173,7 +173,6 @@
               >
                 <DataSink
                   :src="`/me/~notifications`"
-                  v-slot="{ submit }"
                 >
                   <XLayout
                     type="stack"
@@ -184,11 +183,6 @@
                     >
                       <XAlert
                         :variant="variant"
-                        @dismiss="async () => {
-                          submit(Array.from(value))
-                          await nextTick()
-                          refresh()
-                        }"
                       >
                         <ul
                           class="notifications"
@@ -219,7 +213,6 @@
   </div>
 </template>
 <script lang="ts" setup>
-import { nextTick } from 'vue'
 import GithubButton from 'vue-github-button'
 
 import { useEnv, useI18n, useCan } from '@/app/application'
--- a/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
+++ b/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
@@ -54,16 +54,20 @@
                     data-testid="upgrade-check"
                     appearance="info"
                   >
-                    <p>
-                      {{ t('common.product.name') }} update available
-                    </p>
-
-                    <XAction
-                      appearance="primary"
-                      :href="t('common.product.href.install')"
+                    <XLayout
+                      type="separated"
                     >
-                      Update
-                    </XAction>
+                      <p>
+                        {{ t('common.product.name') }} update available
+                      </p>
+
+                      <XAction
+                        appearance="primary"
+                        :href="t('common.product.href.install')"
+                      >
+                        Update
+                      </XAction>
+                    </XLayout>
                   </XAlert>
                 </DataSource>
               </div>
@@ -374,12 +378,10 @@ nav :deep(.app-navigator) > a {
     font-size: $kui-font-size-30;
   }
 
-  .upgrade-alert,
-  :deep(.alert-message) {
+  .upgrade-alert {
     display: flex;
     align-items: center;
     gap: $kui-space-50;
-
   }
 }
 </style>
--- a/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
+++ b/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
@@ -5,170 +5,214 @@
     <XTeleportSlot
       name="modal-layer"
     />
-    <header
-      role="banner"
+    <DataSource
+      :src="`/me/~notifications`"
+      v-slot="{ data: dismissed, refresh }"
     >
-      <div
-        class="horizontal-list"
+      <XNotificationHub
+        v-if="dismissed"
+        uri="application-shell"
+        :dismissed="dismissed"
+        v-slot="hub"
       >
-        <slot name="header">
-          <XAction :to="{ name: 'home' }">
-            <slot name="home" />
-          </XAction>
-
-          <GithubButton
-            class="gh-star"
-            href="https://github.com/kumahq/kuma"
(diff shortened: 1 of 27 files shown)


johncowen:
FYI: I've decided I'm gonna remove the dismiss buttons temporarily once we can some initial feeling of approval here. I figure the unique uris/ids we use that are going to be persisted in localStorage need to be changed to be more consistent. I'd rather do that in another PR as it needs a bit more thought. If I comment out the dismiss buttons, that means we never save anything in localStorage, which means we can merge this PR without folks adding loads of values to localStorage that we will later change.

packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
```diff
@@ -5,179 +5,230 @@
...
+                      v-for="[variant, value] in hub.notifications"
+                      :key="variant"
+                    >
+                      <XAlert
+                        :variant="variant"
+                        @dismiss="async () => {
+                          submit(Array.from(value))
+                          await nextTick()
+                          refresh()
+                        }"
```
johncowen, line 191: Same comment as above, this will be commented/removed before merge.

packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
```diff
@@ -5,179 +5,230 @@
...
-            href="https://github.com/kumahq/kuma"
-            aria-label="Star kumahq/kuma on GitHub"
+        <XNotification
+          v-if="!can('use state')"
+          uri="global-store-type-memory"
+        >
+          <XI18n
+            path="common.warnings.GLOBAL_STORE_TYPE_MEMORY"
+          />
+        </XNotification>
```
johncowen, line 25: I plan on moving this notification out of ApplicationShell and into an AppView once https://github.com/kumahq/kuma-gui/issues/3570 is done. but this shouldn't block approval/merge.

johncowen (commented):
Notes:

packages/kuma-gui/src/app/me/sources.ts
```diff
@@ -3,11 +3,23 @@ import merge from 'deepmerge'
...
 
 type Storage = {
-  get: (key: string) => Promise<object>
+  get: (key: string, d?: object) => Promise<object>
   set: (key: string, value: object) => Promise<object>
 }
 export const sources = ({ get, set }: Storage) => {
   return defineSources({
+    // used for saving dismissed notification id's
+    '/me/~notifications': async () => {
```
schogges, line 12: Is there a specific reason for the `~`? 🙂

schogges (approved):
Great stuff! 🚀 

Just for my complete understanding: Currently it wouldn't be possible to differentiate from a notification perspective to push it up to the "global" hub? Meaning inside an `<AppView :notifications="true"` there is currently no way to still let some notifications pass through?

packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
```diff
@@ -5,179 +5,230 @@
...
-          <XAction :to="{ name: 'home' }">
-            <slot name="home" />
-          </XAction>
-
-          <GithubButton
-            class="gh-star"
-            href="https://github.com/kumahq/kuma"
-            aria-label="Star kumahq/kuma on GitHub"
+        <XNotification
+          v-if="!can('use state')"
```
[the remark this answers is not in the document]
schogges: Not sure if I follow. You mean we need to add some flag to say that we are able to store the dismissal in localStorage?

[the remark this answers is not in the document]
johncowen: Bit of a potential idea this, not sure if its what I'll do.

I'll try explain my thinking a bit better: Moving forwards there are two reasons we'll need the code that is currently in `v-if` (but not use a `v-if`):

1. Decide whether to show the notification or not
2. Decide whether to not show the notification, and additionally remove a previous dismissal key from localStorage (i.e. I dismissed this, made a change to correct it, but then I undo the change, we want the notification to show again)

Because of point 2, we can't use `v-if` from outside the component any more because then XNotification won't be instantiated/mounted when we want to remove the it's key from localStorage.

I'm thinking something like:

```vue
<XNotification
    :uri="key-that-goes-in-local-storage"
    :active="!can('use state')"
>
```

When `active` is `true` we show the notification (bit like `v-if`), but when its `false`, via code inside of XNotifcation, we remove its key from localStorage.

If we house the code to do the "removal from localStorage" inside of XNotification (which would make sense), then the component still needs to be instantiated/mounted, just render nothing when it's `false` (via a `v-if` from inside XNotification instead of outside). When `active` is `true`, we render something but don't bother taking the key out of localStorage.

johncowen:
> Just for my complete understanding: Currently it wouldn't be possible to differentiate from a notification perspective to push it up to the "global" hub? Meaning inside an <AppView :notifications="true" there is currently no way to still let some notifications pass through?

I was thinking about this question after we spoke the other day, and I actually think it is possible, but you'd need to do a tiny bit more plumbing. I'll try and explain below.

I might add something later so you don't even have to do the plumbing, but at the moment I don't think we need this functionality - in saying that I might be able to improve on the `notifications=true` thing at a later date.

----

Currently NotificationHub is in ApplicationShell and AppView, you could potentially forget about the NotificationHub in AppView and instead pass down `hub` all the way from the top, drilling down all the way where you need it (or use `provide/inject` again to avoid all the prop drilling, which would be best):

```vue
/// App.vue
<!-- hub is exported here, we just need to get that into your nested route -->
<NotificationHub v-slot="hub"> 
  <!-- remember this is an inline component so its a -->
  <!-- placeholder for the notification not the notification itself -->
   <XNotification v-if="notificationUri === 'root-uri'" />
   <RouterView :hub="hub" />
</NotificationHub>
```

```vue
/// A deeply nested route where we have drilled or injected `hub`
<template v-for="notificationUri in hub.notifications">
  <!-- remember this is an inline component so its a -->
  <!-- placeholder for the notification not the notification itself -->
   <XNotification v-if="notificationUri === 'nested-route-uri'" />

   <XNotification :uri="root-uri">
      Notification Text for Root Alert
   </XNotification>
   <XNotification :uri="nested-root-uri">
      Notification for The Nested AppView Alert
   </XNotification>
</template>
```

The above would work right now with the code we have here - just we never currently need to define notifications in one place that go into two different "areas", plus we still have the issue of the notification in the root "area" that depends on the nested AppView disappearing when you navigate away from that AppView.

Hopefully that helps a bit on that one

---

Importantly with all this, it shouldn't be too hard to send our notifications to a completely different "notification hub" maybe in some sort of shell application without actually changing any code we have here.

johncowen:
Just gonna disable the dismissals real quick, brb

johncowen:
> Just gonna disable the dismissals real quick, brb

K ☝️ this is done in [3058d44](https://github.com/kumahq/kuma-gui/pull/3566/commits/3058d444b6c8131c717e5b5d51251fcee8019554)

packages/kuma-gui/src/app/me/sources.ts
schogges: Gotcha, thanks 👍

schogges:
> > Just for my complete understanding: Currently it wouldn't be possible to differentiate from a notification perspective to push it up to the "global" hub? Meaning inside an <AppView :notifications="true" there is currently no way to still let some notifications pass through?

> The above would work right now with the code we have here - just we never currently need to define notifications in one place that go into two different "areas", plus we still have the issue of the notification in the root "area" that depends on the nested AppView disappearing when you navigate away from that AppView.
> 
> Hopefully that helps a bit on that one
> 
> Importantly with all this, it shouldn't be too hard to send our notifications to a completely different "notification hub" maybe in some sort of shell application without actually changing any code we have here.

Perfect, thanks for the detailed explanation 👍 🙂

packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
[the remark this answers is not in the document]
schogges: Ah 💡  gotcha thanks 👍

PR #3567, opened 2025-02-18
chore(zones): uses XDownload for zone bundle downloading

johncowen:
Follow up of:

- https://github.com/kumahq/kuma-gui/pull/3547
- https://github.com/kumahq/kuma-gui/pull/3548

In https://github.com/kumahq/kuma-gui/pull/3548 we made XDownload but only added it to dataplanes (ZoneProxy bundling wasn't merged as yet)

This PS uses it for ZoneProxy bundling also.


schogges (approved):
👍

PR #3568, opened 2025-02-19
refactor(policies): use _resources instead of policies endpoint

schogges:
Instead of using the `/policies` endpoint we are now switching over to a more general `/_resources` endpoint that returns a list of all resources. The `/_resources` endpoint is typed through our openapi specs, so I could remove the custom types.
In order to keep the changes as minimal as possible, I've used the data-layer in the policies module to extract the policy types into a separated `policyTypes` entry, which makes it easier to use in the policy templates. Therefore I've also kept the source `/policy-types` and didn't rename it according to the changes.
The `/_resources` endpoint could be reused in other parts of the app that focuses on the non-policy entries and have a separate implementation in the respective data-layer.

---

For the mocks I've extended the `resourceNames` with more resource names and added an optional parameter to return all. As per default it returns a random subset. To be able to differentiate between policies and non-policies I've reused the kuma faker `policyName`, moved the items to a `policyNames` function such that the policy names are in one place.


packages/kuma-gui/src/app/policies/views/PolicyListView.vue
```diff
@@ -27,21 +27,14 @@
...
-                    v-if="type.isInbound"
+                    v-if="type.policy.hasFromTargetRef"
                     appearance="neutral"
                   >
                     {{ t('policies.collection.inbound') }}
                   </XBadge>
 
                   <XBadge
-                    v-if="type.isOutbound"
+                    v-if="type.policy.hasToTargetRef"
```
schogges, line 37: Note: not sure if this is correct 🤔 `isInbound` is `hasFromTargetRef` and `isOutbound` is `hasToTargetRef``

packages/kuma-gui/src/test-support/FakeKuma.ts
```diff
@@ -119,7 +119,10 @@ gbXR5RnEs0hDxugaIknJMKk1b0g=
     return this.faker.helpers.arrayElement<typeof items[number]>(items)
   }
 
-  policyName() {
+  /**
+   * Returns a random subset of policy names in random order. Optionally returns all.
+   */
+  policyNames({ includeAll }: { includeAll?: boolean } = { includeAll: false }) {
```
schogges, line 125: I think there are multiple places in our mocks (fakes and endpoints) that track policies and legacy policies. Was thinking of moving the legacy ones here aswell and optionally include them. 🤔

packages/kuma-gui/src/test-support/mocks/src/resources.ts
```diff
@@ -0,0 +1,44 @@
...
+    'RateLimit',
+    'Retry',
+    'Timeout',
+    'TrafficLog',
+    'TrafficPermission',
+    'TrafficRoute',
+    'TrafficTrace',
+    'VirtualOutbound',
+    'MeshGatewayRoute',
+  ]
```
johncowen, line 18: Should `fake.kuma.policyNames` optionally include/exclude legacy ones?

packages/kuma-gui/src/test-support/mocks/src/resources.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        return {
+          name,
+          scope,
+          readOnly: fake.datatype.boolean(),
+          path: ((_name) => `${_name.includes('Mesh') ? _name.toLocaleLowerCase() : _name.replace(/([a-z0-9])([A-Z])/g, '$1-$2').toLowerCase()}s`)(name),
+          singularDisplayName: name,
+          pluralDisplayName: `${name}s`,
+          includeInFederation: fake.datatype.boolean(),
+          ...([...legacyPolicies, ...policies].find((policy) => policy.includes(name) /* includes legacy policies */) && { policy: {
```
johncowen, line 34: In my eyes this looked a bit weird due to the placement of the comment just infront of the closing bracket, would you be ok moving the comment to the line above?

```suggestion
          // only for resources that are policy type ones, including legacy policy types
          ...([...legacyPolicies, ...policies].find((policy) => policy.includes(name)) && { policy: {
```

Something like that ☝️

johncowen (commented):
Generally looks good to me, glad we are starting to leave old policy things behind 🎉 

I left a few questions and comments, lmk on those, also I think you were commenting as I was so let me read over those also.

packages/kuma-gui/src/test-support/FakeKuma.ts
johncowen: Ah, 😆 I had the same comment, I agree!

packages/kuma-gui/src/app/policies/views/PolicyListView.vue
johncowen: I feel the same as you, kinda sounds like that is the case, but I'm not sure myself, would be good to check with backend to confirm.

lahabana: It's precisely this you can see from the back compat code:
https://github.com/kumahq/kuma/blob/304d2f864ca3ed3acf36a577d789363756a22b08/pkg/api-server/policies_endpoints.go#L28-L29

packages/kuma-gui/src/app/policies/data/index.ts
```diff
@@ -1,10 +1,14 @@
...
 import type { PaginatedApiListResponse } from '@/types/api.d'
 import type {
   PolicyDataplane as PartialPolicyDataplane,
   PolicyEntity as PartialPolicy,
 } from '@/types/index.d'
 
-export type { PolicyType } from '@/types/index.d'
+export type PartialResourceTypes = paths['/_resources']['get']['responses']['200']['content']['application/json']
+export type PartialResourceType = PartialResourceTypes['resources'][number]
+export type PartialPolicyResourceType = PartialResourceType & Required<Pick<PartialResourceType, 'policy'>>
```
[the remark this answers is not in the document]
schogges: Probably not, I'll check 🙂

packages/kuma-gui/src/app/policies/data/index.ts
```diff
@@ -62,3 +66,23 @@ export const Policy = {
   },
 }
 export type Policy = ReturnType<typeof Policy['fromObject']>
+
+export const PolicyResourceType = {
+  isPolicyType(o: PartialResourceType): o is PartialPolicyResourceType {
+    return typeof o.policy !== 'undefined'
+  },
```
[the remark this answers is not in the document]
schogges: I've added this and put it there because I thought it might be useful elsewhere. But surely I can make it inline 👍 🙂

packages/kuma-gui/src/app/policies/data/index.ts
```diff
@@ -62,3 +66,23 @@ export const Policy = {
...
+    return {
+      ...partialPolicyType,
+    }
+  },
+
+  fromCollection(partialResourceTypes: PartialResourceTypes) {
+    return {
+      ...partialResourceTypes,
+      policyTypes: partialResourceTypes.resources.filter(PolicyResourceType.isPolicyType).map(PolicyResourceType.fromObject),
+    }
```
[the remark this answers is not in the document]
schogges: Yes, but both `resources` and `policyTypes` are arrays

```ts
{
  ...<potential_other_stuff_returned_from__resources>,
  "resources": [<all_resources>],
  "policyTypes":: [<only_policy_resources>]
}
```

packages/kuma-gui/src/test-support/mocks/src/resources.ts
schogges: Will do that 👍 going to resolve this one in favor of the convo above (same topic) 🙂

packages/kuma-gui/src/app/policies/data/index.ts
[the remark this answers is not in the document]
schogges: Yup, referencing only from this file. Removed the exports.

packages/kuma-gui/src/test-support/FakeKuma.ts
```diff
@@ -455,25 +464,48 @@ gbXR5RnEs0hDxugaIknJMKk1b0g=
...
+      'RateLimit',
+      'Retry',
+      'Timeout',
+      'TrafficLog',
+      'TrafficPermission',
+      'TrafficRoute',
+      'TrafficTrace',
+      'VirtualOutbound',
+    ] as const
+    return this.faker.helpers.arrayElements(items, { min: includeAll ? items.length : 0, max: items.length })
```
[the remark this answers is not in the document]
schogges: Yes, that sounds really good 👍 will add this

packages/kuma-gui/src/app/policies/data/index.ts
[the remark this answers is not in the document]
schogges: Done. Resolving this one.

packages/kuma-gui/src/app/policies/data/index.ts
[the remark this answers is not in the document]
schogges: Done. Resolving this one.

packages/kuma-gui/src/test-support/FakeKuma.ts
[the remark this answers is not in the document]
schogges: Ah, I first misunderstood your suggestion. I guess you meant to add `min` and `max` as options to `policyNames` and `resourceNames` instead of `includeAll`? 🙂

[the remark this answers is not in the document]
schogges: I'll default to the following:

```ts
min: undefined => 0
max: undefined => Number.MAX_VALUE
```

That means in case of 

```ts
min: undefined, max: Number.MAX_VALUE: all of them
```

I would return a random subset and not all of them. All the other cases you mentioned will work as you described.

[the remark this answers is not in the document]
johncowen: ah yeah 😆  😅 , thats perfect!

packages/kuma-gui/src/test-support/mocks/src/resources.ts
```diff
@@ -0,0 +1,31 @@
...
+    headers: {},
+    body: {
+      resources: fake.kuma.resourceNames({ min: Number.MAX_VALUE }).map((name) => {
+        const scope = fake.helpers.arrayElement(['Mesh', 'Global'])
+
+        return {
+          name,
+          scope,
+          readOnly: fake.datatype.boolean(),
+          path: `${(name.includes('Mesh') ? name : name.replace(/([a-z0-9])([A-Z])/g, '$1-$2')).toLowerCase()}s`,
```
johncowen, line 16: Is this the same way the backend figures out paths?

johncowen (approved):
I left two more comments but I think they can either be addressed here or in a future PR, lmk on those but also feel free to merge and potentially circle back if thats easiest.

packages/kuma-gui/src/test-support/mocks/src/resources.ts
schogges: Looks like this is not consistent atm and I thought the pattern is what I've added here. But from what I can see that is actually not the case.

packages/kuma-gui/src/app/policies/data/index.ts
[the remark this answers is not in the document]
schogges: > I'm just wondering why the PolicyResourceTypes also returns the other policy types in resources.
>
>Should this be Resource.fromCollection => { resources: <all-resources>, policyResources: <only-policy-resources>, >serviceResources: <only-Mesh*Service>, ...}
>
>Then I can do data.policyResources, or data.serviceResources, or maybe data.internalResources (for want of a better >word) for things like Secret etc

>TLDR; Instead of PolicyResourceTypes should it be called Resources (and be a new resources/ module)

Yes, we can do that. As I described in the PR description I wanted to keep the changes minimal for now. I can tackle this in a follow-up 👍

packages/kuma-gui/src/test-support/mocks/src/resources.ts
johncowen: Let's circle back on this then, this is fine for now 👍

johncowen:
> Yes, we can do that. As I described in the PR description I wanted to keep the changes minimal for now. I can tackle this in a follow-up 👍

Totally, makes sense 🙌

packages/kuma-gui/src/test-support/mocks/src/resources.ts
schogges: I think for the mocks it doesn't make a difference, but I've reached out to backend and can circle back here. Lmk 🙂

PR #3571, opened 2025-02-19
feat(policy-type-list-view): make sticky

schogges:
Make the `PolicyTypeListView` sticky to the top. This is a minor improvement to UX as the `PolicyTypeListView` serves as a navigation bar and opens a nested route to the right, which is another potentially large table. When scrolling the table it can happen that the `PolicyTypeListView` scrolls out of the viewport. To reach the navigation again, the user might have to scroll up again.
As there is currently nothing else rendered below the `PolicyTypeListView`, there is nothing that potentially interferes.


johncowen (approved):
Nice improvement/detail 👏 👍

PR #3578, opened 2025-02-20
feat: updated xEmpty state and empty state content

afrzhussain:
### Summary
Updated XEmpty State
- Replaced KEmpty with EntityEmptyState
- Updated Empty state content

### DIFF
--- a/packages/kuma-gui/package.json
+++ b/packages/kuma-gui/package.json
@@ -13,6 +13,7 @@
   "dependencies": {
     "@gera2ld/tarjs": "^0.3.1",
     "@kong-ui-public/app-layout": "^4.3.10",
+    "@kong-ui-public/entities-shared": "^3.17.1",
     "@kong-ui-public/i18n": "^2.2.10",
     "@kong/icons": "^1.21.1",
     "@kong/kongponents": "^9.21.6",
--- a/packages/kuma-gui/package.json
+++ b/packages/kuma-gui/package.json
@@ -73,7 +73,7 @@
     "vite": "^5.4.11",
     "vite-svg-loader": "^5.1.0",
     "vitepress": "^1.6.3",
-    "vitest": "^3.0.5",
+    "vitest": "^3.0.6",
     "vue-tsc": "^2.2.0"
   },
   "browserslist": [
--- a/packages/kuma-gui/package.json
+++ b/packages/kuma-gui/package.json
@@ -63,7 +63,7 @@
     "markdown-it": "^14.1.0",
     "msw": "^2.7.0",
     "openapi-typescript": "^7.6.1",
-    "postcss": "^8.5.2",
+    "postcss": "^8.5.3",
     "postcss-html": "^1.8.0",
     "sass": "^1.85.0",
     "shiki": "^2.3.2",
--- a/packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
+++ b/packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
@@ -78,7 +78,7 @@
                 :name="`${props.type}-x-empty-state-actions`"
               />
               <XAction
-                v-if="href.length > 0"
+                v-if="t(`${prefix}x-growth-empty-state.action.href`, undefined, { defaultMessage: '' }).length > 0"
                 :action="(['docs', 'create'] as const).find((item) => item === t(`${prefix}x-growth-empty-state.action.type`, undefined, { defaultMessage: '' }))"
                 :href="href"
                 :appearance="t(`${prefix}x-growth-empty-state.action.type`, undefined, { defaultMessage: '' }) === 'docs' ? 'secondary': undefined"
@@ -179,7 +179,7 @@ const can = useCan()
 
 const iconMapping: Record<string, unknown> = {
   'zone-cps': LocationIcon,
-  'zone-crud': LocationIcon,
+  'zones-crud': LocationIcon,
   'meshes': MeshIcon,
 }
 </script>
--- a/packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
+++ b/packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
@@ -19,6 +19,7 @@
         :key="title"
       >
         <EntityEmptyState
+          v-if="can('view growth-new-empty-states')"
           :title="title"
           appearance="secondary"
           data-testid="empty-block"
@@ -47,7 +48,7 @@
             <header>
               <h2 class="x-empty-state-title">
                 <XI18n
-                  :path="`${prefix}x-empty-state.title`"
+                  :path="`${prefix}x-growth-empty-state.title`"
                   default-path="components.x-empty-state.title"
                 />
               </h2>
@@ -64,7 +65,7 @@
               v-else-if="body.length > 0"
             >
               <XI18n
-                :path="`${prefix}x-empty-state.body`"
+                :path="`${prefix}x-growth-empty-state.body`"
                 default-path="components.x-empty-state.body"
               />
             </template>
@@ -87,11 +88,74 @@
                   name="docs"
                   :size="KUI_ICON_SIZE_40"
                 />
-                {{ actionLabel }}
+                {{ t(`${prefix}x-growth-empty-state.action.label`, undefined, { defaultMessage: '' }) }}
               </XAction>
             </slot>
           </template>
         </EntityEmptyState>
+        <KEmptyState
+          v-else
+          class="x-empty-state"
+          data-testid="empty-block"
+        >
+          <template
+            #icon
+          >
+            <slot name="icon" />
+          </template>
+          <template
+            #title
+          >
+            <slot
+              name="title"
+            >
+              <template
+                v-if="title.length > 0"
+              >
+                <header>
+                  <h2>
+                    <XI18n
+                      :path="`${prefix}x-empty-state.title`"
+                      default-path="components.x-empty-state.title"
+                    />
+                  </h2>
+                </header>
+              </template>
+            </slot>
+          </template>
+
+          <template
+            v-if="slots.default"
+          >
+            <slot name="default" />
+          </template>
+          <template
+            v-else-if="body.length > 0"
+          >
+            <XI18n
+              :path="`${prefix}x-empty-state.body`"
+              default-path="components.x-empty-state.body"
+            />
+          </template>
+
+          <template
+            #action
+          >
+            <slot name="action">
+              <XAction
+                v-if="href.length > 0"
+                :action="(['docs', 'create'] as const).find((item) => item === actionType)"
+                :href="href"
+              >
+                {{ actionLabel }}
+              </XAction>
+              <XTeleportSlot
+                v-else
+                :name="`${props.type}-x-empty-state-actions`"
+              />
+            </slot>
+          </template>
+        </KEmptyState>
       </template>
     </template>
   </XI18n>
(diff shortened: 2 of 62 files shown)


packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
```diff
@@ -79,18 +76,44 @@
...
     </template>
   </XI18n>
 </template>
 
 <script lang="ts" setup>
-import { KEmptyState } from '@kong/kongponents'
+import { KUI_COLOR_TEXT_DECORATIVE_AQUA, KUI_ICON_SIZE_50 } from '@kong/design-tokens'
+import { LocationIcon, AnalyticsIcon, MeshIcon } from '@kong/icons'
+import { EntityEmptyState } from '@kong-ui-public/entities-shared'
+import '@kong-ui-public/entities-shared/dist/style.css'
```
johncowen, line 89: Could we put this line into here with the others:

https://github.com/kumahq/kuma-gui/blob/e9b0a81d59031989f783a572a1e51a816776d1ad/packages/kuma-gui/src/assets/styles/main.scss#L8-L9

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
```diff
@@ -79,18 +76,44 @@
...
+
 const props = withDefaults(defineProps<{
   type?: string
 }>(), {
   type: '',
 })
 const slots = defineSlots()
+
+const iconMapping: Record<string, unknown> = {
+  'zone-cps': LocationIcon,
```
johncowen, line 99: We recently made a PR (I think it was this one https://github.com/kumahq/kuma-gui/pull/3520) to make our Zone icons all use the same type of icon (i.e. the world looking one). We should either make all zone icons use the LocationIcon as you have done here, or make them all use the World icon. Let me get back to you on this one.

johncowen (commented):
Looks great! I left a few little nits and a question or two, and something for me to follow up also.

Apart from the code nits, I noticed one thing. Should these things be visible when it's empty?:

Screenshot 2025-02-20 at 13 09 25

packages/kuma-gui/package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "@gera2ld/tarjs": "^0.3.1",
     "@kong-ui-public/app-layout": "^4.3.10",
+    "@kong-ui-public/entities-shared": "^3.17.1",
```
johncowen, line 16: It looks like you've installed this but not committed the `package-lock.json` changes?

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
```diff
@@ -79,18 +76,44 @@
...
+  'meshes':MeshIcon,
```
schogges, line 100: nit, format: missing a space between key and value 🙂

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
```diff
@@ -79,18 +76,44 @@
...
+
+const handleLearnMore = (url: string): void => {
+  // open documentation in a new tab
+  window.open(url, '_blank')
+}
+
+const handleCreateAction = (url: string): void => {
+  // open create page
+  window.open(url, '_blank')
+}
```
schogges, line 111: Is it possible to inline them? Two functions that are doing essentially the same thing and do noting more than `window.open`.

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
```diff
@@ -18,52 +18,49 @@
...
       >
-        <KEmptyState
-          class="x-empty-state"
+        <EntityEmptyState
+          :title="title"
+          appearance="secondary"
           data-testid="empty-block"
+          :action-button-text="actionType === 'create' && href.length ? actionLabel : undefined"
+          :learn-more="actionType === 'docs' && href.length"
+          @click:learn-more="handleLearnMore(href)"
```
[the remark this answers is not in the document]
afrzhussain: its not possible directly, we might have to do it via the slots

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
```diff
@@ -18,52 +18,49 @@
         ]"
         :key="title"
       >
-        <KEmptyState
-          class="x-empty-state"
+        <EntityEmptyState
+          :title="title"
```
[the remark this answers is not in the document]
afrzhussain: The EntityEmptyState follows the design where the title is 18px (the current one is 16px). We have updated this design across the product, so keeping it at 16px will look odd and inconsistent.

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
afrzhussain: Screenshot 2025-02-21 at 1 25 55 AM
We already are using the World Icon in networks in konnect

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
afrzhussain: Moved it to a single function

schogges: It's just a nit and non-blocking, but I still feel there is no need for the abstraction and we could just inline `window.open` in the template 🤔

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
johncowen: Ok lets leave this as you have it for now, we can make a decision separately and circle back if necessary 👍

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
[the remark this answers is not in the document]
johncowen: The important thing here is the semantics. I'm guessing you are talking about `font-size`s being different?

Would it be possible to use a `header h2` and add a scoped style within XEmptyState to make it `18px`?

Otherwise we'll have h1's inside of h2 sections, which I believe would fail any sort of a11y audit.

Lmk on this one

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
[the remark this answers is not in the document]
johncowen: Slots would be ideal if that means we can use anchors and keep the a11y goodness there. Let me know if if tricky or not possible tho and we can skip this.

johncowen:
This is looking really good @afrzhussain ! I resolved a few convos, thanks for tweaking those bits!

I had a couple more questions off the back of your responses, lmk on those. I wouldn't say they were super-blockers but would be good if they are possible here.

Other than that I think there was one thing left from @schogges (the inlining comment) and then we need to sort out the DCO check and we should be good to go 👍

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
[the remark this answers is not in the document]
schogges: I agree with @johncowen, I think semantics are generally more important than a few pixels. On top I am questioning what is worse: inconsistency across different parts of the product or being inconsistent in one part of the product 🤔 Generally I think changing it all at once would be best, i.e. through a shared component and design-tokens, but this is rather a long-term solution for these kind of things.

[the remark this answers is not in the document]
afrzhussain: updated

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
afrzhussain: updated the implementation to use anchors instead, removed this bit

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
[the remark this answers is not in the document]
afrzhussain: updated

packages/kuma-gui/src/app/zones/locales/en-us/index.yaml
```diff
@@ -1,7 +1,7 @@
 zone-cps:
   docs: &zone-cps.docs
     type: docs
-    label: Documentation
+    label: Learn more
```
johncowen, line 4: I think this change needs to be applied in `meshes` also

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
```diff
@@ -18,59 +18,67 @@
...
-                    <XI18n
-                      :path="`${prefix}x-empty-state.title`"
-                      default-path="components.x-empty-state.title"
-                    />
-                  </h2>
-                </header>
-              </template>
-            </slot>
+          <template #title>
+            <header v-if="title.length">
```
johncowen, line 42: Does this `v-if` need to go on the template like before? 

Looks like if we could get an empty div rendered if we leave it like this if I understood correctly?

https://github.com/Kong/public-ui-components/blob/b4f6446e145b962bdb6505a9790a8c4a4304677a/packages/entities/entities-shared/src/components/entity-empty-state/EntityEmptyState.vue#L11-L14

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
```diff
@@ -18,59 +18,67 @@
...
+                v-if="slots.icon"
+                name="icon"
+              />
+              <component
+                :is="iconMapping[type]"
+                v-else-if="type && iconMapping[type]"
+                :color="KUI_COLOR_TEXT_DECORATIVE_AQUA"
+                :size="KUI_ICON_SIZE_50"
+              />
+              <AnalyticsIcon v-else />
```
johncowen, line 38: Does this need the same color/size applied?

I'm guessing this is why we see this black-on-blue-bg (but I could be wrong):

Screenshot 2025-02-21 at 15 17 53

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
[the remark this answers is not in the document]
johncowen: awesome, thanks for doing this 👍

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
johncowen: 👏  much better,  less code and we are avoiding re-implementing the platform  🎉 

---

One thing I think we will need to do here, and that is move the XTeleportSlot from underneath (to the right of) the docs link and move it above (to the left of) the docs link instead. I think whatever ends up in that slot will need to be to the left of the docs link?

I also think (different to the previous implementation), we no longer need the `v-else` on the `XTeleportSlot`. I think we need whatever goes in the Teleport to be "additional" instead of a replacement.

You can test out how this will look if someone uses the slot by temporarily adding something like underneath the XTemplateSlot (obv just for testing locally):

```vue
<XTeleportTemplate
  :to="{ name: `${props.type}-x-empty-state-actions` }"
>
  <XAction
    appearance="secondary"
  >
    Something
  </XAction>
</XTeleportTemplate>
```

Does that make sense?

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
afrzhussain: updated

packages/kuma-gui/src/app/zones/locales/en-us/index.yaml
afrzhussain: updated

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
afrzhussain: updated

packages/kuma-gui/src/app/x/components/x-empty-state/XEmptyState.vue
afrzhussain: updated

Issue #3489, opened 2025-01-29
Add more Info to Structured Format of Policy Summary View


schogges:
### Description

After we have split up the summary view into formats, the structured view is too blank in some cases, i.e. the policy summary view only shows a few fields and we can add some more info here.

![Image](https://github.com/user-attachments/assets/2738db3a-4e9b-4fa6-9124-ff0b12b859d3)

lahabana:
Dropdowns should include "Universal YAML" and "Kubernetes YAML".

We should add the yaml of the `spec` (when there a "spec") inline in the Structured view.

lahabana:
2 steps:

1. Add spec to policies structured view (as code block)
2. Make the dropdown a 3 choice option (structured, kubernetes, universal)
3. Remember in localstorage which is picked

PR #3579, opened 2025-02-20
feat(policy-summary): add spec to structured view

schogges:
This is the first part and in a follow-up I'll add more view types for the config. Here I added the spec in a code block to the structured view, so the user can still get all information in this view and doesn't need to switch to YAML.

---

I added a prop to hide the `Copy as K8s` button from the code block. I think the spec is agnostic to universal and k8s.

---

image

Part of #3489

### DIFF
--- a/packages/kuma-gui/src/app/policies/components/PolicySummary.vue
+++ b/packages/kuma-gui/src/app/policies/components/PolicySummary.vue
@@ -67,6 +67,18 @@
           </template>
         </DefinitionCard>
       </div>
+
+      <h3>
+        <XI18n
+          path="policies.routes.item.spec"
+        />
+      </h3>
+
+      <ResourceCodeBlock
+        v-if="policy.spec"
+        :resource="policy.spec"
+        :show-k8s-copy-button="false"
+      />
     </template>
     <template v-else>
       <div>
@@ -83,6 +95,7 @@
 import type { Policy } from '../data'
 import { useI18n, useCan } from '@/app/application'
 import DefinitionCard from '@/app/common/DefinitionCard.vue'
+import ResourceCodeBlock from '@/app/x/components/x-code-block/ResourceCodeBlock.vue'
 
 const { t } = useI18n()
 const can = useCan()
--- a/packages/kuma-gui/src/app/policies/components/PolicySummary.vue
+++ b/packages/kuma-gui/src/app/policies/components/PolicySummary.vue
@@ -68,15 +68,9 @@
         </DefinitionCard>
       </div>
 
-      <h3>
-        <XI18n
-          path="policies.routes.item.spec"
-        />
-      </h3>
-
       <ResourceCodeBlock
         v-if="policy.spec"
-        :resource="policy.spec"
+        :resource="{ spec: policy.spec }"
         :show-k8s-copy-button="false"
       />
     </template>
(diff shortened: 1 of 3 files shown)


johncowen (commented):
Still looking at this but just spotted something.

Totally not related to this PR, but this looks like it could be a pre-existing bug to me? wdyt?

Screenshot 2025-02-21 at 11 20 12

I'm not totally sure what logic it uses to make the targetRef say `[Mesh]` instead of `[MeshService]`?

packages/kuma-gui/src/app/policies/components/PolicySummary.vue
```diff
@@ -67,6 +67,18 @@
...
+
+      <h3>
+        <XI18n
+          path="policies.routes.item.spec"
+        />
+      </h3>
+
+      <ResourceCodeBlock
+        v-if="policy.spec"
+        :resource="policy.spec"
```
johncowen, line 79: I wondered if it would be good to add the `spec:` property in the output, do you follow what I mean, so the output would only show the single `spec:` property, then its super clear that its the `spec` property we are looking at. wdyt?

packages/kuma-gui/src/app/policies/components/PolicySummary.vue
```diff
@@ -67,6 +67,18 @@
...
+        :show-k8s-copy-button="false"
```
johncowen, line 80: Guessing this might be part of a "support the existing stuff while we change some partially related stuff"?

johncowen (commented):
Looks good to me! I had one tiny suggestion, and a non-important curiosity question. Otherwise I reckon this is good to go! lmk on those

johncowen:
> Totally not related to this PR, but this looks like it could be a pre-existing bug to me? wdyt?

Oh wait, I think this depends on a topLevel targetRef, so I think the screengrab I posted here is correct?

schogges:
> > Totally not related to this PR, but this looks like it could be a pre-existing bug to me? wdyt?
> 
> Oh wait, I think this depends on a topLevel targetRef, so I think the screengrab I posted here is correct?

We fall back to `Mesh` in case `spec.targetRef` is not set 🤔 
https://github.com/kumahq/kuma-gui/blob/master/packages/kuma-gui/src/app/policies/components/PolicySummary.vue#L20-L30

packages/kuma-gui/src/app/policies/components/PolicySummary.vue
schogges: Yeah, I agree 🙂 I've added a title `Spec`, but having it inline in the code block is much better 👍

packages/kuma-gui/src/app/policies/components/PolicySummary.vue
schogges: Not sure if I follow 😅 I think in this case we don't need to "Copy as k8s" button, just the normal copy is enough. But there was no way to keep the one but hide the other.

> I added a prop to hide the Copy as K8s button from the code block. I think the spec is agnostic to universal and k8s.

I added this to the PR description and just moved it up above the screen grab 🙂

packages/kuma-gui/src/app/policies/components/PolicySummary.vue
schogges: Done, lmk 🙂

johncowen: Yeah I like this much better 🙌

johncowen (approved):

PR #3580, opened 2025-02-21
chore: ensure x-action anchors get underlines applied within XAlerts

johncowen:
Tiny follow up to https://github.com/kumahq/kuma-gui/pull/3566

### DIFF
--- a/packages/kuma-gui/src/app/x/components/x-action/XAction.vue
+++ b/packages/kuma-gui/src/app/x/components/x-action/XAction.vue
@@ -51,6 +51,9 @@
     </KButton>
     <RouterLink
       v-else
+      :class="{
+        'x-action-appearance-anchor': true,
+      }"
       data-testid="x-action"
       v-bind="$attrs"
       :to="{
@@ -84,6 +87,7 @@
       v-bind="$attrs"
       :href="props.href"
       :class="{
+        'x-action-appearance-anchor': true,
         'action-docs': props.action === 'docs',
       }"
       target="_blank"
@@ -148,7 +152,7 @@
 
     <button
       v-else
-      :class="`appearance-${props.appearance}`"
+      :class="`x-action-appearance-${props.appearance}`"
       data-testid="x-action"
       v-bind="$attrs"
       @click="emit('click')"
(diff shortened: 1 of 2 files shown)


packages/kuma-gui/src/app/x/components/x-action/XAction.vue
```diff
@@ -51,6 +51,9 @@
     </KButton>
     <RouterLink
       v-else
+      :class="{
+        'x-action-appearance-anchor': true,
+      }"
```
schogges, line 56: nit: I think we can just use `class="..."` here?

packages/kuma-gui/src/app/x/components/x-action/XAction.vue
```diff
@@ -84,6 +87,7 @@
       v-bind="$attrs"
       :href="props.href"
       :class="{
+        'x-action-appearance-anchor': true,
```
schogges, line 90: nit: similar to previous comment, I think we can just use `class="..."` for this?

johncowen: This one specifically I'd keep like this due to the fact we are conditionally adding the next line under here (line 91).

I don't mind what we do with the others, but I'd maybe fall on the side of keeping it consistent in this file. The only reason I didn't keep the one on line 155 using a `{}` syntax is because the linter didn't like it because I wanted to use a template literal in the key of the object so I changed that one to be inconsistent. But! Looking deeper now I realized I was missing square brackets 🤦 :

So on line 151 below:

```javascript
{[`x-action-appearance-${props.appearance}`]: true}
```

If I make this change ^ below, that would make them all consistently use the object approach. But I'm still happy to keep all but one using string instead of object. Lmk!

schogges: I have no strong opinion, it just looked a little odd to me. From my pov there is nothing wrong with having both `class` and `:class` on one element. I'll leave it up to you 🙂

johncowen: I'm gonna circle round on this if thats ok, I get the feeling I'm going to be in this file again soon-ish

Issue #3494, opened 2025-01-31
Refactor persistence strategy of user preferences in RouteView component


schogges:
### Description

In order to store and retrieve the user preferences (`size` and `format`) and to reflect them in the URL as query search params, we had to implement some workarounds in #3457. Especially adding another source of truth about the stored preferences by introducing a local state object into the `RouteView` needs to be revisited. Ideally we only take into account what is stored on the global me route (localStorage).

~We also have to consider the following UX/User Stories:~

- ~When I open a deeplink that includes query parameters that are also stored as user preferences it should **not** override my user preferences~
- ~When I open a route that renders a component which tracks the state in the URL and my user preference doesn't match with the state options, it should fall back to some other default and track that in the URL but should **not** override my user preferences~
- ~As a rule of thumb: I expect changes of my user preferences **only** in cases where I intentionally set something as a user that is stored as a user preference~

---

Note: We decided this isn't "User Preferences" so we crossed out the above user stories

johncowen:
Wanted to document here what we spoke about in our UI Sync:

Personally, I'd always imagined these _not_ to be "user preferences". IMO they are "sticky query params" that persist between tabs and sessions, and were built as "sticky query params" not as "user preferences". 

---

If we are doing "user preferences" then we probably need quite a bit of a different design from an application perspective because we can't overload the `route.params` with the query param backed by a default "sticky" localStorage value. If we are doing "user preferences" I would guess route params and preferences need to be separate things.

lahabana:
Let's figure what's complicated in this and explain what we're getting from doing it

johncowen:
Trying to write something here on the back of our conversation on this:

Reaching back into our memory and referencing the TODO comment, we think the issue is the `local` variable. Where we save some state locally that we extract from localStorage. Ideally we wouldn't need to use this `local` variable.

If we comment this line:

https://github.com/kumahq/kuma-gui/blob/e50c98d023f42381f9e4bf26779d681b1a1955e4/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue#L203

We can see what happens if we don't have this local variable by:

1. Going here http://localhost:8081/gui/meshes/default/data-planes?page=1&size=50&dataplaneType=all
2. Clicking the first dataplane in the list on the left
3. Changing the Configuration dropdown from `Structured` to `Kubernetes`
4. Clicking the second dataplane in the list on the left.
5. Seeing that the Configuration dropdown is `Structured` not `Kubernetes`

[...]

Issue #1967, opened 2024-01-08
"Copy as k8s" doesn't work in Safari


johncowen:
### What happened?

I noticed during working on https://github.com/kumahq/kuma-gui/pull/1873 that using the "Copy as k8s" copy functionality doesn't work in Safari.

---

Separately, I looked a little deeper into it an it looks to be due to the browser security features that don't let you perform copy actions without a user interaction. 

In this instance, when the user clicks the button, we perform a HTTP request before using the response of that to write to the users clipboard, hence the write is not really a result of a user interaction, but more a successful HTTP response.

When I briefly read into it, it seems like the "user interaction" is bounded by a time period. Following that I could reliably get the copying to work by using a short `setTimeout`, but as soon as I took the timeout past ~999 the copying reliably stopped working, producing a browser error (which we currently swallow). Other browsers seem to be more lenient on this. Although the assumptions I've written here could be entirely wrong.

lahabana:
Closing in favor of https://github.com/kumahq/kuma-gui/issues/3489

PR #3581, opened 2025-02-21
feat(summaries): add formats universal and kubernetes

### LINKED ISSUE #3489: Add more Info to Structured Format of Policy Summary View

schogges:
Requires

- https://github.com/kumahq/kuma-gui/pull/3579

Splits the format `YAML` into `Kubernetes` and `Universal`. The selected format is stored as preference `k8s` or `universal` in local storage.

Kubernetes

image

Universal

image

Closes #3489


johncowen:
I think we are going to have to decide what to do about our "Global Format" values. Previous to this PR we had 2 possible values `structured` and `yaml`. Now, at least in some places we have 3 `structured`, `yaml-universal` and  `yaml-k8s`.

Of the back of that I have a question, with some potential follow up questions/suggestions.

- I think the plan will be to have 3 options everywhere? Is that right do you think?

---

Separately, I think we will have to build the data interaction so if you don't have Kubernetes selected when landing on the page we don't make the `?format=kubernetes` request, we only actually make it when you switch to kubernetes. If you do already have kubernetes selected when landing on a page, then of course its fine to make the `?format=kubernetes` request when we land on the page.

johncowen:
> I think we will have more options everywhere.

👍 cool

>  I think that would be beneficial for the UX, because as a kubernetes user I'm mainly/only interested in the k8s format. 

Yeah this is where things can get weird. global could be on k8s, but a zone on that install could be universal or vice-versa. Or maybe I just want to see what universal looks like even though I'm totally on k8s.

Given the above maybe we should change the values to `structured`, `universal`, `k8s`? And we just implicitly know (because the default format for universal and k8s is YAML) that `universal` and `k8s` formats mean we are showing it in YAML for the specific software environment?

schogges:
> > I think that would be beneficial for the UX, because as a kubernetes user I'm mainly/only interested in the k8s format.
> 
> Yeah this is where things can get weird. global could be on k8s, but a zone on that install could be universal or vice-versa. Or maybe I just want to see what universal looks like even though I'm totally on k8s.

Yeah, totally agree 👍 

> Given the above maybe we should change the values to `structured`, `universal`, `k8s`? And we just implicitly know (because the default format for universal and k8s is YAML) that `universal` and `k8s` formats mean we are showing it in YAML for the specific software environment?

Sounds good 👍 I'll change it, but would also like to ask wdyt we should do once we add JSON to the group? 🙂

schogges:
The good thing is that in all the other places we currently have the flow of either show `structured` or the codeblock. So regardless if the user selects `kubernetes` or `universal`, in the other places the user will see the codeblock.

johncowen:
> Sounds good 👍 I'll change it, but would also like to ask wdyt we should do once we add JSON to the group? 🙂

Hmm, depends, what will the content of the JSON be? I would guess we aren't doing "JSON Universal" and "JSON Kubernetes", but maybe we are?

schogges:
> > Sounds good 👍 I'll change it, but would also like to ask wdyt we should do once we add JSON to the group? 🙂
> 
> Hmm, depends, what will the content of the JSON be? I would guess we aren't doing "JSON Universal" and "JSON Kubernetes", but maybe we are?

Not sure either, I guess we'll find out then 😄

schogges:
I just realized another thing: we currently only show the select in case there is a `spec`. I think this should be changed now to only include the `structured` in case there is a `spec`, but in any case show the select with options `Kubernetes` and `Universal`.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
```diff
@@ -53,16 +53,17 @@
...
-                  <div v-if="data.spec">
+                  <div>
                     <XSelect
                       :label="t('policies.routes.item.format')"
-                      :selected="route.params.format"
+                      :selected="data.spec || ['universal', 'k8s'].includes(route.params.format) ? route.params.format : 'universal'"
                       @change="(value) => {
                         route.update({ format: value })
                       }"
+                      @vue:before-mount="!data.spec && $event?.props?.selected && ['universal', 'k8s'].includes($event.props.selected) && $event.props.selected !== route.params.format && route.update({ format: $event.props.selected })"
```
schogges, line 63: Note: I'm not sure if this is the best way how to do it, but the problem is that if there is no `spec`, we don't show `structured` and don't provide it as an option to select. The `@vue:before-mount` makes sure that URL syncs to what is `:selected` so there is no case where we have `?format=structured` and the selected option is `Universal` or `Kubernetes`. However, this also overrides the user preference.

schogges: I couldn't find a quick and easy way to prevent updating the user preferences in these cases. I think this requires more work on the `RouteView`. I've added these constraints to https://github.com/kumahq/kuma-gui/issues/3494

schogges:
Hmm potential [new flake](https://github.com/kumahq/kuma-gui/actions/runs/13568149098/job/37926023208?pr=3581)?

image

I'll retrigger

johncowen:
Ah yeah sorry, I forgot to mention, since https://github.com/kumahq/kuma-gui/pull/3599 things related to policies are a lot more random now, so there might be somewhere were we need to specify our mock/setup there a little more.

schogges:
@johncowen would appreciate another review? 🙂

johncowen:
Ah cool, sorry wasn't sure if you'd finished here, 2 tics

schogges:
> Ah cool, sorry wasn't sure if you'd finished here, 2 tics

Sorry, I should've done that switch back to draft and back 😄

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: Q: In what cases do we not get a `spec`?

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
```diff
@@ -73,29 +77,87 @@
...
+                  :query="route.params.codeSearch"
+                  :is-filter-mode="route.params.codeFilter"
+                  :is-reg-exp-mode="route.params.codeRegExp"
+                  :show-k8s-copy-button="false"
+                  @query-change="route.update({ codeSearch: $event })"
+                  @filter-mode-change="route.update({ codeFilter: $event })"
+                  @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+                />
+              </DataLoader>
+            </template>
```
johncowen, line 135: Am I right in thinking we should be in the process of getting rid of `ResourceCodeBlock` and replacing with just `XCodeBlock`?

I'd understood that in order to copy the kubernetes code now you would use the format dropdown to switch to kubernetes format, and then just click the normal copy button to copy the kubernetes content?

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: I'm not totally sure, but I think this is the case for legacy policies on Universal. When I compare `Kubernetes` and `Universal` with https://kuma.io/docs/2.9.x/policies/circuit-breaker/#circuit-breaker and https://kuma.io/docs/2.9.x/policies/meshcircuitbreaker/ I can see that on the old policy (`CircuitBreaker`) on `Universal` there is no `spec`. I can see the `spec` on both `Universal` and `Kubernetes` for `MeshCircuitBreaker`.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: > Am I right in thinking we should be in the process of getting rid of `ResourceCodeBlock` and replacing with just `XCodeBlock`?

Oh, wasn't aware of that. Would it be fine to do that in a follow-up?

> I'd understood that in order to copy the kubernetes code now you would use the format dropdown to switch to kubernetes format, and then just click the normal copy button to copy the kubernetes content?

The Universal still offers both copy buttons. The Kubernetes codeblock just offers one copy, which copies the kubernetes format.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: So this edge case is only for legacy policies when using universal? Is this something we are continuing to support?

If we are looking to continue to support legacy policies with this, I looked at the CircuitBreaker example:

```yaml
type: CircuitBreaker
mesh: default
name: circuit-breaker-example
sources:
- match:
    kuma.io/service: web
destinations:
- match:
    kuma.io/service: backend
conf:
  interval: 1s
  baseEjectionTime: 30s
  maxEjectionPercent: 20
  splitExternalAndLocalErrors: false
  thresholds:
    maxConnections: 2
    maxPendingRequests: 2
    maxRequests: 2
    maxRetries: 2
  detectors:
    totalErrors: 
      consecutive: 20
    gatewayErrors: 
      consecutive: 10
    localErrors: 
      consecutive: 7
    standardDeviation:
      requestVolume: 10
      minimumHosts: 5
      factor: 1.9
    failure:
      requestVolume: 10
      minimumHosts: 5
      threshold: 85

```

If there is no `spec` should we be doing exactly the same as when there is a `spec` but instead use `sources/destinations/conf`? These three things seem to be equivalent information to `spec`?

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: > The Universal still offers both copy buttons

How come we are offering both? The kubernetes one here doesn't work totally across browsers (https://github.com/kumahq/kuma-gui/issues/1967) and we are offering a different way to access k8s format with the dropdown and a more traditional "view then copy" UX that works in all browsers.

> Would it be fine to do that in a follow-up?

Well maybe, but looking at the code there is a lot of stuff that is getting added here that would be unnecessary (all the copy/DataLoader stuff).

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: > So this edge case is only for legacy policies when using universal? Is this something we are continuing to support?

I think we have to support that until we drop supporting legacy policies.

> If there is no `spec` should we be doing exactly the same as when there is a `spec` but instead use `sources/destinations/conf`? These three things seem to be equivalent information to `spec`?

We could potentially do that, but I'm not sure about this 🤔 I feel this is not in scope of this issue/PR as I'm just splitting `YAML` into `Universal` and `Kubernetes`. We should also improve on our mocks to support that differentiation of legacy policy on `Universal` and `Kubernetes`. Let's create an issue and iterate on that, wdyt?

johncowen: > We could potentially do that, but I'm not sure about this 🤔 I feel this is not in scope of this issue/PR

Well wouldn't it make this "but sometimes there is no structured view" thing go away? No only that but it offers the exact same UX for legacy policies as new policies. Then within a certain amount of time there will be nothing without a `spec` anyway?

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: Fair points 👍  will change it.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: > > We could potentially do that, but I'm not sure about this 🤔 I feel this is not in scope of this issue/PR
> 
> Well wouldn't it make this "but sometimes there is no structured view" thing go away? Not only that but it offers the exact same UX for legacy policies as for new policies. Then within a certain amount of time there will be nothing without a `spec` anyway?

Yes, but I would still need the logic to sync URL and select because in subscriptions summary we have no `Kubernetes` 🤔

johncowen: OK well I suppose thats a different thing as we are no longer talking about "with or without `spec`".

Is there anywhere else where you think we wouldn't have a kubernetes output similar to subscriptions?

schogges: > OK well I suppose thats a different thing as we are no longer talking about "Policies with or without `spec`".
> 
> Is there anywhere else where you think we wouldn't have a kubernetes output similar what you've said with subscriptions?

I think `SubscriptionSummaryView` is the only special 🤔 

>If there is no spec should we be doing exactly the same as when there is a spec but instead use sources/destinations/conf? These three things seem to be equivalent information to spec?

I went through some more examples of legacy policies in our docs and depending on the policy there is more than just `sources/destinations/conf`, i.e. [TrafficTrace](https://kuma.io/docs/2.9.x/policies/traffic-trace/) has `tracing` and [TrafficMetrics](https://kuma.io/docs/2.9.x/policies/traffic-metrics/) has `metrics`. Also these are just examples and potentially not fully complete.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: Done 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: > I went through some more examples of legacy policies in our docs and depending on the policy there is more than just sources/destinations/conf, i.e. [TrafficTrace](https://kuma.io/docs/2.9.x/policies/traffic-trace/) has tracing and [TrafficMetrics](https://kuma.io/docs/2.9.x/policies/traffic-metrics/) has metrics. Also these are just examples and potentially not fully complete.

So it feels like is "everything but name and type"?

> I think SubscriptionSummaryView is the only special 🤔

Ok perfect 👍 , I'd say subscriptions aren't config we just show the output, the information we show there is pretty much metrics.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: 👍 are there any more usages of ResourceCodeBlock still left? Can we delete that yet?

johncowen: We just looked and found we still use these in non-summary views, so can't remove it quite yet as we are sticking to just amend summaries in this PR.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: Ok so we just synced on this on this one, to solve subscriptions we are gonna uses ` ?output` instead of `?format` for subscriptions seeing as the output there is not config, its more like metrics. 

For the weird "legacy policies sometimes don't have a spec", we are gonna stick with what is already in this PR, and then potentially we can circle back later to alter slightly if we want to.

Gonna resolve this one as I think we have this one done. But lmk if there any q's off the back of that.

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
johncowen: Just to note what we said in our sync:

I completely didn't realize what we'd need to do to remove the ResourceBlock entirely. So we are gonna keep it for the moment and use the new `:show-k8s-copy=false` flag so we avoid the Safari bug.

Once we've done some more data work probably with a new `resources` modules at a later date, we can circle back and finally remove the ResourceCodeBlock.

Sorry for the back and forth on this one 😞 

Gonna resolve this one

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
```diff
@@ -53,16 +53,20 @@
...
+                    :key="typeof options"
+                  >
                     <XSelect
                       :label="t('policies.routes.item.format')"
-                      :selected="route.params.format"
+                      :selected="options.includes(route.params.format) ? route.params.format : options[0]"
                       @change="(value) => {
                         route.update({ format: value })
                       }"
+                      @vue:before-mount="$event?.props?.selected && options.includes($event.props.selected) && $event.props.selected !== route.params.format && route.update({ format: $event.props.selected })"
```
johncowen, line 66: No action here at all, just wanted to note: whilst its super cool, I'm a little 🤔 about using this seeing as I can't find any documentation for it.

No biggie here, but wanted to note that, if you know of a link to documentation lmk!

packages/kuma-gui/src/app/gateways/views/BuiltinGatewaySummaryView.vue
```diff
@@ -62,16 +62,20 @@
                   <h3>
                     {{ t('gateways.routes.item.config') }}
                   </h3>
-                  <div>
+                  <div
+                    v-for="options in [['structured', 'universal', 'k8s']]"
+                    :key="typeof options"
+                  >
```
johncowen, line 68: Just a question: no action. Is the only place we use `options` here on line 78? I couldn't see it anywhere else, but maybe I missed something?

johncowen: P.S. oh! I guess it makes it consistent with the files where we do need additional access to `options` do do an includes. I think that could be the answer here?

packages/kuma-gui/src/app/subscriptions/views/SubscriptionSummaryView.vue
```diff
@@ -32,25 +32,31 @@
...
-                    v-for="value in ['structured', 'yaml']"
-                    :key="value"
-                    #[`${value}-option`]
+                  <XSelect
+                    :label="t('subscriptions.routes.item.format')"
+                    :selected="route.params.output"
+                    @change="(value) => {
+                      route.update({ output: value })
+                    }"
+                    @vue:before-mount="$event?.props?.selected && options.includes($event.props.selected) && $event.props.selected !== route.params.output && route.update({ output: $event.props.selected })"
```
johncowen, line 45: Do we still need this in places where we don't have "conditional, 2 or maybe 3 options"?

packages/kuma-gui/src/app/gateways/views/BuiltinGatewaySummaryView.vue
schogges: `options` is also used in that `@vue` directive in line 75 🙂

packages/kuma-gui/src/app/subscriptions/views/SubscriptionSummaryView.vue
schogges: Oh, no we don't need that here 👍 I'll remove it

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: It's weird because I couldn't find documentation about this either 🤔

johncowen: Yeah lets leave it as is for now, just we should be cautious of using it if its "undocumented" I suppose.

Will resolve this one

packages/kuma-gui/src/app/gateways/views/BuiltinGatewaySummaryView.vue
johncowen: oh 🤦 , thanks!

packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue
schogges: LLM told me:
>The @vue: prefix you're seeing is part of Vue's new directive system introduced in Vue 3.2. It's a special syntax that provides enhanced functionality compared to regular directives. Let me break down what this means and how it works.

And this is what I search returns on vuejs core: https://github.com/search?q=repo%3Avuejs%2Fcore+%40vue%3A&type=code

johncowen (approved):
👍

Issue #2153, opened 2024-02-07
Show warning notification when NACKs happen on global


kleinfreund:
### Description

(Someone to add more details here)

schogges:
We need more information as it's not super clear what to do right now.

PR #3582, opened 2025-02-24
feat(zones): show warning when NACKs are received

johncowen:
When we receive a non-zero `total.responsesRejected` i.e. NACKs shows the following notification:

Screenshot 2025-02-24 at 11 31 05

Note: I've implemented this using the same approach as before. Once I've added some other notifications in I aim to go through and re-organize at least the i18n a little more consistently.

Closes https://github.com/kumahq/kuma-gui/issues/2153


schogges (approved):
👍

Issue #3521, opened 2025-02-08
Missing pagination for ZoneIngress/Egress


lukidzi:
### Description

When there are more than 100 ingresses, there is no way to navigate to the second page because pagination is missing on the frontend. It appears that the backend supports pagination requests, but the frontend does not provide a way to use them.

<img width="1405" alt="Image" src="https://github.com/user-attachments/assets/a056ad2b-35f2-4059-9c36-d43db81620f8" />

![Image](https://github.com/user-attachments/assets/abc20046-bc9f-4007-9bde-c0084985f23c)

beginning of the response:

```
{
 "total": 442,
 "items": [
  {
   "type": "ZoneIngressOverview",
   "name": "ip-10-36-32-132-zx96dzdw86z4d5xb",
```

johncowen:
Just to note here, we currently hardcode a max of 100 Ingresses or Egresses due to the fact that we can't page these because we show a nested view inside of a zone and the API doesn't (or didn't at least, I think this has been added) let us list ingresses by zone:

https://github.com/kumahq/kuma-gui/blob/385b55c3173a468817018e6fa319f698bdbc1e03/src/app/zone-ingresses/views/ZoneIngressListView.vue#L27

lukidzi:
I've tried 2 options but I cannot see URL accepting a specific zone. We should probably implement it on the backend
`/zone-ingresses/_overview?name=<name>`
`/zone-ingresses/name/<name>`

also kumactl doesn't support 

```
➜  bin ./kumactl get zone-ingresses --help
Show ZoneIngress entities.

Usage:
  kumactl get zone-ingresses [flags]

Flags:
  -h, --help            help for zone-ingresses
  -m, --mesh string     mesh to use (default "default")
      --offset string   the offset that indicates starting element of the resources list to retrieve
      --size int        maximum number of elements to return

Global Flags:
      --api-timeout duration   the timeout for api calls. It includes connection time, any redirects, and reading the response body. A timeout of zero means no timeout (default 1m0s)
      --config-file string     path to the configuration file to use
      --log-level string       log level: one of off|info|debug (default "off")
      --no-config              if set no config file and config directory will be created
  -o, --output string          output format: one of table|yaml|json (default "table")
```

johncowen:
Perfect thanks @lukidzi ! 🙌

lahabana:
[...]

PR #3584, opened 2025-02-24
feat(zones): adds filter based pagination to zone proxies

johncowen:
Following https://github.com/kumahq/kuma/pull/12840 this PR adds support for proper paging of Zone Proxies via the GUI. Previousy we would just request a large number of Zone Proxies and filter for the ones we were interested in, with no paging (max of 100 proxies)

The GUI now pages Zone Ingresses and Zone Egresses just like any other resource in the GUI

Closes https://github.com/kumahq/kuma-gui/issues/3521

### DIFF
--- a/packages/kuma-gui/src/app/zone-egresses/sources.ts
+++ b/packages/kuma-gui/src/app/zone-egresses/sources.ts
@@ -21,18 +21,12 @@ export const sources = (api: KumaApi) => {
   return defineSources({
     '/zone-cps/:name/egresses': async (params) => {
       const { name, size, page } = params
+      const filter = name !== '*' ? {
+        [`labels.${'kuma.io/zone'}`]: name,
+      } : undefined
       const offset = size * (page - 1)
 
-      const res = await api.getAllZoneEgressOverviews({ size, offset })
-      if (name !== '*') {
-        // temporary frontend filtering until we have support for filtering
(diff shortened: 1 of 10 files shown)


packages/kuma-gui/src/app/zone-egresses/sources.ts
```diff
@@ -21,18 +21,12 @@ export const sources = (api: KumaApi) => {
   return defineSources({
     '/zone-cps/:name/egresses': async (params) => {
       const { name, size, page } = params
+      const filter = name !== '*' ? {
+        [`labels.${'kuma.io/zone'}`]: name,
+      } : undefined
```
johncowen, line 26: When we are in standalone mode with only ZoneEgress `name` comes in as `*` (i.e. all zones I don't care which), in this case we don't send the `?filter` param.

packages/kuma-gui/src/types/api.d.ts
```diff
@@ -23,6 +23,7 @@ export interface PaginatedApiListResponse<ResourceType> extends ApiListResponse<
 export interface PaginationParameters {
   size?: number
   offset?: number
+  filter?: Record<string, string>
```
johncowen, line 26: Not really a "pagination" parameter, but seeing as this handwritten type will go soon, I didn't bother worrying about the naming weirdness too much.

johncowen (commented):
Notes:

PR #3598, opened 2025-02-25
feat : new empty state and content changes

afrzhussain:
### Summary
Updated XEmpty State
- Replaced KEmpty with EntityEmptyState
- Updated Empty state content


johncowen:
Replacement for:

xref: https://github.com/kumahq/kuma-gui/pull/3578

johncowen:
@afrzhussain could you add an empty commit on the end here? Looks like the `needs` jobs on our branch protection haven't run for some reason. Hopefully an empty commit should make things rerun and 🤞 the `needs` jobs will run next time.

package.json
```diff
@@ -7,6 +7,7 @@
     "esbuild": "^0.25.0"
   },
   "dependencies": {
+    "@kong-ui-public/entities-shared": "^3.17.2",
```
schogges, line 10: Why do we need to have `@kong-ui-public/entities-shared` both in the root and in `packages/kuma-gui`?

johncowen: Ah good spot, yeah agree this should only be in packages/kuma-gui/packages.json.

🤔 lets circle back and sort this later tho I think, wdyt?

schogges: I'd be fine sorting this out separately, but rather soon as this might cause issues downstream 🤔

schogges: I'm wondering why [this commit](https://github.com/kumahq/kuma-gui/pull/3598/commits/628b3af8f22d67a5dec1bfdf5f4c21041def44d9) caused this 🤔

johncowen: Note: We decided we are gonna sort this out here first/now

schogges: Great, looks good to me now 👍 🙂

package-lock.json
```diff
@@ -2426,6 +2426,30 @@
...
+    "node_modules/@kong-ui-public/core/node_modules/date-fns": {
+      "version": "4.1.0",
+      "resolved": "https://registry.npmjs.org/date-fns/-/date-fns-4.1.0.tgz",
+      "integrity": "sha512-Ukq0owbQXxa/U3EGtsdVBkR1w7KOQ5gIBqdH2hkvknzZPYvBxb/aa6E8L7tmjFtkwZBu3UXBbjIgPo/Ez4xaNg==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/kossnocorp"
+      }
+    },
```
johncowen, line 2452: Doesn't block approval, but more to note for me and @schogges 

---

"Uh-oh" 😄

johncowen (approved):
LGTM! 🎉 

Thanks for the effort here, there was quite a bit of back and forth so thanks for bearing with it!

Issue #2851, opened 2024-02-27
Alert in the GUI when there is no MTP for a given mesh


bcollard:
### Description

Add an alert in the GUI when there is no MTP configured for a given Mesh.

Stumbled upon this when I created an extra Mesh and forgot about a default allow-all MTP.

schogges:
This might address this issue:

![Image](https://github.com/user-attachments/assets/1a1461ce-dd77-4c05-aaf9-0f7720ad93e0)

lahabana:
It should be a "error" level notification when: mtls is enabled (has an enabledBackend) and there's no MeshTrafficPermission (probably from MeshInsight).

johncowen:
> It should be a "error" level notification when: mtls is enabled (has an enabledBackend) and there's no MeshTrafficPermission (probably from MeshInsight).

@lahabana if this is the other way around, i.e. I have a MeshTrafficPermission but I don't have mTLS enabled, I'm guessing the MeshTrafficPermission is not working as expected? Is that right? If so, its probably a higher level of notification, say a red/danger notification in this case? wdyt?

PR #3599, opened 2025-02-25
refactor(meshes): slightly refactors MTP notification

johncowen:
Refactors the MTP notification we have when you have mTLS enabled but no MeshTrafficPermission.

---

While I was doing this I noticed that the faker `policyNames` code wasn't quite the same interface as faker, and I think the randomness wasn't quite working because max was defaulting to a super big value (which I think meant a low chance of getting low numbers) so I made some slight changes there after I looked in fakers code to see exactly how they do this and then made a `maxmin` helper to replicate fakers types/functionality so we keep the same interface everywhere.

This meant:

1. Altering slightly the `{ min, max }` args and how they are defaulted.
2. Using `Number.MAX_SAFE_INTEGER` over `Number.MAX_VALUE` as a "give me all the things" placeholder
3. If you use `Number.MAX_SAFE_INTEGER` to say "potentially include all of them" we don't pass that along to arrayElements
4. I removed includeLegacy where it wasn't needed, mostly because where i was using it I had to write `(undefined, { includeLegacy: true})`

I _think_ this is the functionality we want for `policyNames`

---

Apart from this I also made the policy enabling/disabling rendering work based on `total: 0` as well as just existence, which again is the functionality I _think_ we want?

Closes https://github.com/kumahq/kuma-gui/issues/2851

### DIFF
--- a/packages/kuma-gui/src/test-support/mocks/src/mesh-insights/_.ts
+++ b/packages/kuma-gui/src/test-support/mocks/src/mesh-insights/_.ts
@@ -2,13 +2,12 @@ import type { EndpointDependencies, MockResponder } from '@/test-support'
 export default ({ env, fake }: EndpointDependencies): MockResponder => (req) => {
   const params = req.params
 
-  const policyTypes = fake.kuma.policyNames({ min: Number.MAX_VALUE }, { includeLegacy: true })
-
   const serviceTotal = parseInt(env('KUMA_SERVICE_COUNT', `${fake.number.int({ min: 1, max: 30 })}`))
 
   const standard = fake.kuma.healthStatus()
   const gatewayBuiltin = fake.kuma.healthStatus()
   const gatewayDelegated = fake.kuma.healthStatus()
+
   const gateway = {
     total: (gatewayBuiltin.total ?? 0) + (gatewayDelegated.total ?? 0),
     online: (gatewayBuiltin.online ?? 0) + (gatewayDelegated.online ?? 0),
@@ -41,11 +40,9 @@ export default ({ env, fake }: EndpointDependencies): MockResponder => (req) => 
         gatewayDelegated,
       },
       policies: {
-        ...policyTypes.reduce((prev: Record<string, { total: number }>, item) => {
-          if (fake.datatype.boolean()) {
-            prev[item] = {
-              total: fake.number.int(20),
-            }
+        ...fake.kuma.policyNames().reduce((prev: Record<string, { total: number }>, item) => {
+          prev[item] = {
+            total: fake.number.int(20),
           }
           return prev
         }, {}),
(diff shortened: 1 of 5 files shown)


packages/kuma-gui/src/test-support/mocks/src/mesh-insights/_.ts
```diff
@@ -41,11 +40,9 @@ export default ({ env, fake }: EndpointDependencies): MockResponder => (req) =>
...
       },
       policies: {
-        ...policyTypes.reduce((prev: Record<string, { total: number }>, item) => {
-          if (fake.datatype.boolean()) {
-            prev[item] = {
-              total: fake.number.int(20),
-            }
+        ...fake.kuma.policyNames().reduce((prev: Record<string, { total: number }>, item) => {
+          prev[item] = {
+            total: fake.number.int(20),
```
johncowen, line 45: This was the bit of code that led me to do all this. I wanted to remove the boolean here and just use `policyNames` but I couldn't get it to give me a random amount of names.

packages/kuma-gui/src/test-support/mocks/src/resources.ts
```diff
@@ -1,12 +1,12 @@
 import type { EndpointDependencies, MockResponder } from '@/test-support'
 export default ({ fake }: EndpointDependencies): MockResponder => (_req) => {
-  const policies = Array.from(fake.kuma.policyNames({ min: Number.MAX_VALUE }, { includeLegacy: true }))
-  const legacyPolicies = fake.kuma.policyNamesLegacy({ min: Number.MAX_VALUE })
-  
+  const policies = Array.from(fake.kuma.policyNames({ min: Number.MAX_SAFE_INTEGER }))
+  const legacyPolicies = fake.kuma.policyNamesLegacy({ min: Number.MAX_SAFE_INTEGER })
```
schogges, line 4: The reason why I added the optional `includeLegacy` (with default `true`) was for documentation purposes. Just to make clear that it includes the legacy ones. We might want to keep that? 🙂

schogges (approved):
LGTM 👍 
left a Q 🙂

packages/kuma-gui/src/test-support/mocks/src/resources.ts
schogges: I'm fine either way 🙂 and I agree that this kind of defeats the purpose of a default value, but I also liked the obvious documentation 😄 
Anyways let's just leave it out, it's good as is 👍

johncowen: Yeah, I think related to our other convo on https://github.com/kumahq/kuma-gui/issues/3602, I guess in this instance its at the call site so the fact that I am redefining a default for documentation purposes isn't quite the same as keeping function args for documentation purposes (even though could have TS docs there also), and I suppose this is the difference between the two.

Anyway resolving this one

Issue #1484, opened 2023-09-15
See if theres a CSS lint rule to add spaces around certain things


johncowen:
### Description

See https://github.com/kumahq/kuma-gui/pull/1483/files#r1327015822

```css
.class>*+* {}
.class >*+* {}
.class> * + * {}
```

vs

```css
.class > * + * {}
.class > * + * {}
.class > * + * {}
```

Bit of a nit but if we are keen on this sort of linting in other areas, we may as well add this one also if theres a rule readily available, shouldn't be long to add if there is?

johncowen:
Super nit of an issue, but since its here: its still relevant in the tiniest of ways

johncowen:
Looks like this one https://github.com/stylelint-stylistic/stylelint-stylistic/blob/main/docs/user-guide/rules.md#selector, assigning myself so we can finally do this and close it out

PR #3600, opened 2025-02-25
chore(lint): add CSS selector-combinator-space-after/before always lint

johncowen:
Adds a CSS lint rule to force spaces around selector combinator.

Just to note, this is a code style rule, but at least we all agree on it 😅 .

It also adds a dev dependency 🙈 , so if that dependency ends up in our weekly dependabot's all the time I am 100% ok with removing it again (along with the rule).

Closes https://github.com/kumahq/kuma-gui/issues/1484

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3475,6 +3475,51 @@
         "url": "https://github.com/sponsors/jonschlinkert"
       }
     },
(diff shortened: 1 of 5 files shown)


schogges (approved):
👍

PR #3601, opened 2025-02-25
chore: move some older env vars (strings) over to locales

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/2431

Moves or removes the last of our "strings" to i18n:

- `KUMA_KDS_URL` we no longer need here
- `KUMA_DOCS_URL` I've moved to i18n. The "re-shaping" from `2.9.0` to `2.9.x` is now controlled via i18n, so its an application level concern rather than a framework concern.

We have 2 non-index.html derived env vars left:

- `KUMA_MOCK_API_ENABLED` - actual env var
- `KUMA_VERSION_URL` - whilst this isn't index.html derived, and is a "string" we use it programatically i.e. we don't just "print the thing out to the screen", so I'm a little unsure about putting it in i18n. I'll likely move it/hardcode it directly into `main.ts` at some point or something similar. In any case its not a 100% i18n-able string.

### DIFF
--- a/packages/kuma-gui/.env
+++ b/packages/kuma-gui/.env
@@ -1,3 +1,2 @@
 VITE_VERSION_URL=https://kuma.io/latest_version
 VITE_KUMA_API_SERVER_URL=http://localhost:5681
-VITE_DOCS_BASE_URL=https://kuma.io/docs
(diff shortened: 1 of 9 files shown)


schogges (approved):
👍

PR #3604, opened 2025-02-26
feat: support inboundRules in _rules

schogges:
__TBD__


packages/kuma-gui/src/app/data-planes/views/DataPlaneInboundSummaryOverviewView.vue
```diff
@@ -97,128 +97,137 @@
...
-                :initially-open="0"
-                multiple-open
-                class="stack"
+            <DataLoader
+              :data="[policyTypesData, rulesData]"
+              :errors="[policyTypesError, rulesError]"
+            >
+              <template
+                v-for="policyTypes in [(policyTypesData?.policyTypes ?? []).reduce<Partial<Record<string, PolicyResourceType>>>((obj, policyType) => Object.assign(obj, { [policyType.name]: policyType }), {})]"
+                :key="typeof policyTypes"
```
schogges, line 117: Note: copied over from `DataplanePoliciesView`, because dictionary is easier to use further down in the template. Might be worth to move it to the data layer 🤔

packages/kuma-gui/src/app/services/data/MeshService.ts
```diff
@@ -14,7 +14,7 @@ export const MeshService = {
       namespace,
       labels,
       zone: labels['kuma.io/origin'] === 'zone' && labels['kuma.io/zone'] ? labels['kuma.io/zone'] : '',
-      spec: ((item = {}) => {
+      spec: ((item = { state: 'Unavailable' }) => {
```
schogges, line 17: Note: Had to add this here, because after updating the types from OAS the type became invalid here.

packages/kuma-gui/src/app/data-planes/views/DataPlaneInboundSummaryOverviewView.vue
johncowen: I didn't look super close at this, but does a `Object.groupBy` work here?

packages/kuma-http-api/index.d.ts
```diff
@@ -6360,9 +6610,10 @@ export interface components {
                 /**
                  * @description State of MeshService. Available if there is at least one healthy endpoint. Otherwise, Unavailable.
                  *     It's used for cross zone communication to check if we should send traffic to it, when MeshService is aggregated into MeshMultiZoneService.
+                 * @default Unavailable
                  * @enum {string}
                  */
-                state?: "Available" | "Unavailable";
+                state: "Available" | "Unavailable";
```
johncowen, line 6616: For linkability: `state` went from "possibly unset" to "always set"

packages/kuma-gui/src/app/services/data/MeshService.ts
johncowen: I think we can remove the `((item = {}))` entirely here and things would be valid?

i.e.

`spec.state` is always guaranteed to be set according to the OpenAPI spec. And if thats the case, `spec` itself is guaranteed to be an object.

https://github.com/kumahq/kuma-gui/pull/3604/files#r1977339625

packages/kuma-gui/src/test-support/mocks/src/resources.ts
```diff
@@ -1,12 +1,17 @@
 import type { EndpointDependencies, MockResponder } from '@/test-support'
-export default ({ fake }: EndpointDependencies): MockResponder => (_req) => {
+import type { paths } from '@kumahq/kuma-http-api'
+
+type ResourcesApi = paths['/_resources']['get']['responses']['200']['content']['application/json']
+
+export default ({ fake, env }: EndpointDependencies): MockResponder => (_req) => {
   const policies = Array.from(fake.kuma.policyNames({ min: Number.MAX_SAFE_INTEGER }))
   const legacyPolicies = fake.kuma.policyNamesLegacy({ min: Number.MAX_SAFE_INTEGER })
+  const resourceCount = parseInt(env('KUMA_RESOURCE_COUNT', `${fake.number.int({ min: Number.MAX_SAFE_INTEGER })}`))
```
johncowen, line 9: Do we want it to be at least `Number.MAX_SAFE_INTEGER` when `KUMA_RESOURCE_COUNT` is not set?

Is `Number.MAX_SAFE_INTEGER` exactly the same as `${fake.number.int({ min: Number.MAX_SAFE_INTEGER })}`?

Probably just:

```javascript
env('KUMA_RESOURCE_COUNT', `${Number.MAX_SAFE_INTEGER}`)
```

but maybe I missed something?

packages/kuma-gui/src/app/data-planes/views/DataPlaneInboundSummaryOverviewView.vue
johncowen: Oh look, like this further down I think?

Screenshot 2025-03-03 at 11 36 02

packages/kuma-gui/src/app/data-planes/views/DataPlanePoliciesView.vue
```diff
@@ -65,8 +65,9 @@
                 </template>
 
                 <!-- otherwise, for from rules, group by inbound port and display if we have any -->
+                <!-- filter rules that are being represented as inboundRules (isFromAsRules) -->
                 <DataCollection
-                  :items="rulesData!.rules"
+                  :items="rulesData!.rules.filter((rule) => !Boolean(policyTypes[rule.type]?.policy.isFromAsRules) )"
```
johncowen, line 70: Should this go in the `:predicate` one line below?

johncowen (commented):
I had an initial look at left a few nits, just gonna try it here locally 👍

packages/kuma-gui/src/app/rules/data/Rule.ts
johncowen, line 1: I was wondering why we didn't make a separate `InboundRule.fromObject` here (like we did for `ResourceRule`)

Then instead of calling `inboundRules.map(Rule.fromObject)` you'd call `inboundRules.map(InboundRule.fromObject)`, and lose the union argument in `Rule.fromObject`, making it a little simpler.

I guess, from trying to remember this area of the codebase and looking at this, you could then share `Origin` between `Rule`, `InboundRule` and `ResourceRule`.

For example, I have a question now over why `ResourceRule` takes `resourceMeta` and makes `id`, `namespace`, `zone`, and `port` but an InboundRule/Rule doesn't do this. It looks like they should all use the same code. But maybe I'm not aware of things with this new addition of InboundRule? lmk if so!

packages/kuma-gui/src/app/data-planes/views/DataPlaneInboundSummaryOverviewView.vue
schogges: Ah, yeah that's better 👍

packages/kuma-gui/src/app/data-planes/views/DataPlanePoliciesView.vue
schogges: Yes 👍

packages/kuma-gui/src/test-support/mocks/src/resources.ts
schogges: As we discussed, the `count` for `policyNames` should rather behave like the `count` of `faker.helpers.arrayElements` than like `faker.number.int`. It's been refactored now 👍

packages/kuma-gui/src/app/rules/data/Rule.ts
schogges: I agree and separated the `InboundRules`. It was not as straightforward as `ResourceRules` because we don't use it in `RuleList`. So `InboundRules` have to be closer to `Rules`. However, this is the job of the data layer to shape it for further use.

packages/kuma-gui/src/test-support/FakeKuma.ts
```diff
@@ -507,7 +520,7 @@ gbXR5RnEs0hDxugaIknJMKk1b0g=
       ),
     ]
 
-    return this.faker.helpers.arrayElements<typeof items[number]>(items, minmax({ min: 1, max: items.length }, count))
+    return this.faker.helpers.arrayElements<typeof items[number]>(items, minmax({ min: 1, max: items.length /* 20 */ }, count /* MAX_SAFE_INTEGER */))
```
johncowen, line 523: nit: I think these comments are left over from us messing around yesterday?

johncowen (approved):
I let one more tiny nit, apart from that I think we should merge this as is.

---

Whilst reviewing this I went back and forth a tonne of times, wondering what was the best way to structure the while "Rules Thing", not necessarily the work you've done here in the this PR, but more the stuff you've had to work on top of.

I'm still unsure and I need to do some historical sleuthing, but I "think" what's happened is that we originally just had the `to`/`from` rules which where almost identical, so we flattened their tree structure out _and_ made them look the same (filter by type). Which could have made sense at the time, but I'm thinking now we possibly need to go down the road of still flattening them, but not trying to make them look the same (separate arrays for each type). I'm still finding it hard to look at all 4 types of Rules and see where they are the same and where they aren't!

Since then we've added on `toResourceRules` and now `inboundRules`. From what I remember `toResourceRules` was pretty much "lets add this net new" i.e. only to the Outbound Summary Drawer, and I think `inboundRules` is more rolled out everywhere i.e. the DataPlanePolicyTab and the InboundDrawer.

I think all in all the more we are bolting on new things to the old thing, the approach we took with the old thing might not be making as much sense anymore.

Anyway, wanted to write something out to explain what I'm thinking, and wanted to be clear that its more me being thoughtful over the way we have done things previous to this rather than the way we have done it here.

Apart from that lets :shipit: !

packages/kuma-gui/src/test-support/FakeKuma.ts
schogges: Ah, thanks for catching this 😄

PR #3607, opened 2025-02-27
chore(ci): iterative improvements to releasing

johncowen:
~**DO NOT MERGE at least until the code marked `DO NOT MERGE` is removed**~ 👈 These have been removed

Further iterates on our `release.yml` workflow. The intention here is to simplify it enough so it literally calls `make build` and `make release`. The only reason we split these in two is so we can use GH Actions to generate an SBOM instead of the Makefile itself.

This approach means that we can potentially loop through several packages in our workspace building and releasing each one with scripts controlled by the Makefile itself, not CI YAML (which for reasons cannot be totally versioned)

Separately I also added a similar Makefile target so you can test verify the SBOM locally using `syft`. Note: This is not what we use in CI and is only for testing/verification purposes.

Lastly, this would be good to get in the before we cut for release so that moving forwards subsequent branches only need to support `make build` and `make release`, but no big deal if this slips into 2.11

---

~Firstly looking for an "approval in principle" at which point we'll have to remove some test code here/uncomment some non-test code.~ This is sorted, no looking for a proper approval

### DIFF
--- a/packages/config/src/mk/install.mk
+++ b/packages/config/src/mk/install.mk
@@ -1,10 +1,14 @@
 .PHONY: .install
 .install: check/node
-	@npm install
+	@cd $(NPM_WORKSPACE_ROOT) \
+		&& npm $(if $(CI),clean-install,install) \
+					--ignore-scripts
 
-.PHONY: .install/sync
-.install/sync:
-	npm clean-install
+.PHONY: .update
+.update: check/node
+	@cd $(NPM_WORKSPACE_ROOT) \
+		&& npm install \
+					--ignore-scripts
 
 .PHONY: .dedupe
 .dedupe:
--- a/packages/kuma-http-api/.gitignore
+++ b/packages/kuma-http-api/.gitignore
@@ -1,1 +1,1 @@
-repo
+kuma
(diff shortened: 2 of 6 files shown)


packages/config/src/mk/install.mk
```diff
@@ -1,10 +1,14 @@
...
+					--ignore-scripts
 
-.PHONY: .install/sync
-.install/sync:
-	npm clean-install
+.PHONY: .update
+.update: check/node
+	@cd $(NPM_WORKSPACE_ROOT) \
+		&& npm install \
+					--ignore-scripts
```
johncowen, line 11: Oh I forgot to say I added `---ignore-scripts` throughout

packages/kuma-http-api/.gitignore
```diff
@@ -1 +1 @@
-kuma
+repo
```
schogges, line 1: Why is this needed? 🤔

johncowen: A good spot, can't remember why I added this here, I've undone this one, ty!

schogges (approved):

PR #3623, opened 2025-03-04
chore: remove unnecessary XAction import

johncowen:
I noticed we had an unnecessary XAction import. `X` components are global and therefore injectable, unless of course they are explicitly imported like in this instance.


schogges (approved):
Nice spot 👍

PR #3624, opened 2025-03-04
refactor: replace prismjs with shiki

schogges:
We got a [security warning](https://github.com/kumahq/kuma-gui/security/dependabot/106) for the current version of `prismjs` and took the chance to make the switch to `shiki` as we already have that in our dependencies. The project also looks more active and maintained.


packages/kuma-gui/src/app/x/components/x-code-block/XCodeBlock.vue
```diff
@@ -75,10 +76,12 @@ const emit = defineEmits<{
...
 const isProcessing = ref(false)
 
-async function handleCodeBlockRenderEvent({ preElement, codeElement, language, code }: CodeBlockEventData): Promise<void> {
+async function handleCodeBlockRenderEvent({ codeElement, language, code }: CodeBlockEventData): Promise<void> {
   isProcessing.value = true
 
-  highlightElement(preElement, codeElement, code, language as AvailableLanguages)
+  const highlighted = await highlightElement(code, language as AvailableLanguages)
+  // eslint-disable-next-line no-unsanitized/property
+  codeElement.innerHTML = DOMPurify.sanitize(highlighted)
```
johncowen, line 84: FYI: Looks to me like shiki's `codeToHtml` escapes the HTML? They also have a couple of examples on their docs site using `innerHTML` with no further encoding.

schogges: Yeah, you are probably right. I think it stringifies all provided code and adds safe HTML tags for the highlighting. So I think we can remove the `DOMPurify` dependency and sanitization 👍

packages/kuma-gui/vite.plugins.ts
```diff
@@ -82,7 +82,7 @@ const server = (
         body = body.replace('<meta charset="utf-8" />', `<meta charset="utf-8" /><meta property="csp-nonce" nonce="${nonce}">`)
         res.setHeader('Content-Security-Policy', [
           "default-src 'self'",
-          "script-src 'self'",
+          "script-src 'self' 'wasm-unsafe-eval'",
```
johncowen, line 85: I think we should also consider using the JS engine here (not the wasm one)

schogges: We could, but I thought limiting it to wasm would be "better" 🤔

johncowen: Apparently not:

> The JavaScript engine is best when running in the browser and in cases when you want to control the bundle size.

https://shiki.matsu.io/guide/regex-engines#javascript-regexp-engine

packages/kuma-gui/src/app/x/components/x-code-block/XCodeBlock.vue
```diff
@@ -75,10 +75,32 @@ const emit = defineEmits<{
 
 const isProcessing = ref(false)
 
-async function handleCodeBlockRenderEvent({ preElement, codeElement, language, code }: CodeBlockEventData): Promise<void> {
+async function handleCodeBlockRenderEvent({ codeElement, language, code }: CodeBlockEventData): Promise<void> {
   isProcessing.value = true
 
-  highlightElement(preElement, codeElement, code, language as AvailableLanguages)
+  // TODO(@schogges): use createJavascriptRawEngine for further optimization of bundle size (requires pre-compiled langs)
+  const engine = createJavaScriptRegexEngine()
```
schogges, line 82: I couldn't find the `pre-compiled` langs in the current artifact as described in the docs (v3), but to use the js raw engine I have to use them. So I'll leave that as a TODO for now.

johncowen: Sounds good to me 👍 . 

One little thing here, we create the highlighter every single time the KCodeBlock fires a renderEvent, do you know if its possible/is already being cached that somehow? (or can we can we cache it somehow)

Maybe this isn't necessary, but at first glance I'm guessing we don't need to create a brand new engine and highlight on every render event.

Not a blocker on this and is defo something else we can circle back on it necessary.

schogges: I think the render event has not been triggered multiple times, but anyways I agree and separated it. We definitely shouldn't recreate the highlighter in case the render event is called again 👍

johncowen:
We paired on this and made it render the code whilst waiting for the highlighter to be ready, this way you see the code as soon as possible.

We also figured out what the processing stuff was doing, it shows a small spinner in the CodeBlock, so we kept this in but wrapped it around async code so it works as intended.

johncowen (approved):
LGTM, so glad we removed PrismJS also one less dependencies seeing as we already use shiki in vitepress 🎉

Issue #2934, opened 2024-09-13
Policy table should handle omitted top-level targetRefs


lobkovilya:
### Description

Since Kuma 2.9 it's possible to create policy without the top-level targetRef:
```yaml
type: MeshTimeout
name: mt1
mesh: default
spec:
  to:
    - targetRef:
        kind: MeshService
        name: redis
```
which is an equivalent of:
```yaml
type: MeshTimeout
name: mt1
mesh: default
spec:
  targetRef:
    kind: Mesh
  to:
    - targetRef:
        kind: MeshService
        name: redis
```

In policy table `Target ref` we want to see `Mesh` instead of `–`:
<img width="667" alt="image" src="https://github.com/user-attachments/assets/c29a0579-e8ff-43e4-9893-e4c22e168d77">

PR #3625, opened 2025-03-04
refactor(policy-summary): always show structured format option

### LINKED ISSUE #3608: Always show structured format on PolicySummaryView
### Description

Follow-up to https://github.com/kumahq/kuma-gui/pull/3581

Due to the fact that we are missing a `spec` field on legacy policies on mode universal, we conditionally include the structured view format. Ideally we can always show structured, because in order to keep the state of select in sync with the URL we currently the user preference.

schogges:
To improve on consistencies we decided to always offer the option `structured` as a format in summary drawers. Currently that was missing in `PolicySummaryView` and `DataPlanePolicySummaryView`. Now we just want to only conditionally render the codeblock when there is no `spec` available, which is only the case for legacy policies on universal.

Furthermore this adds the policy `type` to the structured view, such that users can always see the type of the policy that they are looking at.

Closes #3608


schogges:
> I think in the case of a legacy policy the `targetRef` shouldn't show
> 
> Screenshot 2025-03-04 at 14 58 28

Looking at https://github.com/kumahq/kuma-gui/issues/2934 I think we always show `Mesh` in case there is no `spec.targetRef` 🤔 just not sure if this also counts for legacy policies.

johncowen:
Pretty sure legacy policies don't have a `targetRef`

https://kuma.io/docs/2.9.x/policies/general-notes-about-kuma-policies/

Screenshot 2025-03-04 at 15 16 22

Most of them have sources/destinations, the `type: Mesh` ones I believe apply to the mesh itself, but its still not using a `targetRef`, I guess you could switch out the word `targetRef` for `type` just for those legacy `type: Mesh` ones, but at this point, I kinda don't mind either way

johncowen:
I've pushed a commit here that plumbs through `policyType.isTargetRef` (which is basically `legacy: boolean`) this way we know whether to show the `targetRef: Mesh` badge in the structured view or not.

As the structured view for legacy policies was pretty bare, we decided to add the `logging`, `metrics` etc to display a codeblock for legacy policies also as it wasn't much additional to do. This last point isn't totally what we said here https://github.com/kumahq/kuma-gui/issues/3608#issuecomment-2694735083 but we both preferred the outcome so decided to keep it.

schogges:
> I've pushed a commit here that plumbs through `policyType.isTargetRef` (which is basically `legacy: boolean`) this way we know whether to show the `targetRef: Mesh` badge in the structured view or not.
> 
> As the structured view for legacy policies was pretty bare, we decided to add the `logging`, `metrics` etc to display a codeblock for legacy policies also as it wasn't much additional to do. This last point isn't totally what we said here [#3608 (comment)](https://github.com/kumahq/kuma-gui/issues/3608#issuecomment-2694735083) but we both preferred the outcome so decided to keep it.

Can't approve, so I'll leave a ✅

PR #3637, opened 2025-03-06
chore(lintstaged): lint only staged files

schogges:
After moving more and more scripts over to makefile targets, we've noticed that lint-staged takes longer to check the staged files than before. This happens because currently our `lint/js` target lints all files (essentially `npx eslint .`) and doesn't consider the file argument passed over from lint-staged.

Unfortunately there is no option to run `vue-tsc` (same for `tsc`) on specific files only, it always runs for the whole project.


johncowen (commented):
Whats the best way to test this locally? I was using (from packages/kuma-gui)

```bash
make lint/js src/main.ts
```

And its currently showing me:

```bash
make: Nothing to be done for `src/main.ts'.
```

P.S. Oh that is actually linting `src/main.ts`, just the message it gives me is a bit weird 🤔 do you see that also?

schogges:
> Whats the best way to test this locally? I was using (from packages/kuma-gui)
> 
> ```shell
> make lint/js src/main.ts
> ```
> 
> And its currently showing me:
> 
> ```shell
> make: Nothing to be done for `src/main.ts'.
> ```
> 
> P.S. Oh that is actually linting `src/main.ts`, just the message it gives me is a bit weird 🤔 do you see that also?

Yeah I see that too. Thought that it maybe comes from the `--fix` option, but I don't see that message via `npx eslint --fix src/main.ts` 🤔 But I've seen this message already in the past when staging files that have an issue and commit (trigger lint-staged).

schogges:
So looks like this is coming from `make` and the fact that the files that we add to the end (and so does lint-staged) are treated as targets.

when I add

```make
.PHONY: foo
foo:
```

without a recipe and `make foo`, I get `make: Nothing to be done for `foo'.`

johncowen:
I just by chance came across `eslint --cache`:

https://eslint.org/docs/latest/use/command-line-interface#--cache

> Store the info about processed files in order to only operate on the changed ones. Enabling this option can dramatically improve ESLint’s run time performance by ensuring that only changed files are linted.

I'm thinking adding this flag to our lint target and then gitignoring the cache could be a better solution for this? I just briefly tried it and it does look like it works as advertised, wdyt?

schogges:
> I just by chance came across `eslint --cache`:
> 
> https://eslint.org/docs/latest/use/command-line-interface#--cache
> 
> > Store the info about processed files in order to only operate on the changed ones. Enabling this option can dramatically improve ESLint’s run time performance by ensuring that only changed files are linted.
> 
> I'm thinking adding this flag to our lint target and then gitignoring the cache could be a better solution for this? I just briefly tried it and it does look like it works as advertised, wdyt?

Nice 👍 I think we can add that additionally, but in order to actually kick in it requires to maintain that cache.

>If you run ESLint with --cache and then run ESLint without --cache, the .eslintcache file will be deleted.

johncowen:
> but in order to actually kick in it requires to maintain that cache.

The usecase/issue here was for the git hooks right?

If we add `--cache`, the first time we use it the cache will be created, and we get super fast JS linting from there on in (we don't need to do anything with the cache ourselves its all automatic), and it also works when we call `make lint` (not just on a git hook)

schogges:
> > but in order to actually kick in it requires to maintain that cache.
> 
> The usecase/issue here was for the git hooks right?
> 
> If we add `--cache`, the first time we use it the cache will be created, and we get super fast JS linting from there on in (we don't need to do anything with the cache ourselves its all automatic), and it also works when we call `make lint` (not just on a git hook)

Yes, the issue here was about the git hook. But anyways I want to add that flag 👍 but I would probably keep the other changes I did. They are agnostic to the linting tool we use. Wdyt?

schogges:
As discussed I've removed the args part and we just leave it to `--cache`. Reasons:

- not have to maintain the list of file types in another place (looking at `.lintstagedrc.json` 👀 )
- `--cache` only checks the files we changed, so we keep that in the tool itself
- we keep lint-staged mainly to trigger the linting for us whenever we want to commit something
- most likely there is no need to lint specific file(s) without changing them (manual `make lint foo/bar.js` will not likely happen)

johncowen (approved):

PR #3659, opened 2025-03-13
chore: bump node to 22.14

### LINKED ISSUE #3510: Upgrade node to current LTS (v22)
### Description

v22 was promoted to LTS not long ago, whilst not super urgent we have started to get some warnings during on npm install (a new npm should come the new node), so we should bump after the upcoming release.

This should be just a bump of our `.nvmrc`, but there might be the odd engines entry in a `package.json`

schogges:
See title

Closes #3510


johncowen (approved):
🎉

PR #3660, opened 2025-03-13
feat(x-code-block): update shiki version, theme

schogges:
As part of the version bump of shiki I changed some things:

- bump `shiki` to `3.2.1`
- changed the theme to `catppuccin-latte` (light) and `catppuccin-mocha` (dark) 
- changed [`moduleResolution`](https://www.typescriptlang.org/tsconfig/#moduleResolution) to `bundler`. Previous `node` is deprecated.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,10 @@ updates:
       typescript-eslint:
         patterns:
           - "@typescript-eslint/*"
+      shiki:
+        patterns:
+          - "shiki"
+          - "@shikijs/*"
 
   - package-ecosystem: "github-actions"
     open-pull-requests-limit: 50
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,10 +11,6 @@ updates:
       typescript-eslint:
         patterns:
           - "@typescript-eslint/*"
-      shiki:
-        patterns:
-          - "shiki"
-          - "@shikijs/*"
 
   - package-ecosystem: "github-actions"
     open-pull-requests-limit: 50
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,7 +4,6 @@
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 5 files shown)


johncowen (approved):
Nice 👍 

One tiny nit that I think you might have mentioned in our sync.

I noticed in the package-lock we are installing loads of `oniguruma` stuff which we don't use.

You mentioned you went back and forth a little here which different installation methods, so it might have been related to what you were saying there.

Anyway lemme know on that one but not a blocker as I would assume we are not importing what we don't need, its more about potential dependabot PRs in the future. We can always circle back and alter this if it turns out we can.

johncowen:
Oh wait, I think we might be getting a double scrollbar thing here

johncowen:
Yeah we are I think, hard to screengrab, but if you hover over the text and try and scroll you can see this little ghost-like scrollbar that only has a tiny bit of scroll, whereas if you hover over any other part of the window you can scroll as expected.

I noticed that this issue is in `master` already tho, so I'd say lets merge this PR, and see if you can then duplicate what I'm seeing

schogges:
> Yeah we are I think, hard to screengrab, but if you hover over the text and try and scroll you can see this little ghost-like scrollbar that only has a tiny bit of scroll, whereas if you hover over any other part of the window you can scroll as expected.
> 
> I noticed that this issue is in `master` already tho, so I'd say lets merge this PR, and see if you can then duplicate what I'm seeing

Oh, let me have a look at that, brb

schogges:
> > Yeah we are I think, hard to screengrab, but if you hover over the text and try and scroll you can see this little ghost-like scrollbar that only has a tiny bit of scroll, whereas if you hover over any other part of the window you can scroll as expected.
> > I noticed that this issue is in `master` already tho, so I'd say lets merge this PR, and see if you can then duplicate what I'm seeing
> 
> Oh, let me have a look at that, brb

Will circle back on this, thanks for spotting 👍

PR #3661, opened 2025-03-13
chore(deps): create dependabot eslint group

schogges:
Both `eslint` and `@eslint/js` are released from the same repo and always using the same version, so I think we can create a group to reduce the noise.


johncowen (approved):
🎉 Nice! 🙉 😆

PR #3662, opened 2025-03-13
chore(docs): adds improved docs for installing nvm

johncowen:
Adds better instructions/info for installing and running `nvm`

### DIFF
--- a/packages/config/src/mk/check.mk
+++ b/packages/config/src/mk/check.mk
@@ -4,7 +4,7 @@ check/node: NODE_VERSION:=v$(shell head -n1 $(NPM_WORKSPACE_ROOT)/.nvmrc)
 check/node:
 	@node -v | grep $(NODE_VERSION) &> /dev/null || ( \
 		echo "Make sure node-$(NODE_VERSION) is installed (please see the root .nvmrc for nvm installation)"; \
-		echo "Once the correct version of node is installed,  re-run your make target"; \
+		echo "Once the correct version of node is installed, re-run your make target"; \
 		exit 1; \
 	)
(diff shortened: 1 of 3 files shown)


johncowen:
FYI this has been up for a while just because of other priorities, and I'm trying to get these few that have been around for a while over the line now.

packages/config/src/mk/check.mk
```diff
@@ -1,10 +1,10 @@
...
-NPM_VERSION?=$(shell cat $(NPM_WORKSPACE_ROOT)/package.json | jq -r '.engines.npm')
-
 .PHONY: check/node
+check/node: NPM_VERSION:=$(shell cat $(NPM_WORKSPACE_ROOT)/package.json | jq -r '.engines.npm')
+check/node: NODE_VERSION:=v$(shell head -n1 $(NPM_WORKSPACE_ROOT)/.nvmrc)
 check/node:
 	@node -v | grep $(NODE_VERSION) &> /dev/null || ( \
-		echo "Make sure node-$(NODE_VERSION) is installed (see .nvmrc)"; \
+		echo "Make sure node-$(NODE_VERSION) is installed (please see the root .nvmrc for nvm installation)"; \
+		echo "Once the correct version of node is installed,  re-run your make target"; \
```
schogges, line 7: super nit, double space detected

```suggestion
		echo "Once the correct version of node is installed, re-run your make target"; \
```

schogges (dismissed):

packages/config/src/mk/check.mk
johncowen: 🦸 ! 😆

schogges (approved):
👍

PR #3672, opened 2025-03-14
fix(x-code-block): highlighted codeblock additional scrollbar

schogges:
In https://github.com/kumahq/kuma-gui/pull/3660 we spotted that there is a little additional scrollbar showing up on Mac if the system scrollbar setting is **not** set to `Always`.

https://github.com/user-attachments/assets/597d523f-63b7-47d8-9fd5-a71a84e96ae0

---

I've also moved the `background-color: unset` to the global `shiki, shiki span` ruleset.


johncowen (approved):
👍

PR #3676, opened 2025-03-17
fix(vitepress): old usage of shiki getHighlighter

schogges:
When working on https://github.com/kumahq/kuma-gui/pull/3660 I missed to update the usage of shiki in `.vitepress`.


packages/kuma-gui/.vitepress/theme/components/Story.vue
```diff
@@ -44,12 +45,13 @@ const props = withDefaults(defineProps<{
...
 const htmlSource = ref<string>('')
-const highlighter = ref<Awaited<ReturnType<typeof getHighlighter>> | undefined>()
+const highlighter = ref<Awaited<ReturnType<typeof createHighlighterCore>> | undefined>()
 
 ;(async () => {
-  highlighter.value = await getHighlighter({
-    themes: ['nord'],
-    langs: ['javascript']
+  highlighter.value = await createHighlighterCore({
+    themes: [import('shiki/themes/nord.mjs')],
```
schogges, line 52: Note: I didn't change the current theme `nord`, but probably it makes sense to use `catppuccin-mocha` or `catppuccin-latte`?

johncowen: Yeah whichever, I guess I just took this originally from whichever example I was reading from

I guess we could eventually move this to be part of `x` and we can source it from `x`'s service container as `useSyntaxHighligher` or something, then we centralize this.

johncowen (approved):
LGTM!

packages/kuma-gui/.vitepress/theme/components/Story.vue
schogges: Ok, sounds good. I'll leave it as it for now and we can circle back here again when we centralize it 🙂

PR #3687, opened 2025-03-18
chore(deps): move husky and lint-staged to root

schogges:
I hit a case where I installed the dependencies and `lint-staged` ended up in the `packages/kuma-gui/node_modules` (not in the root). This caused that the pre-commit hook failed: `.husky/pre-commit: line 1: lint-staged: command not found`

We decided to move certain dependencies, that operate on the whole repo or are meant to be added to all packages (i.e. `eslint`) to the repository root as `devDepdendencies`. Therefore I am moving `husky` and `lint-staged` to the repository root now.


packages/config/package.json
```diff
@@ -42,6 +42,8 @@
     "cypress-fail-fast": "^7.1.1",
     "cypress-terminal-report": "^7.0.4",
     "eslint": "^9",
-    "stylelint": "^16.10.0"
+    "stylelint": "^16.10.0",
+    "husky": "^9.1.7",
+    "lint-staged": "^15.5.0"
```
johncowen, line 47: Considering the conversation we had this afternoon around peerDependencies and removal of those, I think these should either go:

- Once in root
- Multiple times in each individual project.

I think I'd vote for the former, but I'm not that precious about it, but I'm more "less code is best"

johncowen: > In case we want to use husky and lint-staged (and the other peers) further downstream, we would have to add them there too 

Yeah for sure, but if you think about it this is the model we are moving to i.e. applications no longer depend on "Kuma The Application" but instead each application depends on the individual "micro-frontends" that we have. Each application is then completely standalone so instead of:

Application 1 > Application 2 > Kuma The Application > micro-frontends

we have

- Application 1 > micro-frontends
- Application 2 > micro-frontends
- Kuma The Application > micro-frontends

johncowen: (sorry) Although I'm not sure the above is totally correct I think what we want is:

- Application 1 > config + micro-frontends
- Application 2 > config + micro-frontends
- Kuma The Application > config + micro-frontends

Ideally config would house the CLI tools we need so that are centralized in one place. But something I realized is (which I think helps with understanding): with our Makefiles CLI tooling is executed from the location of the root Makefile _not_ from the location of an `import` statement. If you think about this carefully you begin to understand certain differences to normal node toolchain assumptions.

Thinking about it from a node standpoint, you would assume the Makefile includes (not imports) its `mk/` scripts from `@kumahq/config`, and then things in those scripts are executed from within `@kumahq/config`. If it was JS/node this would mean it would use the executables installed by `@kumahq/config`.

But in reality, we are not importing other scripts, we are just including them so the root of the execution is wherever the root Makefile is, _not_ where the script/included file is.

This means if a script in `@kumahq/config` is executing eslint, its not executing `@kumahq/config/node_modules/.bin/eslint` its executing `(./Makefile)node_modules/bin/eslint`. As we are using `npx` if the executable doesn't exist in the node_modules of where the Makefile is, it will then use node resolution to find the executable higher up in the tree.

I'm wondering if there is a simple way we can make the Makefile scripts act more like node, which would be ideal. But if its not simple, we may just have to accept this fact, and keep it in mind when thinking about things.

schogges: I think I get your point and it would be great, but when moving those dependencies to the root, this means that these dependencies are no longer part of that chain `App > config + micro-frontends` 🤔

johncowen:
@schogges did we say we were gonna close this one seeing as it changes things to use peerDependencies (that we've since said for the moment at least we are going to avoid)

schogges:
> @schogges did we say we were gonna close this one seeing as it changes things to use peerDependencies (that we've since said for the moment at least we are going to avoid)

Yeah, either close it or change it to move these deps to the repo root. In any case these dependencies are in the wrong place.

johncowen:
oh did we say we'd wait to see if it became an issue again, I think we figured that when you hit whatever it was you weren't on master at the time?

schogges:
> oh did we say we'd wait to see if it became an issue again, I think we figured that when you hit whatever it was you weren't on master at the time?

Yeah, I think we said that. But honestly I think even if we don't hit it again, these dependencies are in the wrong place.

schogges:
@johncowen moved them to the repo root 🙂

johncowen (approved):
👍 interested to see what happens elsewhere once we merge this 👀

PR #3688, opened 2025-03-18
chore(ci): add github-actions-ensure-sha-pinned-actions

schogges:
In the first place we want to ensure that we only depend on pinned github actions (externals). We've recently added linting for that, but turns out there is an action that takes care of that.


.github/workflows/main.yml
```diff
@@ -21,6 +21,17 @@ permissions:
...
+    name: Harden Security
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Ensure SHA pinned actions
+        uses: zgosalvez/github-actions-ensure-sha-pinned-actions@25ed13d0628a1601b4b44048e63cc4328ed03633 # v3.0.22
+        with:
+          allowList:
+            kumahq/
+
```
schogges, line 34: Wasn't too sure where to put it, so this is now in `main`, but I could also start a new shareable workflow 🤔

packages/config/src/eslint.cjs
```diff
@@ -135,10 +131,6 @@ function createEslintConfig(
               // see ./package.schema.json patternProperties examples
               schema: packageSchema,
             },
-            {
-              fileMatch: ['.github/**/*.{yaml,yml}'],
-              schema: workflowSchema,
-            },
```
johncowen, line 141: Why are we removing these? They run locally on a git commit hook

schogges: Fair point, will add them back 👍

johncowen: Also realised if we remove these it would mean that we also lose the rule which prevents GH Actions from using `pull_request_target`, so we definitely need to keep this.

.github/workflows/main.yml
johncowen: 🤔 yeah me neither, umm.

Its a linter, and this is where we do linting also, so I think for now at least its fine here. 

Although I'm tempted to say it should be in its own workflow file, 🤔

.github/workflows/main.yml
```diff
@@ -21,6 +21,17 @@ permissions:
...
+  harden_security:
+    name: Harden Security
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Ensure SHA pinned actions
+        uses: zgosalvez/github-actions-ensure-sha-pinned-actions@25ed13d0628a1601b4b44048e63cc4328ed03633 # v3.0.22
+        with:
+          allowList:
+            kumahq/
```
johncowen, line 33: Do we need this allow list currently?

johncowen (commented):
Do we need this job to the needs for branch protection down at the end of this workflow? I think we do?

Also, a bit of a nit, but usually if the "name of the thing" is already the "name of the thing", then usually we don't add another "name of thing" 😅 as it just becomes noise  

So what I'm saying is if the `uses` is called `github-actions-ensure-sha-pinned-actions` then it doesn't need a `name` that says "Ensure SHA pinned actions", similarly for `Harden Security`.

I know in this one single case it doesn't seem like a big deal, but these workflows used to be full of the repeat naming, plus comments to name the thing again, and it was just noise making it hard to grok what was going on at a glance.

So anyway, but of a nit, and I can see these are just from the example, but ideally if we already have the "name of the thing" and there is no benefit in adding another, we generally don't.

.github/workflows/main.yml
schogges: Yes, because of https://github.com/kumahq/kuma-gui/blob/d3cefe4baeb98bba7fa4740858041525fd41b1a1/.github/workflows/lifecycle.yml#L18

.github/workflows/main.yml
schogges: Ok, I've renamed it to `lint_global` and going to leave it here for now. I was tempted to put it as part of `lint-tests`, but this runs as matrix.

.github/workflows/main.yml
johncowen: Ah! 👍

johncowen:
Just to check, are we still adding this job to the branch-protection.needs at the end of the workflow here?

schogges:
> Just to check, are we still adding this job to the branch-protection.needs at the end of the workflow here?

Ah! 🙈 Thanks for the ping on this again 👍

johncowen (approved):

PR #3690, opened 2025-03-18
chore(ci): adds check on jobs.use and adds exclusion for kumahq/

johncowen:
Previously we only checked `jobs.steps.uses`


schogges (approved):
👍

Issue #2001, opened 2024-01-11
Add a vite based server/middleware for our mocks


johncowen:
### Description

Everytime I add new mocks that are slightly more involved, I wish I could just navigate to the API endpoint in my browser and view the response so I can see what is happening.

Currently we can't do this because we use MSW so you have to go to a page in the GUI that you know makes the HTTP request you are building out and view the response in web inspector. If this is a new response that doesn't have any code in the GUI to request it yet, then you have to build this out first.

Vite supports serving up other files other than our HTML and JS files, and I think (from what I remember) it allows you to add middlewares for specific endpoints. We should be able to repurpose our existing mock utility functions to make a middleware for Vite also, just like we do to run them via Cypress. (also I originally had our mocks running in an express server before bringing them to MSW). This would allow us to just navigate to the URL of the API endpoint you are building and see the response.

Whilst it doesn't have to be done here, if you have this running we may aswell not use MSW during local development work and only use it for the PR previews.

---

Decided to make an issue for this rather than run off and find the express code I had and move it to Vite or just think "I'll do it next time I'm building out an endpoint" which never happens 😆

lahabana:
@johncowen please complete with what's left

johncowen:
This currently works fine in Kuma, and is actually running:

From https://github.com/kumahq/kuma-gui/pull/2314: 

> To try this out you need to set KUMA_API_URL=http://localhost:8080. 8080 being the port that vite is using, and also set KUMA_MOCK_API_ENABLED=false to turn off MSW.

Ideally though we need to make this work with our service injection framework, along with vite.

We don't use our service injection currently with vite anywhere. In order to use our service injection framework with vite, it cannot use any of vite's `@` import aliasing capabilities. We should where possible use "native" package names for the things we need here (`fake-api` and our service injection stuff) i.e. `@kumahq/fake-api` etc.

This then hits another problem in that it seems to be non-trivial to write a package in a workspace and share in inside the workspace without having a TS complication step. The good news is, its seems like vite6 (potentially combined with nodes new TS stripping functionality) means it will be easier to implement this.

---

Possibly blocked by https://github.com/kumahq/kuma-gui/pull/3572, the upgrade here includes new functionality to make it much easier to load workspace'd modules without compiling the Typescript first

schogges:
I've created a follow-up issue for the next iteration and I'll close this one.

@johncowen we can potentially create an umbrella for the bigger picture?

PR #3693, opened 2025-03-20
refactor: move vite plugin, cypress interceptor and msw handlers to fake-api package 

schogges:
Creates a new package `@kumahq/fake-api` that serves a vite plugin, msw handlers and cypress interceptor to be used as a server for our mocks. This is still opt-in via `KUMA_API_URL=http://localhost:8080` and `KUMA_MOCK_API_ENABLED=false`.

In order to load `.ts` files without compilation I found two options:

- From https://vite.dev/config/#configuring-vite there is `--configLoader runner` which transforms given files on the fly. This worked great for `vite` only actions, but failed when trying to run the unit tests with `vitest`
- From https://nodejs.org/en/learn/typescript/run-natively using the flag `--experimental-strip-types`, which works without any further configuration on `vite`/`vitest`.

I decided to go for `--experimental-strip-types` only because it works for both `vite` and `vitest`.

Part of #2001


packages/kuma-gui/vite.plugins.ts
```diff
@@ -89,7 +89,7 @@ const server = (
           // its only used for vite dev-time live reloading client
           // the sha256 _will be_ required
           // 'sha256-UtFm94bwcb1Z4CU0svC29YMU26pP5RoZDN8zoniSJhU=' 'sha256-qo7STIM1L/OgU9y0De47mqod1UZFLJfTn36bRC42rfA=' 'nonce-${nonce}'
-          "style-src 'unsafe-inline'",
+          "style-src 'self' 'unsafe-inline'",
```
schogges, line 92: Note: I found when running the preview that csp blocked css on :5681.

packages/config/src/mk/build.mk
```diff
@@ -1,15 +1,17 @@
 .PHONY: .build
 .build: $(if $(NOPRUNE),,prune) install
-	@npx vite \
-		-c ./vite.config.production.ts \
-		build
+	@NODE_OPTIONS=--experimental-strip-types \
+		npx vite \
```
johncowen, line 4: WOuld be great in the future at some point if we centralize some sort of `NODE_RUNNER` or `JS_RUNNER`/`CLI_RUNNER` or something, so that we specify this in one place. Totally a future things though.

johncowen: Actually, does it matter that we don't always need these options when using `npx`, anyway something to think about.

johncowen: @schogges was saying, maybe we think of it like a `VITE_RUNNER`, which sounds like what we are saying here, although ideally without `VITE_` ie agnostic to the tool, anyway food for thought.

packages/fake-api/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+    "target": "esnext",
+    "module": "nodenext",
+    "strict": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
+    "rootDir": "./src"
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules"]
+}
```
johncowen, line 12: When we were talking at some point we need to start combining tsconfig files, try and do something that is simiar to how the create-vue-app generator does it.

packages/fake-api/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "nodenext",
+    "strict": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
```
johncowen, line 7: we are gonna remove this line for now at least

packages/kuma-gui/vite.config.production.ts
```diff
@@ -37,7 +38,8 @@ export const config: UserConfigFn = () => {
         },
       }),
       fakeApi({
-        fs,
+        mockEndpointArgs: dependencies,
+        mockEndpointsDictionary: fs,
```
johncowen, line 42: we are gonna cvall these user facing props, dependencies and fs (based on the original approach the fakeApi plugin)

packages/config/package.json
```diff
@@ -37,8 +37,8 @@
     "stylelint-config-recommended-vue": "^1.6.0"
   },
   "peerDependencies": {
-    "@badeball/cypress-cucumber-preprocessor": "^21.0.2",
-    "cypress": "^13.15.2",
+    "@badeball/cypress-cucumber-preprocessor": "^22",
+    "cypress": "^14",
```
schogges, line 41: Note: bumped `cypress` and `@badeball/cypress-cucumber-preprocessor` to align with the version we have actually installed in the other package(s). Specifically `cypress` is now listed in `fake-api`'s `peerDependencies` with the current version and not aligning them would lead to conflicts.

packages/fake-api/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "nodenext",
+    "allowImportingTsExtensions": true,
+    "noEmit": true,
```
schogges, line 6: Using `module: "nodenext"` requires file extensions on imports and `allowImportingTsExtensions` makes this possible, but can't go without `noEmit` or `emitDeclarationOnly`. This was also necessary to add in `kuma-gui/tsconfig.json` and will most likely be required to be added in all future dependent packages.

schogges (commented):
Some notes

packages/fake-api/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "msw": "^2.7.3"
+  },
+  "devDependencies": {
+    "@kumahq/config": "*",
+    "eslint": "^9.22.0",
+    "typescript": "^5.8.2"
+  },
+  "peerDependencies": {
+    "vite": "^6",
+    "cypress": "^14"
```
schogges, line 34: Theoretically we can add `peerDependencyMeta` to mark both `vite` and `cypress` as optional to avoid auto-installation. At least this is a potential improvement for dependents that use only the `vite` plugin **or** `cypress` interceptor.

schogges:
@johncowen I had to resolve some conflicts after rebasing. When resolving I made some minor changes to the manifests and therefore some of my comments above are marked as outdated, but the changes are essentially the same.

johncowen (commented):
Firstly, I think this looks great, everything is getting a lot better organized now and you've already begun reducing the repetitive stuff like Router etc.

I think there is still a little bit more to do here, and I started adding comments, but then I thought it would maybe better to do that together seeing as there are loads of tiny bits, plus I'll be able to explain next steps a little better.

Umm, I left the beginnings of my comments, but nothing is blocking here and I think we should merge as is as its a vast improvement, and continue separately.

I'm just gonna have one last look incase, plus wait for things upstream to unblock a little, and then I reckon we can merge this 👍

packages/kuma-gui/src/app/data-planes/data/index.spec.ts
```diff
@@ -76,7 +77,7 @@ describe('DataplaneOverview', () => {
       params: {
         name: 'zone',
       },
-    }),
+    }, dependencies),
```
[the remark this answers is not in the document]
schogges: Good point for potential improvement 👍 I've abstracted it and was able to remove `dependencies` from the spec files.

packages/config/package.json
```diff
@@ -37,8 +37,8 @@
     "stylelint-config-recommended-vue": "^1.6.0"
   },
   "peerDependencies": {
-    "@badeball/cypress-cucumber-preprocessor": "^21.0.2",
-    "cypress": "^13.15.2",
+    "@badeball/cypress-cucumber-preprocessor": "^22.0.0",
+    "cypress": "^14.0.0",
```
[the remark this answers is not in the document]
schogges: Yep 🙂 I feel this needs further discussion and is not in the scope of this PR

johncowen (approved):
We had a super useful pairing session to make some last amends/niceties here, during which we also found and fixed a potential replace/replaceAll bug and covered a lot of ground from a knowledge share perspective, so super happy how all this went 🙌 

This LGTM! I've approved but just be aware about making sure this all works fine elsewhere before merge. If we need to come back here for any reason to retweak anything for that, lemme know and I can take another look 👍

PR #3705, opened 2025-03-25
chore: merges release-2.10 branch into master finishing off the 2.10 release cycle

johncowen:
See title

---

My personal opinion on these is still that during the release period, we should still work off of master and backport/cherry-pick back to the release branch.

Basically any approach that means we get to keep the separate commits in the git history, I believe this will need to be squashed and therefore we lose all the individual commits (I could be wrong, we'll see now)

---

In the meantime I need to figure out how to sort the DCO failures

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "kuma-gui",
       "workspaces": [
         "packages/*"
       ],
--- a/packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
+++ b/packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
@@ -1,11 +1,11 @@
 data-planes:
   notifications:
-    recommend-reachable-services: !!text/markdown |
-      We've identified outbound services that would benefit from using
-      <a href="{KUMA_DOCS_URL}/production/dp-config/transparent-proxying/#reachable-backends" target="_blank">Reachable Backends</a>
-      and/or
-      <a href="{KUMA_DOCS_URL}/production/dp-config/transparent-proxying/#reachable-services" target="_blank">Reachable Services</a>
-      to dramatically improve performance.
+    recommend-reachable: !!text/markdown |
+      We've identified outbound services that would benefit from using { mode, select,
+        Exclusive { <a href="{KUMA_DOCS_URL}/production/dp-config/transparent-proxying/#reachable-backends" target="_blank">Reachable Backends</a> }
+        ReachableBackends { <a href="{KUMA_DOCS_URL}/production/dp-config/transparent-proxying/#reachable-backends" target="_blank">Reachable Backends</a> }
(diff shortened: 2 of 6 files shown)


johncowen:
> In the meantime I need to figure out how to sort the DCO failures

Yeah so I think because GH uses the GH `users.noreply.github.com` email addresses to do the final merge for all PRs, during a PR the DCO passes fine, as the git `Author` and the `signed-off-by` match up. But once merged the Author is then the `users.noreply.github.com` email which then doesn't match up, but the DCO isn't re-checked on a merge to `master`.

Following that, when we do a release branch merge back to master (like this PR), the commit authors no longer match up with the sign-off, so the DCO complains.

I _think_ this is what is happening 🤔

johncowen:
Yeah, I rebased, which added me as a git Committer, so git Author `users.noreply.github.com`, git Committer my email, this gives me less DCO errors, but still errors for the schogges commits as neither the git Author nor the fir Committer match up with schogges actual email address that is signing off the commits.

johncowen:
> My personal opinion on these is still that during the release period, we should still work off of master and backport/cherry-pick back to the release branch.

If we did this, the original author of the original commit would also be backporting their original commit, therefore at the point of the original PR the git Author would match up with the sign-off, and at the point of the backport the git Author would match up with the original sign-off, so no DCO errors would happen I don't think.

I guess there are two suggestions here:

- We work off of master and backport to release branch with every PR
- We work off of release branch and forward port to master with every PR.

Main suggestion is we do this for every PR individually rather than with a big merge at the end.

johncowen:
Anyway, this is now ready, sorry for the noise 😅

schogges (approved):

schogges:
> > My personal opinion on these is still that during the release period, we should still work off of master and backport/cherry-pick back to the release branch.
> 
> If we did this, the original author of the original commit would also be backporting their original commit, therefore at the point of the original PR the git Author would match up with the sign-off, and at the point of the backport the git Author would match up with the original sign-off, so no DCO errors would happen I don't think.
> 
> I guess there are two suggestions here:
> 
> * We work off of master and backport to release branch with every PR
> * We work off of release branch and forward port to master with every PR.
> 
> Main suggestion is we do this for every PR individually rather than with a big merge at the end.

Sounds good and makes sense to me 👍 but even though we probably don't need all the commits that we add to master during stabilization have to go into the release branch, it's better because we'd avoid conflicts? Or do you think cherry-picking is fine?

johncowen:
> we probably don't need all the commits that we add to master

Ideally I want to keep all the individual commits on master, whereas right now, I only have a `[Squash and merge]` button to push, so I'm gonna end up with one big commit on master instead of every individual one, bit on the release branch they are all individual. I'd like to check with the rest of the team before we even consider changing any of the process here though.

johncowen:
Gonna go ahead and merge this, I added an issue here https://github.com/kumahq/kuma-gui/issues/3710 to track the things spoken about in the comments here

PR #3706, opened 2025-03-27
chore: ignore attrs on XAnonymous

johncowen:
Removes a bunch of Vue dev time warnings, which also show up in Cypress tests (I found this when working on something else soon to be PRed)

### Before

Screenshot 2025-03-27 at 15 22 59

### After

Screenshot 2025-03-27 at 15 22 41


schogges (approved):
👍

PR #3708, opened 2025-03-27
fix(deps): don't use peerDependencies, we don't have consumers

johncowen:
When we made `@kumahq/config` we started adding peerDependencies to define the version of the "hosts" that we provide config/plugins for.

The thing is as dependbot (rightly) doesn't touch peerDependencies, and the node ecosystem can sometimes move quite fast (major bumps are a common occurrence), this means that our peerDependencies can quite quickly get out of sync with what we have defined as root/application/host dependencies. This is causing issues and unnecessary work to keep these in sync manually (no dependbot).

Moreover, we are in total control of the `@kumahq/config` _and_ our host app, and we don't publicly publish `@kumahq/config`, so in reality whenever we bump the dependency in our host app (that needs to be in sync with the peerDependencies) we should be bumping the peerDependencies to the exact same version. This then begs the question, why are we bumping in multiple places, when we can just bump in one. i.e. in our unique situation what's the point of defining peerDependencies if we are in full control of our entire codebase and we don't provide a comsumable package to anyone else (where peerDependencies would then become useful)

Theres probably more to say here, but the above is pretty much the gist of it. All in all we decided we would remove and not use peerDependencies at all.

This PR removes the peerDependencies we were using and moves what we had there into the correct places.

Whilst doing this I spotted that we use certain cypress plugins in two places just "because Cypress" so I also did the necessary work to make it so we could specify these plugins in one single place in `@kumahq/config` instead of also in `@kumahq/kuma-gui`.

As discussed over the past few days, there may end up being places where we need to ensure a dependency is either in the root of the monorepo, or in its individual workspace/application as a result of removing peerDependencies. If/as and when we discover those we just move them as and when.

Following this we said we would reassess in a couple of months to see what effect this has had on our projects.

As a result of this PR the only cypress related dependencies you need to install in one of our applications that needs e2e testing is `cypress` itself and `@badeball/cypress-cucumber-preprocessor`, which we use in our `steps` file. At some point our steps file will also be moved out of the application and into a shareable packge/module. At which point all an application will need to install locally is `cypress` itself plus `@kumahq/config`

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -43,8 +43,5 @@
     "stylelint-config-html": "^1.1.0",
     "stylelint-config-recommended-scss": "^14.1.0",
     "stylelint-config-recommended-vue": "^1.6.0"
-  },
-  "devDependencies": {
-    "eslint": "^9.23.0"
   }
 }
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -43,5 +43,8 @@
     "stylelint-config-html": "^1.1.0",
     "stylelint-config-recommended-scss": "^14.1.0",
     "stylelint-config-recommended-vue": "^1.6.0"
+  },
+  "devDependencies": {
+    "eslint": "^9.23.0"
   }
 }
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -12,16 +12,24 @@
     ".": {
       "require": "./src/index.cjs",
       "import": "./src/index.ts"
+    },
+    "./cypress/e2e": {
+      "import": "./src/cypress/e2e.ts"
     }
   },
   "dependencies": {
     "@bahmutov/cypress-esbuild-preprocessor": "^2.2.4",
+    "@badeball/cypress-cucumber-preprocessor": "^22.0.1",
+    "@cucumber/gherkin-utils": "^9.2.0",
     "@eslint/js": "^9.23.0",
     "@stylistic/eslint-plugin": "^4.2.0",
     "@stylistic/stylelint-plugin": "^3.1.2",
     "@typescript-eslint/eslint-plugin": "^8.28.0",
     "@typescript-eslint/parser": "^8.28.0",
     "@vue/eslint-config-typescript": "^14.5.0",
+    "cypress-fail-fast": "^7.1.1",
+    "cypress-fail-on-console-error": "^5.1.1",
+    "cypress-terminal-report": "^7.1.0",
     "esbuild": "^0.25.0",
     "escape-string-regexp": "^4.0.0",
     "eslint-import-resolver-alias": "^1.1.2",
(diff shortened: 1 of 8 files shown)


packages/config/package.json
```diff
@@ -35,13 +43,5 @@
     "stylelint-config-html": "^1.1.0",
     "stylelint-config-recommended-scss": "^14.1.0",
     "stylelint-config-recommended-vue": "^1.6.0"
-  },
-  "peerDependencies": {
-    "@badeball/cypress-cucumber-preprocessor": "^22.0.0",
-    "cypress": "^14.0.0",
-    "cypress-fail-fast": "^7.1.1",
-    "cypress-terminal-report": "^7.0.4",
-    "eslint": "^9",
```
schogges, line 44: I think `eslint` would still be a `devDependency` here, as we also want to lint here

johncowen: oh good spot thanks!

---

One thing I noticed separately here, given the assumption we will never move `@kumahq/config` to its own completely separate repository (probably a good assumption, close to a fact). I don't think there will ever be a time when you are using devDependencies without being within our monorepo. Do you think thats a good assumption?

---

Lemme add the eslint here. brb

schogges: > One thing I noticed separately here, given the assumption we will never move `@kumahq/config` to its own completely separate repository (probably a good assumption, close to a fact). I don't think there will ever be a time when you are using devDependencies without being within our monorepo. Do you think thats a good assumption?

yeah, I think that assumption is correct 🙂

johncowen: In https://github.com/kumahq/kuma-gui/pull/3708/commits/9fd7706fee3ec46b1bf705704b1175486d0f3237 I've added the eslint devDependency in multiple places. We know have this duplicated configuration in:

- @kumahq/kuma-gui
- @kumahq/config
- @kumahq/kuma-http-api
- @kumahq/fake-api
- the root of the monorepo

Considering we are aiming to add roughly another 5 to 10 packages in our monorepo, are you still happy to add this duplicated configuration in around 15 different packages? and remember to do so when those packages are made? especially given:

>  I don't think there will ever be a time when you are using devDependencies without being within our monorepo.

i.e. IMO this is a devDependency of the monorepo, personally I don't see reason to duplicate the same configuration 15 times.

Let me know what you think, I'm happy to keep these duplicate in, but as you know I'd prefer to just have one in the root of the monorepo, but I'm easy either way right now.

schogges: Generally I think we should add the dependencies to where they are being used and make the packages self-contained as much as possible. But I see your point and maybe we can define some "global" ones that we want to use in every package (i.e. eslint, typescript, ...).

schogges (approved):

packages/config/package.json
johncowen: We had some more discussion on this one and we are slowly coming to the conclusion that it _might be_ that in order for the package to be added in the root of the workspace, it should _at least_ be a devDependency only.

Still a little up in the air, but looks like this is what we are getting to.

I'm gonna move `eslint` to root only here

packages/fake-api/package.json
```diff
@@ -30,11 +30,7 @@
...
   "devDependencies": {
     "@kumahq/config": "*",
-    "eslint": "^9.22.0",
-    "typescript": "^5.8.2"
-  },
-  "peerDependencies": {
-    "vite": "^6.0.0",
-    "cypress": "^14.0.0"
+    "typescript": "^5.8.2",
+    "vite": "^6.0.0"
```
johncowen, line 34: I guess we will probably end up moving these to root also, but lets wait and see with just `eslint` for now

schogges (approved):

schogges (approved):

Issue #3152, opened 2024-11-06
xds config rendering is very slow


lahabana:
### What happened?

It's almost impossible to use as soon as the config is a little big

johncowen:
I had a quick look and experiment with this, I'm going to keep looking/thinking, but:

1. It's a bit difficult to know what a "big" config is. I would guess it also depends on the machine we are running on. My machine might be fine with a "big" config, but someone else's less so. Commenting this as it makes repros a bit difficult.
2. I don't think this is only about code highlighting, I noticed that when I turn this off, the component we use still renders far far slower than just using a HTML `<pre><code>` to render the code.
3. I looked at the innards of the kongponent we use for the Code block (KCodeBlock), and it looks like there could be some low hanging performance fruit there:
  - I noticed RegExp instantiation in a hot path when using the filter. Pulling this out of the hot-path might give us some more performance
  - KCodeBlock sanitises the code first, in our case seeing as the config and the stats aren't user generated (unless people can put HTML in the Envoy config?) we might not need that sanitising. In any case any sanitising shouldn't be done in a Vue component seeing as thats the view layer/DOM. Moving any sanitising out of the DOM to a data layer always gives us the option of moving this sort of code off the main thread.

Anyway, wanted to make some comments partly for my own benefit, need to think about next steps.

lahabana:
Let's try to see if we can improve the perf of `KCodeBlock`.

johncowen:
Had a little bit more of a dig into this:

1. I went back to 2.7 with a real XDS config (from the demo app), and it does indeed still run slow (i.e. its not something we've added in 2.9)
2. I looked at the nodes produced by the syntax highlighting we have (this is implemented ourselves ontop of KCodeBlock) and it does look like its causing "Too Many Nodes":

![Screenshot 2024-11-14 at 15 59 42](https://github.com/user-attachments/assets/22dd97cc-6b28-4be4-a70b-3391abdea137)

johncowen:
Notes while I'm looking at this:

- There were some improvements to speed made in https://github.com/Kong/kongponents/pull/2562, this PR ended up in `@kong/kongponents@9.17.1` (https://github.com/Kong/kongponents/pull/2562#issuecomment-2597103299)
- Dependabot PR where we upgraded from 9.16.1 to 9.18.4 https://github.com/kumahq/kuma-gui/pull/3460
- Previous commit to the upgrade is 3dc6886
[...]

PR #3711, opened 2025-03-27
fix(codeblock): fix height of codeblock to use DOM recycling scrolling

johncowen:
Also see https://github.com/kumahq/kuma-gui/issues/3152, specifically https://github.com/kumahq/kuma-gui/issues/3152#issuecomment-2758069136

---

Looks like the line numbers we render here can cause a massive slowdown in re-renders, specifically when typing to search large configs.

The upstream KCodeBlock that we use for this does have a virtual scroller for the line numbers built in, but because of the nature of some virtual scrollers, and definitely in this case, you need to fix the height of the container of the thing that contains the virtual scroller. This means that the KCodeBlock now controls the scrolling, not the browser window.

Note: I only implemented this in the "full page" XDS Config pages for now, so if you are viewing the PR preview make sure you are looking at it on this page.

### DIFF
--- a/packages/kuma-gui/src/app/connections/views/ConnectionsXdsConfigView.vue
+++ b/packages/kuma-gui/src/app/connections/views/ConnectionsXdsConfigView.vue
@@ -20,39 +20,48 @@
       <XCard>
         <DataLoader
           :src="uri(sources, `/connections/xds/for/:proxyType/:name/:mesh/:endpoints`, {
-            proxyType: ({ ingresses: 'zone-ingress', egresses: 'zone-egress'})[route.params.proxyType] ?? 'dataplane',
+            proxyType: ({ ingresses: 'zone-ingress', egresses: 'zone-egress' })[route.params.proxyType] ?? 'dataplane',
             name: route.params.proxy,
             mesh: route.params.mesh || '*',
             endpoints: String(route.params.includeEds),
           })"
           v-slot="{ data, refresh }"
         >
-          <XCodeBlock
-            language="json"
-            :code="JSON.stringify(data, null, 2)"
-            is-searchable
-            :query="route.params.codeSearch"
-            :is-filter-mode="route.params.codeFilter"
-            :is-reg-exp-mode="route.params.codeRegExp"
-            @query-change="route.update({ codeSearch: $event })"
-            @filter-mode-change="route.update({ codeFilter: $event })"
-            @reg-exp-mode-change="route.update({ codeRegExp: $event })"
+          <XWindow
+            :resize="true"
+            v-slot="{ resize }"
           >
-            <template #primary-actions>
-              <XCheckbox
-                :checked="route.params.includeEds"
-                label="Include Endpoints"
-                @change="(value) => route.update({ includeEds: value})"
-              />
-              <XAction
-                action="refresh"
-                appearance="primary"
-                @click="refresh"
+            <div
+              ref="$el"
+            >
+              <XCodeBlock
+                :max-height="`${(resize?.target?.innerHeight ?? 0) - ($el?.getBoundingClientRect().top + 146)}`"
+                language="json"
+                :code="JSON.stringify(data, null, 2)"
+                is-searchable
+                :query="route.params.codeSearch"
+                :is-filter-mode="route.params.codeFilter"
+                :is-reg-exp-mode="route.params.codeRegExp"
+                @filter-mode-change="route.update({ codeFilter: $event })"
+                @reg-exp-mode-change="route.update({ codeRegExp: $event })"
               >
-                Refresh
-              </XAction>
-            </template>
-          </XCodeBlock>
+                <template #primary-actions>
+                  <XCheckbox
+                    :checked="route.params.includeEds"
+                    label="Include Endpoints"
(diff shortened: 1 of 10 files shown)


packages/kuma-gui/src/app/connections/views/ConnectionsXdsConfigView.vue
```diff
@@ -20,39 +20,48 @@
...
-                label="Include Endpoints"
-                @change="(value) => route.update({ includeEds: value})"
-              />
-              <XAction
-                action="refresh"
-                appearance="primary"
-                @click="refresh"
+            <div
+              ref="$el"
+            >
```
johncowen, line 36: Bit annoyed about this, the return of meaningless divs.

It seems in order to use a ref to get a reference deep in your component tree (in this case *View.vue > XCodeBlock > XLayout), you have to do a load of `defineExpose` plumbing, which is just going to be brittle and hard to maintain, and at the very least annoying to have to use so much code to get to a deep DOM element.

Instead I just figured I could just wrap the thing in an extra div and get the position of the extra div instead (we need the `.top` of the XCodeBlock to calculate the height it should be)

packages/kuma-gui/src/app/connections/views/ConnectionsXdsConfigView.vue
```diff
@@ -20,39 +20,48 @@
...
+              <XCodeBlock
+                :max-height="`${(resize?.target?.innerHeight ?? 0) - ($el?.getBoundingClientRect().top + 146)}`"
```
johncowen, line 38: Ideally we don't want to hardcode this number, but at the moment it's not obvious how to calculate this exactly, I suspect KCodeBlock's `maxHeight` doesn't take into account its toolbar.

All in all in order to move forwards, I'd be happy with this magical number for the moment at least. Take into account we'll need to do this in three pages, XDS Config, Stats and Clusters.

packages/kuma-gui/src/app/x/components/x-window/XWindow.vue
johncowen, line 1: This file wraps over GlobalEvents (a 3rd party package I added):

1. In case we ever want to move, or roll our own.
2. Maintains our "`x` is for GUI related compoennts"
3. Means we can tailor it to our needs, currently we only support window resize events. If we need more we can add them incrementally as and when we need them here.

johncowen:
Taking this out of draft so its clear we are all happy with the approach, there is still work required before we merge, but for the moment we just need to make sure we are all happy with the general approach

packages/kuma-gui/src/app/connections/views/ConnectionsXdsConfigView.vue
schogges: Would it make sense to define a min-height also?

johncowen: 🤔 I think from a UX perspective that would potentially result in 2 scrollbars (one on the codeblock one on the window), which I reckon we should avoid

Technically, notice we the `max-height` is a "feature" of the underlying KCodeBlock, so if we'd want to add a min-height we'd have to add some custom CSS, which in some cases I'd be fine with but considering the above UX point also, I'd definitely be on the side of not adding extra bits.

schogges: Hmmm, in my case I already see two scrollbars (not sure what is causing this). But I don't think that is bad in general. In case there is more stuff added to that view, we need to have another scrollbar.

image

johncowen: I increased the "magic" number to 200, hopefully you only get one scrollbar now. This is why I'd rather it be calculated, but still I'd be happy with a magic number at least short term. The 2 scrollbars (if they should ever happen) are less bad than a slow search.

schogges: > I increased the "magic" number to 200, hopefully you only get one scrollbar now. This is why I'd rather it be calculated, but still I'd be happy with a magic number at least short term. The 2 scrollbars (if they should ever happen) are less bad than a slow search.

The second scrollbar is gone 🙂 
Yeah that "magic" number is not ideal, but should be fine for now. Unfortunately it's not that easy to just let it grow to fill the rest of the screen, at least from my quick fiddling around 🤔

schogges (approved):

packages/kuma-gui/src/app/connections/views/ConnectionsXdsConfigView.vue
johncowen: > Yeah that "magic" number is not ideal, but should be fine for now. Unfortunately it's not that easy to just let it grow to fill the rest of the screen, at least from my quick fiddling around

Yeah for sure, I couldn't find a calculation that was always correct, I'm happy to go with this for now also. Let me roll it out to the other pages, 🤔 we could make an issue for this maybe, brb

johncowen:
This has been rolled out everywhere we need it now, so just needs another ✅ and we can get this in and close up some issues 🎉

schogges (approved):

PR #3712, opened 2025-03-27
chore(deps): run npm upgrade and npm dedupe

johncowen:
See title, I split in two commits which shows the result of each.


schogges (approved):
🚀

Issue #3655, opened 2025-03-12
Migrate KTable to KTableView


schogges:
### Description

The [KTable](https://kongponents.konghq.com/components/table.html) component is marked as deprecated and newer versions of `kongponents` seem to break the styles.
There are two options for migration:

- [KTableView](https://kongponents.konghq.com/components/table-view.html), which only renders the data
- [KTableData](https://kongponents.konghq.com/components/table-data.html), which has more capabilities like fetching

I think `KTableView` should fit our requirements as we have decoupled the data fetching and other things from the table component already.

johncowen:
Looks like we should close https://github.com/kumahq/kuma-gui/issues/3379 as a dupe?

PR #3713, opened 2025-03-27
chore: migrate from KTable to KTableView

johncowen:
TBD

Closes https://github.com/kumahq/kuma-gui/issues/3655


packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -106,21 +101,14 @@ const resize = (args: TablePreferences) => {
...
-function getRowAttributes(row: Row): Record<string, string> {
+function getRowAttributes(row: Record<string, any>): Record<string, any> {
   if (!row) {
     return {}
   }
 
   const attributes: Record<string, string> = {}
 
-  if (typeof props.isSelectedRow !== 'undefined' && props.isSelectedRow(row)) {
+  if (typeof props.isSelectedRow !== 'undefined' && props.isSelectedRow(row as Row)) {
```
johncowen, line 111: The upgrade to KTableView breaks all our type safety for exporting the data in each slot but keeping the type from the `:items` because KTableView turns these rows into `any`. We know that Row is derived from `:items` and we know that `KTableView:row-attrs` gets called with a single item from `:items`, so even though we are using `as` here to cast to a `Row` we know that things all match up.

packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
```diff
@@ -141,6 +129,18 @@ const click = (e: MouseEvent) => {
...
 }
+const $ref = ref<InstanceType<typeof KTableView> | null>(null)
+const rewrite = () => {
+  const $el = $ref.value?.$el
+  if (
+    ((el: any | HTMLElement): el is HTMLElement => el && typeof el.querySelectorAll === 'function')($el)
+  ) {
+    const $trs = $el.querySelectorAll('tr[tabindex="0"]')
+      ;['tabindex'].forEach(attr => Array.from($trs).forEach(item => item.removeAttribute(attr)))
+  }
```
johncowen, line 140: We remove the `tabindex=0` that KTableView adds to all of our rows, because hitting enter on those rows does absolutely nothing. I had a quick go at making this work (so hitting enter on a row would fire the "action link") but then I could find a way to make it work so that shift+enter would open in a new window. In the end I decided that if we can't replicate the native experience completely I'd rather not have keypress-able rows, you can still tab to the link itself and have all the native features work correctly (link shift+enter etc)

packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
schogges: That's really unfortunate. But I think we can leave it as is for now and try to improve in next iterations. Maybe KTableView brings type-safety back in next versions 🤞 .

packages/kuma-gui/src/app/policies/components/PolicyTypeEntryList.vue
```diff
@@ -156,9 +158,23 @@ function getCellAttributes({ headerKey }: any): Record<string, string> {
...
+  width: 15%;
+}
+
+.cell-config {
+  width: 35%;
+}
+
+.cell-origins {
+  width: 15%;
+}
```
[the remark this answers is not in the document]
johncowen: oh! no idea, ah I think its for the accordion table thing, let me check on these if I can remove them I will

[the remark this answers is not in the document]
johncowen: I found out that this component is only used for legacy policies, and when I checked these rules didn't seem to be affecting the component at all so I removed them

schogges (approved):

PR #3715, opened 2025-03-28
chore(fake-api): lint

schogges:
Adds some common make targets and most importantly `lint` to the new `fake-api` package.


johncowen:
I think this can now be closed following https://github.com/kumahq/kuma-gui/pull/3823

johncowen:
We no longer need this, since we did https://github.com/kumahq/kuma-gui/pull/3823

PR #3721, opened 2025-03-31
chore(deps): unpin kong/shared-entities

johncowen:
See title, I left in some commented out config for future us

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17149,7 +17149,7 @@
       "dependencies": {
         "@gera2ld/tarjs": "^0.3.1",
         "@kong-ui-public/app-layout": "^4.3.16",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -17149,7 +17149,7 @@
       "dependencies": {
         "@gera2ld/tarjs": "^0.3.1",
         "@kong-ui-public/app-layout": "^4.3.16",
-        "@kong-ui-public/entities-shared": "3.21.0",
+        "@kong-ui-public/entities-shared": "^3.21.0",
```
schogges, line 17152: Is it possible to remove `@kong-ui-public/entities-shared` from the root again?

Issue #3605, opened 2025-02-26
add dismissal buttons to our notifications


johncowen:
### Description

Follow up of https://github.com/kumahq/kuma-gui/pull/3566

We need the dismissal actions to not only track when you have dismissed a notification and store that, but also remove that storage once the logic that showed the notification is reversed.

We also want to "garbage collect" the local storage values every-so-often so notifications appear again after say 7 days (you'd have to dismiss again, or actually sort out the issue to prevent it appearing again)

PR #3728, opened 2025-04-01
chore(notifications): normalize notification keys

johncowen:
- Changes all out notifications to use lowercase i18n keys that all live on a `<module-name>.notifications` property.
- Simplifies notifications so that are "calculated" in place instead of some in the data layer and some in the view layer. They are now all in the view layer. This means we can be more intentional about what notifications are shown where. The is a small bug here where we try to show a NACK warning on the zone listing page when we shouldn't.
- Due to the above point, I made some amends to the dataplane data layer
- Noticed a few tiny things in HostnameGenerators data mainly around using the generated types in the mocks rather than our application types.

Note tests are gonna fall on their face here, I wanted to put up a draft PR to share progress, my next steps is fixing up the tests.

Part of https://github.com/kumahq/kuma-gui/issues/3605

I'm considering making a reusable/include-like component here, possibly for the DataplaneListing, still considering it but it shouldn't necessarily block review/merge, we can do a follow up.

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -12,7 +12,7 @@
   >
     <DataSource
       :src="uri(sources, '/connections/stats/for/:proxyType/:name/:mesh/:socketAddress', {
-        proxyType: ({ ingresses: 'zone-ingress', egresses: 'zone-egress'})[route.params.proxyType] ?? 'dataplane',
+        proxyType: ({ ingresses: 'zone-ingress', egresses: 'zone-egress' })[route.params.proxyType] ?? 'dataplane',
         name: route.params.proxy,
         mesh: route.params.mesh || '*',
         socketAddress: props.data.dataplane.networking.inboundAddress,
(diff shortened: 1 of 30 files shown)


packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
```diff
@@ -17,10 +17,10 @@
       >
         <XNotification
           v-if="!can('use state')"
-          uri="global-store-type-memory"
+          uri="main-overview.notifications.store-memory"
         >
           <XI18n
-            path="common.warnings.GLOBAL_STORE_TYPE_MEMORY"
+            path="main-overview.notifications.store-memory"
```
johncowen, line 23: Need to change the root i18n yaml key to `control-planes` at some point

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -49,12 +85,18 @@
...
             :modified="props.data.modificationTime"
           >
-            <DefinitionCard layout="horizontal">
-              <template #title>
+            <DefinitionCard
+              layout="horizontal"
+            >
+              <template
+                #title
+              >
```
schogges, line 93: There is a lot of noise in here that just formats and puts attributes on a new line, is this intended? 🤔

schogges (commented):
One nit, the rest LGTM 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: Exactly, eslint decided to change everything so the PR was messy, so then I tried to put it back how it was, but I probably did too much trying to undo it, so I just gave up and made https://github.com/kumahq/kuma-gui/pull/3731 to avoid all of this shenanigans

schogges: Hmm, ok weird 🤔 let's get this in then and have an eye on it, so maybe I have to revert my standpoint in #3731 🤔

schogges (approved):

PR #3729, opened 2025-04-01
fix: reset z-index of KTable headers to auto

johncowen:
Overwrites kongponents setting KTable headers to `z-index: 2` in https://github.com/Kong/kongponents/pull/2629 which causes the following to happen:

### Before

Screenshot 2025-04-01 at 19 16 52

### After (this PR)

Screenshot 2025-04-01 at 19 16 43

I also filed an issue upstream https://github.com/Kong/kongponents/issues/2678


schogges (approved):
Nice 👍

PR #3731, opened 2025-04-02
chore(lint): adds rule to enforce attributes on their own line always 

johncowen:
Yep linting... 😆 

Whilst working on https://github.com/kumahq/kuma-gui/pull/3728 I noticed that our linting (maybe my IDE) was fiddling with where attributes are. It seems like the rule it is applying is:

- If its a single attribute keep it on the same line as the node
- If its more than one attribute, put them all on separate lines

^ This makes things inconsistent.

---

Moreover on a personal level, I very much prefer line-wise editing, its makes it so much easier to add and remove "settings", and this is probably the best reason for enforcing trailing commas everywhere for example.

All in all, I kinda don't care if the linter didn't mess with these at all, and we just add features, but the linter keeps fiddling with things and adding unrelated changes to PRs, so I decided to PR a rule so it stops doing this.

Happy to hear other opinions if this isn't preferred by others.

### DIFF
--- a/packages/config/src/eslint.cjs
+++ b/packages/config/src/eslint.cjs
@@ -264,6 +264,10 @@ function createEslintConfig(
           ],
           alphabetical: false,
         }],
+        'vue/first-attribute-linebreak': ['error', {
+          'singleline': 'below',
+          'multiline': 'below',
+        }],
         'vue/singleline-html-element-content-newline': ['error', {
           ignoreWhenNoAttributes: true,
           ignoreWhenEmpty: true,
--- a/packages/kuma-gui/src/app/App.vue
+++ b/packages/kuma-gui/src/app/App.vue
@@ -17,7 +17,9 @@
       <ApplicationShell
         class="kuma-application"
       >
-        <template #home>
+        <template
+          #home
+        >
           <img
             class="logo"
             src="@/assets/images/product-logo.png"
@@ -26,7 +28,9 @@
           >
         </template>
 
-        <template #navigation>
+        <template
+          #navigation
+        >
           <AppNavigator
             v-style="'--icon: var(--icon-home)'"
             data-testid="control-planes-navigator"
@@ -67,7 +71,9 @@
           />
         </template>
 
-        <template #bottomNavigation>
+        <template
+          #bottomNavigation
+        >
           <AppNavigator
             v-style="'--icon: var(--icon-configuration)'"
             :active="child.name === 'configuration-view'"
--- a/packages/kuma-gui/src/app/application/components/app-navigator/AppNavigator.vue
+++ b/packages/kuma-gui/src/app/application/components/app-navigator/AppNavigator.vue
@@ -2,7 +2,9 @@
   <li
     class="app-navigator"
   >
-    <slot name="default">
+    <slot
+      name="default"
+    >
       <XAction
         :class="{
           'is-active': props.active,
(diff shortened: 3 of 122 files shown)


schogges:
Honestly I'm not a fan of this 😅 it may makes it easier to edit, but harder to read 🤔 also what would you do about objects with a single entry (i.e. `{ foo: 'bar' }`, function parameters (i.e. `(foo: number)`, ...?

johncowen:
This is only about HTML/Vue component nodes, so:

```vue
<Component single-attribute="" />
but
<Component 
  single-attribute=""
  multiple-attributes=""
/>
```

vs (this PR)

```vue
<Component 
  single-attribute=""
/>
and
<Component 
  single-attribute=""
  multiple-attributes=""
/>
```

schogges:
> This is only about HTML/Vue component nodes, so:
> 
> ```vue
> <Component single-attribute="" />
> but
> <Component 
>   single-attribute=""
>   multiple-attributes=""
> />
> ```
> 
> vs (this PR)
> 
> ```vue
> <Component 
>   single-attribute=""
> />
> and
> <Component 
>   single-attribute=""
>   multiple-attributes=""
> />
> ```

Yeah, but the reasons could also apply to everything else than HTML/Vue. Would you apply this rule to JS/TS aswell?

schogges:
I'd also prefer

```html
<div id="unique" class="foo">
```

over

```html
<div
  id="unique"
  class="foo"
>
```

if it's short enough 🤔

johncowen:
At the moment I think it puts them all on newlines if its more than one.

If you think about it this makes git diffs overly "change-y" also. git diffs are far better if things like this are always linewise (instead of sometimes linewise sometimes not, which is how it is now)

johncowen:
Oh missed this

> Yeah, but the reasons could also apply to everything else but HTML/Vue. Would you apply this rule to JS/TS aswell?

JS/TS isn't making my PRs messy at the minute so unless it started to do that I'd wouldn't mind either way.

schogges:
Looking at other examples like https://vuejs.org/style-guide/rules-strongly-recommended.html#directive-shorthands it seems that the style guide tells that single attributes should be inline 🤔

johncowen:
Ok so we can agree on this from your above link:

> In JavaScript, splitting objects with multiple properties over multiple lines is widely considered a good convention, because it's much easier to read. 

There are two reasons why I think its weird not to do the same with single attributes:

1. Consistency
2. Flip-flopping between having an attribute on a single line and then a multiple line and back again depending on the amount of attributes. Specifically this thing can make git diffs messier than they need to be.

Lastly, I'm good to close this if you aren't convinced, I'm not super precious about it right now, it was just something I bumped into this morning. lmk

johncowen:
> it seems that the style guide tells that single attributes should be inline 🤔

I wouldn't say "tells that single attributes should be inline". This is an example of something totally different saying to consistently use either directive shorthand or always or never.

It also doesn't give examples of using scoped slots to export variables, that we do rather a lot in our codebase.

johncowen:
I also noticed this page https://vuejs.org/guide/components/slots.html#scoped-slots doens't stick to the style guidde of putting multiple attributes on different lines, so I don't think we should use these sorts of examples as "style guides"

Screenshot 2025-04-02 at 11 14 55

johncowen:
Lemme close this one, I'm not that precious over it, but if it comes up again I can always reopen

schogges:
> Lemme close this one, I'm not that precious over it, but if it comes up again I can always reopen

Sorry for the late response. I see from the examples you posted that it's not really consistent in the docs. But we once said that we would like to stick to the recommended and default rules as much as possible.
However, I also understand the reasons you have for adding these exceptions. Therefore I agree, that in the case we hit issues with the recommended/default rules, I am open for the changes. Thanks for bringing this up 🙂

PR #3689, opened 2025-03-18
chore: cleanup some unused global css vars/classes

johncowen:
Removes some global CSS vars and global CSS helper classes we no longer use.

PR #3732, opened 2025-04-02
chore: remove unused CSS and fonts, add fonts we use

johncowen:
Following on from https://github.com/kumahq/kuma-gui/pull/3689 I had another quick review of our global CSS and found a few more things we could remove there.

Interestingly I also found out (from what I could see) that we ship the `Fira Mono` but we don't seem to use it anywhere? Whereas `@kong/design-tokens` tokens use `JetBrainsMono` (`kui-font-family-code`) but we don't ship that even though we say some of our fonts should use it if its available (it has fallbacks but still). Honestly I find it a little weird that `@kong/design-tokens` doesn't ship the fonts it uses, even if they wanted that to be optional, but 🤷 

I removed Fira Mono and added JetBrainsMono from here https://github.com/JetBrains/JetBrainsMono


schogges (approved):
👍 🧹

PR #3733, opened 2025-04-03
feat(x-about): truncate tag-list, list of badges and ports

### LINKED ISSUE #3352: Truncate list of badges in About Sections
### Description

Follow up to #3300 

Next to the adoption of the `AppAboutSection` component, we started to use badges for the information displayed in the About Section card. Some of them show list of badges. In order to improve the layout and reduce the noise in these cards we should consider to truncate these lists.

Some of the lists use the `TagList` component that can be optionally truncated. The default is currently set to `false`. We should consider to set default to `true`.
In other cases we create these lists inline and use `XBadge` or other components (e.g. `KumaPort`) directly. We should consider to align this (if possible) and maybe always use `TagList`.

#### No Truncation
Screenshot 2025-01-06 at 11 14 46
Screenshot 2025-01-06 at 11 14 59

#### Truncation
Screenshot 2025-01-06 at 11 30 34
Screenshot 2025-01-06 at 11 30 44

schogges:
In #3300 we added a new `XAboutCard` component to our detail views. This card can sometimes hold list of badges, tags and more. For better responsiveness we decided to add truncation.

Full width
Screenshot 2025-04-03 at 10 13 57

No truncation
Screenshot 2025-04-03 at 10 14 21

Truncation
Screenshot 2025-04-03 at 10 14 09

---

I've deprecated the `TagList` component, mainly used `XLayout` with `truncate` prop and added information about when to use truncation.

Closes #3352


schogges:
Put it back to draft to amend:

- Use `XLayout` instead of `TagList` (use `truncate` props)
- Deprecate `TagList` component
- Document in `XLayout` to truncate in places where y-axis space is limited (i.e. table rows, about-cards, ...) and don't truncate in all the other places

schogges:
I've entirely removed `TagList` and replaced it with `XLayout` and `XBadges` while using `truncate` only in places where space is limited vertically. Additionally I've updated the docs accordingly.

schogges:
> I've entirely removed `TagList` and replaced it with `XLayout` and `XBadges` while using `truncate` only in places where space is limited vertically. Additionally I've updated the docs accordingly.

Reverted it to stay in the scope of this PR. It's not straightforward to remove `TagList` because of some nested functionality/logic. Therefore this needs to happen separately.

packages/kuma-gui/src/app/services/views/MeshServiceDetailView.vue
```diff
@@ -84,14 +84,16 @@
...
+                <XTruncate>
+                  <KumaPort
+                    v-for="connection in props.data.spec.ports"
+                    :key="connection.port"
+                    :port="{
+                      ...connection,
+                      targetPort: undefined,
+                    }"
+                  />
+                </XTruncate>
```
johncowen, line 96: If you wrap this in an `<XLayout type="separated" truncate>` if will truncate and also fix the ever so tiny spacing discrepancy (same with MMZ).

I 'think' then we don't need a `XTruncate` component?

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
```diff
@@ -54,15 +54,17 @@
...
+                    <XBadge
+                      v-for="([label, value], index) in Object.entries(labels)"
+                      :key="`${label}${value}${index}`"
+                      appearance="neutral"
                     >
-                      <XBadge>
-                        {{ key }}:{{ value }}
-                      </XBadge>
-                    </template>
+                      <span class="label">{{ label }}:</span>{{ value }}
```
johncowen, line 66: How come these are now styled differently than previously? But not like any other XBadges in the PR?

schogges: I think the `label:value` badges are styled differently in some places. That is a left-over from the attempt to align them. Not in scope of this PR, so I remove it.

johncowen: Gotcha, I took a quick look and noticed one place elsewhere in a listing where we do key:**value** I guess its Very Carefully Placed 😅 to be just in that place 🤷 😆 

I 'think' we just need to remove this formatting where we do it in that one place there. Unless I missed somewhere else we do this.

johncowen (approved):
Looks good thanks for adding the doc 👍

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
schogges: Yeah, looks like there was reason to add it there 🙂 Could do a little research again and potentially align it in a follow-up.

PR #3734, opened 2025-04-03
chore(ci): reduce the majority of fake-api to a createFetch function

johncowen:
Refactors some more of fake-api, mostly:

- Reduce everything down to a reusable `createFetch`/`createFetchSync`, that returns something that is very fetch-like (i.e. very recognisable). Note: `createFetchSync` is needed for Cypress which is notorious weird with async things.
- I moved the YAML/JSON merging we do for overwriting responses during tests out of fake-api entirely. If you aren't doing tests like we are then you wouldn't need this in order to build basic fake-apis.

Probably a few other little bits. Unfortunately the PR is a little large, if it was me I'd look at/compare this locally to get a better feel for the improvement here.

We are still not quite done here, but with what we did before extracting things out into its own module, and what is here, I'm far happier with this code now.

### DIFF
--- a/packages/fake-api/src/cypress.ts
+++ b/packages/fake-api/src/cypress.ts
@@ -8,7 +8,7 @@ export const mocker = <T extends object = {}>(
   dependencies: Dependencies<T>,
   fs: FS,
 ): Mocker => {
-  return (path, opts = {}, cb = noop) => {
+  return (path, opts = {}, middleware = noop) => {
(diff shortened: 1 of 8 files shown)


schogges (dismissed):
LGTM 👍

johncowen:
@schogges sorry there was a rebase necessary here that dismissed your review, will need another ✅

PR #3737, opened 2025-04-07
feat: implement x-filter-bar component

### LINKED ISSUE #3707: Create XFilterBar component
### Description

As a first step we have to refactor our current `FilterBar` component. Ideally we define a new `XFilterBar` component.

Requirements:
- Leverages [KInput](https://kongponents.konghq.com/components/input.html)
  - Specifically [before slot](https://kongponents.konghq.com/components/input.html#before-and-after) for the icon 
- Each `tag:value` is highlighted as depicted in designs (see umbrella issue)
- The full input value is still one string, which means that as a user I can click in a `tag:value` and modify/delete what I need to change without loosing the highlighted `tag:value` pair
- On submit the full input value is parsed and returned via callback/event-handler. We can still include the raw input value. If possible we have type support and auto-completion based on the provided fields (props).
- The implementation of the dropdown is **not** in the scope of this issue and will be implemented subsequently

schogges:
As part of the improvement for the filtering on lists, collections and tables I have implemented a new filter bar component `XFilterBar`, which deprecates the old `FilterBar` component.
Initially I hoped that it's possible to leverage our `XInput` component, but unfortunately the way how the underlying `KInput` component does not really allow having a `cover`. At least I didn't find an easy way to achieve that. So I decided to write it from scratch again but use the design tokens.

This is the first part and does not include the dropdown.

Closes #3707

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -24,6 +24,7 @@
       />
       <XCard>
         <search>
+          <XLayout style="display:flex; justify-content:space-evenly; align-items: space-evenly">
           <FilterBar
             class="data-plane-proxy-filter"
             :placeholder="`service:backend`"
@@ -41,8 +42,13 @@
             })"
           />
           <XFilterBar
-            @change="(e) => console.log(e)"
+            :defaultValue="route.params.s"
+            @submit="({ raw, ...rest }) => route.update({
+              page: 1,
+              s: Object.entries(rest).map((v) => v.join(':')).join(' '),
+            })"
           />
+        </XLayout>
 
           <XSelect
             label="Type"
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -24,7 +24,6 @@
       />
       <XCard>
         <search>
-          <XLayout style="display:flex; justify-content:space-evenly; align-items: space-evenly">
           <FilterBar
             class="data-plane-proxy-filter"
             :placeholder="`service:backend`"
@@ -42,13 +41,12 @@
             })"
           />
           <XFilterBar
-            :defaultValue="route.params.s"
+            :default-value="route.params.s"
             @submit="({ raw, ...rest }) => route.update({
               page: 1,
               s: Object.entries(rest).map((v) => v.join(':')).join(' '),
             })"
           />
-        </XLayout>
 
           <XSelect
             label="Type"
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -40,6 +40,9 @@
               ...Object.fromEntries(e.entries()) as Record<string, string | undefined>,
             })"
           />
+          <XFilterBar
+            @change="(e) => console.log(e)"
+          />
 
           <XSelect
             label="Type"
@@ -312,6 +315,7 @@ import FilterBar from '@/app/common/filter-bar/FilterBar.vue'
 import StatusBadge from '@/app/common/StatusBadge.vue'
 import SummaryView from '@/app/common/SummaryView.vue'
 import type { Mesh } from '@/app/meshes/data'
+import XFilterBar from '@/app/x/components/x-filter-bar/XFilterBar.vue'
 const props = defineProps<{
   mesh: Mesh
 }>()
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -24,24 +24,10 @@
       />
       <XCard>
         <search>
-          <FilterBar
-            class="data-plane-proxy-filter"
-            :placeholder="`service:backend`"
-            :query="route.params.s"
-            :fields="{
-              name: { description: 'filter by name or parts of a name' },
-              protocol: { description: 'filter by “kuma.io/protocol” value' },
-              service: { description: 'filter by “kuma.io/service” value' },
-              tag: { description: 'filter by tags (e.g. “tag: version:2”)' },
-              ...(can('use zones') && { zone: { description: 'filter by “kuma.io/zone” value' } }),
-            }"
-            @change="(e) => route.update({
-              page: 1,
-              ...Object.fromEntries(e.entries()) as Record<string, string | undefined>,
-            })"
-          />
           <XFilterBar
+            class="data-plane-proxy-filter"
             :default-value="route.params.s"
+            placeholder="Filter by name, protocol, service or tag..."
             @submit="({ raw, ...rest }) => route.update({
               page: 1,
               s: Object.entries(rest).map((v) => v.join(':')).join(' '),
@@ -315,7 +301,6 @@ import { KUI_ICON_SIZE_40 } from '@kong/design-tokens'
 
 import { sources } from '../sources'
 import AppCollection from '@/app/application/components/app-collection/AppCollection.vue'
-import FilterBar from '@/app/common/filter-bar/FilterBar.vue'
 import StatusBadge from '@/app/common/StatusBadge.vue'
 import SummaryView from '@/app/common/SummaryView.vue'
 import type { Mesh } from '@/app/meshes/data'
--- a/packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
+++ b/packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
@@ -93,6 +93,13 @@
                 ...Object.fromEntries(e.entries()) as Record<string, string | undefined>,
               })"
             />
+            <XFilterBar
+              :defaultValue="route.params.s"
+              @submit="({ raw, ...rest }) => route.update({
+                page: 1,
+                s: Object.entries(rest).map((v) => v.join(':')).join(' '),
+              })"
+            />
           </search>
           <DataLoader
             :src="uri(dataplaneSources, `/meshes/:mesh/dataplanes/for/service-insight/:service`, {
@@ -281,6 +288,7 @@ import ResourceStatus from '@/app/common/ResourceStatus.vue'
 import StatusBadge from '@/app/common/StatusBadge.vue'
 import SummaryView from '@/app/common/SummaryView.vue'
 import { sources as dataplaneSources } from '@/app/data-planes/sources'
+import XFilterBar from '@/app/x/components/x-filter-bar/XFilterBar.vue'
 </script>
 
 <style lang="scss" scoped>
--- a/packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
+++ b/packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
@@ -94,7 +94,7 @@
               })"
             />
             <XFilterBar
-              :defaultValue="route.params.s"
+              :default-value="route.params.s"
               @submit="({ raw, ...rest }) => route.update({
                 page: 1,
                 s: Object.entries(rest).map((v) => v.join(':')).join(' '),
--- a/packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
+++ b/packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
@@ -78,23 +78,10 @@
           </template>
 
           <search>
-            <FilterBar
-              class="data-plane-proxy-filter"
-              :placeholder="`name:dataplane-name`"
-              :query="route.params.s"
-              :fields="{
-                name: { description: 'filter by name or parts of a name' },
-                protocol: { description: 'filter by “kuma.io/protocol” value' },
-                tag: { description: 'filter by tags (e.g. “tag: version:2”)' },
-                ...(can('use zones') && { zone: { description: 'filter by “kuma.io/zone” value' } }),
-              }"
-              @change="(e) => route.update({
-                page: 1,
-                ...Object.fromEntries(e.entries()) as Record<string, string | undefined>,
-              })"
-            />
             <XFilterBar
+              class="data-plane-proxy-filter"
               :default-value="route.params.s"
+              placeholder="Filter by name, protocol or tag..."
               @submit="({ raw, ...rest }) => route.update({
                 page: 1,
                 s: Object.entries(rest).map((v) => v.join(':')).join(' '),
@@ -283,7 +270,6 @@
 import { sources } from '../sources'
 import AppCollection from '@/app/application/components/app-collection/AppCollection.vue'
 import DefinitionCard from '@/app/common/DefinitionCard.vue'
-import FilterBar from '@/app/common/filter-bar/FilterBar.vue'
 import ResourceStatus from '@/app/common/ResourceStatus.vue'
 import StatusBadge from '@/app/common/StatusBadge.vue'
 import SummaryView from '@/app/common/SummaryView.vue'
--- a/packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
+++ b/packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
@@ -1,13 +1,41 @@
 <template>
-  <div class="filter-bar" :style="`--width:${width}px`">
+  <div
+    v-style="`--width:${width}px`"
+    class="filter-bar"
+    @click.stop="inputRef?.focus()"
+  >
+    <div class="icon-wrapper">
+      <XIcon
+        class="icon"
+        name="filter"
+      />
+    </div>
     <div class="container">
       <div class="content-wrapper">
-        <div class="content"><div ref="contentRef" v-html="inputValue.replace(/([\w-\.\/]+:[\w-\.\/]+)/gi, `<span class='highlight'>$1</span>`)"></div></div>
+        <div class="content">
+          <template
+            v-for="(chunk, index) in inputValue.split(/([\w-\.\/]+:[\w-\.\/]+)/gi).filter(Boolean)"
+            :key="chunk+index"
+          >
+            <span :class="{ highlight: /([\w-\.\/]+:[\w-\.\/]+)/gi.test(chunk) }">{{ chunk }}</span>
+          </template>
+        </div>
       </div>
       <div class="wrapper">
-        <div ref="sizerRef" class="sizer"><span>{{ inputValue }}</span></div>
-        <form @submit.prevent="submit">  
-          <XInput :value="props.defaultValue" :placeholder="props.placeholder" appearance="filter" @input="onChange" />
+        <div
+          ref="sizerRef"
+          class="sizer"
+        >
+          <span>{{ inputValue }}</span>
+        </div>
+        <form @submit.prevent="submit">
+          <input
+            ref="inputRef"
+            type="text"
+            :defaultValue="props.defaultValue"
+            :placeholder="props.placeholder"
+            @input="onChange"
+          >
         </form>
       </div>
     </div>
@@ -15,66 +43,97 @@
 </template>
 
 <script setup lang="ts">
-import { ref, watch } from 'vue'
-
-// TODO: dropdown: on submit, the dropdown should close, but the input field should stay focused. when starting to type again, the dropdown should open again/11
+import { ref } from 'vue'
 const props = withDefaults(defineProps<{
   placeholder?: string
   defaultValue?: string
 }>(), {
   placeholder: undefined,
-  defaultValue: ""
+  defaultValue: '',
 })
 
-const inputValue = ref<string>(props.defaultValue);
-const width = ref<number | undefined>();
+const inputValue = ref<string>(props.defaultValue)
+const width = ref<number | undefined>()
 const sizerRef = ref<null | HTMLElement>(null)
-const contentRef = ref<null | HTMLElement>(null)
+const inputRef = ref<null | HTMLInputElement>(null)
 
 const emit = defineEmits<{
   (e: 'submit', value: Record<string, string> & { raw: string }): void
 }>()
 
-const onChange = (event: string) => {
-  if(sizerRef.value) {
-    const sizerWidth = sizerRef.value?.getBoundingClientRect().width
-    const contentWidth = contentRef.value?.getBoundingClientRect().width
-    width.value = contentWidth;
-    // width.value = sizerWidth;
-    console.log("🚀 ~ onChange ~ sizerWidth:", sizerWidth)
-    console.log("🚀 ~ onChange ~ contentWidth:", contentWidth)
-    // console.log(inputValue.value.split(/([a-z0-9]*:[a-z0-9]+)/gi))
-  }
-  inputValue.value = event
+const onChange = (event: Event): void => {
+  const value = (event.target as HTMLInputElement)?.value
+  const sizerWidth = sizerRef.value?.getBoundingClientRect().width
+  width.value = sizerWidth
+  inputValue.value = value
 }
 
 const submit = () => {
-  const values = Object.fromEntries(inputValue.value.split(" ").map((v) => {
-    const [key, value] = v.split(":")
+  const values = Object.fromEntries(inputValue.value.split(' ').map((v) => {
+    const [key, value] = v.split(':')
     if(!key || !value) return []
     return [key, value]
   }).filter((v) => v.length))
   emit('submit', { raw: inputValue.value, ...values })
 }
-
-watch(() => props.defaultValue, () => {
-  inputValue.value = props.defaultValue
-})
-
 </script>
 
 <style scoped lang="scss">
-* {
-  box-sizing: border-box;
-}
-
 .filter-bar {
   --word-spacing: $kui-space-40;
-
-  display: flex;
+  position: relative;
   width: 100%;
+  display: inline-flex;
+  vertical-align: middle;
+  cursor: text;
+  outline: none;
+  align-items: center;
   font-family: $kui-font-family-code;
+  font-size: $kui-font-size-30;
   word-spacing: var(--word-spacing);
+  border-radius: $kui-border-radius-20;
+  box-shadow: $kui-shadow-border;
+  transition: box-shadow $kui-animation-duration-20 ease-in-out;
+  padding: $kui-space-40 $kui-space-50 $kui-space-40 $kui-space-100;
+  color: $kui-color-text;
+
+  &:hover {
+    box-shadow: $kui-shadow-border-primary-weak;
+  }
+  &:focus-within {
+    box-shadow: $kui-shadow-border-primary, $kui-shadow-focus;
+  }
+}
+
+.icon-wrapper {
+  position: absolute;
+  left: $kui-space-50;
+  color: $kui-color-background-primary;
+}
+
+:deep(.icon) {
+  height: $kui-icon-size-40 !important;
+  width: $kui-icon-size-40 !important;
+}
+
+.container {
+  position: relative;
+  display: flex;
+  overflow-x: auto;
+  overflow-y: hidden;
+  align-items: center;
+  scrollbar-width: none;
+  flex: 1;
+  align-self: stretch;
+}
+
+.content-wrapper {
+  position: absolute;
+  display: inline-flex;
+  padding: 0;
+  word-break: break-word;
+  white-space: pre;
+  flex: 1;
 }
 
 :deep(.highlight) {
@@ -85,84 +144,34 @@ watch(() => props.defaultValue, () => {
   word-spacing: var(--word-spacing);
 }
 
-.container {
-  position: relative;
-  display: flex;
-  width: 100%;
-  flex-flow: column nowrap;
-  flex-basis: 0%;
-  flex-grow: 1;
-  flex-shrink: 1;
-  align-self: stretch;
-  flex: 1;
-  width: 100%;
-}
-
-.content-wrapper {
-  position: absolute;
-  display: flex;
-  flex-flow: column nowrap;
-  justify-content: center;
-  left: $kui-space-100;
-  right: $kui-space-50;
-  overflow: hidden;
-  height: 100%;
-}
-
-.content {
-  font-size: $kui-font-size-30;
-  // position: sticky;
-  // right: 0;
-  overflow-wrap: break-word;
-  position: absolute;
-  text-wrap-mode: nowrap;
-  unicode-bidi: isolate;
-  user-select: none;
-  white-space-collapse: preserve;
-  word-break: break-word;
-  min-width: 100%;
-  color: $kui-color-text;
-}
-
 .wrapper {
-  position: relative;
-  display: flex;
   width: 100%;
-  flex-flow: column nowrap;
+  align-self: stretch;
 }
 
 .sizer {
   position: absolute;
+  top: 0;
+  left: 0;
+  height: 0;
+  overflow: scroll;
+  white-space: pre;
   visibility: hidden;
-  height: 5px;
-  font-size: $kui-font-size-30;
 }
 
-:deep(.k-input) {
-  max-width: 100%;
-}
-
-:deep(.input-element-wrapper) {
-  overflow-x: scroll;
-  overflow-y: hidden;
-  scrollbar-width: none;
-  width: 100%;
-  resize: none;
-  position: relative;
-}
-
-:deep(.input) {
-  width: var(--width, 'max-content');
-  min-width: 100%;
-  max-width: 100%;
+input {
+  width: var(--width, "max-content");
   position: relative;
   display: flex;
-  background: transparent;
-  color: transparent;
-  font-size: $kui-font-size-30;
-  caret-color: $kui-color-text;
-  font-family: $kui-font-family-code;
-  word-spacing: var(--word-spacing);
+  min-width: 100%;
+  padding: 0;
+  overflow-x: auto;
+  overflow-y: hidden;
   resize: none;
+  color: transparent;
+  background: transparent;
+  border: 0;
+  outline: none;
+  caret-color: $kui-color-text;
 }
 </style>
--- a/packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
+++ b/packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
@@ -2,26 +2,37 @@
   <div class="filter-bar" :style="`--width:${width}px`">
     <div class="container">
       <div class="content-wrapper">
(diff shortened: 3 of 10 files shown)


packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,182 @@
...
+      />
+    </div>
+    <div class="container">
+      <div class="content-wrapper">
+        <div class="content">
+          <template
+            v-for="(chunk, index) in inputValue.split(/([\w-\.\/]+:[\w-\.\/]+)/gi).filter(Boolean)"
+            :key="chunk+index"
+          >
+            <span :class="{ highlight: /([\w-\.\/]+:[\w-\.\/]+)/gi.test(chunk) }">{{ chunk }}</span>
```
schogges, line 20: The regex will check for `key:value` pairs to highlight, i.e. `foo:bar`, `foo-bar:baz`, ...

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,182 @@
...
+  display: inline-flex;
+  padding: 0;
+  word-break: break-word;
+  white-space: pre;
+  flex: 1;
+}
+
+:deep(.highlight) {
+  // TODO: there is currently no token for this color
+  background: #f0f4f7;
```
schogges, line 145: I couldn't find a design tokens that serves this color (taken from design)

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,182 @@
...
+  inputValue.value = value
+}
+
+const submit = () => {
+  const values = Object.fromEntries(inputValue.value.split(' ').map((v) => {
+    const [key, value] = v.split(':')
+    if(!key || !value) return []
+    return [key, value]
+  }).filter((v) => v.length))
+  emit('submit', { raw: inputValue.value, ...values })
```
schogges, line 77: On submit it will emit the `raw` value from the input field and sanitized `key:value` pairs that can be used for the query, i.e. `{ raw: 'foo:bar abc def bar:baz fyi', foo: 'bar', bar: 'baz' }`.

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,192 @@
...
+  inputValue.value = value
+}
+
+onMounted(() => {
+  const observer = new ResizeObserver(([e]) => {
+    width.value = e?.contentRect?.width
+    containerRef.value?.scrollBy(width.value, 0)
+  })
+  observer.observe(sizerRef.value as HTMLElement)
+})
```
schogges, line 81: Makes sure that we always see the cursor, even when copy&paste

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,201 @@
+<template>
+  <form
+    class="filter-bar-form"
+    @submit.prevent="submit"
+  >
```
johncowen, line 5: Ideally we should try and move the `<form>` into the `*View.vue` template

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,201 @@
...
+    <div
+      v-style="`--width:${width}px`"
+      class="filter-bar-container"
+      data-testid="filter-bar"
+      @click.stop="inputRef?.focus()"
+    >
+      <div class="icon-wrapper">
+        <XIcon
+          class="icon"
+          name="filter"
```
johncowen, line 15: I think we should just use the `search` icon. Can you think of a reason why this is a filter bar and not a search bar? I think its ended up with this naming for historical reasons. Anyway, whatever we call it I don't think we need another `variant` of icon for it.

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,201 @@
...
+      <div
+        ref="containerRef"
+        class="input-container"
+      >
+        <div
+          ref="contentRef"
+          class="content-wrapper"
+        >
+          <template
+            v-for="(chunk, index) in inputValue.split(new RegExp(regex)).filter(Boolean)"
```
johncowen, line 27: Do we need the RegExp wrapper here?

packages/kuma-gui/src/app/services/views/ServiceDetailView.vue
```diff
@@ -78,19 +78,13 @@
...
-                tag: { description: 'filter by tags (e.g. “tag: version:2”)' },
-                ...(can('use zones') && { zone: { description: 'filter by “kuma.io/zone” value' } }),
-              }"
-              @change="(e) => route.update({
+              :default-value="route.params.s"
+              placeholder="Filter by name, protocol or tag..."
+              @submit="({ raw, ...rest }) => route.update({
                 page: 1,
-                ...Object.fromEntries(e.entries()) as Record<string, string | undefined>,
+                s: Object.entries(rest).map((v) => v.join(':')).join(' '),
```
johncowen, line 87: Is this just `raw`?

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,201 @@
...
+        </div>
+        <div class="input-wrapper">
+          <input
+            ref="inputRef"
+            type="text"
+            :defaultValue="props.defaultValue"
+            :placeholder="props.placeholder"
+            data-testid="filter-bar-filter-input"
+            @input="onChange"
+          >
```
johncowen, line 41: Yeah it is a shame that KInput is troublesome here 😭  what is it exactly?

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
```diff
@@ -0,0 +1,201 @@
...
+
+onMounted(() => {
+  const observer = new ResizeObserver(([e]) => {
+    width.value = e?.contentRect?.width
+
+    // keep the cursor position in the view
+    containerRef.value?.scrollBy(inputRef.value?.scrollLeft ?? 0, 0)
+  })
+  observer.observe(contentRef.value as HTMLElement)
+})
```
johncowen, line 93: Is this to keep the input and the highlight content div the same width or something?

johncowen (commented):
This is a super cool approach to doing this! I really like it.

I added a few initial comments, which are mostly nits/non-blockers, but I think I'll ping you so you can take me through

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
schogges: Ok, that would mean that we have to emit the changes on every keystroke. Do you think this is still fine?

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
schogges: Yeah, makes sense 👍

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
schogges: Since I'm using the regex in multiple places I moved it out of the template. For some reason it didn't behave correctly when not wrapped in the `new RegExp` 🤔

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
schogges: The main reason was because in `KInput` the actual `input` element is the wrapper of everything. It has a padding to move the cursor and leave some space for the before and after content. Since I needed to tweak the behaviour of the input element and let it grow the same as the content overlay, this caused a lot of issues and I had to roll back and re-implement.

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
schogges: Yes, this keeps both input and overlay at the same width and also at the same scroll position. The overlay would not scroll automatically on a keystroke.

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
johncowen: Native HTML input@onchange fires on an enter keypress or loss of focus I think (as opposed to input@oninput which I think is also any keypress).

I think something similar to this would be ideal:

https://github.com/kumahq/kuma-gui/blob/61a4de154229b7a9ccbc62cbecf6f0fe1a453751/packages/kuma-gui/src/app/policies/views/PolicyListView.vue#L66-L78

packages/kuma-gui/src/app/x/components/x-filter-bar/XFilterBar.vue
johncowen: Gotcha, thanks for the explanation 👍

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -24,22 +24,17 @@
...
-          />
+          <form
+            class="search-form"
+            @submit.prevent="(e) => route.update({ page: 1, ...onSearch(e) })"
+          >
+            <XSearch
+              class="search-field"
+              name="s"
+              placeholder="Filter by name, protocol, service or tag..."
+              :default-value="route.params.s"
```
johncowen, line 35: We spoke briefly about `default-value` when we synced, and after having a little time to think about it in more detail, I definitely feel like we shouldn't use it. My reasons are:

> The defaultValue property of the [HTMLInputElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) interface indicates the original (or default) value of the [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) element. It reflects the element's [value](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#value) attribute.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/defaultValue

According to the above/my understanding is, the `defaultValue` property is supposed to be the "original value" of the input as rendered to the HTML page before any user interaction of the content/the input field. In our case the `defaultValue` should always be `""` as this is the value of the input field when we load the HTML page. If we wanted to automatically fill the input field with a default value for the user on page load (which we don't) then the `defaultValue` would reflect this value even if the user then edits this content.

By changing the `defaultValue` every time the `?s=` changes (which itself is changed every time the user types something), we are overwriting the defaultValue (`""`) with what the user has typed, essentially making it the same as `.value`. Importantly, if we wanted to know the value of the input field before the user interacted with it, we couldn't do that with native features by inspecting `defaultValue`.

Theres probably a little more I could say here on our own custom component interfaces but I don't think its that relevant if we agree on the above reasoning. But let me know what you think. (in the meantime I'm still reviewing pieces of this following your updates form our chat yesterday 👍 )

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -0,0 +1,181 @@
...
+    <div
+      ref="containerRef"
+      class="input-container"
+    >
+      <div
+        ref="contentRef"
+        class="content-wrapper"
+      >
+        <template
+          v-for="(chunk, index) in inputValue.split(new RegExp(regex)).filter(Boolean)"
```
johncowen, line 23: Could you help me understand exactly what this line is doing?

```javascript
inputValue.split(new RegExp(/([^:\s]+:[^\s]+)/gi)).filter(Boolean)
```

I played with it myself a little yesterday and I couldn't quite get my head around it, even after experimenting a little

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
schogges: The `defaultValue` is only set initially on page load and when the `?s=` changes, which only happens when the user submits the form (not on every keystroke). If we would think that a form submit "refreshes" the page with new content, then again this value is the default. If the `?s=` would change on every keystroke I'd say the value of the input field is controlled by the state in the url query search param, which then is the difference between a controlled and uncontrolled component.
And this to me is the main difference here: we are not controlling the value of the input, just giving it a initial/default value when mounting/loading the page based on the provided query search param.
Keeping this name, means it is aligned with how it's being used on the actual input field.

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: It splits a value of type string by a certain pattern (regex) and filters the result from empty strings to prevent unnecessary iterations on further use. The pattern/regex I'll explain in separate steps:
1. `[^:\s]+`: matches when non-whitespace and no colons and at least one occurrence is found
2. `:` followed by a colon
3. `[^\s]+`: same as 1. but allows `:` to match urls on that side
4. `(...)`: the surrounding braces are necessary to keep the match after the split, because splitting the string by a pattern would mean that it's not part of the result. The braces keep it in the result.

Maybe the following example helps:

```typescript
const result = 'a'.split(/a/) // ['', '']
```
```typescript
const result = 'a'.split(/(a)/) // ['', 'a', '']
```

schogges:
Note: This PR still contains inline text in the template which I'm going to move to our i18n layer in the follow-up when adding the dropdown.

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
johncowen: > The defaultValue is only set initially on page load and not when the ?s= changes

```javascript
:default-value="route.params.s"
```

Maybe I misunderstood, but doesn't this ^ mean the default-value is being updated whenever `?s=` changes?

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
johncowen: Ok thanks, 🤔  did you manage to figure out why we need to re-wrap the regex literal with a further RegExp constructor?

We briefly mentioned yesterday that I should be able to "just type" into the search field and assume that I am going to search on the "name of the thing" (or better said the identifier). So if I type "johns-service" in the field that text should be highlighted, and I should find the service with the name "johns-service". I vaguely remember this was one of the requirements highlighted when the original `FilterBar` was built, so we should try and continue that here.

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
schogges: Right, but it has no effect because the value of the internal `input` will not be updated. My vue knowledge is still limited, so maybe I missed a way to pass a computed value only on mount? 🤔 But even if not, `default-value` (regardless of the `:` in front) indicates that the value which is being passed is only the initial value.
To verify my assumption, I've done the following:

```vue
  <XSearch
    class="search-field"
    name="s"
    :default-value="route.params.s.split('').join(',')" // <== note this
    :keys="['name', 'protocol', 'service', 'tag']"
  />
```

I started on an empty input field, typed `foo bar` and hit enter. The value in the input didn't change even though the query search param `?s=` was added. I refreshed the page with the `?s=` still being set and the default value of my input changed to `f,o,o, ,b,a,r`.

johncowen: > but doesn't this ^ mean the default-value is being updated whenever ?s= changes?

>  indicates that the value which is being passed is only the initial value
> I started on an empty input field, typed foo bar and hit enter. The value in the input didn't change even though the query search param ?s= was added

Are you saying the `.defaultValue` (the initial value) has changed but the `.value` hasn't?

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: > did you manage to figure out why we need to re-wrap the regex literal with a further RegExp constructor?

I will have a look at this now 🙂 

> We briefly mentioned yesterday that I should be able to "just type" into the search field and assume that I am going to search on the "name of the thing" (or better said the identifier). So if I type "johns-service" in the field that text should be highlighted, and I should find the service with the name "johns-service". I vaguely remember this was one of the requirements highlighted when the original FilterBar was built, so we should try and continue that here.

This is still possible, because I didn't change the way how the search is being interpreted. It's just not highlighted, but this is how the new design looks like. I'll bring it up in the sync 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
schogges: > > but doesn't this ^ mean the default-value is being updated whenever ?s= changes?
> 
> > indicates that the value which is being passed is only the initial value
> > I started on an empty input field, typed foo bar and hit enter. The value in the input didn't change even though the query search param ?s= was added
> 
> Are you saying the `.defaultValue` (the initial value) has changed but the `.value` hasn't?

Essentially yes, the prop will be updated but not the value. It is the exact same behaviour compared to

```vue
<input :default-value="computedValue"
```

johncowen: Ok so just to be clear, this is re-setting the "original value" (i.e. the `defaultValue=""`) of the input field every time the `?s=` changes?

schogges: Yes, it does.

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: > did you manage to figure out why we need to re-wrap the regex literal with a further RegExp constructor?

I found the problem:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

> JavaScript [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) objects are stateful when they have the [global](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/global) or [sticky](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky) flags set (e.g., /foo/g or /foo/y). They store a [lastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex) from the previous match. Using this internally, test() can be used to iterate over multiple matches in a string of text (with capture groups).

After I've removed the global flag, it works as expected and I don't even need the global flag.

johncowen: aaaaah wow, good sleuthing! 👍

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -24,22 +24,17 @@
...
+            class="search-form"
+            @submit.prevent="(e) => route.update({ page: 1, ...onSearch(e) })"
+          >
+            <XSearch
+              class="search-field"
+              name="s"
+              placeholder="Filter by name, protocol, service or tag..."
+              :value="route.params.s"
+            />
+          </form>
```
johncowen, line 37: The XSelect below should be part of the `form` no? (i.e. inside the form) This is half the reason for keeping forms outside of the components.

schogges: It could be added, but does not really make sense because the form has no submit button. Unless we add one there would be no way to submit the selection. The select submits `onChange` without a form needed.

johncowen: But semantically its all part of the same form no? Why is XSearch wrapped by a form and the XSelect isn't?

schogges: Yeah, that's right 🙂 I'll add it 👍

schogges:
Back to draft to cover some more things we discussed during our sync 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
schogges: Changed it to `:value` 🙂

schogges:
@johncowen I think this is good for now, the rest will be covered in #3748 and potential follow-ups (i.e. syntax validation)

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -0,0 +1,181 @@
+<template>
+  <div
+    v-style="`--width:${width}px`"
+    class="container"
+    data-testid="filter-bar"
+    @click.stop="inputRef?.focus()"
```
johncowen, line 6: nit: this doesn't have to happen here, but focus is one of those things that should only happen in *View.vue components, otherwise you run the risk of two resuable components fighting for focus. If its in the `*View.vue` its a lot easier to control.

We don't have this issue rn anywhere, but something to keep in mind.

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -0,0 +1,181 @@
...
+
+<script setup lang="ts">
+import { onMounted, ref } from 'vue'
+
+const props = withDefaults(defineProps<{
+  placeholder?: string
+  value?: string
+  name?: string
+}>(), {
+  placeholder: undefined,
```
johncowen, line 52: Tiny nit, I think an empty string would be fine here? If so, I'd prefer a string rather than a `string | undefined`

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -0,0 +1,181 @@
...
+  name: undefined,
+  value: '',
+})
+
+const regex = /([^\s]+)/i
```
johncowen, line 57: nit: does this need to be case insensitive?

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -0,0 +1,181 @@
...
+  width: $kui-icon-size-40 !important;
+}
+
+.input-container {
+  position: relative;
+  display: flex;
+  overflow-x: auto;
+  overflow-y: hidden;
+  align-items: center;
+  scrollbar-width: none;
```
johncowen, line 126: Is this required?

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -0,0 +1,181 @@
...
+  overflow-x: auto;
+  overflow-y: hidden;
+  resize: none;
+  color: transparent;
+  background: transparent;
+  border: 0;
+  outline: none;
+  caret-color: $kui-color-text;
+  line-height: $kui-font-size-70;
+  direction: ltr;
```
johncowen, line 179: Is this required?

johncowen (commented):
Few tinsy things, and one more important thing.

Thanks for the work here, I think its looking great, and I think we are pretty much there 👍

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -24,48 +24,49 @@
...
-            })"
-          />
-
-          <XSelect
-            label="Type"
-            :selected="route.params.dataplaneType"
-            @change="(value: string) => route.update({ page: 1, dataplaneType: value })"
+          <form
+            class="search-form"
+            @submit.prevent="(e) => route.update({ page: 1, ...onSearch(e) })"
```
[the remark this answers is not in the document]
schogges: A benefit of moving the form out of the `FilterBar` and don't add it back to the `XSearch` is that this form could have multiple fields. I don't really see the benefit in adding `@change` to all the input fields. If we had a submit button, I would remove the `@change`-handler from the select.

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: Good point, that's a left-over and not needed anymore 👍

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: Yes, because the as soon as the element grows out of the parent's bounds, the scrollbar would appear. Scrolling is needed otherwise the content would not exactly match the input field anymore.

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: During the testing we found out that this doesn't work well with rtl, therefore I've set it to ltr.

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
johncowen: Oh sorry totally misread this, I thought it was an autofocus, please ignore me! 🙈

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: I have no strong opinion about it, but it replicates more the native behaviour. Using an empty string as a default means that there is always a `placeholder` attribute added to the DOM element

image

I see there is a benefit in having it always `string` if we would operate on the placeholder.

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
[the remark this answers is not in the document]
schogges: Some more thoughts:
- I think the `preventDefault` is needed because otherwise a form submit would refresh the page, which we don't want in SPA environment.
- I think it's a lot easier to just add a name instead of the `@change` over and over again
- The `onSearch` function is not much bigger than a single `@change`
- I would have inlined it into the template but unfortunately we can't use `window` in templates 🤔

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: Ah 😄 with autofocus I would agree and we should not use it here 👍

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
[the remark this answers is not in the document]
johncowen: > I think the preventDefault is needed because otherwise a form submit would refresh the page, which we don't want in SPA environment.

Yeah totally, the screengrab above keeps the `@submit.prevent`, which prevents the form submission/page refresh

> I think it's a lot easier to just add a name instead of the @change over and over again

If it was a form with a submit button and lots of input elements, then maybe I'd agree, but we would probably be using KInput::model-values.

The even more important thing is, its not a @change over and over again. We already have a @change in order to submit the form for the select change, so in reality we'd add one more `@change` with 

```vue
@change="(value: string) => route.update({ page: 1, s: value })"
```

This is one line of code. Nothing else.

Instead we've added, this same one line in a difference place (on the form), and additionally made a custom `onSearch` small function that is outside of the declarative template (which you've noted is problematic to do). This is need to convert the form data back to the actual string that we need to put in the URL, that we already have as a result of the @change event.

Granted we had to make the XSearch component emit a change event, but all our form elements should emit change events (and all of the rest of them do), so arguably this should be added no matter which way we go.

All in all, for me the "its a lot easier" argument doesn't convince me here.

We haven't really discussed that the input element submits differently to the select element. Plus, if you submit the input element both values are submitted whereas if you submit the select element only the select element is submitted.

---

Overall _I am interested in_ us exploring using native form element names here (and overall elsewhere) and I appreciate that you are trying this out now, so I don't want to be to negative on this 😬 , just not sure if now is the right point, especially as I think if we are doing like this its not ideal.

If/when we start using form element names, I would prefer:

- The onSearch function would be moved to `RouteView#form.params(event)` or similar. So it works similar to `route.params` and is available in all RouteViews (seeing as it is a general utility that is commonly and probably exclusively used in routes)
- Whilst names don't necessarily need to be unique, they can cause unexpected problems if they aren't. We have a uniqueId generator available to us as part of RouteView, but this then complicates matters because the name is no longer the name of the route param. Should we not worry about unique ids in this case? I don't know. I'm tempted to say it doesn't matter in this case. We should consider it more carefully.

Maybe we should look at where else we use forms (and even should be using forms), and see if using names would work everywhere? wdyt?

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
johncowen: Where do we set this to use rtl? Isn't ltr the default? Or am I misunderstanding here?

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
johncowen: Ah! So this is why we need overflow-x: auto? But "please don't ever show a scrollbar"?

schogges: Yes, exactly 🙂

schogges: This is how it'd look without `scrollbar-width: none`

image

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: Hm, I set it manually and I think there are extensions that could do that in case people translate a webpage. Also setting `dir='rtl'` on the html tag flips all text.
I've played around a little and this setting actually breaks the behaviour in case the whole page is set to rtl. So I will remove it. Thanks for challenging it 👍

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
johncowen: Perfect! Gotcha, thanks!

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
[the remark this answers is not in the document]
schogges: Ok, synced on it and will change it to use `@change` to stay consistent with the other parts. Will open an issue to rework the overall form behaviour and eventually we come back to this.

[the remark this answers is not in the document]
johncowen: Sounds good to me, thanks again for being so flexible 👍

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -24,48 +24,49 @@
...
+                  v-if="item !== 'all'"
+                  :name="item"
+                />
+                {{ t(`data-planes.type.${item}`) }}
+              </template>
+            </XSelect>
+
+            <input
+              type="submit"
+              hidden
```
johncowen, line 67: Still looking at this but I think this (the hidden input) could have been left over from some experimentation?

schogges: Yes, left-over. Thanks.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
johncowen, line 1: I think we need to add some more bits into the parsing, at least for namespace that we are saying we support here.

I think `namespace:my-namespace` needs to convert to `kuma.io/namespace: my-namespace`.

I'm also not totally sure whether for here (dataplanes) we have labels and tags or just tags, I saw we send tags.

I also just noticed the regex comments and rule between lines 10 and 20 in this file.

I guess, for this PR we should at least do the namespace thing, wdyt?

schogges: Nice catch 👍 I think it's actually `k8s.kuma.io/namespace`.

johncowen: > I'm also not totally sure whether for here (dataplanes) we have labels and tags or just tags, I saw we send tags.

If you look in ZoneIngresses/ZoneEgresses, it uses a `filter=` to search for ingresses/egresses for a particular `kuma.io/zone` and I think we are supposed to use that for searching by label. I think this was the functionality that was added everywhere (so we can use `filter=` by labels on all resources)

I think searching by tags is a different thing and can return a different result set to searching by a label.

Does that make sense what I'm wondering?

schogges: Hmmm, not sure if this is applied everywhere. I've added https://github.com/kumahq/kuma/issues/13395 to check if it is applied to endpoints of DPP listings and if not to implement it. From https://github.com/kumahq/kuma/pull/12840 a lot of endpoints support it, but not sure for DPPs 🤔 

Overall I'm not trying to change the whole logic behind for now, I just don't want to break it while adding the new component. If there is an issue/bug with the filter logic, that's worth a separate PR.

johncowen: Ok if thats the case then we should only mention that we are filtering on `tags` here? I don't think using `tags=kuma.io/service` also looks for labels with the corresponding tag? (they might be the same key/values in labels and tags but sometimes they aren't) I might be wrong on this, but I would imagine using `tags=` for searching doesn't mean you are searching by `labels=`?

packages/kuma-gui/features/mesh/dataplanes/Index.feature
```diff
@@ -101,7 +101,7 @@ Feature: mesh / dataplanes / index
     Then the "$input-search" element isn't disabled
     And I wait for 500 ms
     When I "type" "service:system-1" into the "$input-search" element
-    And I click the "$button-search" element
+    And I "type" "{enter}" into the "$input-search" element
```
johncowen, line 104: Is the `{enter}` thing a Cypress thing do you know? Have you got a link handy that explains that by any chance?

johncowen (approved):
LGTM! Thanks for the work and patience with my queries here 👍

packages/kuma-gui/features/mesh/dataplanes/Index.feature
schogges: Yes, this is how to type specific keys in cypress: https://docs.cypress.io/api/commands/type#Arguments (in the table below)

johncowen: Gotcha, thanks!

PR #3748, opened 2025-04-09
feat(x-search): add dropdown

### LINKED ISSUE #3709: Implement dropdown  for `XFilterBar` component
### Description

As a next step we can implement and add the dropdown for the `XFilterBar` component, which gives more infos and suggestions based on the props of the `XFilterBar` component.

Requirements:
- Should leverage [KDropdown](https://kongponents.konghq.com/components/dropdown.html)
- The first row shows the current filter based on the value in the input and shows possible actions (i.e. "Submit") based on the design (see umbrella issue)
- The second row documents what is possible to filter for and the logic applied to the filters (for now only `AND`)
- The dropdown should open as soon as the user focuses the input and close when unfocused or on submit

schogges:
Adds a dropdown to the `XSearch` component which displays some more information about the filter/search logic, i.e. the actual search query.
In order to make the component more reusable in different places, I've turned the props into optional and provide fallbacks and defaults within the component.

I've added `formatList` (coming from our i18n, `formatjs createIntl().formatList`) to `XI18n` such that we have the dependency to i18n only in one place.

I might change the internal behaviour of `defaultKey`, but first want to see how this component will be used in other places. Probably will change it to fall back to `keys[0]` instead of fixed `name`, since there might be places where we don't have a `name` but still don't want set a `defaultKey` 🤔 

~In order to make the `XI18n` usable inline, I've added an optional prop `tag`, which has one option for now: `span`. So the `XI18n` component can now render as either `div` or `span`. Can eventually grow in case we want to support headlines and more. I probably wouldn't allow any string for this, but also it might be hard to track the whole list of HTML tags that are usually used for text. Therefore I would keep an eye on that and eventually reevaluate the decision here.~

I've also started with some basic documentation and will add more step by step.

Closes #3709

### DIFF
--- a/packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
+++ b/packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
@@ -22,6 +22,7 @@
         v-if="slotName === 'default'"
         :name="slotName"
         :t="t"
+        :format-list="i18n.formatList"
       />
       <XTeleportTemplate
         v-else
@@ -29,6 +30,7 @@
       >
         <slot
           :t="t"
+          :format-list="i18n.formatList"
           :name="slotName"
         />
       </XTeleportTemplate>
@@ -40,6 +42,7 @@
     <slot
       name="default"
       :t="t"
+      :format-list="i18n.formatList"
     />
   </template>
 </template>
--- a/packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
+++ b/packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
@@ -2,9 +2,8 @@
   <template
     v-if="props.path.length > 0"
   >
-    <!-- eslint-disable vue/no-v-html, vue/no-v-html-on-component, vue/no-v-text-v-html-on-component -->
-    <component
-      :is="props.tag ?? 'div'"
+    <!-- eslint-disable vue/no-v-html -->
+    <div
       class="x-i18n"
       data-testid="x-i18n"
       v-bind="attrs"
@@ -64,14 +63,12 @@ const props = withDefaults(defineProps<{
   path?: string
   params?: Record<string, string>
   defaultPath?: string
-  tag?: 'span'
 }>(), {
   strings: undefined,
   prefix: '',
   path: '',
   params: () => ({}),
   defaultPath: undefined,
-  tag: undefined,
 })
 const slots = defineSlots()
 
--- a/packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
+++ b/packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
@@ -2,8 +2,9 @@
   <template
     v-if="props.path.length > 0"
   >
-    <!-- eslint-disable vue/no-v-html -->
-    <div
+    <!-- eslint-disable vue/no-v-html, vue/no-v-html-on-component, vue/no-v-text-v-html-on-component -->
+    <component
+      :is="props.tag ?? 'div'"
       class="x-i18n"
       data-testid="x-i18n"
       v-bind="attrs"
@@ -63,12 +64,14 @@ const props = withDefaults(defineProps<{
   path?: string
   params?: Record<string, string>
   defaultPath?: string
+  tag?: 'span'
 }>(), {
   strings: undefined,
   prefix: '',
   path: '',
   params: () => ({}),
   defaultPath: undefined,
+  tag: undefined,
 })
 const slots = defineSlots()
 
--- a/packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
+++ b/packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
@@ -13,7 +13,7 @@
       data-testid="filter-bar"
       @click="inputRef?.focus()"
     >
-      <XI18n v-slot="{ t }">
+      <XI18n v-slot="{ t, formatList }">
         <div class="icon-wrapper">
           <XIcon
             class="icon"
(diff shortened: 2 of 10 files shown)


packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -1,58 +1,140 @@
...
+          >
+            <span :class="{ highlight: regex.test(chunk) }">{{ chunk }}</span>
+          </template>
+        </div>
+        <div class="input-wrapper">
+          <input
+            ref="inputRef"
+            type="text"
+            :defaultValue="props.value"
+            :placeholder="props.placeholder ?? t('components.x-search.filterBy', { count: props.keys.length, keys: formatList(props.keys, { type: 'disjunction' }) })"
```
schogges, line 42: In the docs of [formatList](https://formatjs.github.io/docs/intl/#formatlist) it's still recommended to check for browser support and eventually use a polyfill, because of the internal use of [Intl.ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat). I think we don't need to polyfill anymore because it's widely adopted and according to [caniuse](https://caniuse.com/?search=Intl.ListFormat) available to ~95% of global users.

packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
```diff
@@ -63,12 +64,14 @@ const props = withDefaults(defineProps<{
   path?: string
   params?: Record<string, string>
   defaultPath?: string
+  tag?: 'span'
```
schogges, line 67: I've seen cases where this prop is named `as`. Wdyt? 🙂

johncowen (commented):
I left a couple of initial thoughts but mainly on this:

> In order to make the XI18n usable inline, I've added an optional prop tag, which has one option for now: span. So the XI18n component can now render as either div or span. Can eventually grow in case we want to support headlines and more. I probably wouldn't allow any string for this, but also it might be hard to track the whole list of HTML tags that are usually used for text. Therefore I would keep an eye on that and eventually reevaluate the decision here.

Unless I've missed something here you should be able to use `t`:


```vue
<span>{{ t() }}</span>
<h1>{{ t() }}</h1>
etc etc
```

You only absolutely need to use the component itself when you are passing in markdown/html in which case you probably want it wrapped in a `div`

Lmk on ☝️ in case I missed something, and I'll continue to review this 👍

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -1,58 +1,140 @@
...
+          class="filter-block"
         >
+          <template
+            v-for="(chunk, i) in inputValue.split(regex).filter(Boolean)"
+            :key="chunk+i"
+          >
+            <span v-if="regex.test(chunk)">
+              <template
+                v-for="([key, ...values], j) in [chunk.split(':')]"
+                :key="key+j"
```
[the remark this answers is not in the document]
schogges: AFAIK it's recommended to not use the index only. Arrays are mutable and the order can change. Therefore I'm using both `key` and the index to set a key that vue can use to optimize rendering. Also the `key` is not unique because a user can type the same key multiple times. I think the "best" `:key` to be used is a uuid.

johncowen:
> I think I was on a wrong path thinking we want to exclusively use x-components in our templates.

Not sure if I misunderstood this ^ , we do still want to exclusively use x-components if possible, but we can still use `t` from either `XI18n` or `RouteView`

schogges:
> > I think I was on a wrong path thinking we want to exclusively use x-components in our templates.
> 
> Not sure if I misunderstood this ^ , we do still want to exclusively use x-components if possible, but we can still use `t` from either `XI18n` or `RouteView`

Well in that case I'm confused 🤔 

> Unless I've missed something here you should be able to use `t`:
> 
> ```vue
> <span>{{ t() }}</span>
> <h1>{{ t() }}</h1>
> etc etc
> ```

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
johncowen: Are there any disadvantages to using the native approach instead of the dependency? Something like this:

```typescript
keys: new Intl.ListFormat(undefined, {type: 'disjunction'}).format(props.keys)
```

johncowen:
> Well in that case I'm confused 🤔

So am I 😆 , honestly I'm still not totally sure I understood right the point you were making. 

I think the PR looks good now and we are using `t` from `XI18n` which is still using x-components in our templates.

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: This would work, but as soon as we add more languages there is a high risk that there is a mismatch in the languages, because there might be different strategies to determine that applied language. Even worse once we'd add a language switcher that applies a different language than the browser.

packages/kuma-gui/src/app/x/components/x-i18n/XI18n.vue
schogges: Removed this part.

schogges:
@johncowen I've changed what we discussed yesterday 🙂

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
johncowen: I realized after we spoke that `i18n.locale` exists, is this the actual locale string, looks it to me? If so we could do:

```vue
:list=(strs, opts) => new Intl.ListFormat(i18n.locale, opts).format(strs)
```

If that works, then not necessary to do it here and now, we can leave for a follow up

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -1,58 +1,141 @@
...
+                >
+                  <template v-if="!values.length">
+                    <dt>{{ props.defaultKey }}:</dt><dd class="text-important">
+                      {{ key }}
+                    </dd>
+                  </template>
+                  <template v-else>
+                    <dt>{{ key }}:</dt><dl class="text-important">
+                      {{ values.join(':') }}
+                    </dl>
```
johncowen, line 79: Just a little note here (I mentioned offline) I think this `dl` should be a `dd`, also one last tiny nit. Now we are using `dt`s and `dd`s you shouldn't need the tailwind-like `text-important` etc, just style the `dd`s.

Lets catch up when you are around tho and I can give this a 👍

schogges: Ah 😆 yes, that should be a `dd`, thanks for catching this 👍 
I'd still keep the `text-important`, because I am using it further down another time inside the second dropdown item.

johncowen (approved):
Looks good 👍

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
schogges: Ah nice find 🙂 will give this a try in a follow up 👍

PR #3756, opened 2025-04-11
chore: use CSS isolation to avoid z-index battles

johncowen:
A better fix for https://github.com/kumahq/kuma-gui/pull/3729. It also means that if we ever need bulk actions with a max-height (this is the only configuration of KTable that requires the  z-index on the thead), the Kongponent will still be displayed as intended.

We might want to add this to every single X component to isolate ourselves from Kongponent z-index related side effects.

Note: I have a much better branch hanging around that has much better isolation of Kongponents which would prevent us from having these sorts of issues, I keep meaning to resurrect it 🤔 

Probably a good FYI to keep in mind from when I was looking into this:

The whole issue is caused by Kongponents providing custom/non-native icons for checkboxes:

Screenshot 2025-04-11 at 12 04 38

Lastly, we are due to change our dropdown filterbar thingie here soon, but I would guess we might hit this again at some point. Also good to keep in mind this checkbox weirdness because I've seen the tick icon disappearing before now and wondered "wow, what is happening here?!?!?"

### DIFF
--- a/packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
+++ b/packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
@@ -144,15 +144,13 @@ onMounted(rewrite)
 </script>
 
 <style lang="scss" scoped>
+.app-collection {
+  isolation: isolate;
+}
 .app-collection :deep(td:first-child a) {
   color: inherit;
   font-weight: $kui-font-weight-semibold;
   text-decoration: none;
-}
-.app-collection :deep(thead) {
-  /* overwrite kongponents setting this to z-index:2 */
-  /* which causes overlay issues with other components */
-  z-index: auto !important;
 }
 
 .app-collection :deep(td:first-child li a) {


schogges (approved):
Great finding 👍 

> We might want to add this to every single X component to isolate ourselves from Kongponent z-index related side effects.

I don't think that it's a solution for everything. Some of the components need to share a stacking context such as tooltips and dropdowns because they should overlay everything else.

johncowen:
> Some of the components need to share a stacking context such as tooltips and dropdowns because they should overlay everything else.

Right, but the things inside the tooltip, or the things inside the dropdown would have its own stacking context, the component itself would use the context of the application. You follow what I mean, the difference between the internals of a component and the application that is made up of components.

schogges:
> > Some of the components need to share a stacking context such as tooltips and dropdowns because they should overlay everything else.
> 
> Right, but the things inside the tooltip, or the things inside the dropdown would have its own stacking context, the component itself would use the context of the application. You follow what I mean, the difference between the internals of a component and the application that is made up of components.

Yes I think I follow 🙂 but what if there is a dropdown, the dropdown content has it's own stacking context and there is tooltip. Would have to check, but my assumption is that the other elements on the page that have a higher stacking order will overlay the tooltip 🤔

johncowen:
> Would have to check, but my assumption is that the other elements on the page that have a higher stacking order will overlay the tooltip

If I follow correctly, this would be the application layer of z-indexing I think. You would give the tooltip a z-index of 1, and it would mean that anything in the dropdown that with a z-index: 2 would still be underneath the tooltip (if dropdown and tooltip are siblings) because both the tooltip and the dropdown have their own separate stacking contexts.

schogges:
> > Would have to check, but my assumption is that the other elements on the page that have a higher stacking order will overlay the tooltip
> 
> If I follow correctly, this would be the application layer of z-indexing I think. You would give the tooltip a z-index of 1, and it would mean that anything in the dropdown that with a z-index: 2 would still be underneath the tooltip (if dropdown and tooltip are siblings) because both the tooltip and the dropdown have their own separate stacking contexts.

Hm yeah, could be that when keeping a z-index on the wrapping elements (like it is already) and only create a new stacking context for inner elements will work.

PR #3767, opened 2025-04-15
chore: import KToggle into XDisclosure

johncowen:
We were playing around making things better and we noticed that this import was missing, I figured I may aswell PR it now rather than just reset my local clone


schogges (approved):
👍

PR #3768, opened 2025-04-15
chore(deps): remove dotenv

johncowen:
Putting this up in draft while I verify this is fine.

---

Removes top-level dependency (actually surprisingly removes the dependency entirely 🎉 , I thought vite would use this but apparently not) and removes a weirdly named `VITE_` var which is only used in Cypress not `VITE_`

Probably related to https://github.com/kumahq/kuma-gui/issues/2431 and our overall work to remove unnecessary environment variables that should either be i18n strings, or something else. In this case the something else is "actual bash env var falling back to hardcoded", rather than "actual bash env var falling back to `.env` file var" (loss of locality).

There'd maybe a case for keeping this if we had a tonne of env vars that we used everywhere, or if we needed to switch locally often - neither of which we need.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7473,19 +7473,6 @@
         "url": "https://github.com/fb55/domutils?sponsor=1"
       }
     },
-    "node_modules/dotenv": {
-      "version": "16.5.0",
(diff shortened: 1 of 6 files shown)


johncowen:
FYI: This will hang around for a while, and depending on workload/priorities might get closed to be reopened at a much later date.

PR #3769, opened 2025-04-15
feat: add connection icons to subscription listings

### LINKED ISSUE #3183: Add new style icons to subscriptions
### Description

One thing I wanted to do with the subscriptions re-design was to add some icons back into each row denoting whether the subscriptions is connected or not. It would look similar to our Zone listings that have universal/kubernetes icons. I was thinking of using the VitalsIcon (https://kong.github.io/icons/?q=vitals) in green and the Remove Icon (https://kong.github.io/icons/?q=remove) which looks like a line, in grey.

Didn't quite get around to doing that, but we should circle back and do that.

schogges:
For easier identification of connection issues, we've decided to add an icon to the subscription listings. The connection is healthy if there is a connection time and no disconnection time. In all other cases the connection is unhealthy.

---

For the icons we already have a `name=connected` icon, which is the `PlugIcon`. The `VitalsIcon` was already used for `health` in one case and I've decided to reuse this one for `healthy` vs `unhealthy` (displayed as `RemoveIcon`).

Closes #3183


packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
```diff
@@ -118,6 +118,7 @@ data-planes:
           title: Supported CAs
       subscriptions:
         title: XDS Connections
+        description: Connections between Control Plane Instances and Data Plane Proxies.
```
schogges, line 121: Not super sure about this 🙂

packages/kuma-gui/src/app/zone-egresses/locales/en-us/index.yaml
```diff
@@ -30,6 +30,7 @@ zone-egresses:
       config: 'Configuration'
       subscriptions:
         title: 'XDS Connections'
+        description: Connections between Control Plane Instances and Data Plane Proxies.
```
schogges, line 33: Not super sure about this 🙂

packages/kuma-gui/src/app/zone-ingresses/locales/en-us/index.yaml
```diff
@@ -31,6 +31,7 @@ zone-ingresses:
       config: Configuration
       subscriptions:
         title: 'XDS Connections'
+        description: Connections between Control Plane Instances and Data Plane Proxies.
```
schogges, line 34: Not super sure about this 🙂

packages/kuma-gui/src/app/zones/locales/en-us/index.yaml
```diff
@@ -77,7 +77,9 @@ zone-cps:
       store-memory: Uses memory store
       global-cp-incompatible: Version mismatch
   detail:
-    subscriptions: 'KDS Connections'
+    subscriptions:
+      title: 'KDS Connections'
+      description: Connections of Control Plane Instances and Data Plane Proxies for Kuma Service Detection (KDS).
```
schogges, line 82: Not super sure about this 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -624,62 +624,78 @@
...
-                  { ...me.get('headers.responses'), label: t('http.api.property.responses'), key: 'responses' },
-                ]"
-                :is-selected-row="item => item.id === route.params.subscription"
-                :items="props.data.dataplaneInsight.subscriptions.map((_, i, arr) => arr[arr.length - (i + 1)])"
-                @resize="me.set"
-              >
-                <template
-                  #instanceId="{ row: item }"
+              <XLayout>
+                <p>{{ t('data-planes.routes.item.subscriptions.description') }}</p>
```
johncowen, line 628: Just need to switch this to use X18n component with markdown

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -733,6 +749,12 @@ const props = defineProps<{
   margin-left: auto;
 }
 
+:deep(td:nth-child(2) a) {
+  color: inherit;
+  font-weight: $kui-font-weight-semibold;
+  text-decoration: none;
+}
+
```
johncowen, line 757: I remembered than because we moved AppCollections to use `data-action` tagging instead of saying "always use the first td a", we can probably amend the selector in AppCollection to use that (possibly even get rid of the old one):

https://github.com/kumahq/kuma-gui/blob/906cd3369f0e7925734edd20a76d10e9cd8e8b77/packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue#L150-L154

```css
.app-collection :deep(a[data-action]),
.app-collection :deep(td:first-child a) { 
   color: inherit; 
   font-weight: $kui-font-weight-semibold; 
   text-decoration: none; 
 } 
```

Good for a followup or here I'm easy

packages/kuma-gui/src/app/services/views/MeshServiceDetailView.vue
```diff
@@ -210,7 +210,7 @@
               data-testid="healthy-dpps"
             >
               <template #icon>
-                <XIcon name="health" />
+                <XIcon name="healthy" />
```
johncowen, line 213: Looks simple but isn't necessarily, good call 👍

packages/kuma-gui/src/app/zones/locales/en-us/index.yaml
```diff
@@ -77,7 +77,9 @@ zone-cps:
       store-memory: Uses memory store
       global-cp-incompatible: Version mismatch
   detail:
-    subscriptions: 'KDS Connections'
+    subscriptions:
+      title: 'KDS Connections'
+      description: Communication channels used to synchronize configuration and policies between the global and zone control planes.
```
johncowen, line 82: I think all these descriptions should use `!!text/markdown |` so that the end up wrapped in p's

johncowen (commented):
Looks great, I left two things, the only blocker is the markdown/p comment, others aren't necessary/can be done whenever

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
schogges: Yes, I thought the same. Would do it in a follow-up.

packages/kuma-gui/src/app/zones/locales/en-us/index.yaml
schogges: Done 👍

johncowen (approved):
LGTM!

PR #3776, opened 2025-04-17
refactor(dataplane-detail-view): move tls card content to about card

### LINKED ISSUE #3411: Consider moving TLS info into "About this Data Plane Proxy" box
### Description

Image

I think this is the last place where we use this strange "small header, large body" design (apart from for stats), feels like at least some of it is also important information "About this Data Plane Proxy" which would probably be waaay below the fold with an install with lots of services.

Feel free to close if we want to leave as is (or amend the issue if we want to change but use a different approach) 👍

schogges:
_TBD_

Closes #3411

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
+++ b/packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
@@ -113,7 +113,7 @@ data-planes:
         regenerations:
           title: Generation
         issued_backend:
-          title: Certificate issued by
+          title: Issued by
         supported_backends:
           title: Supported CAs
       subscriptions:
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -85,148 +85,233 @@
             :created="props.data.creationTime"
             :modified="props.data.modificationTime"
           >
-            <DefinitionCard
-              layout="horizontal"
-            >
-              <template
-                #title
+            <XLayout>
+              <XLayout
+                type="separated"
               >
-                {{ t('http.api.property.status') }}
-              </template>
-
-              <template
-                #body
-              >
-                <XLayout
-                  type="separated"
+                <DefinitionCard
+                  layout="horizontal"
                 >
-                  <StatusBadge :status="props.data.status" />
-                  <DataCollection
-                    v-if="props.data.dataplaneType === 'standard'"
-                    :items="props.data.dataplane.networking.inbounds"
-                    :predicate="item => item.state !== 'Ready'"
-                    :empty="false"
-                    v-slot="{ items: unhealthyInbounds }"
+                  <template
+                    #title
                   >
-                    <XIcon name="info">
-                      <ul>
-                        <li
-                          v-for="inbound in unhealthyInbounds"
-                          :key="`${inbound.service}:${inbound.port}`"
+                    {{ t('http.api.property.status') }}
+                  </template>
+
+                  <template
+                    #body
+                  >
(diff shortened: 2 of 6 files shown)


packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
```diff
@@ -107,11 +109,11 @@ data-planes:
...
         title: 'TLS'
         expiration_time:
-          title: Certificate expiration time
+          title: Certificate expires at
         generation_time:
-          title: Certificate generation time
+          title: Certificate generated at
         regenerations:
-          title: Generation
+          title: Certificate generation
```
johncowen, line 116: 👏 I much prefer the more human texts

packages/kuma-gui/src/app/x/components/x-layout/XLayout.vue
```diff
@@ -28,6 +28,9 @@ const props = withDefaults(defineProps<{
 .stack.small > * + * {
   margin-block-start: $kui-space-40;
 }
+.max {
+  width: 100%;
+}
```
johncowen, line 33: Could you lemme know what this is doing? 👍

johncowen (commented):
One q, but otherwise looks good to me, lemme know one that one 👍

packages/kuma-gui/src/app/x/components/x-layout/XLayout.vue
schogges: Without this, the `XLayout` wouldn't grow to the full size. I specifically needed this to make sure that the visual separator in the about card is not cut off somewhere in the middle. Without this, the size of `XLayout` would only grow with its content.

johncowen: I noticed that this seems to be a consequence of AppAboutSection though I think?

Screenshot 2025-04-30 at 14 17 21

I'd rather not fix AppAboutSection's problems in XLayout if at all possible, is there a way to fix AppAboutSection in a different way?

schogges: Hmm good point, will check that. brb

schogges: No, `.about-section-content` is full width. The purple area is the space that the element can grow. So it's actually 100% already, but the nested element `x-layout` is not taking it.

johncowen: What happens if you remove the `display: flex` on the `.about-section-content`?

schogges: I've set it to `display: block`, but that breaks it in other places, see examplary screengrab below (I've noted what we would probably need to do)
image

schogges: This is something we can do separately. At this point I don't want to add the `.max` to `XLayout` and I think I can add a special case here and set `display: block` to this one case of `XAboutCard`. Wdyt?

johncowen: That sounds better to me. We'd be addressing the problem closer to where the issue is. And I think we would both agree that whilst not totally "solving it" but pragmatism wins and we can merge this in and maybe we come up with something better in the longer term (or maybe not and its fine)

johncowen (approved):
LGTM! We had a random chat whilst reviewing this, and discussed the potential of just using `dl/dt/dd` for the things inside the XAboutCards therefore removing the need to use DefinitionCard. The `dl/dt/dd` would be styled from inside the XAboutCard and be a step towards getting rid of DefinitionCard altogether.

Issue #3570, opened 2025-02-19
routing: `control-planes` routing should be at the root of our application


johncowen:
### Description

At the root our our application routing we currently have three siblings:

- `/` itself (i.e. Home/ControlPlanesDetailView)
- `/meshes` (the Mesh listing MeshListView)
- `/zones` (the Zone listing ZoneListView)

This is incorrect as a mesh and a zone live inside a control plane.

We should have a single root `/` (i.e. "control plane root" which we don't currently have) with two children as siblings

- `/` (i.e. "control plane root" which we don't currently have)
  - `overview` itself (i.e. Home/ControlPlanesDetailView)
  - `meshes` (the Mesh listing MeshListView)
  - `zones` (the Zone listing ZoneListView)

This then gives us a place to apply/display things always for an entire single control plane instead of using `App.vue` to do so.

---

Perhaps separately (or perhaps not) to this, I would really like to get to the point where we 'mount' our routing configurations ontop of each over from the outside rather than from the inside (we currently do this from the inside meaning tight-coupling of dependencies):

i.e.

```javascript

// kinda pseudo-code

controlPlaneRoutes

// mount(theRoutes, 'where.to.mount.them')

mount(meshRoutes, 'control-planes')
mount(serviceRoutes, 'control-planes.meshes')
mount(dataplaneRoutes, 'control-planes.meshes')
mount(subscriptionRoutes, 'control-planes.meshes.dataplanes')
mount(networkingRoutes, 'control-planes.meshes.dataplanes')

mount(zoneRoutes, 'control-planes')
mount(subscriptionRoutes, 'control-planes.zones')

mount(ingressRoutes, 'control-planes.zones')
mount(subscriptionRoutes, 'control-planes.zones.ingresses')
mount(networkingRoutes, 'control-planes.zones.ingresses')

mount(egressRoutes, 'control-planes.zones')
mount(subscriptionRoutes, 'control-planes.zones.egresses')
mount(networkingRoutes, 'control-planes.zones.egresses')

```

This way we have full modular control of our routes (and reusable routes) from The Outside (i.e. looser coupling of dependencies)

johncowen:
Just remembered why/where I noticed this was wrong, and its not clear from the issue so:

[...]

PR #3789, opened 2025-04-22
feat(notifications): adds dismissal, GC and resetting to notifications

johncowen:
At a general level this PR adds a dismissal button to all our notifications.

If you dismiss a an alert box, all notifications within that box are saved to localStorage as dismissed for 7 days at which point it is "un-dismissed" meaning you will see the notification again if you haven't resolved what it is warning you about.

localStorage values also be removed i.e. "un-dismissed" if we detect that you have resolved the reason for the warning, meaning if you then remove whatever you did to resolve the warning, you will be warned again.

Further notes:

- I moved around AppView a little so I could put all the notification display all in one place in AppView (instead of also in ApplicationShell), this also meant I could remove some CSS that we no longer need (we are now using XLayout for the same layout)
- ~I moved the memory store notification to the `App.vue` file so that it is visible all the time, this will mean we will have to duplicate this. Ideally this notification would be in ControlPlaneDetailView, but the routing structure is incorrect (see https://github.com/kumahq/kuma-gui/issues/3570)~ We resolved the linked issue with further work and now the notification is where we want it in ControlPlaneDetailView

Closes https://github.com/kumahq/kuma-gui/issues/3605

### DIFF
--- a/packages/kuma-gui/src/app/me/sources.ts
+++ b/packages/kuma-gui/src/app/me/sources.ts
@@ -3,18 +3,42 @@ import merge from 'deepmerge'
 import { defineSources } from '../application/services/data-source'
 
 type Storage = {
-  get: (key: string, d?: object) => Promise<object>
-  set: (key: string, value: object) => Promise<object>
+  get: (key: string, d?: object) => Promise<[]>
+  set: (key: string, value: object) => Promise<[]>
 }
+
+type Notification = {
+  timestamp: number
+  uri: string
+}
+const DAYS = 1000 * 60 * 24
 export const sources = ({ get, set }: Storage) => {
   return defineSources({
     // used for saving dismissed notification id's
     '/me/~notifications': async () => {
-      return get('~notifications', [])
+      const dismissed: Notification[] = await get('~notifications', [])
+      const current = dismissed.filter(item => {
+        return (item.timestamp + (7 * DAYS)) > Date.now()
+      })
+      if(dismissed.length !== current.length) {
+        await set('~notifications', current)
+      }
+      const res = current.map(item => item.uri)
+      return res
     },
-    '/me/~notifications/:data': async (params) => {
-      const res = merge<object>(await get('~notifications', []), JSON.parse(params.data))
-      set('~notifications', res)
+    '/me/~notifications/dismiss/:data': async (params) => {
+      const dismissed: string[] = JSON.parse(params.data)
+      const res = merge<object>(await get('~notifications', []), dismissed.map(item => ({
+        timestamp: Date.now(),
+        uri: item,
+      })))
+      await set('~notifications', res)
+      return
+    },
+    '/me/~notifications/reset/:data': async (params) => {
+      const reset = new Set<string>(JSON.parse(params.data))
+      const dismissed = new Set<string>(await get('~notifications', []))
+      await set('~notifications', Array.from(reset.difference(dismissed)))
       return
     },
     //
--- a/packages/kuma-gui/src/app/me/sources.ts
+++ b/packages/kuma-gui/src/app/me/sources.ts
@@ -2,21 +2,23 @@ import merge from 'deepmerge'
 
 import { defineSources } from '../application/services/data-source'
 
-type Storage = {
-  get: (key: string, d?: object) => Promise<[]>
-  set: (key: string, value: object) => Promise<[]>
-}
-
 type Notification = {
   timestamp: number
   uri: string
 }
+type Storage = {
+  get(key: '~notifications', d?: Notification[]): Promise<Notification[]>
+  set(key: '~notifications', value: Notification[]): Promise<Notification[]>
+  get(key: string, d?: object): Promise<object>
+  set(key: string, value: object): Promise<object>
+}
+
 const DAYS = 1000 * 60 * 24
 export const sources = ({ get, set }: Storage) => {
   return defineSources({
     // used for saving dismissed notification id's
     '/me/~notifications': async () => {
-      const dismissed: Notification[] = await get('~notifications', [])
+      const dismissed = await get('~notifications', [])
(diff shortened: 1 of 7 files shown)


packages/kuma-gui/src/app/me/sources.ts
```diff
@@ -3,18 +3,42 @@ import merge from 'deepmerge'
 import { defineSources } from '../application/services/data-source'
 
 type Storage = {
-  get: (key: string, d?: object) => Promise<object>
-  set: (key: string, value: object) => Promise<object>
+  get: (key: string, d?: object) => Promise<[]>
+  set: (key: string, value: object) => Promise<[]>
```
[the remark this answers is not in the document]
johncowen: 🤔 lemme look at this a bit closer

packages/kuma-gui/src/app/me/sources.ts
```diff
@@ -2,19 +2,49 @@ import merge from 'deepmerge'
...
+  timestamp: number
+  uri: string
+}
 type Storage = {
-  get: (key: string, d?: object) => Promise<object>
-  set: (key: string, value: object) => Promise<object>
+  get(key: '~notifications', d?: Notification[]): Promise<Notification[]>
+  set(key: '~notifications', value: Notification[]): Promise<Notification[]>
+  get(key: string, d?: object): Promise<object>
+  set(key: string, value: object): Promise<object>
```
schogges, line 13: Hmm 🤔 I'd suggest to use generics for this, i.e. like

```suggestion
  get<T = object>(key: string, d?: T): Promise<T>
  set<T = object>(key: string, value: T): Promise<T>
```

and further down we could do

```ts
const dismissed = await get<Notification[]>('~notifications', [])
```

packages/kuma-gui/src/app/me/sources.ts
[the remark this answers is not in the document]
schogges: Ah 😆 comment ended up as a new conversation. See below.

packages/kuma-gui/src/app/me/sources.ts
johncowen: Hmm, maybe 🤔 

What's then stopping me from doing `await get<Notification[]>('~not-a-notification-shaped-thing', [])`?

I kinda want to say, we control the `~notifications` namespace, and whenever you access that exact namespace (get or set), this is the shape you use/get. With the generic approach I lose that control I think?

Are there any other benefits to the generic approach over the overloading that you can think of here?

johncowen: >> What's then stopping me from doing await get<Notification[]>('~not-a-notification-shaped-thing', [])?

> Nothing 😄 but why would you do that?

What about this? 

```js
await get('~notifications', [])
```

(ie. forget)

> Also is there a guarantee that what is being returned is what we expect? What if it's a remote API? Maybe a type guard would be better?

> What if it's a remote API?

It's not a remote API, but then we would have auto-generated types so I wouldn't have to do anything type-wise.

> Maybe a type guard would be better?

I get the feeling we are overthinking this a little, we don't add type guards to our local storage access anywhere else, this isn't a remote API, and in my opinion we don't have scalability issues with our localStorage properties/access.

Honestly my feeling is, either approach is fine.

But, the overload approach means I can set this in one place and I get type checking on all the places in this file where I access `~notifications` which is exactly what I want. Generics doesn't give me that because I have to specify the type in multiple places. I don't at the moment see any benefits to make it a generic instead because I always want any access to `~notifications` to use the same type. i.e.

```typescript
type LocalStorage = {
  notifications: Notification[]
  [key: string]: object
}
```

schogges: > > > What's then stopping me from doing await get<Notification[]>('~not-a-notification-shaped-thing', [])?
> 
> > Nothing 😄 but why would you do that?
> 
> What about this?
> 
> ```js
> await get('~notifications', [])
> ```
> 
> (ie. forget)

Further use would not be typed and things you try to access would cause typescript errors 🤔 

> > Also is there a guarantee that what is being returned is what we expect? What if it's a remote API? Maybe a type guard would be better?
> 
> > What if it's a remote API?
> 
> It's not a remote API, but then we would have auto-generated types so I wouldn't have to do anything type-wise.

Good point 👍 

> > Maybe a type guard would be better?
> 
> I get the feeling we are overthinking this a little, we don't add type guards to our local storage access anywhere else, this isn't a remote API, and in my opinion we don't have scalability issues with our localStorage properties/access.
> 
> Honestly my feeling is, either approach is fine.
> 
> But, the overload approach means I can set this in one place and I get type checking on all the places in this file where I access `~notifications` which is exactly what I want. Generics doesn't give me that because I have to specify the type in multiple places. I don't at the moment see any benefits to make it a generic instead because I always want any access to `~notifications` to use the same type. i.e.
> 
> ```ts
> type LocalStorage = {
>   notifications: Notification[]
>   [key: string]: object
> }
> ```

Yeah, maybe I'm overthinking it. I guess we can leave it with the overloading for now and come back here in case it becomes cumbersome. My initial comment got resolved, so I'm fine either way now 🙂

johncowen: Ok sounds good lemme know if there is anything else here, will resolve this one

schogges (dismissed):
LGTM 👍

johncowen:
@schogges I've added some further amends following https://github.com/kumahq/kuma-gui/pull/3795, so this PR is now a lot more like how I want things to work. The notification that belongs to `control-planes` is added from inside a `control-planes` route rather than in `App.vue` but we render it in `App.vue`s top-level `AppView:notifications`, just waiting for this to go green and I'll bring it back out of draft for one last approval

schogges (approved):
:+1

PR #3790, opened 2025-04-22
chore(ci): reduce the majority of fake-api to a createFetch function

johncowen:
Note this was reviewed previously and reverted, I've made additional amends to the original PR in the second commit here to keep the changes separate. I've also done some "pre-testing" elsewhere to make sure it works once we merge to try and avoid having to revert again if I messed up.

---

Refactors some more of fake-api, mostly:

- Reduce everything down to a reusable `createFetch`/`createFetchSync`, that returns something that is very fetch-like (i.e. very recognisable). Note: `createFetchSync` is needed for Cypress which is notorious weird with async things.
- I moved the YAML/JSON merging we do for overwriting responses during tests out of fake-api entirely. If you aren't doing tests like we are then you wouldn't need this in order to build basic fake-apis.

Probably a few other little bits. Unfortunately the PR is a little large, if it was me I'd look at/compare this locally to get a better feel for the improvement here.

We are still not quite done here, but with what we did before extracting things out into its own module, and what is here, I'm far happier with this code now.

### DIFF
--- a/packages/fake-api/src/cypress.ts
+++ b/packages/fake-api/src/cypress.ts
@@ -1,28 +1,22 @@
-import { createMerge, Router } from './index.ts'
-import type { Callback, Dependencies, FS, Mocker } from './index.ts'
-
-type Server = typeof cy
-
-type HistoryEntry = {
-  url: URL
-  request: {
-    method: string
-    body: Record<string, unknown>
-  }
-}
-type Client = { request: (request: HistoryEntry ) => void }
+import { createFetchSync } from './index.ts'
+import type { Middleware, Dependencies, FS, Mocker } from './index.ts'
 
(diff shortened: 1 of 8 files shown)


johncowen:
Just to note we applied this:

- https://github.com/kumahq/kuma-gui/pull/3734

and then backed out

- https://github.com/kumahq/kuma-gui/pull/3788

As it caused unforeseen issues elsewhere, and we didn't want to spend too much time investigating rn.

The other thing I noticed which is probably related is that, I missed that we need to mock on full absolute URLs (mainly due to `kuma.io/latest_version` mocking, but probably other things)

I have work to go on top of this but haven't gotten around to pushing it yet, will be back in a bit with that

schogges (approved):
👍

PR #3792, opened 2025-04-24
chore(policies): use resources over policies for CPStatus policy count

johncowen:
Uses `resources` instead of `policies` for calculating the count of policies in your control-plane.

We use the `_resources` endpoint to understand which resources are actually policies and therefore include them in the count.

Closes https://github.com/kumahq/kuma-gui/issues/3793


packages/kuma-gui/src/app/control-planes/data/index.ts
```diff
@@ -34,4 +34,6 @@ export const GlobalInsight = {
   },
 }
 
-export type GlobalInsight = ReturnType<typeof GlobalInsight.fromObject>
+export type GlobalInsight = ReturnType<typeof GlobalInsight.fromObject> & {
+  resources: ReturnType<typeof GlobalInsight.fromObject>['resources'] & PartialGlobalInsight['resources']
+}
```
schogges, line 39: Why is this needed? From what I can see it doesn't really change the type? 🤔

schogges (commented):
Just a Q/nit, otherwise LGTM 🙂

packages/kuma-gui/src/app/control-planes/data/index.ts
johncowen: Yeah is a bit of a weird one. If you remove it it thinks that `.resources` only has `MeshService`, `MeshMultizoneService`, `MeshExternalService` (because of the reshaping code we do just above here), but it also has `[key: string]: Resource` on it. This ensures that `[key: string]: Resource` is kept.

schogges: Ah gotcha, but still not sure why we need it? 😅 
Anyways I think this could be reduced 🤔 

```ts
export type GlobalInsight = ReturnType<typeof GlobalInsight.fromObject> & PartialGlobalInsight
```

or

```ts
export type GlobalInsight = ReturnType<typeof GlobalInsight.fromObject> & {
  resources: PartialGlobalInsight['resources']
}
```

johncowen: > Ah gotcha, but still not sure why we need it? 😅

Here's a with and without, note `resources` is a list of resources including all the policies

## With

Screenshot 2025-04-24 at 15 50 52

## Without

Screenshot 2025-04-24 at 15 51 29

---

I tweaked things to use the second suggestion as the result looked the best

schogges: Synced on this and I agree that it's good for documentation purposes and to know that resources can contain more than `MeshService`, `MeshMultiZoneService` and `MeshExternalService`. At the moment we haven't added more, because we don't use it.

schogges (dismissed):

johncowen:
@schogges I think this one just needs another quick look because I re-added back the mocks I'd removed in the PR (I think removing them is causing a flake, ideally they'd be removed and why that caused the flake being sorted, but not to be done in this PR)

schogges (approved):
👍

PR #3795, opened 2025-04-24
chore(control-planes): move routing to be a child of control-planes

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/3570

After this goes in I'll probably make some immediate amends ontop of https://github.com/kumahq/kuma-gui/pull/3789, so ideally this would go in first.

I probably should have done this PR first tbf. The plan is to put the notification about "control-plane using a memory store" in the control-plane module where it belongs. The only reason we haven't done that already is because we had no where to put it that meant its always visible, so it ended up going in App.vue.

### DIFF
--- a/packages/kuma-gui/src/app/hostname-generators/index.ts
+++ b/packages/kuma-gui/src/app/hostname-generators/index.ts
@@ -3,6 +3,7 @@ import { routes } from './routes'
 import { sources } from './sources'
 import type { ServiceDefinition } from '@/services/utils'
 import { token } from '@/services/utils'
+import type { RouteRecordRaw } from 'vue-router'
 
 type Token = ReturnType<typeof token>
 
@@ -18,12 +19,17 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
       ],
     }],
     [token('hostname-generators.routes'), {
-      service: routes,
-      arguments: [
-        app.can,
-      ],
+      service: () => {
+        return [
+          (item: RouteRecordRaw) => {
+            if (item.name === 'control-plane-root-view') {
+              item.children = (item.children ?? []).concat(routes())
+            }
+          },
+        ]
+      },
       labels: [
-        app.routes,
+        app.routeWalkers,
       ],
     }],
     [token('hostname-generators.locales'), {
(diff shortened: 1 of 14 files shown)


schogges (dismissed):
LGTM 👍

johncowen:
Sorry @schogges I spotted on more thing last minute here and added an additional commit

https://github.com/kumahq/kuma-gui/pull/3795/commits/cc8ab6a981f8e63e1c25922cefbb87fb37dc9d2f

Should be straightforwards 👍

schogges (dismissed):

packages/kuma-gui/src/app/meshes/routes.ts
```diff
@@ -13,7 +13,7 @@ export const routes = (
 ): RouteRecordRaw[] => {
   return [
     {
-      path: '/meshes',
```
johncowen, line 16: These slashes ended up being the main "root" of my issues. Now we are nesting everything within a `/` nothing else should have a `/` in front of them.

packages/kuma-gui/src/app/hostname-generators/index.ts
```diff
@@ -18,12 +19,17 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
...
-      ],
+      service: () => {
+        return [
+          (item: RouteRecordRaw) => {
+            if (item.name === 'control-plane-root-view') {
+              item.children = (item.children ?? []).concat(routes())
+            }
+          },
+        ]
+      },
```
johncowen, line 30: Just to note, previously the parent was in control of where these things get appended to, which meant passing all of the children in to the parent.

With this PR, the child is now in control of where it gets mounted, which means we don't have to pass all of the children into the parent, and the parent has no knowledge of its children.

Ideally we want to totally decouple these entirely, with something like this in the `main.ts` file


```js
mount(meshRoutes, 'control-plane-root-view')
mount(zonesRoutes, 'control-plane-root-view')
mount(hostnameGeneratorRoutes, 'control-plane-root-view')
```

It's not totally clear exactly how we are gonna get there yet, but this is where we are going as it totally decouples the modules, i.e. neither the parent nor the child needs to know about each other. We plumb them together in the wiring root/container config.

schogges (approved):
LGTM 👍

PR #3796, opened 2025-04-25
feat(mesh): add info to activate meshServices

### LINKED ISSUE #3161: Add hints for activation of MeshServices
### Description

In case `meshService` is set to `"Disabled"` (which is default), we should display hints on how to get them activated (including docs link). Indicator to that is when there are no services listed in the mesh service type tabs.

schogges:
In case `meshServices` mode has not been configured or set to `Disabled`, we want to inform the users to migrate to `MeshService`s. Includes a link to [meshservice/migration](https://kuma.io/docs/2.10.x/networking/meshservice/#migration).

Closes #3161

### DIFF
--- a/packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
+++ b/packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
@@ -29,6 +29,15 @@
             path="meshes.notifications.mtls-warning"
           />
         </XNotification>
+        <XNotification
+          :notify="!mesh.meshServices.mode || mesh.meshServices.mode === 'Disabled'"
+          :uri="`meshes.notifications.mesh-service-activation:${route.params.mesh}`"
+          variant="info"
+        >
+          <XI18n
+            path="meshes.notifications.mesh-service-activation"
+          />
+        </XNotification>
         <XLayout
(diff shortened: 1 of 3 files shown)


packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
```diff
@@ -29,6 +29,15 @@
             path="meshes.notifications.mtls-warning"
           />
         </XNotification>
+        <XNotification
+          :notify="!mesh.meshServices.mode || mesh.meshServices.mode === 'Disabled'"
```
johncowen, line 33: `mesh.meshServices.mode` is always one of our states right?

Screenshot 2025-04-25 at 16 09 47

https://github.com/kumahq/kuma-gui/blob/master/packages/kuma-gui/src/app/meshes/data/Mesh.ts#L19-L19

schogges: Yes, just thought double checking doesn't hurt. Anyways I removed the check for `meshService.mode` being set. Also I formatted the e2e test file, was a little off in there 🙂

johncowen (approved):
👍

PR #3797, opened 2025-04-25
fix(notifications): fix notifications sometimes not appearing

johncowen:
In certain places a notification would appear without the text.

I noticed that the conditionals here where slightly wrong, which fixes the issue


schogges (approved):
LGTM, thanks for fixing this 👍

PR #3798, opened 2025-04-25
feat(dpp/overview): add info about transparent proxying 

### LINKED ISSUE #2472: Use transparentProxy / outbounds properties in DataPlane Overview tab to help debugging
### Description

We can tell from the HTTP API whether transparent proxy is enabled for a Dataplane. If this isn't enabled then in some cases we would expect to see some outbounds defined. There is the possibility that a user has misconfigured things/forgotten to add these in this case.

We should/could improve our messaging around this circumstance in the GUI which would help users debug/fix this.

Separately there may be other things we could do in the case that we know whether a Dataplane has transparent proxy enabled or not.

schogges:
In order to inform the user about the potential improvement to configure `transparentProxying` for the dataplane networking, this PR adds an info about this in case there is no `'feature-transparent-proxy-in-dataplane-metadata'` or `'bind-outbounds'` in `dataplaneInsight.metadata.features` list. The variant is just an `info` because it could be intentional to not configure this.
Some notes:
- I've started to integrate the new dataplane types from OAS and I hit some issues while I tried that. Some of the types are not matching the actual response, such as `dataplaneInsight.subscriptions[].connectTime`.
- Even though `networking.transparentProxying` is deprecated and will be removed soon, some users have this configured and therefore I think we should still include that in the condition of wether to show the info or not. So a simple check on `'transparentProxying' in dataplane.networking` should be enough and no need to add this to types or data-layer.

Closes #2472


packages/kuma-gui/src/types/index.d.ts
```diff
@@ -301,6 +301,14 @@ export type DataplaneOutbound = {
   tags: ServiceTags
 }
 
+export type DataplaneNetworkingTransparentProxying = {
+  redirectPortInbound?: number
+  redirectPortOutbound?: number
+  directAccessServices?: string[]
+  reachable_services?: string[]
+  ipFamilyMode?: 'UnSpecified' | 'DualStack' | 'IPv4' | 'IPv6'
+}
```
schogges, line 310: Note: from the docs there is another entry `reachable_backends`, that could be added, but it's a quite complex type and I'd rather leave it out for now until we really need it. Ideally we can get rid of these custom types as soon as we get them auto generated from oas.

schogges:
Will have to come back to this as `transparentProxying` might be deprecated in the near future. Checking with xds metadata and dataplane-insights.

schogges:
This will stay on hold until the indicator about leveraging `transparentProxying` is added to `dataplaneInsights` metadata.

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -65,13 +65,19 @@
                 error: error?.toString() ?? '',
               },
             },
+            {
+              bool: !props.data.dataplane.networking.outbounds.length && !('transparentProxying' in props.data.dataplane.networking) && !props.data.dataplaneInsight.metadata.features.includes('feature-transparent-proxy-in-dataplane-metadata'),
```
lahabana, line 69: Also needs the feature `bind-outbounds` from @lukidzi 's recent work.

packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
```diff
@@ -21,6 +21,10 @@ data-planes:
       <a href="{KUMA_DOCS_URL}/policies/mutual-tls?{KUMA_UTM_QUERY_PARAMS}" target="_blank">Learn about certificates in {KUMA_PRODUCT_NAME}</a>
     stats-not-enhanced: !!text/markdown |
       The below view is not enhanced with runtime stats (Error loading stats: **{ error }**)
+    networking-transparent-proxying: !!text/markdown |
+      The networking configuration of this Dataplane does not include transparent proxying.
```
lahabana, line 25:
```suggestion
      This data plane runs without [transparent proxying]({KUMA_DOCS_URL}/production/dp-config/transparent-proxying).
      It also has no defined outbounds, to communicate to other services in the mesh, make sure to define [outbounds]({KUMA_DOCS_URL}/production/dp-config/dpp-on-universal/)
```

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
lukidzi: Yes, we are going to introduce a new `bind-outbounds` flag. Do you need more details about it?

schogges: Thank you @lahabana and @lukidzi 🙂 I've updated the condition and include `bind-outbounds`. Since this is not implemented yet and the name of the flag might change, I'll keep an eye on this and adjust the name in case it won't match.

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -65,13 +65,19 @@
                 error: error?.toString() ?? '',
               },
             },
+            {
+              bool: !can('use transparent-proxying', props.data),
+              key: 'networking-transparent-proxying',
+              variant: 'info' as const,
```
johncowen, line 71: nit: the `as const` would be better on the end of the array on line 73 I think?

johncowen: Oh wait, I see! Moving the `as const` means all `params` and `variant`s need settingHmm I almost said maybe we should default the variants to warnings anyway?

🤔 I guess don't worry for now, it was a bit of a nit anyway kinda up to you

johncowen (approved):
LGTM! 👍 I left one flip-flop nit comment, which I don't mind about, I'm not even sure what I'd prefer myself 😆

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
schogges: They all default to `warning` already. But this one should really be an `info`. The only other way I currently see is to create a separate loop for `info`s 🤔

PR #3799, opened 2025-04-25
chore: minor X cleanup

johncowen:
I spotted some more tiny bits in our `x` module that will affect our ability to make it into a completely zero dependency module in the future.

- Missing Kongponents imports where we were depending on kongponents being available globally.
- one instance I spotted of direct usage of `@kong-ui-public/i18n`, we should try and use `XI18n` exclusively within `x`. I hope to provide a very simple but injectable instance of i18n eventually making `x` zero dependency.


schogges (approved):
👍

PR #3800, opened 2025-04-25
chore: turn x into a native vue plugin

johncowen:
Turns `x` into a more traditional Vue plugin.

We had a conversation a little while back about how we might move things into modules, and one thing that was spoken about was whether we should split up our modules form the service container.

We had two choices each with Pros and Cons:

### 1. Keep the container in the module

Pros: Easier to manage each module. Its only us using the modules and we always use the container, so pragmatically no massive point in splitting
Cons: modules now need to depend on the container and the utility functions in order to provide a container.

### 2. Have two modules, a "container module" that depends on its "vanilla module"

Pros: We have a "vanilla" module that doesn't need to depend on the container
Cons: 
- We now have 2 modules to manage and support (multiple this by the amount of modules we will need)
- This is almost like a deeper inheritance chain. Whilst the vanilla module is dependency free, we still have to depend on the container which then depends on the vanilla module, just now we have a deeper dependency tree.

I thought of a third option which is contained (sorry!! 😆 ) in this PR

### 3. Make our container install native Vue plugins and make their components injectable

Pros: 
  - We have a "vanilla" module that doesn't need to depend on the container
  - Easier to manage each module, we have one single module, not a "container only module" depending on the "vanilla module"
  - We could theoretically install any Vue module and have the ability to inject components.
  - Theoretically anyone could use `x` as we do with or without our container/component injecting.

Note: This only works if you use our container to install the plugin.

---

I had to do a little bit of moving things around to make things make a bit more sense, please ask if anything isn't clear.


packages/kuma-gui/src/app/application/index.ts
```diff
@@ -260,9 +258,6 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
         app.sources,
       ],
     }],
-
-    ...me(app),
-    ...x(app),
     ...kuma(app),
```
johncowen, line 261: I moved `me` and `x` to be part of `service-mesh` for now. They shouldn't be part of (a generic) `application` module. `kuma` ideally shouldn't be here either but I'd need to do more re-plumbing to move that, so I left that for now.

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
```diff
@@ -81,6 +81,17 @@ common:
...
+  status-badge:
+    tooltip:
+      not_available: |
+        The online or offline status of this entity is unknown
+  x-empty-state:
+    title: 'No data'
+    body: ''
+    action:
+      label: ''
+      href: ''
```
johncowen, line 94: These are specific to our application, not `x`. They could have gone in `kuma` or `service-mesh` but seeing as `kuma` already had a locales file I dumped them in there. I'm tempted to combine `kuma` and `service-mesh` at some point which would make one less module.

packages/kuma-gui/src/app/service-mesh/debug.ts
```diff
@@ -0,0 +1,21 @@
...
+          }
+          return [name, item]
+        },
+      ]
+    },
+    labels: [
+      app.componentWalkers,
+    ],
+  }],
+]
```
johncowen, line 21: We choose to add the `XCopyButtonDebug` for our application, other users of `x` might not want to do this, and we don't want to provide a debug container in `x` to keep it dependency free, so I moved this to `service-mesh`.

This file only gets included during development, I also checked a production build to make sure that `XCopyButtonDebug` isn't compiled into a prod build i.e. `XCopyButtonDebug` is tree-shaken out of `x` when its not imported anywhere.

packages/kuma-gui/src/app/vue/index.ts
```diff
@@ -45,19 +48,45 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
...
+                      return app_component(...args)
+                    }
+                    app_component(...componentWalkers.reduce((prev, item) => item(...prev), args))
+                    return proxyApp
+                  }
+                default:
+                  return target[prop]
+              }
+            },
+          })
```
johncowen, line 82: This is the "meat" of the PR.

Instead of installing components when a plugin is installed, call a function on the component definition passed in _before_ actually installing it with Vue.

This gives us the opportunity to amend it if we need to.

This is very very similar to how our containers decorators work, but instead of doing this at "container build time" we do it at "runtime/component installation time"

packages/kuma-gui/src/app/vue/index.ts
```diff
@@ -148,6 +178,22 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
...
+      labels: [
+        $.directives,
+      ],
+    }],
+    [token('application.componentWalkers'), {
+      service: () => {
+        return []
+      },
+      labels: [
+        $.componentWalkers,
```
johncowen, line 194: Naming of these is getting a bit uncomfortable 😬 we already have things called `walkers` for `routeWalkers`, so I used the same naming for now.

packages/kuma-gui/src/main.ts
```diff
@@ -48,11 +47,10 @@ async function mountVueApplication() {
           ...$,
           ...msw.TOKENS,
           ...fakeFs.TOKENS,
-          ...x.TOKENS,
         }
         return [
           ...application.services(TOKENS),
-          ...xdebug.services(TOKENS),
+          ...serviceMeshDebug.services(TOKENS),
```
johncowen, line 53: This was just because I moved the `x` debug container to be in `service-mesh` instead.

johncowen (commented):
Notes:

schogges (approved):
I'm happy to see that we can keep x free from service containers 👍 LGTM 🚀

johncowen:
I need to decide where to put the locales that came from the new x-search component before we merge this

schogges:
> I need to decide where to put the locales that came from the new x-search component before we merge this

As an iteration I want to use props and slots for the texts and remove the dependency to i18n in this component. So it might be fine to leave them where you put them now?

schogges (approved):

PR #3819, opened 2025-04-29
chore(deps): remove glob from @kumahq/kuma-gui

johncowen:
Tiny PR to remove `glob` from `@kumahq/kuma-gui`, we moved it to `@kumahq/config` a while back where its used but didn't remove it from `@kumahq/kuma-gui` where I can no longer see it being used.


schogges:
I think we use it in [`.vitepress`](https://github.com/kumahq/kuma-gui/blob/4c91c8d9432aa622321c5bb3121864d96cc5c2f8/packages/kuma-gui/.vitepress/config.ts#L17)

johncowen:
Oh so we do! Thanks 👍

PR #3820, opened 2025-04-29
chore(ci): moves lockfile-lint to @kumahq/config

johncowen:
Uses the approach added in https://github.com/kumahq/kuma-gui/pull/3818 to move lockfile-lint.

This should be enough to unblock us 😅 . Once we are unblocked I'll do a single PR to move all of them.

I actually think this is a major win for us 🎉


schogges (approved):
👍

Issue #3757, opened 2025-04-14
Add new x-search to other listings


schogges:
### Description

In order to allow filter and search on Mesh*Service listings we should add the new `XSearch` to `Mesh*ServiceListView`s.

PR that adds filtering by labels: https://github.com/kumahq/kuma/pull/12840

Update:
I've changed the title to cover all the other listings that should adopt the new `XSearch` component. In some listings we there is already an implementation of a search bar and others don't have any yet.

- [x] DPP listing
- [x] legacy service listings
- [x] Mesh*Service listings
- [x] Policies listing (replace current implementation)
- [x] Meshes listing
- [x] Zones listing
- [x] Gateways listing
- [x] RBAC listings
- [x] Zone ingresses & egresses listings

PR #3821, opened 2025-04-29
feat(legacy-services): add x-search

schogges:
Part of rolling out the new `XSearch` to our other listings, starting off with the legacy `internal` and `external` services. These are only filterable by `name`.

Part of #3757


packages/kuma-gui/src/app/external-services/views/ExternalServiceListView.vue
```diff
@@ -105,3 +120,15 @@
...
+  margin-bottom: $kui-space-70;
+}
+.search-form {
+  display: flex;
+}
+.search-field {
+  flex-basis: 310px;
+  flex: 1;
+}
+</style>
```
johncowen, line 134: Can all of this be replaced with an `XLayout` for the vertical spacing and then the rest (horizontal stuff) I don't think is doing anything is it?

packages/kuma-gui/src/app/external-services/data/index.ts
```diff
@@ -8,6 +8,27 @@ export type ExternalService = PartialExternalService & {
...
+          // at this point this would be an invalid query, i.e. `name:`
+          return acc
+        default:
+          return {
+            ...acc,
+            name: key,
+          }
+      }
+    }, {})
+  },
```
johncowen, line 31: Do you have any thoughts on how we might choose to DRY out this and future others. I know we often err on the side of not worrying about copy/pasta every so often but something tells me this is going to be very very common. Not saying we should do something yet, but interested to know if you had any ideas.

Also, is there the opportunity at some point to reuse or merge in with similar code to what's in `DataplaneOverview.search` (I believe it supports `name:thing` and `name: thing`, note the colon trailing space)

schogges: >Do you have any thoughts on how we might choose to DRY out this and future others. I know we often err on the side of not worrying about copy/pasta every so often but something tells me this is going to be very very common. Not saying we should do something yet, but interested to know if you had any ideas.

Yes, I had the same feeling. But I wasn't sure where to put it. I was thinking of adding a `data` module that could house common things of the data layer 🤔 

>Also, is there the opportunity at some point to reuse or merge in with similar code to what's in DataplaneOverview.search (I believe it supports name:thing and name: thing, note the colon trailing space)

Yes, I'd like to align all of this. Also because `name: thing` is not valid anymore. I think it's not causing any issues atm, but the highlighting of the new x-search component does not support this anymore.

packages/kuma-gui/src/app/external-services/views/ExternalServiceListView.vue
schogges: Let me give this a try 👍

schogges: I reduced it down to just setting width on the search field. And thanks for raising this, I caught another issue that I missed after the last changes to `XSearch.logic-block`. It was not wrapping and could overflow the container.

packages/kuma-gui/src/app/external-services/data/index.ts
johncowen: > Yes, I had the same feeling. But I wasn't sure where to put it. I was thinking of adding a data module that could house common things of the data layer 🤔 wdyt?

Do you remember we spoke a while back about a `resources/` module? I hit this and was reminded of this just recently in https://github.com/kumahq/kuma-gui/pull/3792. I think this could maybe fit in there as say `/resources/data/Resource.ts` `Resource.search()`.

> Yes, I'd like to align all of this. Also because name: thing is not valid anymore. I think it's not causing any issues atm, but the highlighting of the new x-search component does not support this anymore. So at the moment it's more about visuals and not functionality.

👍 Yeah I have the same feeling on alignment, so good we are both thinking about possibles.

On the `name: thing` being not valid, its "not valid" only because it doesn't currently support it. One thing I noticed when @lahabana was using it was he immediately tried doing this with a space. But you're right that its not causing any issues currently, but I do think its a valid usecase/"feature". But lets revisit this later, the question here are more to hear your thinking on this.

schogges: > > Yes, I had the same feeling. But I wasn't sure where to put it. I was thinking of adding a data module that could house common things of the data layer 🤔 wdyt?
> 
> Do you remember we spoke a while back about a `resources/` module? I hit this and was reminded of this just recently in #3792. I think this could maybe fit in there as say `/resources/data/Resource.ts` `Resource.search()`.

Hmm 🤔 but isn't `resources` is an actual resource from the API? Wouldn't that mix-up common utilities?

> > Yes, I'd like to align all of this. Also because name: thing is not valid anymore. I think it's not causing any issues atm, but the highlighting of the new x-search component does not support this anymore. So at the moment it's more about visuals and not functionality.
> 
> 👍 Yeah I have the same feeling on alignment, so good we are both thinking about possibles.
> 
> On the `name: thing` being not valid, its "not valid" only because it doesn't currently support it. One thing I noticed when @lahabana was using it was he immediately tried doing this with a space. But you're right that its not causing any issues currently, but I do think its a valid usecase/"feature". But lets revisit this later, the question here are more to hear your thinking on this.

Yeah, but even though it might be a valid use-case, we decided to use space as a separator of different filters. So I'm not sure if we should keep supporting it. The UI should reflect the filters by highlighting and validate them. But I'd also be fine to adjust the regex and allow a space right after a colon: "`foo` `namespace:bar` `zone: baz`"

johncowen: > Hmm 🤔 but isn't resources is an actual resource from the API? Wouldn't that mix-up with common utilities?

Aren't all (or at least the vast majority) resources searchable by `name` and `label`?

It could also be a utility in the `resources` module, but seeing as we only use it at the data layer it might as well go there (and keeping it out of "general utilities" makes it harder for some one to use it outside of the data layer)

> But I'd also be fine to adjust the regex and allow a space right after a colon: "foo namespace:bar zone: baz"

Yeah this is what I mean I think this regex does that:

https://github.com/kumahq/kuma-gui/blob/95261cacfa726eee398d68d56e7102cfa8d3bcd6/packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts#L10-L20

Anyway I like this discussion, but just to be clear I don't think we need to do anything here, any thing we decide would be a follow up. Lmk, then maybe we should add a couple of issues to be done based on the outcome.

packages/kuma-gui/src/app/external-services/views/ExternalServiceListView.vue
johncowen: not a blocker if this doesn't lead to something straight forwards but can you remember why this `width: 0` is needed?

https://github.com/kumahq/kuma-gui/blob/95261cacfa726eee398d68d56e7102cfa8d3bcd6/packages/kuma-gui/src/app/x/components/x-search/XSearch.vue#L284

You can imagine that if possible I'm hopeful to see if we can avoid adding any CSS here at all, that being said totally fine this being a longer term objective also.

schogges: Yes, this was required because of the layout in dataplanes-list-view (the inlined x-select). I will have a look at this separately, but it was mainly about trying to scope the changes when I implemented the component, but don't break anything. The problem that I need to solve is: let the x-search component grow as much as possible, but dont grow on it's own content.

packages/kuma-gui/src/app/external-services/data/index.ts
schogges: > > Hmm 🤔 but isn't resources is an actual resource from the API? Wouldn't that mix-up with common utilities?
> 
> Aren't all (or at least the vast majority) resources searchable by `name` and `label`?

Possibly yes, but I think that's something that can be solved quite easily via params.

> It could also be a utility in the `resources` module, but seeing as we only use it at the data layer it might as well go there (and keeping it out of "general utilities" makes it harder for some one to use it outside of the data layer)

Yeah, that's the thing. Our data layer is part of every module/feature. I think `resources` is an actual resource (looking at `/_resources` endpoint), which could lead to confusion. So in order to DRY the search a little bit and possible other common things, I think we need a different place, but make sure that this will only be used in the data layer.

> > But I'd also be fine to adjust the regex and allow a space right after a colon: "foo namespace:bar zone: baz"
> 
> Yeah this is what I mean I think this regex does that:
> 
> https://github.com/kumahq/kuma-gui/blob/95261cacfa726eee398d68d56e7102cfa8d3bcd6/packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts#L10-L20
> 
> Anyway I like this discussion, but just to be clear I don't think we need to do anything here, any thing we decide would be a follow up. Lmk, then maybe we should add a couple of issues to be done based on the outcome.

Yeah, I also like the discussion and depending on the outcome I'm happy to work on follow-ups 🙂 
I haven't reused that regex, because it doesn't reflect the current highlighting in the UI. So I'd need to adjust the regex in the x-search component to reflect that.

packages/kuma-gui/src/app/external-services/views/ExternalServiceListView.vue
johncowen: ok gotcha, let resolve this one for now and see if we can circle back at some point.

packages/kuma-gui/src/app/external-services/data/index.ts
johncowen: K lets resolve this one, shall we add an issue for potentially using the above regex everywhere at some point?

johncowen (approved):
Few good discussions here for potential follow ups, but this is looks good to me!

PR #3822, opened 2025-04-29
feat(kuma-http-api): update oas types

schogges:
Updates our types after the following changes to kuma OAS. Changes are mainly about adding `DataplaneOverview` types for endpoints `/meshes/{mesh}/dataplanes/_overview` and `/meshes/{mesh}/dataplanes/{name}/_overview`

- https://github.com/kumahq/kuma/pull/13479
- https://github.com/kumahq/kuma/pull/13483

Adoption will happen in a separate PR as it may become a larger effort. Also note that most of the dataplane types are optional and we will need to see where we have to provide fallbacks.

### DIFF
--- a/packages/kuma-http-api/index.d.ts
+++ b/packages/kuma-http-api/index.d.ts
@@ -84,6 +84,38 @@ export interface paths {
         patch?: never;
         trace?: never;
     };
+    "/meshes/{mesh}/dataplanes/{name}/_overview": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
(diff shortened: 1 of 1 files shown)


johncowen (approved):
LGTM! 🎉

schogges:
> Oh wow this is awesome, finally!! 😆
> 
> I understood right didn't I, this PR is fine as we are still using our handwritten types, but now we have these in OpenAPI we can gradually move over to them?

Yes exactly 🙂 I actually started with this already, but I immediately felt that this is too big of a change to be done within this PR. So preferably we do this in small iterations 👍

PR #3823, opened 2025-04-29
chore(ci): move all lint scripts to use resolve/bin

johncowen:
This PR does a few Makefile based things:

- Moves all `Makefile` files to include a single `index.mk`. We used to use different files to provide an abstraction over which targets were available, but we fairly recently moved to a `target: .target` approach which is far more scalable. This means we no longer need to include files individually. Instead of omitting a 'test.mk', just don't add a `test: .test/something`
- Ensures all root `Makefile` always use `target: .target` i.e. makes use of the dot targets.
- Adds a `Makefile` for `fake-api` (plus a couple of other bits)
- Lastly, mostly works on `check.mk` to use our new `resolve/bin`, aswell as moving related dependencies.

I plan to work on other areas to move more dependencies soon such as `test`/cypress things, `build`/vite things. Eventually we should have virtually nothing in devDependencies for `@kumahq/kuma-gui` which will mean we will no longer need to depend on it for development scripts/utilities for any of our (soon to be numerous) workspaces.

P.S. I noticed https://github.com/kumahq/kuma-gui/pull/3715 had already made a bit of a start on this, hope you don't mind me continuing on!


packages/fake-api/tsconfig.json
```diff
@@ -1,13 +1,17 @@
...
+    "moduleResolution": "bundler",
     "noEmit": true,
     "strict": true,
     "skipLibCheck": true,
     "rootDir": "./src",
-    "types": ["node", "cypress"]
+    "types": [
+      "vite/client",
+      "cypress"
+    ]
```
johncowen, line 14: I kinda just added things here until things worked properly, no massive idea on what I was doing.

packages/config/package.json
```diff
@@ -44,6 +44,8 @@
     "stylelint-config-html": "^1.1.0",
     "stylelint-config-recommended-scss": "^14.1.0",
     "stylelint-config-recommended-vue": "^1.6.0",
+    "stylelint": "^16.19.1",
+    "typescript": "^5.8.3",
```
schogges, line 48: We could treat `typescript` as one of those global dependencies and add it to the root manifest. Wdyt?

packages/fake-api/tsconfig.json
schogges: Hm 🤔 I think you are also referring to `module` and `moduleResolution`? It's a little weird to set `moduleResolution` to `bundler` since nothing in `fake-api` is not meant to be bundled. I feel that the correct `module` is `nodenext` 🤔

johncowen: lemme try again, I had to do this because of `import type { Plugin } from 'vite'` in `vite.ts` which typescript said didn't exist until I made these changes. brb

packages/config/package.json
schogges: Hm 🤔 `vue`'s optional peer just means that it does not auto-install it. Since everything still seems to work, I think you are right.

packages/fake-api/tsconfig.json
johncowen: Yeah I dunno, I removed it and fiddled again without any clue what I was doing and couldn't get it to `make lint` successfully. Did you want to pull this branch down and have a go? Maybe thats easier, I kinda don't mind what it uses. Otherwise should we just leave as is? lmk!

schogges: After pulling down your branch, I see what the problem is 👍 So I tried myself and went to the [docs](https://www.typescriptlang.org/tsconfig/#module). I think `module` is the more important configuration as it usually comes with some defaults for `moduleResolution`. I found `module: 'preserve'` and I think this is actually what we should set for this type of package. The reason is that we don't transpile anything and just use what it serves in our other packages (we only use typescript as a linter). ☝️ **But** `preserve` implies `moduleResolution: 'bundler'` 😄

johncowen: 😆 lemme try that one sec

johncowen: K done!

schogges (dismissed):
👍

packages/fake-api/tsconfig.json
johncowen: oh awesome! that would be even better! 👍 

I made a quick issue https://github.com/kumahq/kuma-gui/issues/3824

schogges (approved):

PR #3827, opened 2025-05-05
feat(policy-list-view): use x-search

schogges:
Replaces the current custom implementation of a search field with the new `XSearch` component and allow filtering for `name`, `namespace` and `zone` (in case it's supported by the respective policy type). I've updated the mocks to reflect the filter query.

As discussed in https://github.com/kumahq/kuma-gui/pull/3821, in order to DRY the parsing of the `search` string and as suggested I've added a `Resources.search` to `resources/data/Resource` to be shared with other modules.

Part of #3757

### DIFF
--- a/packages/kuma-gui/src/app/resources/data/Resource.ts
+++ b/packages/kuma-gui/src/app/resources/data/Resource.ts
@@ -1,0 +1,44 @@
+/**
+ * Filters should follow the rules of [kong-aip#160](https://kong-aip.netlify.app/aip/160/).
+ * 
+ * `filters` are the current defaults, but may not necessarily be used depending on the actual requirements of an API.
+ */
+const filters = {
+  namespace: 'filter[labels.k8s.kuma.io/namespace]',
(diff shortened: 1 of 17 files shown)


packages/kuma-gui/src/app/resources/data/Resource.ts
schogges, line 1: Let's discuss again wether this is in the right place now 🙂 I had some other options in mind:

- Add it to `x/resources/search`: Probably we want to leave `x` very generic and the search is probably too tightly coupled to kuma
- Add it to `kuma/services/search` as a search service for kuma
- Add it to `application/services/search` as a search service for the application

johncowen: I think leaving this in `resources` for now at leats is fine, I'd love to start adding things into the `resources` module soon anyway (things that are currently in `policies` that seem to be slowly turning into a more generic `resources` module)

In my head right now, the `kuma` module doesn't have anything data specific in it so whilst I get that this option is an option, I think I'd still stick with `resources`

PR #3828, opened 2025-05-05
feat(mesh-services): add x-search to all types of mesh-services listings

schogges:
Requires https://github.com/kumahq/kuma-gui/pull/3827

Adds `XSearch` to each type of Mesh*Service listings. I've updated the mocks to reflect the filter options:
- `MeshServices`: `name`, `namespace`, `zone`
- `MeshExternalServices`: `name`
- `MeshMultiZoneServices`: `name`, `namespace`, `zone`

Part of #3757


johncowen (approved):
LGTM!

Two failing tests but I took a quick look and they are both known flakes.

Issue #3794, opened 2025-04-24
Uses `resources` in `/mesh-insights` and `/mesh-insights/:mesh`


johncowen:
### Description

See title/parent

![Image](https://github.com/user-attachments/assets/ff93c59e-0422-4737-8add-901792a78a5c)

PR #3846, opened 2025-05-06
chore(meshes): moves mesh-insights to use resources

johncowen:
Uses `resources` field instead of `policies` field for calculating the policy count similar to https://github.com/kumahq/kuma-gui/pull/3792

Closes https://github.com/kumahq/kuma-gui/issues/3794


schogges (approved):
👍

PR #3847, opened 2025-05-07
feat(x-search): add validation

### LINKED ISSUE #3787: Validation on filters in filter-bar
### Description

In order to provide feedback to the user in case the entered filters are not compatible with our API, we should validate and highlight wrong `key:value` pairs.

Some thoughts about potential filter validations:

- filtering for fields which have enumerated values (i.e. `state:available`, `state:unavailbale`, ...)
- filter for whitelisted/supported fields, so `foo:bar` may be highlighted as not valid because we don't have a field `foo`
- filter for labels must meet the requirements of [k8s/labels/syntax](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

schogges:
_TBD_

Closes #3787


schogges:
@johncowen I think this is ready to go now. After https://github.com/kumahq/kuma-gui/pull/3902 I first thought it would be nice if we can reuse the highlight regex also for validation, but I realized that `highlight` and validation needs to be separated. For example we still want to highlight `key:` because the user might not be done at this point and keeps writing, i.e. in this case it's a correct path. Validation happens at a different time, when the user submits. At this time the same pattern becomes invalid.

johncowen:
Could you double check the PR preview for me? I'm not sure if I'm seeing whats expected, I can't seem get it to invalidate?

schogges:
> Could you double check the PR preview for me? I'm not sure if I'm seeing whats expected, I can't seem get it to invalidate?

The validation is currently only added to the `DataplaneListView`. I checked the preview and it does work 🤔

johncowen:
Oh! 🤦 , cool gotcha lemme check there

johncowen:
we just had a quick sync on this we are gonna:

- use the current validation as a good default so we don't have to set it everywhere
- move the rendering of the warnings to the default for the warning slot inside the component, perhaps temporarily, we might come back and redesign something later, but putting it inside in one place for now is fine even though we have i18n things in that.

Once thats done I think this is good to go as it should automatically work everywhere once we've done the above

johncowen (approved):
Lets get this one in also. I have a few questions/opinions on a few things here, but its nothing that can't be sorted through later on (or maybe I end up being convinced otherwise)

Thats for the patience with me on this one 👍

PR #3848, opened 2025-05-07
chore(ci): dynamic matrices

johncowen:
Automatically discover packages that need CI testing.

- adds several new make targets to:
  - print a JSON of all packages (i.e. workspaces plus the root package)
  - print a JSON of all workspaces (i.e. minus the root package)
  - print a JSON of workspaces that have `.feature` files
  - print a JSON of workspaces that have `.spec.ts` files
- using the above targets, we can figure out the matrices for each of our `cli-tests` and `browser-tests` jobs, plus pass _all_ packages to `lint-tests`

All in all this is prep work I've been wanting to do for preparing to split all our modules out into separate packages. This PR means that we can just add new ones and they'll all automatically be picked up by CI. i.e. zero config.

There might be a tiny bit of tweaking/renaming or something at some point in the future, but I'd like to get this in as the `.slug` property we have added here is useful for naming uploadable artifacts that can't have `/`s in their names.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,6 +34,9 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       spec_groups: ${{ steps.set-groups.outputs.result }}
+      packages: ${{ steps.meta.outputs.packages }}
+      unit: ${{ steps.meta.outputs.unit }}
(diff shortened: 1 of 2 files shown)


schogges (approved):
LGTM 👍

PR #3849, opened 2025-05-08
[WIP] chore(ci): see if we can use outputs for with

johncowen:
TBD


johncowen:
Nope, at least  not straight forwards anyway 😢

PR #3851, opened 2025-05-08
chore(ci): resolve some more executables and move deps around

johncowen:
Moves more `@kumahq/config` executables to use node resolution. There shouldn't be anymore now that don't use node resolution.

Also moves around some dependencies as a result of that. I was actually hoping to move more, but things like `cypress` are used as executables _and_ imported via `cypress.config.js`.

I noticed that MacOS `Make` and CI (Ubuntu's) `Make` is slightly different, hence the addition of `GNUMAKEFLAGS` to prevent CI from printing out directory changes, which upsets the Make targets we are using to print paths etc.

---

I also noticed a vite config that I don't think we need, and as it looked strange I removed it. It comes from here https://github.com/kumahq/kuma-gui/pull/581 and I've no idea why it was required, plus since then we've totally changed out test setup, so I'm guessing we don't need it anymore.


schogges:
> Moves more `@kumahq/config` executables to use node resolution. There shouldn't be anymore now that don't use node resolution.

I think there is [one more](https://github.com/kumahq/kuma-gui/blob/4b9d579d6ff464138fce2f6adb89c2f41c73849d/packages/kuma-http-api/Makefile#L32) for `openapi-typescript`. Do you want to change that also to use `resolve/bin`? 🙂

johncowen:
Oh! Good spot, I just tweaked that one also 👍

schogges (approved):
👍

PR #3852, opened 2025-05-09
chore: alters vue service container to require an app argument

johncowen:
Changes `vue` service container so that the `app` service returns a function that requires the result of `createApp()` passed to it, instead of doing that itself.

This means that you have access to the result of `createApp` outside of the service contain and you don't need to build the container to get a reference to it.

It also makes it clearer where `createApp` is being called.

---

We'd already vaguely planned on reconfiguring it like this a very long time ago (see the temporary comment in the vitepress code I removed here)

I also spotted some more error wrapping code that we never use, so I removed that here also


schogges (approved):
👍

PR #3854, opened 2025-05-09
fix(tests): ensures KUMA_MESH_COUNT is 3 for control-planes test

johncowen:
We spotted a flake appearing due to changed mocks which seems to be because the mesh count can sometimes be less than 3.

This PR ensures that we always get 3 meshes in order to partially overwrite 3 meshes in the response


schogges (approved):
👍

Issue #3673, opened 2025-03-17
Migrate CodeBlocks in detail views to use YAML/Kubernetes dropdown


johncowen:
### Description

Until we do this https://github.com/kumahq/kuma-gui/issues/1967 is still an issue

PR #3855, opened 2025-05-09
chore(x): reduce "internal dependencies" in x

johncowen:
Removes the majority of "internal dependencies" of `x`.

"internal dependencies" meaning any of our own dependencies of `x` such as `@/app/application`.

3rd party dependencies (for the moment at least) we are fine with.

Note: The last YAML dependency we have here could either use `js-yaml` straight, or by just removing the `ResourceCodeBlock` component entirely. I'd rather do the latter seeing as we have https://github.com/kumahq/kuma-gui/issues/3673 as an issue already

Also, this one is probably a good "Hide Whitespace" in GH one.


schogges (approved):
👍

PR #3856, opened 2025-05-09
refactor(x-search): use useResizeObserver

schogges:
We hit a flake a couple of times `ResizeObserver loop completed with undelivered notifications` which could be the result of not cleaning up the resize observer we register in `XSearch.onMounted`. While thinking of creating a custom hook to prevent us running into this case again, we decided to leverage one of `@vueuse`'s exported hooks `useResizeObserver` which should already include everything we need instead of reinventing the wheel.

### DIFF
--- a/packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
+++ b/packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
@@ -107,7 +107,8 @@
 </template>
 
 <script setup lang="ts">
-import { onMounted, ref } from 'vue'
+import { useResizeObserver } from '@vueuse/core'
+import { ref } from 'vue'
 
 const props = withDefaults(defineProps<{
   /**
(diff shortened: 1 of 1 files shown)


johncowen (approved):
Still in two minds on this one 😆 😅 

Having looked at the vueuse useResizeObserver code, I wonder if its that that is reinventing the wheel rather than use just using the native code and calling `disconnect` which is what was missing here.

Having said that, there's something (large) to be said for making it hard to forget to call that though (i.e. making the need to call it unnecessary), we both totally forgot about that, just means that we leave native JS behind and wrap everything in Vue specific stuff 🤔 

Anyway, this is fine for now and maybe always, just still a little thoughtful on it

schogges:
> Still in two minds on this one 😆 😅
> 
> Having looked at the vueuse useResizeObserver code, I wonder if its that that is reinventing the wheel rather than use just using the native code and calling `disconnect` which is what was missing here.
> 
> Having said that, there's something (large) to be said for making it hard to forget to call that though (i.e. making the need to call it unnecessary), we both totally forgot about that, just means that we leave native JS behind and wrap everything in Vue specific stuff 🤔
> 
> Anyway, this is fine for now and maybe always, just still a little thoughtful on it

Yeah you are right 🙂 probably what I missed to say was in order to make it reusable and don't forget to cleanup the next time we use `ResizeObserver` we could create this hook ourselves. But instead of doing that and reinventing the wheel, we could use `@vueuse`'s one 🙂

PR #3857, opened 2025-05-09
feat(mesh-list-view): add x-search

schogges:
Adds `XSearch` to `MeshListView` in order to let users filter meshes by `name`.

Part of #3757


johncowen (approved):
👍

PR #3859, opened 2025-05-12
feat(gateways): add x-search

schogges:
Adds `XSearch` to both `BuiltinGatewayListView` and `DelegatedGatewayListView`. User can filter

- builtin gateways by `name` and `zone`
- delegated gateways by `name`

Part of #3757

### DIFF
--- a/packages/kuma-gui/src/app/gateways/views/BuiltinGatewayListView.vue
+++ b/packages/kuma-gui/src/app/gateways/views/BuiltinGatewayListView.vue
@@ -6,6 +6,7 @@
       size: Number,
       mesh: '',
       gateway: '',
+      s: '',
     }"
     v-slot="{ route, t, can, me, uri }"
   >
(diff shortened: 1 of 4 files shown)


packages/kuma-gui/src/app/gateways/views/BuiltinGatewayListView.vue
```diff
@@ -137,3 +150,8 @@ import { sources } from '../sources'
 import AppCollection from '@/app/application/components/app-collection/AppCollection.vue'
 import SummaryView from '@/app/common/SummaryView.vue'
 </script>
+<style lang="scss" scoped>
+.search-field {
+  width: 100%;
+}
+</style>
```
johncowen, line 157: Am I right in thinking the only place where we don't use this is in Dataplanes where we have the additional `Type` menu?

johncowen (approved):

packages/kuma-gui/src/app/gateways/views/BuiltinGatewayListView.vue
schogges: Yeah, but also in this place we need to give it some space. I want to have a look at this separately on how we could leverage `XLayout` to provide the info to let the component grow to fill the space horizontally, but respect additional inline elements. For the moment this was the easiest way to achieve the expected results 🤔 
In the end I want to remove all the added `style` sections again 🙂

johncowen: Gotcha, yeah totally on the same page here. Not a blocker at all, and you're totally right to look at this separately 👍

PR #3861, opened 2025-05-12
chore(ci): turn Makefile meta targets into dot targets

johncowen:
I figured we probably want these targets as "dot targets" so we can overwrite them if we need to.

I also exposed the entire package.json to CI instead of just the name, path and 'slug'. This means that if we want to add any meta data to our package.json's just for CI we can.

I also added a `meta/preview` based on whether a workspace has a `index.html` that we don't use as yet, but I would love to be able to deploy multiple PR preview sites based on this.

Just to note, there may be a little more related work in a few days on top of this, I'm just taking little nibbles at this until I get to a point where I can pull it all together.


schogges (approved):
LGTM 👍

PR #3880, opened 2025-05-13
feat(x-search): simplify filter by labels

schogges:
At this point we mainly support filtering by name, namespace and zone. The majority (if not all) API endpoints also support filtering by labels, but it requires a certain format of `?filter[labels.<key>]=<value>`. To make this easier and more accessible (users shouldn't need to care about this format) this PR turns any `key:value` pair, that is not caught on any short filter (namespace or zone), into the format of `filter[labels.<key>]:<value>`, which is then further used and send as a query parameter.


johncowen (approved):
👍

PR #3884, opened 2025-05-14
fix(kongponents): cast components to Component for KRadio issue

johncowen:
Adds a temporary fix to unblock https://github.com/kumahq/kuma-gui/pull/3883

There's work ongoing upstream so not totally sure if this will get in or be closed depending on the upstream work, lets see!

https://github.com/Kong/kongponents/issues/2753


schogges (approved):
👍

PR #3891, opened 2025-05-14
chore: simplifies data-source

johncowen:
TBD


packages/kuma-gui/src/app/application/services/data-source/index.ts
```diff
@@ -61,74 +69,10 @@ export const useUri = () => {
...
-          }
-        }
-      }
-    }, config)
-  }
-}
-export type Source = ReturnType<typeof getSource>
-
-// its fine to not wait for an unfocussed tab for promise returning sources
-const source = getSource(new (class extends EventTarget { hidden = false })())
```
johncowen, line 131: All of this red is what I wanted to get rid of

packages/kuma-gui/src/app/application/services/data-source/index.ts
```diff
@@ -140,30 +84,55 @@ export const create: Creator = (src, router) => {
       size: parseInt(queryParams.get('size') || '0'),
       page: parseInt(queryParams.get('page') || '0'),
       search: queryParams.get('search') || '',
-      cacheControl: ['no-store', 'no-cache'].reduce((prev, item) => queryParams.has(item) ? item : prev, queryParams.get('cacheControl') ?? ''),
+      cacheControl: [
+        'no-store',
+        'no-cache',
+        'immutable',
```
johncowen, line 90: an `immutable` `cacheControl` means we will only ever request the datasource once throughout the entire lifetime of the application. If you request it again, it will always use the value in the cache

packages/kuma-gui/src/app/application/services/data-source/index.ts
```diff
@@ -140,30 +84,55 @@ export const create: Creator = (src, router) => {
...
+      const cb = () => resolve(cb)
+      target.addEventListener(event, cb, { once: true })
+    })
+  }
+
+  return new CallableEventSource(async function* () {
+    let retry = 0
+    const maxRetry = 3
+    // @TODO: enable retries based on certain errors
+    const enableRetries = false
```
johncowen, line 107: I wanted to add the retry stuff in this PR but not enable it as yet

packages/kuma-gui/src/app/data-planes/sources.ts
```diff
@@ -34,46 +34,12 @@ export type MeshGatewayDataplaneSource = DataSourceResponse<MeshGatewayDataplane
...
-        }
-      }, {
-        retry: (e) => {
-          if (e instanceof OfflineError) {
-            return new Promise((resolve) => setTimeout(resolve, 1000))
-          }
-        },
-      })
-    },
-
```
johncowen, line 76: These were all used only in onboarding, which we no longer have

packages/kuma-gui/src/app/zone-ingresses/sources.ts
```diff
@@ -17,28 +17,8 @@ export type ZoneIngressOverviewCollectionSource = DataSourceResponse<ZoneIngress
...
-          throw new OfflineError()
-        }
-      }, {
-        retry: (e) => {
-          if (e instanceof OfflineError) {
-            return new Promise((resolve) => setTimeout(resolve, 2000))
-          }
-        },
-      })
-    },
```
johncowen, line 41: These were all used only in onboarding, which we no longer have

packages/kuma-gui/src/app/zones/sources.ts
```diff
@@ -25,27 +25,6 @@ export const sources = (source: Source, api: KumaApi) => {
...
-          throw new OfflineError()
-        }
-      }, {
-        retry: (e) => {
-          if (e instanceof OfflineError) {
-            return new Promise((resolve) => setTimeout(resolve, 2000))
-          }
-        },
-      })
-    },
```
johncowen, line 48: These were all used only in onboarding, which we no longer have

johncowen (commented):
Notes:

packages/kuma-gui/src/app/application/services/data-source/index.ts
```diff
@@ -140,30 +84,55 @@ export const create: Creator = (src, router) => {
...
-    let inited = false
-    const eventSource = init instanceof CallableEventSource
-      ? init
-      : source(() => {
-        if (inited) {
-          return Promise.resolve(route.route(params))
+
+  const waitForEvent = (target: EventTarget, event: string) => {
+    return new Promise((resolve) => {
+      const cb = () => resolve(cb)
```
schogges, line 98: nit: why are we passing `cb` as a param to `resolve`?

packages/kuma-gui/src/app/application/services/data-source/index.ts
```diff
@@ -140,30 +84,55 @@ export const create: Creator = (src, router) => {
...
-    eventSource.url = src
-    return eventSource
-  } catch (e) {
-    return source(() => Promise.reject(e))
-  }
+        break
+      } catch (e) {
+        if (enableRetries) {
+          if (retry < maxRetry) {
+            await new Promise(resolve => setTimeout(resolve, 3000 * Math.pow(++retry, 1)))
```
schogges, line 124: nit: Do we need `Math.pow` with an exponent of `1` here? The result is always what `++retry` is? 🤔

schogges (commented):
Just a few nits, the rest LGTM 🙂

packages/kuma-gui/src/app/application/services/data-source/index.ts
johncowen: Oh! I had this configurable previously, with a default of `1`. I guess this is still a little up in the air as to whether we want to do this or not, and then if we do whether we need this to be configurable (in which case default would be `1`).

I almost didn't add this code but thought it would be worth leaving for future convo and playing with. At the moment all of this is statically disabled.

packages/kuma-gui/src/app/application/services/data-source/index.ts
```diff
@@ -140,30 +84,55 @@ export const create: Creator = (src, router) => {
...
-      })
-    eventSource.url = src
-    return eventSource
-  } catch (e) {
-    return source(() => Promise.reject(e))
-  }
+        break
+      } catch (e) {
+        if (enableRetries) {
+          if (retry < maxRetry) {
```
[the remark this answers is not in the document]
johncowen: Honestly, I'd rather not combine a feature flag with a logical condition for the sake of reducing nesting here.

Its not huge nesting, and if/when we decide whether want automatic retries everywhere the feature flag will go and possible the retry logic itself will also go.

Happy to remove if you feel strongly, but as its a nit and the above, I don't think its worth it, lmk!

packages/kuma-gui/src/app/application/services/data-source/index.ts
schogges: Ah ok 🙂 But still it'd be weird to me to let the retry timeout grow exponentially 🤔

packages/kuma-gui/src/app/application/services/data-source/index.ts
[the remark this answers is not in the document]
schogges: Gotcha 👍  Yeah, it's probably a personal preference also 🙂 fine by me either way, going to resolve 👍

packages/kuma-gui/src/app/application/services/data-source/index.ts
johncowen: "exponential backoff" is a common-ish thing, but we can definitely chat about whether we want to do this sort of thing, right now its totally disabled but there so either of us can experiment easily.

schogges: Had a chat about it and we'll leave it in there for documentation purposes and make it easier to experiment 👍 didn't know about that "exponential backoff" 🙂

schogges (approved):

johncowen:
We just cut a release branch so I'm taking this out of draft, will merge pretty soon

PR #3892, opened 2025-05-15
fix(ci): use crossplatform -maxdepth for meta/preview

johncowen:
`-depth` seems to work differently on macOS, `-maxdepth` seems to be the more crossplatform/correct thing to use for restricting `find`s depth 🤦 

This doesn't affect us here yet, which explains why this issue was non-blocking, but I found it when trying things out elsewhere.

## Before 

Screenshot 2025-05-15 at 10 26 23

### After

Screenshot 2025-05-15 at 10 27 11


schogges (approved):
👍

PR #3894, opened 2025-05-16
chore(docs): improve `can`s docs slightly

johncowen:
We were discussing `can` and its usages, and we figured a better way to explain case when you would use it, so we worked on the docs whilst we thought about it to improve that a little.


schogges (approved):
👍

PR #3895, opened 2025-05-16
chore(docs): adds some colouring to our Make errors

johncowen:
If you try and run any of our Make targets without the correct version of node installed it now looks like this:

Screenshot 2025-05-16 at 17 25 28


schogges (approved):
👍

PR #3896, opened 2025-05-19
feat(dataplane-summary): show inbounds and outbounds in structured

schogges:
Brings back `Inbounds` and `Outbounds` section in summary structured view.
After removing `inbounds` in #3267 and having it only shown as part of the config in yaml, we introduced a select to switch the view (`structured`, `universal`, `kubernetes`). Now that it's up to the user to decide which type of view is preferred, the `structured` view should show more info again.

I've partially integrated the dataplane types from OAS.


packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -19,29 +20,38 @@ export type DataplaneInbound = PartialDataplaneInbound & Connection & {
...
-  inboundAddress: string
-  inbounds: DataplaneInbound[]
-  outbounds: DataplaneOutbound[]
-  type: 'sidecar' | 'gateway'
-}
+
 const DataplaneOutbound = {
-  fromObject(item: PartialDataplaneOutbound): DataplaneOutbound {
+  fromObject(item: PartialDataplaneOutbound) {
+    const address = item.address ?? '127.0.0.1'
```
johncowen, line 29: Do we know that `127.0.0.1` can be considered a default?

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -19,29 +20,38 @@ export type DataplaneInbound = PartialDataplaneInbound & Connection & {
...
+    const portStringified = item.port?.toString()
```
johncowen, line 30: if `item.port` is undefined can we call `toString()` on it?

packages/kuma-gui/src/test-support/mocks/src/meshes/_/dataplanes/_overview.ts
```diff
@@ -101,7 +101,8 @@ export default ({ fake, pager, env }: EndpointDependencies): MockResponder => (r
                 outbound: Array.from({ length: outboundCount }).map((_, _i) => {
                   return {
                     port: fake.internet.port(),
-                    tags: fake.kuma.tags({ service }),
+                    tags: fake.kuma.tags({ service: fake.word.noun() }),
+                    address: fake.internet.ipv4(),
```
johncowen, line 105: If address and port are potentially undefined, should we reproduce that here?

johncowen: Oh also, can you only use ipv4 for outbounds? I would have thought these can be ipv6 also?

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
schogges: Yes, from the [docs](https://kuma.io/docs/2.10.x/production/dp-config/dpp/#definitions) (scroll down to `Outbound.address` or search for `Properties kuma.mesh.v1alpha1.Dataplane.Networking.Outbound`, unfortunately can't deeplink).

> IP on which the consumed service will be available to this data plane proxy. On Kubernetes, it's usually ClusterIP of a Service or PodIP of a Headless Service. Defaults to 127.0.0.1

johncowen: Ah! Ok so I'm guessing that the typescript definitions say that this value is potentially undefined?

If so, is this something that it would be better for us to fix up in the backend do you think? If so should we create an issue over in kuma?

schogges:
> I had a couple of questions but generally looks good to me!
> 
> One visual thing I thought, and maybe not a blocker here, but maybe we can space the `bounds out a little or somehow make them more visually separate. When there are multiple its a little hard to figure out where one ends and another starts, wdyt?

Yeah I agree, when scrolling through one could easily oversee where `Outbounds` start 🤔 as a suggestion I've added `size: large` to `XLayout`. Wdyt? I'll remove it in case that's not what we want 🙂

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
schogges: The current typescript definitions say that all properties are potentially `undefined`. I feel that we can generally improve on them 🤔

schogges: I've created an issue: https://github.com/kumahq/kuma/issues/13578

johncowen: Nice 👍

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -19,29 +20,38 @@ export type DataplaneInbound = PartialDataplaneInbound & Connection & {
...
       ...item,
-      name: item.tags['kuma.io/service'],
-      service: item.tags['kuma.io/service'],
-      protocol: item.tags['kuma.io/protocol'] ?? 'tcp',
+      tags,
+      name: tags['kuma.io/service'],
+      service: tags['kuma.io/service'],
+      protocol: tags['kuma.io/protocol'] ?? 'tcp',
+      address,
+      addressPort: `${address}${portStringified?.padStart(portStringified.length + 1, ':')}`,
```
johncowen, line 39: Is all of this just 

```javascript
`${address}${item.port ? `:${item.port}` : ''}`
```

?

schogges: Yes, can be expressed like that too 🙂 I just tried to avoid inline ternary. Anyways looks like you'd prefer that and I'm fine either way.

johncowen (approved):
We spotted a tiny unrelated issue with search here that we are going to work on separately but this LGTM!

PR #3897, opened 2025-05-19
chore: untrack cypress downloads

schogges:
In https://github.com/kumahq/kuma-gui/pull/3880/files I've accidentally checked in my `cypress/downloads/downloads.html`, which I think should not be tracked. Actually I think no file in `cypress/downloads` should be tracked.

https://docs.cypress.io/app/references/configuration#Folders--Files

| Option | Default | Description |
| ---- | ----| ---- |
| downloadsFolder | cypress/downloads | Path to folder where files downloaded during a test are saved. |


johncowen (approved):
oh nice spot, thanks!

PR #3898, opened 2025-05-19
chore(ci): adds and uses bootstrap action

johncowen:
TBD


.github/actions/bootstrap/action.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      if: |
+        !contains(github.event.pull_request.labels.*.name, 'ci/cache-ignore') && inputs.install-cmd != ''
+      with:
+        path: |
+          **/node_modules
+          /home/runner/.cache/Cypress
+        key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+
+    - if: steps.node-modules-cache.outputs.cache-hit != 'true' && inputs.install-cmd != ''
```
schogges, line 24: Mssing `id: node-modules-cache` on the cache step

.github/actions/bootstrap/action.yaml
schogges, line 1: I'm wondering if we use the correct caching strategy 🤔 

1. As a first step (after setup node) I think we should check if the cache is there based on the key. Could also use `lookup-only` for this.
2. Install dependencies if there is no cache hit (this looks correct so far)
3. Cache if there was no hit before
4. No need to restore here. I think we only need to restore in the workflows

johncowen: We totally reconfigured the caching for all the little edge cases, and made it much more explicit and probably made it work as it needs to 😆

schogges (approved):
👍

PR #3902, opened 2025-05-19
refactor(x-search): improve UX

schogges:
Changes the regular expression that highlights the different search values (filters) to be more user friendly.
It now allows to have a space after a `key:` => `foo: bar` and also typing `=` instead of `:` is allowed. Still a `=` is being converted back to `:` behind the scenes.

Generally this aligns more with the current behaviour of dataplane search. In another follow-up I'll remove the implementation in dataplanes and connect it to `Resource.search`.

In order to keep track of the regex in just one place, I've created a mandatory prop for `XSearch`: `searchRegex`.
It's being tracked next to and used in `Resource.search`.

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -30,6 +30,7 @@
           >
             <XSearch
               class="search-field"
+              :search-regex="searchRegex"
               :keys="['name', 'tag', 'zone', 'namespace']"
               :value="route.params.s"
               @change="(s) => route.update({ page: 1, s })"
@@ -311,6 +312,7 @@ import AppCollection from '@/app/application/components/app-collection/AppCollec
 import StatusBadge from '@/app/common/StatusBadge.vue'
 import SummaryView from '@/app/common/SummaryView.vue'
 import type { Mesh } from '@/app/meshes/data'
+import { searchRegex } from '@/app/resources/data/Resource'
 const props = defineProps<{
   mesh: Mesh
 }>()
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -30,7 +30,7 @@
           >
             <XSearch
               class="search-field"
-              :search-regex="searchRegex"
+              :highlight="searchRegex"
               :keys="['name', 'tag', 'zone', 'namespace']"
               :value="route.params.s"
               @change="(s) => route.update({ page: 1, s })"
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -30,7 +30,6 @@
           >
             <XSearch
               class="search-field"
-              :highlight="searchRegex"
               :keys="['name', 'tag', 'zone', 'namespace']"
               :value="route.params.s"
               @change="(s) => route.update({ page: 1, s })"
@@ -312,7 +311,6 @@ import AppCollection from '@/app/application/components/app-collection/AppCollec
 import StatusBadge from '@/app/common/StatusBadge.vue'
 import SummaryView from '@/app/common/SummaryView.vue'
 import type { Mesh } from '@/app/meshes/data'
-import { searchRegex } from '@/app/resources/data/Resource'
 const props = defineProps<{
   mesh: Mesh
 }>()
--- a/packages/kuma-gui/src/app/resources/data/Resource.spec.ts
+++ b/packages/kuma-gui/src/app/resources/data/Resource.spec.ts
@@ -3,112 +3,115 @@ import { describe, expect, test } from 'vitest'
 import { Resource } from './Resource'
 
 describe('Resource.search', () => {
-  describe('commons', () => {
-    test.each([
+  test.each([
     // Name
-      [
-        'foo',
-        {
-          name: 'foo',
-        },
-      ],
-      [
-        'name:foo',
-        {
-          name: 'foo',
-        },
-      ],
+    [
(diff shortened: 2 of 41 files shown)


packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -14,35 +15,45 @@ type SearchOptions = {
...
+
 export const Resource = {
   search(query: string, options: SearchOptions = {}) {
     const { defaultKey = 'name' } = options
-    const parts = query.trim().split(/\s+/)
+    const parts = query.trim().split(searchRegex).map((part) => part?.trim()).filter(Boolean)
 
     return parts.reduce((acc, curr) => {
-      const [key, value] = curr.split(/:(.*)/)
+      const [key, value] = curr.split(/:(.*)/).map((item) => item.trim().replace(/=/, ':'))
```
johncowen, line 26: The equals to colon logic needs checking as we saw currently its not quite right

schogges: Addressed 👍 all `=` in the whole query stream is now being replaced with `:`.

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -30,6 +30,7 @@
           >
             <XSearch
               class="search-field"
+              :search-regex="searchRegex"
```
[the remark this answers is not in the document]
schogges: Done 🙂

packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -14,35 +15,45 @@ type SearchOptions = {
...
+            [defaultKey]: value?.length ? value : key,
+          }
+        case ['label', 'labels'].includes(key): {
+          const [k, v] = value.split(/:(.*)/)
+          return {
+            ...acc,
+            [`filter[labels.${k}]`]: v,
+          }
+        }
+        case key?.length && !isShortFilter(key):
```
johncowen, line 43: Question: is the type guard/narrowing doing anything here? i.e. why do we need to narrow it from a string down to a specific string value?

packages/kuma-gui/src/app/resources/data/Resource.spec.ts
```diff
@@ -0,0 +1,114 @@
+import { describe, expect, test } from 'vitest'
+
+import { Resource } from './Resource'
+
+describe('Resource.search', () => {
+  describe('commons', () => {
```
johncowen, line 6: nit: I don't think we need a double `describe` here, see comment below also

packages/kuma-gui/src/app/resources/data/Resource.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+      ],
+      [
+        'foo:',
+        {},
+      ],
+      [
+        ':foo',
+        {},
+      ],
+    ])('%s works', (search, expected) => {
```
johncowen, line 92: if we change the message of the `it` here to be something like `searching by %s works` I think we can drop the double `describes`.

packages/kuma-gui/src/app/resources/data/Resource.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    for(const defaultKey in ['foo', 'bar']) {
+      test.each([
+        // Name
+        [
+          'foo',
+          {
+            [defaultKey]: 'foo',
+          },
+        ],
+      ])('%s works', (search, expected) => {
```
johncowen, line 108: `using %s as the defaultKey works`  means we can drop the double `describes`

packages/kuma-gui/src/app/resources/data/Resource.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+        {},
+      ],
+    ])('%s works', (search, expected) => {
+      const actual = Resource.search(search)
+      expect(actual).toStrictEqual(expected)
+    })
+  })
+
+  describe('defaultKey', () => {
+    for(const defaultKey in ['foo', 'bar']) {
```
johncowen, line 99: I think its a bit weird that we are using a `test.each` and a `for` loop? Should we just make two test cases in the `test.each` by altering the args for the test a little

johncowen (commented):
Few more comments, mostly nits, but the adding a default highlight property I think we should definitely do. lmk!

packages/kuma-gui/src/app/x/components/x-search/XSearch.vue
```diff
@@ -131,6 +131,10 @@ const props = withDefaults(defineProps<{
    * The default key, that is being used to filter for when there is no `key:value` pair but only a `value`
    */
   defaultKey?: string
+  /**
+   * A regular expression that highlights different values and key:value pairs.
+   */
+  highlight: RegExp
```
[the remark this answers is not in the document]
schogges: Well, I thought about that, but decided against it because we track it in two places now and by accident we change in one place but not in the other.

packages/kuma-gui/src/app/resources/data/Resource.ts
schogges: Not sure if I understand your question 🤔 the code piece you are referring here is checking that it's not a short filter being used. This is covered in the next case block. I.e. `name:foo` would be interpreted as a labels filter because `name` is not a short filter.

schogges: Ah, gotcha 😄 actually needed it to satisfy TypeScript because doing something like

```typescript
filters[key]
```

is causing an error

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ namespace: string; zone: string; }'.
  No index signature with a parameter of type 'string' was found on type '{ namespace: string; zone: string; }'.ts(7053)
```

Therefore I need to turn the type into `key is keyof typeof filters` which I did in the type guard.

PR #3918, opened 2025-05-20
fix(x-search): internal state not updated on props.value change

schogges:
While working on https://github.com/kumahq/kuma-gui/pull/3896 we noticed that the internal state does not update in case the `:value` prop is being updated. This makes sure that when the `value` changes the internal state is being updated and the value properly displayed.


johncowen (approved):
Looks good, thanks! 👍

PR #3928, opened 2025-05-21
chore: switchover from dependabot to renovate

schogges:
Removes dependabot config and moves renovate to `.github` folder for a complete switchover.


johncowen:
shall we remove that on-dependbot-pr.yml here also? https://github.com/kumahq/kuma-gui/blob/master/.github/workflows/on-dependabot-pr.yml

we might want a similar one in the future for renovate, but lets start out with seeing what we can do with renovate on its own first

schogges:
> shall we remove that on-dependbot-pr.yml here also? https://github.com/kumahq/kuma-gui/blob/master/.github/workflows/on-dependabot-pr.yml
> 
> we might want a similar one in the future for renovate, but lets start out with seeing what we can do with renovate on its own first

Good point 👍 will remove it 🙂

johncowen (approved):

PR #3929, opened 2025-05-21
chore(tests): add `name` search test

johncowen:
I spent quite a bit of time poking our new search logic whilst reviewing and I found a search case that our new search implementation covers but our old one didn't 🎉  (old implementation isn't quite totally removed yet but thats on our list)

I figured it would be good to keep a test for this case


schogges (approved):
👍

PR #3948, opened 2025-05-22
refactor(dataplanes): use resource search

schogges:
The dataplanes listings were the last ones not using `Resource.search`. This is now changing and I've removed the old `DataplaneOverview.search`. While doing that I had to make sure that we still keep supporting filter by `tag`, i.e. in some places we still use tags from `dataplaneTags`. I still wanted to migrate as much as possible over to use `label`, which I did by using `zone:<zone>` for the dataplanes listing in `MeshDetailView`.

I've also noticed that in some places I checked for `can('use zones')` in the `:keys` for `XSearch` and in some places I didn't. This is also being fixed here.

Also I made sure that in the detail views where we have filter presets for the dpp listing, that these are removed from the `:keys` of `XSearch` to prevent users from trying to filter for different values.


johncowen (approved):
I'd love to not loose the `service:` and `protocol:` short cuts here, but I'm not sure how much effort it would be to keep those and I might be missing some finer details here.

Other than that lets get this in!

PR #3953, opened 2025-05-26
refactor(resource/search): reuse short filters for both labels and tags

schogges:
Follow-up to #3948 

Initially I thought we will drop support for `tag`s and move all over to `label`s. At least for dataplanes we need to keep supporting `tag`s for a while. This change makes sure that we can use short filters (i.e. `zone` => `kuma.io/zone`) for both `label`s and `tag`s.
To cover that I've added some more cases to the unit test and reverted the changes made in the browser tests.

### DIFF
--- a/packages/kuma-gui/src/app/resources/data/Resource.ts
+++ b/packages/kuma-gui/src/app/resources/data/Resource.ts
@@ -3,17 +3,18 @@
  * 
  * `filters` are the current defaults, but may not necessarily be used depending on the actual requirements of an API.
  */
-const filters = {
-  namespace: 'filter[labels.k8s.kuma.io/namespace]',
-  zone: 'filter[labels.kuma.io/zone]',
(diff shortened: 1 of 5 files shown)


packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -28,33 +29,34 @@ export const Resource = {
...
           }
         }
-        case key?.length && !isShortFilter(key):
-          return {
-            ...acc,
-            [`filter[labels.${key}]`]: value,
-          }
         case !!key?.length: {
-          const _key = isShortFilter(key) ? filters[key] : key
+          const _key = `filter[labels.${isShortFilter(key) ? shortFilters[key] : key}]`
```
schogges, line 59: The case of `key?.length && !isShortFilter(key)` could be dropped in favor of a simple inline ternary.

johncowen (approved):

PR #3973, opened 2025-06-04
chore(zones): add ZoneActionGroup

johncowen:
Creates a specific `ZoneActionGroup` to centralise all types of "actions" for zones.

### DIFF
--- a/packages/kuma-gui/src/app/zones/components/ZoneActionGroup.vue
+++ b/packages/kuma-gui/src/app/zones/components/ZoneActionGroup.vue
@@ -1,0 +1,28 @@
+<template>
+  <XActionGroup
+    v-bind="attrs"
+  >
+    <template
+      v-for="(_, slotName) in slots"
+      :key="slotName"
+      #[slotName]="slotProps"
+    >
+      <slot
+        :name="slotName"
+        v-bind="(slotProps)"
+      />
+    </template>
+  </XActionGroup>
+</template>
+<script lang="ts" generic="T extends { name: string } | undefined" setup>
+import { useAttrs } from 'vue'
+
+
+const slots = defineSlots()
+const attrs = useAttrs()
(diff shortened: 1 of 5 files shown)


packages/kuma-gui/src/app/zones/components/ZoneActionGroup.vue
```diff
@@ -0,0 +1,28 @@
...
+        v-bind="(slotProps)"
+      />
+    </template>
+  </XActionGroup>
+</template>
+<script lang="ts" generic="T extends { name: string } | undefined" setup>
+import { useAttrs } from 'vue'
+
+
+const slots = defineSlots()
```
schogges, line 21: I have a Q about `defineSlots`: if I'm not mistaken we could use `$slots` in the template directly without defining them in the script via `defineSlots`. At least as long as we do not type them any further. When would you use `$slots` vs `defineSlots`?

schogges (approved):

packages/kuma-gui/src/app/zones/components/ZoneActionGroup.vue
johncowen: I think in the past we've avoided using `$` global vars "because global". But personally I don't really mind either way, but at the moment we pretty consistently use `defineSlots` over `$slots`.

In the case of "non-route components" there isn't even the argument for avoiding a script tag by using `$slots`, chances are you are using `defineProps` also

johncowen: Its global is in it comes from nowhere when looking at the file in isolation. It's just there.

schogges: ah ok, understood 👍 thanks for the clarification 🙂

Issue #3730, opened 2025-04-02
flake: policies redirect


johncowen:
### Kuma Version

2.11

### Describe the bug

Looks like we might have altered the mocks somehow, I'm guessing we need to make this more stable during testing, i.e. make sure we are mocking everything thats need in the test.

FYI: I've seen this I think twice now

![Image](https://github.com/user-attachments/assets/02a8965a-25cb-4697-833f-b35905dc3f3f)

https://github.com/kumahq/kuma-gui/actions/runs/14214725939/job/39828714802

PR #3977, opened 2025-06-04
chore(ci): fix flake-y policies redirect test

johncowen:
I also switched/remove XActions `:mounted` prop to use Vue's `@vue:mounted` instead now we have that.

Closes https://github.com/kumahq/kuma-gui/issues/3730


schogges (approved):
👍

PR #3985, opened 2025-06-05
chore(renovate): schedule weekly early Mondays

schogges:
To reduce the noise we want to go back to weekly schedule as we did with dependabot.


johncowen (approved):
👍

PR #3987, opened 2025-06-06
feat(zone-detail): add env-icon to title

schogges:
In order to make it more clear to which environment the zone is deployed to, we want to show the icon of the environment in next to the name of the zone (title), which is a very prominent spot.

### DIFF
--- a/packages/kuma-gui/src/app/zones/views/ZoneDetailTabsView.vue
+++ b/packages/kuma-gui/src/app/zones/views/ZoneDetailTabsView.vue
@@ -23,13 +23,25 @@
       >
         <template #title>
           <XLayout size="small">
-            <h1>
-              <XCopyButton :text="route.params.zone">
-                <RouteTitle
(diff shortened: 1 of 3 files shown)


johncowen:
> Do you have an example of where we don't use k8s as default/fallback?

Nope, not sure if there is one. I kinda feel like if we are doing this in one place we should do it everywhere even though it feels a little weird to assume kubernetes as default everywhere in the GUI.

johncowen (approved):
Looks good, I think theres a flake here so it just needs a re-run.

schogges:
> > Do you have an example of where we don't use k8s as default/fallback?
> 
> Nope, not sure if there is one. I kinda feel like if we are doing this in one place we should do it everywhere even though it feels a little weird to assume kubernetes as default everywhere in the GUI.

Ah, understood 👍 and I think you are right that it's weird to assume we are in k8s env if nothing is set 🤔

PR #3988, opened 2025-06-06
chore: update msw worker script

schogges:
In https://github.com/kumahq/kuma-gui/pull/3936 we missed to update the worker script, causing an error. 

image

---

I'll try to add a post upgrade command to renovate, such that this won't happen again.


johncowen (approved):
🎉  thanks I was just about to look at this myself 👍

PR #4001, opened 2025-06-11
chore: pin actions (master)

schogges:
Actions need to be pinned on SHA version. We've put common actions sourced from kumahq on the allowList, but this seems to be colliding with https://github.com/kumahq/.github/blob/75f72db331df298cdb92be152a3eccce93350a67/.github/workflows/wfc_lifecycle.yml#L102

### DIFF
--- a/.github/workflows/lifecycle.yml
+++ b/.github/workflows/lifecycle.yml
@@ -15,7 +15,7 @@ permissions:
 
 jobs:
   lifecycle:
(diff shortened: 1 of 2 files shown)


schogges:
> Should we be removing this if we aren't using it anymore?
> 
> https://github.com/kumahq/kuma-gui/blob/0753d57f3cc26f4616f160b81a5d8f7dfea1457a/.github/workflows/main.yml#L31-L32

Yes, I think we can remove it 👍

johncowen (approved):

PR #4015, opened 2025-06-12
chore(vite.plugins/server): allow extending csp

schogges:
In order to allow extending the CSP when using the vite server middleware I've turned the optional `csp` parameter from a `boolean` to an object that allows extending each directive individually. For now it's only possible to extend, but not modify. Modification will follow in a separate refactoring.


johncowen (approved):
Sorry, in our chat I totoally misunderstood something, I thin kthis is good to go as is 👍

johncowen (approved):

PR #4017, opened 2025-06-12
chore(meshes): use openapi specs for HTTP API calls

johncowen:
There's a little bit of an unfortunate diff here, Hide Whitespace might help but I'll try and add some inlines:

---

I noticed that trying to use overlays for adding/merging new resources into the original OpenAPI spec wasn't quite doing what I wanted (overlays doesn't expand `$ref`s)

This PR:

- Uses `redocly` CLI to merge two OpenAPI specs instead of trying to use overlays only. We take the upstream OpenAPI and then have our own with resources what aren't present in the upstream one(in this case MeshInsight). Making our own small spec with just MeshInsight and then merging means that all `$refs` are expanded into a final OpenAPI spec with everything in.
- Changes all our app code around Meshes and MeshInsights to use the type-driven HTTP client. This means `/meshes` should now be totally using the auto-generated OpenAPI types.
- Similarly changes `/control-planes` to totally use the auto-generated types. This includes a call to `https://kuma.io` for which we fallback to use the fetch client that the type-driven client uses, seeing as `http://kuma.io` isn't in our OpenAPI specs.

Using this merging and overlays I plan on converting all of the remaining resources/types to OpenAPI:

- ServiceInsights, ExternalServices
- Gateways
- Policies
- Zones, Zone Ingress and Zone Egresses (I might be able use the same approach that we used for Dataplanes for this, i.e. use the protobuf autogenerator from upstream)
- Anything else that is left...


johncowen (commented):
Notes:

schogges (approved):
LGTM

PR #4019, opened 2025-06-13
feat(control-plane): allow nesting summary route

schogges:
In order to have nested routes in the control-plane detail view for action routes, I've added a default slot to the action group and made the routeName dynamic based on the props provided by the router.

### DIFF
--- a/packages/kuma-gui/src/app/control-planes/components/ControlPlaneActionGroup.vue
+++ b/packages/kuma-gui/src/app/control-planes/components/ControlPlaneActionGroup.vue
@@ -1,19 +1,25 @@
 <template>
-  <XActionGroup>
-    <template #control>
-      <XAction
-        action="expand"
-        appearance="primary"
-      >
-        {{ t("main-overview.action_menu.toggle_button") }}
+  <XLayout
+    type="separated"
+    justify="end"
+  >
+    <slot />
+    <XActionGroup>
+      <template #control>
+        <XAction
+          action="expand"
+          appearance="primary"
+        >
+          {{ t("main-overview.action_menu.toggle_button") }}
+        </XAction>
+      </template>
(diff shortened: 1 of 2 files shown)


packages/kuma-gui/src/app/control-planes/components/ControlPlaneActionGroup.vue
```diff
@@ -1,19 +1,25 @@
...
-    </template>
-    <XAction :to="{ name: 'hostname-generator-root-view' }">
-      {{ t("main-overview.action_menu.items.hostname_generators") }}
-    </XAction>
 
-    <slot name="actions" />
-  </XActionGroup>
+      <slot name="actions" />
+    </XActionGroup>
+  </XLayout>
```
johncowen, line 22: I think this looks great, I was about to leave this nit and approve but! I noticed something else, lemme know what you think on the question after the nit below

---

Ideally this would be the same shape as the other ActionGroups specifically the default slot is where the actions go because its an ActionGroup, so the default slot is the Actions.

It's a bit of a nit, but I think it'd be better from a consumers perspective if we can keep these *ActionGroups the same shape. To be honest I didn't even realize we had a ControlPlaneActionGroup 😅 

---

Totally separate, but I was about to say, we should probably make this component injectable instead of adding things here that are possibly not used.

And that made me wonder if we should add this sort of stuff to kuma?

https://kuma.io/docs/2.10.x/guides/federate/#connect-zone-control-plane-to-global-control-plane

So maybe we don't make it injectable and add this sort of stuff ^ here.

Wdyt? Is it exactly the same feature?

johncowen (approved):
I just kicked the tests off one more time as it had hit that "Unniversal" flake you just fixed elsewhere, once thats green this LGTM!

PR #4023, opened 2025-06-16
refactor: move cypress step-definitions to config

schogges:
Moves the cypress step-definitions over to `@kumahq/config`. I've decided to put it into the config package as we already have cypress configuration in there and it's just one more file. Generally I think we should move configs and toolings (as documented) into this package and once we see that it grows too much, we can split it further down again.

### DIFF
--- a/packages/kuma-gui/src/app/application/index.ts
+++ b/packages/kuma-gui/src/app/application/index.ts
@@ -18,11 +18,10 @@ import type { EnvVars } from './services/env/Env'
 import Env from './services/env/Env'
 import I18n from './services/i18n/I18n'
 import storage from './services/storage'
-import type { Source } from '@/app/application/services/data-source'
-import { create, destroy, getSource, DataSourcePool } from '@/app/application/services/data-source'
+import { create, destroy, DataSourcePool } from '@/app/application/services/data-source'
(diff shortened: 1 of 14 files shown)


packages/config/src/cypress/definitions.ts
schogges, line 1: this is mostly copy&paste from `packages/kuma-gui/cypress/support/step_definitions/index.ts`, export as a function and add base types for external `mock` and `client`

packages/kuma-gui/cypress/support/step_definitions/index.ts
```diff
@@ -1,352 +1,8 @@
...
-})
-
-Then(/^the "(.*)" element is empty$/, (selector: string) => {
-  $(selector).should('be.empty')
-})
-
-Then('the page title contains {string}', function (title: string) {
-  cy.title().should('contain', title)
-})
+const client = useClient()
```
johncowen, line 6: Following my very brief look at this yesterday and our very short quick chat, I didn't notice that these are actually still here (I was wondering why we didn't need to use these anymore, but we do still have them)

johncowen:
I checked the `cucumber_language_server` with this also following the query we had yesterday and all looks fine, there is just one little nit which I'll put inline in the code instead of here

PR #4036, opened 2025-06-23
chore: uses DataLoader::variant, removes #loadable

johncowen:
TBD


packages/kuma-gui/src/app/application/components/data-source/DataLoader.vue
```diff
@@ -51,22 +20,27 @@
...
-        v-if="props.loader && typeof slots.loadable === 'undefined'"
+        v-if="props.loader"
         name="connecting"
         :data="undefined"
         :error="srcError"
         :refresh="props.src !== '' ? refresh : () => {}"
       >
         <XProgress
-          v-if="props.src !== ''"
+          v-if="props.data.length > 0 || props.src !== ''"
```
johncowen, line 43: This extra `props.data.length` means XProgress is used correctly even with:

```vue
<DataLoader data="[data]" />
```

type DataLoaders, not just

```vue
<DataLoader src="..." />
```

ones

Previously, only these types were still using the loader from KTable, so all in all this changes nothing from a user perspective but fixes a little non-intentional "bug" in our code.

packages/kuma-gui/src/app/gateways/views/BuiltinGatewayDataplanesView.vue
johncowen, line 1: This file needed more refactoring than others, its probably because its a place that we don't really visit often.

I updated it to use the data that is loaded in via the Tab view, and then gets passed down into the individual views rather than reloading it again.

packages/kuma-gui/src/app/application/components/data-source/DataLoader.vue
```diff
@@ -51,22 +20,27 @@
...
+        name="default"
+        :data="srcData as NonNullable<TypeOf<T>>"
         :error="srcError"
         :refresh="props.src !== '' ? refresh : () => {}"
       />
+    </template>
+
+    <template v-else>
       <slot
-        v-if="props.loader && typeof slots.loadable === 'undefined'"
```
schogges, line 62: Q: What is the purpose of the `loader` prop? If I set it to `false` it errors because `srcData` is not ready so I think `NonNullable<TypeOf<T>>` is incorrect in this case.

schogges (commented):
One Q, the rest LGTM 👍

packages/kuma-gui/src/app/application/components/data-source/DataLoader.vue
johncowen: The loader prop (which defaults to true) is so you can tell it not to monitor/show loading.

From what I can see this is the only place we use it:

Screenshot 2025-06-24 at 11 49 00

i.e. `loader: true` (the default) show a loader, `loader: false` (the default) don't show a loader.

I think there are edge cases where we use this, I also vaguely remember there being a reason why we use `DataLoader` to show this and not just a straight `ErrorBlock` but those might be historical reasons. I'd have to go back and read to see if there is a trail of reasoning.

At the moment I'm not too concerned about having this prop, and I can't see where the error you mentioned would be, could you explain more?

PR #4037, opened 2025-06-23
refactor: replace ResourceCodeBlock with XCodeBlock

schogges:
Generally `ResourceCodeBlock` is deprecated and we want to entirely move over to `XCodeBlock`. Due to the issue described in https://github.com/kumahq/kuma-gui/issues/1967 we decided add a dropdown to select the preferred format (k8s/universal) just as how we did it in the summary views. There will be only one copy button left, instead one to copy the universal config and one to (fetch and) copy the k8s config. This also means that a user can actually see the k8s format of a config without copying it.

While working on it I saw that we still use `ResourceCodeBlock` in the summary views, even though we already implemented the dropdown. I've replaced them also with `XCodeBlock` for a full migration.

Also I've seen that in `MeshExternalSummaryView` and `MeshMultiZoneSummaryView` for the k8s config the source of the dataloader was slightly incorrect using `/.../mesh-service/:name`. I've fixed it and changed it to `/.../mesh-external-service/:name` and `/.../mesh-multi-zone-service/:name` respectively.

---

~Adding the dropdown to switch between `universal` and `k8s` format causes an issue, which is related to https://github.com/kumahq/kuma-gui/issues/3494. In case a user has a preference to `structured` (or no preference at all in which case we fall back to `structured`), the dropdown has no value set and we render the `k8s` format. This is why I added `?format=universal` in the tests. In such case of the `format` is not part of the options we could default to one of the options, but this would cause an update of the user preferences and this is not the expected UX as described in #3494.~ 👈 I've used `environment` route param for detail/config views

Closes https://github.com/kumahq/kuma-gui/issues/3673


packages/kuma-gui/features/mesh/builtin-gateways/Item.feature
```diff
@@ -7,7 +7,7 @@ Feature: mesh / builtin-gateways / item
       | listener-card | [data-testid='listener-card']                    |
       | route-card    | [data-testid='route-card']                       |
       | dataplanes    | [data-testid='data-plane-collection']            |
-      | config        | [data-testid='config']                           |
+      | config        | [data-testid='codeblock-yaml-universal']         |
```
schogges, line 10: > In case a user has a preference to `structured` (or no preference at all in which case we fall back to `structured`), the dropdown has no value set and we render the `k8s` format. This is why I added `?format=universal` in the tests.

I could change the selector to `[data-testid='codeblock-yaml-k8s']` so I don't need to add the `?format=universal` further down in this file. But as described in the PR description the UX is not ideal.

johncowen:
> Adding the dropdown to switch between universal and k8s format causes an issue, which is related to https://github.com/kumahq/kuma-gui/issues/3494. In case a user has a preference to structured (or no preference at all in which case we fall back to structured), the dropdown has no value set and we render the k8s format. This is why I added ?format=universal in the tests. In such case of the format is not part of the options we could default to one of the options, but this would cause an update of the user preferences and this is not the expected UX as described in https://github.com/kumahq/kuma-gui/issues/3494.

If I understand correctly I think we need to do what we did for subscriptions (which uses `output`, not `format`).

i.e.

- `?output`: `structured` or `yaml`
- `?format`: `structured`, or `universal` or `kubernetes`
- `?somethingelse`: `universal` or `kubernetes` - I would suggest `?environment`

Would that solve the issue you are describing?

schogges:
> > Adding the dropdown to switch between universal and k8s format causes an issue, which is related to #3494. In case a user has a preference to structured (or no preference at all in which case we fall back to structured), the dropdown has no value set and we render the k8s format. This is why I added ?format=universal in the tests. In such case of the format is not part of the options we could default to one of the options, but this would cause an update of the user preferences and this is not the expected UX as described in #3494.
> 
> If I understand correctly I think we need to do what we did for subscriptions (which uses `output`, not `format`).
> 
> i.e.
> 
> * `?output`: `structured` or `yaml`
> * `?format`: `structured`, or `universal` or `kubernetes`
> * `?somethingelse`: `universal` or `kubernetes` - I would suggest `?environment`
> 
> Would that solve the issue you are describing?

Ah 💡 thanks for reminding me of this 👍 yes, I'll go for `environment` 🙂

schogges:
@johncowen I was able to remove `ResourceCodeBlock` 👍

PR #4044, opened 2025-06-26
chore(deps/config): remove @typescript-eslint/eslint-plugin and @typescript-eslint/parser

schogges:
We are not directly depending neither on `@typescript-eslint/eslint-plugin` nor on `@typescript-eslint/parser` (we are not importing them anywhere or specifying them as plugins). Both are transitive dependencies of `@vue/eslint-config-typescript`.
I've verified that the typescript rules are still active by removing one of our overrides `'@typescript-eslint/no-explicit-any': 'off'`, which raised errors ✅


johncowen (approved):
👏  great spot!

PR #4045, opened 2025-06-27
chore(cypress): move step setup to main file

schogges:
A little cleanup after moving the step definitions to the config package. There is no need to have a separate file in the search path config.


johncowen (approved):
Nice 👍

PR #4058, opened 2025-07-02
fix: infinite rerenders during development

schogges:
I've noticed that the `ZoneDetailView` blew up in development mode due to `ApppCollection` not wrapped in `DataLoader variant="list"`
The same happened on the `DataplaneDetailView` when rendering the subscriptions listing.

Note: best to review with hidden whitespaces

### DIFF
--- a/packages/kuma-gui/src/app/application/components/data-source/DataLoader.vue
+++ b/packages/kuma-gui/src/app/application/components/data-source/DataLoader.vue
@@ -14,10 +14,12 @@
         :error="allErrors[0]"
         :refresh="props.src !== '' ? refresh : () => {}"
       >
-        <ErrorBlock
-          v-bind="$attrs"
-          :error="allErrors[0]"
-        />
+        <XCard>
+          <ErrorBlock
+            v-bind="$attrs"
+            :error="allErrors[0]"
+          />
+        </XCard>
       </slot>
     </template>
     <template
--- a/packages/kuma-gui/src/app/connections/views/ConnectionsStatsView.vue
+++ b/packages/kuma-gui/src/app/connections/views/ConnectionsStatsView.vue
@@ -16,7 +16,7 @@
       :title="t('data-planes.routes.item.navigation.data-plane-stats-view')"
     />
     <AppView>
-      <XCard>
+      <XCard v-if="props.networking">
         <DataLoader
           :src="uri(sources, '/connections/stats/for/:proxyType/:name/:mesh/:socketAddress', {
             proxyType: ({ ingresses: 'zone-ingress', egresses: 'zone-egress'})[route.params.proxyType] ?? 'dataplane',
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -10,219 +10,102 @@
     name="data-plane-detail-view"
     v-slot="{ route, t, can, me, uri }"
   >
-    <DataSource
-      :src="uri(sources, '/connections/stats/for/:proxyType/:name/:mesh/:socketAddress', {
-        proxyType: ({ ingresses: 'zone-ingress', egresses: 'zone-egress' })[route.params.proxyType] ?? 'dataplane',
-        name: route.params.proxy,
-        mesh: route.params.mesh || '*',
-        socketAddress: props.data.dataplane.networking.inboundAddress,
-      })"
-      v-slot="{ data: traffic, error, refresh }"
+    <AppView
+      :notifications="true"
     >
-      <AppView
-        :notifications="true"
+      <template v-if="props.source.data">
+        <DataSource
+          :src="uri(sources, '/connections/stats/for/:proxyType/:name/:mesh/:socketAddress', {
+            proxyType: ({ ingresses: 'zone-ingress', egresses: 'zone-egress' })[route.params.proxyType] ?? 'dataplane',
+            name: route.params.proxy,
+            mesh: route.params.mesh || '*',
+            socketAddress: props.source.data.dataplane.networking.inboundAddress,
+          })"
+          v-slot="{ error }"
+        >
+          <template
+            v-for="{ bool, key, params, variant } in [
+              {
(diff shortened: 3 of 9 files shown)


packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -10,640 +10,674 @@
...
+          <template v-if="props.source.data">
+            <DataSource
+              :src="uri(sources, '/connections/stats/for/:proxyType/:name/:mesh/:socketAddress', {
+                proxyType: ({ ingresses: 'zone-ingress', egresses: 'zone-egress' })[route.params.proxyType] ?? 'dataplane',
+                name: route.params.proxy,
+                mesh: route.params.mesh || '*',
+                socketAddress: props.source.data.dataplane.networking.inboundAddress,
+              })"
+              v-slot="{ data: traffic, error, refresh }"
             >
```
schogges, line 374: This `DataSource` was previously at the top level just below `RouteView`. Since it already depends on the `props.data` (and `props.source.data` being set, we tried to get around this by conditionally call the source via inline ternary. Unfortunately that removes all the types from the slot props which makes it harder to use. While this was not a big issue for the `ZoneDetailView` it was problematic here. So I've duplicated the `DataSource` in this view and wrapped the pieces that depend on it directly.

packages/kuma-gui/src/app/connections/views/ConnectionsStatsView.vue
```diff
@@ -16,7 +16,7 @@
       :title="t('data-planes.routes.item.navigation.data-plane-stats-view')"
     />
     <AppView>
-      <XCard>
+      <XCard v-if="props.networking">
```
schogges, line 19: While all this seems to work I have two points on the whole new approach now:

1. Since we've replaced `DataLoader` with `DataSource` in the tabs-routes, we can't rely on the provided `data` being defined anymore. Ideally we change all the subroutes to use `source` instead now, but this requires more changes everywhere.
2. I really dislike the fact that we lose typesafety on the subroutes. I think this is a general problem with dynamic routes

```vue
<!-- The parent route -->
<RouterView
  v-slot="{ Component }"
>
  <component
    :is="Component"
    :data="data" <-- potentially undefined, I expect an error because of the defined props in the child
    :source="{ data, error, refresh }"
  />
</RouterView>
```

```vue
<!-- The child route -->
<script lang="ts" setup>
const props = defineProps<{
  data: DataplaneOverview // <-- expects to be defined
}>()
</script>
```

johncowen: > I really dislike the fact that we lose typesafety on the subroutes. I think this is a general problem with dynamic routes

Yeah you're right I think. I've thought the same in the past. I guess its a downside of Vue/Vue Router, there's no guarantee here. I've thought about it in the past but not all the way through.

packages/kuma-gui/src/app/application/components/data-source/DataLoader.vue
```diff
@@ -14,10 +14,12 @@
...
-        <ErrorBlock
-          v-bind="$attrs"
-          :error="allErrors[0]"
-        />
+        <XCard>
+          <ErrorBlock
+            v-bind="$attrs"
+            :error="allErrors[0]"
+          />
+        </XCard>
```
johncowen, line 22: This is a possibly a misunderstanding, but this shouldn't be in here. I do remember asking you to add it while we were pairing tho just to see what happened. So while I did say "could you add a XCard in there for me", it wasn't supposed to stay! 😅

johncowen (commented):
I had a quick look at this, and left a couple of comments.

Aside from that, I'm super cautious about moving forwards with this without thinking a bit further first, I think that fact that I'm already cautious and we hit "code-y" issues and you have some concerns also makes me think we should pause a bit here.

I guess I feel like this isn't as easy as it should be, and I'm wondering if there's something we can do deeper in to make it easier. I guess this is reflected in your concerns also.

I'm actually wondering if part of the solution here is to take advantage of DataSource HTTP request sharing i.e. not pass the component props through but instead use a DataSource in the parent route and use the DataLoader in the child route. If DataSource is being used in the top-most route, then both the DataSource and DataLoader would be placed on the page a the same time, and therefore should use the same request.

Advantages:

- DataLoader gives us type safety
- We remove the props dependency in the child route making the child route almost dependency free
- I think this would make for some far simpler code with less `?` and `if`s checking for existence of things.

Still thinking a little but wanted to share thoughts on this whilst thinking. What do you think of the above idea?

packages/kuma-gui/src/app/application/components/data-source/DataLoader.vue
schogges: Ah yes, I'll remove it 😄 👍

schogges:
> I had a quick look at this, and left a couple of comments.
> 
> Aside from that, I'm super cautious about moving forwards with this without thinking a bit further first, I think that fact that I'm already cautious and we hit "code-y" issues and you have some concerns also makes me think we should pause a bit here.
> 
> I guess I feel like this isn't as easy as it should be, and I'm wondering if there's something we can do deeper in to make it easier. I guess this is reflected in your concerns also.
> 
> I'm actually wondering if part of the solution here is to take advantage of DataSource HTTP request sharing i.e. not pass the component props through but instead use a DataSource in the parent route and use the DataLoader in the child route. If DataSource is being used in the top-most route, then both the DataSource and DataLoader would be placed on the page a the same time, and therefore should use the same request.
> 
> Advantages:
> 
> * DataLoader gives us type safety
> * We remove the props dependency in the child route making the child route almost dependency free
> * I think this would make for some far simpler code with less `?` and `if`s checking for existence of things.
> 
> Still thinking a little but wanted to share thoughts on this whilst thinking. What do you think of the above idea?

Yeah, I totally agree. It doesn't really feel safe at this point and too many hacks, like if-statements, to make this work. So this is causing more work and the state of the codebase becomes worse and not better. So I think this should be approached bottom-up and work on the child routes first to remove the dependencies.

schogges:
Closing this up in favor of https://github.com/kumahq/kuma-gui/issues/4106

PR #4060, opened 2025-07-02
fix: gitignore deep cypress/downloads files

johncowen:
I 'think' this should catch all the files we don't want to add accidentally, also removes an accidental file.


schogges (approved):
👍

PR #4063, opened 2025-07-04
chore(ci): adds issue workflow to automatically set an "inbox" milestone

johncowen:
Guessing this will need merging before we can 100% test this works. I'll create a test issue once merged.

### DIFF
--- a/.github/workflows/issue-opened.yaml
+++ b/.github/workflows/issue-opened.yaml
@@ -1,0 +1,27 @@
+name: |
+  Issue Opened
+
+# Ensures that only one workflow is run per branch at a time.
+concurrency:
+  cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-opened.yaml
```diff
@@ -0,0 +1,27 @@
+name: |
+  Issue Opened
+
+# Ensures that only one workflow is run per branch at a time.
+concurrency:
+  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
```
schogges, line 7: I don't think we need to control concurrency for this. We always want to run the workflow and don't cancel it when another issue is opened while it already runs 🤔

johncowen: Removed

schogges (approved):
LGTM

PR #4066, opened 2025-07-04
fix(dataplane/stats): consider inbound servicePort to be different than port

schogges:
When not using transparent proxy it's possible to set a different `servicePort` than `port`. If not set the `servicePort` internally falls back to `port`. Due to not considering the potential difference of `servicePort` we didn't show the stats because the stats refer to the `servicePort`.
In order to include the potential `servicePort` I've added another field to the inbounds in the data layer: `clusterName`, which falls back to the `name` (`localhost_<port>`) which should provide a name of the grouping between proxy and service based on the networking configuration.

### DIFF
--- a/packages/kuma-gui/src/app/connections/views/ConnectionInboundSummaryClustersView.vue
+++ b/packages/kuma-gui/src/app/connections/views/ConnectionInboundSummaryClustersView.vue
@@ -23,14 +23,14 @@
           name: route.params.proxy,
           mesh: route.params.mesh || '*',
         })"
-        v-slot="{ data , refresh }"
+        v-slot="{ data: connections, refresh }"
       >
         <template
-          v-for="prefix in [route.params.connection.replace('_', ':')]"
+          v-for="prefix in [(props.data.clusterName === route.params.connection ? route.params.connection : props.data.clusterName).replace('_', ':')]"
           :key="typeof prefix"
         >
           <DataCollection
-            :items="data.split('\n')"
+            :items="connections.split('\n')"
             :predicate="item => item.startsWith(`${prefix}::`)"
             v-slot="{ items: lines }"
           >
@@ -63,7 +63,9 @@
 </template>
 <script lang="ts" setup>
 import { sources } from '../sources'
+import { DataplaneInbound } from '@/app/data-planes/data'
 const props = defineProps<{
   routeName: string
+  data: DataplaneInbound
 }>()
 </script>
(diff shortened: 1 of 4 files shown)


packages/kuma-gui/src/app/connections/views/ConnectionInboundSummaryClustersView.vue
```diff
@@ -23,14 +23,14 @@
           name: route.params.proxy,
           mesh: route.params.mesh || '*',
         })"
-        v-slot="{ data , refresh }"
+        v-slot="{ data: connections, refresh }"
       >
         <template
-          v-for="prefix in [route.params.connection.replace('_', ':')]"
+          v-for="prefix in [(props.data.clusterName === route.params.connection ? route.params.connection : props.data.clusterName).replace('_', ':')]"
```
johncowen, line 29: Maybe a bit of a nit here I suppose, but I stared at this for quite a while.

Not sure I get this. If clusterName is the same as connection, use connection, otherwise use clusterName. Well if its the same as connection, wouldn't you just use clusterName all the time?

johncowen (commented):
I left a couple of comments, but this looks really good. You've done it with a lot less code than I thought it would take 👏 

Did you check this is fine when viewing this viz in Zones? I'm pretty sure its fine as its simpler over there, but just to check

packages/kuma-gui/src/app/connections/views/ConnectionInboundSummaryClustersView.vue
schogges: Oh 🤦 I suppose I wanted to prefer `clusterName` if it's set, but always default to `route.params.connection`. As `clusterName` is empty for `gateways` we need to fallback to `connection`, which is set from the `traffic` in case of `gateway`. Only for `proxy` it's set from the name and in case of `proxy` we prefer `clusterName` (which defaults to `name`).

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -429,7 +429,7 @@
                           >
                             <template
                               v-for="stats in [
-                                traffic?.inbounds[item.name],
+                                traffic?.inbounds[item.clusterName || item.name], // check for both because clusterName is empty on gateway proxy types
```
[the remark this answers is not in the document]
schogges: We can't set the `clusterName` to `name`, because in case of `gateway` is the value of `kuma.io/service` tag. We can't use this value to extract the stats from the traffic.

PR #4069, opened 2025-07-04
refactor: move vite config and shared plugins to config

schogges:
Moves the common parts of the vite config to a base config exposed by the config package. The different configs don't depend on each other anymore but each extends the base config individually. Also the common plugin configs are moved to the config package and each vite config imports the plugin configs. Also as the `vite.config.development` just imported and reexported the production config, I have removed it.


packages/kuma-gui/vite.config.preview.ts
```diff
@@ -1,19 +1,38 @@
...
+        yamlLoader,
+        svgLoader,
+      }),
       replicateKumaServer({
         template: './dist/gui/index.html',
       }),
+      fakeApi({
+        dependencies,
+        fs,
+      }),
```
schogges, line 29: The `fakeApi` plugin can't be moved to the config package as it would cause a circular dependency. If we want to serve all common plugins from a package we'd have to create a new package for `vite` related stuff.

schogges: Changed it to only export the configs for the plugins. For the moment we neither move the plugins to the config package nor make them injectable.

packages/kuma-gui/vite.config.preview.ts
```diff
@@ -1,19 +1,34 @@
...
+      port: 5681,
+    },
     plugins: [
+      vue(vuePluginConfig()),
+      yamlLoader(yamlLoaderPluginConfig()),
+      svgLoader(),
       replicateKumaServer({
         template: './dist/gui/index.html',
       }),
+      fakeApi({ dependencies, fs }),
```
johncowen, line 24: I don't think we need any of the plugins for the preview server, it pretty much just runs the static artifacts from a `make build`. The only one it needs is the `replicateKumaServer` to replicate what golang does with the inde.html file.

johncowen (commented):
I definitely think you were right think the previous one to this was overshot, this looks like less stuff now in order to externalize the config, and whilst we said we might still push it forwards a little more a later date, this is a first step that gets us to what we want easier.

I left one comment which I think is important, but other than that I think we'll be able to get this in pretty quick.

packages/kuma-gui/vite.config.preview.ts
schogges: Good point 👍 I realized that the preview config didn't extend the prod config as the dev config did. So I mostly rolled it back to how it was, just kept the order of config entries the same as in prod config.

johncowen (approved):
This looks good to me, I particular like that this feels more loosely coupled that previously. Also glad we are getting rid of the `.development` files and losing the "inheritance feeling" file chain.

I guess you might have tried this elsewhere already, but if not this might be one thats worth giving it a check and before merging here.

Issue #3632, opened 2025-03-06
Improve NACK overview


lukidzi:
### Description

While testing we noticed that we could improve few things:
* show a small warning sign next to subscriptions that had NACKs
* add a link to the message opening a tab with nacked resources
* improve a message to not say to the user that needs to restart control-plane (more like info)
* show rejected resources in a separate section, now it's hard to find what kind of resource was rejected

PR #4070, opened 2025-07-04
chore(subscriptions): improve nack overview

johncowen:
Adds several improvements detailed in the below issue.

Note: I have a follow up PR to this which further cleans some things up in the `...View.vue` file

Closes https://github.com/kumahq/kuma-gui/issues/3632

### DIFF
--- a/packages/kuma-gui/src/app/zones/views/ZoneDetailView.vue
+++ b/packages/kuma-gui/src/app/zones/views/ZoneDetailView.vue
@@ -46,7 +46,24 @@
             <XI18n
               :path="`zone-cps.notifications.${key}`"
               :params="Object.fromEntries(Object.entries(params ?? {}))"
-            />
+            >
+              <template
+                v-if="key === 'global-nack-response'"
+                #link
+              >
+                <XAction
+                  data-action
+                  :to="{
+                    name: 'zone-cp-subscription-summary-view',
+                    params: {
+                      subscription: props.data.zoneInsight.connectedSubscription?.id,
+                    },
+                  }"
+                >
+                  Zone Control Plane Summary
+                </XAction>
+              </template>
+            </XI18n>
           </XNotification>
         </template>
         <XLayout
@@ -130,7 +147,7 @@
             <template #title>
               <h2>{{ t('zone-cps.detail.subscriptions.title') }}</h2>
             </template>
-            
+
             <XLayout>
               <XI18n path="zone-cps.detail.subscriptions.description" />
               <AppCollection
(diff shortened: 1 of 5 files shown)


packages/kuma-gui/src/app/zones/views/ZoneDetailView.vue
```diff
@@ -46,7 +46,24 @@
...
+                <XAction
+                  data-action
+                  :to="{
+                    name: 'zone-cp-subscription-summary-view',
+                    params: {
+                      subscription: props.data.zoneInsight.connectedSubscription?.id,
+                    },
+                  }"
+                >
+                  Zone Control Plane Summary
```
schogges, line 63: I think this should be in i18n? 🙂

packages/kuma-gui/src/app/zones/views/ZoneDetailView.vue
```diff
@@ -46,7 +46,24 @@
...
+              <template
+                v-if="key === 'global-nack-response'"
+                #link
+              >
+                <XAction
+                  data-action
+                  :to="{
+                    name: 'zone-cp-subscription-summary-view',
+                    params: {
+                      subscription: props.data.zoneInsight.connectedSubscription?.id,
```
schogges, line 59: It looks like the `connectedSubscription` is potentially not set. Should we show the link if it's not set?

packages/kuma-gui/src/app/zones/views/ZoneDetailView.vue
johncowen: Ideally I should be able to do `blah blah blah <a href="{url}">Zone Control Plane Summary</a> blah blah blah` in the locales file, which I tried at first, but Vue does not automatically make anchors with application URLs use `pushState` instead of traditional navigation.

I decided to just leave this as is because I want to avoid having using things like:

```yaml
text: The text {link} here
text_link: Zone Control Plane Summary
```

when what we actually want is:

```yaml
text: The text <a href="{url}">Zone Control Plane Summary</a> here
```

As there is no extreme rush to i18n everything, and if there was its a couple of hours work to go through and i18n things that aren't currently if it did become urgent, I'd rather avoid doing something I'd rather not do and instead fix the problem of "Vue does not automatically make anchors with application URLs use `pushState`"

Happy to make an issue to track that and leave this as is until thats solved, or alternatively if you don't agree I can make the changes that I don't really want to do detailed above.

schogges: Ok, fair point 🙂 👍 
Just thinking loud, but probably something like the following would help
```vue
<XI18n components="{ a: <XAction ... /> }" />
```

schogges (approved):

packages/kuma-gui/src/app/zones/views/ZoneDetailView.vue
johncowen: An idea for this was to add an eventlistener to the document and listen for click events on all `a` tags (we will probably want to filter the ones we act on not to be XAction)

Then we could just do:

```
text: The text <a href="{url}">Zone Control Plane Summary</a> here
```

Clicking on this anchor link in the I18n file would be caught by the document level event listener, and instead of doing a traditional browser navigation we do a vue `router.push()` to do a push state.

PR #4084, opened 2025-07-09
chore(deps): removes @types/glob

johncowen:
Instead of upgrading this here https://github.com/kumahq/kuma-gui/pull/4075, I looked to see why we need this, and I found that the `glob` package is in typescript (at least version 11 it seems), so I'm pretty sure we don't need to depend on this anymore.

I also checked in my IDE to make sure I was still getting type hints for glob after uninstalling this and I am.


schogges:
Nice 👍 
A note on `glob`: I'm pretty sure `fs.globSync` was experimental in at least `node@22.16`, but it's stable now: https://nodejs.org/docs/latest-v22.x/api/fs.html#fsglobsyncpattern-options
So I think we can also remove the dependency to `glob` and use native `node:fs`'s `globSync`. Nothing to be done here, just a note 🙂

PR #4085, opened 2025-07-09
chore(deps): upgrade brace-expansion

johncowen:
Upgrades transitive brace-expansion dependency from 1.1.11 to 1.1.12 where necessary using `npm upgrade brace-expansion`

I'd like to super importantly point out that renovate has not upgraded this for us nor notified us, nor is it in its Dependency Dashboard.

Anyway, this PR does what's necessary.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2026,9 +2026,9 @@
       }
     },
     "node_modules/@eslint/config-array/node_modules/brace-expansion": {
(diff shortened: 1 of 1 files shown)


johncowen:
> In case there is a security issue we should, but not just because they are outdated

Exactly, dependabot and `npm audit` told me about this, renovate didn't

PR #4088, opened 2025-07-10
chore: catch more load failures

johncowen:
This expands our "catch a HTTP0 and retry" to catch slightly more TypeErrors, I'm still being cautious to open this up to just catch _all_ TypeErrors, and whilst checking the error message ain't great, its achieves what we want for now.


schogges (approved):
👍

Issue #4040, opened 2025-06-25
Take dataplane authentication type from `dpProxy/zoneProxy` not `dpServer`


johncowen:
### Kuma Version

2.11.x

### Describe the bug

We currently use `dpServer.authn.type` in the GUI, but this no longer exists. We should use `dpProxy` and `zoneProxy`:

https://github.com/kumahq/kuma/blob/df678629aa3bdce62a84c0e3752c7e5d45bf1e98/pkg/config/app/kuma-cp/kuma-cp.defaults.yaml#L600-L627

See in the below grabs where there is an authentication type set but we show `N/A`

![Image](https://github.com/user-attachments/assets/7cd90b5d-4e07-4788-8949-246cc6f4906d)

![Image](https://github.com/user-attachments/assets/8ed4a8bb-ccd2-4a6f-b059-d9af07565ebb)

PR #4103, opened 2025-07-15
fix(zone-cp): dp authentication type

schogges:
Updates the extraction of zone dp `authenticationType` and the related mocks.

---

I thought about inlining the mocks returned by `subscriptionConfig` in `packages/kuma-gui/src/test-support/FakeKuma.ts`, but it's quite big and used in two places. So I left it in `FakeKuma`.

Closes https://github.com/kumahq/kuma-gui/issues/4040

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -34,7 +34,8 @@
     },
     {
       "matchPackageNames": [
-        "npm"
+        "npm",
+        "escape-string-regexp"
       ],
       "enabled": false
     },
--- a/packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
+++ b/packages/kuma-gui/src/app/application/components/app-collection/AppCollection.vue
@@ -44,10 +44,9 @@
 
 <script lang="ts" setup generic="Row extends {}">
 import { KTableView } from '@kong/kongponents'
-import { ref, inject, onMounted, ComponentPublicInstance } from 'vue'
+import { ref, onMounted, ComponentPublicInstance } from 'vue'
 
 
-import { runInDebug } from '../../'
 import type { TableViewHeader as KTableViewHeader, TablePreferences } from '@kong/kongponents'
 type ResizeValue = {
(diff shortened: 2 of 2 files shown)


packages/kuma-gui/src/test-support/FakeKuma.ts
```diff
@@ -1,4 +1,4 @@
-import { Faker } from '@faker-js/faker'
+import { faker, Faker } from '@faker-js/faker'
```
johncowen, line 1: How come we are reimporting faker instead of using this.faker like we do everywhere else? I honestly don't know why we do one over the other, but we should stick to one way and if we want to change we do it throughout.

johncowen (commented):
> I thought about inlining the mocks returned by subscriptionConfig in packages/kuma-gui/src/test-support/FakeKuma.ts, but it's quite big and used in two places. So I left it in FakeKuma.

Yeah totally, this is just a big blob of JSON that we barely access, its not like we have to randomize and control it like other largish data in our mocks, so this one is ok in FakeKuma. The problems arise when small to medium data structures are squirrelled away in FakeKuma and then you end up passing more and more arguments to them that slowly becomes more complex and dense/black-boxy that its just easier to copy/paste the data.

I let one little nit here, but otherwise this is good to go!

packages/kuma-gui/src/test-support/FakeKuma.ts
schogges: The problem is that the function that returns the actual mock is not part of the class. `FakeKuma` itself has a method that uses this function and `deepmerge` to merge any potentially provided fixtures/mocks. I move the function body into the `FakeKuma.subscriptionConfig` such that I can use `this.faker`. Wdyt?

schogges: Sounds good, I'll inject `faker` 👍

schogges: Done

PR #4113, opened 2025-07-21
feat: export env and can from new settings package

schogges:
Part of the overall restructuring and is a pre-effort to outsource the testing service into another package.

---

The `CliEnv`, which is only used during unit testing, should not be part of the common export of the new package and therefore not added to the barrel file. Otherwise it would be part of the bundle and end up in the browser. Therefore I've just added a separate export to the manifest. That's a common problem with barrel files and we should be careful what is added to the barrel files while we restructure the codebase.

---

Regarding unit tests: since we have `test.setupFiles` added to our base config I had to create such files in the packages where I started to add unit tests. I created some placeholder files, but I wasn't quite happy about it. Instead I've just decided to remove the `test.setupFiles` from the base config leaving it up to the consumer to set this.

### DIFF
--- a/packages/config/src/vite/config/vite.config.base.ts
+++ b/packages/config/src/vite/config/vite.config.base.ts
@@ -32,9 +32,6 @@ export const defineConfig: UserConfigFn = () => ({
   test: {
     globals: false,
     environment: 'jsdom',
-    setupFiles: [
-      path.resolve(process.cwd(), './test-support/main.ts'),
-    ],
     deps: {
       optimizer: {
         web: {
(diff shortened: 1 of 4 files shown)


packages/kuma-gui/test-support/index.ts
```diff
@@ -1,10 +1,16 @@
...
 import { config } from '@vue/test-utils'
 
-import CliEnv from '@/app/application/services/env/CliEnv'
 import type { PluginDefinition, ComponentDefinition } from '@/app/vue'
 import type { Component } from 'vue'
 
+const $ = {
+  htmlVars: token<KumaHtmlVars>('htmlVars'),
+}
+export const TOKENS = $
```
schogges, line 12: Sry, just noticed that these are still in here. That's a left over from fiddling around. Needs to be removed.

bartsmykla (approved):

PR #4132, opened 2025-07-28
fix(deps/security-actions/sca): use hash of version release

schogges:
Our release workflow to the kuma repo is currently failing. It looks like there is an issue upstream. It started failing since we've used a hash that is newer than the hash of the actual release of the latest version of `Kong/public-shared-actions/security-actions/sca`. This is an attempt to make it work by using the hash of the actual release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,7 +83,7 @@ jobs:
       - run: |
           make -C ${{ matrix.package.path }} build
 
-      - uses: Kong/public-shared-actions/security-actions/sca@ec3f7e6638900dfb56c6963b61659512fc6dbe54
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -83,7 +83,7 @@ jobs:
       - run: |
           make -C ${{ matrix.package.path }} build
 
-      - uses: Kong/public-shared-actions/security-actions/sca@ec3f7e6638900dfb56c6963b61659512fc6dbe54
```
slonka, line 86: This was bumped by renovate https://github.com/kumahq/kuma-gui/commit/b540d92be0a467b3c4e1dcb9e770e066112a6e82 - does it mean there is a bug in renovate? FYI @bartsmykla

schogges: It's probably because we haven't set the version in the comment afterwards so renovate might think we want to pin hash on master and not a version? 🤔

PR #4150, opened 2025-08-12
feat(data-planes): relocate subscriptions listing

schogges:
The subscriptions listing (XDS connections) was a bit lost down below the traffic listing (inbounds/outbounds), so we've decided to relocate it. The most important information is now part of the DP about section and adds a link to a summary tray that shows the whole listing with all information.

Note: As the subscriptions do not depend on the new DP feature flag (`feature-unified-resource-naming`) I've done the same change for both `data-planes` and `legacy-data-planes`.

https://github.com/user-attachments/assets/e0660561-4511-4b23-84e0-e1e7fba061ee


schogges:
Hm, I'm investigating why the e2e still fails. I can't repro locally. Will turn it back to draft until I found a solution.

schogges:
Well, seems like the odd waiting for 500ms did the trick 😄

bartsmykla (approved):

PR #4176, opened 2025-08-22
feat(data-planes): check for mesh-services mode to use unified-naming

schogges:
We've noticed that when `meshServices.mode` is disabled on the mesh, we can't use `feature-unified-resource-naming`. Until this condition is fully considered on the backend, we are going to check for the `meshServices.mode` being set to `Exclusive`. The backend will work on exposing a new field `dataplaneInsights.features` which will either include the feature flag or not by taking into account if mesh services are enabled or not.

### DIFF
--- a/packages/kuma-gui/src/app/legacy-data-planes/features.ts
+++ b/packages/kuma-gui/src/app/legacy-data-planes/features.ts
@@ -1,4 +1,5 @@
 import type { DataplaneOverview } from '@/app/legacy-data-planes/data'
+import type { Mesh } from '@/app/meshes/data'
 import type { Features } from '@kumahq/settings/can'
 import type { Env } from '@kumahq/settings/env'
 export const features = (_env: Env['var']): Features => {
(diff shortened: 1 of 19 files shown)


packages/kuma-gui/src/app/legacy-data-planes/features.ts
```diff
@@ -9,8 +10,8 @@ export const features = (_env: Env['var']): Features => {
       return ('transparentProxying' in dataplaneOverview.dataplane.networking) ||
         new Set(dataplaneOverview.dataplaneInsight.metadata.features).intersection(new Set(['feature-transparent-proxy-in-dataplane-metadata', 'bind-outbounds'])).size > 0
     },
-    'use unified-resource-naming': (_can, dataplaneOverview: DataplaneOverview) => {
-      return dataplaneOverview.dataplaneType === 'standard' && dataplaneOverview.dataplaneInsight.metadata.features.includes('feature-unified-resource-naming')
+    'use unified-resource-naming': (_can, { dataPlaneOverview, mesh }: { dataPlaneOverview: DataplaneOverview, mesh: Mesh }) => {
+      return mesh.meshServices.mode === 'Exclusive' && dataPlaneOverview.dataplaneType === 'standard' && dataPlaneOverview.dataplaneInsight.metadata.features.includes('feature-unified-resource-naming')
```
schogges, line 14: With this we only allow unified resource naming when `meshServices.mode` is `Exclusive`. Is this enough or should it also allow with `Everywhere`?

johncowen: K sounds like this is ready to go then. Thanks @bartsmykla! 👍

PR #4186, opened 2025-08-25
fix(e2e/tls): flake timing issue

schogges:
Again one of those scenarios where the click on the link happens too fast leading to the routing not being performed and the test fails. I've added a random `wait for 500 ms` to get around this issue.
Furthermore I've deduped the env variable `KUMA_DATAPLANE_TLS_ISSUED_MESHIDENTITY` as this is already set for all scenarios.


schogges:
> Approving this for now, but we need to look into how come we are getting more of these. I think this is a separate issue from the "disabled input field" one, so I"m a little more cautious adding these.
> 
> But anyway, lets go with this for now, these are easy to find at a later date and try to remove.

Yeah I agree, it's weird that it's getting more. We had tests that include clicking on links and asserting for certain URLs and different views before already.

PR #4194, opened 2025-08-27
chore: improve typing of can utility

johncowen:
`can()` has always accepted any strings (and number of arguments), if a 'feature'/'ability' didn't exist then we would return `false` i.e. non-existence meant not enabled.

This PR uses declarations to make `can` autocompletable/type-checked instead by using composable TS interfaces and `declare` to amend `can`s definition after initial declaration.

My main goals here are:

- Add typechecking etc to `can` from The Outside
- Keep our existing composable nature of `can`
- Make it as automatic as possible, I shouldn't have to repeat all the feature declarations, hence the utility type.

I eventually plan on re-building `env()` to use a similar approach, thus finally making env composable also.

### Notes

- There is a ts-ignore here https://github.com/kumahq/kuma-gui/pull/4194/files#diff-5fd72b2766aa02f5cdb96929d8cf41d9d6b8a6111e18d0e4b5f2d1cf346705c4R2. I couldn't find a better way to do this, everything else I tried wasn't as good as just using ts-ignore (form an autocomplete perspective). Happy to hear other suggestions to remove this.
- Currently I added the `declares` in the feature.ts files themselves. In the probably distant future its very likely these could be moved to be declared centrally in main.ts (or another single d.ts file sourced from main)
- This might not be the final approach, but it the approach here is unobtrusive enough that its easy to change/back out later when/if we want to change.

### DIFF
--- a/packages/kuma-gui/src/app/application/index.ts
+++ b/packages/kuma-gui/src/app/application/index.ts
@@ -20,7 +20,6 @@ import storage from './services/storage'
 import { create, destroy, DataSourcePool } from '@/app/application/services/data-source'
 import { services as kuma } from '@/app/kuma'
 import type { ServiceDefinition } from '@kumahq/container'
-import type { Can } from '@kumahq/settings/can'
 import type { EnvVars } from '@kumahq/settings/env'
 import type { Component } from 'vue'
 import type { RouteRecordRaw } from 'vue-router'
@@ -63,6 +62,8 @@ declare module 'vue' {
     $routeName?: string
   }
 }
+export interface Abilities {}
+export type Can = Abilities['can']
 
 const $ = {
   Env: token<Env>('application.Env'),
--- a/packages/kuma-gui/src/app/control-planes/features.ts
+++ b/packages/kuma-gui/src/app/control-planes/features.ts
@@ -10,7 +10,7 @@ export const features = (env: Env['var']) => {
     },
   }
 }
-declare module '@kumahq/settings/can' {
+declare module '@/app/application' {
   export interface Abilities {
     can(...args: Features<ReturnType<typeof features>>): boolean
   }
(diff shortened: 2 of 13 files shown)


johncowen:
@schogges I was just looking at the failing tests here, and from what I remember, I had it properly typechecking these args when I initially pushed this PR. If that is the case I think we should revert your amends ontop of here, but keep the tests.

I'm gonna check this morning and if so I'll revert this PR back to that state and add the tests on top, please yell if you think otherwise.

schogges:
> @schogges I was just looking at the failing tests here, and from what I remember, I had it properly typechecking these args when I initially pushed this PR. If that is the case I think we should revert your amends ontop of here, but keep the tests.
> 
> I'm gonna check this morning and if so I'll revert this PR back to that state and add the tests on top, please yell if you think otherwise.

I just fixed it. It was a typo in `data-planes/features.ts`. Somehow your changes in https://github.com/kumahq/kuma-gui/pull/4195/files were not applied, probably after rebasing, and it correctly showed a type mismatch. I just interpreted it wrong.
So after fixing the typo of `dataPlaneOverview` vs `dataplaneOverview`, it all works as expected now.

johncowen:
Ah great, looks much better 👍 

> At some point I'd like to fix the @ts-expect-errors

Yep totally, we currently only ever use this with the `declare` approach anyway and everything is typed correctly when we do that, but I'm the same, we can fix at some point so its correct, but in the meantime it doesn't affect us.

Gonna go ahead and merge 👍

schogges:
> Oh I can't merge 😆 and I guess you can't either? Do we need an extra ✅ from @bartsmykla ?

Oh 😄 yeah I can't merge either...

bartsmykla (approved):

PR #4198, opened 2025-09-02
refactor(resources/routes): use kri for routing of mid and mtrust

schogges:
We decided to use `kri`s as is whenever possible. The first and probably easiest places where we can do this is the new resources `MeshIdentity` and `MeshTrust` which have been lately being added within the context of `kri`s.
There are still some places that require the `name` instead of the `kri`, but at the very least we are trying to remove the parsing of `kri`s from the views and only keep it in the data-layer and sources.

The following routes now expect a `kri`:
- `mesh-identity-summary-view`: `/meshes/:mesh/meshidentities/:mid` - `:mid` is expected to be a `kri`
- `mesh-trust-summary-view`: `/meshes/:mesh/meshtrusts/:mtrust` - `:mtrust` is expected to be a `kri`

For both `MeshTrust` and `MeshIdentity` I've added to build a `kri` from the provided data and add it to the returned data in the data-layer. For now this is just a minimal `kri`, i.e. `kri_mid_default___identity-1_` with only `shortName`, `mesh` and the `name` of the resource.

---

Furthermore I've fixed some minor things
- the route name of `MeshTrustSummaryView` was set to `mesh-identity-summary-view`
- now that we add more data to the response of `MeshTrust` and `MeshIdentity`, I've added another entry `raw` to preserve the returned data as config.

### DIFF
--- a/packages/kuma-gui/src/app/resources/data/MeshIdentity.ts
+++ b/packages/kuma-gui/src/app/resources/data/MeshIdentity.ts
@@ -1,3 +1,4 @@
+import { Kri } from '@/app/kuma/kri'
 import type { components } from '@kumahq/kuma-http-api'
 
 type PartialMeshIdentityList = components['responses']['MeshIdentityList']['content']['application/json']
@@ -7,6 +8,8 @@ export const MeshIdentity = {
   fromObject: (item: PartialMeshIdentity) => {
     return {
       ...item,
+      kri: Kri.toString({ shortName: 'mid', mesh: item.mesh, name: item.name}),
+      raw: item,
     }
   },
   
--- a/packages/kuma-gui/features/mesh/MeshIdentity.feature
+++ b/packages/kuma-gui/features/mesh/MeshIdentity.feature
@@ -28,10 +28,11 @@ Feature: mesh / mesh-identity
       body:
         items:
           - name: identity-1
+            mesh: default
       """
     When I visit the "/meshes/default" URL
     Then I click the "$mesh-mtls a:first" element
-    Then the URL contains "/meshes/default/overview/meshidentity/identity-1"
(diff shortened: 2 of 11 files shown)


packages/kuma-gui/src/app/resources/data/MeshIdentity.ts
```diff
@@ -7,6 +8,8 @@ export const MeshIdentity = {
   fromObject: (item: PartialMeshIdentity) => {
     return {
       ...item,
+      kri: Kri.toString({ shortName: 'mid', mesh: item.mesh, name: item.name}),
```
johncowen, line 11: I thought these resources already had kris?

schogges: Only partially. I.e. `MeshTrust` has `spec.origin.kri` which is the `kri` of a `MeshIdentity`. Both do not have a `kri` that links to themselves 😢

johncowen: face palm...

johncowen (dismissed):
Looks good 👍 

I had a nitpick suggestion potentially for the future.

How would you feel about making `Kri` feel a bit closer to URL (https://developer.mozilla.org/en-US/docs/Web/API/URL)?

One way to do that would be to make it uppercase `KRI`, lmk - not a biggie at all tho.

schogges:
> Looks good 👍
> 
> I had a nitpick suggestion potentially for the future.
> 
> How would you feel about making `Kri` feel a bit closer to URL (https://developer.mozilla.org/en-US/docs/Web/API/URL)?
> 
> One way to do that would be to make it uppercase `KRI`, lmk - not a biggie at all tho.

Yeah, we can do that 👍 🙂

schogges:
@johncowen had to rebase and fix conflicts, need another approval please 🙂

Issue #4193, opened 2025-08-27
rollout XTable component


johncowen:
### Description

https://github.com/kumahq/kuma-gui/pull/4191 adds a new XTable component, but adds it in one single place as an example.

The component needs to be rolled out replacing everywhere we currently use `<DefinitionCard>`/`class="stacked-with-borders"`

PR #4200, opened 2025-09-02
chore: rollout XTable

johncowen:
Follow up of https://github.com/kumahq/kuma-gui/pull/4191

Closes https://github.com/kumahq/kuma-gui/issues/4193

Replaces DefinitionCard in all SummaryViews with XTable. DefinitionCard should no longer be being used for summary view, border separated tables anywhere.

Note: I made some further changes to XTable CSS here to get the layout exactly the same as previously.

I also noticed that following this change these links are no longer blue, I plan to resolve this in a follow up PR (I already have the change here) (edit: see https://github.com/kumahq/kuma-gui/pull/4207)

Screenshot 2025-09-03 at 15 04 33


schogges (approved):
LGTM

PR #4203, opened 2025-09-03
fix(mesh): mtls warning shows up with mesh identity

schogges:
The mtls warning in the `MeshOverview` should only show up if mtls is not configured. With `MeshIdentity` as issuer the `mtls` field in `/meshes/{mesh}` is not set, so we also have to check for `MeshIdentity`s.

image

This also needs to go into 2.12

---

We've decided to take this a little further. We want to move the list of `MeshIdentity`s to the `mTLS` part of the about section. Also the `MeshTrust` listing will only be rendered conditionally in case there is at least one `MeshTrust` resource.

### DIFF
--- a/packages/kuma-gui/features/mesh/Item.feature
+++ b/packages/kuma-gui/features/mesh/Item.feature
@@ -2,12 +2,13 @@ Feature: mesh / item
 
   Background:
     Given the CSS selectors
-      | Alias              | Selector                                 |
-      | error              | [data-testid="error-state"]              |
-      | service-count      | [data-testid="services-status"]          |
-      | config-universal   | [data-testid='codeblock-yaml-universal'] |
-      | config-k8s         | [data-testid='codeblock-yaml-k8s']       |
-      | select-environment | [data-testid='select-input']             |
+      | Alias              | Selector                                                        |
+      | error              | [data-testid="error-state"]                                     |
+      | service-count      | [data-testid="services-status"]                                 |
+      | config-universal   | [data-testid='codeblock-yaml-universal']                        |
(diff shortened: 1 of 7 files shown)


johncowen:
I feel like having `mTLS: Disabled` showing in this case is misleading?

How about we do something like

- `mTLS Issuer: MeshIdentity / identity.kuma-system` (new style)
- `mTLS Issuer: provided backend / ca-1` (old style)

Then we could lose the MeshIdentities XCard entirely?

We should probably pass that idea by someone else if you like it, but form what I'd understood this morning, this approach would make more sense.

schogges:
> I feel like having `mTLS: Disabled` showing in this case is misleading?
> 
> How about we do something like
> 
> * `mTLS Issuer: MeshIdentity / identity.kuma-system` (new style)
> * `mTLS Issuer: provided backend / ca-1` (old style)
> 
> Then we could lose the MeshIdentities XCard entirely?
> 
> We should probably pass that idea by someone else if you like it, but from what I'd understood this morning, this approach would make more sense.
> 
> P.S. We could actually follow a design / badge we have elsewhere where we show something that looks like `[MeshIdentity:identity.kuma-system]` (the `[]`s are to show the badge)

Yeah, sounds good to me. Let's get some feedback before we continue on this. I'll put it back to draft.

schogges:
> As a totally new feature suggestion, I also think it might be worth showing and/or warning on the expiry of the certificate associated with the MeshIdentity (if I've understood the MeshIdentity response correctly as to what it does)

AFAIK this is not possible as we don't know anything about the expiry of the certificate. With MeshIdentity this is managed externally 🤔

PR #4223, opened 2025-09-05
chore(tsconfig): enable verbatimModuleSyntax

schogges:
Enable [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) to enforce proper type imports.

Example

```sh
src/test-support/mocks/src/meshes/_/meshidentities/_/_.ts:1:10 - error TS1484: 'components' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.

1 import { components } from '@kumahq/kuma-http-api'
```

to fix this

```typescript
```
or
```typescript
import { type components } from '@kumahq/kuma-http-api'
```

One note on this:
In our data-layer we usually create an object that we use to backfill and change some data that we receive as a response from an http request. We export a type with the same name that gives us the type of the new data structure. Due to using the same name this new rule has no effect on these imports, because both would be a valid import of either the type or the object.

Example:
```typescript
export const ZoneOverview = {
  fromObject: (item: PartialZoneOverview) => {
    // ...
  },
  fromCollection: (collection: CollectionResponse<PartialZoneOverview>) => {
    // ...
  },
}
export type ZoneOverview = ReturnType<typeof ZoneOverview.fromObject>
```


packages/kuma-http-api/package.json
```diff
@@ -3,7 +3,14 @@
...
   "private": true,
   "description": "Kuma OpenAPI to typescript converter and generated typescript types",
+  "type": "module",
   "types": "index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./index.d.ts",
+      "types": "./index.d.ts"
+    }
+  },
```
schogges, line 13: Adding `verbatimModuleSyntax` also revealed this issue:

```sh
error during build:
[commonjs--resolver] Failed to resolve entry for package "@kumahq/kuma-http-api". The package may have incorrect main/module/exports specified in its package.json.
```

in order to fix it, I've made the above changes.

johncowen (dismissed):

johncowen (approved):

PR #4226, opened 2025-09-08
chore: delete old FilterBar

johncowen:
I noticed our old filter bar component was still hanging around.

This PR deletes it.


schogges (approved):
Oh nice catch! Forgot to delete it after adding `XSearch`.

🧹 🧹

Issue #3313, opened 2024-12-17
Consider adding indicator for nested anchors in badges


schogges:
### Description

x-ref: https://github.com/kumahq/kuma-gui/pull/3300#discussion_r1888453712

In some cases we add links as content to badges. As a user it is hard to identify just by looking at it that there is a link behind. Only when hovering the mouse becomes a pointer and the text is underlined.
Possible options I had in mind so far is to always add a `text-decoration: underline` or an icon that indicates the clickable link and let it stand out to other badges.

_Without Indicator_
![Screenshot 2024-12-17 at 14 58 46](https://github.com/user-attachments/assets/887648b7-188a-4647-a66d-98781eb4d2db)

_With text-decoration_
![Screenshot 2024-12-17 at 14 59 02](https://github.com/user-attachments/assets/27ea8658-864e-4837-812e-c7c15b2b77bd)

lahabana:
> Triage: design needs to give us a direction here

lahabana:
still need design here

PR #4231, opened 2025-09-10
feat(dataplanes): add link icon to linked badges

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/3313

Unfortunately, I had to add a little `-1px` hack to get the icon to look lined up to my eye. I'm totally happy with this approach, but I don't mind if someone prefers a different approach, as long as it looks aligned correctly.

### DIFF
--- a/packages/kuma-gui/src/app/common/TagList.vue
+++ b/packages/kuma-gui/src/app/common/TagList.vue
@@ -7,20 +7,23 @@
       'tag-list--align-right': props.alignment === 'right',
     }"
   >
-    <XBadge
+    <template
       v-for="(tag, index) in tagList"
       :key="index"
-      class="tag kv"
-      :data-kv-key="tag.label"
-      :data-kv-owner="tag.label.split('/')[0]"
     >
       <component
         :is="tag.route ? 'XAction' : 'span'"
         :to="tag.route"
       >
-        <span class="label">{{ tag.label }}</span>:<span class="value">{{ tag.value }}</span>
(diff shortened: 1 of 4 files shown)


schogges:
What do you think about solving this via CSS? I think currently we don't use the combination of `XBadge` and `XAction` consistently, so either we make it consistent or we'd have to add some CSS to both `XBadge` and `XAction`.
I had something like the following in mind:

```css
/* XAction */
:has(.k-badge) :deep(.badge-content-wrapper) {
  &::after {
    content: url('@/assets/images/chevron-down.svg?url');
    margin-left: 1.25em;
    font-size: 0.75em;
  }
}
```

```css
/* XBadge */
:has(.x-action-appearance-anchor) {
  &::after {
    content: url('@/assets/images/chevron-down.svg?url');
    margin-left: 1.25em;
    font-size: 0.75em;
  }
}
```

johncowen:
Not sure I follow due to the use of chevron? Do you mean use CSS to add the icon? If so I would love to do that (and have wanted to use our icons like this since forever). Can we use kong icons like this? (we need to use the link icon)

johncowen:
Ok we've changed this up to add the XIcon inside of XBadge and then use provide / inject so that XBadge knows when its in a XAction or not, that way we don't add the component via JS if its not required.

Whilst we'd love to add icons via CSS, at the moment at least that's not something we can do, so the JS route feels like the right approach to me for the moment.

Issue #3154, opened 2024-11-07
[Dev]: Doc links should redirect to dev


schogges:
### What happened?

After cutting branches for a new release we bump the version in the `package.json`. In development this version is set in the `kuma-config` and then used for the links to the docs. Clicking a docs link then leads to the 404 page since the version is not there yet. Instead we should redirect to the dev docs.

lukidzi:
> triage: let's check if we can do something on the docs side have a next release tag ready

PR #4251, opened 2025-09-16
chore: only in dev and previews, link docs to /dev

johncowen:
See title

Closes https://github.com/kumahq/kuma-gui/issues/3154

### DIFF
--- a/packages/kuma-gui/src/app/kuma/debug.ts
+++ b/packages/kuma-gui/src/app/kuma/debug.ts
@@ -10,6 +10,22 @@ const $ = {
 }
 
 export const services = (app: Record<string, Token>): ServiceDefinition[] => [
+  [token('kuma.locales.debug'), {
+    service: () => {
+      return {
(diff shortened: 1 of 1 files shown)


packages/kuma-gui/src/app/kuma/debug.ts
```diff
@@ -10,6 +10,21 @@ const $ = {
...
 export const services = (app: Record<string, Token>): ServiceDefinition[] => [
+  [token('kuma.locales.debug'), {
+    service: () => {
+      return {
+        common: {
+          product: {
+            docs: 'https://kuma.io/docs/dev',
+          },
+        },
+      }
```
johncowen, line 21: This service definition file is sourced by Cypress, which doesn't have a loader configured for `.yaml` files, so this means I can't use `import {...} from '....yaml'` here.

I figured I could just not use yaml here seeing as its one little value to get around this.

PR #4253, opened 2025-09-17
refactor: move eslint and stylelint to esm

schogges:
Moves both `eslint` and `stylelint` configs to esm.

### File extensions
The migration to esm requires file extensions on all relative import specifiers and no more folder imports. Imports from packages remain untouched if they can be resolved to file entry points specified in the respective manifest. This means that if files are exposed via the `files` entry in the manifest, importing them requires to add the file extension (i.e. `cypress-fail-fast/plugin.js`). Also refer to https://nodejs.org/api/esm.html#mandatory-file-extensions

### @vue/eslint-config-typescript
Both `defineConfig` and `vueTsEslintConfig` are deprecated and I've taken the chance to use the correct functions/configs: `defineConfigWithVueTs` and `vueTsConfigs`.


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@kumahq/kuma-gui-monorepo",
   "private": true,
   "version": "0.1.0",
+  "type": "module",
```
schogges, line 5: Note: Turns the root workspace into module to save some time for node to figure out which module type loader to use (otherwise determines by using `import` vs `require`)

johncowen (approved):

Issue #3710, opened 2025-03-27
discuss and possibly tweak release cycle branching process


johncowen:
### Description

See discussion here https://github.com/kumahq/kuma-gui/pull/3705

At the very least I would like to figure out if I'm doing something wrong in regards to maintaining individual commits. If this isn't possible currently, I'd like to discuss tweaking the process slightly so we can do this.

---

Let's create a workflow "branch cut" that does:

- [x] run npm update or otherwise make sure transitive dependencies are updated (neither dependabot nor ~renovate do this~ we've configured renovate to allow this to be done via a checkbox https://github.com/kumahq/kuma-gui/pull/4028)
- [ ] ~bump package.json for application packages~
- [ ] ~Ideally I'd like this to cut and push a new branch, but ideally it should have some sort of approval on that branch, open to discussion how.~
- [x] We should have a release checklist for manual steps when cutting the release branch (see https://github.com/kumahq/kuma-gui/pull/4278)

Also in parallel:

- [x] change active-branches.json to be synced from Kuma as a cron or equivalent. See https://github.com/kumahq/kuma-gui/pull/4262
- [ ] look at adopting: https://github.com/kumahq/kuma/blob/master/.github/workflows/merge-release-to-master.yaml to automatically merge release to master during stabilization. ~edit: we've discussed possibly just cherry-picking instead~ We are gonna use backends workflow linked here

lukidzi:
> Triage: We should have something like https://github.com/kumahq/kuma/blob/master/.github/workflows/merge-release-to-master.yaml, also we shouldn't squash. We should add a step to release process to enable merge option and job on kuma-gui

johncowen:
We should also add things that are specific to us here, like automatically bumping package files etc and running `npm update` once we start the new cycle.

I'm going to use this issue to track the automation of these things instead of https://github.com/Kong/kong-mesh-gui/issues/1769 (which I'm going to close)

lahabana:
Now that we're using renovate we should do this.

lahabana:
As we're about to jump on a new release. It's probably a good time to document the things we do at the start of the release.

Putting this at In progress for 2.12 to make sure we keep this one in mind

johncowen:
Currently adding a list of things by editing the PR description

schogges:
> run npm update or otherwise make sure transistive dependencies are updated (neither dependabot nor renovate do this)

What do you exactly mean? 🙂

johncowen:
[...]

PR #4262, opened 2025-09-18
chore(ci): add post release / kuma sync workflow

johncowen:
I plan on making a workflow dispatch only workflow for us to use immediately after a kuma release.

It should at least:

- [x] sync the active-branches.json from the upstream repo
- [ ] depending on what we are doing regarding backports etc, make sure things are in sync (if we don't need to sync back we don't need to ensure we do this one)
- [x] ideally automatically do a renovate package-lock maintenance - but we are already semi-automatic with renovate anyway, so maybe just some sort of message in the PR we create
- [x] create a PR with the synced active-branches.json

Work towards https://github.com/kumahq/kuma-gui/issues/3710

---

edit:

I kinda struggled with the naming of this seeing as `kuma-gui` does a release of its code on every PR merge, whereas `kuma` does a release every few months. They are very different things and I didn't want to confuse things.

In the end I called this "Sync kuma-gui from kuma" which is super clear as to what this currently does.

I plan on merging this with only the above two check boxes checked because:

- its not clear to me what we want to do with backports. Judging by the amount of changes we need to do and then sync to another branch during stabilisation, personally I would rather cherrypick the odd change we make instead of doing a total sync after possibly several weeks. Waiting for a total sync to happen weeks later can sometimes get in the way of work on `master` so I'd prefer not to do that. Happy to set up some sort of [automation](https://github.com/kumahq/kuma/blob/f2128424a4ae2276f597cb4069329e4d9b5dc1f1/.github/workflows/backport.yaml) separately to start this off for us if this is what we settle on.
-  Not totally sure if from this workflow we can tell renovate to do something, maybe we just add something to the description of the PR made here with a link to the dependency dashboard and some text saying "[ ] Ask renovate to perform maintenance on the lock file" **edit**: we added this

Let me know your thoughts on the above, other than that I think the only way to test this properly is to temporarily manually change our active-branches.json, merge to main, then run this to see if it updates as intended. I'd also be curious to know what happens if there are no changes here, I would guess it would just no make the PR.

### DIFF
--- a/.github/workflows/sync-with-kumahq-kuma.yaml
+++ b/.github/workflows/sync-with-kumahq-kuma.yaml
@@ -7,11 +7,11 @@ concurrency:
 
 on:
   # Uncomment the following pull_request event to test this workflow in a PR
-  pull_request:
-    types:
-      - opened
-      - reopened
-      - synchronize
+  # pull_request:
+  #   types:
+  #     - opened
+  #     - reopened
+  #     - synchronize
 
   workflow_dispatch: {}
 


.github/workflows/sync-with-kumahq-kuma.yaml
```diff
@@ -0,0 +1,67 @@
...
+  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+
+on:
+  # Uncomment the following pull_request event to test this workflow in a PR
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
```
johncowen, line 14: This will need commenting out before merge.

Issue #4237, opened 2025-09-11
Align subscriptions between zone and data-plane


schogges:
### Description

In https://github.com/kumahq/kuma-gui/pull/4150 we've introduced a different way to show XDS connections of data-planes. In order to make it consistent we want to align the zone's KDS connections to this new approach.
Especially showing only the active connection on the overview is a great improvement.

I've made a [comment](https://github.com/kumahq/kuma-gui/issues/4233#issuecomment-3280780247) to the another issue that may change the way of displaying the table. Once this is sorted we can fully align both views.

Update: Align with what is tbd in https://github.com/kumahq/kuma-gui/issues/4264

PR #4268, opened 2025-09-22
feat(data-planes): relocate xds connections to new tab

schogges:
Moves the `XDS connections` to a new tab. With having in mind that we want to align it with a zone's `KDS connections` (https://github.com/kumahq/kuma-gui/issues/4237), I've created a new view `SubscriptionsListView` in `subscriptions`.


johncowen:
Just wanted to check and make sure you knew this needs to happen in Dataplanes, Zones _and_ Zone Ingresses/Egresses (all re-using as much as possible of the subscriptions module)

schogges:
> LGTM, guessing you are gonna implement the other modules in the same way, and there might be some tweaks here later on also.

Yes, might need some more tweak in follow-ups 😄

Issue #2876, opened 2024-09-06
figure out a way to keep openapi/ts up to date automatically


johncowen:
### Description

We've recently started using OpenAPI specs to auto generate our TS types.

We've added a make target to generate these for us, whilst this is one command, we have to remember to do this, and we also have to remember to update the upstream repository containing the OpenAPI specs.

Ideally we'd automatically know if the types had changed instead of having to remember to `git pull` on `kumahq/kuma` and run `make build/types`.

PR #4275, opened 2025-09-24
chore(kuma-http-api): workflow for auto-sync

schogges:
This is the first part of having automatic updates on our OAS based types from kuma.
The approach of leveraging `repository_dispatch` events requires a workflow in kuma that runs on every push to the base branch. In case there is a diff between HEAD^ and HEAD in `docs/generated/openapi.yaml` it should perform a repository dispatch to the kuma-gui repo, which will then be picked up by this new workflow.

Part of https://github.com/kumahq/kuma-gui/issues/2876

### DIFF
--- a/.github/workflows/sync-kuma-oas.yaml
+++ b/.github/workflows/sync-kuma-oas.yaml
@@ -7,6 +7,12 @@ on:
         required: true
         type: string
         description: The base branch for which to create a PR
+  # Uncomment the following pull_request event to test this workflow in a PR
+  # pull_request:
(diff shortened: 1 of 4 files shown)


schogges:
> Could we add some sort of debugging code here so we can dry-run things during a PR? I think I added similar in the new sync-kuma-kuma-gui.yaml workflow if you wanted to use that as a starting point

Good point 👍 I've added a similar "debug mode" and also realized that I don't need to add the extra job for bootstrapping when there is only one other job.

PR #4276, opened 2025-09-24
chore: integrate XDl into DataPlaneDetailView

johncowen:
https://github.com/kumahq/kuma-gui/pull/4228 added a new XDl component, one of the reasons for having this was to have a simpler/more-grokkable way to author `dl` type lists.

One area of code where I thought it might prove helpful was our new DataPlaneDetailView.vue, so I figured I'd do an initial rollout here to get a feel for it. Once I'd done this I kinda went a bit further after realizing a few things, and because of this I've split this PR up into commits so its easier to see the other refactors/reorgs.

All in all we've discussed this page a little in regards to it being a little more grokkable and I wanted to see if it could be improved without resorting to splitting things up into a number of smaller single-use somponents.

Rough list of changes:

- Move DefinitionCard usage to XDl 61a6f6412886027bcaa185facb53440b74277faa
- Moved some data loading our of this view and into the drawer where its used ee092b489fc299a20768c0782656ed19e568d93c
- Reorg/refactor the data interactions in this page using our composable DataSource/DataLoader to move the data and loaders closer to where its needed bf330813198b75257988f8638e58b3cdedf8b2d7 77cef1f95dd397055010693fab3ca9879a83a383

Its easier to do these sorts of reorgs when you aren't trying to build a new feature i.e. hindsight is everything, plus I might have totally missed something that I'm not aware of here. If necessary we can pull it back to just the DefinitionCard > XDl migration. Although I really do think all of this combined is a big improvement, and in doing this there are some other things I spotted we could do.

---

## Questions

1. Are Gateway detail pages even supported in KRI mode? i.e. do we just use the old view?

### DIFF
--- a/packages/kuma-gui/features/mesh/dataplanes/Subscriptions.feature
+++ b/packages/kuma-gui/features/mesh/dataplanes/Subscriptions.feature
@@ -44,6 +44,7 @@ Feature: dataplanes / subscriptions
     Then I click the "$about-dp-subscriptions a" element
     Then the URL contains "/meshes/default/data-planes/backend/subscriptions"
     And the "$dp-subscriptions" element exists
+    And I wait for 500 ms
     Then I click the "$dp-subscriptions table tbody tr a" element
     Then the URL contains "/meshes/default/data-planes/backend/subscriptions/subscription/bar"
     And the "$dp-subscription-summary" element exists
--- a/packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
+++ b/packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
@@ -1,3 +1,4 @@
+import { Kri } from '@/app/kuma'
 import type {
   DataplaneGateway as PartialDataplaneGateway,
   DataplaneInbound as PartialDataplaneInbound,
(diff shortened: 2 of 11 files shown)


packages/kuma-gui/features/mesh/dataplanes/Subscriptions.feature
```diff
@@ -44,6 +44,7 @@ Feature: dataplanes / subscriptions
     Then I click the "$about-dp-subscriptions a" element
     Then the URL contains "/meshes/default/data-planes/backend/overview/subscriptions"
     And the "$dp-subscriptions" element exists
+    And I wait for 500 ms
```
johncowen, line 47: I added one of these as I know all this subscriptions stuff is changing so I just add this in here for now, although I suspect there could be something weird going on with the fact that we are linking to a drawer from a drawer (in the same drawer position), something that we don't do anywhere else.

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -27,15 +28,33 @@ export type DataplaneGateway = PartialDataplaneGateway & {}
...
+        }
+      }),
+      outbounds: dataplaneNetworkingLayout.outbounds.map(item => {
+        const kri = Kri.fromString(item.kri)
+        return {
+          ...item,
+          stat_prefix: `${item.proxyResourceName.slice(0, item.proxyResourceName.lastIndexOf('_'))}_${item.port}`,
+          portName: kri.sectionName !== String(item.port) ? kri.sectionName : undefined,
+        }
+      }),
```
johncowen, line 57: Apart from what we spoke about yesterday during our pairing session, I changed this bit and related also.

- I changed the name of `proxyResourcePortName` so `stat_prefix`. The fact that the name had `PortName` in it but was specifically to change things to use the `port` rather than the `portName` confusing me a lot. Envoy uses `stat_prefix` for this i.e. the mapping you can use to access/lookup the stats, and this is exactly what we use this for, so I went with that. I actually think we should probably keep that field name as its super clear what its for and what we are using it for.
- I made both inbounds and outbounds the same shape, we should always try to do this where possible.
- The above changes meant I had to change prop names elsewhere, which brought up some other questions but I can ask about those at a later date.

johncowen:
K I added one more commit on the end here 6bae4a81283f81118f1150ffaa7704f7416dbc1a which:

- copy/pastes the AboutCard into legacy (minus the policy stuff)
- copy/pastes the notifications including placing the error in the slightly different places (now in both new and legacy)

If this turns green we are now good to go again 👍

schogges (approved):

PR #4278, opened 2025-09-24
chore(docs): update RELEASE.md

johncowen:
Makes some amends to our RELEASE.md file for guidance when cutting a release branch etc

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -11,8 +11,11 @@ for the release(s) of Kuma on version 2.1).
 
 Whenever a new release branch is created in
 [kumahq/kuma](https://github.com/kumahq/kuma), a new release branch of the
-_exact_ same name should be created in Kuma GUI. New pull requests in Kuma GUI
-are opened against this branch during the stabilization phase.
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -1,11 +1,38 @@
...
+## Release tasks
 
-Once the new version of Kuma was released, changes from the release branch are to be merged back into the project’s default branch.
+### Creating the release branch
 
-## Creating the release branch
+Whenever a new release branch is created in
+[kumahq/kuma](https://github.com/kumahq/kuma), a new release branch of the
+_exact_ same name should be created in Kuma GUI. New pull requests in Kuma GUI
+are opened against this branch during the stabilization phase.
```
schogges, line 15: All PRs or just the ones that are meant to go into the release? So the release branch becomes the new trunk during stabilization?

johncowen: 👍 nice spot

(not that it matters now but unless I'm misremembering I think this is what we used to do, anyway lemme change this)

PR #4281, opened 2025-09-25
chore(sync-kuma-oas): minor amends

schogges:
- Uncomment `repository_dispatch`
- add concurrency to only run one sync at a time
- shorten kuma commit hash

### DIFF
--- a/.github/workflows/sync-kuma-oas.yaml
+++ b/.github/workflows/sync-kuma-oas.yaml
@@ -1,6 +1,17 @@
 name: Sync Kuma OAS
 
+# Ensures that only one workflow is run per branch at a time.
+concurrency:
+  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+
 on:
+  # Uncomment the following pull_request event to test this workflow in a PR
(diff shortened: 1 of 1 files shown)


johncowen:
Is it a hassle to uncomment the PR testing so its easier to tryout/review these?

Also super nitty question, but how come 12 and not 8 for the SHA? If theres no reason I think its 8 everywhere else.

johncowen:
Oh wait! I just saw https://github.com/kumahq/kuma-gui/pull/4280

johncowen:
This looks great to me! I checked the testing diff output and all looks good 👍 

One tiny thing, I noticed we aren't putting the SHA in the title of the PR, should we do that also?

Otherwise if you wanna comment out the PR hooks and gimme a yell, I can give this a final ✅

schogges:
> This looks great to me! I checked the testing diff output and all looks good 👍
> 
> One tiny thing, I noticed we aren't putting the SHA in the title of the PR, should we do that also?
> 
> Otherwise if you wanna comment out the PR hooks and gimme a yell, I can give this a final ✅

Ah missed that 😆 nice catch 👍

PR #4316, opened 2025-10-15
feat(XCodeBlock): XCodeBlock configurable read/write-ability

johncowen:
Extracts syntax highlighting to be injectable from the outside, we continue to provide shiki as a default highlighter for the moment.

Adds injectable readOnly and readWrite components inside of `XCodeBlock`. We don't have code readWrite ability currently in kuma so we don't want to needlessly load potential readWrite implementations (such as monaco) until we want to have readWrite ability here.

---

P.S. If you look at my commits here you'll see I changed this slightly just to make things easier. Hopefully this will help to explain the comments I've added here also.

### DIFF
--- a/packages/x/src/components/x-code-block/XCodeBlock.vue
+++ b/packages/x/src/components/x-code-block/XCodeBlock.vue
@@ -20,6 +20,7 @@
       :searchable="isSearchable"
       :show-copy-button="showCopyButton"
       :query="props.query"
+      :uri="props.uri"
       theme="dark"
       @code-block-render="handleCodeBlockRenderEvent"
       @query-change="emit('query-change', $event)"
@@ -46,6 +47,7 @@ import {
 
 const props = withDefaults(defineProps<{
   editable?: boolean
+  uri?: string
   id?: string
   code: string
   language: 'json' | 'yaml' | 'bash'
@@ -57,6 +59,7 @@ const props = withDefaults(defineProps<{
   isRegExpMode?: boolean
 }>(), {
   editable: false,
+  uri: '',
   id: () => `x-code-block-${useId()}`,
   isSearchable: false,
   showCopyButton: true,


packages/x/src/components/x-code-block/XCodeBlock.vue
```diff
@@ -72,33 +71,21 @@ const emit = defineEmits<{
...
   // we can ignore eslint no-unsanitized/property as all code content is stringified and shiki adds safe HTML for highlighting.
   // eslint-disable-next-line no-unsanitized/property
-  codeElement.innerHTML = (await codeHighlighter).codeToHtml(code, {
-    lang: language,
+  codeElement.innerHTML = (await syntax()).codeToHtml(code, {
+    lang: props.language,
     themes: {
-      light: 'catppuccin-latte',
+      light: 'catppuccin-mocha', // both dark for now at least
       dark: 'catppuccin-mocha',
```
schogges: What is the reason to only use `catppuccin-mocha` for now at least? 🙂

schogges (dismissed):
Left a Q, the rest LGTM 👍

packages/x/src/components/x-code-block/XCodeBlock.vue
johncowen: Unless I'm mistaken we don't have a switch for this i.e. we only ever use the dark theme (you have to specify something for `light:`)

packages/x/src/components/x-code-block/XCodeBlock.vue
```diff
@@ -1,53 +1,53 @@
...
 import { ref, useId } from 'vue'
 
-import type { CodeBlockEventData } from '@kong/kongponents'
+import {
+  useSyntaxHighlighter,
+} from '@kumahq/x'
 
 const props = withDefaults(defineProps<{
+  editable?: boolean
+  uri?: string
```
schogges: I think it would be great if we start the habit to add short jsdoc comments to each prop that explains what it does or what it's for. I assume this one (`uri`) is about https://microsoft.github.io/monaco-editor/docs.html#classes/Uri.html?

PR #4335, opened 2025-10-24
chore(data): create @kumahq/data

johncowen:
Does the bare minimum to move the majority of our `Data*` components and related into their own `@kumahq/data`.

There are loads of little bits I would like to do here still, but I'd rather merge this as is and then pick of all these bits as time goes on. There's likely to be quite a few little PRs at some point in the future to do this, but for now this at least creates the new package.

---

With the current work that I'm doing I keep bumping into places where I'm like "oh I wish I'd already built `<thing>`, or I wish I'd cleaned up DataSource so I can do `<thing>`".

I decided to bite the bullet with this and the first step is to create our planned `@kumahq/data` package


packages/config/tsconfig/browser.json
schogges: What's the reason for adding this? It looks almost the same as `app.json`. Maybe we just rename `app.json` to `browser.json`?

johncowen: I just took another look at `app` and noticed it has `cypress` types which we don't want here (but we do what for an "app").

I think this is where the composable tsconfig approach would work because we'd be able to use different configs for different files in a more fine grained approach.

Umm, kinda not sure what to do here, I'm kinda fine doing anything so we can get this merged so lmk what you want to do. I'm good leaving as is or changing to whatever.

schogges: IIRC the `composite` feature doesn't work well in a monorepo 🤔 at the very least we've identified another case and maybe we just have this new `browser` as the base of the `app` and `app` just adds the additional types?
Maybe something for a separate PR, lmk 🙂

schogges (approved):
LGTM

PR #4359, opened 2025-11-06
chore: adds new KumaTargetRef component

johncowen:
New KumaTargetRef component for rendering a TargetRef consistently in several different places.

I had to amend XDl slightly as its possible to have a single `dt` (in the case of `Mesh` as opposed to `MeshService: name`)

---

Note: this may be tweaked slightly in the future, but for now it does what I need it to do, including having a slot embedded incase we need to put anything at the end of the badge.


johncowen:
Tests are failing here which looks like for good reason, so just converted to draft whilst I look into it and fix up

Issue #4358, opened 2025-11-05
Pagination queryParams are being passed to details pages unnecessarily


johncowen:
### Kuma Version

2.12.x

### Describe the bug

In at least ServiceInsight listings the pagination query params are also being passed through to the detail page, meaning any detail page "sub-listing" will "begin" on the same page you were on on the previous listing page.

So it sounds like we need to remove things like this:

https://github.com/kumahq/kuma-gui/blob/435eb508638dd92cb60b4a5e503ef50b0b013b99/packages/kuma-gui/src/app/services/views/ServiceListView.vue#L76-L79

But we should also check other list / detail combos and make sure these query params aren't being passed through elsewhere.

slonka:
> Triage: let's fix it. We need to check other pages

PR #4364, opened 2025-11-13
fix: don't pass query params from list to detail views

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/4358

See above issue for details

I noticed that the only places where we did this wrong was where we had a list view with no summary view. Summary views do need these query params passing to them, whereas if the list view links straight to a detail view, then they don't need the query parameters passing to them.

I then also check zone and mesh listings (neither have summary views) and noticed the parameters needed removing there also.


bartsmykla (approved):
Looks good, detail views don't use these params anyway.

Issue #4106, opened 2025-07-16
Show loaders on individual features


schogges:
### Description

A current and more or less common approach is to have a `DataLoader` on a parent route (i.e. `*TabsView.vue`) and pass the fetched data as props to the dynamic child routes. This approach causes certain issues:

- Typesafety: due to the dynamic routing we lose the typesafety on the props. This is a general problem with the dynamic routes and is not directly an issue with the loaders, but the solution of this issue would resolve this problem. So we can hit two birds with one stone here.

```vue
<!-- The parent route -->

<!-- DataSource is non-blocking -->
<DataSource
  ...
  v-slot="{ data }"
>
  <RouterView
    v-slot="{ Component }"
  >
    <!-- at this point `data` is potentially undefined, I expect an error because of the defined props in the child -->
    <component
      :is="Component"
      :data="data"
    />
  </RouterView>
</DataSource>
```

```vue
<!-- The child route -->
<script lang="ts" setup>
const props = defineProps<{
  data: DataplaneOverview // <-- expects to be defined
}>()
</script>
```

- A loader is shown on the entire route. We have different loader variants (spinner, skeleton, ...) and it would be great to display the variants depending on the type of feature that is about the be rendered once the data is fetched.
- Single responsibility: Those parent tabs routes have the responsibility to render the tabs based on the sub-routes of the router. The responsibility to fetch data that is going to be displayed should be moved to where the data is being used. This also means less dependency between parent and child and easier changes (replacement, addition, deletion of sub-routes).

The idea is now to remove the data fetching from the wiring routes as much as possible and move the data fetching closer to the actual features - move it into the sub-routes. Due to the shared `DataSource`/`DataLoader` HTTP requests, this should not cause a tremendous increase of requests.

Also refer to the first approach and the comments in https://github.com/kumahq/kuma-gui/pull/4058. The approach in there didn't work well because it meant a lot of hacky if-statements. So the probably better approach is to tackle this bottom-up and remove the props on the child routes first.

johncowen:
> triage: We agreed that this has been in "Todo" for a few weeks now, it has a low priority so we've decided to take it out of Todo just for the moment (its still in the same milestone, but it wouldn't matter if it slipped)

PR #4365, opened 2025-11-13
chore: DataLoader amends

johncowen:
This PR is part of https://github.com/kumahq/kuma-gui/issues/4106. Diff best viewed with Github's `Hide whitespace` setting.

---

This PR adds 2 things to our much used `DataLoader` component

- Allows non-blocking behaviour if you don't specify data. This allows you to specify only errors. The DataLoader will not block and show its default slot _unless_ there is an error. In which case it will show the error instead of the default slot.
- More of an ergonomic change: The `:data="[]"` prop can now contain errors. On the inside we check `:data` for errors and split them out into our internal `errors` variable. This allows us to load/error on single properties that are of type `Result | Error`. This assumes (a reasonable assumption) that data result will never actually be an `Error` its always some other actual data. Whilst we only absolutely need this is rare cases, its actually ergonomically better to not have to manual give both `:data` and `:error` properties, so just be able to give a single `:data` property may become the default way we use DataLoader in the future.

---

Using the above changes I've made some amends to the homepage to allow us to show loaders inside of our XCards (inside our grey borders)

## Before

Screenshot 2025-11-13 at 14 34 50

## After

Screenshot 2025-11-13 at 14 38 37

### DIFF
--- a/packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
+++ b/packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
@@ -5,53 +5,110 @@
         <h2>{{ t('main-overview.detail.about.title') }}</h2>
       </div>
     </div>
-
-    <XLayout
-      type="columns"
-      class="columns-with-borders"
+    <DataLoader
+      :data="[props.globalInsight]"
     >
-      <ResourceStatus
-        v-if="props.canUseZones"
-        :total="props.globalInsight.zones.controlPlanes.total"
-        data-testid="zone-control-planes-status"
-      >
-        <template #icon>
-          <img
-            class="icon"
-            src="@/assets/images/zone.svg?url"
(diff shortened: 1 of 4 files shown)


packages/kuma-gui/src/app/control-planes/components/ControlPlaneStatus.vue
```diff
@@ -5,133 +5,141 @@
...
           </template>
 
+          <template #title>
+            {{ t('main-overview.detail.about.meshes') }}
+          </template>
+        </ResourceStatus>
+        <template
+          v-for="insight in [props.globalInsight]"
+          :key="typeof insight"
+        >
```
johncowen: I had to add this so that Vue understands that globalInsight is now guaranteed to be set and not an error, previously for some reason Vue couldn't figure that out from within the reduce statement.

PR #4367, opened 2025-11-14
feat(RouteView): adds route.go functionality

johncowen:
Previously we only had `router.back()` which would always go one step back in the Vue routing history (if no history existed we also provide a default route name to go to instead)

This PR adds `router.go()` with an extra `delta` property on the `to` object.

This allows us to go several steps back in the Vue routing history with:

```
go({ delta: -2, name: 'mesh'})
```

...which would take us back two steps, defaulting to `mesh` if enough steps don't exist.

### DIFF
--- a/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
+++ b/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue
@@ -47,6 +47,7 @@
           replace: routeReplace,
           params: routeParams,
           back: routerBack,
+          go: routerGo,
           children,
           child,
           from: getFrom,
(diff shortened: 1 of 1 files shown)


bartsmykla (commented):
Hey, I think there's an issue with the history depth check here:

https://github.com/kumahq/kuma-gui/blob/8eb329dafd8372e707d489c891ab15b3360eb298/packages/kuma-gui/src/app/application/components/route-view/RouteView.vue#L277-L287

The check `win.history.state.back !== null` only tells us there's at least one step in the history, but doesn't verify we have enough steps for the requested delta. 

So if someone calls `go({ delta: -3, name: 'home' })` but there are only 2 steps in the history, `router.go(-3)` will silently do nothing and return early, which means the fallback `routeReplace(args)` never gets called. Users could get stuck on the page instead of falling back to the provided route.

This breaks the safety mechanism from the original `routerBack` implementation (commit 101b3c56) that was specifically added to prevent users from getting stuck when navigation fails.

bartsmykla (approved):

PR #4368, opened 2025-11-14
feat(x-card): add new XCard component

johncowen:
Previously our XCard component was a straight alias to KCard.

We now need to add custom styles to our XCard only if the XCard is inside an XAction (a.k.a. an anchor tag)

This PR creates our own XCard component and passes through to KCard, which allows us to add the conditional custom styling depending on whether we are inside an XAction.


johncowen:
> Should XCard forward attributes to KCard using v-bind="$attrs"? I noticed other wrapper components in the x package (like XTabs and XAction) use defineOptions({ inheritAttrs: false }) and v-bind="$attrs" to forward attributes like class and data-testid to their wrapped components.

Good question! 👍 If you don't use `defineOptions({ inheritAttrs: false })` they automatically get passed through to the top level component, which in this case is KCard, so everything should work as expected

PR #4398, opened 2025-12-04
chore(ci): move husky prepare script to Makefile

johncowen:
We decided we'd rather not use `npm` for calling scripts and such-like, and I spotted that we still had `prepare` in our package files.

This PR:

1. Moves the call to `husky` into a `git/hooks` target
2. Rewrites our `resolve/bin` target which we use to find node binary scripts (it couldn't resolve husky properly). I realized I could use `npm query` to find the `bin` from the package-lock file instead of the packages `package.json`. Potentially changing this script could cause a lot of issues if its wrong as a lot of our tooling depends on it, everything seems fine to me with this change though.
3. Whilst noodling around with this a found an easy way to list all dependencies with postinstall scripts, which is probably useful so I also added a smol target here for this.
4. Now that all references to `husky` and `lint-staged` are now in `@kumahq/config` I moved the dependencies there instead of in the root workspace package.json - I'm guessing any consumers of `@kumahq/config` can also do the same in their repos and no longer directly depend on these packages


johncowen:
> Potentially changing this script could cause a lot of issues if its wrong as a lot of our tooling depends on it, everything seems fine to me with this change though.

Spoke too soon 😆 , leaving this in Draft until I look at the test fail

Issue #4410, opened 2025-12-09
Lets adopt kong's shared renovate


johncowen:
See https://github.com/kong/shared-renovate

We need to be conscious of not adding lots of groups, that aren't fine grained enough to not block certain upgrades (i.e. we don't want groups at all ideally)

PR #4411, opened 2025-12-10
fix(deps): remove majority of renovate grouping

johncowen:
Related: https://github.com/kumahq/kuma-gui/issues/4410

Our renovate groups quite often/can end up being "block-y" because if you have a group of 10 dependencies to be upgraded and for one reason 1 of those can't/shouldn't be upgraded, then the other 9 are blocked.

We've decided to remove the groups for now at least.

Note: I've also removed the `minimumReleaseAge` as this is not quite doing what we want it to do (it doesn't effect transitive dependencies) and is causing PRs titles/descs to have mismatching versions from the actual changes it is making.

In the meantime we should consider the `age` that renovate adds to the PR descriptions - PR CI will still run, but then there is a human review to get past which would spot any strange additions/amends and we don't expose any tokens anywhere in CI.

Lastly we are soon to look to move to kong's shared renovate anyway, so I guess all this will be changed.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -38,26 +38,6 @@
       "dependencyDashboardApproval": true
     },
     {
-      "groupName": "patch",
-      "matchUpdateTypes": [
-        "patch"
-      ],
-      "minimumReleaseAge": "3 days"
-    },
-    {
-      "groupName": "minor",
-      "matchUpdateTypes": [
-        "minor"
-      ],
-      "minimumReleaseAge": "3 days"
-    },
-    {
-      "matchUpdateTypes": [
-        "major"
-      ],
-      "minimumReleaseAge": "3 days"
-    },
-    {
       "matchPackageNames": [
         "npm",
         "escape-string-regexp"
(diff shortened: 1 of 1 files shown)


schogges:
> > Note: I've also removed the minimumReleaseAge as this is not quite doing what we want it to do (it doesn't effect transitive dependencies) and is causing PRs titles/descs to have mismatching versions from the actual changes it is making.
> 
> `minimumReleaseAge` from what I can tell does not affect transitive dependencies so its almost pointless.

Sorry, seems I haven't properly read the description. I think we are not receiving renovate PRs for transitive dependencies other than the ones we're overriding. Need to check the `minimumReleaseAge` behaviour for the ones we are overriding 🤔 

> The other thing I've noticed is, it seems like renovate will open a PR saying its updating to say `1.10.1` but as its late another release has come out, so the upgrade can sometimes actually be `1.10.3`. It would be good to confirm exactly its behaviour here as I'm a bit fuzzy on exactly what this looks like

That would surprise me and I also would like to check this, because this sounds like not intended behaviour by renovate 🤔

johncowen:
Just for history, we had a good catch up on this and decided to merge as we are working on a different approach to this anyhow

PR #4420, opened 2025-12-12
feature(kuma-http-api): add custom overlayer that can deref during overlay

johncowen:
Adds a smol wrapper around https://github.com/thim81/openapi-format (which is the overlay lib/CLI we are using) to allow us to dereference `$ref`s _before_ the overlay happens.

I decided to add our own custom CLI instead of composing CLI calls using `make` because I wanted to avoid making lots of intermediate files though the pipeline. This way you have one overlay source file and a single generated OpenAPI spec.

I've made some inline notes to help review, but also note there are a few formatting changes just due to the multiple bits of serializing/deserializing happening as the YAML goes through the pipeline.

~Lastly, we can't really merge this as yet as we need the upstream issue resolved - worst case scenario we are going to have to patch or fork the upstream lib - or potentially find another overlay tool/lib that we can use. Leaving in Draft for the moment, but a totally open for review and potential informal approval at this stage 👍~

Upstream bugfix is in, I've added a couple of commits on the end of here for the upgrade of `openapi-format` and then one more `make build` - [packages/kuma-http-api/dist/components/schemas/MeshAccessLogItem_Put_RequestBody.yaml](https://github.com/kumahq/kuma-gui/pull/4420/changes#diff-9def993296fca062098e47f80d079e0c65813832713efcbfc6fae4157461f549) is now how we want it 🎉


packages/config/src/mk/openapi.mk
```diff
@@ -1,9 +1,14 @@
...
+OPENAPI_BIN_DIR := $(dir $(abspath $(OPENAPI_MAKEFILE_DIR)))bin
+
+NODE:=node
+OPENAPI_OVERLAY:=$(NODE) $(OPENAPI_BIN_DIR)/openapi-overlay.ts
 
 OPENAPI_TS ?= $(shell $(MAKE) resolve/bin BIN=openapi-typescript)
 OPENAPI_FORMAT ?= $(shell $(MAKE) resolve/bin BIN=openapi-format)
 
-OVERLAY_SRC_FILES := $(shell find ./src -name "*.overlay.yaml" -type f)
+OVERLAY_SRC_FILES := $(shell find ./src -name "*.overlay.yaml" -type f | sort)
```
johncowen: The `sort` here makes sure that the almost-final `openapi.overlay.yaml` is the last thing to be overlaid.

packages/config/src/bin/openapi-overlay.ts
```diff
@@ -0,0 +1,116 @@
...
+    return prev
+  }, {} as Record<string, unknown>) as T
+}
+
+const opts = {
+  keepComments: true,
+  bundle: false,
+}
+
+const __ = process.argv.reduce((prev, flag, i, arr) => {
```
schogges: Is there any particular reason for naming this `__`? How about `const options = ...`?

johncowen:
FYI Added this to the desc here:

> Upstream bugfix is in, I've added a couple of commits on the end of here for the upgrade of `openapi-format` and then one more `make build` - [packages/kuma-http-api/dist/components/schemas/MeshAccessLogItem_Put_RequestBody.yaml](https://github.com/kumahq/kuma-gui/pull/4420/changes#diff-9def993296fca062098e47f80d079e0c65813832713efcbfc6fae4157461f549) is now how we want it 🎉 

Taking out of draft 

Thanks again @thim81 !

Issue #3093, opened 2024-10-24
convert PolicyTypeList to use an unordered list


johncowen:
### Description

Tiny nit perhaps, but semantically it would be better. Plus if it was a `ul li` I wouldn't have hit the issue I found in https://github.com/kumahq/kuma-gui/pull/3092#issuecomment-2434978835

![Screenshot 2024-10-24 at 12 11 24](https://github.com/user-attachments/assets/1ea93a80-c1e0-4375-a175-6162d713413d)

bartsmykla:
> Triage: @johncowen is it still relevant?

johncowen:
still relevant, still super low hanging fruit, still super not urgent, although...

is a good first issue, will add the label, but also a good easy/hopefully quick one for either of us to do just for the sake of being able to close the issue.

johncowen:
This is related to some other work I'm doing, so I'm pulling it into our board and I'll can PR against this issue also

PR #4425, opened 2025-12-15
chore: refactors CSS underlining

johncowen:
Our CSS hover underlining is a little overzealous. We've discussed moving this into XLayout and making it a little more targeted so it doesn't put underlines on things we don't want it to by default.

I had to tweak some of the CSS in the PolicyTypeListView also so I figured I may as well pick up https://github.com/kumahq/kuma-gui/issues/3093 while I was there.

I also "finally" added a `variant` property to `XLayout` and deprecated `type`. Moving forwards we should try to use `y-stack`, `x-stack`, whereas `separated` should only be used for laying out inline lists of words (where you would commonly use a comma to separate)

Closes https://github.com/kumahq/kuma-gui/issues/3093

### DIFF
--- a/packages/x/src/components/x-theme/XTheme.vue
+++ b/packages/x/src/components/x-theme/XTheme.vue
@@ -179,7 +179,7 @@
   }
 
   :where(p, td, dd, li) a:where(:hover, :focus) {
-    text-decoration: underline
+    text-decoration: underline;
   }
 
   /* icons */


packages/x/src/components/x-theme/XTheme.vue
```diff
@@ -170,6 +170,19 @@
...
+  & {
+    --x-anchor-text-color: #{$kui-color-text-primary};
+  }
+
+  a {
+    color: var(--x-anchor-text-color);
+  }
+
+  :where(p, td, dd, li) a:where(:hover, :focus) {
+    text-decoration: underline
```
schogges: typo:


```suggestion
    text-decoration: underline;
```

Issue #4401, opened 2025-12-05
Use envoy stats to show expiration date of externally managed certs


johncowen:
### Description

> As a separate "feature" as @lukidzi suggested we could use the envoy stats to show the expiration date in the TLS view in the detail page. This could be split off as a separate issue, and as its not a bug as such could be in a different milestone.

Taken from https://github.com/kumahq/kuma-gui/issues/4384#issuecomment-3581927828

PR #4427, opened 2025-12-15
feat(dataplanes): with spire/spiffe conditionally show cert expiration

schogges:
With SPIFFE/SPIRE we get the information about certificate expiration as part of the envoy stats. There are entries for each service, but since it's one `MeshIdentity` the expiration time of each entry is always the same i.e. from a local deployment

```
cluster.kri_msvc_default_default_kuma-demo_demo-app-v1_5050.ssl.certificate.spiffe://default.local-zone.mesh.local/ns/kuma-demo/sa/default.expiration_unix_time_seconds: 1765831281
cluster.kri_msvc_default_default_kuma-demo_demo-app-v2_5050.ssl.certificate.spiffe://default.local-zone.mesh.local/ns/kuma-demo/sa/default.expiration_unix_time_seconds: 1765831281
cluster.kri_msvc_default_default_kuma-demo_demo-app_5050.ssl.certificate.spiffe://default.local-zone.mesh.local/ns/kuma-demo/sa/default.expiration_unix_time_seconds: 1765831281
cluster.kri_msvc_default_default_kuma-demo_kv_5050.ssl.certificate.spiffe://default.local-zone.mesh.local/ns/kuma-demo/sa/default.expiration_unix_time_seconds: 1765831281
listener.self_inbound_dp_5050.ssl.certificate.spiffe://default.local-zone.mesh.local/ns/kuma-demo/sa/default.expiration_unix_time_seconds: 1765831281
```

To get one of the entries I've added a `findDeep` utility function that recursively tries to find a key within an object. If that key is found it returns the value. This is needed because the parsing of the stats returns an arbitrary JSON blob and the location of a spiffe related entry of `expiration_unix_time_seconds` is unknown.
Since there might be multiple entries that have `expiration_unix_time_seconds` where some of them are related to SPIFFE and some are not, I've added a nested `findDeep` to first find an entry by RegExp `spiffe://` and then `expiration_unix_time_seconds`.
To pass the `expiration_unix_time_seconds` to the view I've added a top level `$meta` field that stores the information. If the value is not found (`undefined`) the information remains hidden.
Also note that `expiration_unix_time_seconds` is seconds, so I'm making sure that I multiply by `1_000` to turn it into milliseconds.

Note: Best to review with hidden whitespaces

Closes https://github.com/kumahq/kuma-gui/issues/4401


packages/kuma-gui/src/app/connections/data/index.spec.ts
```diff
@@ -59,7 +65,21 @@ describe('ConnectionCollection', () => {
...
-            tcp: {},
+            tcp: {
+              ssl: {
+                certificate: {
+                  'spiffe://default': {
+                    'local-zone': {
+                      mesh: {
+                        'local/ns/kuma-demo/sa/default': {
+                          expiration_unix_time_seconds: 1765820805,
+                        },
```
johncowen: How difficult would it be to do the same thing we do with ip address here?

i.e. for ips we end up with `obj['127.0.0.1'].stuff.here = 0`

would be good if we could end up with:

```
obj['spiffe://default.local-zone.mesh.local/ns/kuma-demo/sa/default'].stuff.here = 0
```

I would guess anything with `://` in it should be treated like this (`://` is impossible to add as a user so its safe to assume its some sort of uri added by internal means)

johncowen (commented):
I left one question for now, I'm gonna check this out locally and take a closer look. One more question:

Are you looking to get this into 2.13?

schogges:
> I left one question for now, I'm gonna check this out locally and take a closer look. One more question:
> 
> Are you looking to get this into 2.13?

I think it would be good if we can get this into 2.13

packages/kuma-gui/src/app/connections/data/index.spec.ts
johncowen: Ok so we had a little pair on this and it looks like its not currently possible because Kubernetes ServiceAccounts can have `.`s in them which means we could have a spiffe uri like:

`spiffe://default.local-zone.mesh.local/ns/kuma-demo/sa/default.stuff`

Currently its impossible for us to know at what point the spiffe uri ends.

Lets chat with backend about possibility of escaping `.`s  somehow, if we are using `.`s as separators then any values within the separators should have their `.`s escaped.

johncowen: Just to share thoughts: The other thing I've noticed is (which might not be true), if the `spiffe://` is always the second to last segment we could use that as the rule to parse out the spiffe as a full uri.

packages/kuma-gui/src/app/connections/data/index.ts
```diff
@@ -220,7 +222,13 @@ export const ConnectionCollection = {
       'meshtrace_opentelemetry',
     ].some(item => key.startsWith(item))))
 
+    const spiffeCertificateExpirationTime = findDeep<number>(findDeep(item, spiffeCertRe), 'expiration_unix_time_seconds')
```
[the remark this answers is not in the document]
schogges: My understanding was that it's always the same for the different services (inbounds/outbounds).

> With SPIFFE/SPIRE we get the information about certificate expiration as part of the envoy stats. There are entries for each service, but since it's one MeshIdentity the expiration time of each entry is always the same

See my description above also with the stats entries from my local deployment ☝️

PR #4428, opened 2025-12-15
chore(config): create renovate base config

schogges:
Creates a new shareable renovate base config to not require setting same options in multiple places.

By adopting the base config `Kong/public-shared-renovate//base/javascript` we've received grouping of all patches and minors. This is something we'd like to avoid so I've added the same setting but set `enabled: false`.


packages/config/src/renovate/base.json
```diff
@@ -0,0 +1,72 @@
...
+      ],
+      "matchPackageNames": [
+        "*"
+      ],
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "matchCurrentVersion": "!/^0/",
+      "enabled": false
```
schogges: I'm not entirely sure if this will work out and removes the grouping 🤞

johncowen: Q: What does the upstream dependency give us apart from grouping? (that we are undo-ing)

schogges: Some more presets and `miniumReleaseAge: 12`, which is a security setting. Additionally upstream has more control over security relevant settings in the future.

johncowen: > I'd propose upstream to remove the grouping, but until then I would like to already try to remove the grouping here.

Sounds good, I'd agree with that.

BTW, IMO `miniumReleaseAge` gives a false sense of security as it doesn't affect transitive dependencies as far as I'm aware (making it almost useless for node deps)

I think overall though, and hopefully I don't sound too negative on all this, I would have rather just stuck with the most minimal renovate possible. Adding all this config has done nothing but complicate matters with very little return 🤷 But anyway, hindsight I suppose.

Issue #4302, opened 2025-10-07
add notification to Envoy related tabs when your dataplane is offline


johncowen:
## Why

We have this issue https://github.com/kumahq/kuma/issues/5216 over in backend which has been open for quite a while. I would guess (and I could be wrong) that this generally happens when your dataplane is `offline`, something we know in the GUI

## What

We should render one of our notifications if you visit any of these Envoy related tabs saying something like:

 "your dataplane is offline so you will not get any information here, but here's the error anyway"

(obvs better wording)

We do this in the Dataplane Traffic viz already, but we should also check that we either cannot open a drawer when this happens or that (if we can) we show similar notifications.

If we do this I would imagine we can close the backend issue, unless I've totally misunderstood the original reaosn for the issue.

johncowen:
> Triage: there is another case where this can error, and that is when we are trying to get stats via several hops, i.e. via the control-plane > zone control-plane > dataplane.

> Lets do two things:

1. If the dataplane is offline. Lets either gray out or show a yellow notification
2. If we get an error when the dataplane is online, lets hardcode our own more human error into the error message. We believe the error message that comes from the backend isn't that useful for the user. Maybe check with @bartsmykla if we can easily change the backend message to be more human also.

If we do these two things we can close this issue and close the upstream backend issue (https://github.com/kumahq/kuma/issues/5216)

schogges:
> > Triage: there is another case where this can error, and that is when we are trying to get stats via several hops, i.e. via the control-plane > zone control-plane > dataplane.
> 
> > Lets do two things:
> 
> 1. If the dataplane is offline. Lets either gray out the tabs or show a yellow notification (my opinion is we show notification seeing as we do this elsewhere, whereas we never gray out tabs)
> 2. If we get an error when the dataplane is online, lets hardcode our own more human error into the error message. We believe the error message that comes from the backend isn't that useful for the user. Maybe check with [@bartsmykla](https://github.com/bartsmykla) if we can easily change the backend message to be more human also.
> 
> If we do these two things we can close this issue and close the upstream backend issue ([kumahq/kuma#5216](https://github.com/kumahq/kuma/issues/5216))

Regarding "2." I think this should be done in the backend. We've recently generally improved how we display error messages and in case we receive an unexpected error i.e. when the dataplane is online we don't know the reasons and therefore hardcoding an error message might be misleading. Though if you have any suggestions for a generic message you have in mind, please let me know 🙂

PR #4461, opened 2026-01-02
feat(dataplanes): show warning when offline

schogges:
When the dataplane is offline most of the dataplane related endpoints will return an error. Adding the warning informs the user why this is the case and also that the information we still display might be outdated.

Closes https://github.com/kumahq/kuma-gui/issues/4302


johncowen (approved):
👍

PR #4475, opened 2026-01-09
chore(routing): add more docs and tests

johncowen:
Note: The majority of this PR was AI generated followed by a pretty deep review myself followed by a fair amount of tweaking

1. Splits up utilities into `query` and `dom` files
2. documents and tests out the majority of these utils
3. removes our old unique id generator and replaces it with the Vue native one (we still keep or extra wrapper ontop of that for RouteView)
4. Some slight readability tweaks and code re-ordering I'll try and point these out inline
5. I found one issue, well AI generated passing tests for it and I saw that it was generating passing tests for a bug 😆 . Talk about people depending on your bugs! 😆  I'll point this one out inline. The bug itself is a theoretical bug not affecting our application.

Lastly there is a bit more work I wanted to do here, but as this is code fiddling and I've got other things I want to get on with I've pushed this up now.

Also sorry that the file move means that the diff isn't that useful, I was hoping it would be better, hopefully the inlines help. Let me know if this is awkward and I can try and roll back a bit and make multiple commits on the PR instead.


packages/routing/src/utils/dom.ts
```diff
@@ -0,0 +1,103 @@
...
+            prev[key].add(value)
+          }
+          return prev
+        }, prev,
+      )
+    }, { class: new Set() })
+
+    const currentClasses = new Set($el.classList).difference(originalClasses)
+    $el.classList.remove(...currentClasses.difference(flat.class))
+    $el.classList.add(...flat.class.difference(currentClasses))
```
johncowen: This function was changed to use `Sets` now we have guaranteed `Set.difference` support, the array based difference was removed.

packages/routing/src/utils/query.ts
```diff
@@ -0,0 +1,139 @@
...
+    case typeof value === 'undefined':
+    case value === false:
+      return
+    case value === true:
+      return null
+    case String(value).length > 0:
+      return encodeURIComponent(String(value))
+    default:
+      return
+  }
```
johncowen: Two things I did here:

1. I moved the `&&`ed conditional into two separate cases, just to save that split second of thought when looking at the cases it covers
2. The "bug" I mentioned in the description is fixed by the `undefined` case (line 99). Previous you could totally pass `undefined` into here and get `'undefined'` due to the `String(value)` on line 104

As I said I don't believe the bug was affecting us anywhere, I guess because we typecheck further up in `RouteView` to prevent `undefined` ever being passed in here, and since we don't use this function anywhere but in `RouteView` we are fine. But if this function were to ever be used elsewhere then thats a different story.

packages/routing/src/utils/query.ts
```diff
@@ -0,0 +1,139 @@
...
+
+
+/**
+ * Converts a normalized URL parameter value into its query string representation.
+ * Handles encoding and special boolean/`null` cases for URL query parameters.
+ *
+ * @param value - The normalized parameter value to serialize
+ * @returns Serialized value for use in query strings:
+ *   - `true` → `null` (flag-style param like `?debug`)
+ *   - `false` → `undefined` (removes flag-style param like `?debug`)
```
schogges: super nit: should this be 

```suggestion
 *   - `false` → `undefined` (removes flag-style param like `?debug=false`)
```

to point out that the flag would be removed, i.e. `?foo=bar&debug=false` -> `?foo=bar`?

PR #4478, opened 2026-01-12
refactor(hostnames,zones): show loaders/errors on individual features

schogges:
Shows loaders and errors on the individual features in hostname-generators and zones. Zone-ingresses and -egresses will follow separately.

---

This is part of https://github.com/kumahq/kuma-gui/issues/4106 and I'll cover the other parts in separate PRs.

Note: Probably best to review without whitespaces.


packages/kuma-gui/src/app/zones/views/ZoneDetailTabsView.vue
```diff
@@ -38,11 +40,7 @@
                 </XIcon>
               </template>
               <h1>
-                <XCopyButton :text="data.name">
-                  <RouteTitle
-                    :title="t('zone-cps.routes.item.title', { name: data.name })"
-                  />
```
schogges: I've removed the `RouteTitle` from here and added it to the `ZoneDetailView`. Due to the delayed rendering it caused that it overrides the route title of other sub-routes.

PR #4508, opened 2026-01-14
chore: adds XDl:y-stack and runtime XTable:justify default

johncowen:
Adds XDl::y-stack variant, and caters for some interactions with XTable

1. Adds a `y-stack` variant to XDl to allow vertically stacked Dls.
2. Automatically takes block level elements in a "`kv` table" always align to the `end`/right. This is preferable to having to wrap elements in a TagList with a right alignment for elements that are probably not tags i.e. we use the Layout component to define layout instead of the TagList component to define layout.

### DIFF
--- a/packages/x/src/components/x-table/XTable.vue
+++ b/packages/x/src/components/x-table/XTable.vue
@@ -76,6 +76,9 @@ table.variant-kv {
       + td {
         font-weight: $kui-font-weight-bold;
         text-align: right;
+        li {
+          list-style-type: none;
+        }
       }
     }
   }


johncowen:
Done. We shouldn't have bullets on right aligned text, even if we aligned the bullet with the start of the text it would look wrong because none of the bullets would be lined up. I'm not actually sure why we ever had bullets here anyway to be fair.

schogges:
So far the only alternative I found is using something like the following within the nested selector for `td` (`... th[scope="row"] + td`)

```css
> * {
  justify-content: flex-end !important;
}
```

I'd like to avoid using `!important`, but seems there is no other way to increase specificity for `*`.

Another option could be

```css
.x-layout {
  justify-content: flex-end;
}
```

This leaks class information and compared to the approach of inject/provide there is no linting to get warned if this doesn't exist anymore or has changed.

johncowen:
Ok cool, thanks for the approval! Yeah I think there are still a lot of things (aswell as what youv'e outlined above) that still make me prefer the approach in this PR.

Plus, at the end of the day we've added a one-liner here, in one place. If we don't like it later on its easy to back out and change for something else we prefer.

I'd just really like to get rid of having to wrap `<TagList alignment="right" />` around everything even though the content isn't a `Tag` and the container should already be right aligning it (but can't without us doing something extra due to technical reasons)

I'll come back for a merge in a little while 👍

PR #4560, opened 2026-02-03
chore(renovate): disable @kong-ui-public/app-layout, allow npm patches, npm version range patches

schogges:
This PR does several things:
- due to conflicts further downstream we temporarily disable renovate for `@kong-ui-public/app-layout`
- we've noticed that since the last node upgrade we have some noise in the lock-files around `peer: true|false`. That happens because we have a version constraint in the `engines` field that is a range of minors for `npm` (`npm: "^11.6.2"`). Due to that renovate uses the latest version of `npm` that satisfies, which is at the time writing `npm@11.8.0`. We just want to use and kind of enforce the version of `npm` that is being shipped with `node`, so we need a constraint. If we'd leave out the constraint renovate would also use the version that is being shipped with `node`. Therefore we:
  - set the range of `npm` versions to patches only in the constraint (`engines`)
  - allow renovate raising `npm` upgrade PRs for patches


johncowen:
Should we group this with the others at the top of the file? We could also add the description there for each, something like:

```js
description: [
  "Disabled:",
  "app-layout: because blah",
  "other-one: because blah"
]
```

schogges:
> > We could, but not at the top of the file. It should be rather at the bottom otherwise it may be overridden by other rules.
> 
> Lets do this then? I'd rather all `disabled`s were together?

Yep, I did that already 😄 however we may find that when we have to disable something else in the future we also need to add another setting, which means another separated rule anyways.

johncowen (approved):
Lets give it a go!

PR #4576, opened 2026-02-06
chore(deps): replace @modyfi/vite-plugin-yaml with our own loader

johncowen:
We found that this package wasn't being properly maintained to our standards, and also that its basically this line:

```ts
const json = load(code, { schema, filename, onWarning: (warning) => console.warn(warning.toString())})
```

So we've replaced it with our own.


schogges (approved):
👍

PR #4565, opened 2026-02-04
chore: update npm to 11.6.4

schogges:
With https://github.com/kumahq/kuma-gui/pull/4564 I was hoping to get renovate to update `npm` to `11.6.4`, but for some reason it's not picking it up. I just found the following in the logs

```
{
  "depType": "engines",
  "depName": "npm",
  "currentValue": "~11.6.2",
  "datasource": "npm",
  "commitMessageTopic": "npm",
  "prettyDepType": "engine",
  "updates": [
    {
      "bucket": "non-major",
      "newVersion": "11.8.0",
      "newValue": "~11.8.0",
      "releaseTimestamp": "2026-01-21T23:15:05.862Z",
      "newVersionAgeInDays": 13,
      "newMajor": 11,
      "newMinor": 8,
      "newPatch": 0,
      "updateType": "minor",
      "isBreaking": false,
      "isRange": true,
      "libYears": 0.2880256011225266,
      "branchName": "renovate/npm-11.x"
    }
  ],
  "packageName": "npm",
  "versioning": "npm",
  "warnings": [],
  "sourceUrl": "https://github.com/npm/cli",
  "registryUrl": "https://registry.npmjs.org",
  "homepage": "https://docs.npmjs.com/",
  "mostRecentTimestamp": "2026-01-21T23:15:05.862Z",
  "isAbandoned": false,
  "currentVersion": "11.6.2",
  "currentVersionTimestamp": "2025-10-08T20:08:50.505Z",
  "currentVersionAgeInDays": 118,
}
           
```

So I'm now manually updating `npm` to `11.6.4` to proceed with the dependency updates.

PR #4577, opened 2026-02-06
chore(deps): removes unnecessary and/or broken npm renovate config

johncowen:
We've seen weird behaviour from renovate opening and closing npm upgrade PRs, which could be caused by this config which isn't working (the PRs opened are minors)

Personally I'd like to have to config renovate as little as possible. From experience these bots never seem to work as intended or required, so I'd rather just not bother trying to find some sort of configuration that works.

Related: Previously we've wanted to just stick to whichever version of npm comes with node, but this raises the question "what if there is a security patch that the node installation isn't using yet?" - do we just wait and hope? Why shouldn't we just upgrade ourselves?

We've recently added a few tiny bits of code so its really straight forward to make sure renovate/CI (automatically) and ourselves (via a nudge) can keep `npm` versions in sync.

I also realized `npm` is just another node dependency just like all our others, apart from the fact that it is the thing we use to install those dependencies I can't think of a reason not to treat it just like every other dependency and let renovate bump within our configured range/constraint.

This PR removes the config that doesn't seem to be doing much, and I added a few more descriptions while I was here. Let me know if any of the new descriptions are wrong or need a tweak.

Once this is in and renovate has had a chance to run and we can see the result I plan on merging https://github.com/kumahq/kuma-gui/pull/4568 and closing https://github.com/kumahq/kuma-gui/pull/4565

---

Note: We realized there is no lock file for the "engines manifest", so if we use a range here there is always the tiniest of possibilities that CI could be ahead of us for a period of time. As this engines manifest isn't consumed by anything else we figured we could consider the engines property in the manifest to actually be the lock. Therefore we used an exact version for `npm` in the `engines` property.


schogges (commented):
Is it intentional to just remove the `npm` configuration about disabling `majors` and `minors`? This was the attempt to receive at least patch upgrades, but this didn't seem to work properly as you've pointed out. However just removing this turns all `npm` upgrades back on and I'd just like to confirm wether this is intended or not because I think that security updates are not prevented by disabling `npm` 🤔

schogges (approved):
👍

PR #4578, opened 2026-02-06
fix(data-planes/inbound-policies): rendering all origins for each rule

schogges:
The issue was that we were rendering all origins for each policy on each rule. The fix is to ignore the origins and just use the KRI that is part of a rule.

**Before**
image

**After**
image


johncowen (approved):
👍

PR #4599, opened 2026-02-11
fix(x-action): allow fallthrough of click-handlers

schogges:
As of https://vuejs.org/guide/components/attrs.html#nested-component-inheritance `defineEmits` prevents fallthrough of defined event-handlers. For the case of `XAction` this meant that any `@click` (`v-on:click`) handlers would be consumed by `XAction` and removed from the `$attrs` object, so we could not check for the presence of `$attrs.onClick` in the template to determine which type of action it should render.
Since we are already binding all `$attrs` in the different cases, I think we can safely remove `defineEmits` and just let `onClick`-handlers fall through.

### DIFF
--- a/packages/x/src/components/x-action/XAction.vue
+++ b/packages/x/src/components/x-action/XAction.vue
@@ -15,7 +15,6 @@
         },
       }"
       :danger="props.appearance === 'danger' ? true : false"
-      @click="emit('click')"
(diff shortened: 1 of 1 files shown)


johncowen (approved):
Thanks for figuring this out! 🙌

PR #4600, opened 2026-02-11
fix(workflows/release): not using bootstrap

schogges:
The release workflow didn't use `actions/bootstrap` yet and therefore didn't use the correct `npm` version.

I've also checked if there are more workflows that didn't use `actions/bootstrap` yet, but seems like they are all up to date.


johncowen (approved):
nice! thanks for checking the others 👍

PR #4604, opened 2026-02-12
refactor(data-loader): use state machine

schogges:
Some more refactorings in `DataLoader` mainly to turn it into a state machine with the states `error`, `connecting` and `default`. The states are inferred from the given props `:src` and `:data` and reflect their state. This means we could also remove `loader` from the props because the state `connecting` only appears if there is a `src` and/or `data` being provided as props. All in all the changes in here are non-breaking.


johncowen (commented):
Left some nits, mainly just a few comment additions and moving some code blocks around. I think the runtime is a lot easier to grok now and I really like this approach!

Lmk!

packages/data/src/vue/components/data-loader/DataLoader.vue
```diff
@@ -147,4 +106,47 @@ provide('data-loader', {
...
+// The state will only ever be connecting if either props.src or props.data is set.
+// In any other case the state will only be error or default.
+const state = computed<'error' | 'connecting' | 'default'>(() => {
+  switch(true) {
+    case allErrors.value.length > 0:
+      return 'error' as const
+    case allData.value.some(item => typeof item === 'undefined'):
+      return 'connecting' as const
+    default:
+      return 'default' as const
```
johncowen: If I'm not mistaken none of the `'error' | 'connecting' | 'default'` or `as const`s are necessary here. I think _if anything_ we might want to tie this to whatever the keys of `defineSlot`s are, but even then they can't really be "tied".

🤔 All in all this comment is all buildtime/TS tho, and I think we are done with runtime so I'm good with the PR as is I think. 1 sec lemme approve this. FYI I'd like to revisit the types in this PR at some point I think they can be simplified slightly.

Although do let me know if you'd like to do that here, but I guess the main outcome right now was to get rid of the deprecated `loading` prop.

PR #4619, opened 2026-02-16
chore(docs): additional pass at `@kumahq/data` docs

johncowen:
Wanted to bring these up-to-date a little.

~We are currently working through some stuff here so putting into Draft for now ready to finish up once we've finished up the code.~

Taking out of draft, we decided to follow up later following the stuff we are working through


schogges (commented):
This is still a draft, but just a little suggestion/fix while I was here 🙂

PR #4623, opened 2026-02-17
chore(css): moves all our sass usage to CSS props via XTheme

johncowen:
We want to be in a position where we can easily theme our application (e.g. light/dark).

This PR is a step towards that by moving the majority of our build-time/static SASS vars to use runtime/dynamic CSS props instead.

**As CSS changes can be hard to spot via automated testing etc, please give the preview site a good click around on, to my eyes it looks fairly identical to before this PR.**

---

This PR is also a part experiment/bounce-around of AI.

The PR has 2 commits, one entirely generated by Claude/Opus4.6 and one with some tiny amends by me. I've included the prompts and plan that Claude created based off that and out of interest I asked it to record the time/tokens used in the frontmatter. Unfortunately this is only the time/tokens to generate the plan, I ran out of my session allowance just as Claude finished up its first pass (the first commit), and I was asking it to record the time/tokens for the implem. also. I guess I could have recorded this myself but I quit Claude and forgot to take a note of the time/tokens 🙈 .

As of right now I plan on deleting the plan before merge, I've left it in here seeing as this is the "modern source code" and we are an Open Source project. But as we haven't defined what we want to do here (if anything) I guess I'd probably delete it once we've reviewed this experiment.

Lastly, this PR still requires some review as I've already spotted one place where I think the colors have altered slightly, but all in all I think it looks pretty promising!

Oh I should add, Claude was running in a docker based sandbox with the project mounted inside along with an iptables based firewall. Be careful out there folks! 😅 

---

Edit: I added another commit with a mostly AI generated README for `XTheme` (I asked it to take a look at the plan again), I made some fairly extensive edits to that myself and added a short README for `x-icon-start` also. 12% of my current session usage for writing the README 😱 .

Lastly I'm planning on deleting the plan from here, in fact I'll do it now, I'm guessing it will still be visible in the commits if folks are interested. edit: plan deleted taking out of Draft

### DIFF
--- a/packages/x/src/components/x-theme/XTheme.vue
+++ b/packages/x/src/components/x-theme/XTheme.vue
@@ -206,12 +206,14 @@
   /* color-background */
   --x-color-background: #{$kui-color-background};
   --x-color-background-inverse: #{$kui-color-background-inverse};
+  --x-color-background-neutral-weak: #{$kui-color-background-neutral-weak};
   --x-color-background-neutral-weaker: #{$kui-color-background-neutral-weaker};
   --x-color-background-neutral-weakest: #{$kui-color-background-neutral-weakest};
   --x-color-background-primary: #{$kui-color-background-primary};
   --x-color-background-primary-weakest: #{$kui-color-background-primary-weakest};
   --x-color-background-success-weakest: #{$kui-color-background-success-weakest};
   --x-color-background-warning-weak: #{$kui-color-background-warning-weak};
+  --x-color-background-danger: #{$kui-color-background-danger};
 
   /* color-border */
   --x-color-border: #{$kui-color-border};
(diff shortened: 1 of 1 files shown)


johncowen:
Oh top spot 👏  , let me sort that here

packages/x/src/components/x-theme/XTheme.vue
schogges: I think we should use `--x-space-30` here

johncowen:
K I think I got those last few, I found some comments in the index.html file also so I changed them for completeness aswell.

Apart from that I've:

1. Reorder XTheme definitions so the tokens are at the top, followed by the icons, followed by some initial application application of the tokens.
2. I remembered kongponents had made our buttons sometimes have underlines, so I fixed that here also seeing as I was here. Happy to move this to another PR but I'm fine having it here.

PR #4628, opened 2026-02-18
refactor(DataLoader): make data slot-prop always an array of all data provided as props

schogges:
This is the last step of the refactorings in `DataLoader` and changes the return type of the `data` default-slot-prop to a consistent array. This means that we always return an array of data regardless of if we've set a `:src`-prop, `:data`-prop or both. This is a breaking change and required to update the usage in all the places where we use `v-slot` of the `DataLoader`.

### DIFF
--- a/packages/data/src/vue/components/README.md
+++ b/packages/data/src/vue/components/README.md
@@ -229,7 +229,7 @@ polling.
 In most cases refetching happens automatically in the background whenever a path or query parameter changes. But in some cases there is an imperative way of refreshing the fetched data required, i.e. to allow the user update the data manually, say for example a `[Refresh]` button. For this use case the `DataSource` component exposes a `refresh` method. The same method is also passed through `DataLoader`.
 Refreshing data by refetching an endpoint will not show a loader as it happens in the background. Once the new data is fetched the view will be rehydrated with the new data.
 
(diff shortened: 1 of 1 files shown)


packages/data/src/vue/components/README.md
```diff
@@ -220,4 +222,39 @@ polling.
...
   }
 },
+```
+
+### Refetching and refreshing
 
+In most cases refetching happens automatically in the background whenever a path or query parameter changes. But in some cases there is an imperative way of refreshing the fetched data required, i.e. to allow the user update the data manually. For this use case the `DataSource` component exposes a `refresh` method. The same method is also passed through `DataLoader`.
+Refreshing data by refetching an endpoint will not show a loader as it happens in the background. Once the new data is fetched the view will be rehydrated with the new data.
+
+**Note:** Calling `refresh` on `DataLoader` only refreshes a given source, but not any data that is passed via the `data`-prop.
```
johncowen:
```suggestion
**Note:** Calling `refresh` on `DataLoader` only refreshes a given source i.e. the data that the `:src` prop refers to, but not any data that is passed via the `:data` prop.
```

Another extremely tiny detail to try and make this super clear, bit of a nit also really.

johncowen (commented):
Just two more tiny suggestions in the docs, then I think this is good to go!

PR #4629, opened 2026-02-18
chore(resources): refactor search parsing/formatting into two steps

johncowen:
We need to use this search functionality elsewhere, but instead of returning the data formatted specifically for a queryString, we need the data as raw json:

```js
{
  labels: {
    'kuma.io/display-name': 'name',
    ...etc
  }
}
```

I began by re-familiarizing myself with the code by refactoring it slightly:

- Setting the key/value to have defaults thus avoiding `?`s later on.
- Splitting up the cases a little and adding some explanatory comments as I re-remembered how things worked.
- Generally, especially if we omit `tags`, we special case `name` (or the `defaultKey`) and convert everything to `labels`, so I moved this one to be the `default` case.

The tests were _extremely_ useful in doing this and gave me pretty good confidence that there is no functionality change here (?)

Once I'd done this I pulled out the `filters[labels.]` stuff and instead returned a json object looking like `{ labels: {} }` (see example above)

I then reformatted this JSON so it was the same as it was before.

Lastly I then split this into two methods `parseSearch` and `search`. I would guess these names and exactly how the sources.ts files use these _may_ change slightly in the future. But for now I wanted to avoid changing the tests for ease of review/confidence and much more churn that isn't super necessary right now.

I plan to use `parseSearch` on its own elsewhere pretty soon, but at some point I might convert the tests to test `parseSearch` instead, and move/change how we then format the result of `parseSearch` but I haven't thought this through enough yet.

Umm I'd maybe advise looking at the entire files separately to look at changes rather than using the changeset, but maybe its just me that finds that easier.

### DIFF
--- a/packages/kuma-gui/src/app/resources/data/Resource.ts
+++ b/packages/kuma-gui/src/app/resources/data/Resource.ts
@@ -52,7 +52,7 @@ export const Resource = {
           }
         // support old style tags, explicitly used `tag:key:value`
         // TODO: this will need to be optional
-        case ['tag', 'tags'].includes(key): {
+        case ['tag'].includes(key): {
           const [k, v = ''] = value.split(kvSeparatorRegex)
           return {
             ...acc,
@@ -66,7 +66,7 @@ export const Resource = {
         // it into a label
         default: {
           // explicitly used `label:key:value` or, if not, it was just `key:value`
-          const [k, v = ''] = ['label', 'labels'].includes(key) ? value.split(kvSeparatorRegex) : [key, value]
+          const [k, v = ''] = ['label'].includes(key) ? value.split(kvSeparatorRegex) : [key, value]
           return {
             ...acc,
             labels: {


packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -14,57 +14,76 @@ type SearchOptions = {
...
         }
-        default:
-          return acc
       }
-    }, {} as Record<string, string | string[]>)
+    }, {} as Record<string, string | Record<string, string>>)
+
+
+  },
+  search(query: string, options: SearchOptions = {}) {
```
schogges: The return type of `search` is not reflecting the actual return value properly anymore. The return type is now just `{ tag?: string[] | undefined }` whereas the actual return type is rather `Record<string, string | string[]>`. We can be also more explicit by explicitely adding `{ tag?: string[] | undefined }`

johncowen: I think this is sorted now, lemme know!

packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -14,57 +14,76 @@ type SearchOptions = {
...
+    const { labels = {}, tags = {}, ...rest } = Resource.parseSearch(query, options)
+
+    return {
+      ...rest,
+      ...(Object.keys(labels).length > 0 ? Object.fromEntries(Object.entries(labels).map(([key, value]) => [`filter[labels.${key}]`, value])) : {}),
```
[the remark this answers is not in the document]
schogges: No strong opinion either, would just save an unnecessary ternary, but you have a reason to keep it so I'm 🤷‍♂️

PR #4645, opened 2026-02-23
fix(dataplanes): show dataplane info even when /stats fails

johncowen:
I found a tiny loading/error bug whilst working on something else.

TLDR: There was a DataLoader waiting on `traffic` but not erroring on `traffic`s error. When I looked at it I don't think we should be using a DataLoader (neither a loader nor an error) and just wait for the data and show the rendered ui once its loaded. Errors are dealt with via error notifications elsewhere.

I added a little test to prevent future regressions (and tested it breaks on master)

Previous to this PR, the `About Dataplane` box would show an eternal loader if this errored (relatively common) even though it could show all of the data about the Dataplane.


schogges (approved):

johncowen:
Sorry for the delay in rebasing, this one is ready to go again @schogges !

PR #4667, opened 2026-03-02
chore(docs): add some basic README/docs to @kumahq/container

johncowen:
TL;DR Adds some initial non-AI assisted / "brain-dump" docs around `@kumahq/container` - I plan on improving these at some point in the short/mid-term.

[Rendered README](https://github.com/johncowen/kuma-gui/blob/b7b20a6057f88d84854cdc84e9f1b0e6de9427d7/packages/container/README.md)

We don't change the interfaces here often, just because there is no real need to. But, the last (and only) time we did change the interfaces meant that the docs we had at that time became out date and irrelevant. We've also moved our repository structure from a normal repo to a workspace'd monorepo and dropped our vitepress documentation site (due to effort of support). The combination of these things has meant some of this documentation has not been updated and eventually deleted to prevent confusion.

This time around I've tried to provide very high level documentation of the approach, not necessarily the exact implementation. This should ensure that at least this part stays relevant no matter what changes happen in the future.

---

Some of the text here has been taken from elsewhere, but I mainly wanted to outline that our service container is literally just turning something like this:

```yaml
data-source:
  service: DataSourcePool
  arguments:
    - sources
    - lifecycle
lifecycle:
  service: DataSourceLifecycle
mesh-sources:
  service: getMeshSources
  labels:
   - sources
data-source-with-logging:
  service: loggingDecorator
  decorates: data-source
```

Into this:

```js
const meshSources = getMeshSources()
const lifecycle = new DataSourceLifeCycle()
const dataSourcePool = new DataSourcePool(
  [meshSources],
  lifecycle
)
const dataSource = loggingDecorator(dataSourcePool)
```

and then giving a few reasons as to why we need this (which have also been documented elsewhere)

---

I plan on revising this more in the near/short term but for now this should act as a good starting point.

### DIFF
--- a/packages/container/README.md
+++ b/packages/container/README.md
@@ -1,0 +1,210 @@
+# @kumahq/container
+
+This module contains code to implement an unobtrusive service container to
+achieve a dependency inversion pattern
+
+This README begins with a high level overview of the approach given that whilst
+we have a model of the problems we are solving and where we want to be, the
+implementation is still in flux and likely will be for some time to come.
+
+Part of the design of the approach (keeping the
+configuration/injection/implementation on The Outside) was decided knowing that
+things would likely change over time and we didn't want to lock ourselves in to
+a particular implementation or dependency. Therefore a minimal declarative
+"configuration-based" API was chosen.
+
+## TL;DR
+
+At a high-level the approach we are aiming for would look like the following
+YAML:
+
+```yaml
+data-source:
+  service: DataSourcePool
+  arguments:
+    - sources
+    - lifecycle
+lifecycle:
+  service: DataSourceLifecycle
+mesh-sources:
+  service: getMeshSources
+  labels:
+   - sources
+data-source-with-logging:
+  service: loggingDecorator
+  decorates: data-source
+```
+
+This would convert into the following "wiring/plumbing" in javascript:
+
+```js
+const meshSources = getMeshSources()
+const lifecycle = new DataSourceLifeCycle()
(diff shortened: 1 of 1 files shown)


johncowen:
> Is there a plan to rewrite this?

No real plan at the moment. I have "pretty much" rewritten this (without the brandi dependency). As to whether it will ever land in here is a different question.

> Will we have v1, v2, ...?

We may (or may not) end up with a `v1`, in which case there will be a period of time where we will have `v0` and `v1` in the package. If this ever happens (I'm not certain it will), once everything is using `v1`, `v0` will be deleted.

> Why do you think we should mark it as v0 internally?

It gives us a clean space / directory should we wish to start an introduction / refactor of this. I wouldn't want both old implementation and new implementation mixed in the same `index.ts` file or root directory.

`v0` also says to me "this was the first version, and is just the shape of how it ended up, ideally we would improve on it given time". If it ends up being the only version here I would guess people would think "oh look its a `v0`" and derive meaning from that.

---

I think all in all what we have is totally fine from a functional perspective - it does exactly what we need. There are some DX things I'd like to improve, but in the scheme of things it's way down on the priority list.

But I am conscious of the fact that some of the documentation for this became obsolete when we changed some things a few years back. So I wanted to write something down that wouldn't become obsolete plus and then include separate documentation on how it is now whilst making it clear that this second part could also become obsolete. The `v0` thing helped with that.

Does that help at all? Let me know if anything else isn't clear or you had other suggestions 👍

PR #4699, opened 2026-03-10
feat(kuma-http-api/mocks/data-planes/inbounds/policies): add kri to rules

### LINKED ISSUE #4678: udpdate mocks for `conf.kri`
### Kuma Version

2.13.x

### Describe the bug

During https://github.com/kumahq/kuma-gui/pull/4578 we updated the policy origins to display differently. We could only do this because the `conf` for each policy now includes the `kri` (`conf.kri`)

I noticed that the mocks for the endpoint we use for this information does not ever have `conf.kri` set and thus during development using mocks we see `Policy Origin: <blank>`

We should update the mocks so that it does.

I think this is the mock https://github.com/kumahq/kuma-gui/blob/41874f3a787a8efd27fede9d6feb824be4d51afd/packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_inbounds/_/_policies.ts#L19

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

schogges:
Adds a `kri` to each entry of the `rules` of `/meshes/:mesh/dataplanes/:dataplane/_inbounds/:kri/_policies` endpoint.

While I was here I also changed the current policies to Mesh* policies and tweaked the mocks a little for more reasonable examples. I've used AI to generate those and prompted to look at the respective policy docs at https://kuma.io.

Closes https://github.com/kumahq/kuma-gui/issues/4678


johncowen (commented):
Couple of qs/suggestoins

johncowen (dismissed):
👍

johncowen (approved):

PR #4708, opened 2026-03-13
fix(kuma-gui/app-collection): coloring of badge links

schogges:
In some cases we have badges in a collection/listing that link to other places. In https://github.com/kumahq/kuma-gui/pull/4608 we ensured that especially buttons infer the color. Using `> *` was too eager and also overruled the coloring of badges. Therefore I've changed it to only `> span`. There would be other ways to fix this i.e. excluding badges via `:not(.k-badge)` from the parent selector or `> *:not(div)`, but I thought what I went for now is a little broader and still more defensive. Still I think this is not the perfect solution, but fixes it for now. We might need to circle back at a later point.

The reordering in the parent selector doesn't actually change anything, but to me it's easier to grok to first have a selector and the specify it further, i.e. in this case "any data-action that is an anchor or button" is easier to me vs "any anchor or button that is data-action". That's totally optional and opinionated and I can revert if not desired.

**Before**
Screenshot 2026-03-13 at 16 52 56

**After**
image


johncowen (approved):
Oh I think I meant to approve this when we spoke, LGTM!

PR #4719, opened 2026-03-16
refactor(x-search): reduce custom layout efforts

schogges:
When introducing `XSearch` back in https://github.com/kumahq/kuma-gui/pull/3737 and the dropdown in https://github.com/kumahq/kuma-gui/pull/3748 there was a need to set the internal width of of the top most internal element to 0. This was due to two exceptional cases we have from a layout perspective:
1. dataplanes listing filters also include an inline select
2. policy listing is nested in a `separated` layout without a parent element giving a certain width that `width: 100%` can refer to

This caused that we had to add a `width: 100%` to all the other cases where `width: 100%` would just work. This PR now turns it around to leave and improve the CSS in the special cases.

---

While I went through the application I noticed that I could remove some more CSS that was not doing anything and most likely was left after some copy&paste and replaced it with `XLayout`.

---

Best to review without whitespaces.

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
@@ -346,10 +346,6 @@ search {
 
 .search-field {
   flex-basis: 310px;
-  flex-grow: 1;
-}
-
-.search-field {
   flex: 1;
 }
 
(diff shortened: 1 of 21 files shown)


johncowen (dismissed):
🎉 Nice cleanup! 👏

johncowen (approved):

PR #4723, opened 2026-03-17
refactor(kuma-gui/control-planes): adopt data-loader changes

schogges:
Removes the need for checking wether a `result` and other props sourced from data-source are `undefined` or of type `Error` by adopting the recent changes to `DataLoader`.

In this case here I've also removed the `computed` property and inlined the calculation into the template. In order to save another wrapping template that remodels the `policyTypes` I've moved this part to the data-layer.


johncowen (approved):
Thanks!

PR #4724, opened 2026-03-17
feat(ci): add initial playwright support

johncowen:
Adds initial playwright browser runner support for our gherkin e2e tests

This is not yet wired into CI, but I've added a `make playwright` which you can use in kuma-gui to test with (and also play with extra flags like `--headed` and `--ui`).

I would like to merge it in this state, and then get it wired into CI as a follow up.

---

This PR is kind of a "middle stage" with an eye to replacing Cypress in CI as soon as possible. There are some semi-obvious TODOs in here that will be done at a later stage. I'll make some inline notes here and we can also chat in sync. Its mostly just a reimplementation of our steps using Playwright APIs.

### DIFF
--- a/packages/fake-api/package.json
+++ b/packages/fake-api/package.json
@@ -18,10 +18,6 @@
       "import": "./src/cypress.ts",
       "types": "./src/cypress.ts"
     },
-    "./playwright": {
-      "import": "./src/playwright.ts",
-      "types": "./src/playwright.ts"
-    },
     "./msw": {
       "import": "./src/msw.ts",
       "types": "./src/msw.ts"


packages/kuma-gui/playwright/main.ts
```diff
@@ -0,0 +1,45 @@
...
+          $.vars,
+        ],
+      }],
+    ],
+  )
+  setupSteps({
+    baseURL: 'http://localhost:8080/gui',
+    dependencies: get($.dependencies),
+    fs: get($.fakeFS),
+  })
```
johncowen: I'm eventually just going to pass `baseURL` and `fetch` into here. Mocking can then default to a native-like fetch if you are not using `fake-api` (theoretically)

packages/kuma-http-api/mocks/FakeKuma.ts
```diff
@@ -115,7 +115,7 @@ sWWnRiRqtJ6pNYTI3+qy4Y/pBcsX9rAFAlMGaNZiyo77Hd8C2gGTraj1T2qwKhAf
 jcrw4Z/qOwintpvJJVIrNKELr/zfQv5ugop05YC2h+rF1eSdwOGyzltnyunwi5kV
 gbXR5RnEs0hDxugaIknJMKk1b0g=
 -----END CERTIFICATE-----`
-    const encode = typeof window?.btoa === 'function' ? window.btoa : (str: string) => Buffer.from(str).toString('base64')
+    const encode = typeof globalThis?.btoa === 'function' ? globalThis.btoa : (str: string) => Buffer.from(str).toString('base64')
     return base64 ? encode(cert) : cert
   }
```
johncowen: playwright doesn't like any references to `window`, makes sense really.

johncowen:
We figured we had to set a TZ to UTC like we do for Cypress here:

https://github.com/kumahq/kuma-gui/blob/080fabd4be1017ecf6a7532b0c20e876e3d2070c/packages/config/src/mk/test.mk#L29

Lemme add in the config and rebase this so we are ready to go.

PR #4760, opened 2026-03-25
feat: add resources

### LINKED ISSUE #4777: Create resources tab
### Description

This is the main part of https://github.com/kumahq/kuma-gui/issues/4683.
It will be put behind a feature flag until the rest of it is done and we can fully replace the policies tab. For better tracking of the progress I've extracted the creation of the resources tab into this separate child issue.

schogges:
Create a new `Resources` tab that will replace the `Policies` tab. It's being put behind a feature flag `KUMA_RESOURCES_ROUTE_ENABLED`.

This mainly affects the current `resources` module:

- Adds new views
  - `ResourceTypesListView` which shows all different resource types (`/_resources`) in the different scopes.
  - `ResourceListView` is a generic list view that renders a list of resources of the selected resource type
  - `ResourceSummaryView` is a generic summary view in a drawer that is rendered relative to the `ResourceListView`
  - `ResourceDetailView` is a generic detail view for a resource that has no respective detail view and currently for the views/routes that do not support KRIs (this will be handled in follow-ups)
- Adds respective APIs in the sources and handlers in the data-layer to handle generic resources
- ~Moves out parts of `MeshIdentity` and `MeshTrust` that would cause conflicts (this may not be the case anymore with the extra namespacing that I've added to the generic resources sources)~ This will be done separately

---

~In order to link to generic resources by a KRI I've extended the `protocolHandler` to handle routing relatively for summaries and separately for navigating to detail views.~ Extending the `protocolHandler` will be done in a follow-up.

Closes https://github.com/kumahq/kuma-gui/issues/4777


packages/kuma-gui/src/app/resources/data/ResourceTypes.ts
```diff
@@ -0,0 +1,73 @@
...
+  ['meshtrafficpermission', ['security']],
+  ['trafficpermission', ['security']],
+  ['meshtrust', ['security']],
+  ['proxytemplate', ['performance', 'traffic']],
+  ['secret', ['security']],
+  ['trafficroute', ['traffic']],
+  ['virtualoutbound', ['traffic']],
+  ['workload', ['traffic']],
+  ['zone', ['traffic']],
+])
```
schogges: These will be reviewed separately and potentially be added to the entries of `/_resources`

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
```diff
@@ -133,7 +133,7 @@
 </template>
 <script lang="ts" setup>
 import type { Mesh } from '../data'
-import type { MeshIdentity } from '@/app/resources/data'
+import type { MeshIdentity } from '@/app/resources/data/MeshIdentity'
```
johncowen: Feels like this will need to be changed again once https://github.com/kumahq/kuma-gui/pull/4776 goes in? Either way I think I'd prefer `@app/module-name/data` to export all their "data types" and then we import them from there.

packages/kuma-gui/src/app/meshes/features.ts
```diff
@@ -0,0 +1,16 @@
+import type { Env } from '@/app/application'
+import type { Features } from '@kumahq/settings/can'
+
+export const features = (env: Env) => {
+  return {
+    'use resources route': (_can: unknown) => {
```
johncowen, line 6: Ideally this would just be `use resources` (i.e. the new "resource module"), erm I don't know how long we are going to have this feature flag so not a biggie.

packages/kuma-gui/src/app/resources/views/ResourceDetailView.vue
```diff
@@ -0,0 +1,207 @@
...
+                :text="data.name"
+              >
+                <RouteTitle
+                  :title="t('policies.routes.item.title', { name: data.name })"
+                />
+              </XCopyButton>
+            </h1>
+          </DataLoader>
+        </template>
+        <XAboutCard
```
johncowen: This will need tweaking now! Sorry!

packages/kuma-gui/src/app/resources/views/ResourceDetailView.vue
```diff
@@ -0,0 +1,207 @@
...
+              v-for="options in [['universal', 'k8s']]"
+              :key="typeof options"
+            >
+              <XSelect
+                :label="t('data-planes.routes.item.format')"
+                :selected="route.params.format"
+                @change="(value) => {
+                  route.update({ format: value })
+                }"
+                @vue:before-mount="$event?.props?.selected && options.includes($event.props.selected) && $event.props.selected !== route.params.format && route.update({ format: $event.props.selected })"
```
johncowen, line 152: I'd still really like to make this more declarative via our RouteView params. I've been wanting to do this for ages just never gotten around to it. No action here btw just talking out loud.

packages/kuma-gui/src/app/resources/views/ResourceTypesListView.vue
```diff
@@ -0,0 +1,184 @@
...
+    v-slot="{ uri, route, t }"
+  >
+    <RouteTitle
+      :render="false"
+      :title="t('resources.routes.items.title')"
+    />
+    <AppView>
+      <DataSource
+        :src="uri(sources, '/resources', {}, {
+          search: `${route.params.category !== 'all' ? `category:${route.params.category}` : ''}`,
```
johncowen: Note to self: Should we be using `search` here or `/resources/:category` 👀

johncowen (commented):
Oh jees 🤦 I have a partial review here from yesterday that wasn't submitted yet, let me submit that now, not sure if things are still relevant since a bunch of things have gone in since yesterday, I'll resolve anything that isn't relevant

johncowen:
ok guessing you might be already working on the rebasing here, ping me once you are ready for another review, I think a lot of my comments will be resolved once rebased 👍

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
schogges: Yep, will cover this when rebasing 👍

packages/kuma-gui/src/app/resources/views/ResourceDetailView.vue
schogges: Yes, will cover this when rebasing 🙂

packages/kuma-gui/src/app/meshes/features.ts
schogges: Hopefully we won't have it long. Resources itself is not a new thing and should not be behind a feature flag, just the routing.

packages/kuma-gui/src/app/resources/views/ResourceTypesListView.vue
schogges: The intention was to be more like usual filtering, but having this in the path is a lot cleaner in the template 👍

schogges:
@johncowen I've rebased it, resolved the conflicts and addressed your comment. This is ready to review now 🙂

packages/kuma-gui/src/app/control-planes/data/index.ts
```diff
@@ -15,25 +15,46 @@ export const GlobalInsight = {
...
+          },
+          ZoneIngress: {
+            ...partialGlobalInsight.resources.ZoneIngress,
+            total: partialGlobalInsight.resources.ZoneIngress?.total ?? partialGlobalInsight.zones.zoneIngresses.total ?? 0,
+          },
+          Mesh: {
+            ...partialGlobalInsight.resources.Mesh,
+            total: partialGlobalInsight.resources.Mesh?.total ?? partialGlobalInsight.meshes.total ?? 0,
+          },
+        }
```
johncowen, line 33: I guess not a blocker for this, but what happens here when a new resource is added and we haven't dealt with it by defaulting to zero here? Maybe this is just the types telling us lies and these totals are guaranteed to always be set to zero? I guess we can check with backend and then we can potentially remove all this.

packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
```diff
@@ -19,13 +19,13 @@
...
       <DataSource
-        :src="uri(meshIdentitySources, '/meshes/:mesh/meshidentities', {
+        :src="uri(resourcesSources, '/meshes/:mesh/meshidentities', {
           mesh: route.params.mesh,
         })"
         v-slot="{ data: meshIdentities }"
       >
         <DataSource
-          :src="uri(meshTrustSources, '/meshes/:mesh/meshtrusts', {
+          :src="uri(resourcesSources, '/meshes/:mesh/meshtrusts', {
```
johncowen: Could you explain how come these are using the resources module now? Not super clear by looking here why and depending on the outcome/reason I may disagree

packages/kuma-gui/src/app/resources/views/ResourceListView.vue
```diff
@@ -0,0 +1,266 @@
...
+              </form>
+            </search>
+            <DataLoader
+              :src="type.scope === 'global' ? uri(sources, '/resources/:path', {
+                path: route.params.resourcePath,
+              }, {
+                page: route.params.page,
+                size: route.params.size,
+                search: route.params.s,
+              }) : uri(sources, '/resources/:path/for/:mesh', {
```
johncowen, line 83: nice, I really really prefer this pattern of keeping the "module prefix" and using `/for/:mesh` at the end. I don't think we need to worry, but if we ever need different `fors` we'd have to do `/for/mesh/:mesh` but I don't think we need to do that now (if at all)

packages/kuma-gui/src/app/resources/routes.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const meshTrustRoutes = (prefix: string): RouteRecordRaw[] => {
+  return [
+    {
+      path: 'meshtrust/:mtrust',
+      name: `${prefix}-mesh-trust-summary-view`,
+      component: () => import('@/app/mesh-trusts/views/MeshTrustSummaryView.vue'),
+    },
+  ]
+}
```
johncowen: How come these routes are still here (the mesh trust and mesh identity ones). Didn't we make a new module for them? Maybe this is left over form the rebase?

packages/kuma-gui/src/app/resources/sources.ts
```diff
@@ -1,17 +1,201 @@
 import createClient from 'openapi-fetch'
 
+import { Resource } from './data'
+import { MeshIdentity } from './data/MeshIdentity'
+import type { KumaMeshIdentity } from './data/MeshIdentity'
+import { MeshTrust } from './data/MeshTrust'
+import type { KumaMeshTrust } from './data/MeshTrust'
+import { Resources } from './data/Resources'
+import { ResourcesTypes } from './data/ResourceTypes'
```
johncowen: Could we change anything that comes from the `./data` folders use the barrel file i.e.:

```ts
import { Mesh, MeshInsight } from './data'
```

packages/kuma-gui/src/app/resources/sources.ts
```diff
@@ -1,17 +1,201 @@
...
   })
   return defineSources({
     '/resources': async () => {
-      return (await http.GET('/_resources')).data!
+      const response = await http.GET('/_resources')
+
+      return ResourcesTypes.fromCollection(response.data!)
+    },
+
+    '/resources/by/:category': async (params) => {
```
johncowen: Little bit of a nit, but if we want to be consistent (probably good idea) we should use the word `/of/:category` here. We use this in dataplanes sources.

packages/kuma-gui/src/app/resources/sources.ts
```diff
@@ -1,17 +1,201 @@
...
+        return normalized
+      }
+
+      return {
+        ...normalized,
+        resources: normalized.resources.filter((resource) => {
+          return resource.categories.includes(category)
+        }),
+      }
+    },
```
johncowen, line 27: I just realized this is frontend filtering, not backend filtering like we do in dataplanes - wondering if its ok that we perform multiple HTTP requests for this, even though we could totally do this on the frontend 🤔

packages/kuma-gui/src/app/resources/sources.ts
```diff
@@ -1,17 +1,201 @@
...
+    '/resources/:path/for/:mesh': async (params) => {
+      const { mesh, path } = params
+      const search = Resource.search(params.search)
+      
+      const response = await http.GET(`/meshes/{mesh}/${path as DynamicPathMesh}`, {
+        params: {
+          path: {
+            mesh,
+          },
+          // @ts-expect-error - incorrect typing in openapi-fetch
```
johncowen: Wondering how come we didn't need these in `policies`?

packages/kuma-gui/src/app/resources/sources.ts
```diff
@@ -1,17 +1,201 @@
...
+            kri: mtrust,
+          },
+          // @ts-ignore
+          query: {
+            format: 'kubernetes',
+          },
+        },
+      })
+      
+      return res.data
```
johncowen: Yeah I'm definitely wondering why all these mesh identity and mesh trust sources are in here?

packages/kuma-gui/src/app/service-mesh/index.ts
```diff
@@ -136,6 +135,5 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
     ...zones(app),
     ...meshes(app),
     ...hostnameGenerators(app),
-    ...resources(app),
```
johncowen, line 139: Wondering how come we deleted this?

johncowen (commented):
I had a bunch more comments/questions. I'm also going to try and take a closer look at this locally in the meantime.

packages/kuma-gui/src/app/control-planes/data/index.ts
schogges: The important part was having ` as PartialGlobalInsight['resources'] & typeof resources` because otherwise it caused type issues.
Needed to preserve the original type from OAS otherwise it would be turned to only having `MeshService`, `MeshMultiZoneService` and `MeshExternalService` which would cause issues when trying to access any `string` from `resources`:

```
resources: {
    [key: string]: {
        total: number;
    };
}
```

packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
schogges: Uff 😮‍💨 didn't check well enough that this all was cleaned up when rebasing and resolving the conflicts

packages/kuma-gui/src/app/resources/routes.ts
schogges: Leftover from rebasing 🙈

packages/kuma-gui/src/app/control-planes/data/index.ts
johncowen: Gotcha, ok I'll take a closer typescript look at this when I pull it down 👍

packages/kuma-gui/src/app/resources/sources.ts
schogges: Leftover from rebasing 🙈

packages/kuma-gui/src/app/resources/sources.ts
schogges: Ah thanks, this wasn't correct. Didn't pass through the pagination params.

packages/kuma-gui/src/app/service-mesh/index.ts
schogges: I added this for mesh-identities and mesh-trusts, but it's not needed here.

packages/kuma-gui/src/app/resources/sources.ts
schogges: Maybe we get backend filtering for this at some point, that's why I wanted it to be as close as possible to this.

packages/kuma-gui/src/app/service-mesh/index.ts
johncowen: How are the new `resources` sources.ts being "installed"?

packages/kuma-gui/src/app/control-planes/data/index.ts
schogges: Forgot to mention that I reverted everything else, just kept the type casting 🙂

packages/kuma-gui/src/app/service-mesh/index.ts
schogges: In meshes

johncowen:
Some comments on the listing/summary view that don't need to happen in this PR (and probably shouldn't be so we can get this in and move forwards with a new PR)

- Category menu should look like the `[title / menu]` layout we have elsewhere (like summaries) i.e. the title and category menu should line up (I 'think' our new `action-group` XLayout variant should help here)

Screenshot 2026-04-16 at 11 30 10

- We shouldn't 50%/50% this layout, it should be the same as policies (which I believe, but not sure, has some sort of special sizing to fit the minimum menu item width). I noticed that the vertical spacing/scrolling is also much better in the policies view (We could consider a `<XLayout variant="vertical-nav">` or similar if this is something we want to reuse, but lets consider/chat first)

Screenshot 2026-04-16 at 11 33 09

- Clicking ResourceListItems in the list/table defaults to structured view but there is no structured view in the summary. We should either make a structured view or default to either universal or kubernetes

- Policies has a selected state for the listing items, we should make this the same

Screenshot 2026-04-16 at 11 38 21

- If I select `Category: Security` and then click any of the policy types in the left column, the `?category=security` is removed from the URL, meaning when I refresh I no longer have that category selected.

- I think that we briefly spoken about this before, but if I select a new filter for my Resource Types, I don't think it should jump the Resource listing itself back to a different resource type. I chose to filter the resource types on the left, not change the listing on the right. I think you had a different opinion when we last spoke, so lets discuss this again.

- I don't think we should use `$` for our "frontend KRIs" as a "magic string" because it can get encoded to `%24`, lets use `~` instead which I don't think gets encoded ever.

I've not looked massively at the detail view yet, and to note again, I don't think any of the above are blocking for this PR but IMO they would be blocking to remove the feature flag. i.e. they are for near-future PRs.

packages/kuma-gui/src/app/service-mesh/index.ts
johncowen: Ok yeah, can we keep this in `service-mesh` for now? There are things in the `resources` module that are necessary for "everything" (i.e. I know we wouldn't but there are things in `resources` that we would use even if we removed the `meshes` or `control-planes` "plugins")

packages/kuma-gui/src/app/resources/sources.ts
johncowen: I doubt backend filtering for such a simple HTTP request/dataset will ever happen. Let me think about this one before we work on it. Leaving this one unresolved for the moment.

packages/kuma-gui/src/app/resources/data/index.ts
```diff
@@ -1 +1,3 @@
 export * from './Resource'
+export * from './ResourceTypes'
+export * from './Resources'
```
johncowen: Did we call this `Resources` because there was already a `Resource`?

packages/kuma-gui/src/app/resources/data/ResourceTypes.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const ResourceTypes = {
+  fromObject(partialResource: KumaResourceTypesCollection['resources'][number]) {
```
johncowen: Should we make a `KumaResourceType` and alias this same type you have here at the top of the file?

packages/kuma-gui/src/app/resources/routes.ts
```diff
@@ -0,0 +1,47 @@
...
+        component: () => import('@/app/resources/views/ResourceSummaryView.vue'),
+      },
+    ]
+  }
+
+  const item = (): RouteRecordRaw[] => {
+    return [
+      {
+        name: 'resource-detail-view',
+        path: 'resources/:resourcePath/:kri/overview',
```
johncowen, line 18: I was just wondering whether we need to put the `resourcePath` in the URLs. The idea is to move completely to KRIs, and with KRIs we should be using the shortName from the KRI/ID rather than something else.

johncowen: Ah maybe we need this actually for the "route config overwriting" we were discussing, lets at least chat about this to be sure.

packages/kuma-gui/src/app/resources/sources.ts
johncowen: Yeah it looks like we also do an additional filter in the template to filter out `insight` and `secret`.

At the very least we should do any massaging/filtering of data in one place.

johncowen:
I gave this another pass over and had a few more comments (thanks for bearing with me, I'm giving this a much more careful eye than usual maybe!). But we also have a sync soon so I think we should spend the time there on this with an aim to getting it approved and merged during the sync.

johncowen:
We had a sync and I'm just gonna drop in our notes here also real quick:

> - Q: What should we do about doing the filter in the `/of/category` source, should it remain there or be in the template. ~At the very least we should have this entire filter in one place.~
> - Q: the filters currently in the template or for `insights` and `secret`, which at least for `secret` should never be visible in the GUI. 
> - The resource filtering in regards to legacy policies needs to show all legacy policies if any one of them has a count of greater than zero.
> - Consider moving some of the `/resource` uris to `/resource-descriptors`, maybe look at this along with the filtering frontend/backend thing
> - So lets move the current `Resources.search` etc, to just be `search` etc functions, exported from a Resource.ts (Resources.ts will move to be `Resource.ts`)
> - Lets avoid parsing KRIs in `Resources.fromObject` if we can lets check for a KRI if it doesn't have one generate it from the JSON blob
> - Make sure to check for a zone origin before setting a zone name in Resource.fromObject
> - ResourceTypeListView

As for what we do here and what can be done later, I'd say anything that is about making this code work/look the same as the code everywhere else in the application what I'd prefer to be done here. I think for me this is mostly (perhaps totally) just the stuff in the `resources/data` folder. I think that is the thing that I found hard to follow mainly because of the `Resource.fromObject`/`Resources.fromObject` thing.

Please ping me in Slack once you're happy with any changes made here, I'd still like to get this PR as soon as poss

Oh I just noticed I think I had more thoughts on this bit:

> - Q: What should we do about doing the filter in the `/of/category` source, should it remain there or be in the template. **At the very least we should have this entire filter in one place.**
> - Q: the filters currently in the template or for `insights` and `secret`, which at least for `secret` should never be visible in the GUI. 

Because, to me it makes sense that we filter out `secret` (and probably `insights`) at a global level (in the data / API smoothing place) because we never ever need these ones, and then the other category etc filters in individual places where necessary. I know we kinda had differences of opinion here, but we did agree I think that it currently could be better, but all in all so I kinda don't mind too much at the moment.

I've crossed the bit out in the quote above also.

schogges:
@johncowen I think I've covered everything now 🙂 thanks for the patience

packages/kuma-gui/src/app/meshes/locales/en-us/index.yaml
```diff
@@ -54,6 +54,7 @@ meshes:
         data-plane-list-view: Data plane proxies
         policy-list-index-view: Policies
         workload-list-view: Workloads
+        resource-types-list-view: Resources
```
johncowen: We changed the name of the view to remove the `s` here so I think we need to remove the s here also. I'm not totally sure if this is just a nit or it means the tab text doesn't show properly, if its just a nit I don't mind, but had to say in case its a bug

packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
```diff
@@ -102,14 +102,13 @@
...
                         <XAction
-                          v-if="item.spec.origin.kri.length > 0"
+                          v-if="item.spec.origin?.kri?.length > 0"
                           :to="{
                             name: 'mesh-mesh-identity-summary-view',
                             params: {
                               mid: item.spec.origin.kri,
                             },
                           }"
-                          data-action
```
johncowen, line 112: How come we lost this?

packages/kuma-gui/src/app/resources/views/ResourceTypeListView.vue
```diff
@@ -0,0 +1,197 @@
+<template>
+  <RouteView
+    name="resource-types-list-view"
```
johncowen: Another `s` ah this is why the locale isn't a bug and just a nit, but ideally these need to exactly match the name of the file (well, "exactly", from camel to kebab)

packages/kuma-gui/src/app/resources/views/ResourceTypeListView.vue
```diff
@@ -0,0 +1,197 @@
...
+                  </XLayout>
+                </template>
+                <DataLoader
+                  :data="[sourceResources, sourceGlobalInsight, sourceMeshInsight]"
+                  :errors="[resourcesError, globalInsightError, meshInsightError]"
+                  v-slot="{ data: [resources, globalInsight, meshInsight] }"
+                >
+                  <template
+                    v-for="hasLegacyPolicies of [resources.resources.some(resource => !resource.policy?.isTargetRef && (meshInsight.policies[resource.name]?.total ?? 0) > 0)]"
+                    :key="typeof hasLegacyPolicies"
```
johncowen, line 66: 👍

packages/kuma-gui/src/app/resources/routes.ts
```diff
@@ -0,0 +1,47 @@
...
+        component: () => import('@/app/resources/views/ResourceDetailView.vue'),
+      },
+    ]
+  }
+
+  return {
+    items: (): RouteRecordRaw[] => {
+      return [
+        {
+          name: 'resource-types-list-view',
```
johncowen: another `s` removal

packages/kuma-gui/src/app/resources/views/ResourceTypeListView.vue
```diff
@@ -0,0 +1,197 @@
...
+                  >
+                    <XLayout
+                      v-for="filtered in [resources.resources.filter((item) =>
+                        (route.params.category === 'all' || item.categories.includes(route.params.category)) &&
+                        (!item.policy || item.policy?.isTargetRef || hasLegacyPolicies) &&
+                        (item.name !== 'ExternalService' || meshInsight.services.external > 0)),
```
johncowen, line 72: I like this much better 👍

packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
schogges: It's not needed and it's more like how it's in other places, where only the first XAction in a table row has this attribute set.

packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -1,3 +1,11 @@
+import { Kri } from '@/app/kuma'
+import type { paths } from '@kumahq/kuma-http-api'
+
+
+type KumaResourceCollection = Record<string, unknown>
```
johncowen, line 8: I think we need to amend this PR ever so slightly one more time (for the `s`s) if so could we add a comment above this to explain why this is `unknown`

packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -93,4 +102,40 @@ export const Resource = {
...
+    const name =  labels['kuma.io/display-name'] ?? partialResource.name ?? ''
+    const zone =  labels['kuma.io/origin'] === 'zone' ? labels['kuma.io/zone'] ?? '' : ''
+
+    const kri = 'kri' in partialResource ? partialResource.kri ?? '' : ''
+
+    return {
+      ...partialResource,
+      config: partialResource,
+      id: partialResource.name,
+      kri: Kri.isKriString(kri) ? kri : Kri.toString({ shortName: `~${partialResource.type?.toLowerCase()}`, mesh, zone, namespace, name }),
```
johncowen, line 122: This might be a nit, but you have something called `const kri` that could potentially not be a KRI (it could be `''`) if this code where to grow, somebody might see the variable `kri` and think "ok this is a valid `kri` already.

In the same way that the `mesh, labels, namespace, zone` variables are all guaranteed correct at this point, personally I'd prefer:

```
const kri = 'kri' in partialResource ? Kri.isKriString(partialResource.kri) ? kri : Kri.toString({ shortName: `~${partialResource.type?.toLowerCase()}`, mesh, zone, namespace, name }) : ''
```

this is a bit of a nit though, so totally up to you

schogges: I'm slightly against inlining this as it gets harder to read, so I'll address this in a follow-up. Maybe just renaming it to `potentialKri` or `maybeKri` 😄

packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
johncowen: Oh I see it not actually the thing we use fir when we click the entire row, its just the badge in the other row 👍

johncowen (approved):
Ok good to go, thanks for bearing with me! 👍

PR #4791, opened 2026-03-30
chore(x-timespan): Use XTimespan everywhere

johncowen:
We recently made XTimespan to smooth over an awkward API of `AppAboutSection` which is just a wrapper around `KCard`(with a worse API)

This PR is an AI generated PR (done whilst I was approving dependency PRs) that refactors to use XTimespan everywhere and then remove XAboutCard (our `AppAboutSection` wrapper)

- Uses `XTimespan` instead of `XAboutCard:created` etc
- Replaces `XAboutCard` with just `XCard`
- Moves `XAboutCard:title` and uses the `XCard` `#title` slot instead (honestly I'd rather also avoid the `#title` slot and just use the default one but seeing as we currently have a mix anyway 🤷 )
- Removes the custom styles everywhere we had `XAboutCard` as the are no longer necessary/inside `XCard`


packages/kuma-gui/src/app/workloads/views/WorkloadDetailView.vue
```diff
@@ -16,12 +16,16 @@
...
-        :title="t('workloads.routes.item.about.title')"
-        :created="typeof props.data !== 'undefined' && !(props.data instanceof Error) ? props.data.creationTime : undefined"
-        :modified="typeof props.data !== 'undefined' && !(props.data instanceof Error) ? props.data.modificationTime : undefined"
+      <XCard
         data-testid="about-workload"
       >
+        <XTimespan
+          :start="typeof props.data !== 'undefined' && !(props.data instanceof Error) ? props.data.creationTime : undefined"
+          :end="typeof props.data !== 'undefined' && !(props.data instanceof Error) ? props.data.modificationTime : undefined"
+        />
```
schogges, line 25: If this was moved into `DataLoader` below we wouldn't need the conditionals

packages/kuma-gui/src/app/zones/views/ZoneDetailView.vue
```diff
@@ -15,13 +15,16 @@
...
-        :title="t('zone-cps.detail.about.title')"
-        :created="props.data && 'creationTime' in props.data ? props.data.creationTime : undefined"
-        :modified="props.data && 'modificationTime' in props.data ? props.data.modificationTime : undefined"
-        class="about-section"
+      <XCard
         data-testid="zone-about-section"
       >
+        <XTimespan
+          :start="props.data && 'creationTime' in props.data ? props.data.creationTime : undefined"
+          :end="props.data && 'modificationTime' in props.data ? props.data.modificationTime : undefined"
```
schogges, line 23: s.a.

packages/kuma-gui/src/app/workloads/views/WorkloadDetailView.vue
johncowen: We are gonna leave this here and follow up when merging the PRs that also touch this area

schogges: Ok, this needs to get sorted separately in the loaders because the DataLoader is not using the slot props yet. I'll deal with the conflicts then in the loaders PRs 🙂

schogges (approved):
LGTM 👍

PR #4819, opened 2026-04-09
refactor(services): show loaders in features

schogges:
See title

Covers both `services` and `external-services` modules.
While I was in the places I've converted `XAboutCards` to combination of `XCards` with `XTimespan` to reduce conflicts with https://github.com/kumahq/kuma-gui/pull/4791.


packages/kuma-gui/src/app/services/views/MeshExternalServiceDetailView.vue
```diff
@@ -13,106 +13,113 @@
...
-                      zone: props.data.zone,
-                    },
-                  }"
-                >
+          <template #title>
+            {{ t('services.mesh-external-service.about.title') }}
+          </template>
+          <DataLoader
+            :data="[props.data]"
+            v-slot="{ data: [service] }"
```
johncowen, line 24: Similar to another PR, do we need errors here? I'm thinking we might need to say "let's get used to using `result`" as that means we just pass one thing through.

johncowen: This one is using result also so all good!

johncowen (dismissed):

johncowen (approved):

PR #4823, opened 2026-04-10
refactor(zone-ingresses,-egresses): show loaders in features

schogges:
See title

Covers both `zone-ingresses` and `zone-egresses` modules and also required minor amends in `connections`.
While I was in the places I've converted `XAboutCards` to combination of `XCards` with `XTimespan` to reduce conflicts with https://github.com/kumahq/kuma-gui/pull/4791.


packages/kuma-gui/src/app/zone-egresses/views/ZoneEgressDetailView.vue
```diff
@@ -10,313 +10,346 @@
...
+                :src="uri(sources, '/connections/stats/for/:proxyType/:name/:mesh/:socketAddress', {
+                  name: route.params.proxy,
+                  mesh: '*',
+                  socketAddress: zoneEgress.zoneEgress.socketAddress,
+                  proxyType: 'zone-egress',
+                })"
+              >
+                <template #connecting>
+                  <XLayout
+                    variant="columns"
```
schogges, line 189: That's a special case here. I thought it might be good to show the loaders in both columns, basically having a `columns` loader variant. We don't have this variant yet and I'd like to get feedback wether this is something we would like to do. I'm happy to go back to just show one loader for now and think about this a little more. We have other places where we probably need some more variants to cover special cases i.e. a vertical list (policy types). Lmk!

johncowen: Looks good to me, yeah I think I'd like to show skeletons rather than two spinners with the same "Loading..." text doubled up, but I think this half way point is fine for now 👍

johncowen (dismissed):

johncowen (approved):

PR #4825, opened 2026-04-10
feat(resources): route to individual summary and detail views

### LINKED ISSUE #4824: [Resources]: Route to individual summary and detail views
### Description

From the resources view we start to route to the generic summary and detail view. Once this is done we can investigate the best way to route to the individual summary and detail views by using the KRI of a resource.

schogges:
Based on https://github.com/kumahq/kuma-gui/pull/4760 to work on the routing to individual resource summary and detail views separately.

---

This might hang for a while until the first part ☝️ is merged.

Closes https://github.com/kumahq/kuma-gui/issues/4824


johncowen:
Same as other related PR: We've done a lot of work on routing and KRIs over the past few months, and I have a feeling we can close this as it was a bit of an experiment towards that that we didn't go with, is that right @schogges ?

schogges:
> Based on #4760 to work on the routing to individual resource summary and detail views separately.
> 
> This might hang for a while until the first part ☝️ is merged.
> 
> Closes #4824

This was hanging a little longer and in the meantime we solved this already. I don't want to make the effort to dig out the PRs that actually solved it though, but the strategy was using the KRI protocol handler. Closing this.

PR #4830, opened 2026-04-13
feat(services): add KRI support

### LINKED ISSUE #4758: Support KRIs on service routes
### Description

In order to properly link from the generic resources listings we need support for KRIs in the services routes (i.e. mesh-service). In the generic resources we completely rely on KRIs and for service resources we currently can't use the information in the respective KRI because of the incomplete hash.

schogges:
This is adding support for KRIs in the `services` module.

Part of https://github.com/kumahq/kuma-gui/issues/4758

Closes https://github.com/kumahq/kuma-gui/issues/4758


schogges:
~https://github.com/kumahq/kuma-gui/pull/4830/changes/737db116ad128a67ff12acca7afb0ce28ed12415 temporarily uses the KRI to route from the different listing views to a service. Before merging I will remove it again, but I thought this makes it easier to test via preview or checking it out locally without needing to change anything. Once the resources are merged, we actually support both KRIs and traditional.~
Overridden with https://github.com/kumahq/kuma-gui/pull/4830/commits/b1046bc07b819d0ed28c5a117fe518348574673a

packages/kuma-gui/src/app/services/data/MeshMultiZoneService.ts
```diff
@@ -1,23 +1,27 @@
...
   search(query: string) {
     return Resource.search(query)
   },
 
-  fromObject(item: Generated) {
+  fromObject(item: KumaMeshMultiZoneService) {
     const labels = item.labels ?? {}
     const name = labels['kuma.io/display-name'] ?? item.name
     const namespace = labels['k8s.kuma.io/namespace'] ?? ''
+    const mesh = labels['kuma.io/mesh'] ?? item.mesh ?? ''
```
johncowen: I think this is the wrong way around? If it potentially has a `.mesh` property shouldn't that override the label?

Also if we are potentially using name for the kri do we need to remove potential namespaces? i.e. `name.namespace`?

I'm not sure how important this is, I think KRI is _always_ set just our OpenAPI says its not

Umm, we use mesh in the final object so maybe that one is important, worth thinking about this to make sure we aren't overwriting the "higher priority data"

packages/kuma-gui/src/app/services/views/MeshExternalServiceDetailView.vue
```diff
@@ -127,57 +127,63 @@
...
+            v-slot="{ data: [service] }"
           >
-            <DataCollection
-              type="hostnames"
-              :items="hostnames.items"
+            <DataLoader
+              :src="uri(servicesSources, '/meshes/:mesh/:serviceType/:serviceName/_hostnames', {
+                mesh: route.params.mesh,
+                serviceType: 'meshexternalservices',
+                serviceName: service.id,
```
johncowen, line 138: I think we can make this into some sort of KRI thing. Is this retrieving HostnameGenerator hostnames?

packages/kuma-gui/src/app/services/views/MeshExternalServiceDetailView.vue
```diff
@@ -232,9 +238,8 @@
 
             <template v-else>
               <DataLoader
-                :src="uri(servicesSources, '/meshes/:mesh/mesh-external-service/:name/as/kubernetes', {
-                  mesh: route.params.mesh,
-                  name: route.params.service,
+                :src="uri(servicesSources, '/mesh-external-service/:kri/as/kubernetes', {
+                  kri: route.params.kri,
```
johncowen, line 242: I'm I understanding right that we need to keep the `mesh-external-service` in the source URI here so that it always returns the same shaped thing?

packages/kuma-gui/src/app/services/sources.ts
```diff
@@ -59,35 +59,36 @@ export const sources = (api: KumaApi) => {
...
             format: 'kubernetes',
           },
         },
       })
-      return res.data!
+
+      return response.data!
     },
 
     '/meshes/:mesh/mesh-multi-zone-services': async (params) => {
```
johncowen, line 94: Its weird that we have a mix of non-mesh prefixed URIs and mesh prefixed URIs.

Lets not do it here as I'm currently only concerned with detail pages and their URLs but I'd like to cicle back and sort this.

I'm thinking something like:

```
/mesh-multi-zone-services/for/mesh/:mesh

// or

/mesh-multi-zone-services/:kri <== kri here is partial kri_mmsvc_default i.e. without a name
```

packages/kuma-http-api/mocks/src/meshes/_/meshexternalservices.ts
```diff
@@ -26,27 +25,34 @@ export default ({ fake, pager, env }: Dependencies): ResponseHandler => (req) =>
...
-            }
-            : {}),
+          ...fake.kuma.timespan(),
+          kri: fake.kuma.kri({ shortName, mesh, zone, namespace: nspace, name: displayName, sectionName: '' }),
+          labels: {
+            ...fake.kuma.labels({
+              name: displayName,
+              ...(zone ? { zone } : {}),
+              ...(k8s ? { namespace: nspace } : {}),
+            }),
```
johncowen, line 54: Doh this makes me think I didn't add the new labels to MeshServices etc 👀

johncowen (approved):
I left a few questions just to help me think about next steps, but otherwise this looks good to me!

packages/kuma-gui/src/app/services/views/MeshExternalServiceDetailView.vue
schogges: According to OAS it returns the (inspect) hostnames for a service. Basically generated hostnames, not a hostname generator.

packages/kuma-gui/src/app/services/views/MeshExternalServiceDetailView.vue
schogges: Yeah, if we want to keep the type-safety and don't want to deal with a union in the views we have to keep them separated

packages/kuma-gui/src/app/services/sources.ts
schogges: I had the same thought 😅 I'd prefer `/mesh-multi-zone-services/for/mesh/:mesh`

packages/kuma-gui/src/app/services/views/MeshExternalServiceDetailView.vue
johncowen: Ah thanks, gotcha 👍

packages/kuma-gui/src/app/services/sources.ts
johncowen: I thought you might 😆 , ok we can discuss later. Lets keep going on the details views 👍

packages/kuma-gui/src/app/services/views/MeshExternalServiceDetailView.vue
johncowen: Just so I'm clear: and those generated hostnames come from a hostname generator?

johncowen:
Don't miss I approved this!

packages/kuma-gui/src/app/services/views/MeshExternalServiceDetailView.vue
schogges: I think so, yes

johncowen: Gotcha, thanks!

FYI: I'm thinking we should move this one to the hostname generator module and rename it something like `/hostname-generators/hostnames/for/:kri` or similar.

Not to do here but just to think out loud.

PR #4836, opened 2026-04-15
[WIP] Use/implement new PageLayout

johncowen:
Putting up in Draft for visibility, I'm guessing this will eventually be close, but it gives us somewhere to write things and share against the code if we need to.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1809,6 +1809,22 @@
         "vue": ">= 3.3.13 < 4"
       }
     },
+    "node_modules/@kong-ui-public/page-layout": {
+      "version": "0.2.12",
+      "resolved": "https://registry.npmjs.org/@kong-ui-public/page-layout/-/page-layout-0.2.12.tgz",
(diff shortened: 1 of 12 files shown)


johncowen:
note: I removed all the warnings stuff in AppView to start closer to a clean slate with the total re-coding of the template. I need to put that back before this is even close to approval

johncowen:
Quick note here: Its highly unlikely that we are going to take the route in this PR. But it was useful to prove how difficult and problematic this component was to use in its current state. Even after all this, this PR cannot be merged because its causes a significant amount of double-rendering.

TL;DR Not exposing the `router-view` to us means we can't set props on RouterViews (see standard Vue documentation https://router.vuejs.org/guide/advanced/router-view-slot.html#Passing-props-and-slots)

I made a larger explanation offline, if anyone is reading this and would like to know more, please ask/ping me.

Luckily we've managed to get some changes made to the PageLayout component to allow us to use `router-view` props, so there will be another PR upcoming. Once that new PR is approved and merged this one can be closed, but we might use it for reference at a later date.

johncowen:
Closing this up, we went a different route for the moment

PR #4837, opened 2026-04-15
fix: fixes an issue with margins of the version header

johncowen:
Fixes an issue where this is too much top margin on our software version info:

## Before

Screenshot 2026-04-16 at 10 09 48

Note: I also removed a piece of global CSS that nobody quite knows why it was there in the first place (putting a margin top on lots of different HTML elements). Whilst the `p` selector here is what is causing the bug, I've removed the entire thing here and it looks to me like it doesn't effect anything.

### DIFF
--- a/packages/x/src/components/x-theme/XTheme.vue
+++ b/packages/x/src/components/x-theme/XTheme.vue
@@ -466,6 +466,11 @@ import '@kong-ui-public/app-layout/dist/style.css'
   a.k-button {
     text-decoration: none !important;
   }
+
+  /* styles for things that come from locales/i18n rendered from markdown */
+  .x-i18n p {
+    margin-bottom: var(--x-space-50);
+  }
 }
 </style>
 
@@ -542,19 +547,6 @@ import '@kong-ui-public/app-layout/dist/style.css'
   ul,
   ol {
     padding-left: var(--x-space-80);
-  }
-
-  h2,
-  h3,
-  h4,
-  h5,
-  h6,
-  p,
-  ul,
-  ol {
-    &:not(:first-child) {
-      margin-top: var(--x-space-50);
-    }
   }
 
   p:empty {
--- a/packages/x/src/components/x-theme/XTheme.vue
+++ b/packages/x/src/components/x-theme/XTheme.vue
@@ -466,11 +466,6 @@ import '@kong-ui-public/app-layout/dist/style.css'
   a.k-button {
     text-decoration: none !important;
   }
-
-  /* styles for things that come from locales/i18n rendered from markdown */
-  .x-i18n p {
-    margin-bottom: var(--x-space-50);
-  }
 }
 </style>
 
--- a/packages/x/src/components/x-i18n/XI18n.vue
+++ b/packages/x/src/components/x-i18n/XI18n.vue
@@ -110,3 +110,14 @@ const safeT: TFunction = (
 }
 
 </script>
+<style lang="scss" scoped>
+/* default styles for markdown content */
+/* note whilst it doesn't seem necessary */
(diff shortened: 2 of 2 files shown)


packages/x/src/components/x-theme/XTheme.vue
```diff
@@ -466,6 +466,11 @@ import '@kong-ui-public/app-layout/dist/style.css'
   a.k-button {
     text-decoration: none !important;
   }
+
+  /* styles for things that come from locales/i18n rendered from markdown */
+  .x-i18n p {
+    margin-bottom: var(--x-space-50);
```
johncowen: I'd actually prefer this to be `1em` i.e. the space between paragraphs should be dependent on the size of the text. Let me know if you agree and we can change. I've not done straight away myself because we usually use tokens everywhere - but I think in this case this is somewhere where it we had a paragraph font size on say 12px in one place and 22px in another place, using this token would be totally wrong.

schogges: I generally agree to prefer relative sizing (i.e. `em`, `rem`), but I also prefer consistency in other places. I would set a token in x-theme and use it in XI18n, rather than adding component specific CSS in here 🤔

johncowen: The reason its not in X18n is that this is theming that should be controlled by the external user of XI18n, therefore shouldn't be added in X18n (ideally, I know we sometimes add kuma things into `x` when absolutely necessary). Overall this is exactly how XTheme should be used. Do you follow my thinking?

johncowen: btw this is a case for X18n only because it renders markdown, and the styling of the markdown should be entirely controlled by the external user of X18n, not X18n itself

schogges: Not sure if I follow 🤔 you mean that XI18n by itself should never have the margin only when used in combination with XTheme? Why? External users could still set margins, no?

johncowen: > Not sure if I follow 🤔 you mean that XI18n by itself should never have the margin only when used in combination with XTheme? 

Yes exactly.

> Why? External users could still set margins, no?

Yes, via XTheme.

Maybe you are asking that we set a good default in X18n but still allow external users to override that via XTheme if necessary?

schogges: > Maybe you are asking that we set a good default in X18n but still allow external users to override that via XTheme if necessary?

Yes, for example via token 🤔

johncowen: Gotcha, brb 👍

johncowen: Ok I added a commit here that puts a default of `1em` inside of `XI18n` and check that it is overridable via the CSS property (but I didn't overwrite it)

Just to note I really want to avoid having lots and lots of CSS properties like:

```
--x-i18n-p-margin-block: ...
--x-i18n-p-text-color: ...
--x-i18n-p-padding-block: ...
--x-i18n-p-font-size: ...
--x-i18n-h1-margin-block: ...
--x-i18n-h1-text-color: ...
--x-i18n-h1-padding-block: ...
--x-i18n-h1-font-size: ...
```

I think I have a better solution to this, but its not "run of the mill" so I went with this for now, and hope to follow up with something that avoids "millions" of CSS properties.

Oh lastly, not sure if I understood, but if you want me to overwrite the 1em with our own token let me know. I left it as 1em for now because that the default, which in my mind (and I think yours) is a better default.

schogges (approved):
👍

PR #4838, opened 2026-04-16
fix(oas): reduce allOf usage for simple types

johncowen:
Removes usage of `allOf` for simple types. Originally we were using this for a bit of reuse, but this is definitely the case that a bit of copy/pasta is better.

It looks like there is quite a bit of churn here, but its just taking the contents of Entity.yaml and PagedCollection.yaml and pasting them into the main entity (MeshInsights.yaml, MeshInsightsCollection.yaml, PolicyCollection.yaml)

All in all it fixes the lint issue currently in https://github.com/kumahq/kuma-gui/pull/4601 (openapi-fetch upgrades somehow has meant that the allOfs are working as intended, removing the allOfs makes the iissue go away)


schogges (approved):
👍

PR #4849, opened 2026-04-20
chore: move tertiary menus out of actions slot

johncowen:
Since now we have `XLayout variant="y-stack"` and `XLayout variant="action-group"` there is no reason to put these in the `AppView #actions` slot. They were only in actions originally for positioning purposes, and they are really "actions"


johncowen:
Took this out of draft as I'm pretty certain we should do this no matter what, but FYI it is very related to some work I am working on.

schogges (approved):

PR #4851, opened 2026-04-21
feat(dataplanes,zones,zone-ingresses,-egresses): support KRI

### LINKED ISSUE #4757: Support KRIs on dataplane and dataplane-like routes
### Description

In order to properly link from the generic resources listings we need support for KRIs in the dataplane and dataplane-like routes (i.e. zone-ingress and zone-egress). In the generic resources we completely rely on KRIs and for dataplane-like resources we currently can't use the information in the respective KRI because of the incomplete hash.

schogges:
This implements support for KRIs in the views of the following modules:
- dataplanes
- legacy-dataplanes
- zones
- zone-ingresses
- zone-egresses

We can only fetch a resource by its KRI via the generic KRI API (`/_kri/{kri}`). Therefore I've changed the sources of fetching a single resource (i.e. `/meshes/:mesh/dataplanes/:name`) to check wether the provided identifier is a KRI. If this is the case then we fetch the generic KRI API.
In a lot of cases we use the computed endpoints for our views (i.e. `/meshes/:mesh/dataplane-overviews/:name`). These endpoints do not support KRIs and therefore I had to wrap these in a call to the generic API first to extract the `name` (normalized to `id`) which is then being used to fetch the computed and other endpoints that do not support KRIs. We might get support via https://github.com/kumahq/kuma/issues/16322.

---

We will support both ways for a while, especially for listings that use overview list endpoints and therefore I've extended our tests with examples to test both strategies.

Closes https://github.com/kumahq/kuma-gui/issues/4757


packages/kuma-gui/src/app/data-planes/views/DataPlaneSummaryView.vue
```diff
@@ -11,238 +11,180 @@
...
+        }) : undefined"
+        v-slot="{ data: [dataplaneOverview] }"
       >
-        <template
-          v-for="item in [proxies[0]]"
-          :key="item.id"
+        <DataCollection
+          :items="typeof props.items !== 'undefined' ? props.items : [dataplaneOverview]"
+          :predicate="item => item.id === route.params.proxy"
         >
```
schogges, line 36: This is pre-work for nesting summary views in the resources outside of the usual dataplane listing. The main difference between the usual dataplane listing and the dataplane listing in the resources is that the current one uses dataplane overviews and the one in the resources uses plain dataplanes (`/meshes/{mesh}/dataplanes`). In the listing in the resources we will use a KRI for the navigation and will not pass any props.

This conditionally makes calls to the generic KRI API and the `_overview` endpoint to retrieve the data. In the `DataCollection` we use either the `items` provided via `props` or the response from the `_overview` endpoint as a single entry.

The same applies to the summary views in `zone-ingresses` and `zone-egresses`.

packages/kuma-gui/src/app/zone-egresses/views/ZoneEgressDetailTabsView.vue
```diff
@@ -7,11 +7,11 @@
...
   >
-    <DataSource
-      :src="uri(sources, '/zone-egress-overviews/:name', {
+    <DataLoader
+      :src="uri(sources, '/zone-egresses/:name', {
         name: route.params.proxy,
       })"
-      v-slot="{ data, error, result }"
+      v-slot="{ data: [zoneEgress] }"
     >
```
schogges, line 15: This is very unfortunate that I had to do this. It conflicts with the pattern to show the loaders in the features. Prefetching the `ZoneEgress`, `ZoneIngress` or `Dataplane` is required to get the `name` for subsequent requests to other endpoints, i.e. the `_overview` endpoints.
I chose this approach for now otherwise I'd need to do the same thing in all the different places.

packages/kuma-http-api/mocks/src/zones/_.ts
```diff
@@ -1,6 +1,26 @@
 import type { Dependencies, ResponseHandler } from '#mocks'
 
 export default ({ fake }: Dependencies): ResponseHandler => (req) => {
+  // this template can be called via the /_kri/kri_<shortName>_:kri endpoint or
+  // the legacy endpoint
+  const kri = req.params.kri ? `kri_ze_${req.params.kri}` : undefined
```
johncowen:
```suggestion
  const kri = req.params.kri ? `kri_z_${req.params.kri}` : undefined
```

packages/kuma-gui/src/app/connections/views/ConnectionInboundSummaryClustersView.vue
```diff
@@ -18,55 +18,63 @@
...
 import { sources } from '../sources'
 import { DataplaneNetworkingLayout } from '@/app/data-planes/data'
-import type { DataplaneInbound } from '@/app/data-planes/data'
+import type { DataplaneInbound, DataplaneOverview } from '@/app/data-planes/data'
+import type { ZoneEgressOverview } from '@/app/zone-egresses/data'
+import type { ZoneIngressOverview } from '@/app/zone-ingresses/data'
 const props = defineProps<{
   routeName: string
   data: DataplaneInbound | DataplaneNetworkingLayout['inbounds'][number]
+  overview: DataplaneOverview | ZoneIngressOverview | ZoneEgressOverview | Error | undefined
```
schogges, line 78: Here and in the other places I've added/aligned the `overview` prop vs `dataPlaneOverview` prop and made sure that it's potentially `Error`/`undefined`.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -21,6 +21,7 @@ const dpTypes = {
 } as const
 
 export const DataplaneOverview = {
+  // TODO: use export type KumaDataplaneOverview = NonNullable<components['schemas']['DataplaneOverviewWithMeta']>
```
schogges, line 24: Need to check in a follow-up, this wasn't straight-forward

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue
```diff
@@ -86,12 +86,7 @@
                   </th>
                   <td>
                     <XAction
-                      :to="{
-                        name: 'zone-cp-detail-view',
-                        params: {
-                          zone: item.zone,
-                        },
-                      }"
+                      :href="t('common.label.href.kuma~io/zone', { name: item.zone })"
```
schogges: Zone routes now use a KRI so I need to link via KRI. In the listings we only have the name, which is derived from the zone label. So I had to use the protocol handler already. I will cover the rest of the linking in a follow-up.

packages/kuma-http-api/mocks/src/zones/_.ts
schogges: Thanks! 👍

packages/kuma-gui/src/types/index.d.ts
```diff
@@ -341,8 +341,9 @@ export interface LabelValue {
 export interface Entity {
   type: string
   name: string
-  creationTime?: string
-  modificationTime?: string
+  kri?: string
+  creationTime: string
+  modificationTime: string
```
johncowen: Lets check this

packages/kuma-gui/src/app/zones/views/ZoneDetailTabsView.vue
```diff
@@ -25,7 +25,7 @@
         <template #title>
           <h1>
             <RouteTitle
-              :title="route.params.zone"
+              :title="Kri.fromString(route.params.zone).name"
```
johncowen, line 28: `route.params.zone` is now a KRI, so in order to get the name we need to pull it out of the KRI. We can;t use `data?.name` because this will load after any statically set child titles and overwrite the child title.

We need a larger change in RouteView to sort this (we've hit this issue a couple of times now from different angles)

packages/kuma-gui/src/app/zone-ingresses/sources.ts
```diff
@@ -21,8 +22,9 @@ export const sources = (api: KumaApi) => {
     fetch: api.client.fetch,
   })
   return defineSources({
-    '/zone-cps/:name/ingresses': async (params) => {
-      const { name, size, page } = params
+    '/zone-cps/:zone/ingresses': async (params) => {
```
johncowen: Shall we change this to from `:zone` to `:kri`?

packages/kuma-gui/src/app/zone-ingresses/data/index.ts
```diff
@@ -17,8 +17,9 @@ export const ZoneIngress = {
   fromObject: (item: PartialZoneIngress) => {
     return {
       ...item,
+      kri: item.kri ?? Kri.toString({}),
```
johncowen: This looks weird, realistically we never use the fallback here, but if we are adding a faillback we should see if we can make a proper KRI, maybe we con't have the information for that here, in which case maybe we make it in whatever is passing the PartialZoneIngress in here.

packages/kuma-gui/src/app/zone-egresses/views/ZoneEgressDetailTabsView.vue
```diff
@@ -7,11 +7,11 @@
     }"
     v-slot="{ route, can, t, uri }"
   >
-    <DataSource
-      :src="uri(sources, '/zone-egress-overviews/:name', {
-        name: route.params.proxy,
+    <DataLoader
+      :src="uri(sources, '/zone-egresses/:kri', {
```
johncowen, line 11: Self-note: this is the thing with the dataplane hash

packages/kuma-gui/src/types/index.d.ts
johncowen: I think we need to added the optionals back to the times.

packages/kuma-gui/src/app/zone-egresses/data/index.ts
```diff
@@ -103,6 +102,8 @@ export const ZoneEgressOverview = {
           type: 'ZoneEgress',
           name: item.name,
           mesh: item.mesh,
+          creationTime: item.creationTime,
+          modificationTime: item.modificationTime,
```
johncowen: If we add the optionals back to the types for these, these can be removed.

packages/kuma-gui/src/app/workloads/views/WorkloadDetailView.vue
```diff
@@ -47,12 +47,7 @@
                 <dd>
                   <XAction
                     v-if="props.data.zone"
-                    :to="{
-                      name: 'zone-cp-detail-view',
-                      params: {
-                        zone: props.data.zone,
-                      },
-                    }"
+                    :href="t('common.label.href.kuma~io/zone', { name: props.data.zone })"
```
johncowen: We wondered about this because we shouldn't be using the label i18n string for these

Ideally it would be:

```vue
:href="`kri://${kri({ resourceName: 'z', name: row.zone})}`"
```

or even this for now:

```vue
:href="`kri://${Kri.toString({ shortName: 'z', name: row.zone})}`"
```

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue
johncowen: We wondered about this because we shouldn't be using the label i18n string for these

Ideally it would be:

```vue
:href="`kri://${kri({ resourceName: 'z', name: row.zone})}`"
```

or even this for now:

```vue
:href="`kri://${Kri.toString({ shortName: 'z', name: row.zone})}`"
```

packages/kuma-gui/src/app/policies/data/PolicyDataplane.ts
```diff
@@ -6,13 +7,21 @@ import type {
 export const PolicyDataplane = {
   fromObject(item: PartialPolicyDataplane) {
     const labels = typeof item.labels !== 'undefined' ? item.labels : {}
+    const mesh = item.mesh
+    const zone = labels['kuma.io/zone'] ?? ''
+    const displayName = labels['kuma.io/display-name']
+    const name = displayName ?? item.name
+    const namespace = labels['k8s.kuma.io/namespace'] ?? ''
```
johncowen, line 13: I think it might be possible to copy/paste the pattern from MeshTrust

https://github.com/kumahq/kuma-gui/blob/abda515e49c8cb36f79385a95c4bdff1e4a66127/packages/kuma-gui/src/app/mesh-trusts/data/MeshTrust.ts#L9-L14

packages/kuma-gui/src/app/legacy-data-planes/views/DataPlaneInboundSummaryOverviewView.vue
```diff
@@ -103,7 +103,7 @@
             <DataSource
               :src="uri(sources, '/meshes/:mesh/rules/for/:dataplane', {
                 mesh: route.params.mesh,
-                dataplane: route.params.proxy,
+                dataplane: props.overview.id,
```
johncowen, line 106: -`route.params.proxy` is now a KRI
-`props.overview.id` is a `name.namespace`

The source and underlying HTTP request requires a `name.namespace` and name is one of those things that can have a hash in it.

packages/kuma-gui/src/app/zone-egresses/views/ZoneEgressDetailTabsView.vue
johncowen: We don't particular love this because we have too many DataLoaders in one page. Ideally we want to try and keep one "overall loading state" (they might be lots of different loaders but one state) in the page.

Something to maybe come back to.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -39,7 +40,7 @@ export const DataplaneOverview = {
     const name = labels['kuma.io/display-name'] ?? item.name
 
     // temporarily make a KRI until we have those from the backend
```
johncowen: Lets remove this comment

packages/kuma-gui/src/app/data-planes/data/Dataplane.ts
```diff
@@ -1,13 +1,15 @@
...
 
-type KumaDataplane = NonNullable<components['schemas']['DataplaneItem']>
+export type KumaDataplane = NonNullable<components['schemas']['DataplaneItem']>
 type KumaDataplaneNetworking = NonNullable<components['schemas']['DataplaneItem']['networking']>
 
 export const Dataplane = {
   fromObject(partialDataplane: KumaDataplane) {
     return {
       ...partialDataplane,
+      kri: partialDataplane.kri ?? '',
```
johncowen: how come this isn't using `Kri.toString()`? Let's check if we have the information needed to make a KRI here (we need to use the displayName label not the name as the name as the weird hash). If we can construct it somehow, let do it. But, even then this is very unlikely to be significant as the backend should always give us a kri.

johncowen:
A big thing we felt here was that moving forwards we can probably request overviews less and less because there is hardly any information that we need from them. It would be better to get say `ZoneIngress` first, and then _only if_ we need subscriptions (where we compute the state from) only then should we request a `ZoneIngressOverview`. Its almost like inverting what we currently have.

If we did this we could maybe stop doing the weird "fetch an overview and then copy its labels onto the resource"

packages/kuma-gui/src/app/data-planes/data/Dataplane.ts
```diff
@@ -1,13 +1,15 @@
 import { DataplaneNetworking } from './DataplaneNetworking'
 import type { components } from '@kumahq/kuma-http-api'
 
-type KumaDataplane = NonNullable<components['schemas']['DataplaneItem']>
+export type KumaDataplane = NonNullable<components['schemas']['DataplaneItem']>
```
johncowen, line 5: self-note: We needed to export this so that we can cast the kri's endpoint call to the correct shape.

johncowen:
We were thinking here as a follow up we could try and think of a test/tripwire that would fail if we tried to make a KRI for a dataplane that used one of the weird hashes. So that if we mistakenly tried to do this the test would tell us not to.

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue
johncowen: We are gonna try do something here.

packages/kuma-gui/src/types/index.d.ts
johncowen: We are gonna do this here.

packages/kuma-gui/src/app/zone-ingresses/sources.ts
johncowen: We are gonna try and do this.

packages/kuma-gui/src/app/zone-ingresses/data/index.ts
johncowen: We are gonna look at this one, but it might be painful if so shout and we can see if we wanna just leave

packages/kuma-gui/src/app/zone-egresses/data/index.ts
johncowen: We are gonna do this here

packages/kuma-gui/src/app/policies/data/PolicyDataplane.ts
johncowen: We are gonna do this here

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
johncowen: We are gonna do this here

johncowen:
```suggestion
```

packages/kuma-gui/src/app/data-planes/data/Dataplane.ts
johncowen: Similar to the Kri.toString things above, lets try do this here if not painful.

packages/kuma-gui/src/app/data-planes/data/Dataplane.ts
johncowen: I just remembered this morning: Lets not worry about doing this here, but a tiny while back I considered this and decided I would prefer to re-import from `{ paths } from 'kuma-http-api'` where-ever we need it and avoid exporting Kuma shaped things from the data layer. My thinking was I want to make these Kuma shapes as hard as possible to import, if its "locally exported" its easier to import than importing from a 3rd party import.

We can circle back and sort this its only a little nit anyhow.

johncowen:
Re-looking at this now

packages/kuma-gui/src/app/data-planes/views/DataPlaneListView.vue
```diff
@@ -185,12 +185,7 @@
               <template #zone="{ row }">
                 <XAction
                   v-if="row.zone"
-                  :to="{
-                    name: 'zone-cp-detail-view',
-                    params: {
-                      zone: row.zone,
-                    },
-                  }"
+                  :href="`kri://${Kri.toString({ shortName: 'z', name: row.zone })}`"
```
johncowen, line 188: Even though I would rather not add an import just to do this, I much prefer this to using the "label to KRI" i18n strings (which are only for looping through arbitrary lists of labels)

Plus if we're gonna add `v-slot="{ kri }"` to `RouteView` that removes my only nit with this.

packages/kuma-gui/src/app/data-planes/sources.ts
```diff
@@ -7,6 +7,7 @@ import {
   MeshGatewayDataplane,
   SidecarDataplane,
   DataplaneNetworkingLayout,
+  type KumaDataplane,
```
johncowen, line 10: I'm sure we used to have a lint rule to make us always use the `import type {} from ''` form for types. No biggie/not a blocker, but this made me wonder.

packages/kuma-gui/src/app/gateways/views/BuiltinGatewayListView.vue
```diff
@@ -84,12 +84,7 @@
                 <template #zone="{ row }">
                   <template v-if="row.labels && row.labels['kuma.io/origin'] === 'zone' && row.labels['kuma.io/zone']">
                     <XAction
-                      :to="{
-                        name: 'zone-cp-detail-view',
-                        params: {
-                          zone: row.labels['kuma.io/zone'],
-                        },
-                      }"
+                      :href="`kri://${Kri.toString({ shortName: 'z', name: row.labels['kuma.io/zone'] })}`"
```
johncowen, line 87: Not a blocker at all, just half thinking out loud:

This is an interesting one, I wondered if there was an `item.zone` to use here, which if I remember correctly has already been calculated depending on `origin` and `zone` labels.

Anyway, this is gateways I suppose so is potentially being deleted soonish.

packages/kuma-gui/src/app/zone-egresses/data/index.ts
```diff
@@ -17,6 +17,8 @@ export const ZoneEgress = {
     return {
       ...item,
       config: item,
+      kri: item.kri ?? '',
```
johncowen, line 20: A default, blank KRI, this might be one of the ones that is difficult to create, lets not block on this. But we should check for these later.

packages/kuma-gui/src/app/zone-egresses/sources.ts
```diff
@@ -21,9 +22,10 @@ export const sources = (api: KumaApi) => {
     fetch: api.client.fetch,
   })
   return defineSources({
-    '/zone-cps/:name/egresses': async (params) => {
-      const { name, size, page } = params
-      const filter = name !== '*' ? {
+    '/zone-cps/:zone/egresses': async (params) => {
```
johncowen, line 25: I'm pretty sure we need to try and make these things called `:kri`, I think we said this is ok to do in sources (mostly). Again one to circle back on.

packages/kuma-gui/src/app/zone-ingresses/views/ZoneIngressSummaryView.vue
```diff
@@ -47,9 +48,9 @@
                     :to="{
                       name: 'zone-ingress-detail-view',
                       params: {
-                        zone: item.zoneIngress.zone,
+                        zone: route.params.zone,
```
johncowen, line 51: Not super important and not a blocker, but wondered why this was changed to take a value from the route params and render/use it rather than use the data we've loaded from the backend.

Whilst its unavoidable in a lot of places we should always prefer data loaded from the backend instead of rendering/using text from the route params.

This will be one of those ones where "it depends" I guess, but I guess just keep in mind to have a preference for data from the backend rather than user editable route params.

johncowen (dismissed):
A left a few comments and "thinking out louds" but nothing blocking.

Thanks for doing this, lots of tedious but careful work 👍

packages/kuma-gui/src/app/zone-egresses/sources.ts
schogges: Oh seems like we missed that one during our review 😅

johncowen: If you wanna sort it real quick ping me when its done and I can give it another 👍

schogges:
@johncowen need another approval after rebasing and resolving conflicts 🙂

johncowen (approved):

PR #4856, opened 2026-04-22
chore(deps): remove actions/github-script dependency

johncowen:
I noticed in https://github.com/kumahq/kuma-gui/pull/4853 that we only have one single usage of `actions/github-script`.

I wondered if we could just use `curl` or `gh` instead (`gh` upgrades and security upgrades etc are managed by the OS)

I asked Claude and it gave me this, which seems fine to be (I also asked if it was "absolutely sure!" 😆 )

We won't know until we try and use it on master though. It it doesn't work we can revert.

If it does its one less dependency, one less renovate PR every so often.

### DIFF
--- a/.github/workflows/dispatch-merged-pr-notification.yml
+++ b/.github/workflows/dispatch-merged-pr-notification.yml
@@ -24,4 +24,4 @@ jobs:
           gh api repos/${{ vars.NOTIFY_REPO_OWNER }}/${{ vars.NOTIFY_REPO_NAME }}/dispatches \
             -X POST \
             -f event_type=${{ vars.NOTIFY_EVENT_TYPE }} \
-            -f client_payload[branch]=${{ github.ref_name }}
+            -f 'client_payload[branch]=${{ github.ref_name }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/dispatch-merged-pr-notification.yml
```diff
@@ -18,16 +18,10 @@ jobs:
...
-                branch
-              },
-            })
+      - env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh api repos/${{ vars.NOTIFY_REPO_OWNER }}/${{ vars.NOTIFY_REPO_NAME }}/dispatches \
+            -X POST \
+            -f event_type=${{ vars.NOTIFY_EVENT_TYPE }} \
+            -f client_payload[branch]=${{ github.ref_name }}
```
schogges: I think we have to wrap it in string, see Examples in https://cli.github.com/manual/gh_api

```suggestion
            -f 'client_payload[branch]=${{ github.ref_name }}'
```

I tried it locally and I had to wrap it.

johncowen: Ah nice 😂 , thankyou! I actually questioned Claude over the syntax as it looked weird to me. Of course he said he was "absolutely certain" I guess he just missed the quotes 😆

schogges (approved):
Let's give it a try 😁

PR #4858, opened 2026-04-22
feat(hostname-generators): support KRIs

### LINKED ISSUE #4854: Support KRIs on HostnameGenerator routes
### Description

In order to properly link from the generic resources listings we need support for KRIs in the hostname generator routes. In the generic resources we completely rely on KRIs and for hostname generator resources we currently can't use the information in the respective KRI.

schogges:
This supports KRI in routes of the `hostname-generators` module.

I've remodeled the mocks a little to be more correct and real world alike.

Closes https://github.com/kumahq/kuma-gui/issues/4854


packages/kuma-gui/features/hostname-generators/Item.feature
```diff
@@ -24,25 +24,55 @@ Feature: hostname-generators / item
...
     And the "$about-section" element contains "zone-1"
     And the "$about-section" element contains "kuma.io/origin:zone"
     And the "$about-section" element contains "kuma.io/env:kubernetes"
 
-  Scenario: Visiting the detail view of HostnameGenerator
-    When I visit the "/hostname-generators/local-mesh-external-service/overview" URL
+    Examples:
+      | Name                                                    |
+      | local-mesh-external-service                             |
+      | kri_hg__zone-1_kuma-system_local-mesh-external-service_ |
```
schogges: Adding Examples to test both approaches: using KRI and not using KRI

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
```diff
@@ -57,12 +53,16 @@
               :errors="[error]"
               v-slot="{ data: [data] }"
             >
+              <XTimespan
+                :start="data.creationTime"
+                :end="data.modificationTime"
+              />
```
schogges, line 65: Nest `XTimespan` into `DataLoader` to avoid conditional chaining

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorDetailView.vue
```diff
@@ -57,12 +53,16 @@
...
               <XLayout
                 variant="y-stack"
               >
                 <XLayout variant="y-stack">
                   <XDl>
-                    <div>
+                    <div v-if="data.namespace">
```
schogges, line 71: Found that we didn't check for this, which could lead to empty description list element

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue
```diff
@@ -10,136 +10,121 @@
...
+        name: route.params.name,
+      }) : undefined"
+      v-slot="{ data: [hostnameGenerator] }"
     >
-      <template
-        #item="{ item }"
+      <DataCollection
+        :items="props.items ?? [hostnameGenerator]"
+        :predicate="item => item.id === route.params.name || item.kri === route.params.name"
       >
```
schogges, line 19: Pre-work to allow nesting `HostnameGeneratorSummaryView` in other routes in which we do not pass props to nested views.

packages/kuma-gui/src/app/hostname-generators/data/HostnameGenerator.ts
```diff
@@ -45,7 +45,7 @@ export const HostnameGenerator = {
     }
   },
 
-  fromCollection(collection: HostnameGeneratorList) {
+  fromCollection(collection: KumaHostnameGeneratorCollection) {
```
johncowen, line 53: Thanks for doing these bits 👍

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue
johncowen: We briefly touched on this yesterday, but as far as I'm aware we don't _need_ to do this. I'd rather leave this out of here until we find that we do need to.

I actually don't mind the pattern and it makes it feel nice that "anyway you want to get the data in will work" but...

I guess we want to discourage loading data in Summary views I went into it briefly yesterday and I'll try find more info (and I'll write something here), but long while back the summary views were a way to very quickly see data from table rows without further making HTTP requests and therefore avoid load waits.

The idea is that we shouldn't be using tiny thin table rows to show loads of data. 

The rows are a way to identify the thing you are interested in and then a way to get somewhere else to see more. Summary views make it a lot easier to _compare_ things before you can click backwards and forwards between rows really easily without waiting for things to load.

Alternatives mean either:

- Overloading the table rows with tonnes of data horizontally that never fits in your screen so you are forever scrolling backwards and forwards (or hiding and showing columns)
- Overloading table rows with tonnes of data vertically making it hard to identify the thing you are looking for without scrolling or narrowing your results via searching (if a search facet is available). This also makes it really hard to compare things as you have way less rows in one page.
- Clicking results in a data reload, which means brief pauses when trying to compare things.

Focussed table rows that are "a way to identify and compare things and/or a way to get somewhere" plus Summary views (plus deeper HTTP loading detail views) are the approach we chose to address these things.

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue
```diff
@@ -162,6 +169,6 @@ import { sources } from '../sources'
 import { YAML } from '@/app/application'
 import type { HostnameGenerator } from '@/app/hostname-generators/data'
 const props = defineProps<{
-  items: HostnameGenerator[]
+  items: HostnameGenerator[] | undefined
```
johncowen: Can items be erroneous?

packages/kuma-gui/src/app/hostname-generators/sources.ts
```diff
@@ -34,33 +36,59 @@ export const sources = (api: KumaApi) => {
...
+          params: {
+            path: {
+              name,
+            },
+          },
+        })
+      }
 
-      return HostnameGenerator.fromObject(res.data!)
+      return HostnameGenerator.fromObject(response.data as KumaHostnameGenerator)
```
johncowen, line 46: I like that we are hiding the transition from old style to new KRI-style from the application engineer 👍

johncowen (commented):
Generally, looks good to me!

I left a (long!) comment on the thing we spoke about yesterday (I'll try find more info on that), plus a thing about errors, lmk once those are covered and I should be able to 👍 this

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue
schogges: No, this is just that there is potentially no prop set. For example if the parent view fetched and errored, there would be no row displayed that could link to a summary view.

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorSummaryView.vue
schogges: I've reverted it and we can check back again when it comes to the actual routing

schogges:
I have resolved all the resolvable previous comments to start fresh 🙂

packages/kuma-http-api/mocks/src/hostnamegenerators/_.ts
```diff
@@ -4,12 +4,25 @@ import type { components } from '@kumahq/kuma-http-api'
 type HostnameGenerator = components['responses']['HostnameGeneratorItem']['content']['application/json']
 
 export default ({ fake, env }: Dependencies): ResponseHandler => (req) => {
-  const { params } = req
-  const name = params.name as string
-  const [displayName = '', namespace = ''] = name.split('.')
-  const meshServiceTypeSelector = fake.kuma.meshServiceTypeSelector()
   const k8s = env('KUMA_ENVIRONMENT', 'universal') === 'kubernetes'
-  const creationTime = fake.date.past()
+  const kri = req.params.kri ? `kri_mal_${req.params.kri}` : undefined
```
johncowen:
```suggestion
  const kri = req.params.kri ? `kri_hg_${req.params.kri}` : undefined
```

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorListView.vue
```diff
@@ -95,7 +95,7 @@
                       :to="{
                         name: 'hostname-generator-detail-view',
                         params: {
-                          name: item.id,
+                          kri: item.kri,
```
johncowen, line 98: I'd like to start using `kri://` links, wdyt of changing these now and using those here and in the SummaryView?

johncowen (commented):
Looks good, looks more straightforwards now we are leaving behind using `.name`

2 tiny things but I think this is ready to go!

packages/kuma-http-api/mocks/src/hostnamegenerators/_.ts
schogges: Oh 😆 thanks!

packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorListView.vue
schogges: I'd prefer to do this in follow-ups. This is mainly about fetching and minor amends to the routes. The linking is a separate topic.

johncowen: Can that be an immediate follow up after this PR (i.e. before we roll this pattern out to other modules)?

schogges: Yeah, I think I can do that 👍

johncowen (approved):
Looks good 👍

PR #4859, opened 2026-04-22
chore(meshes): remove breadcrumb from mesh list

johncowen:
Currently we have a `Meshes` breadcrumb _and_ `Meshes` title on the mesh listing page.

If its in the title we don't need the breadcrumb.

---

Note: This isn't ideally how I'd like to do it, but taking a quick look at zones (where the breadcrumbs are correct), this should be possible by reconfiguring all the routing config. Unfortunately I don't want that much chrurn right now in order to fix this, so I've used a different approach which is perfectly fine from a functional perspective, but don't use the same approach as zones _yet_.

### DIFF
--- a/packages/kuma-gui/src/app/meshes/views/MeshRootView.vue
+++ b/packages/kuma-gui/src/app/meshes/views/MeshRootView.vue
@@ -4,7 +4,7 @@
     v-slot="{ t }"
   >
     <AppView
-      :breadcrumbs="[
+      :breadcrumbs="['mesh-list-view'].includes(String(router.currentRoute.value.name ?? '')) ? [] : [
         {
           to: {
             name: 'mesh-list-view',
@@ -17,3 +17,7 @@
     </AppView>
   </RouteView>
 </template>
+<script lang="ts" setup>
+import { useRouter } from 'vue-router'
+const router = useRouter()
+</script>


johncowen:
~(maybe there is a test fail here 🤔 👀 )~ tests are fine

schogges (commented):
Is there a reason why we can't just do it the same way as in zones where we add the breadcrumbs in `ZoneDetailTabsView`?

johncowen:
I think longer term I guessing no, no reason at all, but I've not looked at that as yet so I could be wrong.

But, I do know that that will add a lot more churn than the code I have here (and potentially time + potentially bugs if I don't do that correctly (see "potentially time" for the likelihood of bugs being introduced)), something I was hoping to avoid short term.

> Note: This isn't ideally how I'd like to do it, but taking a quick look at zones (where the breadcrumbs are correct), this should be possible by reconfiguring all the routing config. Unfortunately I don't want that much chrurn right now in order to fix this, so I've used a different approach which is perfectly fine from a functional perspective, but don't use the same approach as zones yet.

Also note, I don't have to do this just here plus there are other things that need amending that aren't as straightforwards.

---

So I guess TL;DR is: I totally get your concerns, but I'm making the explicit decision to prioritise less churn right now. In any other situation I would be asking the same as you.

(BTW: I'm actually super glad you asked!)

schogges:
> I think longer term I guessing no, no reason at all, but I've not looked at that as yet so I could be wrong.
> 
> But, I do know that that will add a lot more churn than the code I have here (and potentially time + potentially bugs if I don't do that correctly (see "potentially time" for the likelihood of bugs being introduced)), something I was hoping to avoid short term.
> 
> > Note: This isn't ideally how I'd like to do it, but taking a quick look at zones (where the breadcrumbs are correct), this should be possible by reconfiguring all the routing config. Unfortunately I don't want that much chrurn right now in order to fix this, so I've used a different approach which is perfectly fine from a functional perspective, but don't use the same approach as zones yet.
> 
> Also note, I don't have to do this just here plus there are other things that need amending that aren't as straightforwards.
> 
> So I guess TL;DR is: I totally get your concerns, but I'm making the explicit decision to prioritise less churn right now. In any other situation I would be asking the same as you.
> 
> (BTW: I'm actually super glad you asked!)

Alright, let's get this in then and iterate later 🙂

schogges (approved):

johncowen:
I made an issue also https://github.com/kumahq/kuma-gui/issues/4867

PR #4860, opened 2026-04-22
feat(dataplanes): show SPIFFE ID in tls section

### LINKED ISSUE #4684: Show spiffeID in dataplane overview
### Description

Follow-up of https://github.com/kumahq/kuma-gui/issues/4526

Once https://github.com/kumahq/kuma/issues/15751 landed we can show the spiffeID of a dataplane in the about section.

schogges:
The `spiffeId` is an optional entry on a dataplane `_layout` response if there is a matching `MeshIdentity`.

image

`spiffeId` is not part of `_overview` and is part of `_layout` because

> Reasons:
>  1. Already does server-side computation — the layout handler already uses meshContextBuilder which gives access to MeshIdentity resources needed to
  compute SPIFFE IDs
>  2. Already has identity concepts — KRI (Kuma Resource Identifier) is already there; SPIFFE ID is the mTLS-level identity, a natural companion
>  3. Avoids proto changes — _overview is proto-defined (DataplaneOverview = Dataplane + DataplaneInsight), adding spiffeID would require proto schema
  changes + regeneration. _layout uses plain Go/OpenAPI types, much simpler to extend
>  4. Conceptual fit — layout answers "what is this DP's networking identity and connectivity?" — SPIFFE ID is literally its cryptographic identity

Closes https://github.com/kumahq/kuma-gui/issues/4684

### DIFF
--- a/packages/kuma-gui/features/mesh/dataplanes/overview/Tls.feature
+++ b/packages/kuma-gui/features/mesh/dataplanes/overview/Tls.feature
@@ -37,12 +37,18 @@ Feature: mesh / dataplanes / overview / TLS
       body: |
         cluster.kri_msvc_default_east_kuma-demo_my-service_5050.ssl.certificate.spiffe://default.local-zone.mesh.local/ns/kuma-demo/sa/default.expiration_unix_time_seconds: 1765823243
       """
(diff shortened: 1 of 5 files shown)


johncowen (approved):
Looks good, appreciate that chat through on this one 👍

PR #4868, opened 2026-04-23
fix(meshes): use undefined over `[]` for breadcrumbs

johncowen:
Follow up of https://github.com/kumahq/kuma-gui/pull/4859.

I used `[]` for the "empty" conditional. Which actually means "don't show any breadcrumbs at all, not even any above this"

Both `[]` and `undefined` have the same behaviour in Kuma, but if you embed kuma elsewhere then using `[]` means any higher breadcrumbs no longer show.

### DIFF
--- a/packages/kuma-gui/src/app/meshes/views/MeshRootView.vue
+++ b/packages/kuma-gui/src/app/meshes/views/MeshRootView.vue
@@ -4,7 +4,7 @@
     v-slot="{ t }"
   >
     <AppView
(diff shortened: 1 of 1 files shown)


schogges (approved):

johncowen:
Things are queued, but in the meantime I've done this instead https://github.com/kumahq/kuma-gui/pull/4869

I'd rather close this one I think and go with https://github.com/kumahq/kuma-gui/pull/4869 instead, sorry for the double PRs!

johncowen:
This was superseded by https://github.com/kumahq/kuma-gui/pull/4869, closing this

PR #4888, opened 2026-04-28
chore: extracts (npm and GH) constraint linter into separate package

johncowen:
Extraction of our package.json and GH action constraints linter to a new separate `@kumahq` package.

Note: I got an LLM to do this, reined it in, and then heavily amended it following my own review, please take this into account when reviewing to make sure the original functionality remains as-is. I'm also giving this a secondary review myself.

I also tried to make the "nested workspace" functionality we have as an optional (not applied by default) feature (undocumented) seeing as this feature is probably very rarely required.


packages/eslint-package-constraints/src/index.ts
```diff
@@ -0,0 +1,73 @@
...
+type DependencyKey = 'dependencies' | 'devDependencies' | 'peerDependencies'
+
+type Options = {
+  workspaceRoot?: boolean
+  dependencyIgnorePatterns?: Partial<
+    Record<DependencyKey, Record<string, unknown>>
+  >
+  workspaceJSON?: Record<string, unknown>
+}
+export const npmWorkspaceJSON = () => JSON.parse(execSync('npm query :root').toString()).at(0)
```
schogges: nit: Could we rename it to i.e. `getNpmWorkspaceJSON`? I stumbled upon this when I went through and it would read better because `npmWorkspaceJSON` sounds/reads like it's a variable and not a function.

schogges (commented):
Just a nit, the rest LGTM 👍

packages/eslint-package-constraints/src/index.ts
johncowen: one sec, brb

johncowen: Done, lmk! Will need another ✅

schogges (approved):

packages/eslint-package-constraints/src/index.ts
schogges: Thanks 👍

Issue #4855, opened 2026-04-22
Support KRIs on policy routes


schogges:
### Description

In order to properly link from the generic resources listings we need support for KRIs in the policy routes (i.e. MeshCircuitBreaker). In the generic resources we completely rely on KRIs and for policy resources we currently can't use the information in the respective KRI.

schogges:
I'm going to create follow-up issue(s) for using KRI protocol strings instead of vue-router objects.

PR #4894, opened 2026-04-29
feat(policies): support KRIs

schogges:
Finalizes the support of KRIs in policy detail view and adds some tests.

> [!WARNING]
> This covers a lot of code that hasn't been touched for years and I'm putting a lot of trust in our tests. Please make sure to give a good review and click through yourself.

Part of https://github.com/kumahq/kuma-gui/issues/4855

### DIFF
--- a/packages/kuma-gui/features/application/Titles.feature
+++ b/packages/kuma-gui/features/application/Titles.feature
@@ -37,7 +37,7 @@ Feature: application / titles
       | /meshes/default/data-planes                                                                                           | Data plane proxies        |
       | /meshes/default/data-planes/kri_dp_default___data-plane-name_/overview                                                | data-plane-name           |
       | /meshes/default/policies/circuit-breakers                                                                             | Policies                  |
-      | /meshes/default/policies/circuit-breakers/kri_~circuit-breakers_default___program-0_/overview                         | program-0                 |
+      | /meshes/default/policies/circuit-breakers/kri_~circuitbreaker_default___program-0_/overview                           | program-0                 |
       | /meshes/default/resources/meshfaultinjections                                                                         | Resources                 |
       | /meshes/default/resources/meshfaultinjections/kri_mfi_default_adviser_kuma-system_pension-0-959cb35ab-xtzqf_/overview | pension-0-959cb35ab-xtzqf |
       | /meshes/default/workloads                                                                                             | Workloads                 |
--- a/packages/kuma-gui/features/application/Titles.feature
+++ b/packages/kuma-gui/features/application/Titles.feature
@@ -37,7 +37,7 @@ Feature: application / titles
       | /meshes/default/data-planes                                                                                           | Data plane proxies        |
       | /meshes/default/data-planes/kri_dp_default___data-plane-name_/overview                                                | data-plane-name           |
       | /meshes/default/policies/circuit-breakers                                                                             | Policies                  |
-      | /meshes/default/policies/circuit-breakers/program-0/overview                                                          | program-0                 |
+      | /meshes/default/policies/circuit-breakers/kri_~circuit-breakers_default___program-0_/overview                         | program-0                 |
       | /meshes/default/resources/meshfaultinjections                                                                         | Resources                 |
       | /meshes/default/resources/meshfaultinjections/kri_mfi_default_adviser_kuma-system_pension-0-959cb35ab-xtzqf_/overview | pension-0-959cb35ab-xtzqf |
       | /meshes/default/workloads                                                                                             | Workloads                 |
--- a/packages/kuma-gui/features/mesh/policies/Index.feature
+++ b/packages/kuma-gui/features/mesh/policies/Index.feature
@@ -118,7 +118,7 @@ Feature: mesh / policies / index
     Then the "$item:nth-child(1) td:nth-child(2)" element contains "fake-cb-1"
     When I click the "$action-group" element
     And I click the "$view" element
-    Then the URL contains "circuit-breakers/fake-cb-1/overview"
+    Then the URL contains "circuit-breakers/kri_~circuitbreaker_default___fake-cb-1_/overview"
     And the "$detail-view" element contains "fake-cb-1"
     When I click the "$breadcrumbs > .breadcrumbs-item-container:nth-child(3) > a" element
     Then the "$item" element exists 2 times
--- a/packages/kuma-gui/src/app/gateways/sources.ts
+++ b/packages/kuma-gui/src/app/gateways/sources.ts
@@ -1,11 +1,11 @@
 import createClient from 'openapi-fetch'
 
 import { MeshGateway } from './data'
-import { useFetch } from '../kuma'
+import { useDataSource } from '../kuma'
 import type { DataSourceResponse } from '@/app/application'
 import { defineSources } from '@/app/application'
 import type KumaApi from '@/app/kuma/services/kuma-api/KumaApi'
-import type { KumaResourceTypeDescriptorCollection } from '@/app/resources/data'
+import type { ResourceTypeDescriptorCollection } from '@/app/resources/data'
 import { Rule } from '@/app/rules/data'
 import type { PaginatedApiListResponse as CollectionResponse } from '@/types/api.d'
 import type { paths } from '@kumahq/kuma-http-api'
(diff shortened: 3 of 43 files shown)


packages/kuma-gui/src/types/index.d.ts
```diff
@@ -390,6 +390,7 @@ export interface Meta<Type extends string = string> {
   type: Type
   mesh: string
   name: string
+  kri?: string
```
schogges, line 394: This is a little confusing. We are sharing this type with mocks `satisfies` in `kuma-http-api` and in the app. The weird thing about it is that I have to set this to optional here, because we need to support both legacy/old and new policies. In the data-layer we want to enforce the `kri` to be set, so we can rely on it in the view, but in the mocks we don't want to set it for legacy/old policies (just like the real API behaves).

packages/kuma-gui/src/app/policies/sources.ts
```diff
@@ -50,45 +50,66 @@ export const sources = (api: KumaApi) => {
...
-          path: {
-            mesh,
-            name,
+    '/policy-path/:path/policy/:kri/as/kubernetes': async (params) => {
+      const { path, kri } = params
+      const { shortName, mesh, name } = Kri.fromString(kri)
+
+      let response
+
+      if(shortName.startsWith('~')) {
```
schogges, line 86: This is the forking of wether we want to fetch a legacy/old policy that can't resolve a KRI or a new one. KRIs that are generated in the data-layer use a prefix `~` on the `shortName`.

schogges:
Similar to hostname-generators I'll do the linking via kri protocol separately. This is only doing the routing and fetching via KRI.

packages/kuma-gui/features/application/Titles.feature
```diff
@@ -37,7 +37,7 @@ Feature: application / titles
       | /meshes/default/data-planes                                                                                           | Data plane proxies        |
       | /meshes/default/data-planes/data-plane-name/overview                                                                  | data-plane-name           |
       | /meshes/default/policies/circuit-breakers                                                                             | Policies                  |
-      | /meshes/default/policies/circuit-breakers/program-0/overview                                                          | program-0                 |
+      | /meshes/default/policies/circuit-breakers/kri_~circuit-breakers_default___program-0_/overview                         | program-0                 |
```
johncowen: I saw that our GUI shortnames are the `type.toLowerCase()` wouldn't this mean that the CircuitBreaker shortname would be `~circuitbreaker`?

Maybe I misunderstood, but gonna fire this comment off while I keep looking at the rest of the code

packages/kuma-gui/features/mesh/policies/Index.feature
```diff
@@ -118,7 +118,7 @@ Feature: mesh / policies / index
     Then the "$item:nth-child(1) td:nth-child(2)" element contains "fake-cb-1"
     When I click the "$action-group" element
     And I click the "$view" element
-    Then the URL contains "circuit-breakers/fake-cb-1/overview"
+    Then the URL contains "circuit-breakers/kri_~circuitbreaker_default___fake-cb-1_/overview"
```
johncowen, line 121: Here we go this KRI shortName looks correct

packages/kuma-gui/src/app/rules/data/Origin.ts
```diff
@@ -20,6 +21,7 @@ export const Origin = {
       mesh,
       name,
       labels,
+      kri: 'kri' in origin ? String(origin.kri) : Kri.toString({ shortName: `~${type.toLowerCase()}`, mesh, name }),
```
johncowen: Self-note: double check this

packages/kuma-gui/src/types/index.d.ts
johncowen: Do you know specifically where we use this in the mocks? Seeing as these are our handwritten types (not OpenAPI ones) its almost pointless using them in the mocks. We could maybe remove/replace the usage in the mocks and possibly make this easier to work with.

packages/kuma-gui/src/app/rules/data/Origin.ts
johncowen: K after double checking, I'm not sure on a few things here:

- It looks like the type here can be things like a `Dataplane` which wouldn't require a `~`
- Do we even use these in URLs anywhere. Guessing this might be just "lets make sure everything has a KRI" and maybe it not addeed to a URL anywhere? lmk!

johncowen (commented):
Looks good! I left a couple of tiny notes, and possibly an important (or not at all) question

packages/kuma-gui/features/application/Titles.feature
schogges: Oh yes, you are right. I was thinking of the `paths` in `_resources`

packages/kuma-gui/src/app/rules/data/Origin.ts
schogges: Need to check on this.

schogges: I've checked again and `origin.kri` would be used for `policy-detail-view`s and `data-plane-policy-summary-view`. The latter is part of the legacy dataplanes policies rule list.

johncowen: I'm just trying to figure this out, maybe I'm wrong but does the `.origin` ever have a `.kri` from the backend?

johncowen: FYI: I'm just waiting on more info from backend here, this bit could be pretty consequential

packages/kuma-gui/src/types/index.d.ts
schogges: Yeah, definitely 👍 I'd like to try and remove these hand-written types and replace them with OAS. Potentially adding some overlays.

packages/kuma-gui/src/app/rules/data/Origin.ts
johncowen: https://github.com/kumahq/kuma/issues/17140

So this PR would probably be blocked on that ^ that I just made. Wondering if we can do something in the meantime so we can get this PR in 🤔

packages/kuma-gui/src/types/index.d.ts
johncowen: Ok cool, I did a quick search in the mocks for `types/` and found the ones using the handwritten types.

Lets replace these with OpenAPI ones in a follow up PR

packages/kuma-gui/src/app/rules/data/Origin.ts
johncowen: > sync: We are gonna make own KRI here in the meantime. So we'll need to fetch `_resources` and map the `Type` > `shortName` and then make the KRI with that. If we map and it comes back empty that means its an old style policy and needs the `~`

Once backend implement `origin` KRIs we can then remove this.

packages/kuma-gui/src/app/gateways/sources.ts
```diff
@@ -73,6 +75,8 @@ export const sources = (api: KumaApi) => {
     },
 
     '/meshes/:mesh/mesh-gateways/:name/rules': async (params) => {
+      const fetch = useFetch()
+      const resources = await fetch<KumaResourceTypeDescriptorCollection>('/resource-type-descriptors')
```
johncowen: We just figured that this should use our type rather than the backend type

packages/kuma-gui/src/app/policies/data/Policy.ts
```diff
@@ -26,7 +27,7 @@ export const Policy = {
 
     return {
       ...item,
-      kri: 'kri' in item ? item.kri ?? '' : '',
+      kri: 'kri' in item && typeof item.kri === 'string' ? item.kri : Kri.toString({ shortName: `~${item.type.toLowerCase()}`, mesh, zone, namespace, name: id }),
```
johncowen, line 30: Ah here (see https://github.com/kumahq/kuma-gui/pull/5105/changes#r3550839952)

🤔

packages/kuma-gui/src/app/rules/data/Origin.ts
schogges: Done via https://github.com/kumahq/kuma-gui/pull/4894/commits/4efca38a2f9a9e262ed1d6bb44f58212a2fe4e57

johncowen: I took a quick look, remember we said we'd not use the word `fetch` for the token/hook. I left an inline comment to explain a bit better.

schogges: The token/hook is already renamed to `dataSource` and `useDataSource`

packages/kuma-gui/src/app/rules/sources.ts
```diff
@@ -14,6 +16,8 @@ export const sources = (api: KumaApi) => {
   })
   return defineSources({
     '/meshes/:mesh/rules/for/:dataplane': async (params) => {
+      const fetch = useDataSource()
```
johncowen, line 19: Just leaving a note for future:

I'd like this to be:

```js
const data = useDataSource()
data.source('ssdf/dfg/fdg') <- EventTarget / EventSource API

//or

data.fetch('sdf/sdf/sdf') <-- promise/fetch aAPI
```

but we can do this in the future

johncowen (approved):

PR #4898, opened 2026-05-01
chore(x-tabs): patch KTabs

johncowen:
This one is a bit complicated, so I feel like a fairly good description will help, but I'm trying to be mindful of giving the whys/whats whilst keeping things short.

## Context for reviewers

We noticed a tiny issue with our HTML structure that could cause browser/accessibility issues.

We traced it back to https://github.com/Kong/kongponents/pull/2707 and I have a PR up with a clean revert of that that fixes this issue that is currently in review.

The conversation is quite long and its difficult not to get a bit lost in it - so I won't repeat myself here.

A recommendation was made:

> However, my recommendation is still that Kuma GUI should patch this behavior on its side, then remove that patch once XTabs is migrated to pass route targets through KTabs tab.to. 

I'm finding it difficult to get on a zoom call to discuss at the moment:

> In fact, I’ve already organized all the points I could think of in my replies above. Hopefully you’ll be able to find answers there to any questions that may come up.

So, I've gone ahead and drafted a PR that fixes the issue ourselves (it sounds like the reviewer feels its something we should fix ourselves). Hopefully this is some sort of path forwards if we cannot fix the issue for everyone.

## What this PR does.

1. Removes all the javascript, apart from one line (there is some property defining/slot defining that I don't really count as javascript)
2. Greatly simplifies all the code so we only have a `nav > ul > li` structure
3. Removes the attribute removal we were doing to correct the incorrect "aria semantics"
4. Adds `role="none"` to `li`s avoid overly word-y aria.
5. Copies the styles from KTabs, and removes duplicated and unnecessary rules, removes small SCSS mixins and genrally reorganises the structure into clearer CSS removing.
6. Adds a variant="small" which we don't use yet, but I copied this from https://github.com/Kong/kongponents/pull/3178 with help from @portikM

I might make a couple of inline comments.

## What this PR achieves (apart from unblocking us)

1. No javascript, less potential for bugs.
2. No javascript, so nothing super valuable to test (no tests written in potentially buggy javascript either!)
3. Correct ARIA semantics, whilst this is called XTabs this describes how it looks. Behaviourally its a horizontal navigation bar (ARIA is about behaviour not looks, its mainly used by partially impaired users plus others with and without impairments). We _never_ use without using links to update the URL (everyone wants bookmarkable/refreshable pages without jumping you back to the start)
4. Cleaner CSS - I totally understand that this has just been added and added over time and has gradually got a little out of shape. This PR cleans it up so its easier to follow and understand what is going on. To be fair I think @portikM has also done a lot of this work in his above linked PR.
5. The space bar now correctly scrolls the page when a tab is focussed instead of activating the link. Enter activates the link (via code in the browser we didn't write)
6. "Modifier-key clicking" now works.
7. No more incorrect HTML nesting anchors inside of anchors.
8. XTabs is now never responsible for linking behaviour, the engineer is. This approach trusts the engineer to do the right thing. i.e. "we make it easy to do the right thing and hard to do the wrong thing" (one of our shared/guiding themes). If an engineer doesn't use links, it just doesn't work - easy. If the engineer wants to add interactive content to a tab they can wrap it in a further component (we don't need this functionality but I've seen it in plenty of places elsewhere) - easy. It's now a lot harder to get this wrong.
9. There is not one single engineer/component responsible for everything anymore, aria functionality, bugs, styling, linking etc etc etc. The styling is built with HTML/CSS and the linking isn't the responsibility of the component - because in my mind linking isn't the problem to solve here. Linking is "easy", and we can trust engineers to do that correctly.
10. We don't have to re-implement or otherwise work around things here, we just use the browser.

Currently in Draft as I want some sort of informal direction on this before we approve and merge (this may never happen). But I would appreciate reviews from the usual reviewers, and I would also love it if @portikM could take a quick look at the preview builds and code, I've tried my best to stay true to the original here, I hope you can see that 👍


johncowen:
Ah perfect, our tests failed 🎉 

Let me take a quick look 👀 . (any reviewers can disregard the failing tests for the moment)

johncowen:
Ok I've fixed the tests - I didn't realise that I broke the `data-testid` so I put it back to how it was previous to this PR.

If we want to change this we can circle back and change it later. FYI: I would also like to change how the `disabled` things works: We don't use `disabled` ourselves, but since a "disabled hyperlink" is a "piece of text" i.e. a `span` (not an anchor with a disabled attribute and class) , the styles should reflect that. Actually let me add that to the list of things this PR does as it allows us to use potentially use `span`s inside of the tabs if/when we need to.

All in all, we can add/improve these things later.

Following https://github.com/Kong/kongponents/pull/3171#issuecomment-4384903190 I'm gonna take this PR out of Draft ready for approval and merge.

But! I'm still open to hearing any other realistic alternatives if anyone thinks we shouldn't do this. FWIW I think we should do this.

schogges (dismissed):
LGTM

johncowen:
I added a commit here that moved everything to use native CSS props (and `--x` prefixes), plus I noticed that the focus-ring design of KTabs should really be used on the entire app, not just KTabs

johncowen:
@schogges not sure if you've noticed but I made a few changes here (see the last non-merge commit) and it'll need another ✅

schogges (approved):
LGTM

PR #4900, opened 2026-05-04
feat(legacy-data-planes/detail-view): conditionally show SPIFFE ID

### LINKED ISSUE #4684: Show spiffeID in dataplane overview

schogges:
In https://github.com/kumahq/kuma-gui/pull/4860 we added the SPIFFE ID to the about section. This also adds it for legacy dataplanes. Previously I thought we can't do that because we could only use the `_layout` endpoint for "new" dataplanes, but this is not the case.

Closes https://github.com/kumahq/kuma-gui/issues/4684


johncowen (approved):
LGTM 👍

PR #4901, opened 2026-05-04
feat(dataplanes): consider listeners when computing state

### LINKED ISSUE #4749: New DPPs with `listeners` are displayed as `Offline` in GUI
### Description

By looking at the `/_overview` response:
<details>
  <summary>DataplaneOverview</summary>

```json
{
 "total": 1,
 "items": [
  {
   "type": "DataplaneOverview",
   "mesh": "default",
   "name": "kuma-default-ingress-787987f8f9-84n6k-56c6x8cb85v2w864",
   "creationTime": "2026-03-23T12:17:03.310495334Z",
   "modificationTime": "2026-03-23T13:35:20.896706915Z",
   "labels": {
    "app": "kuma-default-ingress",
    "app.kubernetes.io/instance": "kuma",
    "app.kubernetes.io/managed-by": "Helm",
    "app.kubernetes.io/name": "kuma",
    "app.kubernetes.io/version": "0.0.0-preview.vlocal-build",
    "helm.sh/chart": "kuma-0.0.0-preview.vlocal-build",
    "k8s.kuma.io/namespace": "kuma-system",
    "k8s.kuma.io/service-account": "kuma-default-ingress",
    "kuma.io/display-name": "kuma-default-ingress-787987f8f9-84n6k",
    "kuma.io/env": "kubernetes",
    "kuma.io/mesh": "default",
    "kuma.io/origin": "zone",
    "kuma.io/proxy-type": "sidecar",
    "kuma.io/sidecar-injection": "enabled",
    "kuma.io/workload": "kuma-default-ingress",
    "kuma.io/zone": "zone-1",
    "pod-template-hash": "787987f8f9"
   },
   "dataplane": {
    "networking": {
     "address": "10.42.0.44",
     "transparentProxying": {
      "redirectPortInbound": 15006,
      "redirectPortOutbound": 15001,
      "ipFamilyMode": "DualStack",
      "reachableBackends": {}
     },
     "admin": {
      "port": 9901
     },
     "listeners": [
      {
       "type": "ZoneIngress",
       "address": "10.42.0.44",
       "port": 10001,
       "name": "10001"
      }
     ]
    }
   },
   "dataplaneInsight": {
    "subscriptions": [
     {
      "id": "dd81d89d-81bf-4964-b966-5e01c54c9ad5",
      "controlPlaneInstanceId": "kuma-control-plane-57bb5d44bd-xwccb-15f9",
      "connectTime": "2026-03-23T12:17:15.035018731Z",
      "status": {
       "lastUpdateTime": "2026-03-23T12:17:15.208016421Z",
       "total": {
        "responsesSent": "2",
        "responsesAcknowledged": "2"
       },
       "cds": {
        "responsesSent": "1",
        "responsesAcknowledged": "1"
       },
       "eds": {},
       "lds": {
        "responsesSent": "1",
        "responsesAcknowledged": "1"
       },
       "rds": {}
      },
      "version": {
       "kumaDp": {
        "version": "0.0.0-preview.vlocal-build",
        "gitTag": "local-build",
        "gitCommit": "local-build",
        "buildDate": "local-build",
        "kumaCpCompatible": true
       },
       "envoy": {
        "version": "1.37.0",
        "build": "6d9bb7d9a85d616b220d1f8fe67b61f82bbdb8d3/1.37.0/Modified/RELEASE/BoringSSL",
        "kumaDpCompatible": true
       }
      }
     }
    ],
    "metadata": {
     "dataplane.admin.address": "127.0.0.1",
     "dataplane.admin.port": "9901",
     "dataplane.appProbeProxy.enabled": "true",
     "dataplane.dns.port": "15054",
     "dataplane.proxyType": "dataplane",
     "dataplane.readinessReporter.port": "9902",
     "features": [
      "feature-embedded-dns",
      "feature-readiness-unix-socket",
      "feature-strict-inbound-ports",
      "feature-tcp-accesslog-via-named-pipe"
     ],
     "ipv6Enabled": true,
     "metricsCertPath": "",
     "metricsKeyPath": "",
     "spireAgent.socketPath": "",
     "systemCaPath": "/etc/ssl/certs/ca-certificates.crt",
     "version": {
      "envoy": {
       "build": "6d9bb7d9a85d616b220d1f8fe67b61f82bbdb8d3/1.37.0/Modified/RELEASE/BoringSSL",
       "kumaDpCompatible": true,
       "version": "1.37.0"
      },
      "kumaDp": {
       "buildDate": "local-build",
       "gitCommit": "local-build",
       "gitTag": "local-build",
       "version": "0.0.0-preview.vlocal-build"
      }
     },
     "workdir": "/tmp/kuma-dp-1734515325"
    }
   }
  }
 ],
 "next": null
}
```
</details>

I'd say the proxy is "Online" since the last subscription doesn't have `disconnectTime` set. However, GUI shows it as "Offline" 

Image

Note: currently there is no easy way to reproduce exactly this scenario (the feature is not fully merged), I'm just curious to know how "Online/Offline" statuses are computed.

schogges:
When computing the state of a DP we also have to take into account the new `dataplane.networking.listeners` which are being used when the DP is a zone proxy. In this case the `dataplane.networking.inbound` entry is absent.

I've extended the mocks and added a separate test case to cover the new `dataplane.networking.listeners`.

---

Most importantly this adds a new status a DPP can have and changes our interpretation of a DPP being online or offline. Some time ago there was never a case in which a DPP would have no inbounds. This has changed and it's valid case that a DPP has no inbounds and now also no listeners. Still, the DPP would be seen as healthy and therefore it's online. It's only offline if there are inbounds and/or listeners that are all in non-healthy state.
The new status is `Not connected to CP`, which is the first check we have to make. If it's not connected to the CP all the other data we have might be stale and therefore we can't reliably tell wether it's online or offline. Therefore a DP that is online/offline is always connected to the CP.
Some more info about the history around DPP status: https://github.com/kumahq/kuma-gui/pull/4901#discussion_r3267981100

Closes https://github.com/kumahq/kuma-gui/issues/4749


packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -49,13 +49,20 @@ export const DataplaneOverview = {
...
             return states.offline
           case unhealthyInbounds.length > 0:
             // Some inbounds being unhealthy means the Dataplane is partially degraded.
             return states.partiallyDegraded
+          case networking.listeners.length > 0 && unhealthyListeners.length === networking.listeners.length:
+            // All listeners being unhealthy means the Dataplane is offline.
+            return states.offline
+          case unhealthyListeners.length > 0:
+            // Some listeners being unhealthy means the Dataplane is partially degraded.
+            return states.partiallyDegraded
```
johncowen, line 76: We were wondering in the case of:

- There are both inbounds and listeners (is this possible?)
- All of the inbounds are healthy
- All of the listeners are unhealthy

Does this mean the dataplane is offline or partially degraded? 

If I'm reading the code correctly, currently this would give us `offline` which might not be correct if there are healthy inbounds (i.e. only "degraded")

johncowen: > There are both inbounds and listeners (is this possible?)

Looks like these can co-exist

https://github.com/kumahq/kuma/blob/master/docs/madr/decisions/095-mesh-scoped-zone-ingress-egress.md#example

schogges: We agreed to treat both `inbounds` and `listeners` as conceptually the same, meaning that we leave the switch statement as is but merge the `listeners` to the `inbounds` to do the computation. Furthermore we also want to escape early in case there is no connection to the CP. In that case the wording should be `Not connected to CP`.

schogges: I've change the wording from `Not connected to CP` to `Disconnected from CP`

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -43,21 +44,25 @@ export const DataplaneOverview = {
...
 
-        const unhealthyInbounds = networking.inbounds.filter((inbound) => inbound.state !== 'Ready')
+        const networkEndpoints = [...networking.inbounds, ...networking.listeners]
+        const unhealthy = networkEndpoints.filter((endpoint) => endpoint.state !== 'Ready')
         switch (true) {
-          case unhealthyInbounds.length === networking.inbounds.length:
-            // All inbounds being unhealthy means the Dataplane is offline.
+          case unhealthy.length === 0:
+            // All network endpoints being healthy means the Dataplane’s status is online.
+            return states.online
```
johncowen, line 70: Q: If the initial state is `online` how come we added this new case?

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
johncowen: Biggest nit ever 😅 : Does `Disconnected` imply that it was once connected?

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_overview.ts
```diff
@@ -105,7 +128,7 @@ export default ({ fake, pager, env }: Dependencies): ResponseHandler => (req) =>
                       zone: zoneQuery || (isMultizone && fake.datatype.boolean()) ? zone : undefined,
                     }),
                     ...(fake.datatype.boolean() ? {
-                      state: fake.kuma.inboundState(),
+                      state: fake.kuma.state(),
```
johncowen, line 131: Wondering how many other places we use `inboundState`, and in my mind following up from that I would ask how useful it is to have this method, and whether we should deprecate it or not.

packages/kuma-http-api/mocks/FakeKuma.ts
```diff
@@ -156,6 +156,13 @@ gbXR5RnEs0hDxugaIknJMKk1b0g=
     return k8s ? this.k8s.dataplaneSuffix() : ''
   }
 
+  state() {
+    return this.faker.helpers.arrayElement(['Ready', 'NotReady', 'Ignored'])
+  }
+
+  /**
+   * @deprecated use `fake.kuma.state()` instead
+   */
```
johncowen: Oh! 😆 👏 👏 👏 

I did a quick search and in `main` only found 2 places where we use this, one of which you've removed in this PR.

If I'm not mistaken (and we don't use this elsewhere), I think we should just change the remaining instance and remove the method entirely instead of deprecating. WDYT?

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -43,21 +44,25 @@ export const DataplaneOverview = {
       dataplaneInsight,
       dataplaneType: networking.type === 'gateway' ? dpTypes.builtin : typeof networking.gateway !== 'undefined' ? dpTypes.delegated : dpTypes.standard,
       status: (() => {
-        const state = typeof dataplaneInsight.connectedSubscription !== 'undefined' ? states.online : states.offline
-        if (networking.gateway) {
+        const state = typeof dataplaneInsight.connectedSubscription !== 'undefined' ? states.online : states.disconnectedCp
+        if (networking.gateway || state === states.disconnectedCp) {
```
johncowen, line 61: Just wanted to leave a note to remind me that we probably have two states here:

- Connectedness
- Healthiness

At the moment in this PR we mix the two into "status" but there is an argument to say that these states could be treated separately i.e. I could be "disconnected" and "healthy" or "connected" and "unhealthy" or "disconnected" and "unhealthy" etc etc. They all means slightly different things. Whether this tiny nuance is actually important or not is another thing.

No action here, but wanted to leave a note to remind myself.

johncowen (commented):
I left a couple of questions, but I think this is pretty much good to go! 👍 There is one other thing here, but let me think about that for a little while, I'm not sure how important it is.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
schogges: I think you are right, it implies there was a connection before. I've reworded it to what we agreed on before.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
schogges: It's more explicit than adding to the (current) second case `unhealthy.length > 0 && ...`

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_overview.ts
schogges: I've changed the other case where `inboundState` was used and removed it entirely. This means that `fake.kuma.state()` now entirely replaces `fake.kuma.inboundState()`

schogges: > how useful it is to have this method, and whether we should deprecate it or not.

I think it's quite useful because it centralizes these special terms in one place. Otherwise if another state will be added in the future, we'd have to adjust all places.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
johncowen: Thing is, if its not connect to the CP, then we don't know whether its healthy or unhealthy. The healthy state would be what it was the last time it was connected to the CP.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
johncowen: Ah! I wasn't suggesting we add it I guess, but we let it just fall through to the default like we had before. i.e. the initial state is online, and we don't change it unless its not online.

I guess its like "_unless_ the count of unhealthy is the same as the total (offline), or unhealthy is non-zero (partially degraded), then nothing needs to change"

schogges: I'm not entirely sure, but I think it was always the case that a DPP had at least one inbound? Otherwise we wouldn't have the issue now that the DPP shows offline when there are listeners but no inbounds 🤔 therefore we need to explicitly check for the amount of inbounds is greater than 0. Either in a separate case or within the case in which we check if the amount of unhealthy is the same as the total amount.
I chose to be more explicit and have it in a separate case, because it's easier to read/understand.

schogges: Maybe it's now the case that a DPP either has at least one inbound or at least one listener. Anyways I would still argue that it's more bullet proof to explicitly check wether unhealthy is 0.

johncowen: It looks to me like this change changes the previous functionality. What do you think?

packages/kuma-gui/features/mesh/dataplanes/Index.feature
```diff
@@ -63,6 +64,53 @@ Feature: mesh / dataplanes / index
...
+      | Offline             | !!js/undefined              | [{ state: NotReady }]                  | !!js/undefined                         |
+      | Offline             | !!js/undefined              | !!js/undefined                         | [{ state: NotReady }]                  |
+      | Partially degraded  | !!js/undefined              | [{ state: Ready },{ state: NotReady }] | !!js/undefined                         |
+      | Partially degraded  | !!js/undefined              | !!js/undefined                         | [{ state: Ready },{ state: NotReady }] |
+      | Partially degraded  | !!js/undefined              | [{ state: Ready },{ state: NotReady }] | [{ state: Ready },{ state: NotReady }] |
+      | Not connected to CP | 2021-02-17T07:33:36.412683Z | [{ state: Ready }]                     | !!js/undefined                         |
+      | Not connected to CP | 2021-02-17T07:33:36.412683Z | !!js/undefined                         | [{ state: Ready }]                     |
+      | Not connected to CP | 2021-02-17T07:33:36.412683Z | [{ state: Ready },{ state: NotReady }] | [{ state: Ready },{ state: NotReady }] |
+
```
johncowen, line 113: This looks to me like its checking that the implementation is done correctly, wdyt?

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
```diff
@@ -41,6 +42,12 @@ export default ({ env, fake }: Dependencies): ResponseHandler => (req) => {
       case defaultType === 'delegated':
       case name.includes('-delegated'):
         return 'DELEGATED'
+      case name.includes('ingress') && name.includes('egress'):
+        return 'INGRESS-EGRESS'
+      case name.includes('ingress'):
+        return 'INGRESS'
+      case name.includes('egress'):
+        return 'EGRESS'
```
johncowen, line 60: The previous pattern used `-<keyword>` i.e. `-delegated`, I can kinda imagine what this is for but I can only vaguely remember. Bit of a nit, but is it worth using `-ingress` and `-egress`? I guess there was a reason originally for including the `-`

packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
```diff
@@ -27,7 +27,7 @@ data-planes:
     abnormal-traffic-stats: !!text/markdown |
       Abnormal traffic stats detected. Investigate the traffic patterns of individually marked inbounds and outbounds.
     dataplane-offline: !!text/markdown |
-      The dataplane is offline, so we either cannot retrieve any information or the information currently displayed may be out of date.
+      The dataplane is disconnected from the control plane or offline, so we either cannot retrieve any information or the information currently displayed may be out of date.
```
johncowen: Isn't half the reason why we added this extra state so we can tell the user whether the data is either "disconnected" or "offline" specifically? Also, I'm not totally sure but as I understand it, now we know at this point whether a dataplane is disconnected or offline means maybe we also know whether the out-of-date or not (I might be wrong, but if the DP is connected but unhealthy, I would guess that means the data is up-to-date?)

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
schogges: What do you mean?

packages/kuma-gui/features/mesh/dataplanes/Index.feature
schogges: I've deleted most of the cases and left only one case per status. The actual testing of implementation now happens in a unit test.

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
schogges: Not sure why it's necessary to add a `-`. During my testing it wasn't necessary. Probably to avoid any issues when `ingress` or `egress` are part of a generated word?

packages/kuma-gui/src/app/data-planes/locales/en-us/index.yaml
schogges: From how I understood, if the DP is not connected we can't reliably tell if the DP is online or offline. This sentence covers both cases. I'll have a look how I can reword this.

schogges: I've split it up in two separate messages

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
johncowen: It looks to me like previously:

```
inbounds=0
unhealthy=0 (there are no inbounds so unhealthy will always be 0)
```

then the state was `offline`

Now if:

```
inbounds=0
listeners=0
unhealthy=0 (there are no inbounds/listeners so unhealthy will always be 0)
```

then the state is `online`

Is that right?

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: > Not sure why it's necessary to add a `-`

Me neither, but I would guess one of the things could be what you mentioned. I guess because its like that already in previous instances, I'd say its safer and no trouble for us to just do the same as previous (unless we can think of a reason not to)

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
schogges: Yes, that's how I understood it should be 🤔 but now that you are asking I'm not entirely sure anymore 😅

schogges: Ok, I've reverted it back to the logic as it was before. If both `listeners` and `inbound` are empty the dataplane is offline.

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
schogges: Ok, I've splitted up `type` into `type` and `zoneProxyType` because they are separate. It might cause invalid types of having a gateway with listeners, but this is better than how it actually was before with having only gateways with listeners. I've also added the `-` prefix.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
schogges: Actually my change was correct and I have to change it back to how it was before. I've chatted with backend and we should see the dataplane as online if there are neither listeners nor inbounds. It's like that:
The dataplane is online if it is connected to CP and if it's not proven to be offline or partially degraded by any listener/inbound that is not ready.

johncowen (commented):
I think this looks almost there, as we weren't clear on the `inbound` thing here and we have more info now I want to make a comment on that, but I also found some more history in the GUI repo that might help this morning, but I've lost it now. Let me go and find it and I can make the comment, give this a final once over and we should be good to go 👍

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
johncowen: TL;DR:

- Q: Is it valid for a Dataplane to have no inbound/listeners?
A: Previously, no, currently yes.
- Q: If it is valid, what state should we show for a "connected" Dataplane with no inbound/listeners?
A: `online`

---

The following is mostly for my own (and future me's) benefit
Ok there was quite a bit of back and forth here and I wanted to try and understand the history the present and what and why we are doing what we are doing now.

I think there are two questions here (both relevant to historically and now):

- Is it valid for a Dataplane to have no inbound/listeners?
- If it is valid, what state should we show for a "connected" Dataplane with no inbound/listeners?

This whole thing rang bells with me and I'm sure I've worked through this with backend folks myself at some point and confirmed that what we had was the intended behaviour.

I don't think it is intended behaviour now, but I was trying to find any sort of documentation around that so I could try and understand why this is how it is to be sure.

This morning after quite a bit of sleuthing I found https://github.com/kumahq/kuma-gui/pull/1835 which says:

> Removes the inbounds.length === 0 check for the Dataplanes status utility. It’s not needed because non-gateway Dataplanes always have an inbound.

And this is the only doc I could find regarding our questions unfortunately.

GitHub doesn't seem to be working for me at all, but before I continue let post some links. I managed to find the original deletion of the line we added back in in the squashed commit.https://github.com/kumahq/kuma-gui/commit/ec0bb869c7dcd3fdf9241215935c7dfccecfaac2#diff-6cd4b378952d5211d00c3e499e5d33d9d1c1e40cfcc5e470b4940a5d97eaa8a1L35

Screenshot 2026-05-19 at 17 12 21

with the removal of that case, if `unhealthyInbounds.length` is `0` _and_ `inbounds.length` is `0` then they are equal and the status therefore ends up being 'offline'.

Important note: inbounds.length being zero here meant `online` before this was deleted

Back to our questions:

> - Is it valid for a Dataplane to have no inbound/listeners?

Back when the above PR was merged apparently "non-gateway Dataplanes always have an inbound", but we don't know where that information came from. It also feels to me from the statement that the removal of the inbounds.length check _could_ have been a mistake, and neither the author nor the reviewer spotted that `0` would equal `0` and perhaps unintentionally a Dataplane with no inbound would be `offline`. I think this is inconsequential due to the below (_previously_ all non-gateways had an inbound)

Unfortunately we'll never know, and its not that important anymore anyway 🤷 but seeing as I spent a while trying to figure this out I'd rather all this info was together _somewhere_.

---

Separately we've asked the backend about whether it is valid for a Dataplane to have no inbound/listeners.

As a kubernetes `Job` doesn't have any inbounds or listeners then it _is_ valid. But!

Apparently at some point in time there was a workaround that assigned an inbound even to a kubernetes `Job` in order to then be able to assign `tags` to the inbound. I guess this was until fairly recently as part of the reason to add labels would have been (I guess) to be able to stop doing this.

Most importantly, there is **nothing _now_ that prevents us from running a DPP that has no listeners or inbounds**.

So

> - Is it valid for a Dataplane to have no inbound/listeners?

Previously, no, currently yes.

---

> - If it is valid, what state should we show for a "connected" Dataplane with no inbound/listeners?

The fact that you can have a kubernetes `Job` than has no inbounds, and we can't tell whether a DPP corresponds to that or not means we have to say DPPs with no inbound is healthy. It is not about whether services can contact the DPP or not (i.e. no inbounds no inbound traffic?)

@schogges sorry this is a bit of a brain dump! You mentioned you might want to add some of this to the PR, so let me know if you need a little more clarity.

In the meantime I'm going to take a final look at this PR, but I think the change you made here is the right thing to do 👍 

Sorry for the delay here, I've been so busy today I started this morning looking for this and I've only just got around to getting back here to write down what I found.

schogges: Thanks @johncowen for digging through history and shedding some light on this 👍 I'll add some info to the PR description

packages/kuma-gui/features/mesh/legacy-dataplanes/Index.feature
johncowen, line 1: Did we delete this file in another PR and its found its way back in here?

schogges: Thanks for catching it 👍 this came back via rebase 😄

packages/kuma-gui/src/app/data-planes/data/index.spec.ts
```diff
@@ -410,4 +412,109 @@ describe('DataplaneOverview', () => {
...
+          dataplaneInsight: {
+            subscriptions,
+          },
+        }
+        const actual = await fixture.setup((item) => {
+          item.type = expected.type
+          item.name = expected.name
+          item.mesh = expected.mesh
+          item.dataplane.networking.inbound = expected.networking.inbounds as unknown as DataplaneInbound[]
+          item.dataplane.networking.listeners = expected.networking.listeners as unknown as DataplaneListener[]
```
johncowen: Could you explain the issue here? (in relation to the casting)

johncowen: It also looks to me like this is casting to the wrong type

packages/kuma-gui/src/app/data-planes/data/index.spec.ts
```diff
@@ -2,7 +2,9 @@ import { fs } from '@kumahq/kuma-http-api/mocks'
 import { describe, expect, test as _test } from 'vitest'
 
 import { Dataplane, DataplaneOverview } from './'
+import type { DiscoverySubscription } from '@/app/subscriptions/data'
 import { plugin, server } from '@/test-support/data'
+import type { DataplaneInbound, DataplaneListener } from '@/types'
```
johncowen: I'm almost sure we shouldn't be using these types at this point, we should be using the ones from `./data` wherever possible

packages/kuma-gui/src/types/index.d.ts
```diff
@@ -306,6 +311,7 @@ export type DataplaneNetworking = {
   advertisedAddress?: string
   inbound?: DataplaneInbound[]
   outbound?: DataplaneOutbound[]
+  listeners?: DataplaneListener[]
```
johncowen, line 314: I don't think any of these additions are necessary. I'm fairly sure this export can be deleted and linting is fine. Unless this is used elsewhere (which I doubt), I think we should delete this export here instead of adding to it.

If you could double check this thinking for me tho that would be great 👍

packages/kuma-gui/src/app/data-planes/data/index.spec.ts
schogges: There are no types based on OAS in `DataplaneOverview` yet. I followed to current pattern and added the types next to the others. I needed to type cast here because I just added the bare minimum of data to the fixture.

packages/kuma-gui/src/app/data-planes/data/index.spec.ts
schogges: I've added an answer above ☝️ all of this here is related to above

packages/kuma-gui/src/types/index.d.ts
schogges: I've added an answer above ☝️ all of this here is related to above

packages/kuma-gui/src/app/data-planes/data/index.spec.ts
johncowen: Ah! Sorry I followed back `Dataplane` instead of `DataplaneOverview`, sorry! I guess this is another reason to split this `index.spec` file into several files (`{Dataplane,DataplaneOverview}.spec` etc) instead of one so its clearer. There's something else I spotted here also which I might put a PR up for in a sec.

---

I haven't been in these files for quite some time so I can't fully remember the details of this little test harness but I do remember that the intention here was to make it so these things were done with a non-destructive deep merge rather than an overwrite - the same as our how our gherkin tests work. Granted its not as straight-forwards as with our gherkin tests here (thats maybe something we can improve) and maybe there are times where we can't do this with that is there currently 🤷 , but if we can I'd rather do that. Another thing I can vaguely remember was that another intention was to make it so you didn't have to import lots of other types just to get some tests written.

Anyway, I re-wrote the test using this approach (deep merge rather than overwrite) without having to cast anything or import more types here https://github.com/johncowen/kuma-gui/commit/2eafd9d00bd562f50c6fd439960d1b90287b6966 . I don't think I missed anything, but let me know if you want me to push this commit onto your branch or do it some other way. Also happy if you wanted to just apply something similar there and push that.

If it's too much hassle I'm also good to leave as is, but I figured it was a good opportunity to share some background also. Lemme know either way.

schogges: There were still some issues (i.e. type `string` doesn't match with union `'Ready' | 'NotReady' | ...`) in your commit so I just pushed something myself.

johncowen: Ok no problem 👍 , could you explain to me what the issues were in more detail? I'm interested to understand those, ty!

schogges: For example:

```ts
{
  inbounds: [{ state: 'Ready' }],
  listeners: [{ state: 'Ready' }],
  subscriptions: [{ connectTime: '2021-02-19T10:00:00Z', disconnectTime: '2021-02-19T11:00:00Z' }],
  status: 'disconnected_cp',
},
```

causes `Type 'string' is not assignable to type '"Ready" | "NotReady" | "Ignored" | undefined'.`. Therefore need to add `as const` to each `state`.

johncowen: Are you seeing this sort of error when you add a different `state`?:

Screenshot 2026-05-21 at 12 01 43


Or maybe I'm still not following still? lmk

johncowen: Oh I just took a closer look at what you did here, I think you missed the `const` I added in this commit (https://github.com/johncowen/kuma-gui/commit/2eafd9d00bd562f50c6fd439960d1b90287b6966)

Screenshot 2026-05-21 at 12 33 43

If you add this single const at the end, does that solve your issue? (if so I think we can resolve this convo and get this PR approved and merged)

schogges: Done

johncowen (approved):
Lets get this in!

PR #4905, opened 2026-05-05
chore(ci): fixup vite/connect based mocking

johncowen:
Tweaks our vite based mock runner so we can use vite (which uses node connect http server) to serve our mocks.

Please use the below cookie config for the moment (my vite server is currently running on `:8081`)

Screenshot 2026-05-05 at 16 49 10

Running our mocks in vite instead of MSW will soon become the default. MSW will only be used for our preview sites.

I probably will have a little more to write here

---

Wanted to quote this issue from a while back here:

> Whilst it doesn't have to be done here, if you have this running we may aswell not use MSW during local development work and only use it for the PR previews.

https://github.com/kumahq/kuma-gui/issues/2001

### DIFF
--- a/packages/config/src/vite/plugins/server.ts
+++ b/packages/config/src/vite/plugins/server.ts
@@ -29,19 +29,6 @@ const read = async (path: string) => {
   }
   return (await readFile(`${cwd}/${path}`)).toString()
 }
-const version = JSON.parse((await read('./package.json'))).version
-
-export const defaultKumaHtmlVars = {
-  baseGuiPath: '/gui',
-  apiUrl: 'http://localhost:5681',
-  version,
-  product: 'Kuma', // we no longer use this, it can be removed in the backend
-  mode: 'global',
-  environment: 'universal',
-  storeType: 'postgres',
-  apiReadOnly: false,
-}
-
 const exists = async (path: string) => {
   try {
     return (await stat(path)).isFile()
@@ -49,6 +36,30 @@ const exists = async (path: string) => {
     return false
   }
 }
+const Cookie = {
+  parse: (str: string, { prefix = '' } = { prefix: '' }) => {
+    return Object.fromEntries(str.split(';')
+      .map((item) => item.trim())
+      .filter((item) => item !== '')
+      .map((item) => {
+        const [key, ...value] = item.split('=')
+        return [key, value.join('=')] as [string, string]
+      })
+      .filter(([key, value]) => key.startsWith(prefix)))
+  },
+}
+
+export const defaultKumaHtmlVars = {
+  baseGuiPath: '/gui',
+  apiUrl: 'http://localhost:5681',
+  version: JSON.parse((await read('./package.json'))).version,
+  product: 'Kuma', // we no longer use this, it can be removed in the backend
+  mode: 'global',
+  environment: 'universal',
+  storeType: 'postgres',
+  apiReadOnly: false,
+}
+
 const interpolate = (template: string, vars: KumaHtmlVars) => {
   return template
     .replace('{{.BaseGuiPath}}', '/gui')
@@ -63,6 +74,7 @@ export const kumaIndexHtmlVars = (htmlVars: Partial<KumaHtmlVars> = {}): Plugin 
   }
 }
 
+
 const server = ({
   template = './index.html',
   vars = {},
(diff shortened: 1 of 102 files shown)


packages/kuma-http-api/mocks/src/_resources.ts
```diff
@@ -8,7 +8,9 @@ export default ({ env }: Dependencies): ResponseHandler => (_req) => {
   const resourceCount = parseInt(env('KUMA_RESOURCE_COUNT', `${resources.length}`))
 
   return {
-    headers: {},
+    headers: {
+      'Transfer-Encoding': 'chunked',
+    },
```
schogges, line 13: I would prefer to have it more random and not to specific mock handlers.

packages/config/src/vite/plugins/server.ts
```diff
@@ -63,26 +74,30 @@ export const kumaIndexHtmlVars = (htmlVars: Partial<KumaHtmlVars> = {}): Plugin
...
+    }
 
-      const cookies = (req.headers?.cookie ?? '').split(';')
-        .map((item) => item.trim())
-        .filter((item) => item !== '')
-        .reduce((prev, item) => {
-          const [key, value] = item.split('=')
-          prev[key] = value
-          return prev
-        }, {} as Record<string, string>)
```
johncowen, line 85: This was mainly replaced with the `Cookie` util I copy/pasta'ed from elsewhere

packages/config/src/vite/plugins/server.ts
```diff
@@ -63,26 +74,30 @@ export const kumaIndexHtmlVars = (htmlVars: Partial<KumaHtmlVars> = {}): Plugin
...
+    const htmlVars = {
+      ...defaultKumaHtmlVars,
+      apiUrl: `http://${address.address}:${address.port}`,
+      ...vars,
+    }
```
johncowen, line 95: Only when running a vite server, this should now use the port of where-ever the vite instance is running instead of the default `localhost:5681`

packages/fake-api/src/index.ts
```diff
@@ -30,6 +30,12 @@ export function escapeRoute(route: string): string {
   return route.replaceAll('+', '\\+')
 }
 
+export const addOrigin = <T>(fs: Record<string, T>, origin: string) => {
+  return Object.fromEntries(Object.entries(fs).map(([route, response]) => {
+    return [route.includes('://') ? route : `${origin}${route}`, response]
+  }))
+}
+
```
johncowen, line 38: I've been wanting to add this util in for a while - I just took it from elsewhere

johncowen (commented):
Some notes:

packages/fake-api/src/vite.ts
```diff
@@ -9,44 +9,59 @@ type PluginOptions<TDependencies extends object = {}> = {
...
+        })
+        const type = response.headers.get('Content-Type') ?? 'application/json'
+        res.setHeader('Content-Type', type)
+        res.setHeader('Status-Code', response.headers.get('Status-Code') ?? '200')
+        const resp = type.endsWith('/json') ? JSON.stringify((await response.json()), null, 4) : (await response.text())
+        if(response.headers.get('Transfer-Encoding') === 'chunked') {
+          res.write(resp)
+          res.end()
+        } else {
+          // res.removeHeader('Content-Length')
```
schogges: nit: is this still needed?

packages/kuma-http-api/mocks/src/zones/_.ts
```diff
@@ -1,15 +1,19 @@
...
 
-export default (_deps: Dependencies): ResponseHandler => (req) => {
+export default ({ fake }: Dependencies): ResponseHandler => (req) => {
   switch (req.method.toUpperCase()) {
     case 'DELETE':
       return {
-        headers: {},
+        headers: {
+          ...(fake.datatype.boolean() ? { 'Transfer-Encoding': 'chunked' } : {}),
+        },
```
schogges, line 9: `DELETE` response would probably never be streamed

packages/kuma-http-api/mocks/src/_resources.ts
schogges, line 13: Should this also be randomized?

johncowen: I wanted to keep at least one that was always chunked, we should also keep one that is always not chunked come to think of it, brb

packages/kuma-http-api/mocks/src/zones/_.ts
johncowen: Probably but who knows 🤷

packages/fake-api/src/vite.ts
johncowen: I removed it for the moment

packages/kuma-http-api/mocks/src/zones/_.ts
johncowen: I've left this in, I don't think it matters either way

packages/kuma-http-api/mocks/src/_resources.ts
johncowen: I made it so the mesh listing is never chunked, so we have one common one never chunked, one common one always chunked and all the rest are randomly chunked/not-chunked

johncowen: Resolving this one, we added random chunking to the majority of endpoints

schogges (approved):
LGTM 👍

PR #4911, opened 2026-05-07
fix: undo KSlideout z-index-ing

johncowen:
Undoes the z-indexing of KSlideout

Unfortunately KSlideout doesn't support specifying a z-index as `auto`, so we work around that with a bit of strange piece of TS, open to other suggestions to do the same.

### DIFF
--- a/packages/x/src/components/x-drawer/XDrawer.vue
+++ b/packages/x/src/components/x-drawer/XDrawer.vue
@@ -8,6 +8,7 @@
     :max-width="props.width"
     offset-top="var(--x-drawer-offset-top, 0)"
     data-testid="summary"
+    :z-index="zIndex"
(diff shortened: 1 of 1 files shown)


schogges:
Would this work instead of the scripting?

```CSS
:deep(.slideout-container) {
  z-index: auto !important;
}
```

johncowen:
Possibly yeah lemme try, I guess thats safer also 👍 .

We should maybe make an issue in kongponents also to allow `auto` 🤔

johncowen:
Yep LGTM, thanks @schogges 🎉 lemme push another commit one sec

johncowen:
K thats there, taking out of draft!

schogges (approved):

Issue #4889, opened 2026-04-28
[Mesh][ZoneIngress/ZoneEgress]: Reflection in Dataplane traffic listing


schogges:
### Description

Add entries to the traffic listing of a dataplane that shows the traffic going through (mesh scoped) zone proxies.

PR #4912, opened 2026-05-08
feat(dataplanes): listeners in traffic listing

schogges:
Listeners are being rendered next to the inbounds in the traffic listing. Although they are practically the same thing, I think for the moment we have to treat them differently.
This is an early stage and therefore there are some tradeoffs:
1. ~At some point we can fetch policies of listeners via the same request as inbounds: `/meshes/:mesh/dataplanes/:dp/_inbounds/:listener/_policies`. This is currently not fully implemented/errors.~ This seems to work now (tested with https://github.com/kumahq/kuma/commit/e3ded1b68). I've reused the inbound summary and connection views except `clusters`.
2. Stats around listeners are currently very minimalistic and we are missing cluster information in the layout. Once https://github.com/kumahq/kuma/issues/16692 is done, we can add more info. For the moment the summary of a listener will not show a cluster tab.
3. Currently `inbounds` and `listeners` are separate entries in `_overview:dataplane/networking` and `_layout`. Historically `inbound[]` also have `tags`, which is not the case for `listeners`. Basically they do not match 100%. Once this is the case we might be able to merge both `listeners` and `inbounds` either on the client side or in the backend.

Part of https://github.com/kumahq/kuma-gui/issues/4889


johncowen:
I have a feeling this PR is related to the sync we had yesterday (listener in traffic listing), let me know if not. It needs a rebase anyhow, so let me know when this is ready again. Aware this has been up for a while and I've unfortunately not had the time to get here yet, so please pester me for review once its ready again 👍

schogges:
> I have a feeling this PR is related to the sync we had yesterday (listener in traffic listing), let me know if not. It needs a rebase anyhow, so let me know when this is ready again. Aware this has been up for a while and I've unfortunately not had the time to get here yet, so please pester me for review once its ready again 👍

The main issue in this PR is that I added separate views to treat `listeners` different from `inbounds`, although they are more or less the same. At the moment they are not exactly the same. So we either choose to inline conditions and keep the views the same for inbounds and listeners, or we keep the separate views for listeners and may be able to clean things up once they are more aligned.

johncowen:
Ok 👍 would you say this is ready for review then? Or shall i wait a little?

schogges:
> Ok 👍 would you say this is ready for review then? Or shall i wait a little?

Let me give this one last check from my side. I'll let you know.

packages/kuma-gui/src/app/connections/components/connection-traffic/ConnectionCard.vue
```diff
@@ -141,6 +141,10 @@
             <dt>{{ t('data-planes.components.service_traffic_card.cx') }}</dt>
             <dd>{{ t('common.formats.integer', { value: props.traffic.tcp?.[`${props.direction}_cx_total`] }) }}</dd>
           </div>
+          <div>
+            <dt>{{ t('data-planes.components.service_traffic_card.cx_active') }}</dt>
+            <dd>{{ t('common.formats.integer', { value: props.traffic.tcp?.[`${props.direction}_cx_active`] }) }}</dd>
+          </div>
```
schogges, line 147: I've added active connections because it might be useful information for users.

packages/kuma-gui/src/app/connections/components/connection-traffic/ConnectionCard.vue
```diff
@@ -159,9 +163,11 @@
...
     </template>
     <template v-else>
-      <XProgress
-        variant="line"
-      />
+      <slot name="empty">
+        <XProgress
+          variant="line"
+        />
+      </slot>
```
schogges, line 170: Allow setting an empty state. Stats for listeners are very minimalistic and it's likely that there are no stats at all until some traffic went through. Therefore I wanted to add some empty state for listeners that tells the user that there is no data instead of showing a loading indicator which might be misinterpreted.

packages/kuma-gui/src/app/connections/routes.ts
```diff
@@ -66,5 +66,26 @@ export const routes = (prefix: string): RouteRecordRaw[] => {
...
+        },
+        {
+          path: 'xds-config',
+          name: `${prefix}-connection-listener-summary-xds-config-view`,
+          // Note: This uses the same component as the inbound summary xds view because they are practically the same,
+          // but for now we have to handle them as different summary views. This will likely change in the future.
+          component: () => import('@/app/connections/views/ConnectionInboundSummaryXdsConfigView.vue'),
+        },
+      ],
+    },
```
schogges, line 89: This is just routing configuration that might be removed in the future once we know wether we can treat listeners **exactly** the same as inbounds. For now we need to separate them because there are no clusters. The views for `stats` and `xds` are reusable already.

packages/kuma-gui/src/app/connections/sources.ts
```diff
@@ -68,13 +68,17 @@ export const sources = (api: KumaApi) => {
...
-      let inbounds, outbounds, passthrough
+      let inbounds, outbounds, passthrough, listeners
       if (proxyType === 'dataplane') {
         // pick out the listeners/inbounds that start with our ip address (the.ip.address.1_port000)
         inbounds = params.socketAddress === 'localhost' || params.socketAddress === 'self_inbound'
           ? Object.fromEntries(Object.entries(connections.cluster).filter(([key, _value]) => key.startsWith('localhost_') || key.startsWith('self_inbound')))
           : Object.fromEntries(Object.entries(connections.listener).filter(([key, value]) => key.startsWith(`${params.socketAddress}_`) && !value.$clusterName.startsWith('_')))
 
+        listeners = {
+          ...Object.fromEntries(Object.entries(connections.listener).filter(([key]) => key.startsWith('self_zone'))),
```
schogges, line 79: We might be able to merge inbounds and listeners here in the future by just checking for `self_` prefix instead of `self_inbounds` and `self_zone` prefixes.

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -33,10 +34,32 @@ export const DataplaneNetworkingLayout = {
           portName: kri.sectionName !== String(item.port) ? kri.sectionName : undefined,
         }
       }),
+      listeners: dataplaneNetworkingLayout.listeners.map(item => {
+        const kri = Kri.fromString(item.kri)
+        return {
+          ...item,
+          // protocol of ZoneIngress listeners is always tcp, for ZoneEgress it depends and therefore it's not set
+          protocol: item.type === 'ZoneIngress' ? 'tcp' : '',
```
schogges, line 43: @slonka Can you double check and verify this please? I've heard that this is the case during one of our syncs.

schogges (commented):
@johncowen Ok I think it's good for review now. Thanks for the patience 🙂

packages/kuma-gui/src/app/data-planes/views/DataPlaneInboundSummaryOverviewView.vue
johncowen, line 1: I was wondering if we can make DataplaneListeners closer to each other from a shape perspective that way we'd have less conditionals in here to check whether it has a certain prop. Instead we can just use `.length`s etc.

I'm not sure if this is possible as I don't have enough context but it feels like it would be better if we could smooth the old (inbounds) and the new (listeners) styles together so they almost act/are the same.

packages/kuma-gui/src/app/connections/routes.ts
johncowen: Do we only need this because we don't want to show a cluster tab if its a listener (instead of an inbound)?

johncowen (commented):
I guess my overall comment here is: 

How come we can't make `inbounds` and `listeners` the same shape and then just merge the arrays really high up in the data layer. Then everything underneath that in the View will "Just Work"

I think the only other thing that is slightly different is that listeners has no Clusters so we want to hide the Cluster tab in the summary view?

Are listeners always zoneProxies or some type (i.e. ingress, egress or both)?

packages/kuma-gui/src/app/data-planes/views/DataPlaneInboundSummaryOverviewView.vue
schogges: Oh, I did that already, but missed to update the view. But some of the entries remain as is and are truly different between classic inbounds and listeners:

- `tags` are not present on a listener
- `type` is never set on an inbound, it's `ZoneIngress` or `ZoneEgress` for listeners and it's always set
- `serviceAddressPort` does not really make sense for a listener because a zone-proxy does not connect to one specific service

I'd rather leave these three ☝️ as is

schogges:
> I guess my overall comment here is:
> 
> How come we can't make `inbounds` and `listeners` the same shape and then just merge the arrays really high up in the data layer. Then everything underneath that in the View will "Just Work"
> 
> I think the only other thing that is slightly different is that listeners has no Clusters so we want to hide the Cluster tab in the summary view?
> 
> Are listeners always zoneProxies or some type (i.e. ingress, egress or both)?

I think that's the plan for the future, but I'm tempted to leave it as is. This works for now, but I've added comments for the next iteration to make it cleaner. This also depends a little bit on how the backend makes changes regarding listeners and their stats.

packages/kuma-gui/src/app/connections/routes.ts
schogges: Yeah this is what I left after removing all the specific views for listeners.

johncowen: Is it possible to figure out if something is a listener (maybe by its `type`) and then just wrap a conditional around the tab we shouldn't show in the template?

I just feel that for something that is very possibly super temporary it would be easier, less complex and easier to remove when we come to do that might be a better option here.

I feel like there is this routing config, I think there was a bit more routing config code in dataplanes also, and I feel like in this case a little `if` around the Cluster tab would be fine temporarily.

Maybe its not possible to do that, or its not as straight forwards as I'm imagining, but rather share the idea than not

johncowen (dismissed):
Quick approval for this, it looks better to me from the last round of changes. I think we are planning on circling back in this anyway.

schogges:
> Quick approval for this, it looks better to me from the last round of changes. I think we are planning on circling back in this anyway.

I made another iteration, but I think it's not finished yet. There are still some things I'd like to clean up. But we need to get this in asap to be able to test it properly. I've changed the `Closes #xyz` to `Part of #xyz` to keep the issue open. I'll keep working on a follow-up right away.

schogges:
@johncowen had to resolve conflicts, need another approval

johncowen (approved):
👍

packages/kuma-gui/src/app/connections/routes.ts
schogges: That's what I want to try as a follow-up 🙂

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
schogges: Quoting @lobkovilya 

>I n Envoy, there is no concept of a “listener protocol”. Each listener has one or more filter chains, and each filter chain can contain TCP or HTTP filters.
>
> For sidecar inbounds, we have one filter chain per listener, so in Kuma we can unambiguously treat the filter chain protocol as the listener protocol.
>
> However, for ZoneEgress and ZoneIngress, there are multiple filter chains. In the ZoneEgress case, the protocol of each filter chain is defined by the corresponding MeshExternalService. For ZoneIngress, you’re right that all filter chains are TCP. However, since in the general case we can’t reliably infer a single protocol for the listener, I think it’s fine to omit it.

For now, until we have more information, we will leave the `protocol` empty for `ZoneEgress` and set it to `tcp` for `ZoneIngress`

PR #4917, opened 2026-05-11
chore(dataplane): initial support for top-level dataplane labels

### LINKED ISSUE #4570: Review where we are using Dataplane tags internally or user facing
### Description

Tags are going so if we are using them to map things internally we neeed to find a replacement approach.

This issue could be just investagation or also actual replacement implementation

johncowen:
Adds more support for `labels` at the root of Dataplane/DataplaneOverview

The majority of this is mocking (both in the mocks and in the test override mocks)

In the application/data layer code itself, I've tried to keep a fallback to tags whilst preferring labels i.e. if a label is there we prefer that, if not fallback to previous tag functionality.

I noticed that a lot of the tag based functionality we have is using "deeper tags" i.e. in `.gateway` etc, which according to the OpenAPI don't have labels (at least yet). I need to confirm whether I need to consider labels in these deeper areas and then add support (preferably via OpenAPI first). I've loeft some comments in the code to remind me where these bits are and for reviewing. Personally I'd be fine merge these but also happy to remove these before merge if necessary.

Note: This PR does not add additional rendering of labels anywhere.

Part of and possibly closes https://github.com/kumahq/kuma-gui/issues/4570


packages/kuma-gui/features/mesh/dataplanes/Index.feature
```diff
@@ -28,6 +28,8 @@ Feature: mesh / dataplanes / index
         items:
         - name: fake-backend
           mesh: fake-default
+          labels:
+            kuma.io/display-name: !!js/undefined
```
schogges, line 36: Why does this need to be explicitely undefined while the others get set explicitely to the same value as name?

packages/kuma-gui/features/mesh/legacy-dataplanes/Index.feature
```diff
@@ -28,6 +28,8 @@ Feature: mesh / dataplanes / index
         items:
         - name: fake-backend
           mesh: fake-default
+          labels:
+            kuma.io/display-name: !!js/undefined
```
schogges: s.a.

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
```diff
@@ -46,21 +48,29 @@ export default ({ env, fake }: Dependencies): ResponseHandler => (req) => {
...
   return {
     headers: {},
     body: {
       type: 'DataplaneOverview',
       mesh,
       name,
+      labels: {
+        ...fake.kuma.labels({
+          name: displayName,
+          ...(multizone ? { zone } : {}),
```
schogges: I think in case of multizone we also need to set `kuma.io/origin: 'zone'`?

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
```diff
@@ -88,7 +98,7 @@ export default ({ env, fake }: Dependencies): ResponseHandler => (req) => {
                 tags: fake.kuma.tags({
                   protocol: ports[i].protocol,
                   service,
-                  zone: isMultizone && fake.datatype.boolean() ? fake.word.noun() : undefined,
+                  zone: multizone && fake.datatype.boolean() ? fake.word.noun() : undefined,
```
schogges: I think inbounds always have the same zone as the DP?

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
```diff
@@ -46,21 +48,29 @@ export default ({ env, fake }: Dependencies): ResponseHandler => (req) => {
...
 
   return {
     headers: {},
     body: {
       type: 'DataplaneOverview',
       mesh,
       name,
+      labels: {
+        ...fake.kuma.labels({
+          name: displayName,
```
schogges, line 86: Should we fall back to `name` if `displayName` is empty?

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_.ts
```diff
@@ -54,7 +67,7 @@ export default ({ fake, env }: Dependencies): ResponseHandler => (req) => {
                 const tags = fake.kuma.tags({
                   protocol: fake.kuma.protocol(),
                   service: fake.kuma.serviceName(),
-                  zone: isMultizone && fake.datatype.boolean() ? fake.word.noun() : undefined,
+                  zone: multizone && fake.datatype.boolean() ? fake.word.noun() : undefined,
```
schogges, line 83: s.a.

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_overview.ts
```diff
@@ -69,6 +71,13 @@ export default ({ fake, pager, env }: Dependencies): ResponseHandler => (req) =>
         return {
           type: 'DataplaneOverview',
           mesh,
+          labels: {
+            ...fake.kuma.labels({
+              name: displayName,
+              ...(multizone ? { zone } : {}),
```
schogges, line 89: s.a.

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_.ts
```diff
@@ -1,27 +1,40 @@
...
         apiVersion: 'kuma.io/v1alpha1',
-      }),
+      } : {}),
       type: 'Dataplane',
       mesh,
       name,
+      labels: {
+        ...fake.kuma.labels({
+          name: displayName,
+          ...(multizone ? { zone } : {}),
```
schogges, line 47: s.a.

packages/kuma-gui/features/mesh/dataplanes/Index.feature
johncowen: I wanted to keep at least one test with an unset `label['kuma.io/display-name']` just to ensure that if there is no label set it continues to work as before. Previously the mocks (and kuma itself) had no labels here.

packages/kuma-gui/features/mesh/legacy-dataplanes/Index.feature
johncowen: resolving this one, see above for convo

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: See https://github.com/kumahq/kuma-gui/pull/4917/changes#diff-d868f83b2a3e57da334685ff25081d1c5c27a023f4c8d1bbff199e4b999bbc60R326

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_.ts
johncowen: This isn't quite the same as above because this is tags not labels.

The whole PR is about not using `tags` anywhere anymore, and since this has been like this for a veeeeeeeeery long time, I don't think its hugely important.

But I think I'm going to make a change here anyway for something else, so I can look at doing the same as I've done for `fake.kuma.labels` and automatically add the `kuma.io/origin` if you specify a zone, which I think is what we want here? lmk

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: Since:

- This was how it worked before
- I'm not sure either
- In this particular case, I don't think it matters as we aren't trying to recreate a stateful and correct (from a data integrity perspective) application here, only stateless mocks.

I'm inclined to leave it as is. lmk!

packages/kuma-gui/features/mesh/dataplanes/Index.feature
schogges: Sounds like this is a special case you want to keep. I think it's worth adding a comment here.

schogges:
I'm not following the last [commit](https://github.com/kumahq/kuma-gui/pull/4917/changes/9fec5c5002bd10650c666f23715f614037a2e480). What are you trying to improve with this?

johncowen:
Oh sorry I meant to comment on here after I added the commit (I figured it would be of interest so I made sure I did an extra commit). Bare with me a sec and I'll come back and add some comment around it 👍

johncowen:
Ok so the main reason I added the next commit was mainly a result of this comment:

https://github.com/kumahq/kuma-gui/pull/4917#discussion_r3236118387

> Should we fall back to name if displayName is empty?

Its a bit of a long story/thought process but here goes 🤞 :

- I started looking the the fallbacks for displayName and hit the KRI stuff. 
- Then it took me a good while to remember that we did the weird aliasing thing so that both endpoints go to the same file.
- Next, I thought "it would have taken me less time to realise this if the files where on the filesystem", and also we should be adding trying to add new features and providing backwards compatibility rather than keeping old features and providing backwards compatibility.
- This thought process led me to adding the new on-filesystem files, with a simple `export`/alias to the old file. Less importantly but something we should do at some point would be to switch this around so the alias is in the old stuff. That being said we barely make `/_kri` HTTP calls as of right now, so I would guess we would move this around the same time as your KRI PRs that are up.
- I also noticed that the KRI/params code at the top of the files where we have this was a little inconsistent and (I can't remember exactly how) potentially not 100% ideal.
- I then worked on finding an easy copy/pastable chunk that I could copy/pasta and minimally change where necessary. Added a bunch of comments to make it super clear for future me.
- I then finished up by doing the copy/pasta of the same chunk and amending where necessary.
- Finally I changed any previous `label` usage to use the new `fake.kuma.labels` and the newly computed labels instead.
- Oh! Lastly I noticed the workloads endpoint on the filesystem weren't following the same pattern as all the rest of the endpoints (we noticed this with another resource a little while back) so I renamed them to follow the pattern elsewhere.

Ok, hopefully that gives the background, lemme know if you had more q's 👍

packages/kuma-http-api/mocks/src/_kri/kri_dp__.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from '../meshes/_/dataplanes/_/_overview'
```
schogges: Oh, I've just realized that this wrongly used the `_overview` mock, but a KRI of a dataplane should actually return a `Dataplane` and not a `DataplaneOverview`. Also `DataplaneOverview`s do not have a KRI.

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
```diff
@@ -1,12 +1,26 @@
...
+    nspace = k8s ? fake.word.noun() : '',
+    displayName,
+  ] = kri ? kri.split('_') : [
+    'kri', // prefix
+    'dp', // shortName
+    String(req.params.mesh), // mesh
+    fake.helpers.arrayElement(['', fake.word.noun()]), // zone
+    ...String(req.params.name).split('.').toReversed(), // nspace, displayName
+  ]
+  const name = kri ? `${displayName}${nspace ? `.${nspace}` : ''}` : String(req.params.name)
```
schogges: s.a. this is/was actually not correct. We can't retrieve a `DataplaneOverview` via KRI.

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_.ts
schogges, line 1: The KRI handling in `DataplaneOverview` ☝️ should happen in here.

packages/kuma-http-api/mocks/src/zoneegresses/_.ts
```diff
@@ -1,13 +1,29 @@
...
+    zone,
+    // if its not a kri (which always has a nspace, even if it's ''), or the
+    // name has no '.', then, if its k8s use a random nspace, otherwise ''
+    nspace = k8s ? fake.word.noun() : '',
+    displayName,
+  ] = kri ? kri.split('_') : [
+    'kri', // prefix
+    'ze', // shortName
+    String(req.params.mesh), // mesh
+    // @TODO
```
schogges: Is the `@TODO` a leftover?

schogges (commented):
Since the current mock around retrieving a DP via KRI is wrongly returning a `DataplaneOverview` my comments below can be done in a follow-up or here. Lmk 🙂

packages/kuma-http-api/mocks/src/_kri/kri_dp__.ts
johncowen: Ah nice spot thankyou!

packages/kuma-http-api/mocks/src/zoneegresses/_.ts
johncowen: ah perfect! Yes it will be, thanks for nudge!

johncowen:
Sorry for the delay here, just coming back round to this now. Need to rebase then I'll take a look at the feedback 👍

packages/kuma-gui/features/mesh/dataplanes/Index.feature
johncowen: Added a comment during rebase

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: Its funny actually because I think this is one of the things that made me even do this commit. I was looking at this file and thinking "why does this file even have the KRI stuff? You can't access this via a KRI". I'm glad the diffs here have made this clearer and we can sort 👍

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: This was resolved by my second commit here but it turns out this file doesn't even need KRI support, so I'll make another commit here to resolve this 👍

packages/kuma-http-api/mocks/src/zoneegresses/_.ts
johncowen: Resolved in https://github.com/kumahq/kuma-gui/pull/4917/changes/100524763ca8ac06ac839d9420b63fc0b29c660c

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_.ts
johncowen: Resolved in https://github.com/kumahq/kuma-gui/pull/4917/commits/e92510b519681c563fb6322029b29700070fc353

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: Resolved in https://github.com/kumahq/kuma-gui/pull/4917/commits/e92510b519681c563fb6322029b29700070fc353

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: Resolving this one, I've used a new approach including the KRI version of the endpoints to default all the KRI related segments, resolving this one.

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: Resolved this one, but feel free to reopen if necessary 👍

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
```diff
@@ -1,12 +1,17 @@
...
+    mesh,
+    zone,
+    nspace = '',
+    displayName,
+  ] = [
+    String(req.params.mesh), // mesh
+    fake.helpers.arrayElement(['', fake.word.noun()]), // zone
+    ...String(req.params.name).split('.').toReversed(), // nspace, displayName
+  ]
+  const name = String(req.params.name)
```
schogges, line 15: Even though I like this approach to get the `displayName` and `namespace` a lot more than what we had before, I think this is still not 100% correct. Correct me if I'm wrong and misunderstood something. Just from reading I currently assume that this results in incorrect displayName when there is no namespace in the name.

```ts
✅ 
const [ns = '', displayName] = 'foo.bar'.split('.').toReversed() // -> ns: bar, displayName: foo

❌ 
const [ns = '', displayName] = 'foo'.split('.').toReversed() // -> ns: foo, displayName: undefined
```

Furthermore, there may be more dots in the name, i.e. `a.service.kuma-demo`. So we need to join back the dots after splitting up.
I remember myself trying to find a more elegant and better working approach for this to replace what we had before.

Since this is just mocks and it seems to work for now I'm fine putting this on a follow-up, but still wanted to bring it up 🙂

schogges (dismissed):
One thing, but I think this is non-blocking

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
johncowen: Oh yeah, your red X case here does look like this is wrong, one sec lemme look at this again

johncowen: K what we discussed has now been applied everywhere here lmk!

schogges (approved):
LGTM 👍

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_overview.ts
schogges: Thanks, this looks good now 👍

PR #4930, opened 2026-05-15
chore(dataplanes): implements kri 'protocol' links

johncowen:
Allows us to link by full ~or partial~ KRIs i.e. `href="kri://kri_m_____default_"` etc.

Screenshot 2026-05-15 at 18 50 22

I've only implemented this in the non-legacy DataplaneDetailView labels for now, with an eye to using it everywhere else we use labels (or tags) following merge of this. Currently supported labels are `kuma.io/mesh`, `kuma.io/zone` and `kuma.io/workload`.

~Still a bit TBD, and there is still a little work to come on this PR but I'll add some inlines now to help review.~ I think I've finished on this now, so this is definitely good for a review now. I'd also like to use it to apply to labels on other PRs I've been using (those PRs aren't merged yet)

I thought it was useful to include the link to the original PR for protocol handlers https://github.com/kumahq/kuma-gui/pull/4256

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -238,21 +238,27 @@
                         <dt>{{ t('data-planes.routes.item.labels') }}</dt>
                         <dd>
                           <XLayout
+                            v-for="kumaRe in [/^(.+\.)?kuma\.io\//]"
+                            :key="typeof kumaRe"
                             variant="separated"
                             truncate
                           >
-                            <template
-                              v-for="kumaRe in [/^(.+\.)?kuma\.io\//]"
-                              :key="typeof kumaRe"
+                            <XAction
+                              v-for="[key, value] in labels"
+                              :key="key"
+                              :href="t(`common.kri.labelHrefs.${key.replaceAll('.', '~')}`, {
+                                mesh: props.data.mesh,
+                                zone: props.data.zone,
+                                namespace: props.data.namespace,
+                                name: value,
+                              }, { defaultMessage: '' })"
                             >
                               <XBadge
-                                v-for="[key, value] in labels"
-                                :key="key"
                                 :appearance="kumaRe.test(key) ? 'info' : 'decorative'"
                               >
                                 {{ key }}:{{ value }}
                               </XBadge>
-                            </template>
+                            </XAction>
                           </XLayout>
                         </dd>
                       </div>
--- a/packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
+++ b/packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
@@ -1,4 +1,13 @@
 common:
+  kri:
+    shortNames:
+      mesh: m
+      zone: z
+      workload: wl
+    labelHrefs:
+      kuma~io/mesh: kri://kri_m____{name}_
+      kuma~io/zone: kri://kri_z____{name}_
+      kuma~io/workload: kri://kri_wl_{mesh}_{zone}_{namespace}_{name}_
   i18n:
     ignore-error: ""
   product:
--- a/packages/kuma-gui/src/app/service-mesh/index.ts
+++ b/packages/kuma-gui/src/app/service-mesh/index.ts
@@ -3,6 +3,7 @@ import { token } from '@kumahq/container'
 import X from '@kumahq/x'
 import { useLink } from 'vue-router'
 
+import type { Can } from '@/app/application'
 import { services as controlPlanes } from '@/app/control-planes'
 import { services as hostnameGenerators } from '@/app/hostname-generators'
 import { Kri } from '@/app/kuma'
@@ -10,66 +11,41 @@ import { services as me } from '@/app/me'
 import { services as meshes } from '@/app/meshes'
 import { services as zones } from '@/app/zones'
 import type { ServiceDefinition, Token } from '@kumahq/container'
-import type { Router } from 'vue-router'
 
-const findAnchor = (target: HTMLElement) => {
-  // we look for anchors, or any other element that has [data-actionable]
-  const $el = target.tagName.toLowerCase() === 'a' ? target : target.closest('a,[data-actionable]')
-  if($el) {
-    switch(true) {
-      // if its a data-action element we "bubble down" to find a child [data-action]
-      case $el.hasAttribute('data-actionable'):
-        return $el.querySelector('[data-action]')
-      // otherwise we check for rel="x-internal"
-      // which allows us to use vue-router links in our i18n locales files via
-      // <a href="http-link" rel="x-internal" />
-      // and helps us support custom protocols
-      case ($el.getAttribute('rel') ?? '').includes('x-internal'):
-        return $el
-    }
-  }
-  // anything else we do nothing special
-  return null
-}
-const protocolHandler = (router: Router, doc = document) => {
-
-  const base = router.options.history.base
-
-  const listener = (e: Event) => {
-    // if its not a user generated event return
-    // if we ever need this listener to also fire on non-user generated events
-    // we'll have to refine findAnchor and remove this
-    if(!e.isTrusted) {
-      return
-    }
-    const $a = findAnchor(e.target as HTMLElement)
-    if($a) {
-      // anything with x-internal is something for whatever reason we can't/don't want to use RouterLink
-      if (($a.getAttribute('rel') ?? '').includes('x-internal')) {
-        const href = $a.getAttribute('href')
-        if(href) {
-          e.preventDefault()
-          e.stopPropagation()
-          router.push(href.startsWith(base) ? href.substring(base.length) : href)
-        }
-      // anything else
-      } else if('click' in $a && typeof $a.click === 'function') {
-        $a.click()
-      }
-    }
-  }
-  // install
-  doc.body.addEventListener('click', listener)
-
+const protocolHandler = (can: Can) => {
   return (href: string) => {
     const kriProto = 'kri://'
     switch (true) {
       case href.startsWith(kriProto): {
-        const { mesh, name, namespace, shortName } = Kri.fromString(href.substring(kriProto.length))
+        const { mesh, name, zone, namespace, shortName } = Kri.fromString(href.substring(kriProto.length))
         const id = `${name}${namespace !== '' ? `.${namespace}`: '' }`
 
         const to = (() => {
           switch (true) {
+            case shortName === 'm':
+              return {
+                name: 'mesh-detail-view',
+                params: {
+                  mesh: name,
+                },
+              }
+            case shortName === 'z':
+              if(can('use zones')) {
+                return {
+                  name: 'zone-cp-detail-view',
+                  params: {
+                    zone: name,
+                  },
+                }
+              }
+              break
+            case shortName === 'wl':
+              return {
+                name: 'workload-detail-view',
+                params: {
+                  wl: Kri.toString({ shortName, mesh, zone, namespace, name }),
+                },
+              }
             case shortName === 'msvc':
               return {
                 name: 'mesh-service-detail-view',
@@ -113,18 +89,18 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
   return [
     ...me(app),
     [token('service-mesh.plugins'), {
-      service: (i18n, router) => {
+      service: (i18n, can) => {
         return [
           [Kongponents],
           [X, {
             i18n,
-            protocolHandler: protocolHandler(router),
+            protocolHandler: protocolHandler(can),
           }],
         ]
       },
       arguments: [
         app.i18n,
-        app.router,
+        app.can,
       ],
       labels: [
         app.plugins,
--- a/packages/x/src/components/x-action/XAction.vue
+++ b/packages/x/src/components/x-action/XAction.vue
@@ -176,15 +176,83 @@
     </template>
   </template>
 </template>
+<script lang="ts">
+// eslint-disable-next-line import/order
+import SharedPool from '../../utilities/SharedPool'
+// eslint-disable-next-line import/order
+import type { Router , RouteLocationNamedRaw } from 'vue-router'
+
+const findAnchor = (target: HTMLElement) => {
+  // we look for anchors, or any other element that has [data-actionable]
+  const $el = target.tagName.toLowerCase() === 'a' ? target : target.closest('a,[data-actionable]')
+  if($el) {
+    switch(true) {
+      // if its a data-action element we "bubble down" to find a child [data-action]
+      case $el.hasAttribute('data-actionable'):
+        return $el.querySelector('[data-action]')
+      // otherwise we check for rel="x-internal"
+      // which allows us to use vue-router links in our i18n locales files via
+      // <a href="http-link" rel="x-internal" />
+      // and helps us support custom protocols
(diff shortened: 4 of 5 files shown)


packages/kuma-gui/src/app/service-mesh/index.ts
```diff
@@ -3,73 +3,52 @@ import { token } from '@kumahq/container'
...
-          router.push(href.startsWith(base) ? href.substring(base.length) : href)
-        }
-      // anything else
-      } else if('click' in $a && typeof $a.click === 'function') {
-        $a.click()
-      }
-    }
-  }
-  // install
-  doc.body.addEventListener('click', listener)
```
johncowen, line 62: All of this has been moved into XAction, leaving only the specific `kri://` protocol handling as a specific "kuma application" thing

packages/kuma-gui/src/app/service-mesh/index.ts
```diff
@@ -3,73 +3,52 @@ import { token } from '@kumahq/container'
...
+                }
+              }
+              break
+            case shortName === 'wl':
+              return {
+                name: 'workload-detail-view',
+                params: {
+                  wl: Kri.toString({ mesh: mesh || String(params.mesh), zone: zone || 'default', namespace, shortName, name }),
+                },
+              }
```
johncowen, line 48: For the moment we should just add more and more "kri protocol handlers" here as and when we can. If something isn't dealt with here, then the link will not produce a anchor tag.

I'm thinking we could probably somehow make the default for valid KRIs the new generic resource detail page. That way if we get new resources that we don't add handlers for they will just use the generic page.

packages/x/src/components/x-action/XAction.vue
```diff
@@ -176,15 +176,83 @@
     </template>
   </template>
 </template>
+<script lang="ts">
+// eslint-disable-next-line import/order
+import SharedPool from '../../utilities/SharedPool'
+// eslint-disable-next-line import/order
+import type { Router , RouteLocationNamedRaw } from 'vue-router'
```
johncowen, line 183: We really must get rid of the annoying `import/order` rule once and for all, its extremely frustrating and worse than pointless.

packages/x/src/components/x-action/XAction.vue
```diff
@@ -176,15 +176,83 @@
...
+    }
+    case 'acquiring':
+      return item
+    case 'releasing':
+      return item
+    case 'destroying':
+      document.body.removeEventListener('click', item)
+      return item
+  }
+})
```
johncowen, line 246: This is almost just a copy of what was in the `service-mesh/index` file, with added singleton management.

packages/x/src/components/x-action/XAction.vue
```diff
@@ -221,12 +289,16 @@ const group = inject<{
...
+onMounted(() => pool.acquire(router, sym))
+onBeforeUnmount(() => pool.release(router, sym))
+
 const protocolHandler = useProtocolHandler()
 const href = computed(() => props.href.includes('://') ? protocolHandler(props.href) : props.href)
 const target = computed(() => props.href.length > 0 && props.href === href.value ? '_blank' : undefined)
 const rel = computed(() => target.value === '_blank' ? 'noopener noreferrer' : href.value.length > 0 && props.href === href.value ? undefined : 'x-internal')
 
-if(href.value.length > 0 || typeof attrs.onClick === 'function' || props.for || props.to) {
+if(href.value.length > 0 || typeof attrs.onClick === 'function' || props.for || Object.keys(props.to).length) {
```
johncowen, line 301: This is a little bug (one that didn't affect us yet I don't think), similar to another we had recently, in XAction `to` currently defaults to `{}`.

packages/x/src/utilities/SharedPool.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  destroy() {
+    Array.from(this.pool.entries()).forEach(([key, item]) => {
+      Array.from(item.references).forEach((ref) => {
+        this.release(key, ref)
+      })
+    })
+  }
+}
```
johncowen, line 55: This is "sometimes a little copy/pasta is better than a little dependency". I copied this directly with no changes from `@kumahq/data`.

johncowen (commented):
Notes:

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -217,21 +217,22 @@
...
                             >
                               <XBadge
-                                v-for="[key, value] in labels"
-                                :key="key"
                                 :appearance="kumaRe.test(key) ? 'info' : 'decorative'"
                               >
                                 {{ key }}:{{ value }}
                               </XBadge>
-                            </template>
+                            </XAction>
```
johncowen, line 261: The idea here is you need very little code or custom components to do this, and its pretty clear what is happening.

You add an anchor link (XAction), with a href (a KRI href) and it just works.

johncowen:
I put this up last thing last Friday and there were a couple of things I spotted that I wanted to continue on, hence I put it in Draft. But as we are discussing it now I think it'd be best if I take it out of Draft so its clear that I'd like review on this.

Just take into account I might make the odd change still myself.

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
```diff
@@ -1,4 +1,12 @@
 common:
+  kri:
+    shortNames:
+      mesh: m
+      zone: z
+    labelHrefs:
+      kuma~io/mesh: kri://kri_m_{name}___{name}_
+      kuma~io/zone: kri://kri_z__{name}__{name}_
```
schogges: On a first glance this looks incorrect to me 🤔 The KRI of a mesh and a zone is only `kri_m____{name}_` and `kri_z____{name}_` (the name is only in the name section).

I will continue to review, but wanted to leave this comment already

packages/x/src/components/x-action/XAction.vue
schogges, line 1: My main question is: Do we need all of this? Since we are already within `XAction` we are controlling here what element is being rendered (`a`-tag, `KButton`, ...) and we don't need to find the anchor first. So I think it would be possible to just add a click-handler to the rendering of `a`-tag:

```vue
<a
  v-else
  ...
  @click="(e) => {
    if(rel?.includes('x-internal')) {
      e.preventDefault()
      e.stopPropagation()
      const base = router.options.history.base
      router.push(href.startsWith(base) ? href.substring(base.length) : href)
    }
  }"
>
```

packages/x/src/utilities/SharedPool.ts
schogges, line 1: If we can do above ☝️ we don't need this either

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
johncowen: hmm, yeah maybe I was unsure about this to be honest, I originally had the name only in one place. Let me put it back to just the one place. Do you know where the docs are to explain how a mesh/zone KRI should be?

packages/x/src/components/x-action/XAction.vue
johncowen: - What about for clicks on things that aren't links, but contain the link, i.e. table rows and traffic cards?
- What about for `a` links in the locales files where we can't use `@click`? (see comments in the code)

These are the things that immediately come to mind, but I can only vaguely remember the reasons behind all this code (that I pretty much just moved). Let me know if you have further questions and I can try and dig into the history/try remember a little more.

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
schogges: Not sure 🤔 but I see it as it's a scope/context, i.e. a resource in the scope/context of mesh `{name}` has the mesh-section set to the name of the mesh it belongs to.

packages/x/src/components/x-action/XAction.vue
schogges: > * What about for clicks on things that aren't links, but contain the link, i.e. table rows and traffic cards?
> * What about for `a` links in the locales files where we can't use `@click`? (see comments in the code)
> 
> These are the things that immediately come to mind, but I can only vaguely remember the reasons behind all this code (that I pretty much just moved). Let me know if you have further questions and I can try and dig into the history/try remember a little more.

Maybe the moving is the issue then. Would these still work if there was no `XAction` rendered anywhere else on the page, i.e. `onMounted` would never be triggered?

johncowen: I guess for the first one `clicks on things that aren't links, but contain the link`, its for the usecase where there is a XAction inside something else (so a XAction inside a traffic card), so an XAction would be in the page.

I think for the second one, I guess theoretically its possible to have a page with no links in it (take into account _in Kuma_ the main navigation uses XActions), but its a bit of a long shot that we would have a page with no links but we are _also_ using a "locale anchor link". I guess we could improve this a little but I don't think we would ever hit this (and if we did the "bug" would be that the link would still work but it would cause a page refresh). Its not something I'm concerned about making sure we improve here, happy to make an issue though?

schogges: What issue are we trying to solve by moving the code for the listener? If I'm not mistaken it would all still work if we'd leave the stuff where it is? But by moving the code to XAction, even though the chances are very low, we introduce a very minor risk that it might break or there is a regression in the UX.
If the reason for moving the code was about trying to separate the protocol handler from the listener registration, I would argue to put this in a follow-up and think about it separately.

johncowen: Q: What happens when you add event listeners and don't remove them?

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
johncowen: I've removed the mesh/zone segments from these bits (mesh and zone) and just used name

packages/x/src/components/x-action/XAction.vue
schogges: They stay until you close the page, which I think is fine if they get only added once and act for the entire app.

schogges: I'm just trying to do the minimum now and separately search for the best way of doing all of this, considering the different approaches and situations.

johncowen: > if they get only added once and act for the entire app.

remember what changed recently elsewhere. If we don't clean up and keep running this code over and over then we'll get a memory leak

> I'm just trying to do the minimum now and separately search for the best way of doing all of this, considering the different approaches and situations.

Yeah totally get you, as I said I pretty much just moved the code and added some cleanup bookkeeping via SharedPool (that I just took as-is from the other place we have this).

There is a small edge case that you pointed out (if we add an anchor into locales and then use that anchor on a page that has no XActions), my thoughts on this are:

- Every page here uses XActions for the main navigation
- Where we don't use our own main navigation, every page I can think of has an XAction. The only page I can think of that doesn't have an XAction is the configuration page, which we only show here in Kuma.
- In order for the un-intentional thing to happen, we need both a local anchor being used on the same page as a page with no XAction.
- If/when that highly unlikely occurrence happens. The "bug" is that if someone clicks the locale based link, they _still_ go to where they were trying to get to, just instead of a HistoryAPI navigation they will get a traditional browser navigation.

All in all I don't think its worth overthinking things (and changing) too much at this stage, but as I said I'm happy to make an issue to cover this if you'd like.

packages/kuma-gui/src/app/kuma/locales/en-us/index.yaml
johncowen: Gonna resolve this one

packages/x/src/components/x-action/XAction.vue
schogges: > > if they get only added once and act for the entire app.
> 
> remember what changed recently elsewhere. If we don't clean up and keep running this code over and over then we'll get a memory leak

Well, that's an argument 👍 although it might have been good to solve this separately, i.e. the `listener` does not need to be a closure and we could control the registration in the lifecycle of the application.

> > I'm just trying to do the minimum now and separately search for the best way of doing all of this, considering the different approaches and situations.
> 
> Yeah totally get you, as I said I pretty much just moved the code and added some cleanup bookkeeping via SharedPool (that I just took as-is from the other place we have this).
> 
> There is a small edge case that you pointed out (if we add an anchor into locales and then use that anchor on a page that has no XActions), my thoughts on this are:
> 
> * Every page here uses XActions for the main navigation
> * Where we don't use our own main navigation, every page I can think of has an XAction. The only page I can think of that doesn't have an XAction is the configuration page, which we only show here in Kuma.
> * In order for the un-intentional thing to happen, we need both a local anchor being used on the same page as a page with no XAction.
> * If/when that highly unlikely occurrence happens. The "bug" is that if someone clicks the locale based link, they _still_ go to where they were trying to get to, just instead of a HistoryAPI navigation they will get a traditional browser navigation.
> 
> All in all I don't think its worth overthinking things (and changing) too much at this stage, but as I said I'm happy to make an issue to cover this if you'd like.

I think that this is still not the best and final solution and therefore it would be great to have an issue to track it.

johncowen: We had a good deal of chat about this and things are a lot clearer now - ideally I should have put something in the description to explain why I'm moving the code I moved here due to the eventListener thing.

End of the day, we are happy to give this PR a final look, but in regards to this conversation we are happy to approve/merge as-is.

We are also going to make an issue as a follow up for a rethink of exactly how we want to deal with linking from labels and such-like with KRIs. Lets make sure we link the issue here also once thats created.

I'm gonna resolve this.

packages/kuma-gui/src/app/service-mesh/index.ts
johncowen: > I'm thinking we could probably somehow make the default for valid KRIs the new generic resource detail page. That way if we get new resources that we don't add handlers for they will just use the generic page.

I would definitely like to do this, but there are a few little issues we would need to change/alter to be able to do this. I would guess this will be outlined in the issue we are going to make for KRI links, but I need to think about it a little more to be sure - I'm also wondering if we should do this for 2.14.

For the moment I'm going to resolve this comment as I'm definitely not going to add any more here. At the moment an incorrect or unsupported KRI "handler" will just not show a link and just use a span instead.

packages/x/src/utilities/SharedPool.ts
johncowen: I'm going to resolve this comment as I think we've dealt with it in the above comment

schogges (approved):
I've created a follow-up to iterate on this: https://github.com/kumahq/kuma-gui/issues/4972

johncowen:
Just to note for this:

(https://github.com/kumahq/kuma-gui/pull/4930#discussion_r3317521572)

> I guess for the first one clicks on things that aren't links, but contain the link, its for the usecase where there is a XAction inside something else (so a XAction inside a traffic card), so an XAction would be in the page.

> I think for the second one, I guess theoretically its possible to have a page with no links in it (take into account in Kuma the main navigation uses XActions), but its a bit of a long shot that we would have a page with no links but we are also using a "locale anchor link". I guess we could improve this a little but I don't think we would ever hit this (and if we did the "bug" would be that the link would still work but it would cause a page refresh). Its not something I'm concerned about making sure we improve here, happy to make an issue though?

I had an idea to solve this edge case also, but we didn't get a chance to try it out. I will give it a go once I get a chance tho. I can't think of why it wouldn't cover all bases including this edge-case.

Gonna go aheead and merge this!

PR #4935, opened 2026-05-18
feat(meshes): remove ff for resources routes

schogges:
Removes the feature flag for the resources routes in a mesh.


packages/kuma-gui/src/app/meshes/features.ts
```diff
@@ -1,16 +0,0 @@
-import type { Env } from '@/app/application'
-import type { Features } from '@kumahq/settings/can'
-
-export const features = (env: Env) => {
-  return {
-    'use resources route': (_can: unknown) => {
-      return env('KUMA_RESOURCES_ROUTE_ENABLED') === 'true'
```
johncowen, line 7: Could this PR just be defaulting this to `true` for now? If we need this to be controllable from elsewhere we can can then just overwrite this feature definition couldn't we?

johncowen: Bit more information here, ideally this would all be in a `resources` module (well, not `meshes`), that means we can enabled disable the module as a whole. Thing is we've bumped into some super old legacy code here, plus the weird "we have resources as an abstract class type thing, and also as a concrete thing so it difficult to name things".

I think the above, coupled with the fact that we the feature flagging will only be pretty temporary makes me thing that it might be easier to just leave everything as is for the moment (and default to `true`), then if we need to overwrite this elsewhere we can.

Overall I just think it will be easier to get onto other things shorter term and come back here later.

Let's sync on this though as I could be wrong

johncowen (approved):
Lets get this in!

PR #4936, opened 2026-05-18
refactor(resources): categorization

schogges:
The current categorization was temporarily added to gather feedback but was not meant to stay as is. At some point they might even be set in the backend. Based on the [feedback](https://github.com/kumahq/kuma/issues/16002#issuecomment-4171591616) this is another iteration to finish this up for now.

---

~I'm still waiting for more feedback before I'll move on with this, because this is now only categorizing policies. I could assume that i.e. `MeshIdentity` could be categorized under `security` aswell.~ We can use the current categorization in this PR for 2.14, although we might add categories to the other non-policy resources subsequently.


packages/kuma-gui/src/app/resources/data/ResourceTypeDescriptor.ts
```diff
@@ -4,49 +4,45 @@ type KumaResourceTypeDescriptorCollection = paths['/_resources']['get']['respons
...
-  ['zone', ['traffic']],
+  ['trafficlog', ['observability']],
+
+  // testing
+  ['faultinjection', ['testing']],
+  ['meshfaultinjection', ['testing']],
+
+  // advanced
+  ['meshproxypatch', ['advanced']],
+  ['proxytemplate', ['advanced']],
```
johncowen, line 45: Just a comment for the future:

Ideally these need to come from OpenAPI. I believe we have descriptions for these also. Using the `_resources` endpoint for all this stuff is just overloading this endpoint until its pretty much the OpenAPI spec. I'd rather just use the OpenAPI spec.

packages/kuma-gui/src/app/resources/views/ResourceTypeListView.vue
```diff
@@ -36,7 +36,15 @@
...
-                      v-for="options in [['all', 'security', 'traffic', 'observability', 'performance', 'reliability']]"
+                      v-for="options in [[
+                        'all',
+                        'resilience',
+                        'traffic',
+                        'security',
+                        'observability',
+                        'testing',
+                        'advanced',
+                      ]]"
```
johncowen, line 47: Again for circling back and not in this PR. Ideally these need to come from the original data not be hardcoded here.

johncowen (approved):
I left a couple of comments but this LGTM as it stands. Lets wait see a little more incase we can any more feedback on your original question here, but if we don't hear in the next day or so lets go ahead and merge this.

Self-note: This is currently still behind a feature flag

packages/kuma-gui/src/app/resources/data/ResourceTypeDescriptor.ts
schogges: Just for my understanding: are you referring to already existing specs or are you saying that it should be added there?

johncowen: I'm referring to the existing specs in the backend. Every resource should use some sort of `x-field-name` to specify the category (and human description for the resource/component). 

I'm saying they should be added there.

schogges: Gotcha, yeah this needs to be added.

PR #4940, opened 2026-05-19
fix(meshtrust): deprecated spec.origin

### LINKED ISSUE #4884: MeshTrust origin field not displayed (present in API response)
When viewing MeshTrust resources, the "origin" field appears empty in the UI, but it is present in the API response.

Found during Test Friday 2026-04-24.

schogges:
As per OAS the field `spec.origin` of a `MeshTrust` item is deprecated and we should use `status.origin` instead.

Closes https://github.com/kumahq/kuma-gui/issues/4884


johncowen:
Q: Do we need to think about back compat here?

johncowen:
Just linking to the issue (and conversation) we created when I realised what had actually happened here (a field was marked as deprecated but still removed by accident in the backend).

https://github.com/kumahq/kuma/issues/16710

packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
```diff
@@ -102,16 +102,16 @@
...
                             name: 'mesh-mesh-identity-summary-view',
                             params: {
-                              mid: item.spec.origin.kri,
+                              mid: item.origin.kri,
                             },
                           }"
                         >
                           <XBadge appearance="decorative">
-                            {{ item.spec.origin.kri }}
+                            {{ item.origin.kri }}
```
johncowen: How come we aren't using `status.origin`

If I'm not mistaken this is a brand new GUI only field (that we only generally do when there is a reason to provide such alias - like a weird awkward label name or something)

packages/kuma-http-api/mocks/src/meshes/_/meshtrusts/_.ts
```diff
@@ -45,11 +45,13 @@ export default ({ fake }: Dependencies): ResponseHandler => (req) => {
...
-        origin: {
-          kri: fake.kuma.kri({ resourceName: 'MeshIdentity', mesh, namespace, zone }),
-        },
         trustDomain: `${mesh}.${fake.word.noun()}.mesh.local`,
       } satisfies components['schemas']['MeshTrustItem']['spec'],
+      status: {
+        origin: {
+          kri: fake.kuma.kri({ shortName: 'mid', mesh, namespace, zone }),
+        },
+      } satisfies components['schemas']['MeshTrustItem']['status'],
```
johncowen: How come we have 2 `satisfies` in here, could you explain why we can't do `satisfies MeshTrustItem`?

schogges: Some entries are missing: `type` (is currently conditional), `mesh` and `name`

packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
schogges: I pushed it before we synced on this with backend and didn't change it since then.

johncowen: gotcha, what you've got here now looks better now 👍

packages/kuma-http-api/mocks/src/meshes/_/meshtrusts/_.ts
johncowen: Please could we:

- Move inline type to the top of the file with a `type MeshTrustResponse = ...`
- Reference that with `} satisfies MeshTrustResponse` here
- Change the referenced type to a referenced based of the filepath (which corresponds to the HTTP API url) like this https://github.com/kumahq/kuma-gui/blob/ebbe48a1dae12800473c9841db4887cadbfbdcfe/packages/kuma-http-api/mocks/src/mesh-insights/_.ts#L3

If we can we do that here in this PR that would be great 👍 

I have more to say here, but lets concentrate on this PR for the moment.

schogges: Done

packages/kuma-http-api/mocks/src/meshes/_/meshtrusts/_.ts
```diff
@@ -45,11 +49,13 @@ export default ({ fake }: Dependencies): ResponseHandler => (req) => {
           },
           type: 'Pem',
         }],
+        trustDomain: `${mesh}.${fake.word.noun()}.mesh.local`,
+      },
+      status: {
         origin: {
-          kri: fake.kuma.kri({ resourceName: 'MeshIdentity', mesh, namespace, zone }),
+          kri: fake.kuma.kri({ shortName: 'mid', mesh, namespace, zone }),
```
johncowen: I don't think this is important to do here, but if this is a universal (not k8s) install, does this mean that the KRI will have a namespace?

packages/kuma-gui/src/app/meshes/views/MeshDetailView.vue
```diff
@@ -102,16 +102,16 @@
                       </template>
                       <template #origin="{ row: item }">
                         <XAction
-                          v-if="item.spec.origin?.kri?.length > 0"
+                          v-if="item.status.origin.kri.length > 0"
```
johncowen, line 105: Love this change 👍

packages/kuma-http-api/mocks/src/meshes/_/meshtrusts/_.ts
johncowen: I also noticed that the sectionName for a KRI gets filled even though you don't specify one. Am I right in thinking MeshTrusts (and plenty of other things) don't have sectionNames?

schogges: Yeah, most of the resources do not have a sectionName.

```diff
-          kri: fake.kuma.kri({ resourceName: 'MeshIdentity', mesh, namespace, zone }),
+          kri: fake.kuma.kri({ shortName: 'mid', mesh, namespace, zone }),
```

not sure how this diff came in. I'm pretty sure I didn't change `resourceName` to `shortName` 🤔

schogges: > I don't think this is important to do here, but if this is a universal (not k8s) install, does this mean that the KRI will have a namespace?

Yeah, I think we have to generally address the difference of universal vs k8s in the mocks.

packages/kuma-gui/features/mesh/MeshTrust.feature
```diff
@@ -49,7 +49,7 @@ Feature: mesh / mesh-identity
       body:
         items:
           - name: trust-1
-            spec:
+            status:
               origin:
                 kri: kri_mid_default_default_foo_bar_baz
```
johncowen, line 54: Can we take the `sectionName` off these and leave them with the trailing `_`s if thats what they are supposed to look like (?)

schogges: I'd prefer to do this in a follow-up. There might be more cases where this is wrong. These type of changes are not in scope of this PR.

johncowen (approved):
Ok this looks good to go!

PR #4941, opened 2026-05-19
[Experiment]: Use router instead of protocol handler

schogges:
TBD


johncowen:
We are just chatting through this because this was work from a little while ago and we are trying to figure out/remember why this had a lot of routing config work

We are thinking that the other two PRs up are more dealing with "how to render the link to the page", whereas this PR is more about "once you've clicked a link how do we render the page and keep it under the `resources/` URL segment

Not totally sure but as of right now this is what we are thinking

(other related PRs:)

- https://github.com/kumahq/kuma-gui/pull/4825
- https://github.com/kumahq/kuma-gui/pull/4930

johncowen:
We've done a lot of work on routing and KRIs over the past few months, and I have a feeling we can close this as it was a bit of an experiment towards that that we didn't go with, is that right @schogges ?

schogges:
This was just an experiment and I'm closing this now.

PR #4946, opened 2026-05-21
feat: skills for e2e and mocks

schogges:
See title


schogges:
For https://github.com/kumahq/kuma-gui/pull/4946/changes/f4c179b03bcea50760cc5b1bd03e263b77dd7d33 I pulled in the [skill for skills](https://github.com/anthropics/skills/blob/main/skills/skill-creator/SKILL.md), let the LLM iterate on the SKILLs and refined them. I also renamed the `mock` skill to `kuma-http-api-mock`. The `gherkin-playwright` skill should be agnostic to an app and mocks.

johncowen:
> sync: we are very interested in how these get automatically installed (but continue to track the local versions of the files if you make amends). But there nothing stopping us from merging these without an "installation approach" and we just do whatever works for us at the moment.

> we are gonna merge these as it doesn't affect anything just now, but I'll give them a quick review again first (and try them out) just incase I have any immediate comments

johncowen (approved):
I had a fairly good read of these last night, and I was pleased to see stuff in here I wouldn't have thought of adding that I can see how would help getting a more deterministic output.

There are also quite a few things I personally would alter, but I appreciate its probably best to just get these in for the moment, they are currently totally optional to use and in using them we can improve as we go.

I may add some comments into this PR even after merge, just so I have somewhere to write my opinions (they'll mainly be opinions, not requests at the moment and they may well change as I get more of a feel for things)

That being said LGTM! Excited for what's coming!

schogges:
> I had a fairly good read of these last night, and I was pleased to see stuff in here I wouldn't have thought of adding that I can see how would help getting a more deterministic output.
> 
> There are also quite a few things I personally would alter, but I appreciate its probably best to just get these in for the moment, they are currently totally optional to use and in using them we can improve as we go.
> 
> I may add some comments into this PR even after merge, just so I have somewhere to write my opinions (they'll mainly be opinions, not requests at the moment and they may well change as I get more of a feel for things)
> 
> That being said LGTM! Excited for what's coming!

Thanks 🙂 I think it would be good to extract good templates and add them as separate files to the skill folder, instead of the "looking for siblings" 🤔

PR #4955, opened 2026-05-26
refactor(mesh-list-view,mesh-insights-list): show mesh (generic) services count

### LINKED ISSUE #4882: Meshes listing shows old services count
On the meshes listing page, the "services" value still shows old internal mesh services.

Need to do a code search for "internal" - we thought we had them all but some remain.

Meshes services count

Found during Test Friday 2026-04-24.

schogges:
Shows the calculated count of generic mesh-services (`MeshService`, `MeshExternalService`, `MeshMultiZoneService`) instead of old-style internal services.

Closes https://github.com/kumahq/kuma-gui/issues/4882


packages/kuma-gui/src/app/meshes/views/MeshListView.vue
```diff
@@ -96,7 +96,7 @@
                   <template
                     #services="{ row: item }"
                   >
-                    {{ item.services.internal }}
+                    {{ item.resources.MeshServiceGeneric.total }}
```
johncowen: Does this continue to work correctly if you aren't using `mode: Exclusive`?

schogges: I was thinking about it and we should probably calculate the sum of both `item.resources.MeshServiceGeneric.total + item.services.internal` 🤔 wdyt?

johncowen: I think in other places we show either `old`, `new` or `old / new` I don't think we ever sum anyhere? But maybe I'm misremembering

schogges: No, I think you are right that we only show both together but never summed up. I will change it to conditionally show `services.internal` if there are no mesh services.

johncowen: What about if both are non-zero we show `num / num` and otherwise we prefer MeshService and fallback to the old `internal`? I can see it's not a straightforwards as other places where we know the `mode` but from what I remember we chose to have a state where we show both like `num / num`?

schogges: > What about if both are non-zero we show `num / num` and otherwise we prefer MeshService and fallback to the old `internal`? I can see it's not a straightforwards as other places where we know the `mode` but from what I remember we chose to have a state where we show both like `num / num`?

I think it's not easy to understand the `num / num` without further explanation. Would need to add some info icon and provide that information 🤔 

image

johncowen: We could do the same as we currently do for `Data plane proxies (online/total)` in the 3 column?

i.e. 

`Services (MeshServices / Services)`

But only when both the _control-plane_ totals are non-zero.

schogges: Sounds good 👍

schogges: Hmmm 🤔 actually the title of the column would be tricky... We'd always show `Services (MeshServices / Services)`, which might be misleading if there are non or only partial entries that have `num / num`. Probably better would be to just leave the title as is and sum all services (including internal). The user can always dig deeper and find more information about the services.

johncowen: sorry been really busy today, but just to add some notes from our catch up.

I pointed out that we don't usually sum these two things mentioned also in https://github.com/kumahq/kuma-gui/pull/4955#discussion_r3305352819

We outlined several ways we could do this, there are some pros/cons for each option which also depend on whether we are adding this on the control-plane detail page or the mesh listing page.

### Use global insights to decide whether to show `num / num` or just `num`

Further up the page we already check to see if either MeshServices or ServiceInsights are non-zero (from global insights) and act accordingly to show either `num` or `num num`. We could use the same data and approach.

Whilst global insights isn't split per mesh, if we have just one ServiceInsight in global insights we know we need to show `num / num` not just `num`

Cons:

- On the CP detail page we will have to shift some component nesting around (🤷 ) and pass something through to the MeshListing component
- On the Mesh List page, we'll have to make an additional HTTP request to globalInsights as we don't currently do this here.

### Do a pre-loop of the mesh data to figure out if any meshes have service insights to decide whether to show `num / num` or just `num`

Currently we just get the mesh listing data and immediately loop-and-render the table. We could do a first "pre-loop" to look at all meshes and see if any have service insights, if they do show `num / num` otherwise don't.

Pros:
- On the CP page we only ever request your first 10 meshes and don't paginate, so no issues here

Cons:
- On the MeshList view we have pagination

IMO, as the mesh listing page defaults to a "items per page" of 50 and the lowest you can set this is 15. I would say that folks having under 15 meshes is the most common outcome so not paginating to figure out if any meshes have service insights becomes a non issue.

Also IMO, even if someone has 1000 meshes, and as they paginate to a page that happens to have `num / num` and all the rest of the pages just use `num` (i.e. one page happens to have a mesh that has service insights). I don't think this is an issue at all.

---

We also spoke about just always just showing `num / num`. My opinion here is that we are only doing `num / num` to provide backwards compatibility and we want people to upgrade to MeshServices (i.e. `num`). So we shouldn't just show the old thing we don't want to support. We should support the new thing and only backwards compat when necessary.

---

We also think we spotted a couple of backend issues here that we need to confirm before raising with the backend to check whether those things are intentional or not.

schogges: Done

packages/kuma-gui/src/app/meshes/components/MeshInsightsList.vue
johncowen, line 1: Interested to know why you made certain choices here. Could you explain the thinking behind the implementation you chose in this file (and MeshListView) for me?

packages/kuma-gui/src/app/meshes/views/MeshListView.vue
johncowen: Resolving this comment, there is a little more conversation further down on this but its more about implementation

packages/kuma-gui/src/app/meshes/components/MeshInsightsList.vue
johncowen: We had a sync on this and ended up with https://github.com/kumahq/kuma-gui/pull/4955/commits/19740b502f402748eee3f306dc097cbd704de8f6 instead for several reasons

johncowen: Resolving this one also

johncowen (approved):

PR #4956, opened 2026-05-26
chore(dataplanes): remove tags usage for service and protocol

johncowen:
Removes `.service` property entirely, we barely used it (only for a `:key`) and I've replaced it here with the `.name` which is equally as unique.

I also removed `.protocol` which previously was an alias to the `tags['kuma.io/protocol']` tag.

Instead of using the tag we now just look to see if there are stats for `http`, if there are then the protocol is `http`, otherwise its `tcp`. This is the exact same method we use for figuring out protocol for gateways - now we just do the same for "inbounds".

Continuation of https://github.com/kumahq/kuma-gui/issues/4570

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
+++ b/packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
@@ -134,7 +134,7 @@ export const DataplaneNetworking = {
             clusterName: item.servicePort && item.servicePort !== item.port ? `localhost_${item.servicePort}` : name,
             // If a health property is unset the inbound is considered healthy
             state: (typeof item.state !== 'undefined' ? String(item.state) : 'Ready'),
-            protocol: item.protocol ?? '',
+            protocol: item.protocol ?? tags['kuma.io/protocol'] ?? 'tcp',
             address,
             addressPort: `${address}:${item.port}`,
             // inbound serviceAddress, inbound address, networkingAddress because the internal services accessible address
--- a/packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
+++ b/packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
@@ -44,8 +44,8 @@ const DataplaneOutbound = {
     return {
       ...item,
       tags,
-      name: tags['kuma.io/service'],
-      service: tags['kuma.io/service'],
+      // @TODO how do we get the name of the outbound?
+      name: tags['kuma.io/service'] ?? '',
       protocol: tags['kuma.io/protocol'] ?? 'tcp',
       address,
       addressPort: `${address}${typeof item.port === 'number' ? `:${item.port}` : ''}`,
@@ -99,9 +99,8 @@ export const DataplaneNetworking = {
         ? ((tags = {}) => [{
           address: networking.address ?? '',
           tags,
-          name: tags['kuma.io/service'] ?? '',
-          service: tags['kuma.io/service'] ?? '',
-          protocol: tags['kuma.io/protocol'] ?? 'tcp',
+          name: '',
+          protocol: '',
           state: 'Ready',
           // these could be filled out during 'cloning'
           // i.e. these are like template variables to be filled out
@@ -135,8 +134,7 @@ export const DataplaneNetworking = {
             clusterName: item.servicePort && item.servicePort !== item.port ? `localhost_${item.servicePort}` : name,
             // If a health property is unset the inbound is considered healthy
             state: (typeof item.state !== 'undefined' ? String(item.state) : 'Ready'),
-            service: tags['kuma.io/service'] ?? '',
-            protocol: tags['kuma.io/protocol'] ?? 'tcp',
+            protocol: '',
             address,
             addressPort: `${address}:${item.port}`,
             // inbound serviceAddress, inbound address, networkingAddress because the internal services accessible address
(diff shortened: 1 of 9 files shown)


packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -135,8 +134,7 @@ export const DataplaneNetworking = {
             clusterName: item.servicePort && item.servicePort !== item.port ? `localhost_${item.servicePort}` : name,
             // If a health property is unset the inbound is considered healthy
             state: (typeof item.state !== 'undefined' ? String(item.state) : 'Ready'),
-            service: tags['kuma.io/service'] ?? '',
-            protocol: tags['kuma.io/protocol'] ?? 'tcp',
+            protocol: '',
```
schogges: I'm not entirely sure about gateway, but there is a `inbound[].protocol` entry that we should check

johncowen: Oh 🤦 forgot about that, thanks! I'm just triple checking that I don't need to support Dataplanes without this new field, once i"ve done that I'll come back here and tweak 👍

johncowen: K I moved this to use `.protocol`, thanks for the reminder!

packages/kuma-gui/src/app/legacy-data-planes/views/DataPlaneDetailView.vue
```diff
@@ -512,13 +512,14 @@
...
                       }
+                      const keys = Object.keys(value ?? {})
                       return prev.concat([
                         {
                           ...props.data.dataplane.networking.inbounds[0],
                           name: key,
                           clusterName: key,
                           port: Number(port),
-                          protocol: ['http', 'tcp'].find(item => typeof value[item] !== 'undefined') ?? 'tcp',
+                          protocol: keys.includes('grpc') ? 'grpc' : keys.includes('http') ? 'http' : 'tcp',
```
johncowen, line 522: This is probably a little bugfix

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -44,8 +44,8 @@ const DataplaneOutbound = {
     return {
       ...item,
       tags,
-      name: tags['kuma.io/service'],
-      service: tags['kuma.io/service'],
+      // @TODO how do we get the name of the outbound?
+      name: tags['kuma.io/service'] ?? '',
```
schogges, line 48: I _think_ we only get the name of the outbound by looking at the KRIs in the list of outbounds in the response of the layout endpoint 🤔 it seems that unfortunately the information is spread across different sources. Potentially in the future we may be able to just use one source.

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
```diff
@@ -99,9 +99,8 @@ export const DataplaneNetworking = {
         ? ((tags = {}) => [{
           address: networking.address ?? '',
           tags,
-          name: tags['kuma.io/service'] ?? '',
-          service: tags['kuma.io/service'] ?? '',
-          protocol: tags['kuma.io/protocol'] ?? 'tcp',
+          name: '',
+          protocol: '',
```
schogges, line 103: Should we do the same here (protocol of gateways) as we do for standard proxies?

schogges (approved):
LGTM 👍

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
johncowen: I don't think so no.

This code for gateways "acts as a template" and the template always gets filled out/overwritten (in the "legacy" dataplane view anyway):

https://github.com/kumahq/kuma-gui/blob/0a4da9631a01d64028f8fcd36f732074d1ba42c1/packages/kuma-gui/src/app/legacy-data-planes/views/DataPlaneDetailView.vue#L521

I'm not entirely sure what it does when folks are using the new view, but I took a brief look and at initial glance it looks like it uses the dataplaneNeworkingLayout stuff instead. Whichever I can't see the new view using this "template" as an actual "template" and if it is it should already be filled this out like the legacy view does.

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
johncowen: > it seems that unfortunately the information is spread across different sources.

This ^

packages/kuma-gui/src/app/legacy-data-planes/views/DataPlaneDetailView.vue
johncowen: resolving this one

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
schogges: Gateways will never use the new dataplane view, this was one of the restrictions when we added the new view:

https://github.com/kumahq/kuma-gui/blob/0a4da9631a01d64028f8fcd36f732074d1ba42c1/packages/kuma-gui/src/app/data-planes/features.ts#L19

So I think we are fine 👍

johncowen: K resolving this one then

packages/kuma-gui/src/app/data-planes/data/DataplaneNetworking.ts
johncowen: Gonna resolve this one, the TODO was a note for a future PR and still something we need to figure out. BUt generally I get the impression that outbounds support here is less of a priority.

PR #4957, opened 2026-05-26
refactor(mesh-external-service-listing): consider mesh-identity for mtls warning

### LINKED ISSUE #4801: MeshExternalService warning banner even though MeshIdentity is applied
### Description

I'm using mesh without mTLS but with MeshIdentity:

Image

However, Service tab when MeshExternalServices are selected shows warning:

Image

schogges:
We only checked for `mesh.mtlsBackend` when deciding to show the warning about missing mtls. This PR changes it to use `mesh-insight/:name` in the decision making and check wether there is a `mTLS.issuedBackends` entry present.

Closes https://github.com/kumahq/kuma-gui/issues/4801

### DIFF
--- a/packages/kuma-gui/src/app/meshes/data/MeshInsight.ts
+++ b/packages/kuma-gui/src/app/meshes/data/MeshInsight.ts
@@ -40,9 +40,7 @@ export const MeshInsight = {
   fromObject(item: PartialMeshInsight) {
     return {
       ...item,
-      mTLS: {
-        ...item.mTLS,
-      },
+      mTLS: item.mTLS ?? {},
       dataplanes: MeshInsightDataplaneStatistics.fromObject(item.dataplanes ?? {}),
       dataplanesByType: {
         ...item.dataplanesByType,
--- a/packages/kuma-gui/src/app/meshes/data/MeshInsight.ts
+++ b/packages/kuma-gui/src/app/meshes/data/MeshInsight.ts
@@ -40,6 +40,9 @@ export const MeshInsight = {
   fromObject(item: PartialMeshInsight) {
     return {
       ...item,
+      mTLS: {
+        ...item.mTLS,
+      },
       dataplanes: MeshInsightDataplaneStatistics.fromObject(item.dataplanes ?? {}),
       dataplanesByType: {
         ...item.dataplanesByType,
--- a/packages/kuma-gui/src/app/services/views/MeshExternalServiceListView.vue
+++ b/packages/kuma-gui/src/app/services/views/MeshExternalServiceListView.vue
@@ -50,7 +50,7 @@
               v-slot="{ data: [data, meshInsight] }"
             >
               <XNotification
-                :notify="!props.mesh.mtlsBackend && !meshInsight.mTLS"
+                :notify="!meshInsight.mTLS.issuedBackends"
                 :uri="`mes-mtls-warning.${props.mesh.id}`"
               >
                 <XI18n
--- a/packages/kuma-gui/src/app/services/views/MeshExternalServiceListView.vue
+++ b/packages/kuma-gui/src/app/services/views/MeshExternalServiceListView.vue
@@ -14,166 +14,174 @@
       :render="false"
       :title="t(`services.routes.mesh-external-service-list-view.title`)"
     />
-    <AppView
-      :docs="t('services.mesh-external-service.href.docs')"
-      :notifications="true"
+    <DataSource
+      :src="uri(meshSources, '/mesh-insights/:name', {
+        name: route.params.mesh,
+      })"
+      v-slot="{ data: sourceMeshInsight }"
     >
-      <XCard>
-        <XLayout variant="y-stack">
-          <search>
-            <form
-              @submit.prevent
+      <AppView
+        :docs="t('services.mesh-external-service.href.docs')"
+        :notifications="true"
+      >
+        <XCard>
+          <XLayout variant="y-stack">
+            <search>
+              <form
+                @submit.prevent
+              >
+                <XSearch
+                  :keys="['name', 'namespace', ...(can('use zones') ? ['zone'] : []), 'label']"
+                  :value="route.params.s"
+                  @change="(s) => route.update({ s })"
+                />
+              </form>
+            </search>
+            <DataLoader
+              :src="uri(sources, '/meshes/:mesh/mesh-external-services', {
+                mesh: route.params.mesh,
+              },{
+                page: route.params.page,
+                size: route.params.size,
+                search: route.params.s,
+              })"
+              :data="[sourceMeshInsight]"
+              variant="list"
+              v-slot="{ data: [data, meshInsight] }"
             >
-              <XSearch
-                :keys="['name', 'namespace', ...(can('use zones') ? ['zone'] : []), 'label']"
-                :value="route.params.s"
-                @change="(s) => route.update({ s })"
-              />
-            </form>
-          </search>
-          <DataLoader
-            :src="uri(sources, '/meshes/:mesh/mesh-external-services', {
-              mesh: route.params.mesh,
-            },{
-              page: route.params.page,
-              size: route.params.size,
-              search: route.params.s,
-            })"
-            variant="list"
-            v-slot="{ data: [data] }"
-          >
-            <XNotification
-              :notify="!props.mesh.mtlsBackend"
-              :uri="`mes-mtls-warning.${props.mesh.id}`"
-            >
-              <XI18n
-                path="services.mesh-external-service.notifications.mtls-warning"
-              />
-            </XNotification>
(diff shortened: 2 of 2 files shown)


johncowen:
From the issue:

> I'm using mesh without mTLS but with MeshIdentity:

and the title of this PR says:

> consider mesh-identity for mtls warning

but then the description says:

> This PR adds mesh-insight/:name to the decision and check wether there is a mTLS entry present.

question: Is the presence of an mTLS property on `mesh-insight/:name` enough to know whether you have set up mTLS using MeshTrust/MeshIdentity? If we are using "mesh without mTLS", and "considering mesh-identity" how come we are checking for the mTLS property on mesh-insights?

schogges:
> From the issue:
> 
> > I'm using mesh without mTLS but with MeshIdentity:
> 
> and the title of this PR says:
> 
> > consider mesh-identity for mtls warning
> 
> but then the description says:
> 
> > This PR adds mesh-insight/:name to the decision and check wether there is a mTLS entry present.
> 
> question: Is the presence of an mTLS property on `mesh-insight/:name` enough to know whether you have set up mTLS using MeshTrust/MeshIdentity? If we are using "mesh without mTLS", and "considering mesh-identity" how come we are checking for the mTLS property on mesh-insights?

In my local deployment I have no mtls configured on the mesh resource, but I'm using MeshIdentity. I have the following mtls entry in the response of mesh-insight:

```json
 "mTLS": {
  "issuedBackends": {
   "kri_mid_default__kuma-system_identity-kv_": {
    "total": 2,
    "online": 2
   }
  },
  "supportedBackends": {
   "kri_mtrust_default_zone-1_kuma-system_identity-244vffwwb42w64c8_": {
    "total": 1,
    "online": 1
   },
   "kri_mtrust_default_zone-1_kuma-system_identity-kv-7bvv9d7bxf4c6z46_": {
    "total": 1,
    "online": 1
   },
   "kri_mtrust_default_zone-1_kuma-system_trust-zone-2_": {
    "total": 1,
    "online": 1
   },
   "kri_mtrust_default_zone-2_kuma-system_identity-244vffwwb42w64c8_": {
    "total": 1,
    "online": 1
   },
   "kri_mtrust_default_zone-2_kuma-system_identity-kv-7bvv9d7bxf4c6z46_": {
    "total": 1,
    "online": 1
   },
   "kri_mtrust_default_zone-2_kuma-system_trust-zone-1_": {
    "total": 1,
    "online": 1
   }
  }
 },
```

Do you think this is enough information to say that mtls is configured on the mesh?

johncowen:
I think there are a few things here:

- We've used `!meshInsight.mTLS` but I _think_ by default mTLS is `{}` so I think at the very least we need to inspect inside of `meshInsight.mTLS`.
- I'm also thinking about whether things being in here actually means that mTLS is properly enabled. For example I'm pretty sure I've managed to badly setup MeshIdentity and there are entries in `mTLS` even though things aren't working correctly.

The other thing I'm thinking is:

Maybe we should move these notifications into the top-level i.e. above the tabs, especially considering they apply to the Mesh as a whole and we possibly want to show them no matter which tab you are in.

This last thing is a bigger lift, so we should think about it a little first. 

---

🤔 I guess the most important thing rn is the comment on `!meshInsight.mTLS`

schogges:
> I think there are a few things here:
> 
> * We've used `!meshInsight.mTLS` but I _think_ by default mTLS is `{}` so I think at the very least we need to inspect inside of `meshInsight.mTLS`.
> * I'm also thinking about whether things being in here actually means that mTLS is properly enabled. For example I'm pretty sure I've managed to badly setup MeshIdentity and there are entries in `mTLS` even though things aren't working correctly.

You are right, by default `mTLS` is `{}` and I'll change it. I'm actually thinking we can only check `mesh-insights/:mesh` for `mTLS.issuedBackends`. I've quickly checked how it looks like with configuring `mTLS` on the mesh:

```json
 "mTLS": {
  "issuedBackends": {
   "ca-1": {
    "total": 2,
    "online": 2
   }
  },
  "supportedBackends": {
   "ca-1": {
    "total": 2,
    "online": 2
   }
  }
 },
```

So whenever there is an entry of `issuedBackends` we can consider that `mTLS` is configured.

To the second point: how do we know that `mTLS` is configured/enabled properly? Did we know from just looking at `mesh.mTLS`? Not sure how much information the `issuedBackends.[key].online` gives us in that regards. But from just looking at the object we know it's configured, but we don't know if it's setup correctly.

> The other thing I'm thinking is:
> 
> Maybe we should move these notifications into the top-level i.e. above the tabs, especially considering they apply to the Mesh as a whole and we possibly want to show them no matter which tab you are in.
> 
> This last thing is a bigger lift, so we should think about it a little first.

That would a good follow-up 👍

johncowen:
Yeah I think you are right on the `issuedBackends` thing, I had a quick look here:

https://github.com/kumahq/kuma/blob/d3b4948eca41ab1cdd665f012727c2f3fa7ba95a/api/mesh/v1alpha1/mesh_insight.proto

Looks like the contents of each `issuedBackends` is just stats to tell us which Dataplanes using these "mTLS-things" are `online/offline`. I guess the fact that a Dataplane is trying to use an "mTLS-thing" whether its online or not tells us its configured (perhaps incorrectly, but configured).

---

Lets do the above (base it on `issuedBackends`) for now. I was looking at it in more depth this morning and I think we should take another look at all of this from a higher level - I definitely managed to (pretty easily) get this setup wrong and and I could only tell it was wrong by looking at the `status` of the things in the YAML window in the GUI.

schogges:
> Yeah I think you are right on the `issuedBackends` thing, I had a quick look here:
> 
> https://github.com/kumahq/kuma/blob/d3b4948eca41ab1cdd665f012727c2f3fa7ba95a/api/mesh/v1alpha1/mesh_insight.proto
> 
> Looks like the contents of each `issuedBackends` is just stats to tell us which Dataplanes using these "mTLS-things" are `online/offline`. I guess the fact that a Dataplane is trying to use an "mTLS-thing" whether its online or not tells us its configured (perhaps incorrectly, but configured).
> 
> Lets do the above (base it on `issuedBackends`) for now. I was looking at it in more depth this morning and I think we should take another look at all of this from a higher level - I definitely managed to (pretty easily) get this setup wrong and and I could only tell it was wrong by looking at the `status` of the things in the YAML window in the GUI.

Alright, I'll apply the changes 🙂 brb

packages/kuma-gui/src/app/meshes/data/MeshInsight.ts
```diff
@@ -40,6 +40,9 @@ export const MeshInsight = {
   fromObject(item: PartialMeshInsight) {
     return {
       ...item,
+      mTLS: {
+        ...item.mTLS,
+      },
```
johncowen: We don't need this right?

packages/kuma-gui/src/app/services/views/MeshExternalServiceListView.vue
```diff
@@ -23,175 +23,182 @@
...
-                  data-testid="service-collection"
-                  :headers="[
-                    { ...me.get('headers.name'), label: 'Name', key: 'name' },
-                    { ...me.get('headers.namespace'), label: 'Namespace', key: 'namespace' },
-                    ...(can('use zones') ? [{ ...me.get('headers.zone'), label: 'Zone', key: 'zone' }] : []),
-                    { ...me.get('headers.port'), label: 'Port', key: 'port' },
-                    { ...me.get('headers.actions'), label: 'Actions', key: 'actions', hideLabel: true },
-                  ]"
+                <XNotification
+                  :notify="typeof meshInsight.mTLS.issuedBackends === 'undefined'"
```
johncowen: Do we need to maintain the old functionality?

johncowen (commented):
Left a couple of qs

packages/kuma-gui/src/app/meshes/data/MeshInsight.ts
schogges: It saves us a conditional chaining when trying to access `meshInsight.mTLS.issuedBackends` as `meshInsight.mTLS` is potentially `undefined`

packages/kuma-gui/src/app/services/views/MeshExternalServiceListView.vue
schogges: I don't think so

johncowen: This covers both cases because the `mTLS.issuedBackends` field is populated whether you are using the "old school" approach or the "new school" approach. Resolving this

johncowen (dismissed):
LGTM!

schogges:
@johncowen had to resolve conflicts, need another approval

johncowen (approved):

PR #4958, opened 2026-05-27
chore(mesh-identity): data cleanup for MeshIdentity and MeshTrust

johncowen:
Whilst looking at/investigating https://github.com/kumahq/kuma-gui/pull/4957 I noticed there were a few things I could/should amend in the MeshTrust/MeshIdentity data layer.

- Use display-name for the name
- Add an `id` (we don't need it for these resources because these use kri's but added for consistency)
- include mesh, zone, namespace and use the display-name for the default KRI (again highly unlikely this is ever used)
- default `total` in the collection
- add a bunch of "aliases" that we use for other resources

The only user facing amend here is that we no longer show `name.namespace` in the GUI for titles (and badges), instead we just show `name` (minus the `.namespace`)

---

I also then noticed that the mocks where missing quite a lot of information such as:

- KRIs
- paging (plus totals etc)
- status
- displayName labels

So I fixed that up also and updated the necessary tests.

### DIFF
--- a/packages/kuma-gui/src/app/mesh-identities/data/MeshIdentity.ts
+++ b/packages/kuma-gui/src/app/mesh-identities/data/MeshIdentity.ts
@@ -14,15 +14,13 @@ export const MeshIdentity = {
     const name = labels['kuma.io/display-name'] ?? item.name
 
     return {
-      // defaults
-      kri: Kri.toString({ shortName: 'mid', mesh, zone, namespace, name }),
-      creationTime: '',
-      modificationTime: '',
-      labels,
-      // overwrite
       ...item,
+      kri: item.kri ?? Kri.toString({ shortName: 'mid', mesh, zone, namespace, name }),
       name,
       mesh,
+      labels,
+      creationTime: item.creationTime ?? '',
+      modificationTime: item.modificationTime ?? '',
       // aliases
       id,
       namespace,
--- a/packages/kuma-gui/src/app/mesh-identities/data/MeshIdentity.ts
+++ b/packages/kuma-gui/src/app/mesh-identities/data/MeshIdentity.ts
@@ -6,17 +6,38 @@ export type KumaMeshIdentity = components['schemas']['MeshIdentityItem']
 
 export const MeshIdentity = {
   fromObject: (item: KumaMeshIdentity) => {
+    const labels = item.labels ?? {}
+    const id = item.name
+    const mesh = item.mesh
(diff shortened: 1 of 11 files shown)


schogges (dismissed):
LGTM 👍

packages/kuma-gui/src/app/mesh-identities/data/MeshIdentity.ts
```diff
@@ -6,17 +6,38 @@ export type KumaMeshIdentity = components['schemas']['MeshIdentityItem']
...
+    const zone = labels['kuma.io/origin'] === 'zone' && labels['kuma.io/zone'] ? labels['kuma.io/zone'] : ''
+    const namespace = labels['k8s.kuma.io/namespace'] ?? ''
+    const name = labels['kuma.io/display-name'] ?? item.name
+
     return {
-      kri: Kri.toString({ shortName: 'mid', mesh: item.mesh, name: item.name}),
+      // defaults
+      kri: Kri.toString({ shortName: 'mid', mesh, zone, namespace, name }),
+      creationTime: '',
+      modificationTime: '',
```
johncowen: Sorry for the late thought here, I thought last night (😅 ) that I want to change this slightly. Yesterday I thought this was quite a nice way to set the defaults, but now I've noticed that the "older way" is safer I think, so I'm gonna change this. I'll need another approval once I've done that, I'll ping once I've done it.

schogges: I also thought it's a nice way to set the defaults 😄 what exactly do you think is unsafe? The `creationTime` and `modificationTime`?

johncowen: Yeah thats what I thought yesterday also 😄 

I guess if we use this approach "generically" then if something turned out to be `null` or `undefined` as a value then the default `''` value would be overwritten. Even thinking that made me wonder if I'm being swayed by the "this is nice!" thinking, whereas the older approach I know works and it's less likely that I've not thought of something i.e. safer.

schogges: Yeah that's true if something is explicitely set to `undefined` or `null` (the entry is present, but the value is not), then it would override the default. Good catch 👍

johncowen: K I made the change for this in a further commit, lmk on that. Will need another approval 👍

schogges (approved):
LGTM

PR #4959, opened 2026-05-27
chore(dataplanes): add labels to summary view

johncowen:
This PR does two things:

- Adds a labels section to the DataplaneSummaryView
- Adds a conditional check to any Dataplane summary type views for tags (if there are none we don't show the section)


schogges (approved):
LGTM 👍

johncowen:
FYI: I'd really like to get https://github.com/kumahq/kuma-gui/pull/4930 ontop of here so we can link the labels to the relevant detail pages (similiar to how tag linking currently works). I need to double check it myself also but I'd like to get that one in pretty soon if possible so I can get the usage onto the summary view labels also. 🤔 actually I guess I can rebase that PR and add the links to the summary labels also 👀

PR #4965, opened 2026-05-28
fix(workloads): don't render a zone link if zones aren't enabled

johncowen:
## Before

Screenshot 2026-05-28 at 15 07 13

## After

Screenshot 2026-05-28 at 15 08 07

We noticed this is also not working properly in the new resources area so we just added a commit on the end of here to fix that also.


schogges (approved):
Nice catch 👍

PR #4966, opened 2026-05-29
feat(dataplanes): update threshold for cert expires soon warning

schogges:
As the default certificate expiry is currently set to 1 day, the GUI would always show a warning that the certificate expires soon when not increasing the expiration time manually. For now we've decided to change the threshold to show the warning from 1 week to 6 hours. After the backend has changed the default expiration time, the GUI may change aswell again.

Potentially it would be good to change the threshold to a relative number, like a percentage of the difference between `mTLS.certificateExpirationTime` and `mTLS.lastCertificateRegeneration` and max 7 days before 🤔 

Closes https://github.com/kumahq/kuma-gui/issues/4962


johncowen (approved):
👍

PR #4980, opened 2026-06-02
refactor(mesh): make routes injectable

schogges:
See title

---

There are some more dependencies left-over in routes, such as `meshIdentity.summary` which I will address in follow-ups.

### DIFF
--- a/packages/kuma-gui/src/app/data-planes/index.ts
+++ b/packages/kuma-gui/src/app/data-planes/index.ts
@@ -14,13 +14,14 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
     // TODO: Uncomment once the legacy routes are removed
     // [token('data-planes.routes'), {
     //   service: () => {
+    //     const _routes = routes()
     //     return [
     //       (item: RouteRecordRaw) => {
     //         if (item.name === 'mesh-detail-tabs-view') {
-    //           item.children = (item.children ?? []).concat(routes().items())
+    //           item.children = (item.children ?? []).concat(_routes.items())
     //         }
     //         if(item.name === 'mesh') {
-    //           item.children = (item.children ?? []).concat(routes().item())
+    //           item.children = (item.children ?? []).concat(_routes.item())
     //         }
     //       },
     //     ]
--- a/packages/kuma-gui/src/app/data-planes/index.ts
+++ b/packages/kuma-gui/src/app/data-planes/index.ts
@@ -14,10 +14,19 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
     // TODO: Uncomment once the legacy routes are removed
     // [token('data-planes.routes'), {
     //   service: () => {
-    //     return [routes()]
+    //     return [
+    //       (item: RouteRecordRaw) => {
+    //         if (item.name === 'mesh-detail-tabs-view') {
+    //           item.children = (item.children ?? []).concat(routes().items())
+    //         }
+    //         if(item.name === 'mesh') {
+    //           item.children = (item.children ?? []).concat(routes().item())
+    //         }
+    //       },
+    //     ]
     //   },
     //   labels: [
-    //     app.routes,
+    //     app.routeWalkers,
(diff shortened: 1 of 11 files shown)


packages/kuma-gui/src/main.ts
```diff
@@ -33,6 +44,17 @@ async function mountVueApplication() {
...
+    externalServices($),
+    gateways($),
+    dataplanes($),
+    legacyDataplanes($),
+    workloads($),
+    policies($),
+    resources($),
+    rules($),
+    meshIdentities($),
+    meshTrusts($),
```
johncowen, line 57: :awesome_cylon: This looks perfect! 👏 Exactly what we want. Let me take a closer look at the PR in general 👀 but this is looking good!

packages/kuma-gui/src/app/data-planes/index.ts
```diff
@@ -14,10 +14,19 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
...
+    //           item.children = (item.children ?? []).concat(routes().item())
+    //         }
+    //       },
+    //     ]
     //   },
     //   labels: [
-    //     app.routes,
+    //     app.routeWalkers,
     //   ],
     // }],
```
johncowen, line 32: I could be wrong, but I have a feeling that we are never going to remove the legacy routes from 2.14

packages/kuma-gui/src/app/meshes/routes.ts
johncowen, line 1: I noticed we still have references to meshTrust and meshIdentity in here, how easy it is to move those in this PR?

johncowen: Also I "think" but I'm not totally sure tat we don't need the SplitRecordRaw types anymore? Maybe not, but 100% sure.

packages/kuma-gui/src/app/services/index.ts
```diff
@@ -22,13 +23,22 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
...
     [token('services.routes'), {
       service: (can: Can) => {
-        return [routes(can)]
+        return [
+          (item: RouteRecordRaw) => {
+            if (item.name === 'mesh-detail-tabs-view') {
+              item.children = (item.children ?? []).concat(routes(can).items())
+            }
+            if(item.name === 'mesh') {
+              item.children = (item.children ?? []).concat(routes(can).item())
```
johncowen: nit: Here (and possibly other places, but I want make another inline if I see), I think we can call `routes(can)` once the line before the return statement on line 25/26.

packages/kuma-gui/src/main.ts
```diff
@@ -33,6 +44,17 @@ async function mountVueApplication() {
       ...$,
       routes: $.routesLabel,
     }),
+    services($),
+    externalServices($),
+    gateways($),
```
johncowen, line 49: I love that we can just comment this `gateways` line out now and thats `gateways` completely gone 🎉

johncowen (commented):
Left a few little things, mainly a little bit more cleanup and a nit or 2. Lmk!

packages/kuma-gui/src/app/data-planes/index.ts
schogges: Yeah, that might be and we need to think of a good solution. For now I just made the same amends like in the other places.

packages/kuma-gui/src/app/meshes/routes.ts
schogges: Oh I should have mentioned in the PR description that I will cover these in follow-ups. There are more dependencies in routes I want to address.

johncowen: Gotcha, sounds good to me!

packages/kuma-gui/src/app/services/index.ts
schogges: Done

johncowen (approved):
Nice 👍

Issue #4972, opened 2026-06-01
Improve linking from KRIs


schogges:
### Description

Follow-up of https://github.com/kumahq/kuma-gui/pull/4930

With https://github.com/kumahq/kuma-gui/pull/4256 we've added a protocolHandler and a custom `kri` protocol to handle links with `href` of `kri://...`. This included the registration of an event listener on the `document.body` element during the initialization of the app. We've not added a removal of that event listener, which can introduce memory leaks.
With https://github.com/kumahq/kuma-gui/pull/4930 the event handler is moved into `XAction` so we can easily maintain the lifetime of the event listener with the lifecycle of the component. This solves the issue around potential memory leaks, but introduces another issue. The event handler is now tightly coupled with `XAction` and this means that there must always be at least one `XAction` on the page to make this work. This might break if we need to provide a KRI link via locales.

General requirements for linking from KRIs:
- Providing string based KRIs as `href` with `kri` protocol to `XAction` needs to trigger the KRI protocol handler.
- Clicks on a wrapping element needs to trigger a link inside (basically keep the functionality of `findAnchor`).
- Decouple `XAction` from global event handling, all of this needs to work without a `XAction` on the page.
- KRI links are internal links and should not use the browser API but vue router history (no page refresh). We also check for the attribute `x-internal`.
- Currently our KRI links link to detail pages, but we also need to support linking to summary views or other types of nested views

PR #4981, opened 2026-06-02
chore(kri): finishing touches to KRI label links and rollout to details

johncowen:
This PR is unfortunately a little larger than I would like, but I'm half putting the finishing touches to the KRI label links whilst half rolling it out elsewhere.

### Finishing touches

- Replicate font `regular:bold` style of the exiting tag component
- Allow linking by old style `kuma.io/service` to support the old tags
- Replicate coloring of old tags (blue kuma links, grey non-kuma links)
- Regex helper (`r`, like `t`) so we don't need to use `<template>` for Regexp precompiling.

### Rolling out

- Attempts to find and rollout the new pattern to all detail pages we have either labels or tags, using the same pattern for both.
- I mainly searched for `TagList` and replaced this component with the new pattern apart from couple of areas which are almost "dead areas" which we barely touch. Therefore I mostly removed TagList in all detail and summary routes.

I think I have one more PR to make here as I noticed a lot of places where there are no labels in summary panels and there probably should be. I plan on one last follow up PR adding missing labels to summary panels (and if I spot anymore in detail views I will add those in the follow up also)

Lastly, a lot of this has been explained in a sync but I welcome any extra questions here, I can also add some inlines here if more explanation is needed.

### DIFF
--- a/packages/kuma-gui/src/app/application/index.ts
+++ b/packages/kuma-gui/src/app/application/index.ts
@@ -84,6 +84,7 @@ const $ = {
   errorHandler: token<(e: Error) => void>('application.error.handler'),
 
   i18n: token<{t: ReturnType<typeof I18n>['t']}>('i18n'),
+  regexp: token<{r: (str: string) => RegExp}>('i18n'),
   enUs: token('i18n.locale.enUs'),
 
   storage: token<ReturnType<typeof storage>>('application.storage'),
@@ -131,10 +132,10 @@ const addRouteName = (item: RouteRecordRaw) => {
 export const services = (app: Record<string, Token>): ServiceDefinition[] => {
   return [
     [token('application.plugins'), {
-      service: (dataSourcePool, can, env, i18n) => {
+      service: (dataSourcePool, can, env, i18n, regexp) => {
         return [
           [Data, { dataSourcePool }],
-          [Routing, { can, env, i18n }],
+          [Routing, { can, env, i18n, regexp}],
         ]
       },
       arguments: [
@@ -142,6 +143,7 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
         $.can,
         $.env,
         $.i18n,
+        $.regexp,
       ],
       labels: [
         app.plugins,
@@ -203,6 +205,16 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
         $.enUs,
         app.env,
       ],
+    }],
+    [$.regexp, {
+      service: () => {
+        // @TODO regexps will move to locales/yaml
+        const re = new Map<string, RegExp>([
+          ['kuma.label', /^(.+\.)?kuma\.io\//],
+        ])
+        //
+        return { r: (str: string) => re.has(str) ? re.get(str) : new RegExp('') }
+      },
     }],
 
     [$.fetch, {
@@ -275,11 +287,13 @@ export const [
   useEnv,
   useCan,
   useI18n,
+  useRegExp,
   useDataEmptyState,
 ] = createInjections(
   $.env,
   $.can,
   $.i18n,
+  $.regexp,
   $.DataEmptyState,
 )
 export { YAML, get } from './utilities'
--- a/packages/kuma-gui/src/app/application/index.ts
+++ b/packages/kuma-gui/src/app/application/index.ts
@@ -210,6 +210,9 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
       service: () => {
         // @TODO regexps will move to locales/yaml
         const re = new Map<string, RegExp>([
+          // NOTE: regexp for kuma labels only
+          // we don't use this for matching URLs
+          // only to color labels grey or blue (currently)
           ['kuma.label', /^(.+\.)?kuma\.io\//],
         ])
         //
--- a/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
+++ b/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
@@ -7,7 +7,7 @@
       proxyType: '',
     }"
     name="data-plane-detail-view"
-    v-slot="{ route, t, can, uri }"
+    v-slot="{ route, t, can, uri, r }"
   >
     <DataSource
       :src="uri(connectionSources, '/connections/stats/for/:proxyType/:name/:mesh/:socketAddress', {
@@ -217,6 +217,32 @@
                         />
                       </dd>
                     </div>
+                    <div v-if="Object.keys(props.data.labels).length > 0">
+                      <dt>{{ t('data-planes.routes.item.labels') }}</dt>
+                      <dd>
+                        <XLayout
+                          variant="separated"
+                        >
+                          <XAction
+                            v-for="(value, key) in props.data.labels"
+                            :key="key"
+                            :href="t(`common.label.href.${key.replaceAll('.', '~')}`, {
+                              mesh: props.data.mesh,
+                              zone: props.data.zone,
+                              namespace: props.data.namespace,
+                              name: value,
+                            }, { defaultMessage: '' })"
+                          >
+                            <XBadge
+                              :variant="r('kuma.label').test(key) ? 'kuma-label' : 'label'"
+                            >
+                              {{ key }}:<strong>{{ value }}</strong>
+                            </XBadge>
+                          </XAction>
+                        </XLayout>
+                      </dd>
+                    </div>
+                    <!-- @TODO gateways don't use this view at the moment, but we want move away from the legacy view -->
                     <div
                       v-if="props.data.dataplane.networking.gateway"
                     >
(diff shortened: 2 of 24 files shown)


packages/kuma-gui/src/app/application/index.ts
```diff
@@ -204,6 +206,16 @@
         app.env,
       ],
     }],
+    [$.regexp, {
+      service: () => {
+        // @TODO regexps will move to locales/yaml
+        const re = new Map<string, RegExp>([
+          ['kuma.label', /^(.+\.)?kuma\.io\//],
```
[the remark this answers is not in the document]
johncowen: I added a comment in the code that explains that this is not actually regexing against URLs

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
```diff
@@ -733,7 +743,7 @@
                                       </template>
 
                                       <template #body>
-                                        <XBadge variant="decorative">
+                                        <XBadge appearance="decorative">
```
johncowen, line 746: This little mistake is on `master` currently, but I fixed it here because it was blocking the lint tests here.

schogges: Oh wow, how did this not been catched?

packages/kuma-gui/src/app/service-mesh/index.ts
```diff
@@ -46,6 +47,13 @@ const protocolHandler = (can: Can) => {
                   wl: Kri.toString({ shortName, mesh, zone, namespace, name }),
                 },
               }
+            case shortName === '~port':
+              return {
+                name: 'data-plane-list-view',
+                query: {
+                  s: `tag:service:${name}`,
+                },
+              }
```
schogges: To me this case looks very weird. Just by saying that when the `shortName` is `~port` we always want to navigate to the dp listing with a search query.

packages/x/src/components/x-badge/XBadge.vue
```diff
@@ -15,11 +17,36 @@
...
 </script>
+<style lang="scss" scoped>
+.variant-label, .variant-kuma-label {
+  &.k-badge {
+    font-weight: var(--x-font-weight-regular);
+  }
+  :slotted(strong) {
+    font-weight: var(--x-font-weight-semibold);
+  }
+}
```
schogges: I think we can omit the variant classes from the selector and make this more general, i.e. whenever there is a `strong` we set the font-weight to regular except for the strong.

```css
.k-badge {
  :has(strong)  {
    font-weight: var(--x-font-weight-regular);

    :slotted(strong) {
      font-weight: var(--x-font-weight-semibold);
    }
  }
}
```

packages/kuma-gui/src/app/service-mesh/index.ts
johncowen: > Just by saying that when the shortName is ~port we always want to navigate to the dp listing with a search query.

This is exactly what we want.

We only use `~port` for old style `kuma.io/service` tags. For old style kuma.io/services there is no detail page.

Therefore we take you to the dataplane listing filtered by `kuma.io/service` which gives you a list of dataplanes for the old style `kuma.io/service`

packages/x/src/components/x-badge/XBadge.vue
johncowen: Our label (maybe better said `kv`) styling is an outlier, usually we shouldn't break the design system styling which says that _all_ badges should use bold text.

I wouldn't want people (or LLMs) styling other things that aren't `kv`s just by adding a strong tag.

I like that you specifically have to say "this badge is a KV/label" with a `variant="label/kv"` to get it to display like our outlier styling, it's much more semantic i.e. describing what the thing is.

---

Separately we thought it would be a good idea to use `:variant` as much as possible in our component system (`x`) and that variants should ideally be semantic. Moving forwards I'm considering adding more semantic variants to badges so that we no longer use things like `appearance='decorative'` and instead use things like `variant="kv"` variant="protocol" so that the variant describes what we are wrapping - like semantic HTML.

Please let me know what you think of that reasoning, if you have concerns still I'd like to try and address them

---

Just to note, I figured this morning I could change the variant to be `kv` (for kuma labels) and `user-kv` (for non-kuma labels), which just by tweaking the wording here means that `x` stays generic - if it were ever the case that other apps were to use `x` then having `kv` and `user-kv` variants are totally valid and generic.

I'd like to make this change shortly unless you have concerns still (see above)

packages/kuma-gui/src/app/service-mesh/index.ts
schogges: From the perspective of where it's being set and used it might make sense. But from the perspective of the protocol handler I don't see the connection between this shortName and the view

packages/x/src/components/x-badge/XBadge.vue
schogges: > I wouldn't want people (or LLMs) styling other things that aren't kvs just by adding a strong tag.

Fair enough

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: I guess because `variant` could be a HTML attribute

packages/kuma-gui/src/app/service-mesh/index.ts
johncowen: If `kuma.io/service` is only meaningful on old style services (I believe this is the case since we have MeshServices now). There is no detail page for an old style service, there never has been, because an old style `kuma.io/service` is essentially a list of dataplanes all tagged with the same `kuma.io/service`.

If you look in the locales file where I have the "label/tag to shortName" mapping I map the `kuma.io/service` tag to the shortName `~port`.

As there is no detail page for old style services (services tagged with `kuma.io/service`) where else would we link to apart from a list of dataplanes all tagged with the same `kuma.io/service`?

packages/x/src/components/x-badge/XBadge.vue
johncowen: Ok I'm going to change the variants to be called `kv` and `user-kv` which is more generic (and also handily matches our `XTable variant="kv"` which make me think even more that this is a good decision)

johncowen: Changed my mind again slightly on naming see last commit

packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue
johncowen: resolving this one

packages/kuma-gui/src/app/service-mesh/index.ts
schogges: > If `kuma.io/service` is only meaningful on old style services (I believe this is the case since we have MeshServices now). There is no detail page for an old style service, there never has been, because an old style `kuma.io/service` is essentially a list of dataplanes all tagged with the same `kuma.io/service`.

I'm not sure I follow this 🤔 isn't https://github.com/kumahq/kuma-gui/blob/master/packages/kuma-gui/src/app/services/views/ServiceDetailView.vue the detail view of an old style (internal) service?

---

> If you look in the locales file where I have the "label/tag to shortName" mapping I map the `kuma.io/service` tag to the shortName `~port`.

As I said:

> From the perspective of where it's being set and used it might make sense.

In the protocol handler though there is no clear connection to these locales, so to me it looks very confusing to link from a custom `~port` shortName to the dp listing. Something like `kri://kri_dp_{mesh}_{zone}_{namespace}__?s={value}` would make more sense. Due to the absence of the name it can't link to a single DP but a listing and the query parameters can just be appended. Just as an example.

packages/x/src/components/x-badge/XBadge.vue
```diff
@@ -22,7 +22,7 @@ import { inject, computed } from 'vue'
 import type { BadgeAppearance } from '@kong/kongponents'
 const props = withDefaults(defineProps<{
   maxWidth?: string
-  variant?: 'label' | 'kuma-label' | 'default'
+  variant?: 'reserved-kv' | 'kv' | 'default'
   appearance?: BadgeAppearance | string
 }>(), {
```
schogges: There is this linter warning ☝️ "Prop 'appearance' requires default value to be set"

johncowen: Resolving see https://github.com/kumahq/kuma-gui/pull/4981/changes/3d241ac60a621d549d20d20a7d9d5de0e2e42899

packages/kuma-gui/src/app/service-mesh/index.ts
johncowen: Oh sorry I slightly misremembered here, its not _exactly_ that there isn't a detail page for an old style service, it's that there is no _single_ detail page that we can link to for an old style service. This is because given a `kuma.io/service: something_9090` tag, we don't know whether its an Internal service an External service, or a MeshGateway.

See https://github.com/kumahq/kuma-gui/pull/2188 and more specifically https://github.com/kumahq/kuma-gui/pull/2188#issuecomment-1946173768 where we fixed an issue where we were linking to the internal service detail page for external services and meshgateways.

---


> so to me it looks very confusing to link from a custom ~port shortName to the dp listing.

These are all dps that are "advertising a service called <service-name> specifically listening on port <port-number>"

> Something like kri://kri_dp_{mesh}_{zone}_{namespace}__?s={value} would make more sense

This doesn't make sense because that would be a KRI for a single dataplane proxy (a `dp` shortName), not a grouping of dataplane proxies (grouped by servicename/port) a.k.a. a `kuma.io/service`.

Maybe its the `~port` word that is your issue? What if I gave this a `~service` shortName instead. Whilst I'd rather our own custom `~` prefixed shortName was as descriptive as possible I really don't mind what word we use seeing as we have full control over it.

Lmk!

johncowen:
Just to note: ideally this would target 2.1.4

packages/kuma-gui/src/app/service-mesh/index.ts
schogges: If a KRI is pointing to single resources, maybe a KRI doesn't make sense here? I'll approve this now so we can get it in for 2.14, but I think we should address this, maybe through https://github.com/kumahq/kuma-gui/issues/4972

schogges (dismissed):

johncowen:
FYI: I had to repoint the branch to 2.14 here, I decided I may aswell change `~port` to be `~hostport` which describes the new shortName for legacy `kuma.io/service`s slightly better.

@schogges sorry this will need one more approval

schogges (approved):

PR #4982, opened 2026-06-02
fix(dataplanes): don't show static outbounds if there aren't any

johncowen:
Generally its less often that users define outbounds in their configs.

Moreover, we usually display outbounds based on Envoy stats, not actual user defined YAMLs/configs (these can show the _result of_ user defined configs)

I noticed that when `outbounds` are not user defined, we just show an empty state, with no title (see before). Note: it kinda looks like the word `Address` is the title, with an empty state.

I considered changing things to always show an `Outbounds` title with an empty state, but seeing as `outbounds` are one of the only areas we were quite often don't mind if they aren't defined, I figured it would be best to just not show an empty state at all. So, no empty state, no title, no nothing.

I guess a picture (or two) is worth a thousand words:

### Before

Screenshot 2026-06-02 at 17 15 18

### After

Screenshot 2026-06-02 at 17 15 31


schogges (approved):
👍

PR #4993, opened 2026-06-05
chore: remove copy title buttons

johncowen:
This PR mainly removes the XCopyButtons from around the page titles of detail pages.

There are a couple of other amends here to remove extra graphical elements around the title in a couple of other places:

- Any `status` elements like online/offline etc
- Any icons like `kubernetes`/`universal` for zones

Lastly I also noticed a couple of other places where we weren't using RouteTitle on detail pages to set the title of the page. So I added those here also.

### DIFF
--- a/packages/kuma-gui/features/mesh/delegated-gateways/Item.feature
+++ b/packages/kuma-gui/features/mesh/delegated-gateways/Item.feature
@@ -5,7 +5,7 @@ Feature: mesh / delegated-gateways / item
       | Alias       | Selector                                           |
       | tabs-view   | [data-testid='delegated-gateway-detail-tabs-view'] |
       | detail-view | [data-testid='delegated-gateway-detail-view']      |
-    Given the URL "/meshes/default/service-insights/service-1" responds with
+    Given the URL "/meshes/default/service-insights/gateway-1" responds with
       """
       body:
         name: gateway-1
@@ -19,8 +19,8 @@ Feature: mesh / delegated-gateways / item
       """
 
   Scenario: Overview tab has expected content
-    When I visit the "/meshes/default/gateways/delegated/service-1/overview" URL
-    Then the "$tabs-view" element contains "service-1"
+    When I visit the "/meshes/default/gateways/delegated/gateway-1/overview" URL
+    Then the "$tabs-view" element contains "gateway-1"
     Then the "$detail-view" elements contain
       | Value              |
       |       1.2.3.4:8000 |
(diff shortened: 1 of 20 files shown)


packages/kuma-gui/features/mesh/delegated-gateways/Item.feature
```diff
@@ -19,8 +19,8 @@ Feature: mesh / delegated-gateways / item
       """
 
   Scenario: Overview tab has expected content
-    When I visit the "/meshes/default/gateways/delegated/service-1/overview" URL
-    Then the "$tabs-view" element contains "service-1"
```
johncowen, line 23: This was depending on the URL slug instead of the data. As these resource doesn't have "title loaders" yet its better for it to use the data rather than the slug.

packages/kuma-gui/src/app/zones/views/ZoneDetailTabsView.vue
```diff
@@ -23,38 +23,11 @@
...
-                >
-                  <XIcon
-                    :name="env"
-                    :size="KUI_ICON_SIZE_50"
-                  >
-                    {{ t(`common.product.environment.${env}`) }}
-                  </XIcon>
-                </template>
-                <h1>
-                  <XCopyButton :text="zone.name" />
```
johncowen, line 48: I noticed this one had its RouteTitle removed in https://github.com/kumahq/kuma-gui/pull/4478 because the title setting was moved to be async (dependent on the data).

The thing is the current functionality of RouteTitle/RouteView depends on the order of mounting and that order of mounting not being dependent on something like data loading. The way it works is the last thing to mount is the thing that sets the title.

Once we add HTTP requests into the mix we can't rely on this ordering because the HTTP request come back in a completely random order.

---

For the moment I've "reverted" the loader addition here and moved back to using the URL slug (the opposite of what we want), just because of priorities.

But we need to rethink how the titling works especially as titles are being used for more and more things upstream.

For now this unblocks us and isn't a major user negative, but I'd like to solve this properly moving forwards.

packages/kuma-gui/playwright.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig(
     ...(headed ? { workers: 1, fullyParallel: false, maxFailures: 0, forbidOnly: false } : {}),
     use: {
       ...(headed ? { headless: false } : {}),
-      baseURL: process.env.KUMA_BASE_URL || 'http://localhost:5681/gui',
+      baseURL: process.env.KUMA_BASE_URL || 'http://localhost:8080/gui',
```
johncowen, line 18: We set this to 5681 in CI since a little while back, so the most common usecase is trying things out locally on port `:8080`

schogges (approved):

packages/kuma-gui/src/app/zones/views/ZoneDetailTabsView.vue
schogges: Oh, but there are other places that use the combination of `DataLoader` and `RouteTitle` 🤔

PR #4995, opened 2026-06-08
chore(deps): enable @kong-ui-public/app-layout upgrades via renovate

johncowen:
Whilst we need tighter control over `@kong-ui-public/app-layout` (which we have in-place) we still want renovate to do the upgrade for us.

In practice the renovate PR for this upgrade will probably hang around for a while until we choose to upgrade.

### DIFF
--- a/packages/config/src/renovate/base.json
+++ b/packages/config/src/renovate/base.json
@@ -81,11 +81,9 @@
     {
       "description": [
         " Disable:",
-        " - `@kong-ui-public/app-layout`: downstream conflicts due to pinned versions",
         " - `escape-string-regexp`: need to stick with current version because of esm vs cjs"
       ],
       "matchPackageNames": [
-        "@kong-ui-public/app-layout",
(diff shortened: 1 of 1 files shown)


johncowen:
actually I'm not sure if there is a way to let us "choose" exactly which version we want, I think renovate will always give us the latest which is not always what we want 🤔

schogges:
We could potentially add a note to the pr body via [`prBodyNotes`](https://docs.renovatebot.com/configuration-options/#prbodynotes) to be cautious and check upstream 🤔 

For example:

```json
{
  "prBodyNotes": ["{{#if (containsString depName '@kong-ui-public/app-layout')}}Caution: Check back required version upstream.{{/if}}"]
}
```

johncowen:
Moving to draft for the moment, I almost closed it - and I'm almost sure I will, leaving it in draft for a little bit just incase

johncowen:
Closing this, I can always reopen/redo if need be

PR #4998, opened 2026-06-09
fix(x-layout): x-stack elements dont take available space

schogges:
In `Resources` and `Policies` we have nested views that are being layed out via `XLayout` and `x-stack`. That caused that the nested views only take the space that it needed. On wide screens that meant some white space.
~I've changed the `size` attribute on these `XLayout`s to `max` and for that case let children grow. So far we've only used `size=max` in summary views for `variant=separated`, which will not conflict.~

**Before**
Screenshot 2026-06-09 at 12 09 50

Screenshot 2026-06-09 at 12 09 56

**After**

Screenshot 2026-06-09 at 12 08 23

Screenshot 2026-06-09 at 12 08 07

### DIFF
--- a/packages/kuma-gui/src/app/policies/views/PolicyTypeListView.vue
+++ b/packages/kuma-gui/src/app/policies/views/PolicyTypeListView.vue
@@ -25,7 +25,7 @@
         >
           <XLayout
             variant="x-stack"
-            size="max"
+            size="large"
           >
             <XCard
               class="policy-type-list"
@@ -121,6 +121,10 @@ import { sources } from '@/app/policies/sources'
   top: calc(var(--AppHeaderHeight) + var(--x-space-70));
   align-self: flex-start;
   max-width: 500px;
+
+  & + * {
+    flex: 1;
+  }
 }
 ul {
   list-style-type: none;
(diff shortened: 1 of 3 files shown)


johncowen:
The `size` in XLayout currently refers to the size of the gap not the size of the children.

Could we amend the flex-grow in the place where this is problematic currently? i.e. not in `x-stack`

I think longer term we may want to apply `flex-grow: 1` as a default for all "sized" `x-stack` children (but not for `separated`) but if we do that, I'd rather do it intentionally on` master`, rather than as a quick fix. 

Which reminds me I think this needs to target 2.14?

schogges:
> The `size` in XLayout currently refers to the size of the gap not the size of the children.

The `size` currently refers to the layout of elements depending on a given `size`, which indeed includes `gap` but also `justify-content`. In the end it's a set of rules that may change depending on the given `variant`.

> Could we amend the flex-grow in the place where this is problematic currently? i.e. not in `x-stack`
> 
> I think longer term we may want to apply `flex-grow: 1` as a default for all "sized" `x-stack` children (but not for `separated`) but if we do that, I'd rather do it intentionally on` master`, rather than as a quick fix.

Yes, we can do this quick fix in the two places that are currently problematic.

> Which reminds me I think this needs to target 2.14?

Yes.

johncowen:
@schogges not sure if we are wanting this one to merge down to `release-2.14` I think other than that this is fine for now, lmk!

johncowen (approved):
Thanks, we can do something better here later I think. Just I want to be more careful/intentional about it

PR #5003, opened 2026-06-09
chore: adds labels to almost all remaining summaries

johncowen:
Unfortunately quite a large PR adding labels to both data layer and summary panels.

- Adds `Labels` `tr` to almost all remaining summary panels, pretty much copy/pasta
- Adds datalayer labels where necessary, again pretty much copy/pasta
- I moved MeshGateway to use OpenAPI types
- Made ZoneIngress/ZoneEgress YAML displays show labels (similar to what we do in Dataplanes/DataplaneOverview)

~There is currently a unit test failing just because I have to update it to use labels, will sort that shortly.~

I removed two tests that were failing because I added new properties to zoneIngress/zoneEgress. The style of these two were "snapshot" type tests which we want to try to avoid

### DIFF
--- a/packages/kuma-gui/features/mesh/MeshTrust.feature
+++ b/packages/kuma-gui/features/mesh/MeshTrust.feature
@@ -2,11 +2,11 @@ Feature: mesh / mesh-trust
 
   Background:
     Given the CSS selectors
-      | Alias              | Selector                            |
-      | meshtrusts-listing | [data-testid="mesh-trusts-listing"] |
-      | item               | $meshtrusts-listing tbody tr        |
-      | summary            | [data-testid="slideout-container"]  |
-      | summary-title      | $summary [data-testid='slideout-title']  |
+      | Alias              | Selector                                |
+      | meshtrusts-listing | [data-testid="mesh-trusts-listing"]     |
+      | item               | $meshtrusts-listing tbody tr            |
+      | summary            | [data-testid="slideout-container"]      |
+      | summary-title      | $summary [data-testid='slideout-title'] |
 
   Scenario: MeshTrusts are listed in mesh overview
     Given the URL "/meshes/default/meshtrusts" responds with
--- a/packages/kuma-gui/src/app/gateways/data/MeshGateway.ts
+++ b/packages/kuma-gui/src/app/gateways/data/MeshGateway.ts
@@ -1,25 +1,36 @@
+import { Kri } from '@/app/kuma/kri'
 import { Resource } from '@/app/resources/data/Resource'
-import type { PaginatedApiListResponse } from '@/types/api.d'
-import type {
-  MeshGateway as PartialMeshGateway,
-} from '@/types/index.d'
-type PartialMeshGatewayList = PaginatedApiListResponse<PartialMeshGateway>
+import type { components } from '@kumahq/kuma-http-api'
+
+type KumaMeshGatewayList = components['responses']['MeshGatewayList']['content']['application/json']
+type KumaMeshGateway = components['schemas']['MeshGatewayItem']
 
 export const MeshGateway = {
   search(query: string) {
     return Resource.search(query)
   },
 
-  fromObject(item: PartialMeshGateway) {
-    const labels = typeof item.labels !== 'undefined' ? item.labels : {}
+  fromObject(item: KumaMeshGateway) {
+    const labels = item.labels ?? {}
+    const id = item.name
+    const mesh = item.mesh
+    const zone = labels['kuma.io/origin'] === 'zone' && labels['kuma.io/zone'] ? labels['kuma.io/zone'] : ''
+    const namespace = labels['k8s.kuma.io/namespace'] ?? ''
+    const name = labels['kuma.io/display-name'] ?? item.name
+
     return {
       ...item,
+      kri: 'kri' in item ? item.kri ?? Kri.toString({ shortName: 'mgw', mesh, zone, namespace, name }) : '',
+      name,
+      mesh,
       labels,
-      id: item.name,
-      zone: labels['kuma.io/zone'] ?? '',
+      creationTime: String('creationTime' in item ? item.creationTime ?? '' : ''),
+      modificationTime: String('modificationTime' in item ? item.modificationTime ?? '' : ''),
+      // aliases
+      id,
+      namespace,
+      zone,
       origin: labels['kuma.io/origin'] ?? '',
-      name: labels['kuma.io/display-name'] ?? item.name,
-      namespace: labels['k8s.kuma.io/namespace'] ?? '',
       config: item,
       selectors: Array.isArray(item.selectors) ? item.selectors : [],
       conf: ((item = {}) => ({
@@ -30,7 +41,7 @@ export const MeshGateway = {
               ...item,
               // An omitted hostname implies the wildcard hostname (meaning any hostname applies).
               hostname: item.hostname ?? '*',
-              protocol: item.protocol ?? 'TCP',
+              protocol: typeof item.protocol !== 'undefined' ? String(item.protocol) : 'TCP',
             }
           })
           : [],
@@ -38,7 +49,7 @@ export const MeshGateway = {
     }
   },
 
-  fromCollection(collection: PartialMeshGatewayList) {
+  fromCollection(collection: KumaMeshGatewayList) {
     const items = Array.isArray(collection.items) ? collection.items.map(MeshGateway.fromObject) : []
     return {
       ...collection,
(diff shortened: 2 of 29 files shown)


packages/kuma-gui/features/mesh/MeshTrust.feature
```diff
@@ -2,11 +2,11 @@ Feature: mesh / mesh-trust
...
-      | summary            | [data-testid="slideout-container"]  |
-      | summary-title      | $summary [data-testid='slideout-title']  |
+      | Alias              | Selector                                |
+      | meshtrusts-listing | [data-testid="mesh-trusts-listing"]     |
+      | item               | $meshtrusts-listing tbody tr            |
+      | summary            | [data-testid="slideout-container"]      |
+      | summary-title      | $summary [data-testid='slideout-title'] |
 
   Scenario: MeshTrusts are listed in mesh overview
     Given the URL "/meshes/default/meshtrusts" responds with
```
johncowen, line 12: Everything above here is just because I ran `make lint`

packages/kuma-gui/src/app/gateways/data/MeshGateway.ts
```diff
@@ -1,25 +1,36 @@
...
 import { Resource } from '@/app/resources/data/Resource'
-import type { PaginatedApiListResponse } from '@/types/api.d'
-import type {
-  MeshGateway as PartialMeshGateway,
-} from '@/types/index.d'
-type PartialMeshGatewayList = PaginatedApiListResponse<PartialMeshGateway>
+import type { components } from '@kumahq/kuma-http-api'
+
+type KumaMeshGatewayList = components['responses']['MeshGatewayList']['content']['application/json']
+type KumaMeshGateway = components['schemas']['MeshGatewayItem']
```
johncowen, line 6: Changed to use OpenAPI types instead of handwritten TS ones

packages/kuma-gui/src/app/gateways/data/MeshGateway.ts
```diff
@@ -30,15 +41,15 @@ export const MeshGateway = {
               ...item,
               // An omitted hostname implies the wildcard hostname (meaning any hostname applies).
               hostname: item.hostname ?? '*',
-              protocol: item.protocol ?? 'TCP',
+              protocol: typeof item.protocol !== 'undefined' ? String(item.protocol) : 'TCP',
```
johncowen, line 44: OpenAPI says `protocol` is `string | number` 🤷 which mess up further in, so I ensure its a string

packages/kuma-gui/src/app/gateways/views/BuiltinGatewayDataplanesView.vue
```diff
@@ -34,7 +34,7 @@
             <DataLoader
               :src="uri(dataplaneSources, '/meshes/:mesh/dataplanes/for/service-insight/:service', {
                 mesh: route.params.mesh,
-                service: gatewayData.selectors[0].match['kuma.io/service'],
+                service: gatewayData.selectors[0].match?.['kuma.io/service'] ?? '',
```
johncowen, line 37: This was as a result of moving to OpenAPI types

packages/kuma-gui/src/app/gateways/views/BuiltinGatewaySummaryView.vue
```diff
@@ -142,62 +142,52 @@
...
                     </th>
                     <td>
                       <XLayout
                         variant="separated"
                         justify="end"
                       >
-                        <template
-                          v-for="([key, value]) in Object.entries(item.labels)"
+                        <XBadge
+                          v-for="([key, value]) in Object.entries(item.selectors[0].match ?? {})"
```
johncowen, line 156: Another little thing because move to OpenAPI types

packages/kuma-gui/src/app/kuma/services/kuma-api/KumaApi.ts
```diff
@@ -208,11 +210,11 @@ export default class KumaApi extends Api {
     return this.client.get(`/meshes/${mesh}/${path}/${name}`, { params })
   }
 
-  getAllMeshGatewaysFromMesh({ mesh }: { mesh: string }, params?: PaginationParameters): Promise<PaginatedApiListResponse<MeshGateway>> {
+  getAllMeshGatewaysFromMesh({ mesh }: { mesh: string }, params?: PaginationParameters): Promise<KumaMeshGatewayList> {
     return this.client.get(`/meshes/${mesh}/meshgateways`, { params })
   }
 
-  getMeshGateway({ mesh, name }: { mesh: string, name: string }, params?: SingleResourceParameters): Promise<MeshGateway> {
+  getMeshGateway({ mesh, name }: { mesh: string, name: string }, params?: SingleResourceParameters): Promise<KumaMeshGateway> {
```
johncowen, line 217: More because move to OpenAPI types for MeshGateway

packages/kuma-gui/src/app/mesh-identities/data/MeshIdentity.ts
```diff
@@ -1,8 +1,8 @@
 import { Kri } from '@/app/kuma/kri'
 import type { components } from '@kumahq/kuma-http-api'
 
-export type KumaMeshIdentityList = components['responses']['MeshIdentityList']['content']['application/json']
-export type KumaMeshIdentity = components['schemas']['MeshIdentityItem']
+type KumaMeshIdentityList = components['responses']['MeshIdentityList']['content']['application/json']
+type KumaMeshIdentity = components['schemas']['MeshIdentityItem']
```
johncowen, line 5: We should make these very difficult to import anywhere else.

johncowen (commented):
Notes:

packages/kuma-gui/src/app/zone-egresses/data/index.ts
```diff
@@ -64,25 +66,43 @@ export const ZoneEgressOverview = {
...
-    const labels = typeof item.labels !== 'undefined' ? item.labels : {}
+    const labels = item.labels ?? {}
+    const id = item.name
+    const mesh = item.mesh
+    const zone = labels['kuma.io/origin'] === 'zone' && labels['kuma.io/zone'] ? labels['kuma.io/zone'] : ''
+    const namespace = labels['k8s.kuma.io/namespace'] ?? ''
+    const name = labels['kuma.io/display-name'] ?? item.name
+
+
+    const kri = Kri.toString({ shortName: 'ze', mesh, zone, namespace, name })
```
schogges, line 77: I don't think that _overviews have KRIs 🤔 is there a specific reason to compute one ourselves here? I still think that it's a little risky to create them ourselves, but worst case probably is 404 and usually there is always a KRI set on actual resources.

packages/kuma-gui/src/app/zone-ingresses/data/index.ts
```diff
@@ -72,25 +73,42 @@ export const ZoneIngressOverview = {
...
-    }).config
-    const labels = typeof item.labels !== 'undefined' ? item.labels : {}
+    const labels = item.labels ?? {}
+    const id = item.name
+    const mesh = item.mesh
+    const zone = labels['kuma.io/origin'] === 'zone' && labels['kuma.io/zone'] ? labels['kuma.io/zone'] : ''
+    const namespace = labels['k8s.kuma.io/namespace'] ?? ''
+    const name = labels['kuma.io/display-name'] ?? item.name
+
+    const kri = Kri.toString({ shortName: 'zi', mesh, zone, namespace, name })
```
schogges, line 83: s.a.

packages/kuma-gui/src/app/zone-egresses/data/index.ts
johncowen: This is one of the places where we take things off of `ZoneOverview` and apply it to what we show in the summary YAML tab as a `ZoneIngress` _without_ having to make a HTTP request to get the ZoneIngress we already have.

Making our own KRIs shouldn't be risky because they are purposefully extremely easy to compute (and guess), as long as we know/the HTTP API has responded with correct `mesh`, `zone`, `namespace` and `displayName` then the KRI should be correct.

I'm not that concerned about having correct KRIs aliased on the Overviews of things if we can always do this.

packages/kuma-gui/src/app/zone-ingresses/data/index.ts
johncowen: Lets close this one and continue the conversation above

packages/kuma-gui/src/app/zone-egresses/data/index.ts
schogges: Ok, so it basically saves us a request 🤔 I guess it would be good to have the backend send us the entire resource object that is nested in _overviews then?
Anyways I'm fine with the change 👍

schogges (approved):
LGTM

packages/kuma-gui/src/app/zone-egresses/data/index.ts
johncowen: > Ok, so it basically saves us a request

It saves us a request _every time you click on a table row_. If you are looking for or comparing records using the summary panel then this could be lots of requests.

Remember: The entire reason for the summary panel is to let you see expanded data quickly without making further HTTP requests (and avoiding a load for data we already have), otherwise you can just click straight to the detail page which does make extra HTTP requests.

> I guess it would be good to have the backend send us the entire resource object that is nested in _overviews then?

Yep, totally agree.

PR #5007, opened 2026-06-11
fix(resources/summary): structured causes empty default select option

### LINKED ISSUE #4979: resources: generic pages don't have a "structured" view therefore the default selection is empty
### Kuma Version

2.13.x

### Describe the bug

See below:

Image

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

schogges:
Switches from using `format` query param to `environment` in the summary view, as long as there is no `structured` view option.

Closes https://github.com/kumahq/kuma-gui/issues/4979

### DIFF
--- a/packages/kuma-gui/src/app/resources/sources.ts
+++ b/packages/kuma-gui/src/app/resources/sources.ts
@@ -83,9 +83,10 @@ export const sources = (api: KumaApi) => {
           path: {
             kri: params.kri,
           },
-        },
-        query: {
-          format: 'kubernetes',
+          // @ts-expect-error - query parameter not listed in OAS
+          query: {
+            format: 'kubernetes',
+          },
         },
       })
 
--- a/packages/kuma-gui/src/app/policies/sources.ts
+++ b/packages/kuma-gui/src/app/policies/sources.ts
@@ -85,6 +85,10 @@ export const sources = (api: KumaApi) => {
             mesh,
             name,
           },
+          // @ts-expect-error - query parameter not listed in OAS
+          query: {
+            format: 'kubernetes',
+          },
         },
       })
       return YAML.stringify(res.data)
--- a/packages/kuma-gui/src/app/resources/views/ResourceDetailView.vue
+++ b/packages/kuma-gui/src/app/resources/views/ResourceDetailView.vue
@@ -5,7 +5,7 @@
       mesh: '',
       kri: '',
       resourcePath: '',
-      format: String,
+      environment: String,
       codeSearch: '',
       codeFilter: false,
       codeRegExp: false,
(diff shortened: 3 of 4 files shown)


packages/kuma-gui/src/app/resources/sources.ts
```diff
@@ -83,9 +83,10 @@ export const sources = (api: KumaApi) => {
           path: {
             kri: params.kri,
           },
-        },
-        query: {
-          format: 'kubernetes',
+          // @ts-expect-error - query parameter not listed in OAS
+          query: {
+            format: 'kubernetes',
+          },
```
schogges, line 89: While I was here I also found this: `query` wasn't nested in `params`

johncowen:
Yesterday when we were discussing this in other PRs I was unsure whether this was an issue in the detail view also, and yesterday I checked but couldn't get it to show this, but just now I got it to not work on this PR also, could we take a look at that here also?

Screenshot 2026-06-11 at 11 17 32

packages/kuma-gui/src/app/resources/sources.ts
johncowen: good catch 👍 

big shame that this isn't caught by OpenAPI TS

johncowen:
Don't forget to change this one to `release-2.14` also

johncowen:
Please ping me offline when this is ready

packages/kuma-gui/src/app/resources/sources.ts
schogges: > big shame that this isn't caught by OpenAPI TS

Yeah, very unfortunate 😞

johncowen:
Hmm, netlify deploys don't seem to be updating on commit 🤔 it usually does that right, update after a new commit is pushed?

I'm just checking this locally, seems to be fine locally, gimme 2 more sec

johncowen:
@lobkovilya noticed we are missing a:

```
          // @ts-expect-error - query parameter not listed in OAS
          query: {
            format: 'kubernetes',
          },

```

from here also:

https://github.com/kumahq/kuma-gui/blob/5a21d2dbb1372ee97ccac094d57f52d855a9beed/packages/kuma-gui/src/app/policies/sources.ts#L80-L91

Could we add that here?

I think we might want to consider something similar to our titles test to test this across all resources, I'm always cautious of adding these "1 feature file to test all resources" but I think we might want to consider it for this.

That can be a follow up though, lets make the fix here in the meantime 👍

johncowen:
Quick PR ref for the above bug https://github.com/kumahq/kuma-gui/pull/4333, from what I can quickly tell the bug is in 2.12 and 2.13

johncowen:
I also checked all other `sources` files and they all have some sort of `kubernetes` query param for this source, so its just the policy one

schogges:
> @lobkovilya noticed we are missing a:
> 
> ```
>           // @ts-expect-error - query parameter not listed in OAS
>           query: {
>             format: 'kubernetes',
>           },
> ```
> 
> from here also:
> 
> https://github.com/kumahq/kuma-gui/blob/5a21d2dbb1372ee97ccac094d57f52d855a9beed/packages/kuma-gui/src/app/policies/sources.ts#L80-L91
> 
> Could we add that here?
> 
> I think we might want to consider something similar to our titles test to test this across all resources, I'm always cautious of adding these "1 feature file to test all resources" but I think we might want to consider it for this.
> 
> That can be a follow up though, lets make the fix here in the meantime 👍

Will do it 👍 

> I also checked all other sources files and they all have some sort of kubernetes query param for this source, so its just the policy one

Thanks for checking 👍

johncowen (approved):
nice thanks! I checked both policies and resources locally and all looks much better now 👍

PR #5008, opened 2026-06-11
chore: adds more KRIs to Overview-like things

johncowen:
Adds/ensures KRIs are carried through to the summary YAML views.

There is also a bit of copy/pasting for consistency reasons between Dataplanes/ZoneProxies. There are reasons for formatting decisions in the code here and I commented it up quite a bit to explain why rather than just depend on the formatting to be self-explanatory (that alone it not explanatory enough I don't think)

I renamed the Dataplane tests and added back the ZoneProxy tests but renamed them all to make it clearer what they are testing. Usually we avoid mocking/setting up full data objects to test against, but in this case its the most straight-forwards way to assert that we have neither too many fields or not enough fields. If I think of a different way thats just as straight-forwards but also clearer I might tweak this in the future.

Note: We are now expected KRIs to be added to the top-level DataplaneOverviews via the backend. Depending on when this PR gets merged we can either consider this in this PR or a follow up PR once the backend PR is complete.

### DIFF
--- a/packages/kuma-gui/src/app/zones/data/index.ts
+++ b/packages/kuma-gui/src/app/zones/data/index.ts
@@ -87,7 +87,7 @@ export const ZoneOverview = {
 
     const labels = item.labels ?? {}
     const id = item.name
-    // check for label first, fallback to tags
+
     const name = labels['kuma.io/display-name'] ?? item.name
 
     // temporarily make a KRI until we have those from the backend
--- a/packages/kuma-gui/src/app/zones/data/index.ts
+++ b/packages/kuma-gui/src/app/zones/data/index.ts
@@ -90,6 +90,7 @@ export const ZoneOverview = {
     // check for label first, fallback to tags
     const name = labels['kuma.io/display-name'] ?? item.name
 
+    // temporarily make a KRI until we have those from the backend
     const kri = Kri.toString({ shortName: 'z', mesh: '', zone: '', namespace: '', name })
 
     const insight = ZoneInsight.fromObject(item.zoneInsight)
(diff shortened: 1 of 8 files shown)


packages/kuma-gui/src/app/zones/data/index.ts
```diff
@@ -83,6 +84,15 @@ export const ZoneOverview = {
   },
 
   fromObject: (item: PartialZoneOverview) => {
+
+    const labels = item.labels ?? {}
+    const id = item.name
+    // check for label first, fallback to tags
+    const name = labels['kuma.io/display-name'] ?? item.name
```
schogges: nit: the comment doesn't match with the implementation

schogges (dismissed):
One super nit, the rest LGTM

packages/kuma-gui/src/app/zones/data/index.ts
johncowen: 2 tics

johncowen: Done! Thanks!

schogges (approved):
👍

johncowen:
important note: Currently https://github.com/kumahq/kuma/pull/16925 is up in Draft which would mean we could take the backend generated KRI by default and fallback (mostly for TS reasons to our own fallback).

I could either wait for the above PR, or circle back once its gone in (if it goes in in the near future)

I would rather merge this now as it achieves what we want even though we generate the KRI from labels (pretty safe). We can then circle back and take into account the backend generated one once the above backend PR goes in.

Gonna merge!

PR #4997, opened 2026-06-09
chore(resources): add tests

schogges:
Extracts (mostly) generated tests and mocks from https://github.com/kumahq/kuma-gui/pull/4946

PR #5010, opened 2026-06-11
fix(resources): fix i18n based fallbacks

johncowen:
### Improvement
The default "group" for `ResourceTypeDescriptor` if its not set is `others` (it can also be `policy` and `mesh` and `global`, plus anything else that could be added via the backend in the future):

https://github.com/kumahq/kuma-gui/blob/5a21d2dbb1372ee97ccac094d57f52d855a9beed/packages/kuma-gui/src/app/resources/data/ResourceTypeDescriptor.ts#L53

We then use this `type.group` key to discover several i18n keys including a documentation paragraph and `href`

Thing is `others` doesn't have a `description` nor `href` key.

We sometimes used a `default` fallback which in some places would solve the issue of "future groups" coming from the backend, but not if its not set. I guess it highly unlikely that we end up with a `type.group=others` but we've coded for it so we should support it.

I removed the `default` section and used `others` for a default (like we do in the datalayer)

### Fix

I also found a place where we were defaulting to `t('common.product.docs')` which is a bit of a legacy i18n key, which we only really use to set the `KUMA_DOCS_URL` i18n env var:

https://github.com/kumahq/kuma-gui/blob/5a21d2dbb1372ee97ccac094d57f52d855a9beed/packages/kuma-gui/src/app/application/services/i18n/I18n.ts#L45-L48

We also mock out this key only at dev time because we could never get the docs redirects implemented to avoid us having mock this value out.

The mocked key/value didn't make use of the same variables as the non-mocked key, which meant _during dev_ you wouldn't see an i18n error if you did not pass the required variables to the call for `t`. The issue would then fail in CI only, causing a bit of confusion https://github.com/kumahq/kuma-gui/pull/4997#discussion_r3396571856

The fix involves:

- Changing the `common.product.docs` key to something else, I used the "fallback" `others` key.
- I also thought of how we can prevent this weirdness happening in the future and decided to add the same variabels to the debug URL. Whilst these are "noops" they force you to pass the same variables through. I also added comments to "ensure" that if we add any new vars here (unlikely) we are prompted to sync the variables.

A better fix would be to remove or somehow prevent the `common.product.docs` from being used outside of setting `KUMA_DOCS_URL`. Whilst I would prefer this, its needs better thought/less haste.

---

This desc is a little rough, so let me know if there are any further questions.

Lastly, I think it would be a good idea to have a closer look over the new `resources` section with the above in mind to triple check there are no similar issues. I had a relatively "ok" look myself but I'd rather have a second pair of eyes.

### DIFF
--- a/packages/kuma-gui/src/app/kuma/debug.ts
+++ b/packages/kuma-gui/src/app/kuma/debug.ts
@@ -15,7 +15,9 @@ export const locales = (app: Record<string, Token>): ServiceDefinition[] => [
       return {
         common: {
           product: {
-            docs: 'https://kuma.io/docs/dev',
+            // ensure all variables in kuma/locales/en.us/index.yaml are also
+            // used here
+            docs: 'https://kuma.io/docs/dev{ major, select, other {}}{ minor, select, other {}}',
           },
         },
       }
--- a/packages/kuma-gui/src/app/resources/views/ResourceListView.vue
+++ b/packages/kuma-gui/src/app/resources/views/ResourceListView.vue
@@ -40,10 +40,9 @@
                 >
                   {{ t('resources.routes.items.types.collection.outbound') }}
                 </XBadge>
-
                 <XAction
                   action="docs"
-                  :href="t(`resources.routes.items.types.${type.group}.href.docs`, { name: type.name }, { defaultMessage: t('common.product.docs')})"
+                  :href="t(`resources.routes.items.types.${type.group}.href.docs`, { name: type.name }, { defaultMessage: t(`resources.routes.items.types.others.href.docs`, { name: type.name })})"
                   data-testid="policy-documentation-link"
                 >
                   <span class="visually-hidden">{{ t('common.documentation') }}</span>
@@ -57,7 +56,7 @@
             </header>
             <XI18n
               :path="`resources.routes.items.types.${type.group}.description`"
-              default-path="resources.routes.items.types.default.description"
+              default-path="resources.routes.items.types.others.description"
             />
           </XCard>
           <XCard>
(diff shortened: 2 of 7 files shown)


johncowen:
Slightly related:

I noticed during the CI runs on this PR that this one seems a little flakey:

Screenshot 2026-06-11 at 19 34 19

johncowen:
I fixed the tests here also, wasn't a flake was a result of going from `format` to `enviroment` and thee tests were originally written before we did that.

packages/kuma-gui/src/app/kuma/debug.ts
```diff
@@ -15,7 +15,9 @@ export const locales = (app: Record<string, Token>): ServiceDefinition[] => [
       return {
         common: {
           product: {
-            docs: 'https://kuma.io/docs/dev',
+            // ensure all variables in kuma/locales/en.us/index.yaml are also
+            // used here
+            docs: 'https://kuma.io/docs/dev{ major, select, other {}}{ minor, select, other {}}',
```
johncowen, line 20: This is the actual "fix" well, more the tripwire.

Now, in the view code if you try to use this key without specifying these vars it throws a runtime error.

packages/kuma-gui/src/app/resources/views/ResourceListView.vue
```diff
@@ -40,10 +40,9 @@
                 >
                   {{ t('resources.routes.items.types.collection.outbound') }}
                 </XBadge>
-
                 <XAction
                   action="docs"
-                  :href="t(`resources.routes.items.types.${type.group}.href.docs`, { name: type.name }, { defaultMessage: t('common.product.docs')})"
+                  :href="t(`resources.routes.items.types.${type.group}.href.docs`, { name: type.name }, { defaultMessage: t(`resources.routes.items.types.others.href.docs`, { name: type.name })})"
```
johncowen, line 45: Here's the actual fix, we don't use `common.product.docs` but instead a fallback key that then uses `KUMA_DOCS_URL`

schogges: I'm still a little confused because `MeshAccessLog`, the resource being used in the tests, should be classified as `policy` and never use the `defaultMessage` 🤔

schogges:
> I fixed the tests here also, wasn't a flake was a result of going from `format` to `enviroment` and thee tests were originally written before we did that.

Thanks, I missed that and didn't rebase the tests PR on the release branch before merging 👍

schogges (approved):
LGTM, thanks for fixing this 👍

PR #5012, opened 2026-06-12
chore(ci): make playwright fail on any error or console.error

johncowen:
We used to have Cypress plugins to do this sort of thing for us.

With Playwright it looks to be just `pageerror` and `console` listener/passthroughs

This conversation made me thing about this https://github.com/kumahq/kuma-gui/pull/4997#discussion_r3396571856

### DIFF
--- a/packages/gherkin-web/src/playwright/steps/index.ts
+++ b/packages/gherkin-web/src/playwright/steps/index.ts
@@ -18,6 +18,17 @@ export const test = base.extend<{
   baseURL: async ({ baseURL }, use) => {
     await use(baseURL)
   },
+  page: async ({ page }, use) => {
+    page.on('pageerror', (err) => {
+      throw err
+    })
+    page.on('console', (msg) => {
+      if(['error'].includes(msg.type()) && msg.args().length !== 0) {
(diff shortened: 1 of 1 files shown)


packages/gherkin-web/src/playwright/steps/index.ts
```diff
@@ -18,6 +18,17 @@ export const test = base.extend<{
   baseURL: async ({ baseURL }, use) => {
     await use(baseURL)
   },
+  page: async ({ page }, use) => {
+    page.on('pageerror', (err) => {
+      throw err
+    })
+    page.on('console', (msg) => {
+      if(['error'].includes(msg.type()) && msg.args().length !== 0) {
```
johncowen, line 26: Only fail on `console.error` types, but also the browser based native "console errors" lioke network errors also fire this. But I noticed that the browser based ones have no inspectable `args` whereas user ones will always have at least one arg.

This means we can respond with things like `504`s to test error states (which produce browser native errors) without failing the test (because we want test the error states)

schogges (approved):
LGTM 👍

PR #5042, opened 2026-06-18
chore: stop using kongponents

johncowen:
Stop using kongponents i.e. don't install them globally anymore

We use `@kumahq/x` now instead and have done for quite a while.

---

I also did some checking here on bundle reduction size:

```
# master
dist/assets/index-BqGd6Mxt.js                                              1,435.59 kB │ gzip: 439.39 kB
```

```
# this PR
dist/assets/index-D2BHdZfV.js                                              1,110.26 kB │ gzip: 343.58 kB

```

savings of 100kb gzipped and the parsing etc of that 🎉

### DIFF
--- a/packages/config/src/eslint.js
+++ b/packages/config/src/eslint.js
@@ -237,21 +237,20 @@ export function createEslintConfig(
             'search',
             // vue-router
             'RouterView',
-            // @kong/kongponents
-            'K[A-Z].*',
-            //
+            // @kumhq/x
             'X[A-Z].*',
+            // @kumahq/kuma-gui
             'Kuma[A-Z].*',
-            // @kong-ui-public/i18n
-            'I18nT',
-            // Application
-            'AppView',
+            // @kumahq/data
             'DataLoader',
             'DataSource',
             'DataSink',
-            'DataCollection',
+            // @kumahq/routing
             'RouteView',
             'RouteTitle',
+            // Application
+            'AppView',
+            'DataCollection',
             ...componentIgnorePatterns,
           ],
         }],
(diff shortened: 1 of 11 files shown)


schogges (dismissed):
LGTM

I suppose we are aiming to remove `@kong/kongponents` from the kuma-gui manifest. Which requires to have some more exports from `x` that are currently being imported from `@kong/kongponents`.

johncowen:
Oh sorry I forgot to mention, from what I remember this dramtically reduces our bundle size because we no longer globally `use`/`install` kongponents.

> I suppose we are aiming to remove @kong/kongponents from the kuma-gui manifest. Which requires to have some more exports from x that are currently being imported from @kong/kongponents.

If I understood right I think this might be a reference to the couple of kongponents imports that are left in kuma-gui. In which case the answer is yes. Ideally these would also come from `kumahq/x`.

I think the only thing holding me back here is simply choosing a good name for one of the things that are left and then moving the ones that are left (I guess we'll have `XPagination` and some sort of `XDataTable` or something like that.

My main thing here was the reduction in bundle size though, I can circle back soon to remove these last few references and then remove from the kuma-gui manifest entirely.

johncowen:
FYI: I rebased and added a commit to sort out eslint so it tells us if we try and use a kongponent globally, will need another approval 😬

johncowen:
I also did some checking here on bundle reduction size:

```
# master
dist/assets/index-BqGd6Mxt.js                                              1,435.59 kB │ gzip: 439.39 kB
```

```
# this PR
dist/assets/index-D2BHdZfV.js                                              1,110.26 kB │ gzip: 343.58 kB

```

savings of 100kb gzipped and the parsing etc of that 🎉

johncowen:
FYI: I still plan on resurrecting this, its currently blocked because `public-ui-components` AppAboutSection requires kongponents to be installed globally in your app (because it uses KCard without importing it) and thus that currently blocks this PR.

I have a plan to avoid this so we can still stop installing kongponents globally. I'll try and pick this back up again soonish.

PR #5043, opened 2026-06-18
scratch(deps): widen peers

schogges:
This is an experiment to see how renovate treats `peerDependencies` with a `rangeStrategy` of `widen`.

The peer on `kongponents` in x will be removed soon again.


johncowen (approved):
Lets give it a go!

PR #5047, opened 2026-06-18
chore: stop using kong/design-tokens

johncowen:
Similar to https://github.com/kumahq/kuma-gui/pull/5042, we use @kumahq/x now instead and have done for quite a while.

Note: I plan on adding https://stylelint.io/user-guide/rules/no-unknown-custom-properties/ very soon as a follow up.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13256,7 +13256,6 @@
       },
       "devDependencies": {
         "@faker-js/faker": "^10.5.0",
-        "@kong/design-tokens": "^3.0.1",
         "@kumahq/config": "*",
         "@kumahq/fake-api": "*",
         "@kumahq/gherkin-web": "*",
--- a/packages/kuma-gui/src/app/common/TargetRef.vue
+++ b/packages/kuma-gui/src/app/common/TargetRef.vue
@@ -12,7 +12,6 @@
     <XBadge v-else>
       <slot />
     </XBadge>
-
     <TagList
       v-if="props.targetRef.kind === 'MeshServiceSubset' && props.targetRef.tags"
       :tags="props.targetRef.tags"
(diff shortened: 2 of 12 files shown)


packages/kuma-gui/src/app/common/TargetRef.vue
```diff
@@ -26,16 +25,13 @@
       <img
         src="@/assets/images/icon-weight.svg?url"
         alt="Weight"
-        :width="KUI_ICON_SIZE_30"
```
johncowen, line 29: Note this actually looks better with no width (it just takes the 11x12 size of the svg)

I'll probably change this to use `x-icon-start` in a follow up seeing as its not a kong icon

package-lock.json
```diff
@@ -13775,6 +13774,7 @@
         "@kumahq/config": "*"
       },
       "peerDependencies": {
+        "@kong/kongponents": "^9.59.0",
```
schogges: When we did our test with widening peers I think I've missed to run install such that it ends up in the lock-file. Not really an issue and I've cleaned up everything now. Please run install again and I can approve this 🙂

johncowen: Gotcha, yeah I noticed also and was waiting for your PRs, I can rebase this in a sec I think 👍 , will ping after the rebase

package-lock.json
```diff
@@ -13766,7 +13765,6 @@
         "@kumahq/config": "*"
       },
       "peerDependencies": {
-        "@kong/kongponents": "^9.59.0",
```
johncowen: Hmm this is still here, I guess its on master?

johncowen: yeah looks to me like this is on `master` and it shouldn't be, if thats correct happy to leave this fix in here also, lmk!

johncowen:
@schogges I spotted this one has been lost a little I think, still needs an approval after a rebase I think so should be straightforwards to get this finished up

schogges (dismissed):
👍

schogges (approved):

PR #5049, opened 2026-06-19
fix: remove double "visit" for services test

johncowen:
This test has a `Background` that already "visits" the page to be tested, so this secondary one isn't required.

Moreover, I found that have this actually causes an error and now a test failure since we added https://github.com/kumahq/kuma-gui/pull/5012

I'm thinking this is because the first visit starts a HTTP request, which is then cancelled (HTTP0 🎉 ) by the second "visit" which causes an error and fails the test before it even gets started.

(note: `master` is currently failing due to this)


schogges (approved):
Nice catch 👍

PR #5050, opened 2026-06-19
chore: use github-buttons

johncowen:
Instead of `vue-github-button` (which depended on `github-buttons`) and therefore removes one dependency.

(I've also forever noticed that sometimes during dev at least, the way `vue-github-button` works means that sometimes the button doesn't load properly and produces console errors etc)


schogges (approved):
👍

PR #5052, opened 2026-06-22
fix(kuma-http-api/mocks): next query param separator

schogges:
There was no query param separator `&` in the URL for `next`

### DIFF
--- a/packages/kuma-gui/features/zones/zone-cps/ingresses/Navigation.feature
+++ b/packages/kuma-gui/features/zones/zone-cps/ingresses/Navigation.feature
@@ -3,7 +3,7 @@ Feature: zones / ingresses / navigation
   Background:
     Given the CSS selectors
       | Alias         | Selector                                                                    |
-      | row           | [data-testid$='-collection'] tbody tr:first-child                                 |
+      | row           | [data-testid$='-collection'] tbody tr:first-child                           |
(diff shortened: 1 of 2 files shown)


johncowen:
Yiokes! This is one of those things that has been spoken about lots in the past 😆 

Er, I guess the best thing for me to ask here is, can you find anywhere in the current OpenAPI specs that says this should be `string | null`? I cannot find the keyword `nullable` anywhere and everywhere I see `next` its not specified in the `required` array, making it `string | undefined`.

Pretty sure we don't even use `next` anywhere either (ah actually 👀 , ok I just saw where we want to start using it I can probably respond there)

schogges:
> Yiokes! This is one of those things that has been spoken about lots in the past 😆
> 
> Er, I guess the best thing for me to ask here is, can you find anywhere in the current OpenAPI specs that says this should be `string | null`? I cannot find the keyword `nullable` anywhere and everywhere I see `next` its not specified in the `required` array, making it `string | undefined`.
> 
> Pretty sure we don't even use `next` anywhere either (ah actually 👀 , ok I just saw where we want to start using it I can probably respond there)

Yeah, this needs fixing in OAS. We'd need to make it required and `'null'`. We currently only use `nullable` in the overlays. I've removed the fix for the type of `next` in here and think about it separately.

johncowen (approved):
Thanks 👍

PR #5053, opened 2026-06-22
chore(ci): small follow ups to new connect mock server

johncowen:
1. Moves latency mocking into the server
2. For the system-under-test (i.e. the app, not the runner), depends on the in-HTML vars to figure out the `KUMA_API_URL`
3. Adds `bdd` auto-generation to the preview server (we use this to test in CI, and occasionally locally if we want to test a prod build locally.

**Important Note**: (something I forget to mention in the previous PR). We should permanently set `KUMA_MOCK_API_ENABLED=false`. This means MSW is disabled and during development the app will then use the new connect server which uses real HTTP. The only thing now that should be using MSW are PR previews.

I would like to re-think how `KUMA_MOCK_API_ENABLED` functions/is used in the future, but for now just set it to false.

### DIFF
--- a/packages/kuma-gui/vite.config.preview.ts
+++ b/packages/kuma-gui/vite.config.preview.ts
@@ -1,4 +1,4 @@
-import { replicateKumaServer } from '@kumahq/config/vite'
+import { replicateKumaServer, playwrightBdd } from '@kumahq/config/vite'
 import fakeApi from '@kumahq/fake-api/vite'
 import { fs, dependencies } from '@kumahq/kuma-http-api/mocks'
 import { defineConfig } from 'vite'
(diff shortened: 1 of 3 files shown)


packages/kuma-gui/vite.config.preview.ts
```diff
@@ -15,6 +15,7 @@ export const config: UserConfigFn = () => {
         template: './dist/gui/index.html',
       }),
       fakeApi({ dependencies, fs }),
+      playwrightBdd(),
```
schogges, line 18: Why do we need `playwrightBdd` here?

schogges: Ah

> Adds bdd auto-generation to the preview server (we use this to test in CI, and occasionally locally if we want to test a prod build locally.

nvm 👍

schogges (approved):

PR #5057, opened 2026-06-23
fix(ci): only passthrough local URLs

johncowen:
I found that since we added the `passthrough` config, non-local HTTP requests where also "passing through".

This meant requests to `https://kumaio.latest_version` where not being mocked.

I made a little guard to fix this temporarily, but that adds a kuma specific usecase. At somepoint `passthrough` should be hanged to be a callback so any project can decide for itself whether a HTTP request should passthrough or not.

### DIFF
--- a/packages/gherkin-web/src/playwright/steps/index.ts
+++ b/packages/gherkin-web/src/playwright/steps/index.ts
@@ -1,4 +1,4 @@
-import { createFetch } from '@kumahq/fake-api'
+import { createFetch, Cookie } from '@kumahq/fake-api'
 import { expect } from '@playwright/test'
 import { createBdd, test as base } from 'playwright-bdd'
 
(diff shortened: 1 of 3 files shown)


packages/gherkin-web/src/playwright/steps/index.ts
```diff
@@ -89,46 +89,41 @@ export async function setupSteps({
       return context.route(
         (u) => routeToRegexp(route).test(u.toString()),
         async (route, request) => {
-          try {
-            const url = request.url()
-            const cookies = await context.cookies()
-            const envs = Object.entries(env).map(([name, value]) => ( {name, value} ))
-            const headers = {
-              cookie: [...cookies, ...envs].map((c) => `${c.name}=${c.value}`).join('; '),
```
johncowen, line 97: This line has been moved to `Cookie.stringify` (see above)

packages/gherkin-web/src/playwright/steps/index.ts
```diff
@@ -89,46 +89,41 @@ export async function setupSteps({
...
+            ...request.headers(),
+          }
 
-            client.request({
-              url: new URL(url),
-              request: {
-                method: request.method(),
-                body: request.postDataJSON() ?? {},
-              },
+          const url = new URL(request.url())
```
johncowen, line 99: I found we could create a `URL` once at the top of everywhere its needed

packages/gherkin-web/src/playwright/steps/index.ts
```diff
@@ -89,46 +89,41 @@ export async function setupSteps({
...
+          client.request({
+            url,
+            request: {
+              method: request.method(),
+              body: request.postDataJSON() ?? {},
+            },
+          })
+          // @TODO: We should probably turn this check into a callback
+          if(config.passthrough && url.origin === new URL(baseURL).origin) {
```
johncowen, line 108: This is the important bit, we only passthrough if the URLs origin is the same as the origin of the site we are testing, otherwise we continue to use playwright's mocking instead of the local connect server.

The upshot is:

- `http://kuma.io/latest_version` uses playwright mocking (fake HTTP)
- Anything at say `http://localhost:5681/` uses the local connect server (with real HTTP)

packages/gherkin-web/src/playwright/steps/index.ts
```diff
@@ -89,46 +89,41 @@ export async function setupSteps({
...
-          } catch (e) {
-            console.error(e)
-            await route.continue()
+            await route.fulfill({
+              status: parseInt(response.headers.get('Status-Code') ?? '200'),
+              contentType: type,
+              headers: Object.fromEntries(response.headers.entries()),
+              body,
+            })
+
```
johncowen, line 126: This is all green because I unwrapped the `try/catch` which from what I remember (it was a very long time ago) was there for Cypress reasons.

packages/gherkin-web/src/playwright/steps/index.ts
```diff
@@ -176,60 +171,54 @@ export async function setupSteps({
...
-              method: request.method(),
-              body: request.postDataJSON() ?? {},
-            },
+        } else {
+          merged = _yaml
+          await route.fulfill({
+            contentType: type,
+            headers: merged.headers,
+            status: parseInt(merged.headers?.['Status-Code'] ?? response.headers.get('Status-Code') ?? '200'),
+            body: merged.body as string,
```
johncowen, line 219: Similar changes to the very similar (but not quite) changes above:

1. Use `Cookie.stringify`
2. Create a `new URL()` once
3. Remove `try/catch`

Note: when specifically using this step in tests it _always_ use playwright fake HTTP, not real connect-server based HTTP. I would like to change this once we are happy with the entire approach as I think it could also get the HTTP response from the connect server and then merge that with the mock.

packages/kuma-gui/playwright/main.ts
```diff
@@ -24,7 +24,7 @@ export { test } from '@kumahq/gherkin-web/playwright'
         service: () => {
           // these are only fed to the mocks
           return {
-            KUMA_API_URL: () => 'http://localhost:5681',
+            KUMA_API_URL: () => new URL(process.env.KUMA_BASE_URL || 'http://localhost:8080/gui').origin,
```
johncowen, line 27: This means that if you are using `http://localhost:5681` or `http://localhost:8080` to test against, then all the mocks are configured correctly with the correct baseURL without you having to do anything.

johncowen (commented):
Notes:

schogges (approved):
LGTM

PR #5065, opened 2026-06-25
chore(kumahq/x): moves XDrawer to use a teleport layer like modals

johncowen:
Reorganizes XDrawer so it can be relatively positioned.

### DIFF
--- a/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
+++ b/packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
@@ -138,11 +138,15 @@
           </ul>
         </nav>
       </div>
-      <main
-        class="app-main-content"
-      >
+      <main>
+        <div class="drawer">
+          <XTeleportSlot
+            name="drawer-layer"
+          />
+        </div>
         <XLayout
           variant="y-stack"
+          class="app-main-content"
         >
           <div>
             <slot name="default" />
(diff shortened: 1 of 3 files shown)


packages/kuma-gui/src/app/kuma/components/ApplicationShell.vue
```diff
@@ -176,22 +180,30 @@ onMounted(() => {
 </script>
 <style lang="scss" scoped>
 .application-shell {
-  --AppHeaderHeight: 60px;
+  --kuma-app-top: 60px;
```
johncowen, line 183: We own the `--kuma-*` namespace as a follow up we should change the others here to use that namespace.

packages/x/src/components/x-drawer/XDrawer.vue
```diff
@@ -1,54 +1,72 @@
...
   width: '560px',
+  selector: 'main',
 })
 
 const emit = defineEmits<{
   (event: 'close'): void
 }>()
+
+onMounted(async () => {
+  await nextTick()
```
johncowen, line 44: Because we teleport you have to wait a tick for the teleport to happen and the element to be available.

packages/x/src/components/x-drawer/XDrawer.vue
```diff
@@ -68,5 +86,7 @@ const emit = defineEmits<{
 }
 :deep(.slideout-container) {
   z-index: auto !important;
+  top: unset !important;
+  height: calc(100vh - var(--x-drawer-offset-top)) !important;
```
johncowen, line 90: We've already relatively positioned the drawer now because we've moved it higher in the DOM tree via the Teleport, so we don't need a `top` anymore.

We do need a height tho, and we set that to the size of the viewport minus the defined `top` of the app (dynamic or not)

packages/x/src/components/x-drawer/XDrawer.vue
```diff
@@ -1,54 +1,72 @@
...
 
 const emit = defineEmits<{
   (event: 'close'): void
 }>()
+
+onMounted(async () => {
+  await nextTick()
+  const $el = document.getElementById(id)
+  if($el) {
+    onClickOutside(
```
johncowen, line 47: I found that this had broken at some point, I'm not entire sure how (I'm assuming it worked when it was implemented). I haven't looked to see exactly why yet, but I've fixed it here also seeing as its kinda related.

johncowen (commented):
Notes:

schogges (approved):
LGTM 👍

johncowen:
FYI: I had a few more thoughts/improvements here, but I'm going to move forwards with a merge now and follow up with some DX tweaks in a later PR.

PR #5066, opened 2026-06-26
fix(data-planes): fetch layout when mesh-service non-exclusive

### LINKED ISSUE #5035: Can't use DP _layout without meshService enabled
### Kuma Version

2.14

### Describe the bug

We have two DP detail views, one that requires meshServices `Exclusive` (and more, i.e. `unified-resource-naming`) and one as a fallback (legacy). The legacy one could still be used with `meshServices.mode: Exclusive` but without the other requirements for the new view.

In both views we fetch the `_layout` endpoint, but only in the new view we are guaranteed that meshServices are enabled. In the legacy view we fetch the `_layout` endpoint to show the `spiffeId` in the about section. This request fails when meshServices are not enabled.
We have to check wether meshServices are enabled before we fetch the endpoint.

### To Reproduce

_No response_

### Expected behavior

_No response_

Image

schogges:
If `mesh.meshServices.mode` is not `Exclusive`, fetching the `_layout` endpoint causes an error. I've left the data-source in the same place, even though it means that we have to use a ternary for the `:src` and have to add the types. This means that the data-source can stay in the same place when we need to add a `DataLoader` somewhere deeper.

This needs backport to 2.14.

Closes https://github.com/kumahq/kuma-gui/issues/5035


johncowen:
I've only taken a brief look at this, but my first thought is:

Do you have any ideas on what can we do to help engineers avoid a similar very easy mistake to make in the future?

schogges:
> I've only taken a brief look at this, but my first thought is:
> 
> Do you have any ideas on what can we do to help engineers avoid a similar very easy mistake to make in the future?

You mean issues like fetching an endpoint that's only supported in a certain environment?

johncowen:
Yep exactly

schogges:
Hmmm 🤔 Ideally that's somehow documented in OAS. But AFAIK there is no native gating in open api that we could use 🤔 Potentially we could add some `x-` field, use that in our own tooling and construct some sort of discriminative union on top

johncowen:
What about if we made the mock error with a 400 if the `KUMA_MESHSERVICE_MODE` env var isn't `Exclusive`? i.e. replicate what the real backend does.

That way if you tried to build something that used this endpoint you'd be more likely to notice.

schogges:
> What about if we made the mock error with a 400 if the `KUMA_MESHSERVICE_MODE` env var isn't `Exclusive`? i.e. replicate what the real backend does.
> 
> That way if you tried to build something that used this endpoint you'd be more likely to notice.

I think there is still a chance that it would be missed. IIRC I've implemented this with a local deployment, but had `meshServices.mode` set to `Exclusive`. I've just missed that constraint. I could've used mocks that do what you've described and could have still missed it 🤔

johncowen:
I guess my point is: we should always try to mock out actual behaviour, especially when we spot a bug caused by us not having mocked out actual behaviour.

In this case if the mock could check for non-exclusive and return an error if that's the case, its highly likely to be spotted either at dev time or during our randomised testing (if there is a test covers the endpoint usage)

Lets add a mock that covers that?

schogges:
> I guess my point is: we should always try to mock out actual behaviour, especially when we spot a bug caused by us not having mocked out actual behaviour.
> 
> In this case if the mock could check for non-exclusive and return an error if that's the case, its highly likely to be spotted either at dev time or during our randomised testing (if there is a test covers the endpoint usage)
> 
> Lets add a mock that covers that?

Yeah, makes sense to replicate real behaviour as much as we can 👍 I'll try to add something

packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_layout.ts
```diff
@@ -43,8 +43,26 @@ export default ({ env, fake }: Dependencies): ResponseHandler => (req) => {
...
+      body: {
+        'type': '/std-errors',
+        'status': 400,
+        'title': 'Bad Request',
+        'detail': 'bad request: can\'t use _layout endpoint without meshService enabled',
+        'details': 'bad request: can\'t use _layout endpoint without meshService enabled',
+      },
+    }
+  }
+
```
johncowen, line 62: Perfect, looks great! 👍 Let me take a look at the rest of the PR also

packages/kuma-gui/src/app/legacy-data-planes/views/DataPlaneDetailView.vue
```diff
@@ -20,11 +20,11 @@
...
     >
       <DataSource
-        :src="uri(sources, '/meshes/:mesh/dataplanes/:name/layout', {
+        :src="props.mesh.meshServices.mode === 'Exclusive' ? uri(sources, '/meshes/:mesh/dataplanes/:name/layout', {
           mesh: route.params.mesh,
           name: route.params.proxy,
-        })"
-        v-slot="{ data: sourceDataplaneLayout }"
+        }) : ''"
+        v-slot="{ data: sourceDataplaneLayout }: DataSourceResponse<DataplaneNetworkingLayout>"
```
johncowen, line 27: How come we don't just wrap this around the spiffId section down on line ~417?

Then you don't need to do the weird typing here

Something like the below:

```vue
<DataSource
   v-if="props.mesh.meshServices.mode === 'Exclusive'"
    :src="uri(sources, '/meshes/:mesh/dataplanes/:name/layout', {
      mesh: route.params.mesh,
      name: route.params.proxy,
     })"
        v-slot="{ data: sourceDataplaneLayout }"
                        <div
                          v-if="typeof sourceDataplaneLayout !== 'undefined' && sourceDataplaneLayout?.spiffeId?.length > 0"
                        >
                          <dt>
                            {{ t('data-planes.routes.item.mtls.spiffeId.title') }}
                          </dt>
                          <dd>
                            <XBadge appearance="info">
                              {{ sourceDataplaneLayout.spiffeId }}
                            </XBadge>
                          </dd>
                        </div>
</DataSource>

```

packages/kuma-gui/src/app/legacy-data-planes/views/DataPlaneDetailView.vue
```diff
@@ -882,11 +883,12 @@ import ConnectionCard from '@/app/connections/components/connection-traffic/Conn
 import ConnectionGroup from '@/app/connections/components/connection-traffic/ConnectionGroup.vue'
 import ConnectionTraffic from '@/app/connections/components/connection-traffic/ConnectionTraffic.vue'
 import { sources as connectionSources } from '@/app/connections/sources'
-import type { DataplaneOverview, DataplaneInbound } from '@/app/data-planes/data'
+import type { DataplaneOverview, DataplaneInbound, DataplaneNetworkingLayout } from '@/app/data-planes/data'
 import { sources } from '@/app/data-planes/sources'
 import { Kri } from '@/app/kuma'
 import type { Mesh } from '@/app/meshes/data'
 import { useRoute } from '@/app/vue'
+import type { DataSourceResponse } from '@kumahq/data'
```
johncowen, line 891: These are "legacy" type utilities, and we should avoid using them unless absolutely necessary.

packages/kuma-gui/src/app/legacy-data-planes/views/DataPlaneDetailView.vue
schogges: I had the same thought, but I think it's better to keep `DataSource`s hoisted. Also the followin:

1. For some reason it would trigger double requests, potentially due to some rendering issue
2. If we add a `DataLoader` somewhere above spiffeId, it wouldn't start fetching until this data-loader is resolved
3. If we ever need layout data somewhere else in the tree we'd need to move it back up again

johncowen: K lets go with this for now.

johncowen (approved):
I'm good with this, thanks for adding the "mock tripwire"!

PR #5075, opened 2026-06-30
feat(hostname-generator): support kri links

schogges:
Use KRI to link from the `hostname-generator-list-view` to the `hostname-generator-detail-view`.

---

The linking from generic resource summary and list-view to a specific detail view will probably the last iteration on linking from KRIs, so it's not covered here. For now I'm thinking that within the protocolHandler and the default case of the big switch statement could be the link to the generic detail page.

### DIFF
--- a/packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorListView.vue
+++ b/packages/kuma-gui/src/app/hostname-generators/views/HostnameGeneratorListView.vue
@@ -92,12 +92,7 @@
                 <template #actions="{ row: item }">
                   <XActionGroup>
                     <XAction
-                      :to="{
-                        name: 'hostname-generator-detail-view',
-                        params: {
-                          kri: item.kri,
-                        },
-                      }"
+                      :href="`kri://${item.kri}`"
                     >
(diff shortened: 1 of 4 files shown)


johncowen:
> For now I'm thinking that within the protocolHandler and the default case of the big switch statement could be the link to the generic detail page.

Exactly! Did you start on this yet as I have a branch with a bit of this, I can put the PR up, lemme know

johncowen (approved):
LGTM!

schogges:
> > For now I'm thinking that within the protocolHandler and the default case of the big switch statement could be the link to the generic detail page.
> 
> Exactly! Did you start on this yet as I have a branch with a bit of this, I can put the PR up, lemme know

Yeah I have this 🙂 but won't put it up yet as I want to see what I need to do on more complex modules like dataplanes.
Edit: Just found that I also have something here in an old draft https://github.com/schogges/kuma-gui/blob/c577d6aee119c0cb5943a6ddfe1aa6d3d5c6c874/packages/kuma-gui/src/app/service-mesh/index.ts#L162-L169

PR #5079, opened 2026-07-01
fix: make plugins vue app scoped

johncowen:
Our plugins have lazy-loaded singleton services i.e.

1. Nothing gets executed if they are never "used"
2. Once "used" we reuse the same instance of the service

Previously we instantiated these services in the global scope, therefore if you want to use these plugins several times in the same page (using the same import) but in different applications, then they would share services across applications.

Whilst this is an interesting and potentially useful characteristic, it's not what we want for the moment (we might want this as an _additional_/configurable characteristic)

This PR moves the scoped to be per app instead of having a global scope.

Things to keep in mind:

- I'm still using copy/pasta on purpose, so a lot of the changes here are identical
- I'm temporarily using a little code "borrowed" from `@kumahq/container` (createInjections)
- I've wanted to have the option of a fluent builder (instead of just an almost-YAML one) for a while, this is the beginning of it.
- Side-note: This is getting closer and closer to being able to use this for our main container/builder and thus removing the brandijs dependency.

### DIFF
--- a/packages/data/src/vue/index.ts
+++ b/packages/data/src/vue/index.ts
@@ -1,10 +1,12 @@
+import { inject } from 'vue'
+
 import { DataSourcePool, create, destroy } from '../'
 // import {
 //   DataLoader,
 //   DataSink,
 //   DataSource,
 // } from './components'
-import type { Plugin } from 'vue'
+import type { App, Plugin, InjectionKey } from 'vue'
 
 export * from './components'
 
@@ -26,6 +28,7 @@ export * from './components'
 declare const typeSymbol: unique symbol
 type Uri<T = unknown> = { [typeSymbol]: T }
 type TypeOf<T> = T extends Uri<infer UriType> ? UriType : never
+type InjectionHooks<T extends Uri[]> = { [K in keyof T]: T[K] extends Uri ? () => TypeOf<T[K]> : never }
 //
 
 /* nano-container */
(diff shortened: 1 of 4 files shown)


packages/data/src/vue/index.ts
```diff
@@ -40,8 +43,39 @@ const nano = (map = new Map<Uri, unknown>()) => {
...
+type Container = ReturnType<typeof nano>
 /* */
-const { singleton, uri } = nano()
+
+const createInjections = <T extends Uri[]>(...tokens: T): InjectionHooks<T> => {
+  return tokens.map((token) => {
+    return () => {
+      type Service = TypeOf<typeof token>
+      type Getter = () => Service
+      const service = inject(token as unknown as InjectionKey<Getter>) as () => TypeOf<typeof token>
```
johncowen, line 54: I was seeing if I could make use of more Vue here (in the parts relevant to Vue) with InjectionKey, but its far from ideal, even the docs for it use lots of `as`

All in all `InjectionKey` is just a symbol (like a uri/token), so I just recast things for Vue compatibility, but I might well remove this in the future as I don't think its providing the same level of type safety that our container plus tokens does.

FYI: Its very important that the container itself has no Vue dependencies, it must be vanilla JS. So `nano` is vanilla and reusable and there is a Vue "flavour" of `createBuilder` being used here.

packages/routing/src/vue/index.ts
```diff
@@ -53,15 +85,19 @@ const tokens = {
...
+    const services = {
+      ...deps,
+      ...options,
+    }
+    const { build } = createBuilder(nano())
+    build(app)
+      .service(tokens.can, () => services.can)
+      .service(tokens.regexp, () => services.regexp)
+      .service(tokens.env, () => services.env)
+      .service(tokens.i18n, () => services.i18n)
```
johncowen, line 99: You can see how this is starting to look very similar to our main container but in a fluent style

johncowen (commented):
Notes:

johncowen:
Hmm I found that you absolutely need to install the plugin now, and I'd rather it use the defaults if its not installed.

This isn't important right now, but at first glance it looks relatively easy to tweak a tiny bit to make this happen. At some point I'd like to circle back and sort that. For the moment I've installed the plugin even for the tests.

schogges (approved):
👍

PR #5082, opened 2026-07-01
chore(kumahq/container): allows us to use multiple containers at once

johncowen:
Adds `createBuilder` function instead of `build` to let us use multiple containers in one application.

Mostly just grunt work to get this done by passing through extra arguments to thing.

Remember: I'm very close to replacing this entire file and the brandijs dependency so this is very "lets just get this done" mode, with an eye to replacing with my other better thing at some point.


schogges (approved):
LGTM

PR #5093, opened 2026-07-06
chore: make it easier to check a prod build with a prod API

johncowen:
See title. I also made some inline notes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,6 +109,7 @@ jobs:
       - uses: ./.github/actions/bootstrap/
       - env:
           KUMA_BASE_URL: 'http://localhost:5681/gui'
+          KUMA_API_URL: 'http://localhost:5681'
           KUMA_E2E_SHARD_CURRENT: ${{ matrix.container }}
           KUMA_E2E_SHARD_TOTAL: 4
         run: |
--- a/packages/kuma-gui/docs/mock_bookmarklets.html
+++ b/packages/kuma-gui/docs/mock_bookmarklets.html
@@ -1,0 +1,16 @@
+<!DOCTYPE NETSCAPE-Bookmark-file-1>
+<!-- This is an automatically generated file.
+     It will be read and overwritten.
+     DO NOT EDIT! -->
+<META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
+<TITLE>Bookmarks</TITLE>
+<H1>Bookmarks</H1>
+<DL><p>
+    <DT><H3 ADD_DATE="1773668473" LAST_MODIFIED="1778747300">Kuma Scenarios</H3>
+    <DL><p>
+        <DT><A HREF="javascript:(function()%7B((str)%3D>%7Bstr.split('%3B').map((item)%3D>item.trim()).filter((item)%3D>item!%3D%3D'').forEach((item)%3D>%7Bdocument.cookie%20%3D%20%60%24%7Bitem%7D%3BPath%3D%2F%60%3B%7D)%3Blocation.reload()%3B%7D)('KUMA_MOCK_API_ENABLED=true')%7D)()%3B" ADD_DATE="1778747300">Enable MSW</A>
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -109,6 +109,7 @@ jobs:
       - uses: ./.github/actions/bootstrap/
       - env:
           KUMA_BASE_URL: 'http://localhost:5681/gui'
+          KUMA_API_URL: 'http://localhost:5681'
```
johncowen, line 112: This isn't really necessary, but I like it that its super explicit in CI.

packages/kuma-gui/docs/mock_bookmarklets.html
```diff
@@ -0,0 +1,16 @@
...
+        <DT><A HREF="javascript:(function()%7B((str)%3D%3E%7Bstr.split('%3B').map((item)%3D%3Eitem.trim()).filter((item)%3D%3Eitem!%3D%3D'').forEach((item)%3D%3E%7Bdocument.cookie%20%3D%20%60%24%7Bitem%7D%3BPath%3D%2F%60%3B%7D)%3Blocation.reload()%3B%7D)('KUMA_MOCK_API_ENABLED=false')%7D)()%3B" ADD_DATE="1725548308">Disable MSW</A>
+        <DT><A HREF="javascript:(function()%7B((str)%3D%3E%7Bstr.split('%3B').map((item)%3D%3Eitem.trim()).filter((item)%3D%3Eitem!%3D%3D'').forEach((item)%3D%3E%7Bdocument.cookie%20%3D%20%60%24%7Bitem%7D%3BPath%3D%2F%60%3B%7D)%3Blocation.reload()%3B%7D)('KUMA_API_URL=http://localhost:5681')%7D)()%3B" ADD_DATE="1725548308">Use Kuma HTTP API</A>
+        <DT><A HREF="javascript:(function()%7B((str)%3D%3E%7Bstr.split(%27%3B%27).map((item)%3D%3Eitem.trim()).filter((item)%3D%3Eitem!%3D%3D%27%27).forEach((name)%3D%3E%7Bdocument.cookie%20%3D%20%60%24%7Bname%7D%3D%3BPath%3D%2F%3BMax-Age%3D0%60%3B%7D)%3Blocation.reload()%3B%7D)(%27KUMA_API_URL%27)%7D)()%3B" ADD_DATE="1725548308">Use Local HTTP API</A>
+    </DL><p>
+</DL><p>
```
johncowen, line 16: You can either just import these into your browser, or if you already have some bookmarklets like this, if you drag this file into your browser, you can then drag the individual bookmarklets into your browser bookmark bar for each new one you want.

Note I've renamed some bookmarklets to `Enable/Disable MSW` (so its specifically MSW not mocks seeing as we have three ways mocks can be served now)

`Use Kuma HTTP API/Use Local HTTP API` will switch you to either a live kuma at `:5681` or the local server probably at `:8080` but whereever that is.

Note: These bookmarklets do not work on a production build, even when running a production build locally via `make run/e2e`

packages/kuma-gui/vite.config.preview.ts
```diff
@@ -6,13 +6,18 @@ import { defineConfig } from 'vite'
 import type { UserConfigFn } from 'vite'
 
 export const config: UserConfigFn = () => {
+  const kuma = 'http://localhost:5680'
```
johncowen, line 9: Local prod preview will now start on `5680` so we can run a real kuma on `:5681` and use that API by default.

If for some reason you want to run a prod build against a mock server you should use

```bash
KUMA_API_URL=http://localhost:5680 make run/e2e
# or
KUMA_BASE_URL=http://localhost:5681/gui make run/e2e
```

johncowen (commented):
Notes:

packages/kuma-gui/docs/mock_bookmarklets.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE NETSCAPE-Bookmark-file-1>
+<!-- This is an automatically generated file.
+     It will be read and overwritten.
+     DO NOT EDIT! -->
```
schogges, line 4: How is this file auto-generated?

schogges (approved):

packages/kuma-gui/docs/mock_bookmarklets.html
johncowen: I just exported it from my browser and this is how it exports

johncowen: I think I can resolve this one, lemme know if not

PR #5094, opened 2026-07-06
chore: extract kuma and me modules

johncowen:
Extracts `kuma` and `me` modules to the main wiring root.

I also undid using the local `$.sources` token in some modules. Local tokens are often mass exported and can therefore overwrite "global/app" tokens if they are named the same.

`sources` usually use labels for registration anyway so we can just use local "throwaway tokens" to register them.

---

Whilst doing this, I figured I could move our `htmlVars` unit-test-only-injection and make the `testing` module generic enough to be reused, which I then moved into `@/app/vue/testing.ts`

### DIFF
--- a/packages/kuma-gui/test-support/main.ts
+++ b/packages/kuma-gui/test-support/main.ts
@@ -2,24 +2,33 @@
 // When running via vitest this file is added first using
 // vitest's `setupFiles` property, please see `/vite.config.production.ts`
 
+import { defaultKumaHtmlVars as htmlVars } from '@kumahq/config/vite'
 import { get, container, build } from '@kumahq/container'
 import { beforeEach, afterEach } from 'vitest'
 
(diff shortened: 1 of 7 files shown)


packages/kuma-gui/test-support/main.ts
```diff
@@ -2,24 +2,33 @@
...
     testing($),
+    //
+    kuma($),
+    // during testing we don't have access to the index.html vars
+    // so we inject them here so they are available during unit testing
+    [
+      [$.htmlVars, {
+        service: () => htmlVars,
+      }],
+    ],
```
johncowen, line 31: I felt that if I moved this here, it means that the testing module is generic enough to go into `@/app/vue/testing.ts` so its sharable

schogges (approved):

PR #5099, opened 2026-07-08
chore(ci): disable mocks by default everywhere apart from PR previews

johncowen:
See title

Note: locally during development you probably never need to use the KUMA_MOCK_API_ENABLED cookie anymore as it will use the local node server for mocks which is much more reliable.

I guess thew only time you might need to set it is if you happen to have no internet connection and you want the call to `https://kuma.io/latest_version` to resolve with a mock response.

In this case you can still enable the MSW mocks.

TL;DR and most importantly, we no longer have to explicitly set `KUMA_MOCK_API_ENABLED` to `false` locally. In a preview its built with the MSW mocks enabled by default (but you can turn them off whilst previewing with cookies should we need to)

### DIFF
--- a/packages/kuma-gui/README.md
+++ b/packages/kuma-gui/README.md
@@ -38,6 +38,11 @@ Install the project’s dependencies.
 make install
 ```
 
+> [!NOTE]
+> `make install` will also run every time you run `make run`, so you generally
+> don't need to use `make install`.
+
+
 ## Starting the development server
 
 Start a development server (runs on [localhost:8080](http://localhost:8080/) by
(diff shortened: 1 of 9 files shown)


schogges (commented):
I think we should also update the docs accordingly here https://github.com/kumahq/kuma-gui/blob/master/packages/kuma-gui/README.md?plain=1#L68-L72

johncowen:
@schogges FYI I amended the docs here

packages/kuma-gui/README.md
```diff
@@ -59,17 +64,28 @@ at <http://localhost:5681>.
 > [github.com/kumahq/kuma](https://github.com/kumahq/kuma/) to find out how to do
 > that.
 >
+> In order to allow you locally running GUI (on port 8080) to use the real Kuma
```
schogges:
```suggestion
> In order to allow your locally running GUI (on port 8080) to use the real Kuma
```

schogges (commented):
One nit (typo), the rest LGTM

schogges (approved):

PR #5100, opened 2026-07-08
chore(kumahq/x): add XRouter component

### LINKED ISSUE #4972: Improve linking from KRIs

johncowen:
Closes https://github.com/kumahq/kuma-gui/issues/4972

During https://github.com/kumahq/kuma-gui/pull/4930#issuecomment-4600072938 I had an idea to cope for the edge-case of there being no XActions on the page. The edge-case/only other instance where there could possibly be no XActions on the page and needing the click event listener adding was if we had `<a href>`s rendered from markdown via `XI18n`.

This PR extracts the same event listening logic into `XRouter` and composes that component into `XAction` _and_ `XI18n`.

This covers our other edge-case as we have said we should try to avoid using `<a href>`s in our application _apart from_ via markdown/i18n.

This is nice because as a user you don't even need to know about `XRouter` and things just work.

Although, if we did have a case where people _are_ using `<a href>`s and there is even a tiny chance that we think we could end up having no `XAction`s nor `XI18n`s on the page (highly unlikely but technically possible), we can just add a `XRouter` into the root of the application.

Lastly, the HTML element that the `XRouter` is "attached to" is globally configurable via a `routerElement` option for `X`. By default this uses `document.body` but we've overwritten it here to use a `.kuma-application` HTML element which is something we control and can move around if we need to.

---

Taking out of draft for the minute for early review, but I'm still testing this in other places to make sure it does what we want.


packages/x/src/components/x-router/XRouter.vue
```diff
@@ -0,0 +1,113 @@
...
+      [
+        e.metaKey,
+        e.ctrlKey,
+        e.shiftKey,
+        e.altKey,
+        e.defaultPrevented,
+      ].some(item => item)
+    ) {
+      return
+    }
```
schogges, line 43: Nice 👍

schogges (dismissed):
LGTM

johncowen:
Little note: I just realised this enables us to do something else we want to do. Dynamic route injection of non-generic resource pages.

johncowen:
@schogges I had to rebase here, plus I added a tiny bit more (in a last commit on the end)

Also, to note, I'm pretty sure there will be more to add here soon. But this gets me a little further and I can circle back a little later and add a tiny bit more (probably to do with making `push` injectable)

schogges (dismissed):

johncowen:
Had to rebase this one because we had to merge https://github.com/kumahq/kuma-gui/pull/5047 because we had to upgrade design-tokens etc etc

@schogges will need another approval, sorry!

schogges (approved):

PR #5105, opened 2026-07-09
fix(resources): fetch legacy resources by kri

### LINKED ISSUE #5086: [Resources] Can't fetch old policies by KRI
### Kuma Version

2.14

### Describe the bug

The `resource-type-list` conditionally includes old-style policies if there are any detected in mesh-insights. These policies do not support KRIs, but the resources module currently expects and only uses a KRI to fetch single resources.
We need to add a fork in the source and conditionally fetch `/meshes/:meshes/:resourceType/:name` instead.

### To Reproduce

Deploy kuma and add any old-style policy (i.e. `Retry`), go to `resources` and try to inspect the details of this policy.

### Expected behavior

_No response_

_No response_

schogges:
We currently try to fetch all resources by their KRI including resources that do not support KRIs, i.e. policies. This fixes the issue by conditionally fetching the "old-style" endpoints.

Closes https://github.com/kumahq/kuma-gui/issues/5086


packages/kuma-gui/src/app/kuma/index.ts
```diff
@@ -204,6 +207,22 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
...
+          } finally {
+            data.close(src, sym)
+          }
+        }
+
+        return fetch
+      },
+      arguments: [app.dataSourcePool],
+    }],
+
```
schogges: This is the same I did in https://github.com/kumahq/kuma-gui/pull/4894 to fetch old-style policies

packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -119,7 +119,7 @@ export const Resource = {
       ...partialResource,
       config: partialResource,
       id: partialResource.name,
-      kri: Kri.isKriString(kri) ? kri : Kri.toString({ shortName: `~${partialResource.type?.toLowerCase()}`, mesh, zone, namespace, name }),
+      kri: Kri.isKriString(kri) ? kri : Kri.toString({ shortName: `~${partialResource.type?.toLowerCase()}`, mesh, zone, namespace, name: partialResource.name.split('.')[0] }),
```
schogges: This change makes sure we can fetch resources when their name includes a hash that's not present in the displayName. ~This is possibly required for gateways.~ Builtin and delegated gateways are not in the resources.

packages/kuma-gui/src/app/resources/sources.ts
```diff
@@ -66,10 +67,23 @@ export const sources = (api: KumaApi) => {
     },
 
     '/resource/:kri': async (params) => {
+      const { kri } = params
+      const { shortName, mesh, name } = Kri.fromString(kri)
+
+      if(shortName.startsWith('~')) {
+        const fetch = useDataSource()
+        const resources = await fetch<ResourceTypeDescriptorCollection>('/resource-type-descriptors')
+        const path = resources.resources.find((resource) => shortName.includes(resource.name.toLowerCase()))?.path
```
johncowen: Q: Am I right in remembering the legacy shortNames are currently the `Type`.toLowerCase()?

johncowen: x-ref: https://github.com/kumahq/kuma-gui/pull/4894/changes#r3550853041

schogges: Yes, `type.toLowerCase()` are the legacy ones, i.e. shortName would be `~circuitbreaker`

schogges: > x-ref: https://github.com/kumahq/kuma-gui/pull/4894/changes#r3550853041

You are probably referring to https://github.com/kumahq/kuma-gui/blob/7ef9cd07aabd4aebc4afeb781c6870a2fb427ecd/packages/kuma-gui/src/app/resources/data/Resource.ts#L122 (resources)

johncowen: Would this be possible?:

When we request the `_resources` endpoint, we find any empty shortNames (these are legacy) and we fill them out ourselves with `~${path}` (we use `~path` for the shortName not `Type.lowercase`).

When we request a list of legacy policies (we know this because of the `~`), we create the KRIs for them by using the shortName from `_resources` (now "fixed/smoothed over" to have `~path` instead of `""`). This is the _only time_ we request `_resources` for mapping purposes.

We can then use this `~path` KRIs as links, and the `~path` shows up in the URLs in the browser.

When we read the URL to load a detail page, we use the shortName from the KRI (`~path`) that we have got from the URL to know its legacy and use the `path` part for the HTTP API call to load the data.

When the data comes in we use the same `~path` to recreate the KRI on the item data/record. We know this is the correct `~path` and is "safe to use" because otherwise it would 404.

All in all this means we request the resources endpoint for mapping only purposes _once_, when we get the list.

---

Does the above sound possible/like it would work?

johncowen: P.S. I feel like this is probably mixed in somehow with your other policies PR - not sure

schogges: > When we request a list of legacy policies (we know this because of the ~)

I don't think we know this because the `path` doesn't include the `~` only the shortName.

johncowen: Ah, so currently it doesn't but it should i.e. we route everything by shortName/KRI, we map _once_ when the data comes in and thats it.

Is this PR for 2.14?

schogges: This is the PR for main

schogges:
@johncowen I would like to get this in as soon as possible. This is fixing a bug that needs to get backported to 2.14. I have just realised that the bug also appears for `ExternalServices` which are listed in the resource-types listing (left panel) and not only policies. This is fixing the issue for both, basically any resource that is not supporting KRIs yet.

We share the same concerns regarding this extra imperative fetch in the source, however this is focusing on fixing the bug. I have this draft https://github.com/kumahq/kuma-gui/pull/4825 open for a while now and I'd like to finish it after all the issues that need to be backported have been fixed. This should make the changes in the sources obsolete and I can revert them. But this needs some more work and I want to be cautious to keep all legacy resources covered. Also https://github.com/kumahq/kuma-gui/pull/4894 brings us another step closer to the final solution.

johncowen:
Gotcha 👍 FYI: I'm keeping an eye on the patch release timelines.

> I have just realised that the bug also appears for ExternalServices which are listed in the resource-types listing (left panel) and not only policies.

Oh! good spot!

Let me take another look. I've been thinking some more about this and I've realized some other things, so my thoughts on this have changed a little, but not totally as of now. That may change.

---

Can I just confirm your proposed plan is to:

1. Merge this to master
2. Cherry pick the change to release-2.14
3. Immediately revert the change we made on master that we cherry picked to 2.14

?

johncowen:
Just wanted to add:

I believe 2.13 patch releases are going out soon and this PR isn't blocking that.

I'd understood that there was no particular rush for a 2.14 patch release.

johncowen:
Q: Is it an option to not show old style things in the new Resources tab?

My thinking being:

- Old policies are visible in the `[Policies]` tab
- ExternalServices are visible in the `[Services]` tab

Q: Are old-style Services shown in the new Resources tab currently? I saw you said Gateways aren't

> Builtin and delegated gateways are not in the resources.

schogges:
> Just wanted to add:
>
> I believe 2.13 patch releases are going out soon and this PR isn't blocking that.
>
> I'd understood that there was no particular rush for a 2.14 patch release.

Oh, I thought it's 2.14 patch releases. I have just confirmed with backend folks and you are right that we got some more time to get backports ready for 2.14. However, this PR fixes the immediate issue and we don't need to worry about 2.14 anymore.

> Q: Is it an option to not show old style things in the new Resources tab?
> 
> My thinking being:
> 
> * Old policies are visible in the `[Policies]` tab
> * ExternalServices are visible in the `[Services]` tab

This would potentially clash with the decision we want to take on wether we want to replace Policies with Resources or wether we want to keep them both. At least elsewhere.

> Q: Are old-style Services shown in the new Resources tab currently? I saw you said Gateways aren't
> 
> > Builtin and delegated gateways are not in the resources.

We show everything that is being returned by `/_resources`, except some exceptions like secrets. Old style internal services are not included, but external services are included.

schogges:
> Can I just confirm your proposed plan is to:
> 
> 1. Merge this to master
> 2. Cherry pick the change to release-2.14
> 3. Immediately revert the change we made on master that we cherry picked to 2.14
> 
> ?

1. Merge this to master
2. Cherry pick the change to release-2.14
3. Get https://github.com/kumahq/kuma-gui/pull/4894 merged
4. Get back to https://github.com/kumahq/kuma-gui/pull/4825 to ideally revert the changes specifically in the sources (imperative fetch)

johncowen:
> We show everything that is being returned by /_resources, except some exceptions like secrets. Old style internal services are not included, but external services are included.

Gotcha thanks!

Ok so it makes even more sense to me that we don't show ExternalServices in the new `Resources` tab. We don't show normal old-style services (or gateways) there so in my mind we shouldn't really show old-style ExternalServices either.

> This would potentially clash with the decision we want to take on wether we want to replace Policies with Resources or wether we want to keep them both. At least elsewhere.

Totally get you. We don't have the Resources tab right now anywhere else, so this isn't a problem at the moment.

This is only an issue for getting something shipped as an imminent 2.14 patch.

If we filter out all the legacy policies _and_ external services from the new Resources tab. For those that are still using legacy policies and/or ExternalServices can still access them as they did previously.

We can then decide whether we want to bother adding a "passthrough" for legacy policies and/or external services so it shows the actual existing detail views for those instead of the generic resource pages. We have plenty of time to decide that.

Wdyt? Let me know if I missed something

schogges:
You are right, let's not waste energy in trying to support something that is going to be entirely removed in 3.0. I'll refactor it to ignore old-style resources in the resources tab.

schogges:
@johncowen ok, this should be ready to go. Very straightforward now by just filtering out all resources that have no/empty shortName.

packages/kuma-gui/src/app/resources/views/ResourceTypeListView.vue
```diff
@@ -75,10 +75,8 @@
                   >
                     <XLayout
                       v-for="filtered in [resources.resources.filter((item) =>
-                        (route.params.category === 'all' || item.categories.includes(route.params.category)) &&
-                        (!item.policy || item.policy?.isTargetRef || hasLegacyPolicies) &&
-                        (item.name !== 'ExternalService' || meshInsight.services.external > 0)),
-                      ]"
+                        (route.params.category === 'all' || item.categories.includes(route.params.category)) && item.shortName.length > 0,
+                      )]"
```
johncowen: Nice! So we only show in the resources page if the `_resources` endpoint tells us the resource has a `shortName` i.e. supports KRIs 👍 .

johncowen (dismissed):

schogges:
@johncowen I made another minor cleanup and ensured that in the list-view the legacy resources are ignored too.

johncowen (approved):
👍

PR #5122, opened 2026-07-15
fix(resources): ensure global resource counts are shown correctly

johncowen:
Fixes resource counts for meshes, zones, and zone proxies

## Before

Screenshot 2026-07-15 at 11 56 43

## After

Screenshot 2026-07-15 at 11 56 55


schogges (dismissed):
Thanks 👍 I'm wondering if anything has changed in the backend, because I'm pretty sure I've seen this working before 🤔 I'd prefer to have all resources under `resources` and not spread across multiple top-level entries...

johncowen:
Drafting because this isn't urgent because the patch has only just gone out, we need to concentrate on other things for a few days.

schogges (approved):

PR #5123, opened 2026-07-15
chore(resources): use shortNames for URLs for resources

johncowen:
Moves all resource URLs to use shortNames instead of paths in URLs.

i.e. `mal` instead of `meshaccesslog`
i.e. `~circuit-breakers` (squiggle plus path) instead of `~circuitbreaker` (squiggle plus lowercase type)

This means that aswell as being able to easier link to a page for a single resource really easily (just use the `kri`), given the single resource its now really easy to link to a list of the same resource really easily (just use the `shortName` from the `kri`)

Legacy policies are now back in the `Resources` tab, and whilst I know that we aren't supporting these moving forwards, we still need to provide support for them longer term. I would much rather both legacy and "modern" resources all work the same way from a linking/URL perspective, and now they do. They all use KRIs for everything.

As a side-effect and slight but kinda unimportant benefit, for detail pages for resources we've removed the path/shortName segment from the URL which makes the URL slightly shorter. Its now just `meshes/default/resources/:kri` instead of `meshes/default/resources/potentially-really-long-path-name/:kri`

---

Separately I also altered the loading of resource counts slightly as I found that we don't show the list of resources until:

- The list of resources is loaded, plus:
- The counts of resources is loaded from global insights, plus:
- The counts of resources is loaded from mesh insights

With this PR we don't show a list of resource until:

- The list of resources is loaded

Thats it, we still load the counts of things but we don't block showing the listing on those, so if they take a long time to load or error, the user can still use the navigation to view resources.


packages/kuma-gui/src/app/kuma/kri.ts
```diff
@@ -45,3 +45,62 @@ export class Kri {
...
+      //   kri: KRI.stringify({ shortName: 'm', name}),
+      // }))(shortName === 'm' ? name : mesh),
+      // zone: (name => ({
+      //   name,
+      //   kri: KRI.stringify({ shortName: 'z', name}),
+      // }))(shortName === 'z' ? name : zone),
+      // namespace: (name => ({
+      //   name,
+      //   kri: KRI.stringify({ shortName: '~nspace', name}),
+      // }))(namespace),
```
johncowen, line 101: I've left this in but commented out because I'd like to do this in a followup to avoid this PR doing too much.

But with the above uncommented we no longer have to use `Kri.toString` in view files to link to the mesh or zone of a resource, we can just refer to the `zone.name` and `zone.kri` to make the link.

packages/kuma-gui/src/app/policies/views/PolicyListView.vue
```diff
@@ -12,14 +12,9 @@
...
     <DataCollection
-      :predicate="(policyType) => typeof policyType !== 'undefined' && policyType.path === route.params.policyPath"
+      :predicate="(policyType) => policyType.path === route.params.policyPath"
       :items="props.policyTypes ?? []"
     >
-      <template #empty>
-        <XEmptyState>
-          {{ t('policies.routes.items.empty') }}
-        </XEmptyState>
-      </template>
```
johncowen, line 22: In the parent route there is logic preventing this from ever being empty

packages/kuma-gui/src/app/resources/data/Resource.ts
```diff
@@ -106,36 +106,27 @@ export const Resource = {
...
-      kri: Kri.isKriString(kri) ? kri : Kri.toString({ shortName: `~${partialResource.type?.toLowerCase()}`, mesh, zone, namespace, name }),
-      mesh,
-      namespace,
-      name,
-      zone,
-      creationTime: 'creationTime' in partialResource ? partialResource.creationTime ?? '' : '',
-      modificationTime: 'modificationTime' in partialResource ? partialResource.modificationTime ?? '' : '',
-      labels: partialResource.labels ?? {},
+      ...item,
+      ...KRI.split(item.kri ?? KRI.stringify({ shortName, ...item, labels })),
```
johncowen, line 116: I called this `split` as its primary function is to split the KRI into several more KRIs, but then I figured I could also re-return the main KRI itself, plus other things. This may or may not change. I'd like to hear opinions.

johncowen (commented):
Notes: I'll be adding more tomorrow

packages/kuma-gui/src/app/kuma/kri.ts
```diff
@@ -45,3 +45,62 @@ export class Kri {
...
+          namespace,
+          name,
+        })
+      }
+
+      default:
+        return Kri.toString(item)
+    }
+  },
+  split: (kri: string) => {
```
johncowen, line 78: The idea with `split` is that moving forwards every resource we load into the GUI would use `KRI.split` to splat down all "identifier" type things onto the resource object consistently across all resources

packages/kuma-gui/src/app/resources/data/Resource.ts
johncowen: As mentioned above `KRI.split` splats down all "identifier" type things that are taken/split out from a KRI, this includes:

1. Names
2. KRIs
3. shortNames
4. meshes (both name and KRI to be able to link to the mesh)
5. zones(both name and KRI to be able to link to the zone)
6. namespaces (both name and KRI, we don't have a page to link to as yet/if ever)
7. old style `name.namespace` `id`s that we will eventually phase out.

packages/kuma-gui/src/app/resources/data/ResourceTypeDescriptor.ts
```diff
@@ -46,11 +46,12 @@ const resourceCategories = new Map<string, string[]>([
...
 
 export const ResourceTypeDescriptor = {
-  fromObject(partialResource: KumaResourceTypeDescriptor) {
+  fromObject(item: KumaResourceTypeDescriptor) {
     return {
-      ...partialResource,
-      categories: resourceCategories.get(partialResource.name.toLowerCase()) ?? [],
-      group: ('policy' in partialResource ? 'policy' : (partialResource.scope ?? 'others')).toLowerCase(),
+      ...item,
+      shortName: item.shortName || `~${item.path}`,
```
johncowen, line 52: All resources form the `_resources` endpoint are now guaranteed to have a non-empty `shortName` property, even legacy policies.

packages/kuma-http-api/mocks/src/_resources.ts
```diff
@@ -677,17 +697,17 @@ function getResources(): Resources['resources'] {
       'name': 'Zone',
       'path': 'zones',
       'pluralDisplayName': 'Zones',
-      'readOnly': false,
+      'readOnly': true,
       'scope': 'Global',
-      'shortName': '',
+      'shortName': 'z',
```
johncowen, line 702: I took a new copy/paste of a real resource response, because I found that the last time we did this the shortName from a zone was missing.

I have no idea why it flipped readOnly, but we don't use that anyway.

packages/kuma-http-api/mocks/src/meshes.ts
```diff
@@ -1,23 +1,34 @@
...
-          creationTime: '2020-06-19T12:18:02.097986-04:00',
-          modificationTime: '2020-06-19T12:18:02.097986-04:00',
+          name,
+          kri: fake.kuma.kri({ resourceName: 'Mesh', mesh: '', zone: '', namespace, name: displayName, sectionName: '' }),
+          ...fake.kuma.timespan(),
+          labels: {
+            ...fake.kuma.labels({
+              name: displayName,
+              // ...(zone ? { zone } : {}),
+              // ...(k8s ? { namespace } : {}),
```
johncowen, line 29: I think meshes never have zone or namespace labels? Did I remember right?

packages/x/src/components/x-progress/XProgress.vue
```diff
@@ -31,6 +31,11 @@
     data-testid="line-skeleton"
     width="10"
   />
+  <KSkeletonBox
+    v-else-if="props.variant === 'count'"
+    data-testid="line-skeleton"
+    width="1"
+  />
```
johncowen, line 38: Adds a new XProgress variant for the counts. We might want a better name but we can always change it later if we want.

johncowen (commented):
Ok I filled out the PR description here and added a bunch more inline notes, we also have a short sync on this.

schogges:
I've had a thought on this. For now we only support resources that support KRIs and no legacy ones. I think we will stick to that. But the changes in here might be more relevant for the policies. I have started to extract the important pieces into separate PRs (#5140 and potentially more), which is to shorten the URLs. If we decide to support legacy resources in the resources, we might integrate the rest of this PR on top, which should be non-breaking changes after all.

johncowen:
I'm gonna put this in Draft for the moment, just catching up on what has gone in from here, but I think at the very least if we need to add something else from here it will be a new PR, I'll probably close this soonish, but in Draft just incase.

johncowen:
I'm gonna close this, I'm not super clear right now what went in from this and what didn't, either way if we want to add anything more from here it won't be via merging this particular PR. If I close this PR I can still come back and reference it to reimplement anything we want (if there is anything we want that is)

PR #5128, opened 2026-07-17
fix(fake-api/vite): status code

schogges:
I've noticed that when setting a `Status-Code` header in the mocks, it's not properly passed through the vite middleware.

**Before**

image

**After**

image


packages/fake-api/src/vite.ts
```diff
@@ -54,6 +54,7 @@ export default <TDependencies extends object = {}>(opts: PluginOptions<TDependen
         const type = response.headers.get('Content-Type') ?? 'application/json'
         res.setHeader('Content-Type', type)
         res.setHeader('Status-Code', response.headers.get('Status-Code') ?? '200')
+        res.statusCode = parseInt(`${res.getHeader('Status-Code')}`)
```
johncowen: Thanks for sorting 👍 

---

nit: How come we need to use the backticks here, headers are already strings aren't they?

If we _are_ making a tiny tweak here could we do the following:

```ts
const type = response.headers.get('Content-Type') ?? 'application/json'
const statusCode = response.headers.get('Status-Code') ?? '200'

res.setHeader('Content-Type', type)
res.setHeader('Status-Code', statusCode)
res.statusCode = parseInt(statusCode)
```

Then its the same style as how we do `type` i.e. normalize/default everything first, then apply it.

I think the other reason why I prefer this approach is that setting the header and then getting the header again is passing the value through something we don't control - who knows if it gets mutated somehow in between.

Anyway its a bit of a nit, but thought I'd explain why I prefer this style.

schogges: Done

johncowen (approved):
👍 thanks Mo!

PR #5130, opened 2026-07-17
chore(@kumahq/x): add push and href injectable functions

johncowen:
Adds `useHref` and `usePush` and allows you to inject the implementation of these per application.

- `useHref` returns a `href({to: ...})` function (pretty much `router.resolve`)
- `usePush` returns a `push` function (pretty much `router.push`)

### DIFF
--- a/packages/kuma-gui/src/app/kuma/index.ts
+++ b/packages/kuma-gui/src/app/kuma/index.ts
@@ -15,7 +15,7 @@ import KumaApi from '@/app/kuma/services/kuma-api/KumaApi'
 import { RestClient } from '@/app/kuma/services/kuma-api/RestClient'
 import type { ServiceDefinition } from '@kumahq/container'
 import type { DataSourcePool } from '@kumahq/data'
-import type { Router } from 'vue-router'
+import type { Router, RouteLocationAsRelative } from 'vue-router'
 
 export * from './utils'
 export { Kri } from './kri'
@@ -27,12 +27,6 @@ declare module 'vue' {
     KumaPort: typeof KumaPort
     KumaTargetRef: typeof KumaTargetRef
   }
-}
-export const TOKENS = {
-  httpClient: token<RestClient>('httpClient'),
-  api: token<KumaApi>('KumaApi'),
-  htmlVars: token('kuma.html.vars'),
-  dataSource: token<<T>(src: string) => Promise<T>>('app.dataSource'),
 }
 function getConfig() {
   const pathConfigNode = document.querySelector('#kuma-config')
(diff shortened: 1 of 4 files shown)


packages/kuma-gui/src/app/kuma/index.ts
```diff
@@ -153,24 +148,57 @@ const protocolHandler = (can: Can) => {
...
+    }
+    console.error(e)
+  }
+  return ''
+}
+const push = (router: Router) => (href: string) => {
+  const base = router.options.history.base
+  const h = href.startsWith(base) ? href.substring(base.length) : href
+  return router.push(h.length > 0 ? h : '/')
+}
```
johncowen, line 169: `href` and `push` (a.k.a. router.resolve and router.push) are now completely controlled injected by the application. If the application needs to do something weird with these (such as do something extra if routes don't exist) then they can.

packages/kuma-gui/src/app/kuma/index.ts
```diff
@@ -179,6 +207,21 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
     [app.storagePrefix, {
       service: () => 'kumahq.kuma-gui',
     }],
+    [app.routerElement, {
+      service: () => () => document.querySelector('.kuma-application'),
```
johncowen, line 210: This is now injectable also, in case you want to change this per application.

packages/kuma-gui/src/app/kuma/index.ts
```diff
@@ -264,9 +307,7 @@ export const services = (app: Record<string, Token>): ServiceDefinition[] => {
   ]
 }
 export const [
-  useKumaApi,
```
johncowen, line 267: This is super old and we never call `useKumaApi` anymore, doesn't _need_ to be done here but while I was here I removed it.

packages/x/src/index.ts
```diff
@@ -219,6 +219,10 @@ const deps = {
   },
   routerElement: () => document.body,
   protocolHandler: (href: string) => href,
+  href: (_to: unknown) => {
+    return ''
+  },
+  push: (_href: string) => {},
```
johncowen, line 225: By default `useHref` and `usePush` don't do anything, `kuma` injects its own impl. which are just `vue-router` `.resolve` and `.push`

packages/x/src/index.ts
```diff
@@ -271,11 +279,15 @@ export const [
   useI18n,
   useSyntaxHighlighter,
   useProtocolHandler,
+  useHref,
+  usePush,
```
johncowen, line 283: These are possible candidates for "exporting" from RouteView. Not necessary yet but it might be helpful to have them there in the future

johncowen (commented):
Notes:

schogges (dismissed):
LGTM

schogges (dismissed):

johncowen:
Holding merge for a bit, might merge Monday

johncowen:
@schogges I had to rebase this, needs another approval 👍

schogges (approved):

PR #5222, opened 2026-08-17
fix(kongponents): override default links styles for table actions

johncowen:
We use accessible links for our click/navigation functionality for table rows.

From what I can see, others generally use non-accessible dev/onclick handlers for the rows.

This difference means we get the "secondary link" styling on our "primary links" for table rows.

I'd rather keep using accessible links instead of re-inventing wheels with onclick handlers instead of native links, so for the moment at least I've overridden the default link styling that was added recently. This makes everything look the same as elsewhere.

If elsewhere changes their look again soon then we can just revert this again, but I have a feeling that won't happen.


schogges (approved):
👍

PR #5224, opened 2026-08-17
feat(kuma-http-api): add v2 oas

schogges:
TBD


schogges:
https://github.com/kumahq/kuma-gui/pull/5268 supersedes this one. The result is kind of the same but https://github.com/kumahq/kuma-gui/pull/5268 is less code to maintain. Closing this one.

PR #5226, opened 2026-08-18
chore(search): move search from resources to kuma

johncowen:
TL;DR: move our generic/abstract `search` out of `resources` (now a non-abstract module) to `kuma`

---

Originally our `resources` module (mainly its data layer) was like an abstract class for any resource type to inherit from. Therefore we had a generic resource `search` function attached to it: `Resource.search`.

Since then we've added a `Resource` resource, so the `resources` module being an "abstract" class no longer made sense.

Therefore we decided we would move the `search` function off of `Resource` to somewhere else.

That's what this PR does, although for the moment it keeps a passthrough from `Resource.search` to `search`, apart from for the new `legacy-services` module (which is mainly for illustration purposes of what is to come in a follow up PR).

I chose to move the `search` code into `kuma` seeing that is now our "we use this everywhere" bucket/module.

Depending on my workload, I plan on migrating all other modules away from using `Resources.search` so that they just use `search` instead, imported from `kuma`.


schogges (approved):
👍

PR #5227, opened 2026-08-18
chore: move StatusBadge > KumaStatusBadge

johncowen:
See title.

Consequently as `Kuma*` components are global (but namespaced with `Kuma`) we no longer need to import these.

This is part of finally removing the `common` module that we have random components in.


schogges (approved):
👍

PR #5230, opened 2026-08-18
chore: move Accordion to XAccordion

johncowen:
It kinda pains me to do this PR but generic UI components belong in `x` no matter in what state they are in.

I think `AccordionList/AccordionItem` are our last generic components that aren't in `x` because I've always been of the mindset to improve/update super old components before moving them to `x`. Thing is I really want to get rid of our `common` module asap and this is one of the last remaining things to do.

Whilst I really don't like our Accordion (it doesn't even save its open/closed states to the URL) I decided to move it without messing with its internals or anything, just to be able to stand a chance of removing `common` entirely, which right now is more valuable to me.


packages/kuma-gui/src/app/common/AccordionList.spec.ts
schogges, line 1: Q: Why are we removing these tests?

schogges (commented):
Any opinions on integrating [`KCollapse`](https://kongponents.konghq.com/components/collapse.html)?

johncowen:
> Any opinions on integrating [KCollapse](https://kongponents.konghq.com/components/collapse.html)?

One day (maybe maybe not) we will re-build our accordions, but our current ones date from over 4 years ago so priority wise rebuilding them is way down on the list as much as I'd like to do it.

Right now I'm aim is to get rid of the `/common/` bucket so we don't have random components that may or may not be used anywhere in this repository.

We should end up with `/x/` (generic `X*` GUI components), `/kuma/` (`Kuma*` components used in several places across the product). Then anything else is in the module it belongs to. 

Taking into account we very rarely have non-route components `Kuma` and any module components should be compatively rare, but at least we won't have things like a `TargetRef` component with no namespace (like `KumaTargetRef` or `XTargetRef`) that is only used in say the `gateways` module (and that fact being very unclear)

packages/kuma-gui/src/app/common/AccordionList.spec.ts
johncowen: There's only 3 tests here, and in my opinion they aren't very valuable.

Two of the 3 tests test API we don't even need/use. The final test doesn't seem very valuable either, in fact I just realised it also tests API we don't need/use.

I guess my biggest thing is this component is basically a visualisation of flipping a boolean. I'd rather just remove the tests. If we ever re-wrote this, we wouldn't use the same tests.

schogges (approved):

PR #5232, opened 2026-08-18
chore(kumahq/x): remove old mr/ml helper classes

johncowen:
I found we could remove the `mr`/`ml` helper classes.

---

I would like to remove the `mt`/`mb` ones also but we have a few usages of them still dotted around. Depending on my workload I may try and get rid of the last few and there we don't have to support them in eternity (just use XLayout instead)


schogges (approved):
👍 🧹

PR #5233, opened 2026-08-18
chore: remove some old types

johncowen:
I found I could remove a bunch of old types. Mostly `ControlPlaneConfig` which it looks to me like we don't ever use, but also I removed some legacy-service related types by either removing them or copy pasting them into the respective legacy modules. This result in a duplicate, but I'd rather than than wondering where these types get used.


schogges (approved):
Nice 🧹

PR #5234, opened 2026-08-19
fix(ci): use cache for playwright install

johncowen:
TBD

FYI: Waiting for CI first to make sure this PR can find a cached playwright to use


johncowen:
Screenshot 2026-08-19 at 11 19 36

Ah well I was hoping it would pick it up from the cache, closing this. Ideally we'd cache a lot of this stuff, but it feels like now is not the right time to try this out seeing as GitHub is being Github. Closing this for now

PR #5237, opened 2026-08-20
chore(deps): remove @types/semver

johncowen:
I found that whilst we depend on `@types/semver` we don't actually depend on `semver` itself. I do remember years back we distinctly removed `semver`, I guess this has been unnecessarily in there since then 🤷


schogges (approved):
👍 🧹

johncowen:
This is gonna need a little tweak elsewhere, as this isn't blocking anything right now I'll sort this later on.

PR #5241, opened 2026-08-21
fix: fixes table row action dropdown styling

johncowen:
### Before

Screenshot 2026-08-21 at 12 25 27

---

### After

Screenshot 2026-08-21 at 12 24 34


schogges (approved):
👍

PR #5248, opened 2026-08-25
chore: remove zone proxies

johncowen:
See title


johncowen:
Opening this one up for review seeing as we should be unblocked here soon

schogges (approved):
LGTM

PR #5249, opened 2026-08-25
chore(kri): use KRI links for service-insight links

johncowen:
In https://github.com/kumahq/kuma-gui/pull/5189 `useDataPlaneServiceLinks` was added:

> In some tables we are rendering (service) links to the service-detail-view. These are now also injectable and default to rendering plain text. I have covered this in new tests, as we had only assertions for the text being rendered but no checking for any links.

Our `kri://` protocol links are already injectable and will default to plain text if a certain KRI shortName isn't supported.

We should use `kri://` protocol links everywhere we can. This PR replaces the `useDataPlaneServiceLinks` with `kri://` links. Bit of an awkward PR but hopefully folks see my reasoning.

---

Note: There are a few PRs up around removing zone-ingresses (https://github.com/kumahq/kuma-gui/pull/5248), I've done this PR imagining we haven't done that yet, and if the removal PR goes in before this one, the zone-ingress portion will have to be removed here also.

Additionally, we are soon to be removing `gateways` and `legacy-services`, ideally this PR would go in before then, but if not and they are removed first those portions here will also have to be removed.

I think my ideal order of work would be:

- Merge https://github.com/kumahq/kuma-gui/pull/5248
- Remove ZoneIngress stuff from here
- Merge this PR
- Continue to remove gateways and legacy-services

I might leave this in draft for a bit until I know its a good time to merge this, please don't let that hold off on review. Most importantly, I might have totally missed a reason for not using `kri://` links here in the first place. If I have missed something please let me know and we can close this.


packages/kuma-gui/src/app/kuma/index.ts
```diff
@@ -121,6 +121,32 @@ const protocolHandler = (can: Can, router: Router) => {
...
               }
+            /// for legacy TargetRef component reasons we default to the current mesh
+            case shortName === '~standard':
+              return {
+                name: 'service-detail-view',
+                params: {
+                  mesh: mesh || router.currentRoute.value.params.mesh,
+                  service: name,
+                },
+              }
```
schogges, line 132: From the point of view of the KRI protocol handler the `shortName` `~standard` is not clear enough that it refers to the dataplane type. The dataplane type should not be used in a KRI for a service. `~standard` could be related to anything. The `shortName` of a KRI should always be the `shortName` of the resource type or at least make it clear what resource type it is.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
```diff
@@ -41,15 +41,26 @@ export const DataplaneOverview = {
...
+    const dataplaneType = (() => {
+      switch (true) {
+        case networking.type === 'gateway':
+          return dpTypes.builtin
+        case typeof networking.gateway !== 'undefined':
+          return dpTypes.delegated
+        default:
+          return dpTypes.standard
+      }
+    })()
```
schogges, line 53: Ideally this bit is also injectable, i.e. if the `gateways` module is registered a dataplane **can** be of type `internal` or `delegated`. If not, a dataplane can only be of type `standard`.
During the extraction of the gateways logic I decided to leave this bit untouched for now because it's rather a micro improvement. But ideally this can be injected.

schogges (commented):
The main idea for the `useDataPlaneServiceLinks` is that each module keeps the relevant information and adding/removing a module wouldn't need changes in other places. If we want to switch to KRIs I'd love to achieve the same result.

johncowen:
> The main idea for the useDataPlaneServiceLinks is that each module keeps the relevant information and adding/removing a module wouldn't need changes in other places. If we want to switch to KRIs I'd love to achieve the same result.

Yes exactly! For the moment we just have a big switch statement, but those can/will be injected almost like multiple `protocolHandlers`. Its a very similar approach to what you have here, we build a list of "KRI handlers" via label injection and iterate through that instead of the big hardcoded switch statement. There is no massive rush to do this yet, but that doesn't mean we can't do this now (or later).

My main thing is everything should work with KRIs and all the functionality to do that is exists already (including not rendering a link if a KRI isn't "registered"), so I want to try and avoid having several implementations of "render links that sometimes shouldn't be rendered or we don't know about"

packages/kuma-gui/src/app/kuma/index.ts
johncowen: I'm happy to change these whenever, I kinda don't care what they are as they are legacy names just so that everything uses the same approach, they aren't exposed to users (unlike legacy policy shortnames).

I just used the strings for dataplane types that we already use.

If you had 3 different terms/shortNames in mind happy to replace these, but from what I remember it was much easier/less complicated from a code perspective to just use the existing dataplaneTypes for the string.

packages/kuma-gui/src/app/data-planes/data/DataplaneOverview.ts
johncowen: Probably, but I did consider merging gateways into legacy-services, but figured we should just assume they are always installed together.

packages/kuma-gui/src/app/kuma/index.ts
johncowen: Shall we use `~sidecar`? Or `~service-insight`

Lets go with `service-insight` and we leave the `delegated`/`builtin` as they are 👍

johncowen:
Lets remove `zone-ingress`, `zone-egress`, `legacy-services` and `gateways` before we come back to apply this work.

johncowen:
Part of replacing this is here https://github.com/kumahq/kuma-gui/pull/5308, gonna close this I think I can always come back to the closed PR to use for reference

PR #5267, opened 2026-09-02
chore(services): move MeshService to use dataplaneLabels w/fallback

johncowen:
Moves `MeshService` to use `dataplaneLabels.matchLabels` instead of `dataplaneTags`.

For the moment, perhaps temporarily, we fallback to the old `dataplaneTags`.

---

I also noticed that these can contain things like `kima.io/mesh` so I made them conditionally link depending on that, just like `labels` everywhere else.

I also removed the truncation, truncate should only be used in height restricted space (such as table rows), for the summary and detail pages we have as much vertical space as we like because the user can use the native scroll bar to see more.


schogges (approved):
👍

PR #5268, opened 2026-09-02
chore(kuma-http-api): moves index.d.ts to v2.d.ts and re-points

johncowen:
Moves moves `kuma-http-api/index.d.ts` to `kuma-http-api/v2.d.ts` and re-points `kuma-gui` types via the package.json file in `kuma-http-api`.

---

The idea is our `kuma-http-api` `make build` script will emit the new v3 types to `kuma-http-api/index.d.ts` and we can easily check v2/v3 compatibility and/or figure out what needs moving/removing easily by just changing `kuma-http-api/v2.d.ts`

Things to note:

- `kuma-http-api/v2.d.ts` no longer gets updated, neither locally nor via CI. i.e. it is frozen. Any changes needed here would happen on the `release-v2.14` branch, not `master`
- `kuma-http-api/v2.d.ts` is just a short term thing to help us move between v2 and v3, it will soon be deleted completely.
- I almost did a rebuild of `kuma-http-api` against `kuma` `master` to give us v3 types in `index.d.ts`, but I noticed that it might be helpful to easily see a diff between `v2.d.ts` and `index.d.ts`, and if I committed `index.d.ts` (v3 types) here it wouldn't be as easy because a full set of changes would be spread over several interspersed commits. Once we get closer to being finished (not necessarily totally finished) with the v2 > v2 migration, we can commit in the new `index.d.ts`.

Finally I think this is an alternative to https://github.com/kumahq/kuma-gui/pull/5224 which we were keeping hanging around in case we wanted to merge that to help us with the migration. The thing is v2 is finished. Is/should be frozen. So its easy enough (IMO) to just commit a `v2.d.ts` and not bother providing build/CI support etc. Let me know if you feel like there are any issues doing this, but I can't think of any considering `v2.d.ts` should only be around for a matter of a few weeks.


johncowen:
I just thought as a follow up we could maybe add the v3 `index.d.ts` and get a non-blocking lint check against it in CI to help preview. Although it might be to much hassle for something that is only gonna be like this for a matter of a week or 2.

schogges:
> I just thought as a follow up we could maybe add the v3 `index.d.ts` and get a non-blocking lint check against it in CI to help preview. Although it might be to much hassle for something that is only gonna be like this for a matter of a week or 2.

At some point we'd have to do this anyways, no?

johncowen:
> At some point we'd have to do this anyways, no?

My suggestion here was to potentially have the current CI, which checks the types against v2 and fails if anything is wrong, and _add_ a non-failing check against v3. i.e. have two lint runs one for each version.

I'm in two minds about doing that though because we are almost finished with the transition its hardly worth setting up an _additional_ check.

---

What we'd have to do anyway is, once we are fully on v3 run the hard fail CI/lint against v3 types. We should eventually no longer need to type check against v2. We should no longer have to worry about v2 at all. It's gone, we don't need to support it here (eventually).

schogges (approved):

PR #5269, opened 2026-09-02
chore(meshes): make mesh.mtlsBackend "transitional"

johncowen:
We are slowly getting rid of old v2 props, but there are certain things which will be a little tricky initially.

I found I could remove `metricsBackend` and `routing` entirely, it looks to me like we don't use them anywhere, plus they are super old so it doesn't suprise me that we moved to something else a long time back.

`mtlsBackend` is a little trickier so for the moment I did some data reshaping to continue to support the v2 shape plus the v3 shape. `mtlsBackend` itself seems to be used in a few places, which I think we can carefully remove at a later date.

For the moment this should help unblock us without too much churn.

I added some `@TODO(types)` comments to help us find these later, and I also used https://github.com/kumahq/kuma-gui/pull/5268 to help me to check things.

Oh also the new types I added are just copy/pasted from the old `types/index.d.ts` file - eventually we'll just remove this old file.


packages/kuma-gui/src/app/meshes/data/Mesh.ts
```diff
@@ -14,16 +26,14 @@ export const Mesh = {
...
-      routing: ((item = {}) => {
-        return {
-          ...item,
-          zoneEgress: typeof item.zoneEgress !== 'undefined' ? item.zoneEgress : false,
+      // @TODO(types) provide support for v2 and v3
+      mtlsBackend: ((mtls: MeshBackend | undefined) => {
+        if(typeof mtls === 'undefined') {
+          return
         }
-      })(item.routing),
```
johncowen, line 26: Just to mention a a few properties (metricsBackend and routing) were completely removed here, we don't use them anywhere currently (and therefore we'll never use them in v2) so I just got rid.

packages/kuma-gui/src/app/meshes/data/Mesh.ts
```diff
@@ -1,5 +1,17 @@
...
+// @TODO(types) provide support for v2 and v3
+type Backend = {
+  name: string
+  type: string
+}
+type MeshBackend = {
+  enabledBackend?: string
+  defaultBackend?: string
+  backends?: Backend[]
+}
```
johncowen, line 12: I just copied these out of the old types/index.d.ts file

schogges (approved):
👍

PR #5275, opened 2026-09-03
chore(meshes): make meshServices.mode "transitional"

johncowen:
TL;DR

We hide meshServices.mode behind `can` features.

`legacy-services` looks at `meshServices.mode` (v2)
`services` just returns hardcoded booleans (v3)

As long as we use `can('use service-insights')` and `can('use mesh-services')` everywhere instead of referring to `meshServices.mode` then when we remove legacy services everything should work as before

This PR was a bit trickier than the other "transitional" PRs I've done because in order to provide something that works by only changing the data layer is more difficult because in v2 a missing `meshServices.mode` means `Disabled`, whereas in v3 a missing meshServices.mode would mean the equivalent of `Exclusive`, so I don't think there's a way to reliably do this only in the data layer.

The different way I've done in this PR checks a potentially empty/non-existent field/prop every single time, instead of starting with a good default (like we did previous to this PR), which side-steps the problem of our original v2 default being the opposite of what we want in v3.

---

~Note: I haven't checked this as well as others because I'd need the other PRs in and rebase here before I can see there are no other Mesh related TS errors. I'll take this out of Draft once the other mesh related oneis in~


johncowen:
Just waiting for some more PRs to go in before picking this back up (it crosses over a little with other work). Guessing the test fails are because of test for when we have MeshServices disabled, I could be wrong but I hope its that. Will check when I pick this back up

packages/kuma-gui/features/mesh/Warnings.feature
johncowen, line 1: This file and all of the above removals would need to be maintained or moved elsewhere if we are supporting v2. I tried to pick out a remove the bits that are no longer relevant for v3

packages/kuma-http-api/mocks/src/meshes/_.ts
```diff
@@ -40,38 +37,6 @@ export default ({ fake, env }: Dependencies): ResponseHandler => (req) => {
...
-                },
-              },
-            },
-            {
-              name: 'ca-2',
-              type: 'BUILTIN',
-            },
-          ],
-        },
-      }),
```
johncowen, line 74: These are no longer in v3 I don't think, if we were to maintain support for v2 these would somehow need to be moved elsewhere

schogges (approved):
LGTM 👍

PR #5276, opened 2026-09-04
chore(gateways,legacy-services): remove

schogges:
Removes `gateways` and `legacy-services`, related code, types and mocks.


johncowen (approved):
amazing 👍

PR #5277, opened 2026-09-04
chore(kuma-http-api): remove leftover path overlays

schogges:
In https://github.com/kumahq/kuma-gui/pull/5276 I somehow missed to remove the overlay references for some paths. I also found there is this reference of `Zone` schema which doesn't exist (anymore).

### DIFF
--- a/packages/kuma-http-api/src/openapi.overlay.yaml
+++ b/packages/kuma-http-api/src/openapi.overlay.yaml
@@ -15,15 +15,6 @@ actions:
       /mesh-insights/{name}:
         $ref: "paths/mesh-insights_{name}.yaml"
 
-      /meshes/{mesh}/service-insights:
-        $ref: "paths/meshes_{mesh}_service-insights.yaml"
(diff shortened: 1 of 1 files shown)


packages/kuma-http-api/src/openapi.overlay.yaml
```diff
@@ -43,8 +34,6 @@ actions:
     update:
       Policy:
         $ref: "components/schemas/Policy.yaml"
-      Zone:
-        $ref: "components/schemas/Zone.yaml"
```
johncowen, line 47: How come we can delete this? Is it in the upstream OAS now? If so do we need to delete `components/schemas/Zone.yaml` also?

johncowen: Oh wait 🤔 👀

johncowen: I think I had some files hanging around from when I've been checking the v3 types, ignore!

johncowen (approved):
👍

PR #5279, opened 2026-09-07
feat(legacy-data-planes): decouple and remove

schogges:
Decouples the `legacy-data-planes` module and removes it. Additionally removes the `legacy-data-planes` related e2e tests.

For now I've left the data-plane mocks in place. I'm planning to go through them separately and remove what's not needed anymore.


packages/kuma-gui/src/app/data-planes/index.ts
```diff
@@ -17,31 +19,36 @@ export type DataPlaneServiceLink = (item: { params: Record<string, string> & { s
...
+      arguments: [
+        $.rootViewRoute,
+      ],
+      labels: [
+        app.routeWalkers,
+      ],
+    }],
+    [$.rootViewRoute, {
+      service: () => routes().item,
+    }],
```
schogges, line 51: Allows to override the `RootView` if needed. The `DataPlaneRouteGuard` has been removed as it's not required anymore.

johncowen: If you wanted to, I'm pretty sure you don't even need to expose a token here, the old l`egacy-data-planes` route walker overwrites the same routeNames, so as long as its installed in the correct order (`legacy-data-planes` last) then it should overwrite whatever `data-planes` adds.

We can also do this as a follow up - I think there are gonna be quite a few little follow ups for the v2 removal stuff, I just spotted a meshService one also.

johncowen (approved):

packages/kuma-gui/src/app/data-planes/index.ts
schogges: Ah good point 👍 I will check it separately, thank you 🙂

PR #5280, opened 2026-09-07
chore(dataplanes): remove gateways from the detail page

johncowen:
As gateways never did/don't use the newer DataplanesDetailView (we don't have gateways anymore) I spotted we can remove them from the DataplanesDetailView page.

This should leave us with less to clean up regarding `gateways` now, only the dp listing/summary views


schogges (approved):
👍

PR #5282, opened 2026-09-08
fix(mocks): fixes up some types for mocks for v3

johncowen:
Amends some mocks for new v3 types

### DIFF
--- a/packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_layout.ts
+++ b/packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_layout.ts
@@ -4,7 +4,8 @@ import type { components } from '@kumahq/kuma-http-api'
 type DataplaneNetworkingLayout = components['schemas']['DataplaneNetworkingLayout']
 type DataplaneInbound = components['schemas']['DataplaneInbound']
 type DataplaneOutbound = components['schemas']['DataplaneOutbound']
-type DataplaneListener = components['schemas']['DataplaneListener']
+// @TODO(types) rmeove the clusters form here once we are on v3 as its already in the spec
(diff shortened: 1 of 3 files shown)


packages/kuma-http-api/mocks/src/meshes/_/dataplanes/_/_layout.ts
```diff
@@ -125,6 +109,25 @@ export default ({ env, fake }: Dependencies): ResponseHandler => (req) => {
...
+                zone: fake.word.noun(),
+                namespace: nspace,
+                name: fake.word.noun(),
+                sectionName: String(port),
+              }),
+              port,
+              protocol: fake.kuma.protocol(),
+              proxyResourceName: fake.kuma.contextualKri({ context: `${type.toLowerCase()}_dp`, name: String(port) }),
+            }
+          }),
```
johncowen, line 131: This is a new bit, we don't use this as yet, so I just filled it out enough to satisfy the types

packages/kuma-http-api/mocks/src/meshes/_/meshaccesslogs/_.ts
```diff
@@ -41,34 +41,13 @@ export default ({ fake, env }: Dependencies): ResponseHandler => (req) => {
       },
       kri: fake.kuma.kri({ shortName, mesh, zone, namespace: nspace, displayName }),
       spec: {
-        from: [
+        to: [
```
johncowen, line 44: We don't use the contents of the `spec` anywhere that I know of (apart from just displaying it) so I just removed the old `from` and turned it into a `to`

johncowen (commented):
Notes:

schogges (approved):

PR #5298, opened 2026-09-10
fix(meshes): keep mtls between v2 and v3 temporarily

johncowen:
Temporarily adds `Mesh.mtls` back if its set in the response.


schogges (approved):
👍

PR #5301, opened 2026-09-11
feat(policies): remove types and list views

### LINKED ISSUE #5220: [v3/policies]: Remove policies tab
### Description

In v3 we want to remove the policies tab from a mesh and only keep the resources tab.

Related to https://github.com/kumahq/kuma-gui/issues/4683

schogges:
Removes `policy-type-list-view` and `policy-list-view` and instead re-routes navigations to `mesh-resource-list-view`. Also removes related sources, route configs and tests.

Closes https://github.com/kumahq/kuma-gui/issues/5220


packages/kuma-gui/src/app/data-planes/views/DataPlaneOutboundSummaryOverviewView.vue
```diff
@@ -92,102 +92,108 @@
...
+                :errors="[resourcesError, policiesError]"
+                v-slot="{ data: [resourcesData, policiesData] }"
               >
-                <template
-                  v-for="policyTypes in [Object.groupBy((policyTypesData.policyTypes), ({ name }) => name)]"
-                  :key="`${typeof policyTypes}`"
+                <DataCollection
+                  :items="resourcesData.resources"
+                  :predicate="(resource) => typeof resource.policy !== 'undefined'"
+                  v-slot="{ items: policies }"
```
schogges, line 102: I have removed the `/policy-types` source from policy sources. Here and in some other places I've replaced it with `/resource-type-descriptor` and `DataCollection` which filters for `.policy` to be present.

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
```diff
@@ -1,143 +1,170 @@
...
+                          name: 'policy-list-view',
+                          params: {
+                            mesh: props.mesh.name,
+                            policyPath: policy.path,
+                          },
+                        } : {
+                          name: 'mesh-resource-list-view',
+                          params: {
+                            mesh: props.mesh.name,
+                            shortName: policy.shortName,
```
schogges, line 62: Instead of injecting a different component I thought it's a little easier to do this little fork here and in a few other places. Checks wether there is a `policy-list-view` which means everything would work as is, but if not then route to `mesh-resource-list-view` with the `shortName`.

packages/kuma-gui/src/app/data-planes/views/DataPlaneOutboundSummaryOverviewView.vue
johncowen: This sounds awesome, I was hoping this would be done all at the same time 👍

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
```diff
@@ -1,143 +1,170 @@
...
-              variant="separated"
+          <XTimespan
+            :start="props.mesh.creationTime"
+            :end="props.mesh.modificationTime"
+          />
+          <DataCollection
+            :items="resourcesData.resources"
+            :predicate="(resource) => ['MeshTrafficPermission', 'MeshMetric', 'MeshAccessLog', 'MeshTrace'].includes(resource.name)"
+            v-slot="{ items: policyTypes }"
+          >
```
johncowen, line 29: DataCollection by default renders an empty state if the resulting collection is empty. I know its highly unlikely, but if a Kuma user doesn't install any of these resources (or better said disables all 4) then this would produce a rendered empty state. I actually think this is never going to happen, but theoretically its possible.

I actually like that you reached for DataCollection for these things as it read nicely. But noticed here and in a few other places that we might not want a "theoretical empty state" and thought I'd mention in case I misunderstood that an empty state would never happen.

FYI: We can use `empty=false` I think to disable this auto-empty state. If we are totally 100% sure we'll never get an empty collection though its fine to leave out.

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
johncowen: This feels like we should be using our KRI links which don't render as a link if the route doesn't exist (and we can control from The Outside what a certain KRI/shortName links to). I'm just doing an initial review but I might come back here in a bit to double check, I could be wrong.

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
```diff
@@ -1,143 +1,170 @@
 <template>
-  <XI18n
-    v-slot="{ t }"
+  <DataSource
+    :src="uri(resourceSources, '/resource-type-descriptors', {})"
+    v-slot="{ result: sourceResourcesResult }"
   >
```
johncowen, line 5: How come we are loading data in this component now? Looks like previously `policies` was passed in? At the very least, do we still use the `policies` prop that is passed in?

packages/kuma-gui/src/app/policies/views/PolicyDetailTabsView.vue
```diff
@@ -8,122 +8,147 @@
...
+                </DataLoader>
+              </template>
+              <template
+                #actions
+              >
+                <PolicyActionGroup
+                  :item="data"
+                  :type="{ path: policyType?.path }"
+                  @change="() => route.replace(
+                    router.hasRoute('policy-list-view') ? {
```
johncowen, line 85: I don't like using `router.hasRoute` so deep in the application here. Can we think of a way to abstract/control this from The Outside.

Off the top of my head, `can('list policies')` with the router being injected into the `features.ts` file would be ok at a pinch. Longer term there might be a nicer way.

🤔 actually did I make `useHref` accept KRIs I think I did.

`href('kri://policy-kri')` could give you the URL of policy-list-view if policies are enabled, but the mesh-resource-list-view if they are not.

Would be good to quickly try both of these options. If neither fit, then we can come back later. Obvs I would prefer the KRI option, but `can` would be a second best at a pinch.

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
johncowen: To keep the convo in the same place, see convo below on this one. One extra thing I realized after I wrote my comment is, these are listing URLs not detail URLs.

packages/kuma-gui/src/app/policies/views/PolicyDetailTabsView.vue
johncowen: Extra note:  I realized over the weekend that these are list endpoints and whilst we chatted about adding list URL support for KRI based links, I don't think we ever did. I'd still really like to do that as it avoid all these in-app forks.

If adding list KRI support here is out of scope for this PR, I'd like to come back and do it. If that is the case then I guess I'm fine with whatever here for the minute.

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
schogges: Ah! Thanks for pointing at this 👍

schogges: I've replaced new `DataCollection`s with `template v-for` filtering

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
schogges: We need to load the `resource-type-descriptors` now to get the `path` and `shortName` for proper linking. Previously the param was set as `:policy-path="`${policy.toLowerCase()}s`"`

packages/kuma-gui/src/app/policies/views/PolicyDetailTabsView.vue
schogges: Yeah that was my thinking also. I don't really want to use a KRI to link to a listing view 🤔 so I think for now I'd just stick to this.

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
johncowen: Are these to link the policies that we filter for by hardcoded names?

packages/kuma-gui/src/app/policies/views/PolicyDetailTabsView.vue
johncowen: FYI: I am still 100% keen on using KRIs everywhere (including listings),  even more so now.

Lets get this in for now, there is a tonne of KRI work I've put in the "circle back" category.

johncowen: Resolving this one

johncowen (approved):
Lets go with this for the moment. I'd like to circle back on the "forks for links" thing. But we can cover that at a later date

packages/kuma-gui/src/app/meshes/components/MeshStatus.vue
schogges: Yeah 🙂 but I didn't want to hardcode paths and shortNames

PR #5305, opened 2026-09-14
feat(resource-type-list-view): fall back to 0 values in case of insight error

schogges:
In case of an error on `/mesh-insight/:name` or `/global-insight` we don't want to show an error on the `ResourceTypeListView`, but instead show an error where we normally would display the count.

While I also double checked the same behaviour on the `PolicyTypesListView`, I've update the usage of `DataLoader` and replaced `data!.` with the slot prop from `DataLoader`.


packages/kuma-gui/src/app/resources/views/ResourceTypeListView.vue
```diff
@@ -64,10 +64,11 @@
                     </template>
                   </XLayout>
                 </template>
+                <!-- block on resource-type-descriptors types but not *Insight -->
                 <DataLoader
-                  :data="[sourceResources, sourceGlobalInsight, sourceMeshInsight]"
-                  :errors="[resourcesError, globalInsightError, meshInsightError]"
-                  v-slot="{ data: [resources, globalInsight, meshInsight] }"
+                  :data="[sourceResources]"
```
johncowen, line 69: If I follow right here that we don't show any sort of loader for the insights? I vaguely remember in my other PR I added an additional loader into where we loop each policy and added a `count` variant to the loader to show a small skeleton where the number would appear. Could we take a quick look and see if we can add that also?

schogges: I've picked the changes you've made in #5123 and did an extra bit to show a warning icon with a tooltip as an error state.

johncowen (approved):
👍

PR #5308, opened 2026-09-15
chore(kuma): make KRI handlers modular

### LINKED ISSUE #5264: Make KRI handler to be able to inject multiple KRIs
### Description

This feature is kinda of half of what we rolled back here https://github.com/kumahq/kuma-gui/pull/5249

So similar type of approach but instead of having a `useDataPlaneServiceLinks` we inject different "KRI handlers" into "The Big Switch Statement" here:

https://github.com/kumahq/kuma-gui/blob/b6f137cb5b8bd809cd31038c3f9f681e3438c99c/packages/kuma-gui/src/app/kuma/index.ts#L61-L67

The should belong to the `kuma` module, and the `kuma` module should expose a DI label for other kuma related modules to use for injection. In turn the `kuma` module is then responsible for adding the `kri://` protocolHandler (which it already does)

johncowen:
Makes "KRI handlers" modular so we can install (or not) them per module.

Fallback will fallback to the generic Resource Detail View page is there is no handler associated with a KRI

Closes https://github.com/kumahq/kuma-gui/issues/5264


packages/kuma-gui/src/app/kuma/index.ts
```diff
@@ -229,14 +90,84 @@ export const TOKENS = {
...
+              const item = Kri.fromString(href.substring(kriProto.length))
+              const kri = Kri.toString({
+                ...item,
+                // old style names can have _ in them that are replaced with `~`
+                name: item.name.replaceAll('~', '_'),
+              })
+              const args = {
+                ...Kri.fromString(kri),
+                kri,
+              }
```
schogges: I think `kri` could be wrong here:

```ts
Kri.toString({ ...item, name: 'my~old~style~name'.replaceAll('~', '_') }) // => 'kri_dp_default___my_old_style_name_'
```

johncowen: I do think there is a little issue here, but I'm not sure about your example, could you explain a little more? Are you saying a user might have used `my~old~style~name` for a string? Or something else?

schogges: > I do think there is a little issue here, but I'm not sure about your example, could you explain a little more? Are you saying a user might have used `my~old~style~name` for a string? Or something else?

That's what the comment is saying and essentially `item.name.replaceAll('~', '_')` is doing to replace the `~` in a name?

> // old style names can have _ in them that are replaced with `~`

schogges: Ah sorry, I think we need to turn around `item.name.replaceAll('~', '_')` to `item.name.replaceAll('_', '~')`?

schogges: Do we want

```typescript
Kri.toString({ ...item, name: 'my_old_style_name'.replaceAll('_', '~') }) // => 'kri_dp_default___my~old~style~name_'
```

johncowen: So in the application we replace any `_` in old style names with `~` so that the name doesn't get split into several chunks when we split the KRI:

```
...old_style_name.replace('_', `~`) --> old~style~name --> KRI.fromString(`...old~style~name`) -> old~style~name.replaceAll('~', '_')
```

As old style names should never have `~` in them (as its not DNS compatible) this is should be a safe encoding method.

Does that ^ answer what you were concerned about? If it doesn't maybe I misunderstood. If my terrible pseudo code doesn't help 😅 lmk and we can leave this til we sync

schogges: I think we should do the `.replaceAll('~', '_')` after we've parsed the KRI and when passing it to `args`. Otherwise we could mess up the KRI.

```typescript
const kri = Kri.toString(item)
const { name, ...rest } = Kri.toString(kri)
const args = {
  ...rest,
  name: name.replaceAll('~', '_'),
  kri,
}
```

johncowen: Ah now this bit is the bit I think is wrong 1 sec, lemme push this commit.

johncowen: This (https://github.com/kumahq/kuma-gui/pull/5308/changes/10a9039aea40ac5749363ddd11aedfa30b5749b9) is better right?

schogges: It's better, but still could lead to an invalid KRI further downstream:

```typescript
const item = Kri.fromString(href.substring(kriProto.length))
// old style names can have _ in them that are replaced with `~`
// name = 'my~old~style~name'
const name = item.name.replaceAll('~', '_') // name = 'my_old_style_name'
const kri = Kri.toString({
  ...item,
  name,
}) // kri = 'kri_dp_default___my_old_style_name_'
const args = {
  ...item,
  name, // 'my_old_style_name' is correct
  kri, // 'kri_dp_default___my_old_style_name_' is an invalid KRI
}
```

johncowen: Why is this invalid:

> kri, // 'kri_dp_default___my_old_style_name_' is an invalid KRI

?

schogges: Because a KRI has 7 sections separated by `_`. `'kri_dp_default___my_old_style_name_'` is invalid because it now has 10 sections as each `my`, `old`, `style` and `name` would be interpreted as another section. If we'd try to parse it further downstream it would cause an "empty" KRI because of https://github.com/kumahq/kuma-gui/blob/master/packages/kuma-gui/src/app/kuma/kri.ts#L9-L18. Also I assume the backend could not interpret it either 🤔

schogges: I'm wondering how this is even possible:
>// old style names can have _ in them that are replaced with `~`

johncowen: Ok I've removed the decoding here now, I realized we don't need to decode here anymore as we only supporting v3 here. I think this is now ready for another review. Thanks for nudging on this one 👍

packages/kuma-gui/src/app/hostname-generators/index.ts
```diff
@@ -10,6 +10,26 @@ type Token = ReturnType<typeof token>
...
   return [
+    [token('hostname-generators.kri'), {
+      service: () => {
+        return [
+          ({ shortName, mesh, kri }: { shortName: string, mesh: string, kri: string }) => {
+            if(shortName === 'hg') {
+              return {
+                name: 'hostname-generator-detail-view',
+                params: {
+                  mesh,
```
schogges: There should be no `mesh` param for `hostname-generator-detail-view`

packages/kuma-gui/src/app/workloads/index.ts
```diff
@@ -11,6 +11,25 @@ type Token = ReturnType<typeof token>
...
   return [
+    [token('workloads.kri'), {
+      service: () => {
+        return [
+          ({ shortName, kri }: { shortName: string, kri: string }) => {
+            if(shortName === 'wl') {
+              return {
+                name: 'workload-detail-view',
+                params: {
+                  zone: kri,
```
schogges: `workload-detail-view` needs params `mesh` and `wl` (`kri`)

schogges (commented):
Found two cases of wrong params, the rest LGTM

packages/kuma-gui/src/app/hostname-generators/index.ts
johncowen: Done!

packages/kuma-gui/src/app/workloads/index.ts
johncowen: Done!

johncowen:
Whoops! Thanks 👍 I think this is ready now then

schogges (dismissed):
👍

johncowen:
I wanted to make a little additional tweak here which was to move this stuff:

```yaml
common:
  label:
    href:
      kuma~io/mesh: kri://kri_m____{name}_
      kuma~io/zone: kri://kri_z____{name}_
      kuma~io/workload: kri://kri_wl_{mesh}_{zone}_{namespace}_{name}_
```

TO go into each module instead of all in `kuma`

johncowen:
@schogges just a quick note here, the small change I commented about has now been done, sorry for the post approval tweak!

schogges (approved):

PR #5314, opened 2026-09-16
chore(oas): remove some old, now unnecessary oas overlays

johncowen:
Some of our old overlays are no longer necessary as they have been added upstream. This removes the majority of them.

There is still 1 path and a few components left that are related to legacy-policy support. I'll see if I can remove these in an upcoming PR as at first glance it looks to be a little more involved than just removing.

I also noticed that the upstream types are still al little wrong. For the moment I fixed these up in our datalayer/mocks with an Omit. Once these are fixed we can remove these.


schogges (approved):
👍

PR #5329, opened 2026-09-18
feat(resources): handle legacy policy flags

### LINKED ISSUE #5309: Stop reading removed PolicyDescription fields from GET /_resources
## Context

Since the 3.0 cleanup in `kumahq/kuma`, every policy is targetRef based and the `from`-style targetRef is gone. The `policy` object returned by `GET /_resources` reflects that:

- `policy.hasFromTargetRef` and `policy.isFromAsRules` were hardcoded to `false` and are now removed from the response (kumahq/kuma PR linked below).
- `policy.isTargetRef` is now `true` for every policy. It is kept on the wire for now only because the GUI still reads it, and will be removed in a follow-up once the GUI no longer depends on it.
- `policy.hasToTargetRef` and `policy.hasRulesTargetRef` are the real signals for "applies to outbound" and "applies to inbound".

## What the GUI does today

- `packages/kuma-gui/src/app/resources/views/ResourceListView.vue` shows the **Inbound** badge from `type.policy?.hasFromTargetRef`. That badge has never rendered against a 3.0 control plane because the value was always `false`.
- `ResourceListView.vue` and `packages/kuma-gui/src/app/data-planes/views/DataPlanePolicySummaryView.vue` read `policy.isTargetRef` to decide whether to show the zone column / search key and whether to render a policy as `legacy`. Once the field is removed from the response, `undefined` is falsy and every policy would be treated as legacy.
- `packages/kuma-http-api/mocks/src/_resources.ts` and `mocks/src/policies.ts` still generate the removed fields.

## Requested changes

1. Show the **Inbound** badge from `policy.hasRulesTargetRef` instead of `policy.hasFromTargetRef`.
2. Stop reading `policy.isTargetRef`. Every policy is targetRef based, so the `legacy` rendering path and the `isTargetRef` conditions on the zone column and search keys can go.
3. Drop `hasFromTargetRef`, `isFromAsRules` and `isTargetRef` from the `_resources` and `policies` mocks.

After this lands and the GUI is bumped in kuma, `policy.isTargetRef` will be removed from the control plane response as well.

## References

- Removal PR in kuma: https://github.com/kumahq/kuma/pull/18733

schogges:
Backfills some resource policy flags that are about to be removed.
While looking into this I found that we still have `policies/data/PolicyResourceType`, which was only used for the types. I've removed this file and replaced the types with `ResourceTypeDescriptor` from `resources`.

Closes https://github.com/kumahq/kuma-gui/issues/5309


packages/kuma-gui/src/app/resources/data/ResourceTypeDescriptor.ts
```diff
@@ -51,6 +51,16 @@ export const ResourceTypeDescriptor = {
     return {
       ...item,
       group,
+      ...(typeof item.policy !== 'undefined' && {
+        policy: {
+          ...item.policy,
+          // Backfills some legacy flags that are gone from OAS, but we keep reading them in the views.
+          isTargetRef: 'isTargetRef' in item.policy ? item.policy.isTargetRef as boolean : true,
+          // @ts-expect-error - missing item.policy.hasRulesTargetRef in OAS, refactor once it's there
+          hasFromTargetRef: 'hasFromTargetRef' in item.policy ? item.policy.hasFromTargetRef as boolean : (item.policy.hasRulesTargetRef ?? true),
```
johncowen: Could you explain this one to me again? I always get confused with all these `isHasFromToRules` things

Did I understand right that `hasRulesTargetRef` is in v3 but is missing from the OAS?

schogges: Thanks for asking 👍 that revealed a misunderstanding on my side, brb
`hasRulesTargetRef` should already be there...

schogges: I've reworded the directive

schogges: Ok, I've changed it again as this was a little weird and I've already added to consider `hasRulesTargetRef` when evaluating the inbound badge in
[`packages/kuma-gui/src/app/resources/views/ResourceListView.vue`](https://github.com/schogges/kuma-gui/blob/37c9e403f0676e57afce9df929ec8a2d19153168/packages/kuma-gui/src/app/resources/views/ResourceListView.vue#L31)

packages/kuma-gui/src/app/resources/data/ResourceTypeDescriptor.ts
```diff
@@ -51,6 +51,15 @@ export const ResourceTypeDescriptor = {
     return {
       ...item,
       group,
+      ...(typeof item.policy !== 'undefined' && {
+        policy: {
+          ...item.policy,
+          // Backfills some legacy flags that are gone from OAS, but we keep reading them in the views.
+          isTargetRef: 'isTargetRef' in item.policy ? item.policy.isTargetRef as boolean : true,
+          hasFromTargetRef: 'hasFromTargetRef' in item.policy ? item.policy.hasFromTargetRef as boolean : true,
+          isFromAsRules: 'isFromAsRules' in item.policy ? item.policy.isFromAsRules as boolean : true,
```
schogges, line 60: @lobkovilya could you give us another pair of 👀 on this? Especially if the fallbacks to `true` make sense?

johncowen: I think `hasFromTargetRef` and possibly even `isFromAsRules` might need to default to `false` for v3. Let's definitely get confirmation from @lobkovilya tho 👍

PR #5334, opened 2026-09-21
feat(data-planes): remove check for unified-resource-naming

### LINKED ISSUE #5299: data-planes: Remove checks for unified-resource-naming
### Description

In v3 the feature flag around `unified-resource-naming` is being removed and this becomes the default. There is still one check for this: https://github.com/kumahq/kuma-gui/blob/0bfc96e35e20dc75feb806b6fa2f9bf6b41b273b/packages/kuma-gui/src/app/data-planes/views/DataPlaneDetailView.vue#L477

schogges:
As this becomes the default in v3 and there is only one `dataplaneType` (we have removed `zone-ingresses` and `zone-egresses`), we don't need to check anymore.

Closes https://github.com/kumahq/kuma-gui/issues/5299


johncowen (approved):
👍

PR #5336, opened 2026-09-21
feat(zones): only show incompatible version warning with subscriptions

### LINKED ISSUE #5327: Hide or improve version mismatch notification when a zone has never connect
### Kuma Version

3.0.0

### Describe the bug

> When connecting a zone for the first time, there is a notification that says theres a version mismatch. But, its not that there is a version mismatch, its that we've never connected to so don't know any details. We should work on this to improve it, we can probably figure out from subscriptions if we have ever connected and show something different (or not show a notification at all until its connected).

### To Reproduce

_No response_

### Expected behavior

_No response_

_No response_

schogges:
Only show the warning about incompatible versions between global and zone CP if there have been any connections (`zoneInsight.subscriptions`).

Closes https://github.com/kumahq/kuma-gui/issues/5327


johncowen (approved):
Nice Mo! 👍

PR #5342, opened 2026-09-22
chore(meshzoneaddress): add mocks for MeshZoneAddress

johncowen:
Adds mocks for MeshZoneAddress

---

Entirely LLM generated using our skill


packages/kuma-http-api/mocks/src/meshes/_/meshzoneaddresses/_.ts
```diff
@@ -0,0 +1,63 @@
...
+              name: displayName,
+              mesh,
+              env: k8s ? 'kubernetes' : 'universal',
+              ...(zone ? { zone } : {}),
+              ...(k8s ? { namespace: nspace } : {}),
+            }),
+          },
+        }
+        return k8sFormat ? { metadata } : metadata
+      })()),
```
johncowen, line 56: Just wondering if we even need to do this metadata thing for `k8s`, it overly complicates our mock and its only for the "Display as K8s" stuff.

Not a blocker here but we might want to remove this codebase wide so that the LLM doesn't copy it if we decide that we don't need it.

johncowen (commented):
I gave this a review myself and it LGTM I think, I left one comment for the future.

packages/kuma-http-api/mocks/src/meshes/_/meshzoneaddresses/_.ts
schogges: > Just wondering if we even need to do this metadata thing for `k8s`, it overly complicates our mock and its only for the "Display as K8s" stuff.
> 
> Not a blocker here but we might want to remove this codebase wide so that the LLM doesn't copy it if we decide that we don't need it.

This ^ or we might be able to somehow abstract it away to FakeKuma 🤔

johncowen: > This ^ or we might be able to somehow abstract it away to FakeKuma 🤔

Yeah this was also my initial thought, but then I thought "we don't actually need this", but yeah we can discuss

johncowen: Personally I'm more inclined to just not do this, I think I added it ages ago and I think I was going a little overboard in trying to make things look like K8s, but we only display K8s config we don't actually use it as data, so as long as something changes when we click to display `[Kubernetes]` (in this case the `apiVersion` at the top) then I don't think we need this.

If we were to make a helper you'd need to pass in multiple things, labels, mesh, name, whether its k8s or not etc. and its for something we don't need to use.

schogges (approved):

PR #5347, opened 2026-09-22
chore(meshes): add Zones tab to see Zones per mesh 

johncowen:
TBD

Needs rebasing onto https://github.com/kumahq/kuma-gui/pull/5342


johncowen:
Not taking this out of Draft just yet, I want to try and make this clearer that these aren't "Zones" per se, more "only the Zones that your services and workloads can connect to", possibly `[Connected Zones]`.

PR #5348, opened 2026-09-22
chore(resources): moves control-plane resources into the resource module

johncowen:
Moves the ControlPlane Resources things into `resources` from `control-planes` because they are more related to `resources` not `control-planes`.

Mostly just a file move, but I did some little cleanup while I was here.


schogges (approved):

johncowen:
sorry @schogges I had to rebase here, this is good to go again but needs another approval

schogges (approved):

packages/kuma-gui/src/app/resources/routes.ts
```diff
@@ -33,24 +32,27 @@ export const routes = (prefix: string) => {
...
+          component: () => import('./views/ResourceTypeListView.vue'),
           children: [
             {
               name: `${prefix}-resource-list-view`,
               path: ':shortName',
               props: {
                 routePrefix: prefix,
               },
-              component: () => import('@/app/resources/views/ResourceListView.vue'),
+              component: () => import('./views/ResourceListView.vue'),
```
schogges: Why are we actually changing these to relative imports?

PR #5351, opened 2026-09-23
fix(x-search): styles

schogges:
The colors for highlighted and incorrect values didn't work well with theming (dark&light) elsewhere. Also the width and padding were not properly overridden.


packages/x/src/components/x-search/XSearch.vue
```diff
@@ -305,10 +305,11 @@ watch(() => props.value, () => {
     padding: var(--x-space-10) 0px;
 
     &.highlight {
-      background: var(--x-search-color-token-background);
```
johncowen, line 308: Could we also remove this variable from XTheme.vue if its not being used for anything else. Looks like it was only added because this originally used a custom color that isn't in our design-tokens.

schogges: Done

johncowen (approved):
👍

PR #5352, opened 2026-09-23
chore(kuma): use decorated fetch instead of class based hierarchy

johncowen:
Part 3 of removing KumaApi and friends.

This one actually removes `KumaApi`, `Api` and `RestClient`, finally...

We now just use `fetch` (with a very small wrapper to make it throw), that we pass to our OpenAPI client.

Note: I'm probably going to do something with `ApiError` also, possibly just move it, possibly a tinsy bit more. But I need to think about it a little more.

Related:

- Part 2: https://github.com/kumahq/kuma-gui/pull/5349


schogges (approved):
LGTM

